PR #7558, opened 2021-02-24
Fix parent.config timeouts to return a 504 not 502 on timeout

rob05c:
This fixes requests that used parent.config to return a `504 Gateway Timeout` instead of a `502 Bad Gateway` when `*_timeout_out` is hit.

Requests not using parent.config are already doing this in `HttpTransact::handle_server_died`. This fixes `handle_parent_died` to do the same.

### DIFF
--- a/proxy/http/HttpTransact.cc
+++ b/proxy/http/HttpTransact.cc
@@ -7471,10 +7471,18 @@ HttpTransact::handle_parent_died(State *s)
 {
   ink_assert(s->parent_result.result == PARENT_FAIL);
 
-  if (s->current.state == OUTBOUND_CONGESTION) {
+  switch (s->current.state) {
+  case OUTBOUND_CONGESTION:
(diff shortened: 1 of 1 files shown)


ezelkow1:
Do we need to add any more of the server_died states here? Just wondering if we should knock them all out if applicable

jrushford (approved):
Looks good, I approve.

rob05c:
> Do we need to add any more of the server_died states here? Just wondering if we should knock them all out if applicable

I can. They're all `502 Bad Request`, so the behavior here is correct. I think the only real difference is giving a slightly more informative HTML response.

I was just trying to keep this PR small, but I could add that if we want. Or abstracting it to reduce the duplication between them would be even better.

bryancall (approved):

rob05c:
I think I'd vote to keep this small and backport to 9, and do the larger duplication removal as a separate PR in 10

zwoop:

PR #7537, opened 2021-02-18
Enable origin server session cache by default

duke8253:
As mentioned in #7479 , a separate PR for enabling origin server session cache by default for the 10.0.0 branch.

PR #7479, opened 2021-02-03
Origin session reuse

duke8253:
Allow sessions to be reused when connecting to origin servers.

PR #7559, opened 2021-02-24
Check verify policy and properties

duke8253:
Another bug fix for Origin Session Reuse #7479 

The problem this time is that the verify property and policy can also be different from rule to rule, this adds the policy and property string to the lookup key for sessions. 

Now enabling session reuse for origin servers by default #7537 should not fail any tests.

### DIFF
--- a/iocore/net/SSLUtils.cc
+++ b/iocore/net/SSLUtils.cc
@@ -1997,11 +1997,8 @@ SSLConnect(SSL *ssl)
   if (!sess && SSLConfigParams::origin_session_cache == 1 && SSLConfigParams::origin_session_cache_size > 0) {
     std::string sni_addr = get_sni_addr(ssl);
     if (!sni_addr.empty()) {
-      SSL_CTX *ctx = SSL_get_SSL_CTX(ssl);
-      std::stringstream ctx_ss;
-      ctx_ss << static_cast<const void *>(ctx);
       std::string lookup_key;
-      ts::bwprint(lookup_key, "{}:{}", sni_addr.c_str(), ctx_ss.str().c_str());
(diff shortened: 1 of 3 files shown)


iocore/net/SSLUtils.cc
```diff
@@ -2065,6 +2062,54 @@ get_sni_addr(SSL *ssl)
   return sni_addr;
 }
 
+std::string
+get_verify_str(SSL *ssl)
+{
+  std::string verify_str;
+
+  SSLNetVConnection *netvc = SSLNetVCAccess(ssl);
```
maskit, line 2070: SSLNetVCAccess 😞 I don't see it as a blocker because QUIC connection to origin is probably years away, but we should try to avoid using it for new stuff.

maskit:
> Now enabling session reuse for origin servers by default #7537 should not fail any tests.

Did you run autest locally and confirm it? I think we could avoid making this PR if we did it on the previous fix. I don't want another fix after this.

duke8253:
> > Now enabling session reuse for origin servers by default #7537 should not fail any tests.
> 
> Did you run autest locally and confirm it? I think we could avoid making this PR if we did it on the previous fix. I don't want another fix after this.

Yeah, sorry about that. I ran it locally to confirm it is working with #7537

iocore/net/SSLUtils.cc
duke8253: Yes, we probably could store something in the `ex_data` of the `SSL` object.

maskit (approved):

zwoop:
Moving out to 9.2.x, too many bugs.

PR #7560, opened 2021-02-24
Adding lower_ support to stats and bonding_slave data points for port status

ezelkow1:
Newer kernels have changed the wording from slave_ to lower_ for interfaces as part of a bond, as well as having a transition period between the 2. Adding support for lower_ to be recognized as well.

Also adding ad_actor_oper_port_state and ad_partner_oper_port_state data for each interface from its bonding_slave dir if it exists as this gives information on the actual state of that interface port


bneradt (approved):
Makes sense.

PR #7567, opened 2021-03-01
Remove ProxyTransaction::set_proxy_ssn

maskit:
A ProxySession is set to ProxyTransaction by calling `set_proxy_ssn`. However, the function should not exist because a session for a transaction should not be changed in the middle of a transaction.

This PR adds `ProxySession *session` param to ProxyTransaction's constructor to guarantee a session is always set and it won't be accidentally changed later.

Also, this integrates `Http2Stream::init` into Http2Stream's constructor.

### DIFF
--- a/proxy/ProxyTransaction.cc
+++ b/proxy/ProxyTransaction.cc
@@ -26,7 +26,7 @@
 
 #define HttpTxnDebug(fmt, ...) SsnDebug(this, "http_txn", fmt, __VA_ARGS__)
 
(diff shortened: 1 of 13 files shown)


a-canary (approved):
+1

shinrich (approved):
Looks good. I'll sync up with the H2 to origin branch.

zwoop:
Need this in 9.1.x to support other PRs needed.

PR #7571, opened 2021-03-02
Tidy up session/transaction destruction process

maskit:
The primary goal is to call destructors of ProxySession, ProxyTransaction and their subclasses.

ProxySession
  Made `free()` pure virtual function -- ProxySession does not know how to free itself and current code does not free.
  Added destructor -- The code originally in `free()` is now in the destructor.

ProxyTransaction
  Removed `destroy()` -- I don't see any reasons to have this as a normal function. The existence is confusing.
  Added destructor -- The code originally in `destroy()` is now in the destructor

Http1ClientSession
  Added destructor -- To call ProxySession's destructor automatically. The normal destruction process is easier to follow.
  Changed a function name to call -- See Http1Transaction.

Http1ServerSession
  Added destructor -- Just for consistency.
  Added `free()` -- It does nothing because it's freed in `destroy()` unlike Http1ClientSession.

Http1Transaction
  Renamed `destroy()` to `reset()` -- `Http1Transaction::destroy()` didn't actually destroy itself, but reset its state.

Http2ClientSession
  Added destructor -- To call ProxySession's destructor automatically.

Http2Stream
  Converted `destroy()` to the destructor
  Unified the process to destroy itself

Http3Session
  Added `free()` -- It has never been freed.

### DIFF
--- a/proxy/http/Http1ClientSession.h
+++ b/proxy/http/Http1ClientSession.h
@@ -54,6 +54,7 @@ class Http1ClientSession : public ProxySession
 public:
   typedef ProxySession super; ///< Parent type.
   Http1ClientSession();
+  ~Http1ClientSession() = default;
 
   // Implement ProxySession interface.
   void new_connection(NetVConnection *new_vc, MIOBuffer *iobuf, IOBufferReader *reader) override;
--- a/proxy/ProxySession.cc
+++ b/proxy/ProxySession.cc
@@ -30,6 +30,18 @@ ProxySession::ProxySession() : VConnection(nullptr) {}
 
 ProxySession::ProxySession(NetVConnection *vc) : VConnection(nullptr), _vc(vc) {}
 
+ProxySession::~ProxySession()
+{
+  if (schedule_event) {
+    schedule_event->cancel();
+    schedule_event = nullptr;
+  }
+  this->api_hooks.clear();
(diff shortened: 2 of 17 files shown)


maskit:
[approve ci clang-format]

maskit:
This depends on #7567.

ywkaras:
Looks like you forgot to set the Destruct_of_free parameter to the ClassAllocator instantiations to true.

maskit:
> Looks like you forgot to set the Destruct_of_free parameter to the ClassAllocator instantiations to true.

I know you added the flag, but the PR was not merged yet when I made this change. And I don't think we should use the flag on this PR, because that would unnecessarily introduces a dependency for your change. Do you mind if I make the change separately on another PR? It would allows us to backport or revert our changes separately.

proxy/http/Http1ClientSession.h
```diff
@@ -54,6 +54,7 @@ class Http1ClientSession : public ProxySession
 public:
   typedef ProxySession super; ///< Parent type.
   Http1ClientSession();
+  ~Http1ClientSession() = default;
```
ywkaras, line 57: This isn't necessary.

ywkaras (approved):

zwoop:
Need this for 9.1.x for other PRs to work.

PR #7572, opened 2021-03-02
FQ pacing plugin memory leak fix

ezelkow1:
Add tscontdestroy when transaction is closed and pacing rate is reset


shukitchan (approved):
This looks good.

PR #7575, opened 2021-03-03
add GitHub workflow configuration

bryancall:


.github/workflows/c-cpp.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  build:
+
+    runs-on: macos-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: configure
+      run: brew install automake openssl llvm; autoreconf -if; CC=/usr/local/Cellar/llvm/11.1.0/bin/clang CXX=/usr/local/Cellar/llvm/11.1.0/bin/clang++ ./configure --with-openssl=/usr/local/Cellar/openssl\@1.1/1.1.1j
```
maskit, line 17: I don't think I made this link by myself.
```
$ ll /usr/local/opt/openssl@1.1
lrwxr-xr-x  1 mkitajo  admin  28 Mar  3 13:29 /usr/local/opt/openssl@1.1 -> ../Cellar/openssl@1.1/1.1.1j
```

jrushford: Are you going to build ATS with this workflow @bryancall?  I have this workflow in apache trafficcontrol that builds currently, ATS 8.1, see the trafficserver job in the workflow.  Just thought you might be interested.  The workflow is used to run some integration tests in trafficcontrol that require an ATS RPM.  The workflow checks the 8.1 branch for new commits and build's an RPM if there are otherwise it uses the cached RPM from a previous run.  Just thought you might be interested.  I think the GitHub Actions can be used for all our trafficserver CI.

PR #7577, opened 2021-03-04
Fix the final consumer write size from unchunked to chunked tunnel

shinrich:
Another issue I ran into with the H2 to origin work.  This case may not occur outside of that.

Client is Http/1.1.  Server is Http/2.  The server returns a header with no content length set, so the data is coming back in a series of data chunks.  Thus, it must be returned to the client chunked.  The response tunnel is set up for do_chunking.  That all works well, but the final call on the consumer side of the tunnel sets the total number of bytes read incorrectly. It only sets the number of data bytes.  It does not include the header bytes which is tracked by p->chunked_handler.skip_bytes.

In my case, only 5 bytes were being returned to the client. From the client's perspective, it looked like a stall.

### DIFF
--- a/proxy/http/HttpTunnel.cc
+++ b/proxy/http/HttpTunnel.cc
@@ -1010,7 +1010,7 @@ HttpTunnel::producer_handler_dechunked(int event, HttpTunnelProducer *p)
       HttpTunnelConsumer *c;
       for (c = p->consumer_list.head; c; c = c->link.next) {
         if (c->alive) {
-          c->write_vio->nbytes = p->chunked_handler.chunked_size;
+          c->write_vio->nbytes = p->chunked_handler.chunked_size + p->chunked_handler.skip_bytes;
           // consumer_handler(VC_EVENT_WRITE_COMPLETE, c);
         }
       }
--- a/proxy/http/HttpTunnel.cc
+++ b/proxy/http/HttpTunnel.cc
@@ -1001,10 +1001,12 @@ HttpTunnel::producer_handler_dechunked(int event, HttpTunnelProducer *p)
 
   // We only interested in translating certain events
   switch (event) {
-  case VC_EVENT_READ_READY:
   case VC_EVENT_READ_COMPLETE:
   case HTTP_TUNNEL_EVENT_PRECOMPLETE:
   case VC_EVENT_EOS:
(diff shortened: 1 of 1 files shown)


bryancall:
@masaori335 Can you please look at this?

masaori335 (approved):
This change looks reasonable. Because the header is written into the `chunked_buffer` (which is the buffer of this `c->write_vio`) at the beginning of the tunnel. 
https://github.com/apache/trafficserver/blob/f4f4f59237018b91824f9130d57539e46c1def30/proxy/http/HttpTunnel.cc#L782

masaori335:
Do you see the same issue with the HTTP/1.0 response without the Content-Length header? Which should go through this chunking code.

masaori335:
This change seems reasonable but I don't see the stall with HTTP/1.0 response. It looks like `c->write_vio->nbytes` is finally adjusted at `HttpTunnel::finish_all_internal()`. Does your H2 to origin come here?

https://github.com/apache/trafficserver/blob/f4f4f59237018b91824f9130d57539e46c1def30/proxy/http/HttpTunnel.cc#L1512-L1514

shinrich:
I assume that we were hitting that logic in finish_all_internal, because the path I was testing was the PRECOMPLETE path.  Assuming the client handler was being executed on the step after the reenable, the client's would be not alive by the time finsh_all_internal was called.

In any case, the logic for calculating the final size for chunking and dechunking should be the same for both.  I'll look at pulling that out into a common function called both places.

shinrich:
Actually, I am just going to update the producer_handler_dechunked to call final_consumer_writes_to_byte too.  If there is a cached consumer, this change isn't write in any case..

shinrich:
[approve ci autest]

shinrich:
[approve ci autest]

masaori335 (approved):

shinrich:
Asking for 9.1.x to ease testing of the h2 to origin branch against 9.1.x

zwoop:

bneradt:
I'm adding this to the 9.0.x project because a user experienced this issue there: #7880 . I've verified locally that it cherry-picks cleanly to that branch.

zwoop:

zwoop:

PR #7579, opened 2021-03-04
Clean up /r/n between transactions on Http1Client

shinrich:
Another issue seen while debugging H2 to origin.  I'm not entirely sure this should be pushed to master.  If it should be, I can add an autest to exercise the issue.

Was tracking down spurious ERR_INVALID_REQ failures being recorded in our access logs.  Turns out some of our health checking scripts were adding and extra /r/n or just /n at the end of the health checking GET request.  ATS would try to interpret the extra white space and fail resulting in the ERR_INVALID_REQ log entry.

For scripting languages like Perl and Python, it is pretty easy to accidentally slap on extra whitespace.  I don't think our scripts send multiple requests on the same connection, so this cleanup is just a matter of keeping confusing cruft out of the logs.


shinrich:
Group consensus.  Should just fix the bad clients.

PR #7580, opened 2021-03-05
 Refine connection failure logging and messages and eliminate spurious connection errors

shinrich:
These are changes I made working with @djcarlin to clean up our error messages on connection failures.  These changes are in a build that we are rolling out.  Currently running in a couple pods.

The problem was that messages in errors.log about connection closing were confusing and in some cases missing information. One issue is that the HttpSM had two pieces of information that were tracking the reason for a connection failure (t_state.current.server->connect_result and t_state.cause_of_death_errno), and they weren't always being set consistently. This PR uses an updated version of set_connect_fail() to set them both consistently.

The PR also changes the logic of setting those values.  The original would set EIO if nothing else was set.  The PR sets EIO and clears it on success.  Later calls to set_connect_fail will override the EIO..

The PR also attempts to make the SQUID log codes more detailed in the case of connection failure.

Finally, the PR adjusts the logs to error.log. There was one failure path that was not logging, so I added HttpTransact::error_log_connection_failure. and called that from both locations.  Also added an error.log entry for the DNS failure case.

I see that there are some changes in HttpTunnel and Http2Stream which seem unrelated.  I just wrote quite a lot here, so I'll go ahead and set up the PR and push up another commit to eliminate those changes.

### DIFF
--- a/proxy/http/HttpSM.cc
+++ b/proxy/http/HttpSM.cc
@@ -1222,10 +1222,12 @@ HttpSM::state_raw_http_server_open(int event, void *data)
 
     netvc->set_inactivity_timeout(get_server_inactivity_timeout());
     netvc->set_active_timeout(get_server_active_timeout());
+    t_state.current.server->clear_connect_fail();
     break;
 
   case VC_EVENT_ERROR:
   case NET_EVENT_OPEN_FAILED:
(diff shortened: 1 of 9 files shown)


vmamidi (dismissed):

vmamidi (approved):

zwoop:

PR #7582, opened 2021-03-06
Add pointer/reference downcast function that is checked in debug builds.

ywkaras:
Added to include/tscore/ink_assert.h.  Checking is disabled in release
build for performance.

### DIFF
--- a/include/tscore/ink_assert.h
+++ b/include/tscore/ink_assert.h
@@ -53,13 +53,51 @@ inkcoreapi void _ink_assert(const char *a, const char *f, int l) TS_NORETURN;
 
 #ifdef __cplusplus
 }
-#endif /* __cplusplus */
 
-/* workaround a bug in the  stupid Sun preprocessor
+/*
+Use cast_to_derived() to cast a pointer/reference to a dynamic base class into a pointer/reference to a class
+that inherits directly or indirectly from the base class.  Uses checked dynamic_cast in debug builds, and
+static_cast in release (optimized) builds.  Use examples:
 
-#undef assert
-#define assert __DONT_USE_BARE_assert_USE_ink_assert__
-#define _ASSERT_H
-#undef __ASSERT_H__
-#define __ASSERT_H__
+class A { public: virtual ~A(); };
+class B : public A {};
+B * foo(A *a) { return cast_to_derived<B>(a); }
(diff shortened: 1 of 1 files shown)


include/tscore/ink_assert.h
```diff
@@ -53,13 +53,50 @@ inkcoreapi void _ink_assert(const char *a, const char *f, int l) TS_NORETURN;
...
+class B : public A {};
+B * foo(A *a) { return dbg_dyn_cast<B>(a); }
+B & foo2(A &a) { return dbg_dyn_cast<B>(a); }
+B const & foo3(A const &a) { return dbg_dyn_cast<B const>(a); }
 
-#undef assert
-#define assert __DONT_USE_BARE_assert_USE_ink_assert__
-#define _ASSERT_H
-#undef __ASSERT_H__
-#define __ASSERT_H__
```
ywkaras, line 64: Deleted these kruft comments while I was in here.

ywkaras:
[approve ci CentOS]

include/tscore/ink_assert.h
```diff
@@ -53,13 +53,50 @@ inkcoreapi void _ink_assert(const char *a, const char *f, int l) TS_NORETURN;
...
 */
+
+template <class Derived, class Base>
+Derived *
+dbg_dyn_cast(Base *b)
```
maskit, line 71: The behavior seems more like static cast because it doesn't return nullptr. We can use this as a replacement for static_cast but not dynamic_cast.

ywkaras: Hmm I was thinking this would only be used in cases where b was expected to never be null.  Probably I should have noted that in the comments.

maskit: > I was thinking this would only be used in cases where b was expected to never be null.

I think I understand where to use this. I was trying to say the name is confusing. People use static_cast in those cases, like the code for non-debug build does, and the behavior is more like static_cast since it never returns null, but it has dyn(amic) in its name. Why do we want to call it a kind of dynamic_cast?

I'd suggest `dbg_static_cast` or something doesn't mention dynamic nor static.

ywkaras: ok

maskit:
Isn't it downcast?

maskit (approved):
+1

PR #7588, opened 2021-03-09
Build the test library for tls_engine consistently

shinrich:
@bneradt fixed all the other tests so that the supporting programs and libraries are built in the main build.  This PR fixes up the remaining test library, async_engine.so to be built the same way.

PR #7591, opened 2021-03-11
Use return values to fix ubuntu release build error

shinrich:
Fix ubuntu release build error introduced in PR #7588

### DIFF
--- a/contrib/openssl/async_engine.c
+++ b/contrib/openssl/async_engine.c
@@ -27,6 +27,11 @@
  * operations spawns a thread to sleep for 5 seconds before resuming
  * the asynchronous job
  */
+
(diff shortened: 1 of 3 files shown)


randall (dismissed):

shinrich:
Turns out the ubuntu build warnings actually exposed a real issue in the async_engine.  In the case where we had already created the synchronization object, the engine was not actually causing ASYNC_pause_job.  So by the time my test delay thread tried to signal, the SSL connection and associated synchronization objects had already been cleaned up.

shinrich:
[approve ci autest]

randall (approved):

zwoop:

Issue #7471, opened 2021-02-02
ATS 9.0.1: current_server_connections, current_server_transactions metrics broken


mtorluemke:
Feels similar to https://github.com/apache/trafficserver/issues/7163 and https://github.com/apache/trafficserver/issues/7054
Fix for #7054 was #7056. 
Fix for #7163 was #7258

Both of these metrics function like a counter, instead of a guage:
proxy.process.http.current_server_connections
proxy.process.http.current_server_transactions

c-taylor:
This affects both:
proxy.process.http.current_parent_proxy_connections
proxy.process.http.current_server_connections

Behaviour is constantly additive. Do not see decrements.

zwoop:
proxy.process.http.current_server_connections is a derivative metric too, which derives from 

```
  {"proxy.process.current_server_connections",
   RECD_INT,
   {"proxy.process.http.current_server_connections", "proxy.process.http.current_parent_proxy_connections"}}};
```

c-taylor:
I believe this issue also gives a 'false positive' for `proxy.config.http.per_server.connection.max`.

Trying to confirm this now...

shinrich:
Interesting, in our 9.0.x build both look ok.  From a server running at low time.

```
proxy.process.http.current_server_transactions 179
proxy.process.http.current_server_connections 526
```

shinrich:
Hmm.  Interestingly enough current_server_transactions is broken on my H2 to origin build.

shinrich:
One difference in our environments is that you use parent proxy.  Does you proxy.process.http.current_parent_proxy_connections setting look ok?

c-taylor:
> This affects both:
> proxy.process.http.current_parent_proxy_connections
> proxy.process.http.current_server_connections
> 
> Behaviour is constantly additive. Do not see decrements.

maskit:
What a mess.
```
$ git grep http_current_server_transactions_stat | grep MENT_DYN
proxy/http/Http1ServerSession.cc:    HTTP_DECREMENT_DYN_STAT(http_current_server_transactions_stat);
proxy/http/HttpSM.cc:    HTTP_DECREMENT_DYN_STAT(http_current_server_transactions_stat);
proxy/http/HttpSM.cc:    HTTP_DECREMENT_DYN_STAT(http_current_server_transactions_stat);
proxy/http/HttpSM.cc:  HTTP_INCREMENT_DYN_STAT(http_current_server_transactions_stat);
```

[...]

PR #7594, opened 2021-03-12
Call do_io_close instead of HTTP2_SESSION_EVENT_FINI handler

maskit:
While I was looking into #7471, I realized that `Http2ClientSession::do_io_close` is not called if clients send GOAWAY.

Calling `do_io_close` is not equivalent to calling `HTTP2_SESSION_EVENT_FINI` event handler, and touching this area is always scary, but I don't think we should skip `do_io_close`. In case this change causes something we should keep `do_io_close` and fix the new issue.


shinrich:
Agreed that messing with this area seems scary, but moving forward to more standard shutdown sequences probably worth dealing with the potential fallout.

shinrich (approved):

PR #7595, opened 2021-03-12
Add an assert to check do_io_close is called

maskit:
I think we should always call `do_io_close` if `do_io_read/write` is called.

This should not be merged until #7594 is merged.


bneradt:
[approve ci autest]

randall:
[approve ci autest]

shinrich:
[approve ci autest]

maskit:
Autest won't pass until #7594 gets merged.

maskit:
[approve ci autest]

maskit:
Gave up for now because this would require quite a few changes. It would be better to wait for H2 to Origin at minimum.

`do_io_close` is not called if ATS sends GOAWAY frame.

PR #7598, opened 2021-03-13
Add client_allow_list.so experimental plugin.

ywkaras:
The plugin will fail TLS handshakes if the none of names of the certificate sent by the client appears
is a configured list of allowed certificates.  The list of allowed certificate names can be varied
based on the SNI server name the client sends.


plugins/experimental/client_allow_list/main.cc
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  // Is this a good name or not?
+  TSEvent reenable_event = TS_EVENT_CONTINUE;
+  X509_STORE_CTX *ctx    = reinterpret_cast<X509_STORE_CTX *>(TSVConnSslVerifyCTXGet(ssl_vc));
+  if (ctx) {
+    STACK_OF(X509) *chain = X509_STORE_CTX_get1_chain(ctx);
+    // X509 *cert = X509_STORE_CTX_get_current_cert(ctx);
+    bool retval = false;
+    for (int i = 0; i < sk_X509_num(chain) && !retval; i++) {
```
maskit, line 138:
```
experimental/client_allow_list/main.cc:138:23: warning: comparison of integer expressions of different signedness: ‘intand ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]
  138 |     for (int i = 0; i < sk_X509_num(chain) && !retval; i++) {
      |                     ~~^~~~~~~~~~~~~~~~~~~~
```

There's very similar code in `tests/tools/plugins/ssl_client_verify_test.cc`.

plugins/experimental/client_allow_list/main.cc
```diff
@@ -0,0 +1,180 @@
...
+#if TS_VERSION_MAJOR < 9
+#define TSVConnSslConnectionGet TSVConnSSLConnectionGet
+#endif
+
+namespace
+{
+inline const char *
+ASN1_get_data(ASN1_STRING *cn)
+{
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
```
maskit, line 45: I don't recommend to use `OPENSSL_VERSION_NUMBER` for new code because it's going to be deprecated.

plugins/experimental/client_allow_list/client_allow_list.h
```diff
@@ -0,0 +1,301 @@
...
+  };
+
+  struct _Eq {
+    bool operator()(char const *lhs, char const *rhs) const;
+  };
+
+  std::unordered_map<char const *, std::vector<unsigned>, _Hash, _Eq> _map;
+};
+
+class HashFnv1a32
```
maskit, line 144: If there's some reasons to not use `ATSHash32FNV1a` we might want to improve it.

maskit (changes requested):
Please address the compile issue at least.

plugins/experimental/client_allow_list/util.cc
```diff
@@ -0,0 +1,349 @@
...
+
+      std::size_t si = 0, di = 1;
+      _init->_insert_tmp[0] = '^';
+
+      while (si < name.size()) {
+        if (_init->_insert_tmp.size() < (di + 8)) {
+          _init->_insert_tmp.resize(di + 128);
+        }
+        if ('.' == name[si]) {
+          std::memcpy(_init->_insert_tmp.data() + di, "\\.", 3);
```
alficles, line 313: This unnecessarily copies in the null terminator after this string. Because the realloc above always reserves at least 8 bytes, this won't break anything as it is, but it's somewhat confusing, especially given the escape character. The copy just a few lines down does not do this.

plugins/experimental/client_allow_list/util.cc
```diff
@@ -0,0 +1,349 @@
...
+        } else if ('*' == name[si]) {
+          std::memcpy(_init->_insert_tmp.data() + di, ".{0,}", 5);
+          di += 5;
+        } else {
+          _init->_insert_tmp[di++] = name[si];
+        }
+        ++si;
+      }
+      _init->_insert_tmp[di++] = '$';
+      _init->_insert_tmp[di]   = '\0';
```
alficles, line 324: This routine escapes only dots. If I understand the plugin correctly, though, we're validating CNs as well, which don't have to be DNS names. That means the CNs we're looking for might contain regex magic within them.

Also, it's worth noting that if an attacker can control the contents of these entries (you can imagine a self-service system where users can set the allowed CNs for certs connecting specifically with their domain, when ATS is hosting domains for multiple users), there's a fairly straightforward ReDoS attack. That might be fine under most threat models and working around it might be hard, so documenting the situation might be easier than fixing it if arbitrary regular expressions are allowed here. If the plan isn't to allow arbitrary regexen, this concern disappears when the input is sanitized.

plugins/experimental/client_allow_list/util.cc
```diff
@@ -0,0 +1,349 @@
...
+
+      while (si < name.size()) {
+        if (_init->_insert_tmp.size() < (di + 8)) {
+          _init->_insert_tmp.resize(di + 128);
+        }
+        if ('.' == name[si]) {
+          std::memcpy(_init->_insert_tmp.data() + di, "\\.", 3);
+          di += 2;
+        } else if ('*' == name[si]) {
+          std::memcpy(_init->_insert_tmp.data() + di, ".{0,}", 5);
```
alficles, line 316: This will match multiple domains, which is undoubtedly intended. For example, `*.example.com` should match `www.site.example.com` with no surprise. But someone who writes `www.example.*` might well be surprised to see it match `www.example.evil.com`. Is there a use case for a wildcard anywhere but the very front of a DNS name? Or is this to allow matches for things like "Bill *" to allow matches for all persons named Bill?

alficles (commented):
I had some observations, but this is experimental and seems like a decent place to start. It feels like there is one main use case here: where the operator controls all signing keys for a zone and can ensure that all issued certificates contain names exclusively of the required form.

It does feel like you should be able to use RFC 5280 Name Constraints to prevent intermediate parties from signing certs that would inappropriately match domains. But I'm really not sure this works well, even with the latest version of OpenSSL doing the verification. CNs don't have to contain DNS names and indeed for end users, they can quite reasonably contain personal names instead. If your PKI allows organizations with content served by your ATS instances to sign their own certs, I'm not sure the approach here will prevent them from signing certs with names that match other domains.

For the second use case, you'd probably prefer to use a separate CA for that domain in sni.yaml for the verification in any case.

I think some of the logic around the wildcards is going to want some thinking, but this seems like a reasonable start.

mlibbey:
FWIW, our org uses a field inside the Subject DN eg, when connecting to the ATS with SNI cdn.example.com, only accept mTLS requests with Subject DN field containing the UID=ourorg.group.123, and with specific Issuer DN values. The security peeps also suggest/mandate/whatever that the cert should chain to specific roots. Even if this functionality doesn't make it into this version, might be nice to think through the config language to be able to add it later.

plugins/experimental/client_allow_list/client_allow_list.h
ywkaras: Seems like we should move it to ts/util from tscore it we're going to use it in plugins.

plugins/experimental/client_allow_list/main.cc
ywkaras: What is the alternative?

maskit: Checking availability of the function.

ywkaras:
[approve CI FreeBSD]

plugins/experimental/client_allow_list/main.cc
ywkaras: In configure.ac?

maskit: Yes.

ywkaras: With a herring?

maskit: That's up to you. This one is not a blocker.
You could do it with a magic as well. This spell is already in our configure.ac.
```
AC_CHECK_FUNC([ASN1_STRING_get0_data], [],                                                                                                                    
              [AC_DEFINE([ASN1_STRING_get0_data], [ASN1_STRING_data], [Added in OpenSSL 1.1])])
```

plugins/experimental/client_allow_list/unit_tests/test.cc
```diff
@@ -0,0 +1,360 @@
...
+    char const *args[] = {"dummy plugin name", DIR "bad10.yaml"};
+    REQUIRE(bad_config(2, args));
+  }
+  UT_CHECK_OUTPUT("client_allow_list: YAML::Exception \"yaml-cpp: error at line 4, column 5: the referenced "
+                  "anchor is not defined\" when parsing YAML config file ./" DIR "bad10.yaml\n");
+
+  return 0;
+}
+
+#include <../src/tscore/HashFNV.cc>
```
maskit, line 359: Do we need this? Seems like ssl_session_reuse uses HashFNV without including the cc files. And there are some plugins that include files under include/tscore.

maskit (commented):
I didn't review this from security perspective, but it would be better to leave a note on the documentation if there are any concerns, discussion or future plans.

Including cc files is questionable but not a blocker.

plugins/experimental/client_allow_list/unit_tests/test.cc
ywkaras: These includes are only for unit test / make check.  Is there a better way to access this code in the unit test executables.

maskit: Tests for access_control and sslheaders link libtscore. I'm not sure if it's better but looks more normal.

ywkaras: Yes that would work, but it makes the dependencies of the test executable more ambiguous.

bryancall:
@ywkaras How about switching to std::hash and use string_view?

ywkaras:
> @ywkaras is going to change it to use std::hash and string_view

I had forgotten that the reason I didn't use std::string_view as the unordered_map key (with default hash and compare) is that, because the key is the SNI servername, it must be case insensitive ( https://tools.ietf.org/html/rfc3546#page-8 ).  I could normalize the servername to lowercase before doing the lookup here:  https://github.com/apache/trafficserver/pull/7598/files#diff-a2fcfdba174f68b6444d283948e88441c38cfa09799fc67ae93fadf9e81429e0R203 .  But that would require using non-standard variable length stack allocation (alloca() or a variable length array).  It seems unlikely an extremely long SNI server name could be seen at this point (allowing for a stack blow DOS attack) but I guess I'd have to double check.

PR #7584, opened 2021-03-09
Call constructors and destructors for H1/2 Session/Transaction via ClassAllocator

maskit:
#6241 enabled to call constructors which receive arguments, and destructors.

PR #7600, opened 2021-03-13
Fix double destuct on Http2Stream termination

maskit:
The bug was introduced by #7584.


shinrich (approved):

zwoop:

PR #7601, opened 2021-03-15
Update -with-profile and add some profiling documentation

shinrich:


shukitchan (dismissed):
This looks good.

maskit:
Why was heap profiler removed? If we are going to remove it, can we remove the include line as well?
```
$ git grep gperftools
src/traffic_server/traffic_server.cc:#include <gperftools/profiler.h>
src/traffic_server/traffic_server.cc:#include <gperftools/heap-profiler.h>
```

bryancall:
I would like to know why the heap profiler was remove also.

shinrich:
The symbols did not exist with the new header files.   Probably different symbols.  Since I was concentrating on CPU performance, I didn't track it down..

maskit:
It seems like the three functions exist one the latest header file. Am I looking at an old version or a different file?
https://github.com/gperftools/gperftools/blob/master/src/gperftools/heap-profiler.h

shinrich:
Pushed another commit that added back the heap profiling calls.  Built for me now.  Don't now what was my problem before.

maskit (approved):

zwoop:

Issue #7521, opened 2021-02-12
Crash in OutboundConnTrack - per_server.connection.max


c-taylor:
I will try to repro this in a controlled environment, but sharing the stack trace early...

ATS 9.0.1
Crash/trace shortly after per_server.connection.max triggered WARNING messages.

This appeared after 1-2h of load:
```
/some/path/bin/traffic_server(crash_logger_invoke(int, siginfo_t*, void*)+0x99)[0x559787130dc9]
/lib64/libpthread.so.0(+0xf630)[0x7f4e0247b630]
/some/path/bin/traffic_server(OutboundConnTrack::TxnState::Warn_Blocked(OutboundConnTrack::TxnConfig const*, long, int, sockaddr const*, char const*)+0x41)[0x55978718cd11]
/some/path/bin/traffic_server(HttpSM::do_http_server_open(bool)+0x1455)[0x55978719a745]
/some/path/bin/traffic_server(HttpSM::set_next_state()+0x32e)[0x5597871b089e]
/some/path/bin/traffic_server(HttpSM::send_origin_throttled_response()+0x10e)[0x5597871a2d2e]
/some/path/bin/traffic_server(HttpSM::do_http_server_open(bool)+0x12d8)[0x55978719a5c8]
/some/path/bin/traffic_server(HttpSM::set_next_state()+0x32e)[0x5597871b089e]
/some/path/bin/traffic_server(HttpSM::call_transact_and_set_next_state(void (*)(HttpTransact::State*))+0x10c)[0x559787196bfc]
/some/path/bin/traffic_server(HttpSM::state_cache_open_write(int, void*)+0x3f5)[0x5597871a68a5]
/some/path/bin/traffic_server(HttpSM::main_handler(int, void*)+0x1b1)[0x5597871937b1]
/some/path/bin/traffic_server(HttpCacheSM::state_cache_open_write(int, void*)+0x48a)[0x5597871ee44a]
/some/path/bin/traffic_server(CacheVC::callcont(int)+0x7b)[0x5597872d4b9b]
/some/path/bin/traffic_server(Cache::open_write(Continuation*, ats::CryptoHash const*, HTTPInfo*, long, ats::CryptoHash const*, CacheFragType, char const*, int)+0x40e)[0x5597872de97e]
[...]

PR #7602, opened 2021-03-16
Fix the connection limit crash while using parents

vmamidi:
Fix for #7521. do_http_server_open can be called recursively when there are multiple parents for a domain. Warn_Blocked can crash in this scenario.


zwoop:

zwoop:
Setting an arbitrary Milestone here, since this was never properly changelogg'ed :-/.

PR #7604, opened 2021-03-16
Fix the connection limit crash while using parents

vmamidi:
Fix for #7521. do_http_server_open can be called recursively when there are multiple parents for a domain. Warn_Blocked can crash in this scenario.


SolidWallOfCode (approved):

zwoop:
This has merge conflicts, but I think we can change #7602 to be CP'ed into 9.0.x as well.

PR #7613, opened 2021-03-22
Add AuTest for Background Fill

masaori335:
Test case 1 ( Background Fill on HTTP/1.1 over TLS ) will fail for now. Requires fix of #7603.

### DIFF
--- a/tests/gold_tests/cache/background_fill.test.py
+++ b/tests/gold_tests/cache/background_fill.test.py
@@ -1,0 +1,140 @@
+'''
+'''
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


ezelkow1:
Can one of the admins verify this patch?

tests/gold_tests/cache/background_fill.test.py
```diff
@@ -0,0 +1,125 @@
...
+    def __checkProcessAfter(self, tr):
+        tr.StillRunningAfter = self.httpbin
+        tr.StillRunningAfter = self.ts
+
+    def __testCase0(self):
+        """
+        HTTP/1.1 over TCP
+        """
+        tr = Test.AddTestRun()
+        self.__startProcess(tr)
```
bneradt, line 78: It could be helpful to rewrite consolidate __startProcess and __checkProcessBefore into a single call site for the test run functions that uses a static variable to see whether the processes have been started already and, if not, calls StartBefore but, if it had been called, calls StillRunningBefore. Then the user can easily comment out all but one of the cases and things will work.

This is just a suggestion. It would be pretty easy to do with the way you organized this. But feel free to ignore it.

bneradt (dismissed):
This is a very well organized test.

tests/gold_tests/cache/background_fill.test.py
masaori335: Good suggestion. I combined them.

bneradt:
Looks good to me.

masaori335:
[approve ci autest]

masaori335:
@bneradt Would you take another look? The fix (#7641) is merged. This is ready for review.

bneradt (approved):
Excellent.

PR #7614, opened 2021-03-22
Change ROUNDUP from function-like macro to function template.

ywkaras:
Making it safe for argument expressions with side effects.

### DIFF
--- a/include/tscore/ink_defs.h
+++ b/include/tscore/ink_defs.h
@@ -104,10 +104,6 @@ countof(const T (&)[N])
 #include <hwloc.h>
 #endif
 
-#ifndef ROUNDUP
-#define ROUNDUP(x, y) ((((x) + ((y)-1)) / (y)) * (y))
-#endif
-
 #if defined(MAP_NORESERVE)
 #define MAP_SHARED_MAP_NORESERVE (MAP_SHARED | MAP_NORESERVE)
 #else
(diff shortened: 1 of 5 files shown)


ezelkow1:
Can one of the admins verify this patch?

bryancall:
@alficles Can you please review this?

include/tscore/ink_defs.h
```diff
@@ -126,6 +122,21 @@ int ink_sys_name_release(char *name, int namelen, char *release, int releaselen)
...
+template <typename ArithmeticV, typename ArithmeticM>
+constexpr ArithmeticV
+ROUNDUP(ArithmeticV value, ArithmeticM m)
+{
+  ArithmeticV remainder = value % m;
+  if (remainder) {
+    value += m - remainder;
+  }
+  return value;
+}
```
alficles, line 137: This only provides it when it's included from a C++ file. It's a little odd to remove the C version when converting to C++. This version is unambiguously better, though, since it will refuse to compile with non-integer types.

alficles (approved):
This is an improvement, for sure. Reducing duplication and improving type safety is a win. It'd be nice to not "lose" the C version when included from a C file, but since it's not used, we can probably let whoever needs it in C provide it and not maintain it if nobody ever does.

include/tscore/ink_defs.h
ywkaras: ink_defs.h is included by .c files, but they don't use ROUNDUP.

ezelkow1 (approved):

PR #7618, opened 2021-03-24
Reload server session inactivity timeout before placing a session into the pool

### LINKED ISSUE #7471: ATS 9.0.1: current_server_connections, current_server_transactions metrics broken

maskit:
This may fix #7471. 

I'm not sure if this really solves the issue, but there's a possibility that inactivity timeout is set to 0 on some server sessions and those sessions stay in the pool forever.

Setting active timeout to the value of `keep_alive_no_activity_timeout_out` looks odd, but it's actually the original behavior before [TS-863](https://issues.apache.org/jira/browse/TS-863). Since there should be no activity, it doesn't really matter, like the comment on line 214 says.

### DIFF
--- a/proxy/http/HttpSessionManager.cc
+++ b/proxy/http/HttpSessionManager.cc
@@ -212,8 +212,10 @@ ServerSessionPool::releaseSession(PoolableSession *ss)
   ss->do_io_write(this, 0, nullptr);
 
   // we probably don't need the active timeout set, but will leave it for now
-  ss->set_inactivity_timeout(ss->get_netvc()->get_inactivity_timeout());
-  ss->set_active_timeout(ss->get_netvc()->get_active_timeout());
+  HttpConfigParams *http_config_params = HttpConfig::acquire();
+  ss->set_inactivity_timeout(HRTIME_SECONDS(http_config_params->oride.keep_alive_no_activity_timeout_out));
+  ss->set_active_timeout(HRTIME_SECONDS(http_config_params->oride.keep_alive_no_activity_timeout_out));
+  HttpConfig::release(http_config_params);
   // put it in the pools.
   m_ip_pool.insert(ss);
   m_fqdn_pool.insert(ss);
@@ -270,8 +272,8 @@ ServerSessionPool::eventHandler(int event, void *data)
                 "[%" PRId64 "] [session_bucket] session received io notice [%s], "
                 "resetting timeout to maintain minimum number of connections",
                 s->connection_id(), HttpDebugNames::get_event_name(event));
-          s->get_netvc()->set_inactivity_timeout(s->get_netvc()->get_inactivity_timeout());
-          s->get_netvc()->set_active_timeout(s->get_netvc()->get_active_timeout());
+          s->get_netvc()->set_inactivity_timeout(HRTIME_SECONDS(http_config_params->oride.keep_alive_no_activity_timeout_out));
+          s->get_netvc()->set_active_timeout(HRTIME_SECONDS(http_config_params->oride.keep_alive_no_activity_timeout_out));
           found = true;
           break;
         }


shinrich:
I don't think this is causing the connection count problem.

In  HttpSM you see the pattern

```
      server_session->set_inactivity_timeout(HRTIME_SECONDS(t_state.txn_conf->keep_alive_no_activity_timeout_out));
      server_session->release(nullptr);
```
I'd suggest either, adding an ink_release_assert in ServerSessionPool::releaseSession to verify that the timeout is currently set to a non-zero value.  And/or pass along the keep_alive timeout value as an argument to release, so they are done together always.

I think the reason for the get/set is to make sure that the timeout is at its full level.  But truthfully, I don't see the path that would take the session out of the pool and not go through the release path.  So perhaps these lines aren't necessary at all.

maskit:
I can see the pattern in HttpSM, but those are not the only places calling `Http1ServerSession::release()`. It's called from Http1ClientSession as well and the timeout is not reset there. 
```
$ git grep -E -- "->release\([^\)]"
plugins/prefetch/fetch.cc:    ret &= _policy->release(url);
plugins/prefetch/fetch.cc:  permitted     = _unique->release(url);
plugins/prefetch/fetch.cc:      _state->release(_cachekey);
plugins/prefetch/fetch.cc:      _state->release(cachekey);
plugins/prefetch/plugin.cc:            state->release(data->_cachekey);
plugins/prefetch/plugin.cc:          state->release(data->_cachekey);
proxy/http/Http1ClientSession.cc:    bound_ss->release(nullptr);
proxy/http/Http1ClientSession.cc:    bound_ss->release(nullptr);
proxy/http/Http1ClientSession.cc:  this->release(&trans);
proxy/http/HttpSM.cc:      server_session->release(nullptr);
proxy/http/HttpSM.cc:    ua_txn->release(ua_buffer_reader);
proxy/http/HttpSM.cc:    ua_txn->get_proxy_ssn()->release(ua_txn);
proxy/http/HttpSM.cc:        existing_ss->release(nullptr);
proxy/http/HttpSM.cc:      existing_ss->release(nullptr);
proxy/http/HttpSM.cc:      server_session->release(nullptr);
proxy/http/HttpSessionManager.cc:    to_return->release(nullptr);
```

> I'd suggest either, adding an ink_release_assert in ServerSessionPool::releaseSession to verify that the timeout is currently set to a non-zero value. And/or pass along the keep_alive timeout value as an argument to release, so they are done together always.

I don't think those are great idea. I could add ink_release_assert to check whether it's really because of setting timeout to 0, but that would be a disaster if it is. And passing timeout value to the release function doesn't guarantee anything because you can still pass 0.

IMO, ServerSessionPool or HttpSessionManager should be responsible for setting a valid timeout before placing server sessions into the pool. How server sessions are managed in the pool is none of HTTP state machine's business. We could have another timeout for sessions in the pool in addition to inactivity timeout for sessions in use. I think requiring session pool users to set correct values before returning a session is too much and it's unreliable.

duke8253:
This is interesting, I'm actually dealing with a similar issue for one of our teams here. I'll check the configs they are using, though I don't have access to their deployment currently, might take a while.

maskit:
@duke8253 Does this work for your teams? It didn't work for us - no good/bad effects.

duke8253:
@maskit not really , tested for a few days and didn't see any good or bad effects.

maskit:
I don't have a big motivation for this since the cause of #7471 was not inactive timeout, but it still doesn't make sense to me that inactivity timeout value from NetVC that might have been set to 0 is used to rearm inactivity timeout.

Are there any reason to not reload the original value in config?

shinrich (approved):

zwoop:
@maskit Do we really want this for 9.0.x, or should we defer for 9.1.x or 9.2.x ?

maskit:
I'm fine with deferring to 9.1 or 9.2 because it seems like nobody is suffered from this bug.

PR #7619, opened 2021-03-24
Use standard isdigit library function

shinrich:
Noticed there were two is_digit functions floating around in HTTP.cc.  One defined within the file and another from the ParseRules namespace.  It sees that these days the isdigit() OS function should be at least as efficient as the ParseRules implementation.

So this PR removes the file-local implementation, and replaces all calls to is_digit() to the OS library function isdigit().

At some point we may want to review and pare down the ParseRules logic, but I will leave that for another day.


bneradt (approved):

masaori335 (approved):

zwoop:

PR #7623, opened 2021-03-25
New rate_limit plugin for simple resource limitations

zwoop:
This current version has only one limiter, which implements
a basic active_in limitation. However, at least for now,
these connections that are queued wills still count against
the active connection metrics and limits that are in the core.

I'm open for a refactoring of this, if or when we want to have
different types of limiters. Similar to the policies for the
cache_promote plugin.

### DIFF
--- a/doc/admin-guide/plugins/rate_limit.en.rst
+++ b/doc/admin-guide/plugins/rate_limit.en.rst
@@ -65,12 +65,17 @@ are available:
 .. option:: --header
 
    This is an optional HTTP header name, which will be added to the client
-   request header IF the transaction was delayed (queued). This can be useful
-   to for example log the delays for later analysis.
+   request header IF the transaction was delayed (queued). The value of the
+   header is the delay, in milliseconds. This can be useful to for example
+   log the delays for later analysis.
 
    It is recommended that an `@` header is used here, e.g. `@RateLimit-Delay`,
-   since this header will not leave the ATS server instance. The value here is
-   appended to the header should one already exist.
+   since this header will not leave the ATS server instance.
+
+.. option:: --maxage
+
+   An optional `max-age` for how long a transaction can sit in the delay queue.
+   The value (default 0) is the age in seconds.
 
 Examples
 --------
--- a/doc/admin-guide/plugins/rate_limit.en.rst
+++ b/doc/admin-guide/plugins/rate_limit.en.rst
@@ -37,7 +37,7 @@ are available:
 
 .. option:: --limit
 
-   The maximum number of active connections.
+   The maximum number of active client transactions.
 
 .. option:: --queue
 
@@ -45,6 +45,12 @@ are available:
    on a FIFO queue. This option (optional) sets an upper bound on how many
    queued transactions we will allow. When this threshold is reached, all
    additional transactions are immediately served with an error message.
+
+   The queue is effectively disabled if this is set to `0`, which implies
+   that when the transaction limit is reached, we immediately start serving
+   error responses.
+
+   The default queue size is `UINT_MAX`, which is essentially unlimited.
 
 .. option:: --error
 
@@ -60,3 +66,10 @@ code `429` is used when queue is full.
 
     map http://cdn.example.com/ http://some-server.example.com \
       @plugin=rate_limit.so @pparam=--limit=128 @pparam=--queue=256
+
+
+This example would put a hard transaction (in) limit to 256, with no backoff
+queue:
+
+    map http://cdn.example.com/ http://some-server.example.com \
+      @plugin=rate_limit.so @pparam=--limit=256 @pparam=--queue=0
--- a/doc/admin-guide/plugins/rate_limit.en.rst
+++ b/doc/admin-guide/plugins/rate_limit.en.rst
@@ -80,7 +80,7 @@ are available:
 Examples
 --------
 
-This example shows a simple rate limiting of `128` concurently active client
+This example shows a simple rate limiting of `128` concurrently active client
 transactions, with a maximum queue size of `256`. The default of HTTP status
 code `429` is used when queue is full.
 
--- a/doc/admin-guide/plugins/rate_limit.en.rst
+++ b/doc/admin-guide/plugins/rate_limit.en.rst
@@ -75,7 +75,7 @@ are available:
 .. option:: --maxage
 
    An optional `max-age` for how long a transaction can sit in the delay queue.
-   The value (default 0) is the age in seconds.
+   The value (default 0) is the age in milliseconds.
 
 Examples
 --------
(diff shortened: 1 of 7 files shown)


zwoop:
[approve ci autest]

zwoop:
Found a bug, TSContScheduleEveryOnPool() does not play nicely with config reloads and TSContDestroy() :-/.

plugins/experimental/rate_limit/limiter.h
```diff
@@ -0,0 +1,162 @@
...
+  bool
+  full() const
+  {
+    return (_size == max_queue);
+  }
+
+  void
+  push(TSHttpTxn txnp, TSCont cont)
+  {
+    QueueTime now = std::chrono::system_clock::now();
```
shukitchan, line 104: Should we use TShrtime() instead?

shukitchan (commented):
I think we have proxy.config.net.max_connections_in / proxy.config.http.server.max_connections but they are applied globally. 

So we need this on a "per remap rules" basis ? Is this supposed to be a stop gap before we can make the above 2 "overridable" (i.e. can be changed by conf_remap) ?

plugins/experimental/rate_limit/limiter.h
zwoop: I'd prefer not to :-). The APIs around TShrtime are lame :).

zwoop:
[approve ci]

zwoop:
[approve ci]

zwoop:
[approve ci]

zwoop:
Fixed the spelling errors, please review again :).

plugins/experimental/rate_limit/limiter.h
```diff
@@ -0,0 +1,182 @@
...
+#include <atomic>
+#include <cstdio>
+#include <chrono>
+#include <cstring>
+#include <string>
+
+#include <ts/ts.h>
+
+constexpr char const PLUGIN_NAME[]  = "rate_limit";
+constexpr unsigned QUEUE_DELAY_TIME = 100; // Examine the queue every 100ms.
```
bneradt, line 30: Might want to consider:

```
constexpr auto QUEUE_DELAY_TIME = std::chrono::milliseconds{100};

...

_action = TSContScheduleEveryOnPool(_queue_cont, QUEUE_DELAY_TIME.count(), TS_THREAD_POOL_TASK);
```

bneradt (approved):
Looks great. I approve. Including a few minor suggestions for your consideration. Implement or drop them as you see fit.

doc/admin-guide/plugins/rate_limit.en.rst
```diff
@@ -0,0 +1,115 @@
...
+The limit counters and queues are per remap rule only, i.e. there is
+(currently) no way to group transaction limits from different remap rules
+into a single rate limiter.
+
+All configuration is done via :file:`remap.config`, and the following options
+are available:
+
+.. program:: rate-limit
+
+.. option:: --limit
```
ywkaras, line 38: Why all the double dashes?

plugins/experimental/rate_limit/limiter.cc
```diff
@@ -0,0 +1,159 @@
...
+RateLimiter::queue_process_cont(TSCont cont, TSEvent event, void *edata)
+{
+  RateLimiter *limiter = static_cast<RateLimiter *>(TSContDataGet(cont));
+  QueueTime now        = std::chrono::system_clock::now(); // Only do this once per "loop"
+
+  // Try to enable some queued txns (if any) if there are slots available
+  while (limiter->size() > 0 && limiter->reserve()) {
+    QueueItem item                  = limiter->pop();
+    TSHttpTxn txnp                  = std::get<0>(item);
+    std::chrono::microseconds delay = std::chrono::duration_cast<std::chrono::milliseconds>(now - std::get<2>(item));
```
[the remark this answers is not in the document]
zwoop: Bryan hates Auto :-)

doc/admin-guide/plugins/rate_limit.en.rst
zwoop: Cause that's how we do long-opts?

zwoop: Cause that's how we do this?

plugins/experimental/rate_limit/limiter.cc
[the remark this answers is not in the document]
zwoop: Bryan hates auto :).

doc/admin-guide/plugins/rate_limit.en.rst
ywkaras: Do what?

ywkaras: Why not just omit them?  I could see the reason for double dashes before options for a global plugin but not so much for a remap plugin.  Seems redundant with the the @pparam.

zwoop: This is how like every plugin works. With the --'s, we can use standard libC getopt_long() APIs.

bneradt (approved):

plugins/experimental/rate_limit/limiter.h
```diff
@@ -0,0 +1,182 @@
...
+#include <cstring>
+#include <string>
+
+#include <ts/ts.h>
+
+constexpr char const PLUGIN_NAME[] = "rate_limit";
+constexpr auto QUEUE_DELAY_TIME    = std::chrono::milliseconds{100}; // Examine the queue every 100ms
+
+using QueueTime = std::chrono::time_point<std::chrono::system_clock>;
+using QueueItem = std::tuple<TSHttpTxn, TSCont, QueueTime>;
```
ywkaras, line 33: Call me a loom smasher but I think I would just do:
```
struct QueueItem
{
  TSHttpTxn txnp;
  TSCont contp;
  QueueTime start_time;
};
```

bneradt (approved):

zwoop:

sudheerv:
Nice and useful plugin to have. 

A comment/observation though - 

1) Given that this is a remap plugin, it feels like the rate limiting of the transactions is effectively on the Origin transactions (at least, until H2 to Origin is supported)  as opposed to the client side. 
2) I also wonder if this is potentially close to what https://docs.trafficserver.apache.org/en/9.0.x/admin-guide/files/records.config.en.html#proxy-config-http-per-server-connection-max does already. There is also a queueing support with the outbound connection tracking.

Fwiw, we use a similar rate limiter plugin running in production with the difference being it is a global plugin and tracks concurrency against explicitly configured limits (ie not via remap.config). We've a bunch of business logic (including several sideways calls) that is run before remap state, so, queuing or limiting at remap is a little bit too late if we wanted to cap the resource usage more efficiently.

PR #7627, opened 2021-03-26
Add pooled_server_connections metric

shinrich:
Originally added by PR #7622.  The new stat was useful for debugging lost connections.  @maskit pointed out that the metric would be useful independent of the H2 to origin logic, so I've pulled it out separately.


maskit (approved):
This will be helpful to narrow down the cause of #7471.

zwoop:

PR #7628, opened 2021-03-26
Expose URL element methods through HTTPHdr

shinrich:
Another change originally made via PR #7622.  These methods are needed to correctly create the outgoing HTTP/2 headers.  But the could also be useful independent of the HTTP/2 to origin logic.


maskit (approved):

zwoop:

Seems benign now, and required for testing H22O.

PR #7629, opened 2021-03-26
Add proxy.config.cache.log.alternate.eviction

bneradt:
This commit adds the ability to toggle the logging of cache eviction of
alternates. This can be helpful for tuning
proxy.config.cache.limits.http.max_alts and detecting problematic
Vary: behavior with origins.

### DIFF
--- a/iocore/cache/CacheWrite.cc
+++ b/iocore/cache/CacheWrite.cc
@@ -98,6 +98,30 @@ CacheVC::updateVector(int /* event ATS_UNUSED */, Event * /* e ATS_UNUSED */)
       if (od->move_resident_alt && get_alternate_index(write_vector, od->single_doc_key) == 0) {
         od->move_resident_alt = false;
       }
+      if (cache_config_log_alternate_eviction) {
+        CacheHTTPInfo *info = write_vector->get(0);
+        HTTPHdr *request    = info->request_get();
+        if (request->valid()) {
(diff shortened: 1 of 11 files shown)


bneradt:
I've marked this for the 9.1.x and 9.2.x projects because the feature is simple and low risk (the relevant code path is not exercised if the default-disabled feature is disabled) and has already proved useful to our production engineers. It should also merge easily.

randall (approved):
👍

iocore/cache/CacheWrite.cc
```diff
@@ -98,6 +98,30 @@ CacheVC::updateVector(int /* event ATS_UNUSED */, Event * /* e ATS_UNUSED */)
...
+      if (cache_config_log_alternate_eviction) {
+        CacheHTTPInfo *info = write_vector->get(0);
+        HTTPHdr *request    = info->request_get();
+        if (request->valid()) {
+          // In contrast to url_string_get, this url_print interface doesn't
+          // use HTTPHdr's m_heap which is not valid at this point because the
+          // HttpSM is most likely gone.
+          int url_length = request->url_printed_length();
+          ats_scoped_mem<char> url_text;
+          url_text   = static_cast<char *>(ats_malloc(url_length + 1));
```
masaori335, line 110: FWIW, `ts::LocalBuffer` is faster when url length is small by avoiding using heap.

PR #7630, opened 2021-03-26
Add default implementation for allow_half_open

shinrich:
Adding implementation for allow_half_open in ProxyTransaction.  Only one case needs anything other than return false.  I had added in PR #7622  but @maskit noted that this fix should be pulled out and applied consistently.


maskit (approved):

zwoop:

PR #7632, opened 2021-03-26
Disable JemallocNodumpAllocator with config

duke8253:
The JemallocNodumpAllocator should be disabled when `proxy.config.allocator.dontdump_iobuffers` is set to `0`.


duke8253:
closing for now

PR #7635, opened 2021-03-27
generator: allow for POST requests

randall:
POST body is received and discarded


ywkaras:
Do you have any notes describing how you tested this change?  Don't you need to change:  https://docs.trafficserver.apache.org/en/latest/admin-guide/plugins/generator.en.html ?

randall:
> Do you have any notes describing how you tested this change? Don't you need to change: https://docs.trafficserver.apache.org/en/latest/admin-guide/plugins/generator.en.html ?

Good point. Modified generator docs to reflect change.

bryancall (approved):

PR #7639, opened 2021-03-28
Remove reference to client.verify.server from tests and other bits

randall:
Option originally removed in 9f15e7750030d298078ab43d62534d331c15d63c


randall:
I also went a bit overboard and removed `proxy.config.ssl.server.cipher_suite` where it looked like it was copy pasta (and tests still pass). I can pull that change out separately if preferred

shinrich (approved):

tests/gold_tests/tls_hooks/tls_hooks18.test.py
```diff
@@ -56,23 +54,23 @@
...
-tr.StillRunningAfter = ts
-tr.StillRunningAfter = server
-tr.Processes.Default.Command = 'curl -k -H \'host:example.com:{0}\' https://127.0.0.1:{0}'.format(ts.Variables.ssl_port)
-tr.Processes.Default.ReturnCode = 0
-tr.Processes.Default.Streams.stdout = "gold/preaccept-1.gold"
+tr.StillRunningAfter=ts
+tr.StillRunningAfter=server
+tr.Processes.Default.Command='curl -k -H \'host:example.com:{0}\' https://127.0.0.1:{0}'.format(ts.Variables.ssl_port)
+tr.Processes.Default.ReturnCode=0
+tr.Processes.Default.Streams.stdout="gold/preaccept-1.gold"
```
bneradt, line 64: Interesting. Is this spacing is accepted by pep 8? Normally for variable instantiations like this it prefers spaces around the `=`.

randall: agreed! This was the committed output from autopep8 target.

bneradt: Oh, I think I see what's happening. If you check out line 45, I think that has a mistake in the patch that is throwing off the parser:

https://github.com/apache/trafficserver/blob/7504f6726fed96e1b11b5bad1348af4932d1c202/tests/gold_tests/tls_hooks/tls_hooks18.test.py#L45

I'll create a patch.

bneradt: Here you go: #7643

randall: good find! How did this test pass?

PR #7651, opened 2021-03-30
Fixing Throttler.h for older clang and gcc compilers

bneradt:
This is a workaround for older gcc and clang compilers which implemented
an older version of the standard which made atomic's noexcept
construction specification not compatible with Throttler's use of
time_point's undecorated constructor.

### DIFF
--- a/include/tscore/Throttler.h
+++ b/include/tscore/Throttler.h
@@ -105,16 +105,27 @@ public:
 private:
   /// Base clock.
   using Clock = std::chrono::system_clock;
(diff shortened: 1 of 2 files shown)


bneradt:
[approve ci ubuntu]

randall (approved):

bneradt:
Log throttling, for which this change fixes older clang builds, only went into 9.2.x. So this is needed for the 9.2.x release, but not other releases.

PR #7668, opened 2021-04-01
AuTest: address various permissions issues

bneradt:
By default, the Traffic Server process runs as user nobody. This often isn't a
problem with the AuTest runs, but in certain circumstances this process will
not be able to either read or write from certain locations. For instance, by
default, shells in docker containers run as root and the home directory is
/root with permissions set to 550. There's nothing fundamental about AuTest
that prevents this, but since the ATS process is unprivileged it will not be
able to read or write to anything in /root.  There can be other manifestations
of permissions limitations like this.

The AuTest Traffic Server extension already sets up various directories in the
sandbox chowned to nobody for the ATS process to access. This tweaks a few of
the tests that assumed ATS access outside of these directories to use these
designated ones. This allows the test to run in more environments.

### DIFF
--- a/tests/gold_tests/pluginTest/combo_handler/combo_handler.test.py
+++ b/tests/gold_tests/pluginTest/combo_handler/combo_handler.test.py
@@ -16,6 +16,8 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import os
+
 Test.Summary = '''
 Test combo_handler plugin
 '''
@@ -104,12 +106,8 @@ ts.Disk.remap_config.AddLine(
     'map http://localhost/sub/ http://127.0.0.1:{}/sub/'.format(server.Variables.Port)
 )
 
-ts.Disk.File(ts.Variables.CONFIGDIR + "/ctwl.txt", id="ctwl_cfg", typename="ats:config")
-ts.Disk.ctwl_cfg.AddLine("# test  ")
-ts.Disk.ctwl_cfg.AddLine("")
-ts.Disk.ctwl_cfg.AddLine("		text/javascript  # test side comment")
-ts.Disk.ctwl_cfg.AddLine("  application/javascript")
-ts.Disk.ctwl_cfg.AddLine("text/css")
+# Configure the combo_handler's configuration file.
+ts.Setup.Copy("ctwl.txt", ts.Variables.CONFIGDIR)
 
 tr = Test.AddTestRun()
 tr.Processes.Default.StartBefore(ts)
--- a/tests/gold_tests/pluginTest/lua/lua_watermark.test.py
+++ b/tests/gold_tests/pluginTest/lua/lua_watermark.test.py
@@ -16,6 +16,8 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import os
+
 Test.Summary = '''
 Test lua functionality
 '''
@@ -41,8 +43,11 @@ server.addResponse("sessionfile.log", request_header, response_header)
 
 ts.Disk.remap_config.AddLine(
     'map / http://127.0.0.1:{}/'.format(server.Variables.Port) +
-    ' @plugin=tslua.so @pparam={}/watermark.lua'.format(Test.TestDirectory)
+    ' @plugin=tslua.so @pparam=watermark.lua'
 )
+
+# Configure the tslua's configuration file.
+ts.Setup.Copy("watermark.lua", ts.Variables.CONFIGDIR)
 
 ts.Disk.records_config.update({
     'proxy.config.diags.debug.enabled': 1,
--- a/tests/gold_tests/pluginTest/url_sig/url_sig.test.py
+++ b/tests/gold_tests/pluginTest/url_sig/url_sig.test.py
@@ -63,35 +63,36 @@ ts.Disk.ssl_multicert_config.AddLine(
 
 # Use unchanged incoming URL.
 #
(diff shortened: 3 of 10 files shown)


bneradt:
I marked this PR for the various 9.x releases since it is a test-only change that will improve the usability of the tests in various environments. It's not a critical patch though. If it doesn't merge cleanly anywhere, we can just skip it for that branch.

tests/gold_tests/pluginTest/combo_handler/combo_handler.test.py
```diff
@@ -104,12 +106,12 @@ def add_server_obj(content_type, path):
...
-ts.Disk.ctwl_cfg.AddLine("")
-ts.Disk.ctwl_cfg.AddLine("		text/javascript  # test side comment")
-ts.Disk.ctwl_cfg.AddLine("  application/javascript")
-ts.Disk.ctwl_cfg.AddLine("text/css")
+# Configure the combo_handler's configuration file.
+ts.Setup.Copy("ctwl.txt", ts.Variables.CONFIGDIR)
+ctwl_cfg_path = os.path.join(ts.Variables.CONFIGDIR, "ctwl.txt")
+ts.Setup.Chown(ctwl_cfg_path, "nobody", "nobody", ignore=True)
+# covers ubuntu's unprivileged group
+ts.Setup.Chown(ctwl_cfg_path, "nobody", "nogroup", ignore=True)
```
dragon512, line 114: I am a little worried about this as we found issues where nobody or nogroup causes errors.

tests/gold_tests/pluginTest/lua/lua_watermark.test.py
```diff
@@ -41,9 +43,16 @@
...
 ts.Disk.remap_config.AddLine(
     'map / http://127.0.0.1:{}/'.format(server.Variables.Port) +
-    ' @plugin=tslua.so @pparam={}/watermark.lua'.format(Test.TestDirectory)
+    ' @plugin=tslua.so @pparam=watermark.lua'
 )
 
+# Configure the tslua's configuration file.
+ts.Setup.Copy("watermark.lua", ts.Variables.CONFIGDIR)
+watermark_path = os.path.join(ts.Variables.CONFIGDIR, "watermark.lua")
+ts.Setup.Chown(watermark_path, "nobody", "nobody", ignore=True)
```
dragon512, line 52: I am seeing a number of these items... Why is trafficserver.ext not doing this correctly??

tests/gold_tests/pluginTest/url_sig/url_sig.test.py
```diff
@@ -63,30 +63,31 @@
 
 # Use unchanged incoming URL.
 #
+ts.Setup.Copy("url_sig.config", ts.Variables.CONFIGDIR)
 ts.Disk.remap_config.AddLine(
     'map http://one.two.three/ http://127.0.0.1:{}/'.format(server.Variables.Port) +
-    ' @plugin=url_sig.so @pparam={}/url_sig.config'.format(Test.TestDirectory)
```
dragon512, line 68: just a general note... This could be clean up more via saying 
`f'map http://one.two.three/ http://127.0.0.1:{server.Variables.Port}/' +`

This can be done to a number of cases below

dragon512 (commented):
I leaving a few comments.. not blocking this

dragon512 (dismissed):

tests/gold_tests/pluginTest/combo_handler/combo_handler.test.py
bneradt: This was needed because the esi plugin used default mode of read and write. Fixes the plugin obviates the need for these chowns.

tests/gold_tests/pluginTest/lua/lua_watermark.test.py
bneradt: These Chowns are actually not needed.

bneradt:
> I leaving a few comments.. not blocking this

Thank you for the thoughtful feedback, @dragon512 !

Rather than Chown twice throughout the extension, I wrapped this in a helper function. The only test that needs the Chown now is the runroot one which, due to using a one-off runroot, needs the Chown for the test. It calls the helper function.

bneradt:
@shukitchan Can you please review this as well? I have a tweak to the esi combo_handler.

tests/gold_tests/pluginTest/url_sig/url_sig.test.py
bneradt: Yes, I always use this for new code. But I kept the method consistent with the rest of the file which generally uses the older syntax.

bneradt: Actually, for this file it unambiguously improves the file. I'll apply it throughout.

shukitchan (approved):
good change to do on esi combo handler allow list.

ezelkow1 (approved):
looks good, in the future we should add a `user` option to autest to specify the user to run as for these directories. That way if a user builds ats with --with-user=insert_user, they can also run autest against that build with the same insert_user and not hit permissions issues

bneradt:
> looks good, in the future we should add a `user` option to autest to specify the user to run as for these directories. That way if a user builds ats with --with-user=insert_user, they can also run autest against that build with the same insert_user and not hit permissions issues

Yeah, that's a good idea. This change puts us a step in that direction with the `chownForATSProcess` function added.

dragon512 (approved):

PR #7673, opened 2021-04-02
Change debug tags and make sure sni is printed on certain logs

duke8253:


zwoop:
[approve ci]

zwoop:
[approve ci autest]

zwoop:
[approve ci].

zwoop:
[approve ci].

zwoop:
[approve ci].

zwoop:
[approve ci].

ezelkow1:
[approve ci]

iocore/net/SSLClientUtils.cc
```diff
@@ -82,6 +82,11 @@ verify_callback(int signature_ok, X509_STORE_CTX *ctx)
       } else {
         sni_name = buff;
       }
+      if (is_debug_tag_set("ssl_verify")) {
+        Debug("ssl_verify", "Core server certificate verification failed for (%s). Action=%s Error=%s server=%s(%s) depth=%d",
+              sni_name, enforce_mode ? "Terminate" : "Continue", X509_verify_cert_error_string(err),
+              netvc->options.ssl_servername.get(), buff, depth);
+      }
```
bneradt, line 89: Are you wrapping these in the conditional because the computation of the parts of the message are expensive?

bneradt (approved):

maskit (commented):
Changing tag is probably ok, but changing `Warning` to `Debug` may be not. Warnings are printed regardless of `diags.debug.enabled` and `diags.debug.tags`, and those would be disappeared on prod. If those warnings are not really important on prod then I'm ok with making them debug log, but users watching the warnings would be surprised. We might want to mark this incompatible. I didn't check how `SSLError` works but I have the same concern on it as well.

duke8253:
> Changing tag is probably ok, but changing `Warning` to `Debug` may be not. Warnings are printed regardless of `diags.debug.enabled` and `diags.debug.tags`, and those would be disappeared on prod. If those warnings are not really important on prod then I'm ok with making them debug log, but users watching the warnings would be surprised. We might want to mark this incompatible. I didn't check how `SSLError` works but I have the same concern on it as well.

I thought about that too, but the message logged in Diags also seems reasonable to be in debug logs as well. We will have our ops team test this out for a few days and see what they prefers, I'll make adjustments according to their comments.

shinrich:
Oh, I didn't catch that the Warning messages were going away.  I thought you were just adding Debug messages in parallel so in the more detailed case all would show up in traffic.out.

I know that our ops guys rely on the certificate verification messages always showing up.  They use that to check when to yell at origins for having bad certs.  They don't run with debug enabled enough for that to be equivalent.

bneradt:
[approve ci]

bryancall:
@randall is going to review this

duke8253:
[approve ci autest]

randall (approved):

PR #7675, opened 2021-04-03
Automatically marks PRs and issues stale

bryancall:
Marks PRs and issues with the Stale tag and makes a comment on the PR and issue.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,6 +1,9 @@
 name: Mark stale issues and pull requests
 
 on:
+  push:
+    paths:
+      - .github/workflows/stale.yml
   schedule:
   - cron: "30 1 * * *"
 
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -16,8 +16,8 @@ jobs:
     - uses: actions/stale@v3
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        stale-issue-message: 'This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.'
-        stale-pr-message: 'This pull request has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.'
+        stale-issue-message: 'This issue has been automatically marked as stale because it has not had recent activity. Marking it stale to flag it for further consideration by the community.'
+        stale-pr-message: 'This pull request has been automatically marked as stale because it has not had recent activity. Marking it stale to flag it for further consideration by the community.'
         stale-issue-label: 'Stale'
         stale-pr-label: 'Stale'
         days-before-pr-stale: 90
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,24 @@
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+  - cron: "30 1 * * *"
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.'
+        stale-pr-message: 'This pull request has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.'
+        stale-issue-label: 'Stale'
+        stale-pr-label: 'Stale'
+        days-before-pr-stale: 90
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,24 @@
...
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.'
+        stale-pr-message: 'This pull request has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.'
+        stale-issue-label: 'Stale'
+        stale-pr-label: 'Stale'
+        days-before-pr-stale: 90
+        days-before-issue-stale: 365
+        days-before-pr-close: -1
+        days-before-issue-close: -1
```
bneradt, line 26: Looks like this can be consolidated to `days-before-close: -1` if we want.

On the other hand, having both explicitly stated (as you have it here) can be helpful in the future if we want to tweak one of these. Either way is fine with me.

shukitchan (dismissed):
this is good.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,24 @@
...
+  stale:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.'
+        stale-pr-message: 'This pull request has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.'
```
bneradt, line 20: Just to make sure I'm reading this correctly: the message says it will be closed if no action takes place, but it looks like this mechanism will not automatically close it. Is that right? I assume the -1 values for days before close mean that it will not be closed by this system. I suppose the intention is that things marked stale will be manually triaged and potentially closed if deemed appropriate.

We might want to re-word the issue message to express our intention. When I submit a PR, I generally have responsibility to shepherd my PR to completion. For instance, I currently have an old PR that's relatively low priority for the project and myself, but it will have marginal value and I intend to get it done eventually. But the fact that it is stale by these standards is my responsibility as the submitter of the PR. If I get this message on my PR, I will feel nudged to do something with it, and that seems good. I should take some action.

I wonder, on the other hand, how someone will feel if an issue they filed gets this message. To them it was probably an annoyance or something wrong enough such that they felt a responsibility to raise the issue to the community. They may be a user rather than a developer, and getting a message saying that their issue will be closed if there is no activity may make them feel a bit miffed: there's not much they can do to resolve the issue. Perhaps they can re-articulate things, but beyond that they might feel helpless. 

Perhaps something like:

'This issue has been automatically marked as stale because it has not had recent activity. Marking it stale to flag it for further consideration by the community."

bryancall:
@shukitchan Thanks for the review.  I updated it to it will also run when the file has been modified.  That way it is run when this PR is modified.  Right now I have it set to run in debug mode and won't modify the PRs and issues.  Debug mode should be removed before merging this PR.

bneradt (dismissed):
Thanks for working on this. This looks like it can be helpful for categorizing and flagging issues and PRs. For it to be useful we should probably decide to making it a part of some kind of regular process of ours to look at these. Nothing elaborate is needed, but perhaps we should have a monthly meeting to review these. We're currently meeting weekly, but hopefully once we get through the current set a monthly meeting will do.

.github/workflows/stale.yml
bryancall: My thought is we would automatically close PRs and issues when there has been no activity.  I didn't want to do that yet and just want to label them Stale and make a comment for people to possibly updated the PR and Issue.

It is more about testing this out since we don't have any actions, yet, that automatically edit PRs and issues.

bneradt: Understood. You might want to consider my comment about the wording of the issue message though and what that may communicate to the submitter. Unless my understanding of the intention is wrong.

.github/workflows/stale.yml
bryancall: Yes, in the future we should automatically close PRs and issues and those are just place holders in case we want to have specific times for each.

jrushford (dismissed):
+1 on @bneradt comment regarding the messaging.  Otherwise it looks fine to me.

jrushford (approved):
This looks good to me.  I ran the workflow and it operates fine and finds PR's and Issues that should be marked as stale.  I approve the change in the messaging.

bneradt (approved):
Looks good. Thanks Bryan.

PR #7678, opened 2021-04-05
sslheaders AuTest: Skip if plugin does not exist

### LINKED ISSUE #7666: sslheaders autest not checking for plugin
sslheaders is an experimental plugin but the test doesnt do an existence check for the plugin before running, so it will fail unless experimental plugins were built

bneradt:
Adding a skip condition on the sslheaders.test.py that verifies the
plugin exists (i.e., was configured and build) before running the test.

Fixes #7666

### DIFF
--- a/tests/gold_tests/pluginTest/sslheaders/sslheaders.test.py
+++ b/tests/gold_tests/pluginTest/sslheaders/sslheaders.test.py
@@ -22,6 +22,7 @@ Test sslheaders plugin.
 '''
 
 Test.SkipUnless(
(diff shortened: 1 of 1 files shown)


bneradt:
This is a trivial test-only AuTest update. Might as well pull it into 9.2.x to keep the test consistent across the branches. I'll file a separate PR for 8.1.x because I notice it doesn't cherry-pick cleanly to that branch.

ezelkow1 (approved):

bneradt:
cherry-pick PR to 8.1.x: #7681

PR #7682, opened 2021-04-05
Updating to AuTest 1.10.0

bneradt:
This new release supports case-insensitive gold files. We intend to use
this for HTTP/2 curl AuTest gold files which, depending upon the version
of curl, may or may not lowercase field names.

### DIFF
--- a/tests/Pipfile
+++ b/tests/Pipfile
@@ -26,7 +26,7 @@ pyflakes = "*"
 [packages]
 
 # Keep init.cli.ext updated with this required autest version.
-autest = "==1.8.1"
+autest = "==1.10.0"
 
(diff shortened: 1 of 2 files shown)


bneradt:
I'm marking this for all the supported branches for which we will run CI: we've found in Yahoo's CI that later versions of curl lower-case HTTP/2 field names, which is correct per the RFC. However, this means that older versions of curl will leave them upper case, while newer versions will lower them. To support these differences in output, I plan to make use of this new AuTest feature to use case-insensitive gold files. We'll want this for all the branches on which we run CI.

randall (approved):

zwoop:

zwoop:

PR #7685, opened 2021-04-05
Fix has_consumer_besides_client to deal with no clients

shinrich:
Introduced by PR #7641.   has_consumer_besides_clent should return false if there are no live consumers.

I ran into problems with the h2spec test on the h2 to origin branch once I rebased in this change.


masaori335 (approved):
My bad, I didn't assume the case. Thanks!

zwoop:

zwoop:

PR #7687, opened 2021-04-06
Use only one map in session cache instead of two

duke8253:
This should help reduce the CPU utilization some more, @shinrich found this when testing a new build in prod. The `std::unordered_map` was supposed to provide speed when `getSession` takes up majority of the session cache workload. But to retain the functionalities it had to work along side `std::map`, and looks like that is messing things up a bit.

```
   4.51%  traffic_server                                 [.] std::_Hashtable<SSLSessionID, std::pair<SSLSessionID const, SSLSession*>, std::allocator<std::pair<SSLSessionID const, SSLSession*> >, std::__detail::_Select1st, std:
   4.28%  libwaflzts.so                                  [.] ns_waflz::ac::find
   1.96%  libc-2.17.so                                   [.] __memcmp_sse4_1
```


randall:
[approve ci]

randall:
[approve ci]

maskit:
Is it guaranteed that session id monotonically increases?

duke8253:
> Is it guaranteed that session id monotonically increases?

You're right, for some reason I thought std::map keeps insertion order not the key order. Looks like I'll have to implement something of my own, closing this PR for now till I have more time to think this through.

PR #7694, opened 2021-04-08
Deny unknown transfer encoding values

### LINKED ISSUE #7590: How to handle non-chunked transfer-encoding headers
As part of PR #7473, the Transfer-Encoding header was marked hop-by-hop.  This means all Transfer-encoding headers were are removed before the proxy request is created.  Transfer-encoding headers are only added back as appropriate for the sending from ATS to the origin.

This change was made to address issue #6907.

The mozilla specification https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding indicates that the Transfer-Encoding is a hop-by-hop header, so assuming we interpret the incoming transfer-encoding header the change in PR #7473 is correct and lines up with the spec.

However @bryancall pointed out that ATS really only interprets the chunked value of the Transfer-Encoding header.  According to the Mozilla document there are several other values that deal with compressing the data as it is transferred.

```
Transfer-Encoding: compress
Transfer-Encoding: deflate
Transfer-Encoding: gzip
Transfer-Encoding: identity
```

In addition in our logic to translate HTTP/1.1 headers to HTTP/2, the transfer-encoding header is stripped.  Which again is fine for chunked, but ignores any compression-oriented values of the header.

It seems these days that the vast majority of uses of transfer-encoding headers are just chunked.  In fact it is very hard to find articles discussing the other options.  So we don't need a particularly efficient solution here.  In fact it may be sufficient to just log warnings when we drop these headers.

shinrich:
This PR implements the deny approach outlined in issue #7590.

This closes #7590

### DIFF
--- a/proxy/http/HttpSM.cc
+++ b/proxy/http/HttpSM.cc
@@ -2059,7 +2059,7 @@ HttpSM::state_read_server_response_header(int event, void *data)
   case PARSE_RESULT_DONE:
 
     if (!t_state.hdr_info.server_response.check_hdr_implements()) {
-      t_state.http_return_code = HTTP_STATUS_NOT_IMPLEMENTED;
+      t_state.http_return_code = HTTP_STATUS_BAD_GATEWAY;
       call_transact_and_set_next_state(HttpTransact::BadRequest);
       break;
     }
--- a/proxy/http/HttpSM.cc
+++ b/proxy/http/HttpSM.cc
@@ -848,6 +848,12 @@ HttpSM::state_read_client_request_header(int event, void *data)
   case PARSE_RESULT_DONE:
     SMDebug("http", "[%" PRId64 "] done parsing client request header", sm_id);
 
+    if (!t_state.hdr_info.client_request.check_hdr_implements()) {
+      t_state.http_return_code = HTTP_STATUS_NOT_IMPLEMENTED;
+      call_transact_and_set_next_state(HttpTransact::BadRequest);
+      break;
+    }
+
     if (_from_early_data) {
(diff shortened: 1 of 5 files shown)


maskit (dismissed):

proxy/http/HttpSM.cc
```diff
@@ -848,6 +848,12 @@ HttpSM::state_read_client_request_header(int event, void *data)
   case PARSE_RESULT_DONE:
     SMDebug("http", "[%" PRId64 "] done parsing client request header", sm_id);
 
+    if (!t_state.hdr_info.client_request.check_hdr_implements()) {
+      t_state.http_return_code = HTTP_STATUS_NOT_IMPLEMENTED;
```
bryancall, line 852: We should be responding with 415

https://tools.ietf.org/html/rfc7231#section-3.1.2.2
   An origin server MAY respond with a status code of 415 (Unsupported
   Media Type) if a representation in the request message has a content
   coding that is not acceptable.

bryancall (changes requested):

proxy/http/HttpSM.cc
shinrich: Dueling RFC's.  I'm moving the discussion back to issue #7590.

bryancall (approved):

shinrich:
Based on the discussion in the issue and the discussion in yesterday's PR review pushed a commit to change the status code to 502 (Bad gateway), if the origin sends us a transfer_encoding header with a non-chunked value.

bryancall (approved):

zwoop:

Issue #7241, opened 2020-10-06
Pre-warm TLS Tunnel


masaori335:
# Abstract
On TLS Forward Tunnel or TLS Partial Blind Tunnel, clients always have to wait while ATS establishes connections with origin servers. An idea to fix this issue is "pre-warm" connections between ATS and origin servers before clients make requests to ATS.

The overview is below.
<img width="740" alt="95164780-98f48900-07e5-11eb-9393-c35a6de7e9ae" src="https://user-images.githubusercontent.com/741391/95164919-ecff6d80-07e5-11eb-8eb9-ac6a3266811f.png">

Using the pre-warmed connection, clients can establish TLS Forward Tunnel or TLS Partial Blind Tunnel immediately. We can expect this improves time to first byte drastically.

# Status/Plan
We're testing this concept, and we found some related features are required. The development branch becomes bigger than expected includes these features, so I'll break it down into PRs per feature. All PRs will be linked to this issue.

# Features
- [x] Log negotiated ALPN: #7491
- [x] ALPN Support on TLS Partial Blind Tunnel: #7511
- [x] Outbound ProxyProtocol v1/v2: #7444
- [x] Pristine SNI: #7533
- [x] Overwrite SNI Policy from sni.yaml: #7703
- [x] Dynamic Stats: #7704
- [x] Active Tunnel Stats: #8323 
- [x] Pre-warm TLS Tunnel: #7661

# Related PRs
- [x] Cleanup #7453

PR #7703, opened 2021-04-13
Override proxy.config.ssl.client.sni_policy from sni.yaml

masaori335:
Part of #7241.


shinrich (approved):
Looks good.  So this is needed in the prewarming case because the actual HttpSM is just a blind tunnel and it has no proxy.config arguments of its own?

masaori335:
That's right.

PR #7704, opened 2021-04-13
Add DynamicStats

masaori335:
This is the core version of TSStat* APIs to register stats dynamically - e.g. based on host name, parent name, etc.

I wrote this for pre-warmimg pool based stats for Pre-warming TLS Tunnel. You can see some usages on #7661.

https://github.com/apache/trafficserver/blob/a20ba1993762c2d8cdf9443a8354f26d1a5ce2fd/proxy/http/PreWarmManager.cc#L1159

https://github.com/apache/trafficserver/blob/a20ba1993762c2d8cdf9443a8354f26d1a5ce2fd/proxy/http/PreWarmManager.cc#L701-L704

Part of #7241.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,6 +90,7 @@ src/tscore/test_X509HostnameValidator
 src/tscore/test_tscore
 src/tscpp/util/test_tscpputil
 lib/records/test_librecords
+lib/records/test_librecords_on_eventsystem
 lib/perl/lib/Apache/TS.pm
 
 iocore/net/test_certlookup
--- a/lib/records/DynamicStats.h
+++ b/lib/records/DynamicStats.h
@@ -1,0 +1,170 @@
+/** @file
+
+  Dynamic Stats
+
+  @section license License
+
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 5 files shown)


include/tscore/DynamicStats.h
```diff
@@ -0,0 +1,153 @@
...
+    int r = RecRegisterRawStat(_rsb, rec_type, name, data_type, RECP_PERSISTENT, stat_id, sync_cb);
+    if (r != REC_ERR_OKAY) {
+      return -1;
+    }
+  } else {
+    int r = RecRegisterRawStat(_rsb, rec_type, name, data_type, RECP_NON_PERSISTENT, stat_id, sync_cb);
+    if (r != REC_ERR_OKAY) {
+      return -1;
+    }
+  }
```
ywkaras, line 90: Why not:
```
  int r = RecRegisterRawStat(_rsb, rec_type, name, data_type, is_persistent ? RECP_PERSISTENT : RECP_NON_PERSISTENT, stat_id, sync_cb);
  if (r != REC_ERR_OKAY) {
    return -1;
  }
```
?

masaori335: Because it's a template argument.

https://github.com/apache/trafficserver/blob/e91ca3caf92554c57c5188485682382a44bb9a69/lib/records/I_RecProcess.h#L50-L51

https://github.com/apache/trafficserver/blob/e91ca3caf92554c57c5188485682382a44bb9a69/lib/records/I_RecDefs.h#L97

masaori335: Is it better to make this function template like below?
```
  template <RecPersistT T>
  int
  create(RecT rec_type, const char *name, RecDataT data_type, RecRawStatSyncCb sync_cb)
```

ywkaras: Usiing the if statement rather than a template would probably generate less object code:  https://godbolt.org/z/31oY5zMsn .

masaori335: OK, let's go with the if statement.

masaori335:
Pushed new commit. Please take another look.
- Moved `DynamicStats.h` from `include/tscore/` to `lib/records/`
- Add simple unit tests

masaori335:
[approve ci autest]

masaori335:
[approve ci]

masaori335:
[approve ci]

ywkaras (approved):

PR #7706, opened 2021-04-13
Clean up the logic storing the last negotiated server version

shinrich:
Clean up the logic to store the last negotiated server http protocol.  Originally in the HTTP2 to origin PR #7622.

Also removing the extra HTTP 0.9 logic since I don't think we support 0.9 anymore.


shinrich:
Closing in deference to #7766

PR #7711, opened 2021-04-16
Provide macros for easy logging of string_view/TextView values.

ywkaras:
Using printf-like functions.

### DIFF
--- a/include/tscore/Diags.h
+++ b/include/tscore/Diags.h
@@ -194,3 +194,17 @@ extern inkcoreapi Diags *diags;
 #define is_diags_on(_t) 0
 
 #endif // TS_USE_DIAGS
+
+#if !defined(TS_SVP)
+
+// Macros for convinient output of string_view (and thus also ts::TextView) instances using printf-like
+// functions.
+//
+#define TS_SVPF "%.*s"
+#define TS_SVP(SV__) static_cast<int>((SV__).size()), (SV__).data()
+
+// Usage example:
+// std::string_view yada;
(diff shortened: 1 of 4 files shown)


ywkaras:
[approve ci]

include/tscore/Diags.h
```diff
@@ -194,3 +194,17 @@ extern inkcoreapi Diags *diags;
 #define is_diags_on(_t) 0
 
 #endif // TS_USE_DIAGS
+
+#if !defined(TS_SVP)
+
+// Macros for convinient output of string_view (and thus also ts::TextView) instances using printf-like
+// functions.
+//
+#define TS_SVPF "%.*s"
```
bneradt, line 203: This functions kind of like the `PRIuMAX` and related types, right? Maybe call this TS_PRIsv?

bneradt (changes requested):
Thanks Walt. This looks helpful. I'd consider using it.

maskit:
Do you mind if I keep using the standard unsweetened way? I'd rather wait for C++20 and remember std::format format.

ywkaras:
Do we have a target date to switch to C++20?

Using string_view instead of std::string avoids a heap allocation.  The problem is that it's easier to do printf-line logging output with std::string than with string_view.

ywkaras:
Looks like std::format outputs an std::string instance.  What are the advantages over using std::stringstream?

maskit:
I'm not suggesting to use std::format for debug logs. I'm saying I don't want to remember the TS specific format specifiers and the macro just for one less parameter.

Issue #7712, opened 2021-04-16
Differences in communicating TLS client cert failure with TLS 1.3


shinrich:
Found this while working with @SolidWallOfCode on his down-server branch.  He had just rebased and the second test in tls_client_cert autest started failing on him.  This test should fail because ATS is presenting a client cert to the test origin that is not signed by a CA used by that test origin.

Specifically the test was failing due to an assert that t_state.current.server->had_connect_fail() should be true.  When I ran the test on Centos7, the test passed.  When @SolidWallOfCode ran the test in the same branch on Fedora, the assert would occur.

We did packet captures on both.  The CentOS7 run had negotiated TLS1.2, and looking at the packets, the origin sent TLS assert and rest before the ATS sent any data packets.

The Fedora run had negotiated TLS1.3, so looking at the packets was not so informative (TLS1.3 encrypts the handshake packets as soon as it can), but from looking at the size of packets and looking at the path through the code, it appears that the client handshake succeeded as soon as it sent out the client cert and key exchange message.  Then it moved on and sent out the request to the origin.  Then the origin sent a TLS alert about the bad handshake and resets the connection.  So from ATS's perspective, the connection initiation succeeded and the the origin failed the connection during the data exchange.

When @SolidWallOfCode adjusted the test so ATS would not negotiate TLS1.3 (proxy.config.ssl.client.TLSv1_3), the test passed on Fedora.

I will put up a PR to make sure we are propagating the connection failures and general original failures appropriately.  @SolidWallOfCode will adjust the assert in his branch.  Beyond that, it is not clear there is much more to fix.  Filing the issue, because it seems like a change in an edge case that we should be aware of.

PR #7714, opened 2021-04-16
Propagate TLS errors

shinrich:
Addresses some issues found while investigating issue #7712


iocore/net/SSLNetVConnection.cc
```diff
@@ -546,7 +546,7 @@ SSLNetVConnection::net_read_io(NetHandler *nh, EThread *lthread)
   // If the key renegotiation failed it's over, just signal the error and finish.
   if (sslClientRenegotiationAbort == true) {
     this->read.triggered = 0;
-    readSignalError(nh, static_cast<int>(r));
+    readSignalError(nh, (errno) ? errno : -ENET_SSL_FAILED);
```
maskit, line 549: Which function's error is this checking?

iocore/net/SSLNetVConnection.cc
```diff
@@ -621,7 +621,7 @@ SSLNetVConnection::net_read_io(NetHandler *nh, EThread *lthread)
           Debug("ssl", "ssl handshake for vc %p, expired, release the connection", this);
           read.triggered = 0;
           nh->read_ready_list.remove(this);
-          readSignalError(nh, VC_EVENT_EOS);
+          readSignalError(nh, EPIPE);
```
maskit, line 624: I don't think EOS is correct, but EPIPE doesn't seem appropriate as well. I'd use ETIMEDOUT if we can change it to any error code and want to choose one from standard ESOMETHING error codes.

iocore/net/SSLNetVConnection.cc
shinrich: This is based on the regeneration failure set in ssl_callback_info.  Probably best as just ENET_SSL_FAILED rather than trying to pick up the random errno along the way.

iocore/net/SSLNetVConnection.cc
shinrich: You are correct.  Will fix this up.

shinrich:
Pushed another commit to address @maskit's comments

iocore/net/SSLNetVConnection.cc
```diff
@@ -736,7 +736,7 @@ SSLNetVConnection::net_read_io(NetHandler *nh, EThread *lthread)
     break;
   case SSL_READ_ERROR:
     this->read.triggered = 0;
-    readSignalError(nh, static_cast<int>(r));
+    readSignalError(nh, (errno) ? errno : -ENET_SSL_FAILED);
```
maskit, line 739: What abut this one? I guess it's checking errno set by `ssl_read_from_net`, but it can be changed in `readSignalAndUpdate`. I'm not opposed to propagating errno but we should secure the value right after calling a function.

shinrich: Afraid I don't understand the comment.  We are passing the errno that would have been generated from ssl_read_from_net.  It may change during the call to readSignalError, but the argument will be the value from the call point.  Would you like me to save the errno value directly after the all to ssl_read_from_net?

maskit: > Would you like me to save the errno value directly after the all to ssl_read_from_net?

Not sure if it has to be all, but yes. A problem is that `errno` could be changed if I added some code between `ssl_read_from_net` (or `readSignalError`) and this line. If that's another I/O related code then it may be ok, but it might be something for completely different things (e.g. debugging, logging, etc).

So the current code is fine. It just looks fragile.

shinrich: pushed a commit to introduce a local var to keep the ssl read errno.

maskit (approved):

zwoop:

PR #7437, opened 2021-01-21
Small fix to regex_remap PR # 7347.

ywkaras:

PR #7347, opened 2020-11-21
Allow for regex_remap of pristine URL.

ywkaras:
Adds a new optional pparam "pristine" to the regex_remap core plugin.  When present, this pparam causes
regex_remap to match on the pristine URL, and to use the pristine URL for $ substitutions.

PR #7718, opened 2021-04-19
Fix simple remapping in regex_remap plugin.

ywkaras:
Adds a test run to successfully remap a URL with host example.three, resulting in a 200 response.
It includes a bug fix for regex_remap.cc, needed so the test run will succeed.  The bug was probably
introduced by commit fee5ba1.


bryancall:
@ywkaras Please reference the PR that broke this and write up a description.

ywkaras:
Verizon Media production has tested and approved this change.

randall (approved):

randall:
Should this also go into 9.0.x since the other 2 related PRs (#7347 and #7437) went there?

zwoop:

PR #7719, opened 2021-04-19
Fix cpu utilization problem in session cache

duke8253:
This should help reduce the CPU utilization some more, @shinrich found this when testing a new build in prod. The `std::unordered_map` was supposed to provide speed when `getSession` takes up majority of the session cache workload. But to retain the functionalities it had to work along side `std::map`, and looks like that is messing things up a bit.

```
   4.51%  traffic_server                                 [.] std::_Hashtable<SSLSessionID, std::pair<SSLSessionID const, SSLSession*>, std::allocator<std::pair<SSLSessionID const, SSLSession*> >, std::__detail::_Select1st, std:
   4.28%  libwaflzts.so                                  [.] ns_waflz::ac::find
   1.96%  libc-2.17.so                                   [.] __memcmp_sse4_1
```

After testing a few different ways to store sessions, `std::unordered_map` always takes up around 4% cpu whenever it's paired with other data structures, but by itself it's not even showing up in the perf top list. Since session id reuse rate is extremely low now (around 0.1-0.2% in our production), I think it's ok to just remove any random session in the cache. And after testing, removing random sessions still provided around 0.1-0.2% reuse rate.


duke8253:
This is however still a problem in origin session cache, but we're just starting to deploy the origin session cache in production, will need some time to test it out before making any change.

maskit:
> std::unordered_map always takes up around 4% cpu whenever it's paired with other data structures, but by itself it's not even showing up in the perf top list.

Are you saying erasing a session by key (a session id obtained from the other data structure) in removeOldestSession was the heavy part? It implies find operation to get a cached session is heavy too, but you don't see it on the perf top list?

duke8253:
> > std::unordered_map always takes up around 4% cpu whenever it's paired with other data structures, but by itself it's not even showing up in the perf top list.
> 
> Are you saying erasing a session by key (a session id obtained from the other data structure) in removeOldestSession was the heavy part? It implies find operation to get a cached session is heavy too, but you don't see it on the perf top list?

No, I'm saying that whenever `std::unordered_map` is used with something else, such as `map` or `queue` or `list` or `CountQueue`, the utilization goes up unexpectedly. And the reason for needing a `queue` is because in order to implement `removeOldestSession` we need to keep track the insertion order.

maskit:
How does the utilization go up by using std::unordered_map with something else? I'd understand if the something else took  up CPU, but you say the unordered_map itself took up CPU if you use it with something else. The only difference I can think of was the way to erase an entry from std::unordered_map. By Key from the something else or by Iterator from std::unordered_map::begin.

duke8253:
> How does the utilization go up by using std::unordered_map with something else? I'd understand if the something else took  up CPU, but you say the unordered_map itself took up CPU if you use it with something else. The only difference I can think of was the way to erase an entry from std::unordered_map. By Key from the something else or by Iterator from std::unordered_map::begin.

That could explain why the utilization is high, but I don't think it is the problem here. As I understand it, erase by key is a combination of find by key then erase by iterator. Searching via key is used inside insert to make sure no duplications, one more search should not increase the utilization by that much.

maskit:
> That could explain why the utilization is high, but I don't think it is the problem here.

I think it is, because erasing random entries is not a right thing to do. We shouldn't cut corners.

I found that `find()` (or an equivalent that is internally called inside `erase()`) is actually not that fast and it's waste of time if we don't use the found entry. The code below is not a very fair comparison, but test1 is the original, test2 is the proposed approach, and test3 is another approach. Performance of test2 and test3 are pretty similar but test3 erases old entries first. Would you try it out?
```cpp
#include <unordered_map>
#include <random>
#include <ctime>
#include <iostream>

constexpr int N = 20000000;

int randomKey()
{
    return std::rand() / (RAND_MAX / 1000000);
}

void test1()
{
    std::unordered_map<int, int> map;
    int miss = 0;
    map.reserve(1000);
    int prev;
    for (int i = 0; i < N; ++i) {
        if (map.size() > 1000) {
            map.erase(prev);
        }
        map.insert({prev = randomKey(), 1});

        if (map.find(randomKey()) == map.end()) {
            ++miss;
        }
    }
    std::cout << miss << std::endl;
}
void test2()
{
    std::unordered_map<int, int> map;
    int miss = 0;
    map.reserve(1000);

    for (int i = 0; i < N; ++i) {
        if (map.size() > 1000) {
            map.erase(map.begin());
        }
        map.insert({randomKey(), 1});

        if (map.find(randomKey()) == map.end()) {
            ++miss;
        }
    }
    std::cout << miss << std::endl;
}
void test3()
{
    std::unordered_map<int, int> map1;
    std::unordered_map<int, int> map2;
    std::unordered_map<int, int> *current_map = &map1;
    std::unordered_map<int, int> *old_map = &map2;
    int miss = 0;
    map1.reserve(500);
    map2.reserve(500);

    for (int i = 0; i < N; ++i) {
        if (current_map->size() > 500) {
            old_map->clear();
            std::swap(current_map, old_map);
        }
        current_map->insert({randomKey(), 1});

        auto k = randomKey();
        if (current_map->find(k) == current_map->end()) {
            if (old_map->find(k) == old_map->end()) {
                ++miss;
            }
        }

    }
    std::cout << miss << std::endl;
}

int main(int argc, char **argv) {
    std::srand(std::time(0));
    test1();
    test2();
    test3();
    return 0;
}
```

duke8253:
I added another test
```
void test4()
{
    std::unordered_map<int, int> map;
    int miss = 0;
    map.reserve(1000);
    int prev;
    for (int i = 0; i < N; ++i) {
        if (map.size() > 1000) {
            auto node = map.find(prev);
            map.erase(node);
        }
        map.insert({prev = randomKey(), 1});

        if (map.find(randomKey()) == map.end()) {
            ++miss;
        }
    }
    std::cout << miss << std::endl;
}
```
perf-top shows the following:
test1: 6.9~7.1%
test2: 4.6~4.8%
test3: 7.4~7.6%
test4: 6.3~6.5%

I also thought about swapping things around like in test 3, but decided not to since it requires more memory space in order to keep the same amount of entries. Now it seems that it requires more computation as well.

maskit:
Thank you for testing. The perf-top result is from real ATS session cache code, right? It's interesting because test3 is faster than test1 if I run the test code.

duke8253:
It's not from real ats code. The implementation of test3 is in fact a proposal I made internally as well, I called it batch delete since it essentially removes multiple entries at once. It will be faster no doubt since instead of doing a remove and insert every time a new session arrives, it removes only once when the cache is full. But the problem with that is we're essentially shrinking the size of the cache. e.g. cache size is 1000, but once we reach 1000 the first 500 will be removed, so the next we search we only have 501 in the cache not the 1000 we promised.

maskit:
I understand that batch delete would require double of memory space than current if we keep the same number of sessions in the cache, but would it be unacceptably large?

I didn't put this because I thought we'll see similar results, but this is what I got for test 1, 2 and 3.
image

I do think ATS shouldn't spend CPU time for session cache eviction too much, but a sad thing of this change is that we are completely losing entry prioritization. Prioritizing entries by creation time may be not a good strategy if the cache hit rate is as bad as random eviction. But prioritizing entries by other info (e.g. protocol version, algorithm, User-Agent, etc.) may work and it may reduce CPU usage in total if it improves cache hit rate, which could be done by just changing the key of `std::map`.

I'd have to test this on prod if it changes the functionalities.

duke8253:
Looks like using `std::map` doesn't have the high cpu utilization problem.

duke8253:
[approve ci autest]

duke8253:
[approve ci clang-analyzer]

shinrich:
these changes fixed the increased CPU utilization I noticed when running 9.1 in production

maskit (approved):
Interesting. `std::map::erase(key)` is lighter than `std::unordered_map::erase(key)` for this? It may depends on how many entries we keep. In any cases, the latest code looks more reasonable than random eviction. I didn't compare the performance difference between the two but I'm fine with the latest code if it fixes the CPU utilization problem.

PR #7724, opened 2021-04-20
if transaction status non-success, bypass intercept plugin

randall:
Don't allow the generator or statichit plugin to intercept the transaction
if a prior plugin has changed the transaction status to something non-success.

### DIFF
--- a/plugins/experimental/statichit/statichit.cc
+++ b/plugins/experimental/statichit/statichit.cc
@@ -563,6 +563,12 @@ TSRemapInit(TSRemapInterface * /* api_info */, char * /* errbuf */, int /* errbu
 TSRemapStatus
 TSRemapDoRemap(void *ih, TSHttpTxn rh, TSRemapRequestInfo *rri)
 {
+  const TSHttpStatus txnstat = TSHttpTxnStatusGet(rh);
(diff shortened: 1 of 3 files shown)


plugins/experimental/statichit/statichit.cc
```diff
@@ -563,6 +563,12 @@ TSRemapInit(TSRemapInterface * /* api_info */, char * /* errbuf */, int /* errbu
 TSRemapStatus
 TSRemapDoRemap(void *ih, TSHttpTxn rh, TSRemapRequestInfo *rri)
 {
+  const TSHttpStatus txnstat = TSHttpTxnStatusGet(rh);
+  if (0 != static_cast<int>(txnstat) && txnstat != TS_HTTP_STATUS_OK) {
```
ywkaras, line 567: Why not:
```
if (TS_HTTP_STATUS_NONE != txnstat && txnstat != TS_HTTP_STATUS_OK) {
```
?

randall: Good point. That's what I get for copy/pasting.

randall: Fixed. And added a bonus cleanup based on this feedback.

shinrich (approved):
Seems reasonable.

zwoop:

PR #7725, opened 2021-04-20
compress.test.py: Reference config file from Test.RunDirectory

bneradt:
In certain environments in which AuTest is run, Test.TestDirectory is
not readable to the ATS process. Alter the test to reference the
compress.config file from the sandbox, which is curated to be readable
by the ATS process.


bneradt:
This is a simple change that can save people hassle when running the AuTests in a location not readable by the nobody ATS process user. If it doesn't cherry-pick cleanly to any of these branches, not a big deal. I think it should cherry-pick cleanly though.

randall (approved):

ezelkow1 (approved):
Tested on the new CI which was having this issue

zwoop:

Issue #7656, opened 2021-03-30
Build fails on FreeBSD 13+


Johanhen:
The build of traffic server 8.1.1 (and probably other versions as well fails on FreeBSD 13.x and newer.

As noted by Brian in the mailing list

FreeBSD 13 added eventfd(2) [1], which means that the check on HAVE_EVENTFD in EventNotify.cc [2] can no longer assume that if you have eventfd(2) you also have epoll.

1. https://github.com/freebsd/freebsd-src/commit/7a202823aa54ba18c485bdbcf355269bcfee1ab9#diff-7ea785358f7ef658d52a5d985ea53f03c09cd0bfee8e6a7403063774d3dc58a8
2. https://github.com/apache/trafficserver/blob/3d266379b816227928171d1f07de747f6a9a254b/src/tscore/EventNotify.cc#L34

Johanhen:
Unfortunately i still receive that error after the replacement off #ifdef HAVE_EVENTFD by #if defined(HAVE_EVENTFD) && defined(TS_USE_EPOLL)

[...]

PR #7730, opened 2021-04-22
Fix build on FreeBSD 13

bgaff:
FreeBSD 13 added eventfd(2), which means that now having eventfd no
longer means that you'll also have epoll. This change updates
EventNotify to check for both eventfd(2) and epoll before trying to
use epoll.

Issue #7656


SolidWallOfCode (approved):

bryancall (approved):

zwoop:

PR #7736, opened 2021-04-23
Augment test cases for tls_verify_override test

shinrich:
Added a case for verify.properties = NULL and verify.properties = ENFORCED.  A customer had an issue with this case and I realized that we had not test case for this combination.  Went ahead and reviewed all the cases while I was in there.


shinrich:
[approve ci autest]

rob05c (approved):
LGTM. Test code looks good, makes sense. Ran tests, they passed.

randall (approved):

PR #7742, opened 2021-04-23
Adding DNS TTL AuTests.

bneradt:
Add AuTests that verify the behavior of ATS when DNS goes down. The following configurations are exercised in this test:

            "proxy.config.hostdb.ttl_mode"
            "proxy.config.hostdb.timeout"
            "proxy.config.hostdb.serve_stale_for"


vmamidi (approved):

zwoop:

PR #7744, opened 2021-04-26
Update Nexthop strategies to limit the number of simultaneous host retries

jrushford:
Updates to Nexthop strategies to limit the number of simultaneous
transactions that will retry a parent once it's retry_time window
has elapsed.  The limiting prevents thundering retries on a parent
that is still failed and would cause many transactions to hang or
take much longer than necessary to use a good parent without the
limiting.  This PR mirrors the limiting added to Parent selection.

The maximum number of retriers is configured  with
'proxy.config.http.parent_proxy.max_trans_retries' which defaults to 2.

### DIFF
--- a/proxy/http/remap/NextHopConsistentHash.cc
+++ b/proxy/http/remap/NextHopConsistentHash.cc
@@ -302,20 +302,26 @@ NextHopConsistentHash::findNextHop(TSHttpTxn txnp, void *ih, time_t now)
       host_stat = TS_HOST_STATUS_UP;
     }
   }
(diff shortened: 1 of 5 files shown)


ywkaras (approved):

jrushford:
[approve ci]

jrushford:
@zwoop this should be cherry picked to 9.1 and 9.0

zwoop:
@jrushford Neither 9.0.x or 9.1.x cherry-pick's this cleanly... If we need these for those branches, either let me know what other PRs are needed (pre-reqs), or make new back port PRs. My preference is generally the former, identifying the dependencies.

PR #7748, opened 2021-04-27
Remove proxy.config.http.down_server.abort_threshold

### LINKED ISSUE #7735: Review default value of proxy.config.http.down_server.abort_threshold = 10 seconds
@shinrich noticed default value of 10s for proxy.config.http.down_server.abort_threshold might be a little low.  Example,  If an HTTPS API call takes longer to return a response than 10s, the server is marked down. 

Specifically, how does this setting interact with no_activity timeouts like proxy.config.http.transaction_no_activity_timeout_out ?  We were having origin connections erroneously marked dead until we  set proxy.config.http.down_server.abort_threshold to exceed proxy.config.http.transaction_no_activity_timeout_out - effectively turning the setting off.   This made me wonder if this setting is useful at all. 

Even after this change, very occasionally ran into conflict with proxy.config.net.default_inactivity_timeout = 300 as well.

shinrich:
This setting is confusing and it could cause origins to be marked down unexpectedly.  I had originally proposed increasing the default, but in yesterday's PR review the consensus was just to remove the setting and possibly taking the incompatibility knock to bring this back.  

I left the enum as a placeholder to allow for the possibility of bringing this into some version of 9.x.  Plus I had to leave around more than I wanted to keep the override regression test happy. If we bring this back to 9.x, we will need another PR to remove any last remnants for 10.x.  If we are not going to bring this back to 9.x, should update this PR to remove the enum and related test code.

This partially closes #7735


shinrich:
[approve ci clang-format]

randall (approved):

bryancall (approved):

zwoop:

PR #7752, opened 2021-04-27
AuTest extension: check for unrecognized configurations

bneradt:
This adds a trafficserver AuTest extension check to verify that the
tester did not accidentally add an unrecognized configuration option.


randall (dismissed):
Awesome!

randall (approved):

PR #7753, opened 2021-04-27
Make when_to_revalidate setting available on HTTPS

maskit:
A setting to specify when to revalidate cache (proxy.config.http.cache.when_to_revalidate) was only available for cleartext HTTP.

### DIFF
--- a/proxy/http/HttpTransact.cc
+++ b/proxy/http/HttpTransact.cc
@@ -7331,17 +7331,13 @@ HttpTransact::what_is_document_freshness(State *s, HTTPHdr *client_request, HTTP
 
   TxnDebug("http_match", "[what_is_document_freshness] fresh_limit:  %d  current_age: %" PRId64, fresh_limit, (int64_t)current_age);
 
-  /////////////////////////////////////////////////////////
-  // did the admin override the expiration calculations? //
-  // (used only for http).                               //
-  /////////////////////////////////////////////////////////
   ink_assert(client_request == &s->hdr_info.client_request);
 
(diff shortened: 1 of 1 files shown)


maskit:
The setting has been unavailable on HTTPS for long time and it may be just not maintained, but I'd say it's a bug and it should be fixed on 9.x as well. If it's too late for 9.1.0 then 9.1.1 is fine.

proxy/http/HttpTransact.cc
```diff
@@ -7331,17 +7331,13 @@ HttpTransact::what_is_document_freshness(State *s, HTTPHdr *client_request, HTTP
...
-  // did the admin override the expiration calculations? //
-  // (used only for http).                               //
-  /////////////////////////////////////////////////////////
   ink_assert(client_request == &s->hdr_info.client_request);
 
-  if (s->txn_conf->cache_when_to_revalidate == 0) {
-    ;
-    // Compute how fresh below
-  } else if (client_request->url_get()->scheme_get_wksidx() == URL_WKSIDX_HTTP) {
+  if (auto scheme = client_request->url_get()->scheme_get_wksidx(); scheme == URL_WKSIDX_HTTP || scheme == URL_WKSIDX_HTTPS) {
```
shinrich, line 7336: Is there much of a point to check the scheme here at all?

shinrich (approved):

bryancall (approved):

proxy/http/HttpTransact.cc
maskit: Maybe not, but the intention was to exclude WebSocket.

zwoop:

PR #7335, opened 2020-11-20
Unify all the connect timeouts into one

shinrich:
Eliminate proxy.config.http.parent_proxy.connect_attempts_timeout and proxy.config.http.post_connect_attempts_timeout and   instead use proxy.config.http.connect_attempts_timeout and associated overrides to control all connection to origin timeouts.

I sent a mail to the dev mailing list to start a discussion on this.  I forget about the proxy.config.http.post_connect_attempts_timeout option and we have left this at the default of 1800 seconds.  

Differentiating on the method may have been useful when the connect timeouts were really measuring TTFB.  But now that they are only applied to the time for the TCP or TLS handshake, it seems more confusing to separate them.

Similarly having a different timeout for the parent connect case seems like more confusion than it is worth.

Since I did not leave placeholders in the override enums, this PR is a breaking change. If it was brought into the 9.x releases after 9.0.0 is released, we would have to add some placeholder enums.

PR #7756, opened 2021-04-28
Remove references to removed options

randall:
These were originally removed in #7335


shinrich (approved):

bneradt (approved):

bneradt:
Marking this as Incompatible since the #7335 PR, upon which it depends, is Incompatible.

Issue #7638, opened 2021-03-28
9.0.x: Connection errors to healthy origin


mtorluemke:
From docs:

20210328.17h19m28s CONNECT:[0] could not connect [CONNECTION_ERROR] to 192.168.3.1 for 'http://192.168.3.1:2000/' connect_result=5 src_port=0
20210328.17h19m37s CONNECT:[0] could not connect [CONNECTION_ERROR] to 192.168.3.1 for 'http://192.168.3.1:2000' connect_result=5 src_port=0
20210328.17h19m39s CONNECT:[0] could not connect [CONNECTION_ERROR] to 192.168.3.1 for 'http://192.168.3.1:2000/' connect_result=5 src_port=0
20210328.17h20m28s CONNECT:[0] could not connect [CONNECTION_ERROR] to 192.168.3.1 for 'http://192.168.3.1:2000/' connect_result=5 src_port=0

But, connectivity is fine: 

root@origin1 # telnet 192.168.3.1 2000
Trying 192.168.3.1...
Connected to 192.168.3.1.
Escape character is '^]'.
^CConnection closed by foreign host.
root@origin1 #

Error is from:
https://github.com/apache/trafficserver/blob/660bd3b2e9a252a319f67ccd62bae91b176844e0/proxy/http/HttpTransact.cc#L3849 ...which hasn't changed in a few years.

So, one theory is that connectivity had an interruption, and did not recover when the connectivity is restored.

shinrich:
@djcarlin has been tracking down similar issues for us since before Christmas.  We have been running with the changes in PR #7580 to help track down the issue.

djcarlin:
Do you see log entries in error.log with "(setting last failure time)" ?  We were triggering "dead/down" origin behavior inappropriately sometimes, which would prevent subsequent connections from working for a time.

In our case, I noticed squid.log entries with "ERR_CONNECT_FAIL 502" when a POST was happening, and Content-Length sent by client request differed from %<cqql> log field - "Client request header and content length combined, in bytes."  In our case, Content-Length was tens of megabytes and cqql was only hundreds of bytes.  This was marking origin down and breaking other valid requests.

djcarlin:
If you think you're bumping against dead/down origin behavior, see https://github.com/apache/trafficserver/issues/7290 and https://github.com/apache/trafficserver/issues/7283 - I think the conditions on when an origin is marked dead/down need to be reviewed.

mtorluemke:
[...]

Issue #7290, opened 2020-10-22
Clarification and/or configurability of what it means for a server to be dead/down


shinrich:
While working with @djcarlin and @bryancall to understand an issue from deploying the dead server no retry feature added in PR #7142, we were surprised by when exactly servers are marked dead.

In the case where transactions are retryable, things pretty much worked as expected.  If the handshake failed or the request is sent but the origin fails to return data and is retryable, the address is tried the number of times specified in proxy.config.http.connect_attempts_rr_retries (at least once PR #7288 is applied) before marking the IP address as down and moving onto the next IP address.

However, if the transaction failed after sending the header and it is not retryable (e.g. a POST request), the ip address is marked down immediately (the retry count in proxy.config.http.connect_attempts_rr_retries is ignored).  If the origin only times out now and again due to larger requests, taking it down immediately seems bad particularly using the new feature that avoids the retries against the down server in the dead period.  However, if the server is consistently failing to respond to post requests, it should be marked down.

Probably this down decision criteria needs to be configurable.  Some origins need different criteria than others.  Some should only be marked down in the initial handshake fails.  Others should be marked down but only if no data was returned.  Or maybe you want to mark things down only for specific origin connection failures.

djcarlin:
I noticed squid.log entries with "ERR_CONNECT_FAIL 502" when a POST was happening, and Content-Length sent by client request differed greatly from %\<cqql> log field - "Client request header and content length combined, in bytes."

In our case, Content-Length was tens of megabytes and cqql was only hundreds of bytes. This was marking origin down and breaking other valid requests.

shinrich:
I ran across the proxy.config.http.down_server.abort_threshold setting.  I think the comparison is backwards from the description.  But if the client times out in more than this many seconds and the origin hasn't sent anything back yet, it will count towards marking that origin down.   The default value for this setting is 10..

The intent is to include "slow servers" in the set of dead servers to be avoided.  But if your client side inactivity timeout is less than your origin side inactivity, the bad client scenario that @djcarlin identifies will trigger this and increment the counts for the origin connection to be down.

It is easy enough to take proxy.config.http.down_server.abort_threshold out of the equation by setting it higher than any of your inactivity timeouts.  But I think we may want to reconsider this setting.  The intent of marking "slow" servers as down is admirable, but identifying the slowness as an issue on the origin side vs the client side is sketchy at best.

shinrich:
[...]

PR #7757, opened 2021-04-28
Add dead server policy control and metric. Improve messages.

shinrich:
This should help debug #7638.  Also addresses some issues in #7290

We are running with this changes in production.  I have exercised the new proxy.config.http.connect.dead.policy setting in production on a box, but @djcarlin needs to exercise it more widely.

This PR adds a setting proxy.config.http.connect.dead.policy to add some control over what kind of failures should count towards a dead server.  In our deployment, we want to only mark a server dead it if is failing the TCP level connection.  TLS failures like bad certificates may just be applicable to certain traffic to the server, so failures there should not mark the entire server dead.

The PR also adds a metric proxy.process.http.dead_server.no_requests to track how many client requests are returned immediately because the server is dead and retries is set to 0.

With this changes and dead.policy set to 1, we should be able to take advantage of PR #7142 and set the dead server retries to 0.


shinrich:
[ci approve clang-format]

shinrich:
[approve ci clang-format]

vmamidi (dismissed):

vmamidi (approved):

zwoop:
I would prefer this in 9.2., but willing to listen to arguments :-).

maskit:
I'm curious how `proxy.config.http.connect.dead.policy` would work for QUIC connections. The current description explicitly says "TCP handshake failures", but it's not limited to the "handshake", right? I assume it includes connection failures due to DNS issue and packet routing issue as well.

If so, we might want to generalize the description (or clarifying the border) before we make a release with this to avoid incompatible changes in the future. That would also help implementing this for QUIC connections. If I read current description literally, QUIC connection failures would never be counted.

zwoop:
@bneradt This has some merge conflicts, and as discussed, we'd like to change those Errors() for now for the 9.1.0 branch.

zwoop:
Backported in #8213

PR #7758, opened 2021-04-28
Traffic Dump: Add an HTTP/3 AuTest

bneradt:
Adding an HTTP/3 test for Traffic Dump. This test is currently skipped
because ATS does not have the TS_EVENT_HTTP_SSN_* hooks implemented for
QUIC connections yet. Although the test is skipped, I plan to commit
this change now because it adds the following value:

* With minor modifications the test should work when these hooks are
  implemented.
* The test can be used to help development when we decide to implement
  these hooks for QUIC.
* This test provides an example AuTest for completed HTTP/3 transactions
  which we currently do not have.
* This commit updates the Proxy Verifier extensions for HTTP/3 which will
  be helpful for future QUIC tests using Proxy Verifier.

### DIFF
--- a/tests/gold_tests/pluginTest/traffic_dump/replay/http3.yaml
+++ b/tests/gold_tests/pluginTest/traffic_dump/replay/http3.yaml
@@ -1,0 +1,64 @@
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


tests/gold_tests/pluginTest/traffic_dump/replay/http3.yaml
```diff
@@ -0,0 +1,64 @@
...
+        reason: OK
+        headers:
+          fields:
+          - [ Content-Length, 16 ]
+          - [ Connection, close ]
+
+sessions:
+- protocol: [ {name: http, version: 3},
+              {name: tls, sni: www.tls.com},
+              {name: tcp},
```
maskit, line 33: H3 over TCP?

Also, I'm not sure if it should have tls. I can see the stack has sni key, but strictly speaking, tls is not on QUIC.

bneradt: Thank you for the observations,

Good point: in this case, the tcp is not used by Proxy Verifier. It is over UDP. That is: `{name: http, version 3}` dictates the whole stack which is over QUIC over UDP. I'll update that protocol node.

The `{name: tls}` node indicates the cryptographic characteristics of QUIC (SNI in this case). Does QUIC support unencrypted traffic? I think it has to be encrypted, yes?

maskit: Right, it's always encrypted at least on version 1.

traeak (approved):
These changes look good and they don't seem to interfere with the normal autests.

zwoop:
Cherry-picked to 9.1.x, to resolve some merge conflicts for other PRs.

PR #7760, opened 2021-04-28
doc: header_rewrite random function not inclusive

mlibbey:
In testing %{RANDOM:3}, ATS only prints 0,1, and 2. The code
`rand_r(&_seed) % _max)`
implies that the n value would not be chosen.


shukitchan (approved):
nice catch

zwoop:

PR #7762, opened 2021-04-29
Add traffic_server command line option for debugging in Au test.

ywkaras:
See added code comments for details.

### DIFF
--- a/src/traffic_server/traffic_server.cc
+++ b/src/traffic_server/traffic_server.cc
@@ -174,6 +174,16 @@ static int poll_timeout         = -1; // No value set.
 static int cmd_disable_freelist = 0;
 static bool signal_received[NSIG];
 
+/*
+To be able to attach with a debugger to traffic_server running in an Au test case, temporarily add the
+parameter block_for_debug=True to the call to Test.MakeATSProcess().  This means Au test will wait
+effectively indefinitely (10 hours) for traffic_server to initialize itself.  Run the modified Au test,
+attach the debugger to the traffic_server process, set one or more breakpoints, set the variable
+cmd_block to 0, then continue.  On linux, the command 'ps -ef | fgrep -e --block' will help identify the
+PID of the traffic_server process (second column of output).
+*/
+static int cmd_block = 0;
+
 // 1: the main thread delayed accepting, start accepting.
 // 0: delay accept, wait for cache initialization.
 // -1: cache is already initialized, don't delay.
@@ -217,6 +227,7 @@ static ArgumentDescription argument_descriptions[] = {
   {"bind_stderr", '-', "Regular file to bind stderr to", "S512", &bind_stderr, "PROXY_BIND_STDERR", nullptr},
   {"accept_mss", '-', "MSS for client connections", "I", &accept_mss, nullptr, nullptr},
   {"poll_timeout", 't', "poll timeout in milliseconds", "I", &poll_timeout, nullptr, nullptr},
+  {"block", '-', "block for debug attach", "T", &cmd_block, nullptr, nullptr},
   HELP_ARGUMENT_DESCRIPTION(),
   VERSION_ARGUMENT_DESCRIPTION(),
   RUNROOT_ARGUMENT_DESCRIPTION(),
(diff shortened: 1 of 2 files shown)


src/traffic_server/traffic_server.cc
```diff
@@ -1753,6 +1767,12 @@ main(int /* argc ATS_UNUSED */, const char **argv)
   command_index = find_cmd_index(command_string);
   command_valid = command_flag && command_index >= 0;
 
+  // Attach point when TS is blocked for debugging is in this loop.
```
brbzull0, line 1770: Would it make sense doing this only if `DEBUG`?  I mean just to make it extremely obvious.

bneradt:
Thanks for working on making debugging ATS easier.

I could use some help understanding the use case for this. When I want to run traffic server in a debugger in AuTest, I generally run with `--clean=none` which leaves behind the sandbox.  I then prepend `gdb --args` to the final command line in replay.sh from the sandbox. Then I replay it, add breakpoints, `run` and so forth.

When would this block command be used rather than that method? I have had issues when the AuTest runs via traffic_manager. Is this for those situations?

Don't interpret this question as critical of this patch. I'm just trying to understand its use case further.

src/traffic_server/traffic_server.cc
ywkaras: In some cases the test case failure will only happen in optimized code, so I'd prefer to have this available without --enable-debug .

ywkaras:
Yes, you can generally use replay instead.  But I typically find it's more convenient to be able to just rerun the test and attach to the traffic_server process.

ywkaras:
I don't think you can use this with traffic_manager.  Because you have to use command='traffic_manager' as the command parameter to MakeATSProcess().

ywkaras:
This capture illustrates how you would use this, after starting the modified Au test from another terminal:
```
$ pwd
/Users/wkaras/REPOS/TS2/tests/_sandbox/strategies_ch/ts/bin
$ ps -ef | fgrep -e --block
root      5201  4915  0 18:38 ?        00:00:00 traffic_server --block
root      5205  5010  0 18:41 pts/1    00:00:00 grep -F --color=auto -e --block
$ gdb traffic_server 5201
GNU gdb (GDB) Red Hat Enterprise Linux 7.6.1-110.el7
...
Missing separate debuginfos, use: debuginfo-install glibc-2.17-222.el7.x86_64 hwloc-libs-1.11.8-4.el7.x86_64 libattr-2.4.46-13.el7.x86_64 libcap-2.22-9.el7.x86_64 libgcc-4.8.5-28.el7_5.1.x86_64 libstdc++-4.8.5-28.el7_5.1.x86_64 libtool-ltdl-2.4.2-22.el7_3.x86_64 numactl-libs-2.0.9-7.el7.x86_64 pcre-8.32-17.el7.x86_64 xz-libs-5.2.2-1.el7.x86_64 zlib-1.2.7-17.el7.x86_64
(gdb) b HttpTransact.cc:1595
Breakpoint 1 at 0x5d9496: file HttpTransact.cc, line 1595.
(gdb) set cmd_block=0
(gdb) c
Continuing.
[New Thread 0x7f8fd4870700 (LWP 5211)]
...
[New Thread 0x7f8f8bbe1700 (LWP 5233)]
[Switching to Thread 0x7f8fd2eab700 (LWP 5215)]

Breakpoint 1, HttpTransact::HandleRequest(HttpTransact::State*) () at HttpTransact.cc:1595
1595	  if (s->http_config_param->no_dns_forward_to_parent && s->scheme != URL_WKSIDX_HTTPS &&
(gdb) 
```

src/traffic_server/traffic_server.cc
```diff
@@ -217,6 +235,7 @@ static ArgumentDescription argument_descriptions[] = {
   {"bind_stderr", '-', "Regular file to bind stderr to", "S512", &bind_stderr, "PROXY_BIND_STDERR", nullptr},
   {"accept_mss", '-', "MSS for client connections", "I", &accept_mss, nullptr, nullptr},
   {"poll_timeout", 't', "poll timeout in milliseconds", "I", &poll_timeout, nullptr, nullptr},
+  {"block", '-', "block for debug attach", "T", &cmd_block, nullptr, nullptr},
```
bneradt, line 238: This looks promising.

I suggest we update the MakeATSProcess trafficserver extension in autest-site to take a flag that configures this for the user. Maybe `block_for_detach=False` and someone wanting to use this can pass `block_for_detatch=True` and this will set things up for that, doing the necessary to make things work for cmd being traffic_server or traffic_manager.

bneradt (changes requested):

ezelkow1 (approved):
Tested here using autest and blocking, attach with gdb and continued

bneradt (approved):
Thanks Walt. Looks good.

@willwendorfvm : this might be helpful for debugging the runroot issues you're working on.

ywkaras:
Marking this for 9.2 backport, but if it would be useful in 9.1.x or 9.0.x, very low risk of introducing bugs I think.

Issue #7769, opened 2021-04-30
Need priv elevation for traffic_manager for SSL cert reload using traffic_ctl


sudheerv:
See bunch of these messages from traffic_manager when traffic_ctl config reload is issued:

{code}
[Apr 28 01:16:50.968] traffic_manager NOTE: [ConfigManager::ConfigManager] xxxxxxxxxxxxxxxxxxxxxxxxxxx Unable to load: Permission denied
{code}

PR #7770, opened 2021-04-30
Elevate privileges for traffic_manager during SSL cert reload

sudheerv:
This fixes issue #7769 

traffic_manager now watches the raw SSL certs as well and needs the privs
to stat those files.

### DIFF
--- a/src/traffic_manager/AddConfigFilesHere.cc
+++ b/src/traffic_manager/AddConfigFilesHere.cc
@@ -45,14 +45,14 @@ testcall(char *foo, char * /*configName */)
 }
 
 void
-registerFile(const char *configName, const char *defaultName, bool isRequired)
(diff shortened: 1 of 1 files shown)


src/traffic_manager/AddConfigFilesHere.cc
```diff
@@ -45,7 +45,7 @@ testcall(char *foo, char * /*configName */)
 }
 
 void
-registerFile(const char *configName, const char *defaultName, bool isRequired)
+registerFile(const char *configName, const char *defaultName, bool isRequired, bool isElevateNeeded = false)
 {
```
shinrich, line 49: Where is the new isElevateNeeded argument used?

shinrich (approved):
Looks good.

PR #7279, opened 2020-10-16
Add throttled log functions

bneradt:
Some messages get excessively noisy under high traffic conditions if
something about their mechanism goes wrong. The pipe logging feature,
for instance, will emit warning and error messages on every single log
event if the reader goes down or the pipe buffer fills up. This can
result in thousands of log messages being emitted per second, which
makes reading the logs difficult and causes disk space issues.

This commit addresses this issue by adding throttled versions of the
common logging messages so they only emit a message on some set
interval (60 seconds as a default). The following functions are added:

SiteThrottledStatus
SiteThrottledNote
SiteThrottledWarning
SiteThrottledError
SiteThrottledFatal
SiteThrottledAlert
SiteThrottledEmergency

As a bonus, these are implemented using a generic Throttler class which may also be
useful in other applications where throttling is desired.

PR #7772, opened 2021-04-30
Apply log throttling to HTTP/2 session error rate messages

bneradt:
These messages may become voluminous under certain circumstances.
Applying log throttling to them so they don't become too noisy.


bryancall:
@masaori335 Can you please take a look at this?

masaori335 (approved):
Looks reasonable. I agree with this warning message could be noisy in some situations.

zwoop:

bryancall:
This PR depends on #7279

PR #7787, opened 2021-05-05
Fixes an issue where next hop unit tests crash when run on macOS.

jrushford:
Fixes an issue when running the Next Hop unit tests on macOS.  macOS crashes due to a SIGABRT when a read/write lock cannot be obtained to read a config variable in the NextHopSelectStrategy constructor.  This PR basically creates a mock around REC_ReadConfigInteger().

### DIFF
--- a/proxy/http/remap/unit-tests/nexthop_test_stubs.cc
+++ b/proxy/http/remap/unit-tests/nexthop_test_stubs.cc
@@ -126,6 +126,18 @@ build_request(int64_t sm_id, HttpSM *sm, sockaddr_in *ip, const char *os_hostnam
 }
 
 void
+GetConfigInteger(int *v, const char *n)
+{
+  if (v != nullptr && n != nullptr) {
+    if (strcmp(n, "proxy.config.http.parent_proxy.max_trans_retries") == 0) {
+      *v = 1;
+    } else {
+      *v = 0;
+    }
+  }
+}
+
+void
 PrintToStdErr(const char *fmt, ...)
 {
   va_list args;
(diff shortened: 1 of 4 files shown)


proxy/http/remap/unit-tests/nexthop_test_stubs.cc
```diff
@@ -125,6 +125,14 @@ build_request(int64_t sm_id, HttpSM *sm, sockaddr_in *ip, const char *os_hostnam
   sm->t_state.txn_conf               = reinterpret_cast<OverridableHttpConfigParams *>(_my_txn_conf);
 }
 
+void
+GetConfigInteger(int *v, const char *n)
+{
+  if (strncmp(n, "proxy.config.http.parent_proxy.max_trans_retries", strlen(n)) == 0) {
```
ywkaras, line 131: Why not just use strcmp()?

jrushford: @ywkaras I think strncmp() is safer.  Why do you prefer strcmp()?

ywkaras: I think the safety issue is that a C string might have a missing null terminator.  But strncmp() only helps with  that if it's got a constant as the length.  strcmp() is a safe as strncmp() if one of the strings is a string literal.

jrushford: @ywkaras I made the change to strcmp() and added some nullptr checks.

ywkaras (approved):

PR #7789, opened 2021-05-05
tests: Fixes spelling

randall:


shinrich (dismissed):

bneradt (changes requested):
Looks good. Thanks for fixing the spelling.

tests/gold_tests/h2/httpbin.test.py
```diff
@@ -83,7 +83,7 @@
 # Test Cases
 # ----
 
-# Test Case 0: Basic request and resposne
+# Test Case 0: Basic request and Response
```
bneradt, line 86: Looks like this is still "Response" with a capital R. Was that supposed to be lower-cased?

Thanks,
Brian

bneradt (changes requested):

tests/gold_tests/h2/httpbin.test.py
randall: You're absolutely right. Should be fixed in the latest version

bneradt (approved):
Looks good. Thanks Randall!

PR #7790, opened 2021-05-05
Fix MLoc assert caused by s3auth

### LINKED ISSUE #7780: 9.1 assertion in TSHandleMLocRelease
After running solidly all weekend, I had a bunch of repeated asserts with the following stack from our production box running 9.1.x.  The stack looks weird because the invoking plugin is written in Lua. (as a side note if anyone know how to get symbols from lua plugins I would appreciate any hints).

```
#0  0x00002b8d28cbe3d7 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:55
#1  0x00002b8d28cbfac8 in __GI_abort () at abort.c:90
#2  0x00002b8d2621861b in ink_abort (message_format=message_format@entry=0x2b8d2623ac27 "%s:%d: failed assertion `%s`")
    at ../../../../../../_vcs/trafficserver9.1/src/tscore/ink_error.cc:99
#3  0x00002b8d26214b95 in _ink_assert (expression=expression@entry=0x7efd82 "!\"invalid mloc\"", 
    file=file@entry=0x7edf30 "../../../../../_vcs/trafficserver9.1/src/traffic_server/InkAPI.cc", line=line@entry=2117)
    at ../../../../../../_vcs/trafficserver9.1/src/tscore/ink_assert.cc:37
#4  0x0000000000505d90 in TSHandleMLocRelease (bufp=0x2b8f904f3828, parent=<optimized out>, mloc=0x2b8f59d54308) at ../../../../../_vcs/trafficserver9.1/src/traffic_server/InkAPI.cc:2117
#5  0x00002b8f3a719748 in ?? ()
#6  0x00002b8f904f3000 in ?? ()
#7  0x00002b8f904f3828 in ?? ()
#8  0x00002b8f59d54088 in ?? ()
#9  0x00002b8f59d54308 in ?? ()
#10 0x00002b8d31bfeedc in ?? ()
#11 0x00002b8f4ddc8120 in ?? ()
#12 0x00002b8f376da840 in ?? ()
#13 0x00002b8f904f3000 in ?? ()
#14 0x000000000000ea64 in ?? ()
#15 0x0000000000502f00 in INKContInternal::handle_event(int, void*) () at ../../../../../_vcs/trafficserver9.1/src/traffic_server/InkAPI.cc:1137
#16 0x0000000000517037 in handleEvent (data=0x2b8f904f3000, event=60004, this=0x2b8f4ddc8120)
    at /sd/workspace/src/git.vzbuilders.com/Edge/build/_build/build_release_posix-x86_64_gcc_8/trafficserver9.1/build/../../../../_vcs/trafficserver9.1/iocore/eventsystem/I_Continuation.h:219
#17 handleEvent (data=0x2b8f904f3000, event=60004, this=0x2b8f4ddc8120)
    at /sd/workspace/src/git.vzbuilders.com/Edge/build/_build/build_release_posix-x86_64_gcc_8/trafficserver9.1/build/../../../../_vcs/trafficserver9.1/iocore/eventsystem/I_Continuation.h:215
#18 APIHook::invoke(int, void*) const () at ../../../../../_vcs/trafficserver9.1/src/traffic_server/InkAPI.cc:1374
```

Specifically it assert is in the default case and the obj->m_type == 0, which implies that mloc has already been freed or had never been allocated.
```
TSReturnCode
TSHandleMLocRelease(TSMBuffer bufp, TSMLoc parent, TSMLoc mloc)
{
  MIMEFieldSDKHandle *field_handle;
  HdrHeapObjImpl *obj = (HdrHeapObjImpl *)mloc;

  if (mloc == TS_NULL_MLOC) {
    return TS_SUCCESS;
  }

  sdk_assert(sdk_sanity_check_mbuffer(bufp) == TS_SUCCESS);

  switch (obj->m_type) {
  case HDR_HEAP_OBJ_URL:
  case HDR_HEAP_OBJ_HTTP_HEADER:
  case HDR_HEAP_OBJ_MIME_HEADER:
    return TS_SUCCESS;

  case HDR_HEAP_OBJ_FIELD_SDK_HANDLE:
    field_handle = (MIMEFieldSDKHandle *)obj;
    if (sdk_sanity_check_field_handle(mloc, parent) != TS_SUCCESS) {
      return TS_ERROR;
    }

    sdk_free_field_handle(bufp, field_handle);
    return TS_SUCCESS;

  default:
    ink_release_assert(!"invalid mloc");
    return TS_ERROR;
  }
}
```

We did not see this assert in our 9.0.x branch.  Adding HDR_HEAP_OBJ_EMPTY to the success cases, the assert goes away and things seem to run well, but would be good to understand the change in behavior.

shinrich:
The problem is that the S3auth object contains a URL MLoc reference in _url_mloc.  In our configuration, one of the plugins called after s3auth changes the server_request URL in such as way that the _url_mloc reference is old and destroyed, see http_hdr_url_set in proxy/hdrs/HTTP.cc.  deallocate_obj clears the m_type which causes the later assert in MLocRelease.

By causing the S3Request object to clean up before calling TSHttpTxnReenable, the _url_mloc is not stale when MLocRelease is called.

We only see this problem in 9.0 not in 9.1, and I don't know why.  All of the relevant pieces of code are the same, and we are running with the same configs.  It is possible there are other changes that cause ATS to take the other path in http_hdr_url_set which updates the URLImpl object in place rather than just updating a pointer.  

This fixes #7780

### DIFF
--- a/plugins/s3_auth/s3_auth.cc
+++ b/plugins/s3_auth/s3_auth.cc
@@ -991,36 +991,40 @@ S3Request::authorizeV2(S3Config *s3)
 int
 event_handler(TSCont cont, TSEvent event, void *edata)
 {
-  TSHttpTxn txnp = static_cast<TSHttpTxn>(edata);
-  S3Config *s3   = static_cast<S3Config *>(TSContDataGet(cont));
-
-  S3Request request(txnp);
-  TSHttpStatus status  = TS_HTTP_STATUS_INTERNAL_SERVER_ERROR;
+  TSHttpTxn txnp       = static_cast<TSHttpTxn>(edata);
+  S3Config *s3         = static_cast<S3Config *>(TSContDataGet(cont));
   TSEvent enable_event = TS_EVENT_HTTP_CONTINUE;
 
-  switch (event) {
-  case TS_EVENT_HTTP_SEND_REQUEST_HDR:
-    if (request.initialize()) {
-      while (s3->reload_waiting) {
-        std::this_thread::yield();
+  {
+    S3Request request(txnp);
+    TSHttpStatus status = TS_HTTP_STATUS_INTERNAL_SERVER_ERROR;
+
+    switch (event) {
+    case TS_EVENT_HTTP_SEND_REQUEST_HDR:
+      if (request.initialize()) {
+        while (s3->reload_waiting) {
+          std::this_thread::yield();
+        }
+
+        std::shared_lock lock(s3->reload_mutex);
+        status = request.authorize(s3);
       }
 
-      std::shared_lock lock(s3->reload_mutex);
-      status = request.authorize(s3);
+      if (TS_HTTP_STATUS_OK == status) {
+        TSDebug(PLUGIN_NAME, "Successfully signed the AWS S3 URL");
+      } else {
+        TSDebug(PLUGIN_NAME, "Failed to sign the AWS S3 URL, status = %d", status);
+        TSHttpTxnStatusSet(txnp, status);
+        enable_event = TS_EVENT_HTTP_ERROR;
+      }
+      break;
+    default:
+      TSError("[%s] Unknown event for this plugin", PLUGIN_NAME);
+      TSDebug(PLUGIN_NAME, "unknown event for this plugin");
+      break;
     }
-
-    if (TS_HTTP_STATUS_OK == status) {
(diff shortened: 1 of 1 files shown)


ywkaras:
Looks good to me, request formal approval from non-Yahoo.

zwoop:

PR #7792, opened 2021-05-06
Fix to cookie_remap.cc to avoid Intel compiler warning.

ywkaras:
The Intel compiler warns if an anonymous class has a
non-trival constructor.  Presumably because that implies
an unnamed function.


ywkaras:
Yahoo would like TS to build with the Intell compiler (icc).

ywkaras:
@dragon512 do we no longer need this?

vmamidi (approved):

randall (approved):

PR #7793, opened 2021-05-06
use sendmsg and recvmsg

vmamidi:
replace read and write with sendmsg and recvmsg. This would help in implementing time stamping (https://www.kernel.org/doc/html/latest/networking/timestamping.html),  zero copy ( https://www.kernel.org/doc/html/v4.15/networking/msg_zerocopy.html )


bryancall (approved):
👍🏻 Looks good

zwoop:
@bryancall  On your spiffy setup, can you see if there's a noticeable CPU cost in setting up all these msg structs on every write now? Hopefully there is not.

zwoop:

PR #7795, opened 2021-05-07
Make HTTP/2 configs reloadable

masaori335:
~~The doc says HTTP/2 configs are reloadable, but all HTTP/2 configs require restarting ATS to apply.~~

This introduces `Http2Config` and `Http2ConfigParams` like other subsystems have.

### DIFF
--- a/proxy/http2/Http2Frame.cc
+++ b/proxy/http2/Http2Frame.cc
@@ -76,11 +76,6 @@ Http2DataFrame::write_to(MIOBuffer *iobuffer) const
 int64_t
 Http2HeadersFrame::write_to(MIOBuffer *iobuffer) const
 {
-  // Validation
-  if (this->_hdr_block_len > Http2::max_frame_size) {
-    return -1;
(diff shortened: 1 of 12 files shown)


proxy/http2/Http2Frame.cc
```diff
@@ -76,11 +76,6 @@ Http2DataFrame::write_to(MIOBuffer *iobuffer) const
 int64_t
 Http2HeadersFrame::write_to(MIOBuffer *iobuffer) const
 {
-  // Validation
```
masaori335, line 79: I just got rid of this validation, because the `_hdr_block_len` is already checked before making `Http2HeadersFrame` and too noisy to read or pass the config from `ua_session` here.

masaori335:
[approve ci autest]

masaori335:
Close this. All HTTP/2 configs are reloadable. I misunderstand what `REC_EstablishStaticConfigInt32U()` does.

Issue #7705, opened 2021-04-13
Assertion in HttpCacheSM captive_action.cancelled == 0


maskit:
I realized that an assertion below fails on my laptop while I was running h2spec. Reverting #7667 solves the issue, but I'm not sure how it is related.

```
Fatal: HttpCacheSM.cc:161: failed assertion `captive_action.cancelled == 0`
2021-04-13 11:33:40.743573+0900 traffic_server[76017:10095094] Fatal: HttpCacheSM.cc:161: failed assertion `captive_action.cancelled == 0`
Process 76017 stopped
* thread #3, name = '[ET_NET 1]', stop reason = signal SIGABRT
    frame #0: 0x00007fff2037a462 libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
->  0x7fff2037a462 <+10>: jae    0x7fff2037a46c            ; <+20>
    0x7fff2037a464 <+12>: movq   %rax, %rdi
    0x7fff2037a467 <+15>: jmp    0x7fff203746a1            ; cerror_nocancel
    0x7fff2037a46c <+20>: retq   
Target 0: (traffic_server) stopped.
(lldb) bt
* thread #3, name = '[ET_NET 1]', stop reason = signal SIGABRT
  * frame #0: 0x00007fff2037a462 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff203a8610 libsystem_pthread.dylib`pthread_kill + 263
    frame #2: 0x00007fff202fb720 libsystem_c.dylib`abort + 120
    frame #3: 0x0000000103a2ac8a libtscore.10.dylib`ink_abort(message_format="%s:%d: failed assertion `%s`") at ink_error.cc:99:3
    frame #4: 0x0000000103a20727 libtscore.10.dylib`::_ink_assert(expression="captive_action.cancelled == 0", file="HttpCacheSM.cc", line=161) at ink_assert.cc:37:3
    frame #5: 0x00000001001f58e9 traffic_server`HttpCacheSM::state_cache_open_write(this=0x0000000110ea2c30, event=1109, data=0xffffffffffffb04f) at HttpCacheSM.cc:161:3
    frame #6: 0x000000010000aab5 traffic_server`Continuation::handleEvent(this=0x0000000110ea2c30, event=1109, data=0xffffffffffffb04f) at I_Continuation.h:219:12
[...]

PR #7798, opened 2021-05-10
Revert "Add class to normalize handling of pending action (#7667)"

maskit:
This reverts commit 739994f21f9105d581212dec2b943ee8f885371b (#7667).

At least 3 people faced the assertion failure I reported as #7705. If the issue is not going to be resolved soon, I'd like to take back my approval for the change and revert the commit causing the failure because the change was supposed to be just a cleanup.

The commit is already backported to 9.1.x branch and I confirmed the branch has the same issue.


maskit:
#7807 will fix the issue.

PR #7814, opened 2021-05-12
Cleanup: Remove a meaningless Http2Stream::do_io_close() call

masaori335:
At the beginning of `Http2Stream::initiating_close()`, the `closed` flag is set `true`. Calling `Http2Stream::do_io_close()` after setting the flag does nothing.

`Http2Stream::initiating_close()`
https://github.com/apache/trafficserver/blob/31a580d0d89b2c141655ea167e4ad1b6b4e4973c/proxy/http2/Http2Stream.cc#L500

`Http2Stream::do_io_close(int /* flags */)`
https://github.com/apache/trafficserver/blob/31a580d0d89b2c141655ea167e4ad1b6b4e4973c/proxy/http2/Http2Stream.cc#L425

### DIFF
--- a/proxy/http2/Http2Stream.cc
+++ b/proxy/http2/Http2Stream.cc
@@ -539,7 +539,6 @@ Http2Stream::initiating_close()
       }
     } else if (!sent_write_complete) {
       // Transaction is already gone or not started. Kill yourself
-      do_io_close();
       terminate_stream = true;
       terminate_if_possible();
     }


maskit:
Is `do_io_close` called at somewhere else with `closed` flag off? A right way might be setting `closed` flag `true` after calling `do_io_close`.

Having this pattern below at two places seems very odd -- it implies calling either of those is enough but I think `do_io_close` should be always called if `do_io_read/write()` was called, like `open()` and `close()` system calls.
```
if (closed) {
  closed = true;
  ...
}
```

masaori335:
Basically, HttpSM/HttpTunnel should call ProxyTransaction's do_io_close() because it called do_io_read()/write().

For the closed flag, the current code is pretty wired. I'm working on getting rid of these flags and fixing the closing sequence of Http2Stream.

masaori335:
[approve ci autest]

maskit (commented):
> Basically, HttpSM/HttpTunnel should call ProxyTransaction's do_io_close() because it called do_io_read()/write().

I agree, that is ideal. However, this comment implies that there was an exception and we had to call `do_io_close` here at some point.

```
// Initiated from the Http2 side
```

I'm ok to remove the call if that helps cleanup because it does nothing, but I'd leave it as a hint until I make the fix for clean closing sequence.

+0

shinrich:
Good point about the closed flag and do_io_close.

shinrich (approved):

PR #7418, opened 2021-01-12
PoolableSession

a-canary:
Http1ServerSessions now inherit from an abstract PoolableSession which adds the following for outbound pooling:
+ linkage for intrusive dlist
+ KA status 
+ OutboundConnTrack::Group 

Its is expected that H2 and H3 session will also need to inherit from this, or something similar.
We will continue to refactor the HttpSM and ProxyTransaction to use PoolableSession* for all outbound transations.

PR #7828, opened 2021-05-12
Fix so EOS are delivered to sessions in the pool

### LINKED ISSUE #7827: 9.1.x runs with slightly more ERR_CONNECT_FAIL's than 9.0.x
We have been testing 9.1.x on a box in production for a couple weeks.  All is looking good except that the 9.1.x box has slightly more ERR_CONNECT_FAILS that its 9.0.x peer and slightly more 50x responses.  The difference is slight but consistent.  We need to figure out the cause before deploying 9.1.x more widely.

shinrich:
This fixes an issue introduced in PR #7418.  I just filed issue #7827 about it.  The 9.1.x servers would consistently have slightly more ERR_CONNECT_FAILS than its 9.0.x peer.  I finally got one of the transactions in a packet trace and it looks like the origin had closed the connection.  Then a short while later ATS would start sending traffic on the connection.  Then the resets would fly and the transaction would fail due to an EOS in state_read_response_header.

I finally tracked it down to a change in how the session is released to the pool.  In 9.0.x, the release involves doing do_io_read to the pool with max int bytes against the server session's buffer.  In 9.1.x the buffer is null, so the vc is not enabled.  So the EOS is not delivered while the session is in the pool. I confirmed this with additional warning messages in a test build.

With this change we add back the buffer into the do_io_read.  The EOS's are being delivered while the session is in the pool.  With this change the ERR_CONNECT_FAILS are consistent between the 9.1.x box and its 9.0 peer in production traffic.

This closes #7827.

### DIFF
--- a/proxy/PoolableSession.h
+++ b/proxy/PoolableSession.h
@@ -83,6 +83,8 @@ public:
 
   void set_netvc(NetVConnection *newvc);
 
+  virtual IOBufferReader *get_reader() = 0;
+
 private:
   // Sessions become if authentication headers
   //  are sent over them
--- a/proxy/http/Http1ClientSession.cc
+++ b/proxy/http/Http1ClientSession.cc
@@ -468,7 +468,7 @@ Http1ClientSession::attach_server_session(PoolableSession *ssession, bool transa
     //  have it call the client session back.  This IO also prevent
     //  the server net conneciton from calling back a dead sm
     SET_HANDLER(&Http1ClientSession::state_keep_alive);
-    slave_ka_vio = ssession->do_io_read(this, 0, nullptr);
+    slave_ka_vio = ssession->do_io_read(this, INT64_MAX, ssession->get_reader()->mbuf);
     ink_assert(slave_ka_vio != ka_vio);
 
     // Transfer control of the write side as well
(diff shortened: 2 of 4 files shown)


masaori335 (dismissed):
Looks reasonable. To get EOS, we need to enable the read of the vc.

masaori335:
Might be off topic, but the commit disabled the read on `Http1ClientSession::attach_server_session()` too. Is this safe?

https://github.com/apache/trafficserver/commit/d4f202c808dd1b76255ce8991960ff83c07a9c46?branch=d4f202c808dd1b76255ce8991960ff83c07a9c46#diff-87039ca39b184d8c124c6556c3ce15202a5c643747d0eb107bff28a799e140c1L481-R475

https://github.com/apache/trafficserver/blob/fa6aa624a80184b16b105c860d5565f1bb872636/proxy/http/Http1ClientSession.cc#L471

shinrich:
The change to the slave_ka_vio is probably harmful in the same way.  We are not currently using that feature, so I did not run into it.  Will update this PR to address that as well.

masaori335 (approved):

zwoop:

PR #7831, opened 2021-05-13
Fix stall on sending response for request with trailer header

masaori335:
Prior to the change, `Http2Stream::change_state()` assumes the END_STREAM flag comes from the DATA frame on the OPEN state. But HEADERS frame can send END_STREAM flag too[*1]. The trailer header case is exactly the case. ( An example is on RFC 7540 Section 8.1.3.[*2] ) When Http2Stream falls in this bug, it stalls because the stream is not restarted on receiving WINDOW_UPDATE frames.

The prior AuTest of the trailer header didn't detect this bug, because the response was smaller than the initial window.

[*1] 5.1. Stream States
https://datatracker.ietf.org/doc/html/rfc7540#section-5.1

[*2] 8.1.3.  Examples (trailing header fields)
https://datatracker.ietf.org/doc/html/rfc7540#page-59


shinrich (approved):

bryancall (approved):

zwoop:

PR #7834, opened 2021-05-13
AuTest: use exteneded help output to determine curl feature support

bneradt:
I ran the AuTest suite with a recent development version of curl and
noticed that the proxy_protocol.test.py test was skipped because it
thought that the built curl version didn't have haproxy-protocol
support. It did have this support, but that version of curl didn't show
that option in the brief `--help` output but it was shown if `--help
all` was used. This change makes checking curl feature support use
`--help all` for all feature requests.

### DIFF
--- a/tests/gold_tests/autest-site/conditions.test.ext
+++ b/tests/gold_tests/autest-site/conditions.test.ext
@@ -79,7 +79,7 @@ def HasCurlOption(self, option):
         return False
 
     return self.CheckOutput(
-        ['curl', '--help'],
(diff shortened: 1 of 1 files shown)


randall (approved):
Seems good. Old versions of curl that don't have `--help all` support don't fail. 👍

bneradt:
Removing this from the 8.1.x branch since the patch isn't applicable there: the conditions.test.ext doesn't have the function that tests curl's feature support.

zwoop:

PR #7837, opened 2021-05-13
Update NextHop strategies so that unavailable server retry codes are configurable.

jrushford:
Update NextHop strategies so that unavailable server retry codes are configurable in the YAML instead of using a hard coded range of codes.  This conforms to the same configuration capability in parent.config.


ywkaras (approved):

zwoop:
Moving to 9.2.x for now, unless you have a strong argument that this needs to be in 9.1.x.

PR #7849, opened 2021-05-17
Make HttpSM server reference a Transaction instead of a Session

shinrich:
This is derived as a subset of the H2 to origin PR #7622.  This PR adds no support for H2 to origin, but it performs the restructuring so the HttpSM refers to the server connection via a ProxyTransaction instead of a PoolableSession.  

This was requested by @maskit in hopes this work would assist them in discovering the cause of the server session leak described in issue #7471.  This restructuring should simplify server connection leaks because the server_txn object stays around until HttpSM::kill_this() when server_txn->transaction_done() is called.  If the server session should be returned to the origin pool, it will have previously been marked so.  Otherwise the server session will be closed.  Similarly earlier calls to do_io_close on the server transaction will be held until transaction_done() is called.  

In the case of an origin connection failure and subsequent retry, server_txn->transaction_done() is called in HttpSM::do_http_server_open() before another connection attempt is made.

The PR also includes some stats I added to help debug a pool buildup issue I ran into debugging this PR against our 9.1.x build in master.

I also pulled into the following commits to make the move to our 9.1.x branch from master easier.

PR #6241 - but I see that just got picked into 9.1.x already
PR #7571 - I think that one is there now too.
PR #7584 - which also just got pulled into 9.1.x
PR #7600

### DIFF
--- a/iocore/net/libinknet_stub.cc
+++ b/iocore/net/libinknet_stub.cc
@@ -21,11 +21,11 @@
   limitations under the License.
  */
 
-#include "HttpSessionManager.h"
+class EThread;
+class Continuation;
 void
 initialize_thread_for_http_sessions(EThread *, int)
 {
-  ink_assert(false);
 }
 
 #include "P_UnixNet.h"
--- a/proxy/PoolableSession.h
+++ b/proxy/PoolableSession.h
@@ -97,7 +97,7 @@ public:
   // singleton that keeps track of the connection counts.
   OutboundConnTrack::Group *conn_track_group = nullptr;
 
-  virtual IOBufferReader *get_reader() = 0;
+  virtual IOBufferReader *get_remote_reader() = 0;
 
 private:
   // Sessions become if authentication headers
--- a/proxy/PoolableSession.h
+++ b/proxy/PoolableSession.h
@@ -33,9 +33,11 @@ class PoolableSession : public ProxySession
 public:
   enum PooledState {
     INIT,
-    SSN_IN_USE,  // actively in use
-    KA_RESERVED, // stuck to client
-    KA_POOLED,   // free for reuse
+    SSN_IN_USE,    // actively in use
+    KA_RESERVED,   // stuck to client
+    KA_POOLED,     // free for reuse
+    SSN_CLOSED,    // Session ready to be freed
+    SSN_TO_RELEASE // Session reaady to be released
   };
 
   /// Hash map descriptor class for IP map.
@@ -72,16 +74,28 @@ public:
   TSServerSessionSharingMatchMask sharing_match = TS_SERVER_SESSION_SHARING_MATCH_MASK_NONE;
   TSServerSessionSharingPoolType sharing_pool   = TS_SERVER_SESSION_SHARING_POOL_GLOBAL;
 
-  // Keep track of connection limiting and a pointer to the
-  // singleton that keeps track of the connection counts.
-  OutboundConnTrack::Group *conn_track_group = nullptr;
+  void enable_outbound_connection_tracking(OutboundConnTrack::Group *group);
+  void release_outbound_connection_tracking();
+
+  void attach_hostname(const char *hostname);
 
   void set_active();
   bool is_active();
   void set_private(bool new_private = true);
   bool is_private() const;
 
-  void set_netvc(NetVConnection *newvc);
+  virtual void set_netvc(NetVConnection *newvc);
+
+  // Used to determine whether the session is for parent proxy
+  // it is session to origin server
+  // We need to determine whether a closed connection was to
+  // close parent proxy to update the
+  // proxy.process.http.current_parent_proxy_connections
+  bool to_parent_proxy = false;
+
+  // Keep track of connection limiting and a pointer to the
+  // singleton that keeps track of the connection counts.
+  OutboundConnTrack::Group *conn_track_group = nullptr;
 
   virtual IOBufferReader *get_reader() = 0;
 
@@ -191,4 +205,35 @@ inline bool
 PoolableSession::FQDNLinkage::equal(CryptoHash const &lhs, CryptoHash const &rhs)
 {
   return lhs == rhs;
+}
+
+inline void
+PoolableSession::enable_outbound_connection_tracking(OutboundConnTrack::Group *group)
+{
+  ink_assert(nullptr == conn_track_group);
+  conn_track_group = group;
+}
+
+inline void
+PoolableSession::release_outbound_connection_tracking()
+{
+  // Update upstream connection tracking data if present.
+  if (conn_track_group) {
+    if (conn_track_group->_count >= 0) {
+      (conn_track_group->_count)--;
+      conn_track_group = nullptr;
+    } else {
+      // A bit dubious, as there's no guarantee it's still negative, but even that would be interesting to know.
+      Error("[http_ss] [%" PRId64 "] number of connections should be greater than or equal to zero: %u", con_id,
+            conn_track_group->_count.load());
+    }
+  }
+}
+
+inline void
+PoolableSession::attach_hostname(const char *hostname)
+{
+  if (CRYPTO_HASH_ZERO == hostname_hash) {
+    CryptoContext().hash_immediate(hostname_hash, (unsigned char *)hostname, strlen(hostname));
+  }
 }
--- a/proxy/ProxyTransaction.cc
+++ b/proxy/ProxyTransaction.cc
@@ -221,12 +221,6 @@ ProxyTransaction::has_request_body(int64_t request_content_length, bool is_chunk
   return request_content_length > 0 || is_chunked;
 }
 
-bool
-ProxyTransaction::is_read_closed() const
-{
-  return false;
-}
-
 void
 ProxyTransaction::attach_transaction(HttpSM *attach_sm)
 {
--- a/proxy/ProxyTransaction.cc
+++ b/proxy/ProxyTransaction.cc
@@ -57,8 +57,8 @@ ProxyTransaction::new_transaction(bool from_early_data)
     }
   }
 
-  this->increment_client_transactions_stat();
-  _sm->attach_client_session(this, _reader);
+  this->increment_transactions_stat();
+  _sm->attach_client_session(this);
 }
 
 bool
@@ -181,7 +181,7 @@ void
 ProxyTransaction::transaction_done()
 {
   SCOPED_MUTEX_LOCK(lock, this->mutex, this_ethread());
-  this->decrement_client_transactions_stat();
+  this->decrement_transactions_stat();
 }
 
 // Implement VConnection interface.
@@ -219,6 +219,24 @@ bool
 ProxyTransaction::has_request_body(int64_t request_content_length, bool is_chunked) const
 {
   return request_content_length > 0 || is_chunked;
+}
+
+bool
+ProxyTransaction::is_read_closed() const
+{
+  return false;
+}
+
+void
+ProxyTransaction::attach_transaction(HttpSM *attach_sm)
+{
+  _sm = attach_sm;
+}
+
+HTTPVersion
+ProxyTransaction::get_version(HTTPHdr &hdr) const
+{
+  return hdr.version_get();
 }
 
 bool
--- a/proxy/ProxyTransaction.h
+++ b/proxy/ProxyTransaction.h
@@ -39,15 +39,17 @@ public:
   /// Virtual Methods
   //
   virtual void new_transaction(bool from_early_data = false);
+  virtual void attach_transaction(HttpSM *attach_sm);
   virtual bool attach_server_session(PoolableSession *ssession, bool transaction_done = true);
   Action *adjust_thread(Continuation *cont, int event, void *data);
-  virtual void release(IOBufferReader *r) = 0;
-  virtual void transaction_done();
+  virtual void release()          = 0;
+  virtual void transaction_done() = 0;
 
   virtual void set_active_timeout(ink_hrtime timeout_in);
   virtual void set_inactivity_timeout(ink_hrtime timeout_in);
   virtual void cancel_inactivity_timeout();
   virtual void cancel_active_timeout();
+  virtual bool is_read_closed() const;
 
   // Implement VConnection interface.
   VIO *do_io_read(Continuation *c, int64_t nbytes = INT64_MAX, MIOBuffer *buf = nullptr) override;
@@ -63,8 +65,9 @@ public:
   virtual int get_transaction_priority_weight() const;
   virtual int get_transaction_priority_dependence() const;
   virtual bool allow_half_open() const;
-  virtual void increment_client_transactions_stat() = 0;
-  virtual void decrement_client_transactions_stat() = 0;
+
+  virtual void increment_transactions_stat() = 0;
+  virtual void decrement_transactions_stat() = 0;
 
   virtual NetVConnection *get_netvc() const;
   virtual bool is_first_transaction() const;
@@ -84,6 +87,10 @@ public:
 
   // Returns true if there is a request body for this request
   virtual bool has_request_body(int64_t content_length, bool is_chunked_set) const;
+
+  sockaddr const *get_remote_addr() const;
+
+  virtual HTTPVersion get_version(HTTPHdr &hdr) const;
 
   /// Non-Virtual Methods
   //
@@ -114,14 +121,16 @@ public:
   // This function must return a non-negative number that is different for two in-progress transactions with the same proxy_ssn
   // session.
   //
-  void set_rx_error_code(ProxyError e);
-  void set_tx_error_code(ProxyError e);
+  virtual void set_rx_error_code(ProxyError e);
+  virtual void set_tx_error_code(ProxyError e);
 
   bool support_sni() const;
 
   /// Variables
   //
   HttpSessionAccept::Options upstream_outbound_options; // overwritable copy of options
+
+  IOBufferReader *get_reader();
 
 protected:
   ProxySession *_proxy_ssn = nullptr;
@@ -273,4 +282,20 @@ ProxyTransaction::adjust_thread(Continuation *cont, int event, void *data)
     }
   }
   return nullptr;
+}
+
+inline IOBufferReader *
+ProxyTransaction::get_reader()
+{
+  return _reader;
+}
+
+inline sockaddr const *
+ProxyTransaction::get_remote_addr() const
+{
+  if (_proxy_ssn) {
+    return _proxy_ssn->get_remote_addr();
+  } else {
+    return nullptr;
+  }
 }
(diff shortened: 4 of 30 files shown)


iocore/net/libinknet_stub.cc
```diff
@@ -21,11 +21,11 @@
   limitations under the License.
  */
 
-#include "HttpSessionManager.h"
+class EThread;
+class Continuation;
```
shinrich, line 25: Needed to avoid pulling in unnecessary stuff to the unit test.

proxy/PoolableSession.h
```diff
@@ -192,3 +206,34 @@ PoolableSession::FQDNLinkage::equal(CryptoHash const &lhs, CryptoHash const &rhs
 {
   return lhs == rhs;
 }
+
```
shinrich, line 209: Promoted the conntecting tracking calls to a super class.

proxy/http/Http1ServerSession.cc
```diff
@@ -223,3 +189,71 @@ void
...
 }
+
+bool
+Http1ServerSession::is_chunked_encoding_supported() const
+{
+  return true;
+}
+
+void
+Http1ServerSession ::release_transaction()
```
shinrich, line 200: Called from transaction_done to adjust the released_transacations count and cause any pending close or server session pool release to occur.

proxy/http/HttpSM.cc
```diff
@@ -3260,12 +3282,12 @@ HttpSM::is_bg_fill_necessary(HttpTunnelConsumer *c)
...
   if (c->producer->alive &&          // something there to read
                                      //      server_entry && server_entry->vc &&              // from an origin server
-                                     //      server_session && server_session->get_netvc() && // which is still open and valid
+                                     //      server_txn && server_txn->get_netvc() && // which is still open and valid
       c->producer->num_consumers > 1 // with someone else reading it
   ) {
     HttpTunnelProducer *p = nullptr;
 
-    if (!server_entry || !server_entry->vc || !server_session || !server_session->get_netvc()) {
+    if (!server_txn || !server_txn->get_netvc()) {
```
shinrich, line 3290: May need to adjust this test, since server_txn will not be null at this point even if we did already call do_io_close on it.  The get_netvc() may still be returning null, so it may still be ok.

shinrich:
[approve ci autest]

shinrich:
[ci approve clang-analyzer]

shinrich:
[approve ci clang-analyzer]

proxy/http/HttpSessionManager.cc
```diff
@@ -484,3 +489,50 @@ HttpSessionManager::release_session(PoolableSession *to_release)
...
+
+void
+ServerSessionPool::removeSession(PoolableSession *to_remove)
+{
+  EThread *ethread = this_ethread();
+  SCOPED_MUTEX_LOCK(lock, mutex, ethread);
+  char peer_ip[INET6_ADDRPORTSTRLEN];
+  if (is_debug_tag_set("http_ss")) {
+    ats_ip_nptop(to_remove->get_remote_addr(), peer_ip, sizeof(peer_ip));
+    Debug("http_ss", "Remove session %p %s m_fqdn_pool size=%" PRId64 " m_ip_pool_size=%" PRId64, to_remove, peer_ip,
```
maskit, line 501: Please use `%zu` for `size_t`. https://github.com/apache/trafficserver/pull/7830

```
HttpSessionManager.cc:502:11: error: format specifies type 'long long' but the argument has type 'size_t' (aka 'unsigned long') [-Werror,-Wformat]
          m_fqdn_pool.count(), m_ip_pool.count());
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../include/tscore/Diags.h:158:35: note: expanded from macro 'Debug'
      log_message.debug(tag, loc, __VA_ARGS__);               \
                                  ^~~~~~~~~~~
HttpSessionManager.cc:502:32: error: format specifies type 'long long' but the argument has type 'size_t' (aka 'unsigned long') [-Werror,-Wformat]
          m_fqdn_pool.count(), m_ip_pool.count());
          ~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
../../include/tscore/Diags.h:158:35: note: expanded from macro 'Debug'
      log_message.debug(tag, loc, __VA_ARGS__);               \
                                  ^~~~~~~~~~~
HttpSessionManager.cc:509:113: error: format specifies type 'long long' but the argument has type 'size_t' (aka 'unsigned long') [-Werror,-Wformat]
    Debug("http_ss", "After Remove session %p m_fqdn_pool size=%" PRId64 " m_ip_pool_size=%" PRId64, to_remove, m_fqdn_pool.count(),
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
../../include/tscore/Diags.h:158:35: note: expanded from macro 'Debug'
      log_message.debug(tag, loc, __VA_ARGS__);               \
                                  ^~~~~~~~~~~
HttpSessionManager.cc:510:11: error: format specifies type 'long long' but the argument has type 'size_t' (aka 'unsigned long') [-Werror,-Wformat]
          m_ip_pool.count());
          ^~~~~~~~~~~~~~~~~~
../../include/tscore/Diags.h:158:35: note: expanded from macro 'Debug'
      log_message.debug(tag, loc, __VA_ARGS__);               \
                                  ^~~~~~~~~~~
4 errors generated.
```

proxy/http/HttpSessionManager.cc
```diff
@@ -484,3 +489,50 @@ HttpSessionManager::release_session(PoolableSession *to_release)
...
+    HTTP_DECREMENT_DYN_STAT(http_pooled_server_connections_stat);
+  }
+  if (is_debug_tag_set("http_ss")) {
+    Debug("http_ss", "After Remove session %p m_fqdn_pool size=%" PRId64 " m_ip_pool_size=%" PRId64, to_remove, m_fqdn_pool.count(),
+          m_ip_pool.count());
+  }
+}
+
+void
+ServerSessionPool::testSession(PoolableSession *ss)
```
maskit, line 515: It seems like this functions is unnecessary. It's called by `Http2ServerSession::test_session()` on #7622, but `Http2ServerSession::test_session()` is not called.

proxy/http/HttpSessionManager.h
```diff
@@ -67,6 +67,9 @@ class ServerSessionPool : public Continuation
   static bool validate_host_sni(HttpSM *sm, NetVConnection *netvc);
   static bool validate_sni(HttpSM *sm, NetVConnection *netvc);
   static bool validate_cert(HttpSM *sm, NetVConnection *netvc);
+  void removeSession(PoolableSession *ssn);
```
maskit, line 70: Can you make `removeSession` and `addSession` private functions for now because H1ServerSession doesn't need to call them directly?

I know the new functions are used by Http2ServerSession, and we can discuss the detail later on H2-to-Origin PR, but having `removeSession` and `addSession` as public functions in addition to `acquireSession` and `releaseSession` doesn't make sense to me. The interface would be confusing. I think the both H1ServerSession and H2ServerSession should be handled by HttpSessionManager consistently - unnecessary processes should be skipped inside HttpSessionManager.

proxy/ProxyTransaction.h
```diff
@@ -85,6 +88,10 @@ class ProxyTransaction : public VConnection
   // Returns true if there is a request body for this request
   virtual bool has_request_body(int64_t content_length, bool is_chunked_set) const;
 
+  sockaddr const *get_remote_addr() const;
+
+  virtual HTTPVersion get_version(HTTPHdr &hdr) const;
```
maskit, line 93: We don't need this at the moment.

Since`ProxySession::get_version` receives `HTTPHdr`, a proxy session can always return correct version. I wonder if there will be a session that has multiple versions of transactions. I'm not sure if we are allowed to use multiple versions on a traditional HTTP connection like below:

```
[Connect]
GET /a HTTP/1.0

GET /b HTTP/1.1
Host: example.com

[Close]
```

proxy/ProxyTransaction.h
```diff
@@ -114,15 +121,17 @@ class ProxyTransaction : public VConnection
...
+  virtual void set_rx_error_code(ProxyError e);
+  virtual void set_tx_error_code(ProxyError e);
 
   bool support_sni() const;
 
   /// Variables
   //
   HttpSessionAccept::Options upstream_outbound_options; // overwritable copy of options
 
+  IOBufferReader *get_reader();
```
maskit, line 133: It's almost bike-shedding and this is not the only place, but I don't like this name very much because the data source (or the direction) is unclear. My understanding is that this reader always reads data from a peer (i.e. a remote host) regardless of the type of transaction (ClientTransaction or ServerTransaction).

How about `get_remote_data_reader` ?

proxy/http/HttpSM.cc
```diff
@@ -1800,6 +1798,65 @@ HttpSM::handle_api_return()
...
+  // session object.
+  if (s.outbound_conn_track_state.is_active()) {
+    Debug("http_connect", "[%" PRId64 "] max number of outbound connections: %d", this->sm_id, s.txn_conf->outbound_conntrack.max);
+    retval->enable_outbound_connection_tracking(s.outbound_conn_track_state.drop());
+  }
+  return retval;
+}
+
+void
+HttpSM::create_server_txn(NetVConnection *netvc, PoolableSession *new_session)
```
maskit, line 1841: This function is really confusing.

The name implies that we call this function whenever we want to create a server transaction, but that is probably incorrect. This function seems to be only used for the initial server transaction on a session.

Furthermore, this is a public function despite that HttpSM has to be the only caller. Anybody can accidentally call this function based on the wrong assumption above.

Can you make this a private function and rename it to `create_initial_server_txn`? I haven't understood why server transactions have to be created here and there, and I think it should be done at only one place, but the name would tell that is the current design.

proxy/ProxyTransaction.cc
```diff
@@ -221,6 +221,24 @@ ProxyTransaction::has_request_body(int64_t request_content_length, bool is_chunk
   return request_content_length > 0 || is_chunked;
 }
 
+bool
+ProxyTransaction::is_read_closed() const
```
zwoop, line 225: Is this used in some subsequent PRs? Seems a little odd to have this "no-op" method, which is never called as far as I can tell.

proxy/http/Http1ClientSession.cc
```diff
@@ -345,11 +352,21 @@ Http1ClientSession::state_keep_alive(int event, void *data)
 {
   // Route the event.  It is either for vc or
   //  the origin server slave vc
-  if (data && data == slave_ka_vio) {
-    return state_slave_keep_alive(event, data);
-  } else {
-    ink_assert(data && data == ka_vio);
-    ink_assert(read_state == HCS_KEEP_ALIVE);
+  if (data) {
```
zwoop, line 355: Would this possibly read nicer as a switch statement?

proxy/http/HttpConfig.cc
```diff
@@ -393,6 +393,25 @@ register_stat_callbacks()
...
+                     (int)http_origin_no_sharing, RecRawStatSyncCount);
+  RecRegisterRawStat(http_rsb, RECT_PROCESS, "proxy.process.http.origin.body", RECD_INT, RECP_NON_PERSISTENT, (int)http_origin_body,
+                     RecRawStatSyncCount);
+  RecRegisterRawStat(http_rsb, RECT_PROCESS, "proxy.process.http.origin.private", RECD_INT, RECP_NON_PERSISTENT,
+                     (int)http_origin_private, RecRawStatSyncCount);
+  RecRegisterRawStat(http_rsb, RECT_PROCESS, "proxy.process.http.origin.close_private", RECD_INT, RECP_NON_PERSISTENT,
+                     (int)http_origin_close_private, RecRawStatSyncCount);
+  RecRegisterRawStat(http_rsb, RECT_PROCESS, "proxy.process.http.origin.raw", RECD_INT, RECP_NON_PERSISTENT, (int)http_origin_raw,
+                     RecRawStatSyncCount);
+
```
zwoop, line 414: This is great!

proxy/http/HttpSM.cc
```diff
@@ -1800,6 +1798,65 @@ HttpSM::handle_api_return()
   }
 }
 
+void
+HttpSM::set_server_txn(ProxyTransaction *txn)
+{
+  ink_release_assert(server_txn == nullptr);
```
zwoop, line 1804: I would probably argue that this should be a debug assert, it kinda feels like we're too liberal sometimes with the release assert. Looking where this is called, it can't be nullptr (assuming that to_return->new_transaction() is not allowed to fail).

proxy/http/HttpSM.cc
```diff
@@ -5095,11 +5123,26 @@ HttpSM::do_http_server_open(bool raw)
...
-      existing_ss->release(nullptr);
+      existing_ss->set_inactivity_timeout(HRTIME_SECONDS(t_state.txn_conf->keep_alive_no_activity_timeout_out));
+      existing_ss->release(server_txn);
       ua_txn->attach_server_session(nullptr);
     }
   }
+
+  if (!try_reuse) {
+    HTTP_INCREMENT_DYN_STAT(http_origin_make_new);
+    if (TS_SERVER_SESSION_SHARING_MATCH_NONE == t_state.txn_conf->server_session_sharing_match) {
```
zwoop, line 5134: This code is slightly alarming; It feels like the server session would be in one of several conditions, as seen by which metric we increment. But, the logic to decide what "state" it is in is based on a number of conditions and member variables, rather than a single "state". That feels like e.g. someone could set "raw" and "is_private", and the behavior is now in a unknown state.

proxy/http/HttpSM.cc
```diff
@@ -5492,14 +5534,13 @@ HttpSM::release_server_session(bool serve_from_cache)
...
   }
 
-  ink_assert(server_entry->vc == server_session);
-  server_entry->in_tunnel = true;
-  vc_table.cleanup_entry(server_entry);
-  server_entry   = nullptr;
-  server_session = nullptr;
+  if (server_entry) {
+    server_entry->vc       = nullptr;
+    server_entry->read_vio = server_entry->write_vio = nullptr;
```
zwoop, line 5541: This would look a lot better if you don't do the double assignment on line 5541 :).

proxy/http/HttpSessionManager.cc
```diff
@@ -165,9 +165,11 @@ ServerSessionPool::acquireSession(sockaddr const *addr, CryptoHash const &hostna
     }
     if (zret == HSM_DONE) {
       to_return = first;
-      HTTP_DECREMENT_DYN_STAT(http_pooled_server_connections_stat);
-      m_fqdn_pool.erase(first);
-      m_ip_pool.erase(to_return);
+      this->removeSession(to_return);
+    } else {
+      if (first != m_fqdn_pool.end()) {
```
zwoop, line 170: Bike shedding, but for easier read, why not make this an } else if { ?

proxy/http/Http1ServerSession.cc
```diff
@@ -93,26 +98,19 @@ Http1ServerSession::new_connection(NetVConnection *new_vc, MIOBuffer *iobuf, IOB
...
-Http1ServerSession::enable_outbound_connection_tracking(OutboundConnTrack::Group *group)
+Http1ServerSession::do_io_close(int alerrno)
 {
-  ink_assert(nullptr == conn_track_group);
-  conn_track_group = group;
-  if (is_debug_tag_set("http_ss")) {
-    ts::LocalBufferWriter<256> w;
-    w.print("[{}] new connection, ip: {}, group ({}), count: {}\0", con_id, get_server_ip(), *group, group->_count);
-    Debug("http_ss", "%s", w.data());
+  if (state == SSN_CLOSED) { // Already been closed
```
maskit, line 103: I prefer this way. It has less dup code.
```
if (state != SSN_CLOSED) {
  state = SSN_CLOSED
  release outbound connection tracking
  close vc
  decrement stats
}

if (transact_count == released_transactions) {
  this->destroy();
}
```

proxy/http/HttpSessionManager.cc
shinrich: Yes, this one can be removed.

proxy/http/HttpSessionManager.h
shinrich: Sure, in this use case they can be made private

proxy/ProxyTransaction.h
shinrich: I am not sure CONNECT makes my head hurt.  I would think that only one nested method would be interpreted over the CONNECT method.  And ATS would only interpret it if the CONNECT was made over a non-TLS connection.  Otherwise ATS treats the body as a blind tunnel.

Will have to read the RFC tea leaves on this.

proxy/ProxyTransaction.h
shinrich: I'm ok with get_remote_reader or remote_data_reader.  Just added this as a wrapper method around the _reader member.

proxy/http/HttpSM.cc
shinrich: create_initial_server_txn is probably a better name.  And it can be made private.  The goal of the function is to wire in the netvc into a ProxyTransaction object (Http1ServerTransaction in this case).   It is taking logic that was inline in the HttpSM::state_http_server_open function.  It is called a bit later once H2 to origin is introduced because we need to wait to see what protocol is negotiated to create the correct subclass of ProxyTransaction.

proxy/ProxyTransaction.cc
shinrich: Probably leftover from the H2 to origin pull back.

proxy/http/HttpSM.cc
shinrich: Agree.  I use the release_asserts liberally while debugging in production.  Don't always get them cleaned up.  Looking forward to using your enable-all asserts option.

proxy/http/HttpSM.cc
shinrich: I added these stats for debugging to solve my last issue before posting this PR.  No doubt could be tidied up.  Or removed entirely.

proxy/http/HttpSM.cc
shinrich: Reworked so create_server_txn is called in HttpSessionManager too instead of calling new_transaction directly

proxy/ProxyTransaction.cc
shinrich: Removed for now.  Will be added back with the H2 to origin logic.

proxy/http/Http1ClientSession.cc
shinrich: Cannot have variables as case elements.

maskit (commented):
Although I can't say this doesn't break things, but it looks good to me. I'm going to approve this when conflicts are resolved.

shinrich:
[approve ci autest]

maskit (approved):

PR #7859, opened 2021-05-18
Thread safe Mersenne Twister 64 using c++11

bryancall:
Looking at using this to replace drand48() and InkRand in our code.

### DIFF
--- a/include/tscore/Random.h
+++ b/include/tscore/Random.h
@@ -1,0 +1,56 @@
+/** @file
+
+  Pseudorandom Number Generator
+
+  @section license License
+
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+
+#include <random>
+
+namespace ts
+{
+class Random
+{
+public:
+  Random() = delete;
+
+  static uint64_t
+  random()
+  {
+    return _dist_int(_engine);
+  }
+
+  static double
+  drandom()
+  {
+    return _dist_real(_engine);
+  }
+
+  static void
+  seed(uint64_t s)
+  {
+    _engine.seed(s);
+  }
+
+private:
+  thread_local static std::mt19937_64 _engine;
+  thread_local static std::uniform_int_distribution<uint64_t> _dist_int;
+  thread_local static std::uniform_real_distribution<double> _dist_real;
+};
+}; // namespace ts
--- a/include/tscore/Random.h
+++ b/include/tscore/Random.h
@@ -33,13 +33,15 @@ public:
   static uint64_t
   random()
   {
-    return _dist_int(_engine);
+    std::uniform_int_distribution<uint64_t> dist{0, UINT64_MAX};
+    return dist(_engine);
   }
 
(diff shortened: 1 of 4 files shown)


zwoop:
Are you making the other changes (e.g. removing ink_rand.{cc,h} in another PR?

bryancall:
Graph of the first number generated per thread:

Screen Shot 2021-05-19 at 8 43 36 AM

src/tscore/unit_tests/test_Random.cc
```diff
@@ -0,0 +1,52 @@
...
+  }
+  auto diff = std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::high_resolution_clock::now() - start);
+  std::cout << (double)diff.count() / 1000000 << " ns per Random::random()" << std::endl;
+
+  start = std::chrono::high_resolution_clock::now();
+  for (auto i = 0; i < 1000000; ++i) {
+    x.random();
+  }
+  diff = std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::high_resolution_clock::now() - start);
+  std::cout << (double)diff.count() / 1000000 << " ns per Random::random()" << std::endl;
```
bneradt, line 51: per InkRand::random()

bneradt (approved):
Looks good.

bryancall:
[approve ci autest]

src/tscore/unit_tests/test_Random.cc
```diff
@@ -0,0 +1,52 @@
...
+TEST_CASE("test random", "[libts][random]")
+{
+  ts::Random::seed(13);
+  InkRand x(13);
+
+  for (auto i = 0; i < 1000000; ++i) {
+    REQUIRE(ts::Random::random() == x.random());
+    REQUIRE(uint64_t(ts::Random::drandom() * 100000000) == uint64_t(x.drandom() * 100000000)); // they are pretty close to the same
+  }
+
```
ywkaras, line 38: I would `#if 0` out the speed test, as I don't think there's value in running it in regression.

include/tscore/Random.h
```diff
@@ -0,0 +1,56 @@
...
+
+  static double
+  drandom()
+  {
+    std::uniform_real_distribution<double> dist{0, 1};
+    return dist(_engine);
+  }
+
+  static void
+  seed(uint64_t s)
```
ywkaras, line 48: Seems like this would have to be called exactly once on each thread.  How would that be done?

src/tscore/Random.cc
```diff
@@ -0,0 +1,29 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+
+#include "tscore/Random.h"
+
+namespace ts
+{
+thread_local std::mt19937_64 Random::_engine{std::random_device{}()};
```
ywkaras, line 28: std::mt19937_64 has a constructor that sets. default seed ( https://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine/mersenne_twister_engine ).  I don't think we can be confident that _engine will get initialize, because our threads are not created by std::thread objects.  cppreference says thread-local objects are initialized "initialized as part of thread launch" ( https://en.cppreference.com/w/cpp/language/initialization#Non-local_variables ).  Seem likely the initialization is done in the std::thread constructor.

include/tscore/Random.h
bryancall: This is for people that want to seed if after it has been seeded.  There are some tests were we seed with a constant value, so we have reproducible tests.

src/tscore/unit_tests/test_Random.cc
bryancall: done

ywkaras:
Could you throw your test program into a gist?

ywkaras:
I tinkered a bit with the test program ( https://github.com/ywkaras/MiscRepo/blob/master/THREAD_LOCAL/b.cc ).  But nonetheless it finds no problems with thread_local using clang on MacOS and gcc on RHEL(ish ylinux).

I approve this, but not telling Github, so non-Yahoos will see they need to +1 it.

randall (approved):

ywkaras:
@bryancall did you want to make changes before merging?

PR #7825, opened 2021-05-12
Eliminate next dup call using stale mime field mloc in s3_auth plugin.

ywkaras:

PR #7860, opened 2021-05-19
Eliminate next dup call using stale mime field mloc is s3_auth plugin (#7825)

ywkaras:
(cherry picked from commit f8ddad16c3be2e142c8f6782449fda35e0020318)


ywkaras:
@ezelkow1

PR #7868, opened 2021-05-21
Fixed ASAN issues with MMH test

bryancall:
This fixes the master build on Fedora 32 that are using ASAN

### DIFF
--- a/src/tscore/unit_tests/test_MMH.cc
+++ b/src/tscore/unit_tests/test_MMH.cc
@@ -83,15 +83,18 @@ TEST_CASE("MMH", "[libts][MMH]")
       printf("********** collision %d\n", xy);
   }
 
-  unsigned char *s  = (unsigned char *)MMH_x;
-  int l             = sizeof(MMH_x);
-  unsigned char *s1 = (unsigned char *)ats_malloc(l + 3);
+  unsigned char *s       = (unsigned char *)MMH_x;
+  int l                  = sizeof(MMH_x);
+  unsigned char *s1      = (unsigned char *)ats_malloc(l + 4);
+  unsigned char *free_s1 = s1;
   s1 += 1;
   memcpy(s1, s, l);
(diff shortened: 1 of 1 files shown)


src/tscore/unit_tests/test_MMH.cc
```diff
@@ -83,15 +83,18 @@ TEST_CASE("MMH", "[libts][MMH]")
       printf("********** collision %d\n", xy);
   }
 
-  unsigned char *s  = (unsigned char *)MMH_x;
-  int l             = sizeof(MMH_x);
-  unsigned char *s1 = (unsigned char *)ats_malloc(l + 3);
+  unsigned char *s       = (unsigned char *)MMH_x;
+  int l                  = sizeof(MMH_x);
+  unsigned char *s1      = (unsigned char *)ats_malloc(l + 4);
```
bneradt, line 88: Makes sense. What do you think of using `sizeof(int32_t)` instead of `4`? It would make the intention more explicit.

`unsigned char *s1      = (unsigned char *)ats_malloc(l + sizeof(int32_t));`

bryancall: I made a change to use sizeof(uint32_t)

bneradt (approved):

src/tscore/unit_tests/test_MMH.cc
```diff
@@ -83,15 +83,18 @@ TEST_CASE("MMH", "[libts][MMH]")
       printf("********** collision %d\n", xy);
   }
 
-  unsigned char *s  = (unsigned char *)MMH_x;
-  int l             = sizeof(MMH_x);
-  unsigned char *s1 = (unsigned char *)ats_malloc(l + 3);
+  unsigned char *s       = (unsigned char *)MMH_x;
```
randall, line 86: should the c-style casts be static_casts?

zwoop (approved):
I'm fine leaving the c-style casts, but amc will disagree :-).

PR #7873, opened 2021-05-24
Remove second call to TRANSACT_RETURN while handling cache write lock

### LINKED ISSUE #7872: Double calling TRANSACT_RETURN from HttpTransact::handle_cache_write_lock
Noticed this while walking code trying to see how an assert from @ezelkow1 could occur.  Looking at the last if/else clause in HttpTransact::handle_cache_write_lock

```
  if (s->cache_info.write_lock_state == CACHE_WL_READ_RETRY) {
    TxnDebug("http_error", "calling hdr_info.server_request.destroy");
    s->hdr_info.server_request.destroy();
    HandleCacheOpenReadHitFreshness(s);
  } else {
    StateMachineAction_t next;
    next = how_to_open_connection(s);
    if (next == SM_ACTION_ORIGIN_SERVER_OPEN || next == SM_ACTION_ORIGIN_SERVER_RAW_OPEN) {
      s->next_action = next;
      TRANSACT_RETURN(next, nullptr);
    } else {
      // hehe!
      s->next_action = next;
      ink_assert(s->next_action == SM_ACTION_DNS_LOOKUP);
      return;
    }
    TRANSACT_RETURN(next, nullptr);
  }
```
If the outer else is taken and the next is SM_ACTION_ORIGIN_SERVER_OPEN, then TRANSACT_RETURN is called in the if body and also on return through the outer else.

Doesn't seem like a good thing.  Appears to be accidental change from PR #1794

shinrich:
Discussion in the issue.

This closes #7872

### DIFF
--- a/proxy/http/HttpTransact.cc
+++ b/proxy/http/HttpTransact.cc
@@ -3265,8 +3265,6 @@ HttpTransact::handle_cache_write_lock(State *s)
       ink_assert(s->next_action == SM_ACTION_DNS_LOOKUP);
       return;
     }
-
-    TRANSACT_RETURN(next, nullptr);
   }
 }
 


vmamidi (approved):

shinrich:
Rob pointed out that the TRANSACT_RETURN is a macro that actually returns.  So that makes this PR at most a clean up.  I think we should go ahead and do it for the cleanup benefits.

rob05c (approved):
LGTM. The immediate prior block does `if { TRANSACT_RETURN } else { return }`, and `TRANSACT_RETURN` is a macro that ends in `return`. So it shouldn't be possible for this to change behavior.

ezelkow1 (approved):

PR #7875, opened 2021-05-24
Restrict HTTP versions allowed on the HTTP request line

sudheerv:
This change restricts allowed HTTP versions to 1.0, 1.1 on the
HTTP request line to prevent potential mishandling, request
smugging or other vulns due to random/arbitrary version tags

Note that HTTP/2.0 and HTTP/3.0 are negotiated via ALPN on TLS
and not via the HTTP request line.

### DIFF
--- a/proxy/hdrs/HTTP.cc
+++ b/proxy/hdrs/HTTP.cc
@@ -623,12 +623,36 @@ http_hdr_type_set(HTTPHdrImpl *hh, HTTPType type)
 }
 
 /*-------------------------------------------------------------------------
+  RFC2616 specifies that HTTP version is of the format <major>.<minor>
+  in the request line.  However, the features supported and in use are
+  for versions 1.0, 1.1 and 2.0 (with HTTP/3.0 being developed). HTTP/2.0
+  and HTTP/3.0 are both negotiated using ALPN over TLS and not via the HTTP
+  request line thus leaving the versions supported on the request line to be
+  HTTP/1.0 and HTTP/1.1 alone. This utility checks if the HTTP Version
+  received in the request line is one of these and returns false otherwise
   -------------------------------------------------------------------------*/
 
-void
+bool
+is_version_supported(const uint8_t major, const uint8_t minor)
+{
+  if (major == 1) {
+    return minor == 1 || minor == 0;
+  }
+
+  return false;
+}
+
+bool
+is_http_hdr_version_supported(const HTTPVersion &http_version)
+{
+  return is_version_supported(http_version.get_major(), http_version.get_minor());
+}
+
+bool
 http_hdr_version_set(HTTPHdrImpl *hh, const HTTPVersion &ver)
 {
   hh->m_version = ver;
+  return is_version_supported(ver.get_major(), ver.get_minor());
 }
 
 /*-------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


proxy/hdrs/HTTP.cc
```diff
@@ -625,7 +625,27 @@ http_hdr_type_set(HTTPHdrImpl *hh, HTTPType type)
 /*-------------------------------------------------------------------------
   -------------------------------------------------------------------------*/
 
-void
+bool
+is_version_supported(const uint8_t major, const uint8_t minor)
```
maskit, line 629: Can you clarify what "supported" means? Having a (doxygen) comment for this function would be helpful.

Do you think we should not have 3(.0) here yet?

proxy/hdrs/HTTP.cc
```diff
@@ -625,7 +625,27 @@ http_hdr_type_set(HTTPHdrImpl *hh, HTTPType type)
...
+  return false;
+}
+
+bool
+is_http_hdr_version_supported(const HTTPVersion &http_version)
+{
+  return is_version_supported(http_version.get_major(), http_version.get_minor());
+}
+
+bool
```
maskit, line 648: This function doesn't return a value.

maskit:
It seems like the parser still accepts an invalid request line like `GET / HTTP/2.0` since 2.0 is "supported". Shouldn't we have only `1.1` and `1.0` for the restriction?

proxy/hdrs/HTTP.cc
sudheerv: Ugh, copy paste error from internal repo to ASF. Fixed.

proxy/hdrs/HTTP.cc
sudheerv: Added a comment. I think we can update this when 3.0 is supported

sudheerv:
> It seems like the parser still accepts an invalid request line like `GET / HTTP/2.0` since 2.0 is "supported". Shouldn't we have only `1.1` and `1.0` for the restriction?

Hmm, yeah, good point. I'd be okay to remove 2.0 if you prefer

sudheerv:
[approve ci autest]

maskit:
I don't have a strong opinion on whether we allow the invalid request line. However, I think the restriction should be either a) accept all the versions (i.e. `1.1`, `1.0`, `2.0` and `3.0`) that are on the RFC you quoted && can be handled by ATS , or b) accept only `1.1` and `1.0` that are the only valid versions in reality.

Option a) could be complicated. While ATS has experimental support for H3, it is only available if ATS is built with a SSL library that supports QUIC (special OpenSSL or BoringSSL). We may want to have `ifdef` in the function if we take option a). This is why I wanted you to clarify what you mean by "supported". Strictly speaking, we should check whether ALPN (or NPN) is available to say `2.0` is supported in that sense.

My biggest question was why the function doesn't allow 3.0 where it allows 2.0, but now I'm inclined to option b) because of the complexity above. I don't see benefits to have the complicated function at the moment.

sudheerv:
> I don't have a strong opinion on whether we allow the invalid request line. However, I think the restriction should be either a) accept all the versions (i.e. `1.1`, `1.0`, `2.0` and `3.0`) that are on the RFC you quoted && can be handled by ATS , or b) accept only `1.1` and `1.0` that are the only valid versions in reality.
> 
> Option a) could be complicated. While ATS has experimental support for H3, it is only available if ATS is built with a SSL library that supports QUIC (special OpenSSL or BoringSSL). We may want to have `ifdef` in the function if we take option a). This is why I wanted you to clarify what you mean by "supported". Strictly speaking, we should check whether ALPN (or NPN) is available to say `2.0` is supported in that sense.
> 
> My biggest question was why the function doesn't allow 3.0 where it allows 2.0, but now I'm inclined to option b) because of the complexity above. I don't see benefits to have the complicated function at the moment.

+1 

Agree with the reasoning. Updated to remove HTTP/2.0 as well for consistency. 

Thanks for the review!

maskit (approved):

bryancall:
Cherry picked to 9.1.x

PR #7877, opened 2021-05-25
Adds a new config to parallelize loading of ssl_multicert.config

zwoop:
The new setting is (default: 0):

    CONFIG proxy.config.ssl.server.multicert.concurrency INT 24

A setting of -1 will create one loader thread per system CPU. Still testing performance implications here, but this likely is only a noticeable win with many, many thousands of certificates, or, a very slow "file system".

I intentionally left the line parser as-is, to avoid too much breakage and change. I think we should redo this parser, and use standard STL streams etc. to read the file line by line rather than loading the entire file into memory.

### DIFF
--- a/doc/admin-guide/files/records.config.en.rst
+++ b/doc/admin-guide/files/records.config.en.rst
@@ -3463,6 +3463,13 @@ SSL Termination
    :file:`ssl_multicert.config` file successfully load.  If false (``0``), SSL certificate
    load failures will not prevent |TS| from starting.
 
+.. ts:cv:: CONFIG proxy.config.ssl.server.multicert.concurrency INT 0
+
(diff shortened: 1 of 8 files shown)


maskit:
This is just FYI, but certs are loaded twice if you enable QUIC because QUICMultiCertConfigLoader is created and it does basically the same thing as SSLMultiCertConfigLoader with a couple of different options for SSL_CTX for QUIC.

randall:
[ci approve autest]

zwoop:
if anyone wants to test / play with this (I'm still baffled that loading certs is so CPU expensive!), I have a tar-ball with 20,000 certificates and the accompanying ssl_multicert.config file.

bryancall:
@shinrich is going to review
@zwoop please fix the conflicts

Issue #7852, opened 2021-05-18
HTTP/2 Session Refactoring


masaori335:
# Overview
Change Http2Session and Http2Stream to have appropriate states as Continuation Handler instead of having a bunch of flags.

# Motivation

For a long time, we've been struggling bunch of issues with the HTTP/2 subsystem. IMO, one of the root causes is Http2Session and Http2Stream are handling every event on only `main_event_handler` and handling many corner cases with flags.
Now, we're going to support the HTTP/2 Session on the origin server side too, this might make the situation worse. 

/cc #7622

masaori335:
Below is the first idea of states (v0.1).

## HTTP/2 Session
- https://gist.github.com/masaori335/f16a851b50bec2f69e4034844a40ab59#file-http2ssn-uml
- #7845 has piece of code for this 
<img width="829" alt="Screen Shot 2021-05-18 at 9 56 36" src="https://user-images.githubusercontent.com/741391/118574484-813fa300-b7bf-11eb-8946-9f0c3aacd053.png">

## HTTP/2 Transaction
- https://gist.github.com/masaori335/f16a851b50bec2f69e4034844a40ab59#file-http2txn-uml
- We might want a state after signaling EOS or ERROR to the HttpSM with this.
<img width="627" alt="Screen Shot 2021-05-18 at 10 09 38" src="https://user-images.githubusercontent.com/741391/118575370-463e6f00-b7c1-11eb-95f3-7fe1f5c485a5.png">

maskit:
Please keep #5675 in mind, although it doesn't need to be done on the first step of this.

PR #7845, opened 2021-05-17
HTTP/2 Session & ConnectionState Cleanup

masaori335:
Part of address #6877. One of things makes HTTP/2 Session complicated is `Http2ClientSession` & `Http2ConnectionState` are calling each other as `Continuation`. To make this simple, I propose below.

1). Make `Http2ConnectionState` just a class
2). Introduce states of HTTP/2 session as Continuation Handler

Screen Shot 2021-07-07 at 15 50 21

https://gist.github.com/masaori335/f16a851b50bec2f69e4034844a40ab59#file-http2ssn-uml

PR #7878, opened 2021-05-26
Cleanup: Get rid of HTTP2_SESSION_EVENT_INIT

masaori335:
Part of #7852. The definition will be removed by #7845 with other deprecated events.

### DIFF
--- a/proxy/http2/Http2ConnectionState.cc
+++ b/proxy/http2/Http2ConnectionState.cc
@@ -1040,8 +1040,9 @@ Http2ConnectionState::Http2ConnectionState() : stream_list()
 }
 
 void
-Http2ConnectionState::init()
+Http2ConnectionState::init(Http2ClientSession *ssn)
 {
+  ua_session         = ssn;
   this->_server_rwnd = Http2::initial_window_size;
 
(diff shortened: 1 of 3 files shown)


proxy/http2/Http2ConnectionState.cc
```diff
@@ -1100,33 +1125,6 @@ Http2ConnectionState::main_event_handler(int event, void *edata)
   }
   ++recursion;
   switch (event) {
-  // Initialize HTTP/2 Connection
-  case HTTP2_SESSION_EVENT_INIT: {
-    ink_assert(this->ua_session == nullptr);
-    this->ua_session = static_cast<Http2ClientSession *>(edata);
-    REMEMBER(event, this->recursion);
```
maskit, line 1107: We've been losing hints by removing calls for REMEMBER(). Isn't it worth keeping it? I'm fine with remove it temporally but want to put them back before we release 9.2.x (assuming we don't have this change on 9.1).

masaori335: Fair enough, I'll back REMEMBER.

shinrich (approved):

PR #7879, opened 2021-05-26
Cleanup: Get rid of HTTP2_SESSION_EVENT_RECV

masaori335:
Part of #7852. The definition will be removed by #7845 with other deprecated events.

### DIFF
--- a/proxy/http2/Http2ClientSession.cc
+++ b/proxy/http2/Http2ClientSession.cc
@@ -523,7 +523,8 @@ Http2ClientSession::do_complete_frame_read()
   ink_release_assert(this->_read_buffer_reader->read_avail() >= this->current_hdr.length);
 
   Http2Frame frame(this->current_hdr, this->_read_buffer_reader, this->cur_frame_from_early_data);
-  send_connection_event(&this->connection_state, HTTP2_SESSION_EVENT_RECV, &frame);
+  connection_state.rcv_frame(&frame);
+
   // Check whether data is read from early data
   if (this->read_from_early_data > 0) {
     this->read_from_early_data -=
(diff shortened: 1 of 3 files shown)


maskit:
I appreciate making small changes, but is it safe to land these separately? The logic is not completely equivalent - zombie event stuff won't be processed and `recursion` is not counted.

masaori335:
I agree this is not completely equivalent. As you pointed out `recursion` of `Http2ConnectionState` is not incremented. But this change is safe because the `rcv_frame()` call is guarded by `recursion` of Http2ClientSession.

Could you give me a pointer where the zombi event is checking the `recursion`?

maskit:
I don't thinks the two are related. I just pointed out two separated things that are currently done in `main_event_handler`.

`ready_to_free()` is checked at the bottom of `main_event_handler` when `recursion` is 0, and `ua_session->free()` is called. Is a similar thing done in somewhere else, or we don't need to do it when we call `rcv_frame`?

masaori335:
https://github.com/apache/trafficserver/blob/050b2dfe9a89348b4105183a2b31e7cc59fc389d/proxy/http2/Http2ConnectionState.cc#L1253-L1254

It looks like when the `main_event_handler` handles `HTTP2_SESSION_EVENT_RECV`, `ua_session->is_recursing()` is always true. Because it only comes from Http2ClientSession SessionHandler.

masaori335:
I'm fine with keeping this (and #7878) as a draft and merge with #7845 if reviewers want. This is separated out from it to minimize the change.

masaori335:
[approve ci autest]

maskit:
> I'm fine with keeping this (and #7878) as a draft and merge with #7845 if reviewers want.

I'm not requesting that, it's an option if other reviews want it though. I was just not sure the points above.

shinrich (approved):

shinrich:
Looks good.  Code more clear.

PR #7882, opened 2021-05-26
Add an HTTP/2 304 "Not Modified" AuTest.

### LINKED ISSUE #3147: Add AuTest for HTTP/2 and 304 Not Modified
I tried to cover the case of #3146 with curl (*1), but it looks like curl doesn't wait END_STREAM when it received 304 responses. It'd be grate if this case is covered by someway.

(*1) : https://github.com/masaori335/trafficserver/pull/3

bneradt:
This adds an HTTP/2 304 "Not Modified" test to verify that we handle it
correctly.

This fixes #3147

### DIFF
--- a/tests/gold_tests/headers/cachedIMSRange.test.py
+++ b/tests/gold_tests/headers/cachedIMSRange.test.py
@@ -81,16 +81,34 @@ response_etagfill_header = {
 server.addResponse("sessionlog.json", request_etagfill_header, response_etagfill_header)
 
 # ATS Configuration
-ts = Test.MakeATSProcess("ts")
+ts = Test.MakeATSProcess("ts", enable_tls=True)
 ts.Disk.plugin_config.AddLine('xdebug.so')
+ts.addDefaultSSLFiles()
+ts.Disk.ssl_multicert_config.AddLine(
+    'dest_ip=* ssl_cert_name=server.pem ssl_key_name=server.key'
+)
 ts.Disk.records_config.update({
     'proxy.config.diags.debug.enabled': 1,
     'proxy.config.diags.debug.tags': 'http',
     'proxy.config.http.response_via_str': 3,
+    'proxy.config.ssl.server.cert.path': '{0}'.format(ts.Variables.SSLDir),
+    'proxy.config.ssl.server.private_key.path': '{0}'.format(ts.Variables.SSLDir),
 })
 
-ts.Disk.remap_config.AddLine(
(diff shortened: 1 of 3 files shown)


mlibbey:
- Should we add a range request that updates a stale object?
- with h2 expected to have more requests per stream, should we test multiple requests?
- Should we add any cases from https://cache-tests.fyi/#update304 ?

maskit:
Having more tests is a good thing so I'm not going to block this, however, I wouldn't put the test under h2 directory because 304 is not an H2 specific thing. I wrote this on some previous PR before, but I think we should put tests under h2 directory only if those check H2 specific behavior. Other tests should be under other directories accordingly and the tests should be ran on all the HTTP versions.

bneradt:
> Having more tests is a good thing so I'm not going to block this, however, I wouldn't put the test under h2 directory because 304 is not an H2 specific thing. I wrote this on some previous PR before, but I think we should put tests under h2 directory only if those check H2 specific behavior. Other tests should be under other directories accordingly and the tests should be ran on all the HTTP versions.

Sounds good. I've moved the test to under cachedIMSRange.test.py which has other 304 tests with it already. I've also added HTTP and HTTPS versions of this test as well.

Thanks for the review.

bneradt:
> * Should we add a range request that updates a stale object?
> * with h2 expected to have more requests per stream, should we test multiple requests?
> * Should we add any cases from https://cache-tests.fyi/#update304 ?

These are good ideas. Thank you. For this PR, however, I'd like to stick to just this more narrow coverage. This PR is made to address a particular test need described in #3147 which couldn't be satisfied at the time due to a deficiency in curl's HTTP/2 read behavior.

maskit (approved):
Looks good to me, but I'm not sure if we can close #3147. The motivation of #3147 seems like to be testing 304 response with an empty body.

maskit:
Use of Proxy Verifier was suggested. Is it possible to test the case with curl now?

bneradt:
> Looks good to me, but I'm not sure if we can close #3147. The motivation of #3147 seems like to be testing 304 response with an empty body.

The 304 from ATS has an empty body.

> Use of Proxy Verifier was suggested. Is it possible to test the case with curl now?

Yes, Proxy Verifier was suggested because curl wasn't dealing with the end of the stream appropriately. But it seems to be now. When I add debug to the curl I'm using, I see that the on_stream_close callback is appropriately called.

PR #7886, opened 2021-05-26
Enforce case for well known methods

shinrich:
https://datatracker.ietf.org/doc/html/rfc2616#section-5.1.1 Says “The method is case-sensitive.”

### DIFF
--- a/proxy/hdrs/HdrToken.cc
+++ b/proxy/hdrs/HdrToken.cc
@@ -549,6 +549,30 @@ hdrtoken_tokenize_dfa(const char *string, int string_len, const char **wks_strin
 }
 
 /*-------------------------------------------------------------------------
+  Have to work around that methods are case insensitive while the DFA is
+  case insensitive.
+  -------------------------------------------------------------------------*/
+
+int
+hdrtoken_method_tokenize(const char *string, int string_len)
+{
+  const char *string_out;
+  int retval = -1;
+  if (hdrtoken_is_wks(string)) {
+    retval = hdrtoken_wks_to_index(string);
+    return retval;
+  }
+  retval = hdrtoken_tokenize(string, string_len, &string_out);
+  if (retval >= 0) {
+    if (strncmp(string, string_out, string_len) != 0) {
+      // Not a case match
+      retval = -1;
+    }
+  }
+  return retval;
+}
+
+/*-------------------------------------------------------------------------
   -------------------------------------------------------------------------*/
 
 int
--- a/tests/gold_tests/headers/good_request_after_bad.test.py
+++ b/tests/gold_tests/headers/good_request_after_bad.test.py
@@ -27,7 +27,7 @@ ts = Test.MakeATSProcess("ts", enable_cache=True)
 
 ts.Disk.records_config.update({'proxy.config.diags.debug.tags': 'http',
                                'proxy.config.diags.debug.enabled': 0,
-Test.ContinueOnFail = True
+                               Test.ContinueOnFail=True
                                })
 
 server = Test.MakeOriginServer("server")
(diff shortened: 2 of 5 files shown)


tests/gold_tests/headers/good_request_after_bad.test.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+# Make a good request to get item in the cache for later tests
+tr = Test.AddTestRun("Good control")
+tr.Processes.Default.StartBefore(server)
+tr.Processes.Default.StartBefore(Test.Processes.ts)
+tr.Processes.Default.Command = 'printf "GET / HTTP/1.1\r\nHost: bob\r\n\r\n" | nc  127.0.0.1 {}'.format(ts.Variables.port)
+tr.Processes.Default.ReturnCode = 0
+
+# Mixed case method name;  Should fail but ATS is treating gET as GET
```
bneradt, line 52: Not with your fix in this patch. I suggest something like:

```
# Methods are case sensitive. Verify that "gET" is not confused with "GET".

**Review by bneradt — CHANGES_REQUESTED** [review]

**Comment by masaori335** [comment]
> [RFC 7230 3.1.1.  Request Line](https://datatracker.ietf.org/doc/html/rfc7230#section-3.1.1)
> The method token indicates the request method to be performed on the
> target resource.  The request method is case-sensitive.
>
> method         = token
>
> The request methods defined by this specification can be found in
> Section 4 of [RFC7231], along with information regarding the HTTP
> method registry and considerations for defining new methods.
> ...
> Recipients of an invalid request-line SHOULD respond with either a
> 400 (Bad Request) error or a 301 (Moved Permanently) redirect with
> the request-target properly encoded.

If we think the lower-case or mixed-case method is "invalid request-line", we should send 400 or 301.

> [RFC 7231 4.  Request Methods](https://datatracker.ietf.org/doc/html/rfc7231#section-4.1)
> 4.1.  Overview
> ...
> By convention, standardized methods are defined in all-uppercase
   US-ASCII letters.
   ...
> The set of methods allowed by a target resource can be listed in an
   Allow header field (Section 7.4.1).  However, the set of allowed
   methods can change dynamically. When a request method is received
   that is unrecognized or not implemented by an origin server, the
   origin server SHOULD respond with the 501 (Not Implemented) status
   code.  When a request method is received that is known by an origin
   server but not allowed for the target resource, the origin server
   SHOULD respond with the 405 (Method Not Allowed) status code.

If we think it's "unrecognized" or "not implemented" method, we SHOULD send 501.

**Comment by shinrich** [comment]
I vote for "not recognized" (501) for mixed case versions of methods that ATS does interact with. 

You could make a case either way though, and I would change the PR if others feel strongly about using the other status code.

**Review by bneradt — DISMISSED** [review]
Looks good.

File: proxy/hdrs/HdrToken.cc [hunk 2]
```diff
@@ -548,6 +548,30 @@ hdrtoken_tokenize_dfa(const char *string, int string_len, const char **wks_strin
   return wks_idx;
 }
 
+/*-------------------------------------------------------------------------
+  Have to work around that methods are case insensitive while the DFA is
+  case insensitive.
+  -------------------------------------------------------------------------*/
+
+int
+hdrtoken_method_tokenize(const char *string, int string_len)
```
masaori335, line 557: Note: performance perspective, it looks like this has some overhead, but it might not be matter. Because `GET` is processed in the fast path and doesn't come here.

https://github.com/apache/trafficserver/blob/f002cdbae85509f339f020051d68f8cecdf17bd1/proxy/hdrs/HTTP.cc#L935-L940

masaori335 (dismissed):
I'm +1 on returning `501` and the check looks good. 

Please resolve the conflicts of AuTest.

masaori335:
[approve ci autest]

masaori335 (approved):

zwoop:
@shinrich In hdrtoken_method_tokenize(), why does it have to call hdrtoken_is_wks(string) first, when hdtoken_tokenize() does exactly the same thing when you call that ?

shinrich:
You are right it looks like the separate hdrtoken_is_wks is unnecessary.  Does save a string compare, but likely pretty short string and not worth the added complexity.  Should probably to the comparison in any case.  I'll put up another PR to simplify tomorrow.

zwoop:

PR #7887, opened 2021-05-26
Fail parsing the request line version if there are too many characters

shinrich:


maskit:
Running autest for testing text parser sounds like a great idea 😛

tests/gold_tests/headers/gold/bad_good_request_v10.gold
```diff
@@ -0,0 +1,22 @@
...
+
+<BODY BGCOLOR="white" FGCOLOR="black">
+<H1>Bad Request</H1>
+<HR>
+
+<FONT FACE="Helvetica,Arial"><B>
+Description: Could not process this request.
+</B></FONT>
+<HR>
+</BODY>
```
bneradt, line 22: I wonder whether this gold file is verifying too much? Any small changes in this output, even by one character, will cause it to fail. Perhaps:

```
HTTP/1.0 400 Invalid HTTP Request
Date: ``
Server: ATS/``
Cache-Control: no-store
Content-Type: text/html
Content-Language: en
Content-Length: ``

``
<TITLE>Bad Request</TITLE>
``
Description: Could not process this request.
``
```

bneradt (changes requested):
Looks good.

tests/gold_tests/headers/gold/bad_good_request_v10.gold
shinrich: I agree to a certain extent, but the test is also ensuring that the following request is not processed which the wildcard at the end would not catch.  Guess I could also do this via Contains and Excludes tests.

shinrich: But the problem of the excludes test is if the following request is processed in a different way than I would expect, the excludes test would not catch it.

proxy/hdrs/URL.cc
```diff
@@ -1194,6 +1194,12 @@ url_is_strictly_compliant(const char *start, const char *end)
 } // namespace UrlImpl
 using namespace UrlImpl;
 
+bool
+url_is_compliant(const char *start, const char *end)
+{
+  return url_is_strictly_compliant(start, end);
+}
+
```
shinrich, line 1202: Have to do this wrapper because url_is_strictly_compliant is in the UrlImpl space.   I tried taking it out and it wouldn't build because of a missing hash table maybe because it is calling ParseRules?

tests/gold_tests/headers/gold/bad_good_request_v10.gold
bneradt: Ah, OK. That makes sense.

bneradt (approved):

proxy/hdrs/HTTP.cc
```diff
@@ -1088,6 +1088,11 @@ http_parser_parse_req(HTTPParser *parser, HdrHeap *heap, HTTPHdrImpl *hh, const
     }
     url_end += 1;
 
+    // Make sure we didn't pickup odd characters in the url between the method and version
+    if (!url_is_compliant(url_start, url_end)) {
```
maskit, line 1092: The same check will be done later when `url_parse` is called, if `strict_uri_parsing` is `true`. In other words, there might be a case we want to accept an invalid URL? Adding this check here effectively makes the flag void.

shinrich:
@maskit that is a very good point.  I have adjusted the PR to only change the default value of proxy.config.http.strict_uri_parsing from 0 to 1.  Probably having a default of 0 (disabled) made sense 15 years ago when many more clients played very loose with the spec, but these days good clients should be within spec and we should be in the more secure stance out of the box.

shinrich:
Hmm, looks like a fair number of our clients like to use non-RFC compliant characters in their requests.  It seems like having a middle ground check for white space and control characters would be useful.  I'm experimenting with that.

shinrich:
Will be opening a different PR which adds an intermediate option to the strict url parsing setting.

PR #7897, opened 2021-05-27
Adds a new peering ring mode to next hop selection strategies.

jrushford:
Adds a new peering consistent hash ring mode to next hop strategies.  This lets you build an edge group
of caches that share their cache.  When a request comes to a cache, that caches peers are checked using
consistent hash for the requested object.  If the peer does not have the object in it's cache, it will request the
object from an upstream parent or origin.  In the case where a request caches to the host itself, the request is
forwarded to the upstream parent or origin.  A new configuration parameter to the next hop strategies.yaml.
allows you to disable caching when a requested object is fetched from a peer in the cache group or pod.  This
allows for cache sharing among hosts at the edge.

### DIFF
--- a/configs/strategies.yaml.default
+++ b/configs/strategies.yaml.default
@@ -95,15 +95,24 @@
 #     hash_key: hostname # optional key to use for Hashing. Enum of 'url' or 'uri' or 'hostname' or 'path' or 'path+query' or 'cache_key' or 'path+fragment'
 #     go_direct: true # transactions may routed directly to the origin true/false default is true.
 #     parent_is_proxy: false  # next hop hosts  are origin servers when set to 'false', defaults to true and indicates next hop hosts are ats cache's.
+#     cache_peer_result: true # only used when the 'ring_mode' is set to 'peering_ring' and the policy is 'consistent_hash'.  The default value is
+#                        true which means to always allow caching responses.  When set to 'false' reponses received from peer hosts will not be
+#                        cached, only responses received from upstream parents or origins will be cached.
 #     groups: # groups of hosts, these groups are used as rings in consistent hash and arrays of host groups for round_robin.
 #       - *g1
 #       - *g2
 #     scheme: http
 #     failover:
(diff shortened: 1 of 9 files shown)


configs/strategies.yaml.default
```diff
@@ -95,15 +95,24 @@
...
 #     scheme: http
 #     failover:
 #       max_simple_retries: 2 # default is 1, indicates the maximum number of simple retries for the listed response codes.
-#       ring_mode: exhaust_ring # enumerated as exhaust_ring or alternate_ring
+#       ring_mode: exhaust_ring # enumerated as exhaust_ring, alternate_ring, or peering_ring
 #           #1) in 'exhaust_ring' mode all the servers in a ring are exhausted before failing over to secondary ring
 #           #2) in 'alternate_ring' mode causes the failover to another server in secondary ring.
+#           #3) 'peering_ring' is implemented for only a policy of 'consistent_hash' and requires that the strategy
+#               has two host groups defined.  The first group is the 'peer' group of caches that also includes this host
+#               itself.  The second group is the 'upstream' group of caches.  All parent host lookups are looked up from
```
ywkaras, line 112: Isn't it possible that the upstream group uses a VIP? In that case would the upstream group just have one member, the VIP hostname?

jrushford: sure that's possible but, that vip should be added to the 2nd group and it's fine if it only has one member.  Are you asking for a change in the documentation @ywkaras ?

jrushford:
[approve ci autest]

ywkaras (approved):

zwoop:
I'm going to move this to 9.2.x. Please make an argument for why we need this into 9.1.x, this late in the game :-).

jrushford:
@zwoop It's not that big of a change but, it does add the peering capability we talked about.  I think it would be a big plus to get it into 9.1.   It is a feature that we wish to use at Comcast and Verizon wishes to use it as well.  Walt has written an autest specifically to test this feature.  However, it cleanly cherry picks to 9.1 so, we can always just cherry pick it to our internal build.

PR #7904, opened 2021-05-28
Ensure the HTTP portion of the protocol string is upper case

shinrich:
My reading of https://datatracker.ietf.org/doc/html/rfc2616#section-3.1 seems to imply that the HTTP portion of the protocol string should always be upper case, so we can streamline our protocol string parsing.

### DIFF
--- a/proxy/hdrs/HTTP.cc
+++ b/proxy/hdrs/HTTP.cc
@@ -1056,19 +1056,19 @@ http_parser_parse_req(HTTPParser *parser, HdrHeap *heap, HTTPHdrImpl *hh, const 
     }
     goto parse_url;
   parse_version4:
-    if ((*cur != 'P') && (*cur != 'p')) {
+    if (*cur != 'P') {
       goto parse_url;
     }
     GETPREV(parse_url);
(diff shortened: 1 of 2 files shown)


shinrich:
[approve ci autest]

masaori335:
RFC 7230 obsoletes RFC 2616. It clarified HTTP-version is case-sensitive.

[RFC 7230 A.2.  Changes from RFC 2616](https://datatracker.ietf.org/doc/html/rfc7230#appendix-A.2)
> The HTTP-version ABNF production has been clarified to be case-
> sensitive.  Additionally, version numbers have been restricted to
> single digits, due to the fact that implementations are known to
> handle multi-digit version numbers incorrectly.  (Section 2.6)

[RFC 7230 Section 2.6](https://datatracker.ietf.org/doc/html/rfc7230#section-2.6)
> HTTP-version is case-sensitive.
>
>     HTTP-version  = HTTP-name "/" DIGIT "." DIGIT
>     HTTP-name     = %x48.54.54.50 ; "HTTP", case-sensitive

masaori335 (approved):

zwoop:

zwoop:

Issue #7880, opened 2021-05-26
Trafficserver holds chunked response for backend down case


lukenowak:
We have a situation where the backend returns a response with body size > 2048 bytes without Content-Lenght. This leads to TrafficServer returning it further with Transfer-Encoding: chunked. If the body size is <= 2048 TrafficServer adds Content-Length by itself and serves the content happily.

We are using TrafficServer 8.1.1, but the same behaviour is with 9.0.1.

Our configuration:

- proxy.config.http.negative_revalidating_enabled 1
- proxy.config.http.cache.max_stale_age 604800
- proxy.config.http.negative_revalidating_lifetime 0
- proxy.config.http.chunking_enabled 1 (default)
- proxy.config.http.chunking.size 4096 (default)
- proxy.config.http.server_ports 10.0.117.32:23432
- proxy.config.http.keep_alive_no_activity_timeout_in 10 (this is for the test environment, but higher values in live environment just extend the case)

If the backend is up (returns 200) and negative_revalidating_enabled does not kick in, TrafficServer returns the full response (headers and body) immediately.

The backend up case communication between client and TrafficServer:

![backend-up-case](https://user-images.githubusercontent.com/334019/119633233-d1e37b80-be11-11eb-8923-5c89dcf4f267.png)

So we happily support the case when the backend is down, but after returning headers it is giving part of the body as chunked, but does not finish it with \r\n0\r\n, which leads to the client being “stuck” for the time of proxy.config.http.keep_alive_no_activity_timeout_in, when then the TrafficServer sends FIN/ACK to the client, somehow releasing it. Seems that [RFC](https://datatracker.ietf.org/doc/html/rfc7230#section-4.1) is not followed.

The backend down case communication between client and TrafficServer:

![backend-down-case](https://user-images.githubusercontent.com/334019/119633352-f4759480-be11-11eb-84e4-4a2b81ca3ead.png)

Setting proxy.config.http.chunking_enabled to 0 fixes the issue, but we can't accept it as a workaround, as we rely on chunked encoding for many cases, especially for backend-up ones.

bneradt:
I'm working on reproducing this but haven't been able to yet. We have a negative-revalidating test which I've modified like so:

```$ git diff
diff --git a/tests/gold_tests/cache/negative-revalidating.test.py b/tests/gold_tests/cache/negative-revalidating.test.py
index 1f4975206..901e65f95 100644
--- a/tests/gold_tests/cache/negative-revalidating.test.py
[...]

PR #7907, opened 2021-05-28
Add a chunked negative revalidating test.

bneradt:
This adds a negative revalidating test for a chunk encoded response.
This functions as a regression test for #7880 

It also happens to verify the fix provided in the following PR: https://github.com/apache/trafficserver/pull/7577


bneradt:
Requesting this for the 9.1.x since that also has the fix this test verifies and it should cherry-pick cleanly there.

bneradt:
I've also set this for the 9.0.x target because, talking with @bryancall, we feel the fix for the associated PR https://github.com/apache/trafficserver/pull/7577 should go into 9.0.x. Since that fix will go into 9.0.x, we should also merge in this associated test being added for that fix. If we decide to not merge in PR 7577 to 9.0.x, then this 7907 PR should also not be merged into 9.0.x.

ezelkow1 (approved):

zwoop:

PR #7913, opened 2021-06-02
Ensure TS_VCONN_CLOSE_HOOK hook is called during TS_EVENT_VCONN_CLOSE.

bneradt:
This causes the VCONN_CLOSE hooks to be called more consistently in the
event of TS_EVENT_VCONN_CLOSE.  Without this patch it got missed under
some circumstances, causing situations where plugins would receive a
TS_EVENT_VCONN_START but never a TS_EVENT_VCONN_CLOSE for some
connections.

### DIFF
--- a/iocore/net/SSLNetVConnection.cc
+++ b/iocore/net/SSLNetVConnection.cc
@@ -1694,39 +1694,40 @@ SSLNetVConnection::callHooks(TSEvent eventId)
   Debug("ssl", "sslHandshakeHookState=%d eventID=%d", this->sslHandshakeHookState, eventId);
 
   // Move state if it is appropriate
-  switch (this->sslHandshakeHookState) {
-  case HANDSHAKE_HOOKS_PRE:
-  case HANDSHAKE_HOOKS_OUTBOUND_PRE:
-    if (eventId == TS_EVENT_SSL_CLIENT_HELLO) {
-      this->sslHandshakeHookState = HANDSHAKE_HOOKS_CLIENT_HELLO;
-    } else if (eventId == TS_EVENT_SSL_SERVERNAME) {
-      this->sslHandshakeHookState = HANDSHAKE_HOOKS_SNI;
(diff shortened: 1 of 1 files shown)


bneradt:
[approve ci autest]

iocore/net/SSLNetVConnection.cc
```diff
@@ -1694,6 +1694,12 @@ SSLNetVConnection::callHooks(TSEvent eventId)
   Debug("ssl", "sslHandshakeHookState=%d eventID=%d", this->sslHandshakeHookState, eventId);
 
   // Move state if it is appropriate
+  if (eventId == TS_EVENT_VCONN_CLOSE) {
+    // Regardless of state, if the connection is closing, then transition to
+    // the DONE state. This will trigger us to call the appropriate cleanup
+    // routines.
+    this->sslHandshakeHookState = HANDSHAKE_HOOKS_DONE;
+  }
```
maskit, line 1702: Looks reasonable. We may want to put the `switch` statement below into `else` clause to ensure `sslHandshakeHookState` won't be accidentally overwritten in the `switch` statement.

bneradt: Good idea.

iocore/net/SSLNetVConnection.cc
```diff
@@ -1694,39 +1694,40 @@ SSLNetVConnection::callHooks(TSEvent eventId)
...
+        this->sslHandshakeHookState = HANDSHAKE_HOOKS_CERT;
+      }
+      break;
+    default:
+      break;
     }
-    break;
-  default:
-    break;
   }
```
bneradt, line 1731: View this patch using the "Hide whitespace changes" option via github.

maskit (approved):

bneradt:
Marking this for the 9.1.x train. If plugins rely upon the VCONN_CLOSE hook to free memory, this will result in a memory leak. Or any host of other unexpected behaviors when the hook isn't called when expected.

zwoop:

PR #7919, opened 2021-06-03
Adds TS API TSHttpTxnCacheDiskPathGet

elsloo:
This is a backport of the API added by PRs #7783 and #7806.

PR #7784, opened 2021-05-04
Adds new X-Cache-Info header to the xdebug plugin

elsloo:
* Adds new X-Cache-Info header to the xdebug plugin
* Injects the `X-Cache-Info` header with value of `path=/path/to/disk/from/storage/config; volume=n` where `n` is the volume number
* Fixes a small bug inside `HttpCacheSM.h` in `get_volume_number()` that leads to `-1` being returned on cache miss
  * NOTE: this will change behavior of any plugin that relies on `TSHttpTxnInfoIntGet` with a key of `TS_TXN_INFO_CACHE_VOLUME`
* NOTE: This PR depends on #7783

PR #7783, opened 2021-05-04
Adds new TS API TSHttpTxnCacheDiskPathGet

elsloo:
* Adds new TS API TSHttpTxnCacheDiskPathGet to obtain the path of the disk that contains the requested object

PR #7920, opened 2021-06-03
Adds new X-Cache-Info header to the xdebug plugin (#7784)

elsloo:
This is a backport of changes made to the xdebug plugin via PR #7784 and it depends on backport PR #7919.

(cherry picked from commit 7433038df97229a056775de24008c2c7ef7b608f)

### DIFF
--- a/plugins/xdebug/xdebug.cc
+++ b/plugins/xdebug/xdebug.cc
@@ -31,6 +31,7 @@
 #include <unistd.h>
 
 #include <ts/ts.h>
+#include "ts/experimental.h"
 #include "tscore/ink_defs.h"
 #include "tscpp/util/PostScript.h"
 #include "tscpp/util/TextView.h"
@@ -81,6 +82,7 @@ enum {
   XHEADER_X_REMAP          = 1u << 8,
   XHEADER_X_PROBE_HEADERS  = 1u << 9,
   XHEADER_X_PSELECT_KEY    = 1u << 10,
+  XHEADER_X_CACHE_INFO     = 1u << 11,
 };
 
 static TSCont XInjectHeadersCont  = nullptr;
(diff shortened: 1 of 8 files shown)


rob05c:
I see this is a backport, I missed the original, sorry. But, the IETF has asked people not to use the `X-` prefix anymore. 

Mostly because things tend to become standards, and then it's hard to remove the `X-`.

https://datatracker.ietf.org/doc/html/rfc6648
> Deprecating the "X-" Prefix and Similar Constructs
>                        in Application Protocols

Is it too late to change this to just `Cache-Info`?

elsloo:
> I see this is a backport, I missed the original, sorry. But, the IETF has asked people not to use the `X-` prefix anymore.
> 
> Mostly because things tend to become standards, and then it's hard to remove the `X-`.
> 
> https://datatracker.ietf.org/doc/html/rfc6648
> 
> > Deprecating the "X-" Prefix and Similar Constructs
> > in Application Protocols

Yes, I'm aware of this and chose consistency over following the best practice you referenced. `X` is used throughout the plugin, the headers, and its name. Having a single header without the prefix would lead to a less intuitive experience when using this plugin.

> Is it too late to change this to just `Cache-Info`?

IMO, yes, this is a backport as you noted above.

ezelkow1:
I am +1 (once we get #7917 in so this builds). This is a debug plugin, its not adding actionable headers to the core but in an optional debug plugin that is also easily changed if somehow any of these become standards. Plus its a backport anyway and it already exists in the code base for future versions

ywkaras:
It looks like https://github.com/apache/trafficserver/pull/7783 will have to be backported to 9.1 before this?

ezelkow1:
> 
> 
> It looks like #7783 will have to be backported to 9.1 before this?

yea that backport is in 7919, thats why it fails CI for now

elsloo:
> > It looks like #7783 will have to be backported to 9.1 before this?
> 
> yea that backport is in 7919, thats why it fails CI for now

Correct, however none of it really matters because @zwoop will likely cherry-pick these manually into 9.1.x, at which point, both of the PRs will be closed.

PR #7921, opened 2021-06-03
Fix race condition on server session state

shinrich:
Fixes issue introduced in PR #7849.  Found while running those changes against our 9.1.x on a production box this afternoon. Two times we triggered the assert in Http1ServerSesssion::release().  The state was set to KA_POOLED.  The state was getting set to KA_POOLED both from the HttpSessionManager and within Http1ServerSession::release_transaction after calling httpSessionManager.release_session().

We are running with global pool. I think the session had been reassigned to another thread (the one that ultimately asserted) when the original thread set the state variable again.  We should not be touching the server session again after releasing it in any case, and the change is unnecessary since it has already been set by the HttpSessionManager.

### DIFF
--- a/proxy/http/Http1ServerSession.cc
+++ b/proxy/http/Http1ServerSession.cc
@@ -228,8 +228,6 @@ Http1ServerSession ::release_transaction()
       // The session was successfully put into the session
       //    manager and it will manage it
       // (Note: should never get HSM_NOT_FOUND here)
-      // Set our state to KA for stat issues
-      state = KA_POOLED;
       ink_assert(r == HSM_DONE);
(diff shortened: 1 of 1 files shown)


shinrich:
[approve ci autest]

shinrich:
Ran this fix on our production box without issue overnight.

shinrich:
[approve ci autest]

maskit (approved):
Looks reasonable.

This is a common thing on ATS codebase, but it's unclear which class is responsible for changing `state`. I guess we could remove the other instead, but I'm inclined to changing state on HttpSessionManager side.

PR #7923, opened 2021-06-04
Adjust vc read errors

### LINKED ISSUE #7915: h2spec.test.py has become flakey
I've noticed that CI has started having intermittent failures with the h2spec.test.py test. I reproduced it locally, running the test 20 times, and noticing that it would fail consistently within a few runs. I did a git bisect and here's the results:

```
$ git bisect run /var/tmp/build_and_run_h2spec_test.sh
running /var/tmp/build_and_run_h2spec_test.sh
building: f80ed7354 (HEAD) Do not delete the continuation twice (#7862)
    tests failed for: f80ed7354 (HEAD) Do not delete the continuation twice (#7862)
Bisecting: 5 revisions left to test after this (roughly 3 steps)
[a0c267cdc88586679ccba4ec3e394f386d52f806] Add URI Signing cdnistd Claim Implementation (#7822)
running /var/tmp/build_and_run_h2spec_test.sh
building: a0c267cdc (HEAD) Add URI Signing cdnistd Claim Implementation (#7822)
    tests failed for: a0c267cdc (HEAD) Add URI Signing cdnistd Claim Implementation (#7822)
Bisecting: 2 revisions left to test after this (roughly 2 steps)
[19dfdd4753232d0b77ca555f7ef5f5ba3d2ccae1] Clean up lua plugin doc for overridable configurations (#7844)
running /var/tmp/build_and_run_h2spec_test.sh
building: 19dfdd475 (HEAD) Clean up lua plugin doc for overridable configurations (#7844)
    tests failed for: 19dfdd475 (HEAD) Clean up lua plugin doc for overridable configurations (#7844)
Bisecting: 0 revisions left to test after this (roughly 1 step)
[23174f43ba543c9f56916e5d6aadb7a22f349355] Save and propagate epoll network error (#7809)
running /var/tmp/build_and_run_h2spec_test.sh
building: 23174f43b (HEAD) Save and propagate epoll network error (#7809)
    tests failed for: 23174f43b (HEAD) Save and propagate epoll network error (#7809)
Bisecting: 0 revisions left to test after this (roughly 0 steps)
[fbdbb5b733c389e537d777938298219be43e619e] Add method to write an IpAddr value to a sockaddr. (#7821)
running /var/tmp/build_and_run_h2spec_test.sh
building: fbdbb5b73 (HEAD) Add method to write an IpAddr value to a sockaddr. (#7821)
    tests passed for: fbdbb5b73 (HEAD) Add method to write an IpAddr value to a sockaddr. (#7821)
23174f43ba543c9f56916e5d6aadb7a22f349355 is the first bad commit
commit 23174f43ba543c9f56916e5d6aadb7a22f349355
Author: Susan Hinrichs <shinrich@verizonmedia.com>
Date:   Mon May 17 19:23:48 2021 -0500

    Save and propagate epoll network error (#7809)

 iocore/net/NetEvent.h            | 17 +++++++++++++++++
 iocore/net/UnixNet.cc            | 19 ++++++++-----------
 iocore/net/UnixNetVConnection.cc | 12 ++++++++++++
 3 files changed, 37 insertions(+), 11 deletions(-)
bisect run success

```

Those results seem to indicate that this PR has made the h2spec test to pass inconsistently:
https://github.com/apache/trafficserver/pull/7809

For those trying to reproduce these results:

* Docker image: `controller.trafficserver.org/ats/centos:8`
* `./configure --prefix /var/tmp/build --with-openssl=/opt/openssl-quic --enable-experimental-plugins --enable-example-plugins --enable-werror --enable-debug --enable-wccp`
* `./autest.sh --sandbox /tmp/sb_master --ats-bin /var/tmp/build/bin/ --clean=none  -f h2spec`

shinrich:
Changing how we handle the read error.  I think the h2spec test case is failing with fair frequency because the data delivery (to read) and the epoll error are notified in the same epoll interval.  Normally this wouldn't matter much since the connection has closed in any case, so actually processing the data from the client wouldn't really matter.  But in the failing h2spec test case, we are verifying that ATS was able to process the data frame. 

This PR attempts to perform the read first always (making it like the the code pre #7809).  Assuming the socket had failed with a EPIPE (which is what I was seeing), the reads will succeed until all the data is read from the TCP buffer then the read will.  If for some reason the read did not fail,  this change then sends along the read error event.

This closes #7915

### DIFF
--- a/iocore/net/UnixNetVConnection.cc
+++ b/iocore/net/UnixNetVConnection.cc
@@ -199,12 +199,6 @@ read_from_net(NetHandler *nh, UnixNetVConnection *vc, EThread *thread)
     return;
   }
 
-  if (vc->has_error()) {
-    vc->lerrno = vc->error;
-    vc->readSignalAndUpdate(VC_EVENT_ERROR);
-    return;
-  }
-
   // It is possible that the closed flag got set from HttpSessionManager in the
   // global session pool case.  If so, the closed flag should be stable once we get the
   // s->vio.mutex (the global session pool mutex).
@@ -316,6 +310,7 @@ read_from_net(NetHandler *nh, UnixNetVConnection *vc, EThread *thread)
   } else {
     r = 0;
   }
+  ink_release_assert(!vc->has_error());
 
   // Signal read ready, check if user is not done
   if (r) {
--- a/iocore/net/UnixNetVConnection.cc
+++ b/iocore/net/UnixNetVConnection.cc
@@ -310,7 +310,6 @@ read_from_net(NetHandler *nh, UnixNetVConnection *vc, EThread *thread)
   } else {
     r = 0;
   }
-  ink_release_assert(!vc->has_error());
 
   // Signal read ready, check if user is not done
   if (r) {
@@ -332,6 +331,7 @@ read_from_net(NetHandler *nh, UnixNetVConnection *vc, EThread *thread)
       }
     }
   }
+  ink_release_assert(!vc->has_error());
   // If here are is no more room, or nothing to do, disable the connection
   if (s->vio.ntodo() <= 0 || !s->enabled || !buf.writer()->write_avail()) {
     read_disable(nh, vc);
--- a/iocore/net/UnixNetVConnection.cc
+++ b/iocore/net/UnixNetVConnection.cc
@@ -330,8 +330,12 @@ read_from_net(NetHandler *nh, UnixNetVConnection *vc, EThread *thread)
         return;
       }
     }
+  } else if (vc->has_error()) {
+    vc->lerrno = vc->error;
+    read_signal_and_update(VC_EVENT_ERROR, vc);
+    return;
   }
-  ink_release_assert(!vc->has_error());
+
   // If here are is no more room, or nothing to do, disable the connection
   if (s->vio.ntodo() <= 0 || !s->enabled || !buf.writer()->write_avail()) {
     read_disable(nh, vc);
--- a/iocore/net/UnixNetVConnection.cc
+++ b/iocore/net/UnixNetVConnection.cc
@@ -330,10 +330,6 @@ read_from_net(NetHandler *nh, UnixNetVConnection *vc, EThread *thread)
         return;
       }
     }
-  } else if (vc->has_error()) {
-    vc->lerrno = vc->error;
-    read_signal_and_update(VC_EVENT_ERROR, vc);
-    return;
   }
 
   // If here are is no more room, or nothing to do, disable the connection


shinrich:
[approve ci autest]

shinrich:
Running autest a few times to make sure this change isn't just a fluke.

shinrich:
[approve ci autest]

shinrich:
tls_client_cert_override failed that time.  Looks like traffic_manager was not able to bring up traffic_server.

shinrich:
[approve ci autest]

shinrich:
[approve ci autest]

shinrich:
[approve ci autest]

bneradt:
This looks good. Locally I did the following to verify this patch:

1. Created a new environment using the docker image described in the #7915 and built trafficserver with latest master.
2. Reproduced the problem by re-running h2spec.test.py in a loop. On the fifth AuTest run, the h2spec.test.py test failed.
3. I then cherry-picked the patch from this PR into the local trafficserver repo and rebuilt trafficserver.
4. I then re-ran the h2spec.test.py test in a loop.

After 25 iterations it did not fail.

This patch seems to make the test run much more consistently.

iocore/net/UnixNetVConnection.cc
```diff
@@ -199,12 +199,6 @@ read_from_net(NetHandler *nh, UnixNetVConnection *vc, EThread *thread)
     return;
   }
 
-  if (vc->has_error()) {
-    vc->lerrno = vc->error;
-    vc->readSignalAndUpdate(VC_EVENT_ERROR);
```
masaori335, line 204: I'm +1 on getting rid of this block. It's making the problem.

iocore/net/UnixNetVConnection.cc
```diff
@@ -336,7 +330,12 @@ read_from_net(NetHandler *nh, UnixNetVConnection *vc, EThread *thread)
         return;
       }
     }
+  } else if (vc->has_error()) {
```
masaori335, line 333: `read_from_net()` is a pretty long method and I'm confused. We come here only if the `toread` is `0`?

masaori335: I concern this raise `VC_EVENT_ERROR` in unexpected timing like
- the reading buffer is full (`toread == 0`)
- the peer sent all data and reset the connection

shinrich: Hmm.  The intent was to issue the read error if the reading had not otherwise hit an error, but that is not what we're doing here.  You are correct.

Probably best to remove the error sending on the read side, but leave it on the write side (from the original) PR.  I'll updated.

masaori335 (approved):
Looks reasonable.

zwoop:

PR #7924, opened 2021-06-07
Ensure that URL components are valid when alternate eviction is logged

bneradt:
When alternate eviction logging takes place, we retrieve the URL. Under
some circumstances the internal members of the request were invalid.
This change sets a flag on the request to ensure the members are valid
when the URL is printed.

The code this PR is patching was introduced in #7629


bneradt:
[approve ci autest]

masaori335 (approved):
Looks good

PR #7935, opened 2021-06-10
Remove scrap log object dead code

### LINKED ISSUE #7933: We should delete global_scrap_object code
While doing some logging work, I noticed that our code instantiates a so-called scrap LogObject:

https://github.com/apache/trafficserver/blob/b85d9fd8481bc55547aade92f5b9cf7a8db5099f/proxy/logging/Log.cc#L1091=#L1095

This object is never written to, thus it is dead code. Digging through our history, it looks like this used to be associated with log collation, a feature which was removed in #5353. Since that no longer exists, we can simply remove the scrap log object from our code.

bneradt:
The scrap log object was a part of log collation, which was removed in
commit:
82d564486e058ead160bfb8ad15e40b0507ff8aa

This object is therefore no longer needed and is not used. This commit
removes it.

---
This fixes #7933

### DIFF
--- a/proxy/logging/Log.cc
+++ b/proxy/logging/Log.cc
@@ -58,9 +58,7 @@
 // Log global objects
 inkcoreapi LogObject *Log::error_log = nullptr;
 LogFieldList Log::global_field_list;
-LogFormat *Log::global_scrap_format = nullptr;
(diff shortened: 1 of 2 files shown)


randall (approved):

zwoop:

PR #7939, opened 2021-06-11
regex_revalidate: optionally retain rule epoch state across restarts

traeak:
With this --state-file/-f <filename> option defined the regex_revalidate plugin will write out the current loaded ruleset with included epoch rule epoch time.

When ats is restarted the regex_revalidate plugin will load the regex_revalidate.config file, load this state file, update the epoch time of any matching rule, then immediately write out a new state file.

Every time the regex_revalidate.config file is reloaded a new state file will be generated.

This has been added to counteract any ill affects of using the new regex_revalidate MISS type (marking an asset as MISS instead of the default STALE).

### DIFF
--- a/plugins/regex_revalidate/regex_revalidate.c
+++ b/plugins/regex_revalidate/regex_revalidate.c
@@ -38,12 +38,14 @@
 #include <pcre.h>
 #endif
 
-#define LOG_PREFIX "regex_revalidate"
 #define CONFIG_TMOUT 60000
 #define FREE_TMOUT 300000
 #define OVECTOR_SIZE 30
 #define LOG_ROLL_INTERVAL 86400
 #define LOG_ROLL_OFFSET 0
+
+static const char *const PLUGIN_NAME = "regex_revalidate";
+static const char *const DEFAULT_DIR = "var/trafficserver"; /* Not perfect, but no better API) */
 
 static char const *const RESULT_MISS    = "MISS";
 static char const *const RESULT_STALE   = "STALE";
@@ -77,9 +79,10 @@ typedef struct invalidate_t {
 
 typedef struct {
   invalidate_t *invalidate_list;
-  char *config_file;
+  char *config_path;
   time_t last_load;
   TSTextLogObject log;
+  char *state_path;
 } plugin_state_t;
 
 static invalidate_t *
@@ -115,21 +118,23 @@ free_invalidate_t(invalidate_t *i)
 }
 
 static void
-free_invalidate_t_list(invalidate_t *i)
+free_invalidate_t_list(invalidate_t *iptr)
 {
-  if (i->next) {
-    free_invalidate_t_list(i->next);
+  while (NULL != iptr) {
+    invalidate_t *const next = iptr->next;
+    free_invalidate_t(iptr);
+    iptr = next;
   }
-  free_invalidate_t(i);
 }
 
 static plugin_state_t *
 init_plugin_state_t(plugin_state_t *pstate)
 {
   pstate->invalidate_list = NULL;
-  pstate->config_file     = NULL;
+  pstate->config_path     = NULL;
   pstate->last_load       = 0;
   pstate->log             = NULL;
+  pstate->state_path      = NULL;
   return pstate;
 }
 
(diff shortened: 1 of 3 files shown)


traeak:
[approve ci freebsd]

plugins/regex_revalidate/regex_revalidate.c
```diff
@@ -77,9 +79,10 @@ typedef struct invalidate_t {
 
 typedef struct {
   invalidate_t *invalidate_list;
-  char *config_file;
+  char *config_path;
```
ywkaras, line 82: Weren't we going to convert this plugin to C++?  Would be easier to maintain if this was a std::string.

plugins/regex_revalidate/regex_revalidate.c
```diff
@@ -115,21 +118,23 @@ free_invalidate_t(invalidate_t *i)
...
 static void
-free_invalidate_t_list(invalidate_t *i)
+free_invalidate_t_list(invalidate_t *iptr)
 {
-  if (i->next) {
-    free_invalidate_t_list(i->next);
+  while (NULL != iptr) {
+    invalidate_t *const next = iptr->next;
+    free_invalidate_t(iptr);
+    iptr = next;
```
ywkaras, line 126: Looks like this data structure could be an std::list.

plugins/regex_revalidate/regex_revalidate.c
```diff
@@ -221,6 +229,96 @@ prune_config(invalidate_t **i)
...
+            iptr->epoch = inv->epoch;
+          }
+          break;
+        }
+
+        if (NULL == iptr->next) {
+          break;
+        }
+        iptr = iptr->next;
+      }
```
ywkaras, line 309: Why not just:
```
do {
  // ...
  iptr = iptr->next;
} while (iptr);
```

plugins/regex_revalidate/regex_revalidate.c
```diff
@@ -221,6 +229,96 @@ prune_config(invalidate_t **i)
...
+
+      free_invalidate_t(inv);
```
ywkaras, line 311: Why cant `inv` just be a stack variable?

tests/gold_tests/pluginTest/regex_revalidate/regex_revalidate_state.test.py
```diff
@@ -0,0 +1,132 @@
...
+Test.ContinueOnFail = False
+
+# configure origin server
+server = Test.MakeOriginServer("server")
+
+# Define ATS and configure
+ts = Test.MakeATSProcess("ts", command="traffic_manager")
+
+# **testname is required**
+#testName = "regex_reval"
```
ywkaras, line 42: You can't delete these comments?

tests/gold_tests/pluginTest/regex_revalidate/regex_revalidate_state.test.py
```diff
@@ -0,0 +1,132 @@
...
+regex_revalidate plugin test, MISS (refetch) functionality
+'''
+
+# Test description:
+# If MISS tag encountered, should load rule as refetch instead of IMS.
+# If rule switched from MISS to IMS or vice versa, rule should reset.
+
+Test.SkipUnless(
+    Condition.PluginExists('regex_revalidate.so'),
+    Condition.PluginExists('xdebug.so')
```
ywkaras, line 31: Why does xdebug have to exist?

Non-experimental plugins should always be present so I don't think checking for regex_revalidate.so is good, test should simply be allowed to fail if it's not presesnt.

plugins/regex_revalidate/regex_revalidate.c
traeak: becaise free_invalidate_t free's the variable as well.  every single free_invalidate_t(i) would request TSfree(i) after it.

plugins/regex_revalidate/regex_revalidate.c
traeak: perhaps.  Although std::list, linked lists are the worst of the worst for performance.  std::vector pretty much wins out every time, mostly due to memory fragmentation and modern cpu caches.

tests/gold_tests/pluginTest/regex_revalidate/regex_revalidate_state.test.py
traeak: xdebug not needed for this test.

tests/gold_tests/pluginTest/regex_revalidate/regex_revalidate_state.test.py
traeak: resolved, comments at the top of the test updated as well.

plugins/regex_revalidate/regex_revalidate.c
traeak: perhaps in another pr.

ywkaras (approved):

PR #7944, opened 2021-06-15
Compilation error fixes for QUIC unit tests

bneradt:
When configured with a QUIC version of OpenSSL, `make check` runs the
QUIC unit tests. These have fallen under disrepair, with recent changes
causing the build of them to fail. This addresses those compilation
issues. All the `check` make target tests build and pass with this
change.

select_next_protocol got moved from QUICConnection in commit:
cb247c6af699cb08286f04bd2ef481031b5babc7

HTTPHdr received an iterate change in:
98176750ecb42ab5ec97a49570c5aff1175a014a

### DIFF
--- a/iocore/net/quic/Mock.h
+++ b/iocore/net/quic/Mock.h
@@ -194,13 +194,6 @@ class MockQUICConnectionInfoProvider : public QUICConnectionInfoProvider
     return NET_VCONNECTION_OUT;
   }
 
-  int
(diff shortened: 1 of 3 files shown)


bryancall:
I tested the patch on Fedora 34

maskit (approved):
The patch works on macOS + BoringSSL as well.

PR #7949, opened 2021-06-17
Cleanup: Constify MIMEHdr

masaori335:
With a little help from running `readability-make-member-function-const` of clang-tidy.

https://releases.llvm.org/12.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/readability-make-member-function-const.html

### DIFF
--- a/proxy/hdrs/MIME.h
+++ b/proxy/hdrs/MIME.h
@@ -115,7 +115,7 @@ struct MIMEField {
   }
 
   bool
-  is_cooked()
+  is_cooked() const
   {
     return (m_flags & MIME_FIELD_SLOT_FLAGS_COOKED) ? true : false;
   }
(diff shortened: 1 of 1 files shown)


bneradt (approved):

proxy/hdrs/MIME.h
```diff
@@ -1790,7 +1790,7 @@ MIMEHdr::get_cooked_pragma_no_cache()
   -------------------------------------------------------------------------*/
 
 inline void
-MIMEHdr::set_cooked_cc_need_revalidate_once()
+MIMEHdr::set_cooked_cc_need_revalidate_once() // NOLINT(readability-make-member-function-const)
```
masaori335, line 1793: @bryancall This is a good example. The `readability-make-member-function-const` suggests make this const function, but this is changing the `m_mask`.
If we run `clang-tidy` with `readability-make-member-function-const`, we need to check all suggestions. :(

Issue #7946, opened 2021-06-16
Intermittent failure on AuTest (double)


masaori335:
AuTest (double) is smoking. I saw some failures with unrelated changes.

https://ci.trafficserver.apache.org/view/github/job/autest-github/15874/console
https://ci.trafficserver.apache.org/view/github/job/autest-github/15876/console
https://ci.trafficserver.apache.org/view/github/job/autest-github/15889/console
```
 Test: double: Failed
    File: double.test.py
    Directory: /var/jenkins/workspace/autest-github/src/tests/gold_tests/continuations
   Starting Test double : No issues found - Passed
      Reason: Started!
   Process: ts: Passed
     Setting up : Making Directory /var/tmp/ausb-7765.15876/double/ts - Passed
     Setting up : Copying '/var/jenkins/workspace/autest-github/15876/install/bin' to '/var/tmp/ausb-7765.15876/double/ts/bin' - Passed
     Setting up : Making Directory /var/tmp/ausb-7765.15876/double/ts/config - Passed
     Setting up : On /var/tmp/ausb-7765.15876/double/ts/config with uid: nobody gid: nobody - Passed
     Setting up : On /var/tmp/ausb-7765.15876/double/ts/config with uid: nobody gid: nogroup - Passed
     Setting up : Copying '/var/jenkins/workspace/autest-github/src/tests/gold_tests/autest-site/min_cfg/ip_allow.yaml' to directory '/var/tmp/ausb-7765.15876/double/ts/config' as 'ip_allow.yaml'' - Passed
     Setting up : Copying '/var/jenkins/workspace/autest-github/src/tests/gold_tests/autest-site/min_cfg/readme.txt' to directory '/var/tmp/ausb-7765.15876/double/ts/config' as 'readme.txt'' - Passed
     Setting up : Copying '/var/jenkins/workspace/autest-github/src/tests/gold_tests/autest-site/min_cfg/storage.config' to directory '/var/tmp/ausb-7765.15876/double/ts/config' as 'storage.config'' - Passed
     Setting up : Copying '/var/jenkins/workspace/autest-github/15876/install/etc/trafficserver/body_factory' to '/var/tmp/ausb-7765.15876/double/ts/config/body_factory' - Passed
     Setting up : Making Directory /var/tmp/ausb-7765.15876/double/ts/cache - Passed
[...]

PR #7955, opened 2021-06-18
Temporarily skip the double AuTest to unblock PRs

bneradt:
There is an issue opened to track the intermittent failures in the
double.test.py, see #7946.

In the meantime, concurrently with that being actively investigated,
this PR skips the double.test.py. This is a temporary workaround to
remove the friction this test is adding to our PR process. This patch
should not remain in place for more than a few days.


bryancall (approved):

bneradt:
Closing this since Susan has a fix that addresses the core problem already: #7956. With that fix in, we won't need to temporarily skip the test.

PR #7962, opened 2021-06-21
Pass through expect header and handle 100-continue response

### LINKED ISSUE #7346: proxy.config.http.send_100_continue_response default should be changed to 1
I'm opening a new issue rather than reopening effectively the same older issue #2543.  This original issue was closed deciding that it would be more standards compliant for the origin to product the 100-continue response.  And in general I agree.  However, with this setting set to 0, the exchange between the user agent and ATS is not standards compliant if the Expect header is set to 100-continue.

In the current default case, the following steps occur.
1. Client sends POST request with Expect header set to 100-continue
2. ATS sends post request to origin
3. ATS sets up tunnel to send post body from client to origin
4. ATS receives and relays 100-continue response from origin
5. ATS receives and relays the POST response

So from the client's perspective, 
1. Send POST request
2. Send POST body
3. Receive 100-continue response
4. Receive POST body response

In this default sequence the 100-continue does absolutely nothing.  If the client was implementing things strictly, the client should stall sending the post body because it hasn't received the 100-continue response.

The reason I stumbled on this is that I was helping something debug "performance" issues with adding ATS to an exchange.  Adding ATS adding 1 second to the latency of a relatively small post request made by curl.  Even those the ATS and the origin were sitting right next to each other.  The issue is that proxy.config.http.send_100_continue_response was set to 0.  Curl didn't stall but it seemed to wait for around 1 second before sending along the post body anyway.

Since many people evaluate ATS using tools like curl, I think a more standards compliant and better out of the box experience would be to have proxy.config.http.send_100_continue_response  set to 1 by default.

The other alternative would be to adjust the ATS behavior to properly delay transmitting the POST body until the 100-continue response has been propagated.  That is a much bigger change, and I don't see anyone taking the time to do that in the near future..

shinrich:
The fixes the code to correctly send the request header and wait for the 100-continue response before setting up the post tunnel. Must of the logic was already in the code, but just not wired up correctly.  If you did not have proxy.config.http.send_100_continue_response set to 1 so ATS returns the 100 continue immediately, ATS would go ahead and set up the post tunnel before sending the request header and giving the origin a chance to respond with 100 continue.  Curl would still continue to work in the scenario although with higher latency because it would eventually timeout and send the post data anyway.

I had to adjust one of the tests because it appears that microserver does not correctly respond to the Expect header.

This closes #7346

### DIFF
--- a/proxy/http/HttpSM.cc
+++ b/proxy/http/HttpSM.cc
@@ -876,26 +876,26 @@ HttpSM::state_read_client_request_header(int event, void *data)
 
     if (t_state.hdr_info.client_request.version_get() == HTTP_1_1 &&
         (t_state.hdr_info.client_request.method_get_wksidx() == HTTP_WKSIDX_POST ||
-         t_state.hdr_info.client_request.method_get_wksidx() == HTTP_WKSIDX_PUT) &&
-        t_state.http_config_param->send_100_continue_response) {
+         t_state.hdr_info.client_request.method_get_wksidx() == HTTP_WKSIDX_PUT)) {
       int len            = 0;
       const char *expect = t_state.hdr_info.client_request.value_get(MIME_FIELD_EXPECT, MIME_LEN_EXPECT, &len);
-      // When receive an "Expect: 100-continue" request from client, ATS sends a "100 Continue" response to client
-      // immediately, before receive the real response from original server.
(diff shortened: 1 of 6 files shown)


shinrich:
[approve ci autest]

randall (approved):

PR #7963, opened 2021-06-21
Note YAML parser library bug, and work-around, in documentation.

ywkaras:
Currently the YAML parsing library has a bug where line number counting (for error messages) ignores comment lines that start with #.  A work-around is to put a space before the #.

### DIFF
--- a/doc/admin-guide/files/index.en.rst
+++ b/doc/admin-guide/files/index.en.rst
@@ -86,3 +86,9 @@ Configuration Files
 
 :doc:`volume.config.en`
     Defines cache space usage by individual protocols.
+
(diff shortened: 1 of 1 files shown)


SolidWallOfCode (approved):

randall (approved):
Curious, is there an yaml-cpp issue we can link to from this PR?

ywkaras:
> Curious, is there an yaml-cpp issue we can link to from this PR?

I don't see that there is one.  We copied the code into our repo, so it's hard to be sure if it hasn't been fixed.

zwoop:

PR #7968, opened 2021-06-21
STEK Share Plugin

duke8253:
This is a plugin used to share STEK among a cluster of ATS instances. It'll replace the current session reuse plugin.

### DIFF
--- a/plugins/experimental/stek_share/common.h
+++ b/plugins/experimental/stek_share/common.h
@@ -21,6 +21,7 @@
   limitations under the License.
 
  */
+
 #pragma once
 
 #include <iostream>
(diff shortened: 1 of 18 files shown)


plugins/experimental/stek_share/common.h
```diff
@@ -0,0 +1,67 @@
...
+#include <iomanip>
+#include <string>
+#include <cstring>
+#include <mutex>
+#include <deque>
+#include <cmath>
+
+#define PLUGIN "stek_share"
+
+class PluginThreads
```
shinrich, line 37: Is the PluginThreads class used?

duke8253: Yeah, this is for shutting down threads.

shinrich: Oh nevermind.  I didn't have all the files opened up.  I see its use now.

PR #7969, opened 2021-06-21
ESI plugin documentation updates.

bneradt:
This makes some minor tweaks to the EPI plugin document. The only
significant change to the doc is to remove the language saying that
`--enable-experimental-plugins` is needed to configure ATS to build the
ESI plugin. That's removed because it's not true: the ESI plugin has
been upgraded from an experimental plugin for a few releases already.

### DIFF
--- a/doc/admin-guide/plugins/esi.en.rst
+++ b/doc/admin-guide/plugins/esi.en.rst
@@ -1,4 +1,5 @@
 .. _admin-plugins-esi:
+.. include:: ../../common.defs
 
 ESI Plugin
 **********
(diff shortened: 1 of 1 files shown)


randall (approved):

randall:
EPI or ESI?

bneradt:
> EPI or ESI?

Oh, ESI! The patch is fine, but I need to update the commit.

Good catch.

bneradt:
Replaced with the following which doesn't accidentally reference EPI in any of the commit metadata:
https://github.com/apache/trafficserver/pull/7970

PR #7970, opened 2021-06-21
ESI plugin documentation updates.

bneradt:
This makes some minor tweaks to the ESI plugin document. The only
significant change to the doc is to remove the language saying that
`--enable-experimental-plugins` is needed to configure ATS to build the
ESI plugin. That's removed because it's not true: the ESI plugin has
been upgraded from an experimental plugin to a standard plugin for a few
releases already.


randall (approved):

shukitchan (approved):
It looks good. Thanks.

zwoop:

PR #7976, opened 2021-06-22
Be ready to process response headers sent before the post tunnel is finished

### LINKED ISSUE #4006: ATS does not propagate response header if response is sent and connection closed while post data tunnel is active
I've put up some fixes for this, but sadly was more difficult than I first thought.  So I'm filing the issue, so we don't forget this issue.

In the current code ATS will not try to read from the origin until it has completed the post tunnel.  If the origin gets bored and goes home, the post tunnel will fail on the write to origin.  ATS will not read from the origin connection.  Instead it will send back a canned 502 response.

At least one of the cases our customer see is the origin returning 413 Request Entity Too Large.  Moving forward with HTTP/2 not supporting 100-continue, I expect we will run into this case more often.

I tried to solve this by launching a server response header read in parallel with the post tunnel.  But tracking down all the edge cases was causing a whack-a-mole of crashes.  Perhaps we should use the option to read the entire post body into memory.  Then read from origin while writing from buffer.  No tunnel, maybe this is easier to handle.

Alternatively, maybe from the tunnel failure, we can try to read from the origin assuming the socket has not yet been closed.

I had to move onto other things, but we will need to deal with this eventually.

shinrich:
This PR sets up the response header read in parallel with the post tunnel.  In some cases the origin may return an error response header for a POST before processing all the post data.  With the current code that response is lost if the server closes the connection.

This PR is built upon PR #7962.  Once that is committed, I will rebase this branch.   All references to 100-continue belong with that PR.

I have added some tests to try and exercise the early server response.

This closes #4006

### DIFF
--- a/proxy/http/Http1ServerSession.cc
+++ b/proxy/http/Http1ServerSession.cc
@@ -207,6 +207,7 @@ Http1ServerSession ::release_transaction()
       HTTP_INCREMENT_DYN_STAT(http_origin_close_private);
     }
     this->do_io_close();
+    ink_release_assert(transact_count == released_transactions);
   } else if (state == SSN_TO_RELEASE) {
     _vc->control_flags.set_flags(0);
 
--- a/proxy/http/Http1ServerSession.cc
+++ b/proxy/http/Http1ServerSession.cc
@@ -207,7 +207,6 @@ Http1ServerSession ::release_transaction()
       HTTP_INCREMENT_DYN_STAT(http_origin_close_private);
     }
     this->do_io_close();
-    ink_release_assert(transact_count == released_transactions);
   } else if (state == SSN_TO_RELEASE) {
     _vc->control_flags.set_flags(0);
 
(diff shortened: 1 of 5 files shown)


shinrich:
Have been doing some production testing.  Another commit coming to fix an edge case.  But an interesting observation from the testing.  Our number of ERR_CONNECT_FAIL in a 15 minute access log file drops to single digits from triple digits (~250) on a peer running without this patch.  Digging in the differs seem to be POST requests to one of our origins.  We the patches, we see non 200 responses coming back for the origin (503's and 400's) instead of non origin status and 502 sent back to the user agent.

shinrich:
Testing on 1 production machine.  Added commit to clean up correctly in the cases where the server returns early with an error (i.e. EOS) during the dual post sending and response reading time.

bryancall:
Talked in the PR scrub and @shinrich is going to close the connection to the origin.

shinrich:
Updated to close connection to origin and adjustments made based on production experience.  Ran on production machines in two different colos for a week.

shinrich:
Pulled into our standard build.

bryancall:
@randall is going to look at this.

randall (approved):

PR #7981, opened 2021-06-23
Reject TLS if client offers alpn with no match

### LINKED ISSUE #7980: ATS should reject connections where the client offered ALPN does not have a match with the ATS list
Internally we've been discussing the cross-application attacks discussed in https://alpaca-attack.com/ALPACA.pdf.  And those discussions got us looking at how ATS handles ALPN.

Currently ATS will try to find a matching protocol if the client offers an ALPN entry.  If there is no match, ATS continues on and assumes HTTP/1.1.  If no ALPN string is offered (presumably an older client), ATS continues on and assumes HTTP/1.1

Looking at a snapshot of one of our production boxes logs, we see 92% of the client connections offer ALPN.  In this sample, no client requests that offered ALPN failed to get a protocol match with our ATS server which was willing to accept h2 and http/1.1.

To support older clients (8% of our requests), it is not feasible to stop accepting requests without ALPN. However, we should stop accepting requests where the ALPN string has no match.  Based on our logging, that seems unlikely to be an innocent mistake, but rather that the attacker is trying to poke at our security stance.

shinrich:
Motivation discussed in associated issue.

This closes #7980

### DIFF
--- a/iocore/net/ALPNSupport.cc
+++ b/iocore/net/ALPNSupport.cc
@@ -98,18 +98,24 @@ ALPNSupport::select_next_protocol(SSL *ssl, const unsigned char **out, unsigned 
 {
   const unsigned char *npnptr = nullptr;
   unsigned int npnsize        = 0;
-  if (this->getNPN(&npnptr, &npnsize)) {
+  int retval                  = SSL_TLSEXT_ERR_ALERT_FATAL;
+
+  if (this->getNPN(&npnptr, &npnsize) && npnsize > 0) {
     // SSL_select_next_proto chooses the first server-offered protocol that appears in the clients protocol set, ie. the
     // server selects the protocol. This is a n^2 search, so it's preferable to keep the protocol set short.
     if (SSL_select_next_proto(const_cast<unsigned char **>(out), outlen, npnptr, npnsize, in, inlen) == OPENSSL_NPN_NEGOTIATED) {
       Debug("ssl", "selected ALPN protocol %.*s", (int)(*outlen), *out);
-      return SSL_TLSEXT_ERR_OK;
+      retval = SSL_TLSEXT_ERR_OK;
+    } else {
+      *out    = nullptr;
+      *outlen = 0;
     }
+  } else {
+    *out    = nullptr;
+    *outlen = 0;
+    retval  = SSL_TLSEXT_ERR_NOACK;
   }
-
-  *out    = nullptr;
-  *outlen = 0;
-  return SSL_TLSEXT_ERR_NOACK;
+  return retval;
 }
 
 void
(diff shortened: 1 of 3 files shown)


shinrich:
[approve ci clang-analyzer]

iocore/net/ALPNSupport.cc
```diff
@@ -98,18 +98,24 @@ ALPNSupport::select_next_protocol(SSL *ssl, const unsigned char **out, unsigned
 {
   const unsigned char *npnptr = nullptr;
   unsigned int npnsize        = 0;
-  if (this->getNPN(&npnptr, &npnsize)) {
+  int retval                  = SSL_TLSEXT_ERR_ALERT_FATAL;
```
maskit, line 101: I don't think we can use this error code for 1.0.2.
https://www.openssl.org/docs/man1.0.2/man3/SSL_set_alpn_protos.html

maskit:
This works for QUIC connection as well - a connection closes appropriately.

We could remove the case `alpn.empty()` is empty here because we won't see such a case with this change.
https://github.com/apache/trafficserver/blob/427c0b153c3c6838c7d6e2db34c32e8091649d67/proxy/http3/Http3SessionAccept.cc#L63-L70

iocore/net/ALPNSupport.cc
shinrich: Hmm, how does one express the failure without the SSL_TLSEXT_ERR_ALERT_FATAL?  I'll dig into the 1.0.2 code base and add an ifdef if necessary.

shinrich:
So no additional code changes needed for QUIC?

maskit:
> So no additional code changes needed for QUIC?

Correct. I was just saying the code in Http3SessionAccept.cc (more specifically, line 67) will be unreachable. If you want to avoid touching HTTP/3 code I can remove it later.

iocore/net/ALPNSupport.cc
maskit: Returning `SSL_TLSEXT_ERR_ALERT_FATAL` isn't harmful, but it's impossible to distinguish no-match from no-negotiation.
https://github.com/openssl/openssl/blob/12ad22dd16ffe47f8cde3cddb84a160e8cdb3e30/ssl/t1_lib.c#L2087-L2105

maskit (approved):
We'd need some additional mechanism to do this with OpenSSL 1.0.2. Maybe we could have a flag and close the connection, but I'd say it's a library limitation.

iocore/net/ALPNSupport.cc
shinrich: Hmm that is unfortunate.  However, I think our time for worrying about 1.0.2 is coming to a close.

PR #7940, opened 2021-06-11
Pin Sphinx to 3.x to unblock `make html`

bneradt:
The latest version of Sphinx, 4.0.2, fails to build our docs via the
`make html` target. The latest 3.x release, currently 3.5.4, builds our
docs fine. Pinning to 3.x.

---

Fixes #7938

PR #7987, opened 2021-06-23
Pin Sphinx to 3.x to unblock `make html`

bneradt:
The latest version of Sphinx, 4.0.2, fails to build our docs via the
make html target. The latest 3.x release, currently 3.5.4, builds our
docs fine. Pinning to 3.x.

This also makes a fix for an API change in which sphinx.locale no longer
has `l_`.

### DIFF
--- a/doc/Pipfile
+++ b/doc/Pipfile
@@ -22,7 +22,16 @@ verify_ssl = true
 [dev-packages]
 
 [packages]
-sphinx = "2.0.1"
+
+# The latest 4.x sphinx release, currently 4.0.2, fails `make html`.  For
+# details, see: https://github.com/apache/trafficserver/issues/7938
+#
+# The 3.x releases build fine, however. So we currently pin to that.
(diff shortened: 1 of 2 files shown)


randall (approved):

zwoop (approved):
What could go wrong ...

bneradt:
I noticed this got marked for the 9.1.x project, but this is not needed there (and probably would result in a merge conflict if tried). The following PR for 9.1.x/master has already been merged in for this change: #7940 

I know this is a bit confusing, but this present PR (#7987) is a superset of that 9.1.x/master one (#7940), because the `doc/ext/traffic-server.py` file had to be updated in 9.0.x as well as the Sphinx pin, so a cherry-pick to 9.0.x of that latter PR would not have been sufficient. Thus this PR was filed separately.

In any case, this PR should not go to 9.1.x nor master. This is just for 9.0.x.

PR #7994, opened 2021-06-25
Allow to TLS handshake to error out on TSVConnReenable

shinrich:
Identified by @zwoop.  The TSVConnReenable with an ERROR event would only fail the TLS handshake from the cert verification hook.  This change would cause a failure from any of the hooks.


shinrich:
[approve ci clang-analyzer]

zwoop (approved):

zwoop:

We need this for the rate_limit plugin to work properly.

PR #7998, opened 2021-06-25
Timing and permissions update for regex_revalidate test

bneradt:
The regex_revalidate_state.test.py was failing when the autest was run
as a root user. This ensures that the state file it uses has the
appropriate permissions set so that the ATS process can access the file.
It also adds a ts wait condition upon the TS Process updating the
state file to address any possible timing related issues.


traeak (approved):
These changes improve the test.

PR #8004, opened 2021-06-28
Fix HPACK eviction iterator manipulation

### LINKED ISSUE #7838: HPACK crash in 9.0.1
We have been getting some background crashes with variations of the following stack since we moved from a 9.0.0 based build to a 9.0.1 based build.

```
[ 00 ] traffic_server               get_header_field                                                      ( HPACK.cc:357 )
[ 01 ] traffic_server               HpackIndexingTable::lookup(char const*, int, char const*, int) const  ( HPACK.cc:251 )
[ 02 ] traffic_server               hpack_encode_header_block                                             ( HPACK.cc:908 )
[ 03 ] traffic_server               http2_encode_header_blocks                                            ( HTTP2.cc:669 )
[ 04 ] traffic_server               Http2ConnectionState::send_headers_frame(Http2Stream*)                ( Http2ConnectionState.cc:1651 )
[ 05 ] traffic_server               Http2Stream::update_write_request(IOBufferReader*, long, bool)        ( Http2Stream.cc:626 )
[ 06 ] traffic_server               Http2Stream::do_io_write(Continuation*, long, IOBufferReader*, bool)  ( Http2Stream.cc:348 )
[ 07 ] traffic_server               HttpTunnel::producer_run(HttpTunnelProducer*)                         ( HttpTunnel.cc:958 )
[ 08 ] traffic_server               HttpTunnel::tunnel_run(HttpTunnelProducer*)                           ( HttpTunnel.cc:700 )
[ 09 ] traffic_server               HttpSM::state_api_callout(int, void*)                                 ( HttpSM.cc:1543 )
[ 10 ] traffic_server               HttpSM::state_api_callback(int, void*)                                ( HttpSM.cc:1350 )
[ 11 ] traffic_server               TSHttpTxnReenable                                                     ( InkAPI.cc:6230 )
[ 12 ] header_rewrite.so (deleted)                                                                                                                   
```

I will try to dig into an actual core file to get some more details.  From what I can see in backtracio, the HPACK table looks ok and the index seems reasonable given the size of the map.

shinrich:
Found this while investigating #7838.  According to https://en.cppreference.com/w/cpp/container/deque/pop_back#:~:text=std%3A%3Adeque%3A%3Apop_back&text=Removes%20the%20last%20element%20of,the%2Dend%20iterator%20is%20invalidated. After pop_back is called "Iterators and references to the erased element are invalidated".  In the current logic, the increment is called on the old iterator after pop_back is called.

I'm not 100% sure this is causing our crash.  If most of the time only one item is evicted, this case won't happen very often and maybe most of the time you get lucky.

I will get this patched rolled into our build. Even if it does not fix the crash the new logic is safer and looks cleaner.

This closes #7838

### DIFF
--- a/proxy/http2/HPACK.cc
+++ b/proxy/http2/HPACK.cc
@@ -461,17 +461,18 @@ HpackDynamicTable::_evict_overflowed_entries()
     return;
   }
 
-  for (auto h = this->_headers.rbegin(); h != this->_headers.rend(); ++h) {
+  while (!this->_headers.empty()) {
+    auto h = this->_headers.back();
     int name_len, value_len;
-    (*h)->name_get(&name_len);
-    (*h)->value_get(&value_len);
+    h->name_get(&name_len);
+    h->value_get(&value_len);
 
     this->_current_size -= ADDITIONAL_OCTETS + name_len + value_len;
 
     if (this->_mhdr_old && this->_mhdr_old->fields_count() != 0) {
-      this->_mhdr_old->field_delete(*h, false);
+      this->_mhdr_old->field_delete(h, false);
     } else {
-      this->_mhdr->field_delete(*h, false);
+      this->_mhdr->field_delete(h, false);
     }
 
     this->_headers.pop_back();


shinrich:
[approve ci freebsd]

duke8253 (approved):

bryancall:
@masaori335 will take a look

masaori335 (approved):
Looks reasonable regardless of this fixes the crash or not.

zwoop:

PR #8008, opened 2021-06-29
Change header rewrite query string handling to allow empty

### LINKED ISSUE #4670: Allow header_rewrite to remove query parameters
header_rewrite config:

`set-destination QUERY ""`

result:

`[Nov 29 22:45:49.326] {0x2af11f66d700} DIAG: (header_rewrite) Would set destination QUERY to an empty value, skipping`

The docs actually mention this - but it seems useful in this case:

> You must supply a non-zero length value, otherwise this operator will be an effective no-op (though a warning will be emitted to the logs if debugging is enabled).

https://docs.trafficserver.apache.org/en/latest/admin-guide/plugins/header_rewrite.en.html#set-destination

ezelkow1:
Change header rewrite query string handling to allow empty

Currently query requires a non-empty string. This changes that to allow empty to erase queries. If QSA is added to the rule then the existing query string will be appended so if the new one is empty its effectively a no-op but will still be processed
This allows us to be able to drop query strings via header rewrite, currently this is mostly done using the cachekey plugin however you can only use a single instance of cachekey per remap. Adding this allows us to do it via header rewrite and free up cachekey usage for a simple case

closes #4670

### DIFF
--- a/plugins/header_rewrite/operators.cc
+++ b/plugins/header_rewrite/operators.cc
@@ -210,23 +210,22 @@ OperatorSetDestination::exec(const Resources &res) const
 
     case URL_QUAL_QUERY:
       _value.append_value(value, res);
-      if (value.empty()) {
-        TSDebug(PLUGIN_NAME, "Would set destination QUERY to an empty value, skipping");
-      } else {
-        // 1.6.4--Support for preserving QSA in case of set-destination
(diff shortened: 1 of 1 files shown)


ezelkow1:
Closing for now, after some discussion we should make this a first class operation adding an rm-destination for query (and possibly other components that dont have a current rm equivalent)

PR #8012, opened 2021-06-29
Add option to mostly strictly check URL characters

shinrich:
Augments proxy.config.http.strict_uri_parsing to take three values rather than just two.  Adding a mostly strict URI checking mode that will not enforce the full RFC restrictions but will ensure that white space and unprintable characters are not allowed.  This restricting will be enough to detect other problems in the request line that result in odd URI's based on how ATS parses the request line.

Updates docs, unit test, and autest.  Does not change the default value of proxy.config.http.strict_uri_parsing but may want to consider making 2 the default on the next major release.

### DIFF
--- a/doc/admin-guide/files/records.config.en.rst
+++ b/doc/admin-guide/files/records.config.en.rst
@@ -1221,8 +1221,10 @@ mptcp
 
 .. ts:cv:: CONFIG proxy.config.http.strict_uri_parsing INT 0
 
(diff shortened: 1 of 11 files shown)


shinrich:
[approve ci autest]

maskit (approved):
Not a big deal, but we may want to update places that pass boolean values.

zwoop:

ezelkow1:
Can we get a backport to 8.1.x for this?

Issue #6249, opened 2019-12-10
Custom proxy.config.output.logfile has no effect


xeron:
trafficserver version: 8.0.5.
OS: Amazon Linux v1 (CentOS 6 based)

Related config parts:

```
CONFIG proxy.config.log.logfile_dir STRING /mnt/logs/trafficserver
CONFIG proxy.config.output.logfile STRING /mnt/logs/trafficserver/traffic.out
```

Actual files:
```
# ls -l /mnt/logs/trafficserver/
total 56
-rw-r--r-- 1 ats ats  6522 Dec 10 08:44 access.log
-rw-r--r-- 1 ats ats 19891 Dec 10 09:04 diags.log
-rw-r--r-- 1 ats ats 21905 Dec 10 09:04 manager.log

# ls -l /opt/trafficserver/var/log/trafficserver/
total 8
-rw-r--r-- 1 root root 5844 Dec 10 09:04 traffic.out
```

xeron:
Other issues I noticed which are kind of related:

* The following options have no effect on diags/manager/error logs and only affect `access.log`:
```
CONFIG proxy.config.log.rolling_enabled INT 3
CONFIG proxy.config.log.rolling_offset_hr INT 3
CONFIG proxy.config.log.rolling_interval_sec INT 86400
CONFIG proxy.config.log.rolling_size_mb INT 1000
CONFIG proxy.config.log.auto_delete_rolled_files INT 1
```

* The following options have no effect on `traffic.out`:
```
CONFIG proxy.config.output.logfile.rolling_enabled INT 3
CONFIG proxy.config.output.logfile.rolling_interval_sec INT 86400
CONFIG proxy.config.output.logfile.rolling_size_mb INT 1000
```

xeron:
Just found that default init script sets `--bind_stdout` and `--bind_stderr` on redhat… could be the reason I guess?

bneradt:
Yes, I happened to just come across this code while working on #7937. Our documentation is a bit misleading about `proxy.config.output.logfile`. It currently, it says:

https://docs.trafficserver.apache.org/en/latest/admin-guide/files/records.config.en.html?highlight=proxy.config.output.logfile#proxy.config.output.logfile

> The name and location of the file that contains warnings, status messages, and error messages produced by the Traffic Server processes. If no path is specified, then Traffic Server creates the file in its logging directory.

[...]

PR #8013, opened 2021-06-29
Doc: clarify logfile description

bneradt:
The traffic_manager and traffic_server stdout and stderr streams can be
bound via the --bind_stdout and --bind_stderr command line options. If
this is done, then `proxy.config.output.logfile` can be used to identify
that path for log rolling. This change clarifies that the latter
configuration is only for log rolling configuration identification
purposes and that the filename that ATS uses is actually set via the
--bind options.

---

This fixes: #6249

### DIFF
--- a/doc/admin-guide/files/records.config.en.rst
+++ b/doc/admin-guide/files/records.config.en.rst
@@ -213,13 +213,25 @@ System Variables
 
 .. ts:cv:: CONFIG proxy.config.output.logfile  STRING traffic.out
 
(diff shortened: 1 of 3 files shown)


bneradt:
This is a doc-only clarification that applies to the 9 release branches as well as master, so marking for those branches as well. Looking through the 8.1.x docs, this doesn't apply to that branch, so not  marking it for 8.1.x.

randall (approved):

zwoop:

PR #7298, opened 2020-10-27
Adds OCSP support for BoringSSL

randall:

PR #8014, opened 2021-06-29
Dual cert support for BoringSSL

randall:
BoringSSL removed support for slotted SSL_CTXs. This is key for
 supporting multiple certs for a CTX. When BoringSSL is used,
 multiple CTXs will be created if more than one cert type is specified.
 When a client connection does come in, ATS will interrogate the
  connection and present the approriate CTX (preferring ECDSA)

### DIFF
--- a/iocore/net/SSLUtils.cc
+++ b/iocore/net/SSLUtils.cc
@@ -38,6 +38,7 @@
 #include "P_OCSPStapling.h"
 #include "P_SSLSNI.h"
 #include "P_SSLConfig.h"
+#include "BoringSSLUtils.h"
 #include "ProxyProtocol.h"
 #include "SSLSessionCache.h"
 #include "SSLSessionTicket.h"
@@ -253,16 +254,18 @@ ssl_rm_cached_session(SSL_CTX *ctx, SSL_SESSION *sess)
 }
 
 static int
-set_context_cert(SSL *ssl)
+set_context_cert(SSL *ssl, void *arg)
 {
   shared_SSL_CTX ctx  = nullptr;
   SSL_CTX *verify_ctx = nullptr;
   SSLCertContext *cc  = nullptr;
   SSLCertificateConfig::scoped_config lookup;
-  const char *servername   = SSL_get_servername(ssl, TLSEXT_NAMETYPE_host_name);
-  SSLNetVConnection *netvc = SSLNetVCAccess(ssl);
-  bool found               = true;
-  int retval               = 1;
+
+  const char *servername     = SSL_get_servername(ssl, TLSEXT_NAMETYPE_host_name);
+  SSLNetVConnection *netvc   = SSLNetVCAccess(ssl);
+  bool found                 = true;
+  int retval                 = 1;
+  SSLCertContextType ctxType = SSLCertContextType::GENERIC;
 
   if (!netvc || netvc->ssl != ssl) {
     Debug("ssl.error", "set_context_cert call back on stale netvc");
@@ -279,11 +282,19 @@ set_context_cert(SSL *ssl)
     goto done;
   }
 
+#ifdef OPENSSL_IS_BORINGSSL
+  if (arg != nullptr) {
+    const SSL_CLIENT_HELLO *client_hello = (const SSL_CLIENT_HELLO *)arg;
+    const bool client_ecdsa_capable      = BoringSSLUtils::isClientEcdsaCapable(client_hello);
+    ctxType                              = client_ecdsa_capable ? SSLCertContextType::EC : SSLCertContextType::RSA;
+  }
+#endif
+
   // The incoming SSL_CTX is either the one mapped from the inbound IP address or the default one. If we
   // don't find a name-based match at this point, we *do not* want to mess with the context because we've
   // already made a best effort to find the best match.
   if (likely(servername)) {
-    cc = lookup->find(const_cast<char *>(servername));
+    cc = lookup->find(servername, ctxType);
     if (cc) {
       ctx = cc->getCtx();
     }
@@ -342,6 +353,7 @@ set_context_cert(SSL *ssl)
   }
 
   verify_ctx = SSL_get_SSL_CTX(ssl);
+  // set_context_cert found SSL context for ...
   Debug("ssl", "ssl_cert_callback %s SSL context %p for requested name '%s'", found ? "found" : "using", verify_ctx, servername);
 
   if (verify_ctx == nullptr) {
@@ -410,14 +422,48 @@ ssl_client_hello_callback(SSL *s, int *al, void *arg)
   }
   return SSL_CLIENT_HELLO_SUCCESS;
 }
+#elif defined(OPENSSL_IS_BORINGSSL)
+static ssl_select_cert_result_t
+ssl_client_hello_callback(const SSL_CLIENT_HELLO *client_hello)
+{
+  SSL *s = client_hello->ssl;
+  TLSSNISupport *snis = TLSSNISupport::getInstance(s);
+
+  if (snis) {
+    snis->on_client_hello(client_hello);
+    int ret = snis->perform_sni_action();
+    if (ret != SSL_TLSEXT_ERR_OK) {
+      return ssl_select_cert_error;
+    }
+  } else {
+    // This error suggests either of these:
+    // 1) Call back on unsupported netvc -- Don't register callback unnecessarily
+    // 2) Call back on stale netvc
+    Debug("ssl.error", "ssl_client_hello_callback was called unexpectedly");
+    return ssl_select_cert_error;
+  }
+
+  SSLNetVConnection *netvc = SSLNetVCAccess(s);
+  if (!netvc || netvc->ssl != s) {
+    Debug("ssl.error", "ssl_client_hello_callback call back on stale netvc");
+    return ssl_select_cert_error;
+  }
+
+  bool reenabled = netvc->callHooks(TS_EVENT_SSL_CLIENT_HELLO);
+
+  if (!reenabled) {
+    return ssl_select_cert_retry;
+  }
+  return ssl_select_cert_success;
+}
 #endif
 
 /**
  * Called before either the server or the client certificate is used
- * Return 1 on success, 0 on error, or -1 to pause
+ * Return 1 on success, 0 on error, or -1 to pause, -2 to retry
  */
 static int
-ssl_cert_callback(SSL *ssl, void * /*arg*/)
+ssl_cert_callback(SSL *ssl, void *arg)
 {
   SSLNetVConnection *netvc = SSLNetVCAccess(ssl);
   bool reenabled;
@@ -430,13 +476,17 @@ ssl_cert_callback(SSL *ssl, void * /*arg*/)
 
   // If we are in tunnel mode, don't select a cert.  Pause!
   if (HttpProxyPort::TRANSPORT_BLIND_TUNNEL == netvc->attributes) {
+#ifdef OPENSSL_IS_BORINGSSL
+    return -2; // Retry
+#else
     return -1; // Pause
+#endif
   }
 
   // Do the common certificate lookup only once.  If we pause
   // and restart processing, do not execute the common logic again
   if (!netvc->calledHooks(TS_EVENT_SSL_CERT)) {
-    retval = set_context_cert(ssl);
+    retval = set_context_cert(ssl, arg);
     if (retval != 1) {
       return retval;
     }
@@ -463,7 +513,7 @@ ssl_servername_callback(SSL *ssl, int *al, void *arg)
   TLSSNISupport *snis = TLSSNISupport::getInstance(ssl);
   if (snis) {
     snis->on_servername(ssl, al, arg);
-#if !TS_USE_HELLO_CB
+#if !TS_USE_HELLO_CB && !defined(OPENSSL_IS_BORINGSSL)
     // Only call the SNI actions here if not already performed in the HELLO_CB
     int ret = snis->perform_sni_action();
     if (ret != SSL_TLSEXT_ERR_OK) {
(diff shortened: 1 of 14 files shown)


randall:
[approve ci autest]

randall:
[approve ci autest]

iocore/net/SSLUtils.cc
```diff
@@ -1153,99 +1290,132 @@ setClientCertCACerts(SSL *ssl, const char *file, const char *dir)
...
+  std::vector<SSLLoadingContext> ret;
+  std::vector<std::vector<std::string>> cert_names;
+  std::vector<std::vector<std::string>> key_names;
+  std::vector<std::string> key_names_list;
+  ink_assert(data.cert_names_list.size() == data.cert_type_list.size());
+  //  ink_assert(data.key_names_list.size() == data.key_names_list.size());
 
-  SSL_CTX *ctx = this->default_server_ssl_ctx();
+#ifdef OPENSSL_IS_BORINGSSL
+  for (auto name : data.cert_names_list) {
```
bneradt, line 1307: `auto const& name`

iocore/net/SSLUtils.cc
```diff
@@ -1153,99 +1290,132 @@ setClientCertCACerts(SSL *ssl, const char *file, const char *dir)
...
+  }
+#else
+  cert_names.emplace_back(data.cert_names_list);
+  key_names.emplace_back(data.key_list);
+#endif
 
-  // disable selected protocols
-  SSL_CTX_set_options(ctx, params->ssl_ctx_options);
+  int i = 0;
+  for (auto cert_names_list : cert_names) {
```
bneradt, line 1319: `auto cosnt& cert_names_list`

iocore/net/SSLUtils.cc
```diff
@@ -1473,27 +1643,37 @@ SSLCreateServerContext(const SSLConfigParams *params, const SSLMultiCertConfigPa
...
-      } else if (!SSL_CTX_check_private_key(ctx.get())) {
-        SSLError("SSLCreateServerContext(): server private key does not match server certificate.");
+  SSLCertContextType cert_type;
+  if (!loader.load_certs_and_cross_reference_names(cert_list, data, params, sslMultiCertSettings, common_names, unique_names,
+                                                   &cert_type)) {
+    return nullptr;
+  }
+
+  std::vector<SSLLoadingContext> ctxs = loader.init_server_ssl_ctx(data, sslMultiCertSettings, common_names);
+  for (auto loaderctx : ctxs) {
```
bneradt, line 1655: `auto const&`

iocore/net/SSLUtils.cc
```diff
@@ -1549,17 +1733,17 @@ SSLMultiCertConfigLoader::_store_ssl_ctx(SSLCertLookup *lookup, const shared_SSL
     return false;
   }
 
-  if (!common_names.empty()) {
-    shared_SSL_CTX ctx(this->init_server_ssl_ctx(data, sslMultCertSettings.get(), common_names), SSL_CTX_free);
-
-    if (!ctx || !sslMultCertSettings || !this->_store_single_ssl_ctx(lookup, sslMultCertSettings, ctx, common_names)) {
+  std::vector<SSLLoadingContext> ctxs = this->init_server_ssl_ctx(data, sslMultCertSettings.get(), common_names);
+  for (auto loadingctx : ctxs) {
```
bneradt, line 1737: `auto const&` ?

bneradt (dismissed):
Looks good. In general the patch looks fine, I just have a few suggestions about using `auto&` or `auto const&` if it's possible in some places. It would probably be helpful to have someone more familiar with TLS than I review this as well.

randall:
> Looks good. In general the patch looks fine, I just have a few suggestions about using `auto&` or `auto const&` if it's possible in some places. It would probably be helpful to have someone more familiar with TLS than I review this as well.

Thanks for the early feedback I think I hit all those autos (and maybe some more in that file)!

configure.ac
```diff
@@ -1276,9 +1279,15 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#include <openssl/ssl.h>]],
                     AC_MSG_RESULT([no])
                   ])
 
+AC_ARG_ENABLE([quic],
+  AS_HELP_STRING([--disable-quic], [disable QUIC support]), [
```
maskit, line 1283: I'm fine with having this option if you want to completely disable QUIC, but it doesn't make much sense to me because I think there's only overhead on startup if ports for QUIC are not configured.

Can you tell me what made you want to add this option?

randall: Purely for building in Xcode. I have a project I maintain and wanted to continue to use Xcode for developing the BoringSSL changes, but using that would force QUIC support. And I'm lazy and didn't want to add the QUIC files to it (especially since default openssl doesn't have the appropriate support).

maskit: >  didn't want to add the QUIC files to it

Can you tell me a bit more detail about this part? Do those files bother Xcode somehow?

build/crypto.m4
```diff
@@ -365,3 +365,31 @@ AC_DEFUN([TS_CHECK_SESSION_TICKET], [
 
   AC_SUBST(has_tls_session_ticket)
 ])
+
+dnl
+dnl Weird versions of BoringSSL do not have SSL_CTX_set_tmp_dh
```
maskit, line 370: Just FYI, SSL_CTX_set_tmp_dh (and DH_get_2048_256) will be deprecated on OpenSSL 3.

configure.ac
randall: It's a manual process of add/removing files. The files themselves are fine as long as they're not part of the build, which I need if I'm building against OpenSSL). but that requires either me maintaining a separate branch/copy of the project for QUIC support. 

If this is controversial compile time option, I'll just drop it from the changeset before I merge.

maskit: I'm purely curious why the option is needed, and I still don't get the problem... maybe because I don't use Xcode, but if it's helpful for some situations I don't against adding it as long as you are not making changes that are incompatible with QUIC support (having QUIC support as todo is fine).

I'd make another PR for this compile time option so that people can easily cherrypick the change for their branch.

Current expected behavior of the configure script is to enable QUIC support whenever SSL library supports QUIC. In other words, the files should not be part of the build if the library doesn't support QUIC even if you use BoringSSL.

build/crypto.m4
randall: I'm going to drop this patch and introduce it separately (with conditions for DH_get_2048_256)

randall:
[approve ci clang-analyzer]

configure.ac
randall: I've removed the option from this PR (and will work with a local copy of if it in the future)

iocore/net/YamlSNIConfig.cc
```diff
@@ -136,7 +136,7 @@ std::set<std::string> valid_sni_config_keys = {TS_fqdn,
                                                TS_client_sni_policy,
                                                TS_http2,
                                                TS_ip_allow,
-#if TS_USE_HELLO_CB
+#if TS_USE_HELLO_CB || defined(OPENSSL_IS_BORINGSSL)
                                                TS_valid_tls_versions_in,
```
maskit, line 140: We may want to update the documentation.

https://github.com/apache/trafficserver/blame/3412c4f69552b10779cfe0ce1bc6a71971734491/doc/admin-guide/files/sni.yaml.en.rst#L99-L100

iocore/net/SSLUtils.cc
```diff
@@ -412,12 +424,48 @@ ssl_client_hello_callback(SSL *s, int *al, void *arg)
 }
 #endif
 
+#ifdef OPENSSL_IS_BORINGSSL
+static ssl_select_cert_result_t
+ssl_client_hello_callback(const SSL_CLIENT_HELLO *client_hello)
```
shinrich, line 429: Maybe you want a different name for the BORINGSSL version of this function, which is really a utility function for the callback instead of the proper callback.  Took me a few minutes looking around to resolve in my head that the boring version of the function and openssl version with the same name weren't going to conflict with each other.

iocore/net/SSLUtils.cc
```diff
@@ -430,13 +478,17 @@ ssl_cert_callback(SSL *ssl, void * /*arg*/)
 
   // If we are in tunnel mode, don't select a cert.  Pause!
   if (HttpProxyPort::TRANSPORT_BLIND_TUNNEL == netvc->attributes) {
+#ifdef OPENSSL_IS_BORINGSSL
+    return -2; // Retry
```
shinrich, line 482: What is the difference between pause and retry with boring?  Guess I should go back and look more carefully at your docs.

shinrich:
I notice that you enabled one of the tlshook checks with this test (removed the openssl requirement).  But I see that the tls_check_dual_cert* checks have no restriction to only openssl.  Are they failing when built against boringssl without this PR?

bryancall:
@shinrich is going to take a look

randall:
> I notice that you enabled one of the tlshook checks with this test (removed the openssl requirement). But I see that the tls_check_dual_cert* checks have no restriction to only openssl. Are they failing when built against boringssl without this PR?

Currently, a large number of tls tests fail when run under BoringSSL. I added https://github.com/apache/trafficserver/pull/8014 recently which added the IsOpenSSL condition, and marked up a few tests that I knew depended on OpenSSL. I didn't add the dual cert tests because I knew I had this PR and https://github.com/apache/trafficserver/pull/7298 in the wings. 

And even after this PR, there are still a few tls_hooks test that still fail which I probably won't worry about since we don't depend on them. I'll document(by adding the Condition.IsOpenSSL) to them.

iocore/net/SSLUtils.cc
maskit: I don't think this function should have a different name, although it's a shame that we have two very similar functions.

The lambda expression in `_set_handshake_callbacks` mimics OpenSSL's behavior, and it enables reusing two other callback functions as well. In that sense I think having the same name make sense.

Having this in a `#elif` clause may ease the confusion.
```cpp
#if TS_USE_HELLO_CB
static int
ssl_client_hello_callback(SSL *s, int *al, void *arg)
{
}
#elif defined(OPENSSL_IS_BORINGSSL)
static ssl_select_cert_result_t
ssl_client_hello_callback(const SSL_CLIENT_HELLO *client_hello)
{
}
#endif
```

And we may also be able to have exactly the same function signature by passing `SSL_CLIENT_HELLO` to `arg`.

iocore/net/SSLUtils.cc
maskit: Seems like there is not "pause" in BoringSSL. Maybe they just use different terms.

OpenSSL
> The callback can then inspect the passed ssl structure and set or clear any appropriate certificates. If the callback is successful it MUST return 1 even if no certificates have been set. A zero is returned on error which will abort the handshake with a fatal internal error alert. A negative return value will suspend the handshake and the handshake function will return immediately. SSL_get_error(3) will return SSL_ERROR_WANT_X509_LOOKUP to indicate, that the handshake was suspended. The next call to the handshake function will again lead to the call of cert_cb(). It is the job of the cert_cb() to store information about the state of the last call, if required to continue.

BoringSSL
```cpp
enum ssl_select_cert_result_t BORINGSSL_ENUM_INT {
  // ssl_select_cert_success indicates that the certificate selection was
  // successful.
  ssl_select_cert_success = 1,
  // ssl_select_cert_retry indicates that the operation could not be
  // immediately completed and must be reattempted at a later point.
  ssl_select_cert_retry = 0,
  // ssl_select_cert_error indicates that a fatal error occured and the
  // handshake should be terminated.
  ssl_select_cert_error = -1,
};
```

iocore/net/SSLUtils.cc
randall: Doing the #elif here makes sense.

The callback return values are different types (and different values for equivalent states):

```
enum ssl_select_cert_result_t BORINGSSL_ENUM_INT {
  // ssl_select_cert_success indicates that the certificate selection was
  // successful.
  ssl_select_cert_success = 1,
  // ssl_select_cert_retry indicates that the operation could not be
  // immediately completed and must be reattempted at a later point.
  ssl_select_cert_retry = 0,
  // ssl_select_cert_error indicates that a fatal error occured and the
  // handshake should be terminated.
  ssl_select_cert_error = -1,
};
```
vs openssl:
```
# define SSL_CLIENT_HELLO_SUCCESS 1
# define SSL_CLIENT_HELLO_ERROR   0
# define SSL_CLIENT_HELLO_RETRY   (-1)
```

maskit: I'm fine with the current code, but I guess we could have aliases for the type and also the values.
https://github.com/apache/trafficserver/blob/870bbdb18eb267e4fd0ebebf3d667ac0dd1d5479/iocore/net/P_SSLUtils.h#L45-L49

iocore/net/SSLUtils.cc
```diff
@@ -1550,17 +1697,17 @@ SSLMultiCertConfigLoader::_store_ssl_ctx(SSLCertLookup *lookup, const shared_SSL
...
+    if (!sslMultCertSettings || !this->_store_single_ssl_ctx(lookup, sslMultCertSettings, ctx, loadingctx.ctx_type, common_names)) {
       retval = false;
       std::string names;
-      for (auto name : data.cert_names_list) {
+      for (auto const &name : data.cert_names_list) {
         names.append(name);
         names.append(" ");
       }
-      Warning("(%s) Failed to insert SSL_CTX for certificate %s entries for names already made", this->_debug_tag(), names.c_str());
+      Warning("(%s) Failed to insert SSL_CTX for certificate %sentries for names already made", this->_debug_tag(), names.c_str());
```
shinrich, line 1710: Lost a space between %s and entries

shinrich:
I made on rather nit-picky comment on a Debug message.  Once the clang-format is happy, I am willing to approve.

iocore/net/SSLUtils.cc
randall: I did it on purpose(message ends up having double space in log message), but not worth worrying about

shinrich (approved):

PR #8015, opened 2021-06-29
Update documentation of the url log fields

### LINKED ISSUE #4684: ATS 7.1 logging - cquc: Client Request
Hi
I want to track **urls of client requests** so in the logging.config:
```
-- Squid Log Format.
squid = format {
  Format = '%<cqts> %<ttms> %<chi> %<crc>/%<pssc> %<psql> %<cqhm> %<cquc> %<caun> %<phr>/%<pqsn> %<psct>'
}

log.ascii {
  Format = squid,
  Filename = 'squid'
}
```
In the ATS documentation:
`
cquc | Client Request | Canonical URL from the client request to Traffic Server. This field differs from cqu by having its contents URL-escaped (spaces and various other characters are replaced by percent-escaped entity codes).
`

I thought `cquc `will be the client request (the url I paste in my browser) but it is actually the origin server.

shinrich:
Trying to clarify the documentation for the client/proxy request logging fields.

This closes #4684

### DIFF
--- a/doc/admin-guide/logging/formatting.en.rst
+++ b/doc/admin-guide/logging/formatting.en.rst
@@ -811,17 +811,20 @@ paths), from transactions processed by |TS|.
 ===== ============== ==========================================================
 Field Source         Description
 ===== ============== ==========================================================
-cqu   Proxy Request  URI of the client request to |TS| (a subset of cqtx_). In
-                     reverse proxy mode, |TS| logs the rewritten/mapped URL
-                     (according to the rules in :file:`remap.config`), not the
-                     pristine/unmapped URL.
-cquc  Client Request Canonical URL from the client request to |TS|. This field
+cqu   Proxy Request  URL of the proxy request from |TS| to the origin. If a
+                     remap rule is used (:file:`remap.config`), the original
+                     client URL is replaced with the URL in the target of the
+                     remap rule.
+cquc  Proxy Request  Canonical URL from the proxy request to origin. This field
                      differs from cqu_ by having its contents URL-escaped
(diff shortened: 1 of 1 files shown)


maskit:
I don't think the changes are correct because "cq" in log field names mean "Client reQuest". In fact, the data source of `cqu` is `LogAccess::m_client_request` but not `LogAccess::m_proxy_request`.

Are we sure `m_client_request` and `m_proxy_request` always have the same URL regardless of transaction state? In other words, can we really change the data source to `m_proxy_request` ?

If the answer is yes, the new descriptions probably match actual outputs. However, using "cq" for data from Proxy Request doesn't seem correct. I'd deprecate confusing `cqu` and rename it to `pqu`, and also change the data source to `m_proxy_request`.

If the answer is no, the doc should not say the source is Proxy Request (nor URL in a request form proxy to origin).

Log field names for content length is very clear. We should have followed this manner.
```
cqcl | Client Request | Client request content length, in bytes.
pqcl | Proxy Request | Content body length of the Traffic Server proxy request to the origin server.
pscl | Proxy Response | Content body length of the Traffic Server proxy response.
sscl | Origin Response | Content body length of the origin server response to Traffic Server.
```

shinrich:
Agreed that the original log format names (cq) are confusing.  The updated description matches reality to the best of my reading of the code and reviewing log output.  I'll do as you suggest and add some new format names and deprecate the old ones.

rob05c (approved):
LGTM

shinrich:
Filled #8196 to track fixing the log formats as @maskit suggests.

Issue #5702, opened 2019-07-09
Ignore If_modified_since even if if_none_match field's value is empty. 


scw00:
https://tools.ietf.org/html/rfc7232#section-3.3

>A recipient MUST ignore If-Modified-Since if the request contains an If-None-Match header field;

IIUC we should ignore If-Modified-Since once the request contains If-None-Match header field. `Although the value is empty`.

bryancall:
Look like this was fixed in #5701 and merged

PR #5701, opened 2019-07-09
fix If-Match and If-Unmodified-Since priority problem

WolfgangBai:
A recipient MUST ignore If-Unmodified-Since if the request contains
   an If-Match header field，about rfc https://tools.ietf.org/html/rfc7232#section-3.3.

PR #8018, opened 2021-06-30
Ignore if_modified_since when if_none_match is present

constreference:
Addresses https://github.com/apache/trafficserver/issues/5702
https://datatracker.ietf.org/doc/html/rfc7232#section-3.3


bryancall:
[approve ci]

bryancall:
Look like this was fixed in #5701 and merged - It looks like this PR is just cleaning up initialization?

constreference:
@bryancall, That PR addresses If-Match and If-Unmodified-Since priority, while this looks at If-None-Match and If-Modified, Currently if an etag is not present, we move to check the If-Modified-Since presence

bryancall:
[approve ci]

PR #8025, opened 2021-07-01
header_rewrite rm-destination support

### LINKED ISSUE #4670: Allow header_rewrite to remove query parameters

### LINKED ISSUE #8024: Add rm-destinations for various parts in header rewrite
We have a couple components right now that to delete them expect to be used with a set-destination and empty string, i.e. port, an upcoming PR for query, etc.

We should add first class rm-destination support for these as well

ezelkow1:
Adds a new `rm-destination`, this lets you specify either QUERY or PATH, and be able to drop them from the incoming request

closes #8024 
closes #4670

### DIFF
--- a/doc/admin-guide/plugins/header_rewrite.en.rst
+++ b/doc/admin-guide/plugins/header_rewrite.en.rst
@@ -689,6 +689,18 @@ used as its replacement. You must supply a non-zero length value, otherwise
 this operator will be an effective no-op (though a warning will be emitted to
 the logs if debugging is enabled).
 
+rm-destination
+~~~~~~~~~~~~~~
+::
+
+  rm-destination <part>
+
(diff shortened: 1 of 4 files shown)


zwoop (approved):
Love it!

zwoop (approved):

PR #7501, opened 2021-02-08
JeMalloc performance improvments

duke8253:
Testing to see whether this helps with memory allocation lock contention.

PR #8028, opened 2021-07-01
Use thread_local to speed things up in JeAllocator

duke8253:
As mentioned in #7501 , testing whether `thread_local` helps with performance. @bryancall will help with benchmarking.

### DIFF
--- a/include/tscore/JeAllocator.h
+++ b/include/tscore/JeAllocator.h
@@ -70,15 +70,13 @@ public:
 
 private:
 #if JEMALLOC_NODUMP_ALLOCATOR_SUPPORTED
-  static extent_alloc_t *original_alloc;
-  static extent_hooks_t extent_hooks;
-  static std::shared_mutex je_mutex;
-  static std::unordered_map<pthread_t, unsigned int> arenas;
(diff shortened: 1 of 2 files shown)


bryancall:
I ran benchmarks on this PR on our AMD Epyc 64/128 servers and the rps went from 35K rps to 98K rps with this change.  This PR removes the reader/writer locks to fetch the arenas.

bryancall (approved):

zwoop (approved):
Yeh, this is good!

This is why I want that "ScheduleOnEveryThread()" function, so a plugin can use thread local, but (re)initialize its local thread data on demand :-).

zwoop:
I believe (please correct me if I'm wrong) that this fixes the locking issues introduced via #7501 , which is not in 9.1.x. Do we really want to pull in #7501 for 9.1.x ?

duke8253:
#7501 sill has some performance improvements over the original versions actually, mainly creating a seperate arena for each thread, and it cleaned up some unnecessary code as well, so it might be a cleaner cherry pick with it.

zwoop:

PR #6609, opened 2020-04-02
Add hook for loading certificate and key data from plugin 

shinrich:
Note: I'm marking this for consideration for 9.0.  I am ok with it not going into the first version.  My organization needs this functionality, but we can patch it in for a bit.  Would much rather get some version of 9.0 released sooner.

This branch adds functionality for plugins to load certificate and key information on configuration load and reload.  The new plugin hook triggers for both client certificates and server certificates,

This branch adds a hook (TS_LIFESTYLE_SSL_SECRET_HOOK) and a pair of TS API's, TSSslSecretSet/Get.  The hook gets triggered before the SSL_CTX's are created.  The continuation is called with the name of the certificate or key as the data argument.  The plugin then can load the certificate/key information from some other source (e.g. a key server or just someplace else on disk as the ssl_secret_load_test.cc does in this branch).  The plugin calls TSSslSecretSet to update the secret table.  After the hooks are called, the core looks for the data in the secret_map.  If it is not present, the core loads the data from disk and puts it in the map.  The SSL_CTX creation logic was updated to operate on a mem_bio instead of using the file oriented forms of the cert and key loading calls.

For the client certs specified by conf_remap.so, the secrets are loaded as needed likely after the config load/reload phase.  Same plugin flow applies.

The branch adds a pre-step for the config load so the secrets are loaded before the regular SSL_CTX creation is performed.

The branch adds a secret_map which is effectively filled with the pem encoded certificates and keys.  This is an increase in memory foot print.  The core logic could clear the secret entry after creating the SSL_CTX, but of the same certificate/key is used for multiple SSL_CTX, this will result in extra calls to the plugin or disk.  For config load/reload we could just clear the secret map after the config load stage is done.  For conf_remap updates, the getCTX table should have the same SSL_CTX if the certificate/key is used multiple times, so freeing the entries after the individual load should be fine.

I will be sending a mail out on the dev list highlighting the API change and pointing to this PR.

After we get the client and server side plugin-based certificate loading working, the next step is allowing a plugin to push in changes without a full config reload.

Issue #4808, opened 2019-01-16
Bad SSL Certificate CN causes freeze


jhg03a:
#### Observed Behavior:
All traffic ceases to flow through ATS with the addition of an improperly crafted SSL wildcard certificate.

#### Expected Behavior:
ATS should fail to reload config and report an error

#### Steps to reproduce:
1. Create a wildcard SSL certificate with a CN like `CN=DNS:*.subdomain.domain.com` and SANs like `[DNS:*.subdomain.domain.com, DNS:subdomain.domain.com]`
2. Add certs and keys to disk like normal
3. Update ssl_multicert.config like `ssl_cert_name=edge_subdomain_domain_com_cert.cer ssl_key_name=edge.subdomain.domain.com.key`
4. Perform an ATS config reload
5. Use either openssl or curl to try and send traffic

#### Found in:
trafficserver-7.1.4_rc0

bryancall:
This is resolved in the current 10.1 release (10.1.2). A malformed certificate is now handled gracefully: cert-load validation fails the entry, and on a config reload the new (invalid) config is discarded and the previous config retained with a load-failure error logged, rather than freezing traffic. The Server Name Indication (SNI) lookup is also constant-time with no recursion, so a bad wildcard name can't hang request processing. Closing as fixed; please reopen with a reproduction if you see a freeze on 10.1.2 or later.

Issue #3931, opened 2018-07-06
Certificates not reloaded unless ssl_multicert.config is touched


d2r:
Certificates that are pointed to by `ssl_multicert.config` can change, but ATS only reloads them if the timestamp on `ssl_multicert.config` is updated, which may not change as certificates do.

The current work-around is to update the timestamp via a tool like `touch` to trigger reloading of the current certificates.

It would be nice to reload the config if the certificate files are changed, since it is confusing to admins who expect certificates to be reloaded while the server is up.

Related to #1672

zizhong:
I remember I had a patch to address this. Not sure why it's not functioning. You can check https://github.com/apache/trafficserver/pull/301/files.
I'd take a look when I got time.

d2r:
This is still a valid issue.

shinrich:
I'll take a look at @zizhong's logic.  I've been through there recently.  Hopefully we just need to tweak something small to get the watching signaling propagating correctly.  

Now that we are also referencing and reloading client cert and key files in ssl_server_name.yaml, would be good to use a similar technique there.

bryancall:
The change went into commit: 4d2c262c3be3bbcbfce3214826ecb16dee2cd027

shinrich:
I verified that you no longer have to touch the ssl_multicert.config file if a referenced file is updated.  Getting ready to add an autest to exercise that case.  I thought it went in with my recent PR #6609, but looking at @zizhong's commit, all my PR did was rearrange what he had already added, so the fix is likely in whatever versions include   https://github.com/apache/trafficserver/commit/4d2c262c3be3bbcbfce3214826ecb16dee2cd027

PR #8030, opened 2021-07-01
Fix server cert reload

shinrich:
Noticed this while trying to reproduce the situation described in issue #4808.  Updating a cert file in place that is referenced in the ssl_multi_cert.config file was doing nothing.  The Secret file was not being cleared during the reload before the server cert reloading would occur.  I believe this problem was introduced by PR #6609.  I am surprised we haven't been hit by this.  We must either be running our cert update plugin or restarting the process on each server cert update.

I added a test to exercise config reload with an updated server certificate file.  We already had such test for the client cert update.  

While exercising this I noticed that you no longer need to touch the ssl_multi_cert.config file when only the referenced cert and key files are updated as described in issue #3931.  I think that has been fixed for a while.  The commit from @zizhong mentioned in the issue seems to be the logic that fixed it.

### DIFF
--- a/iocore/net/SSLClientCoordinator.cc
+++ b/iocore/net/SSLClientCoordinator.cc
@@ -35,6 +35,7 @@ SSLClientCoordinator::reconfigure()
   // into it.
   SSLConfig::reconfigure();
   SNIConfig::reconfigure();
+  SSLCertificateConfig::reconfigure();
 }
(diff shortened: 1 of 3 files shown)


shinrich:
[approve ci autest]

randall (approved):

PR #8032, opened 2021-07-01
Stablize trace curl test in good_request_after_bad

shinrich:
I noticed that "Trace request with a chunked body via curl" test in good_request_after_bad fails with some frequency.  It may just be an oddity of the curl output where short output is not always displayed if it is closely followed by a FIN.  Or there may be a real problem where ATS is cutting off the body.  This PR attempts to write the output to a file using the -o option and test that.

Of course it always works for me locally.

### DIFF
--- a/tests/gold_tests/headers/gold/bad_good_request_header.gold
+++ b/tests/gold_tests/headers/gold/bad_good_request_header.gold
@@ -1,0 +1,5 @@
+``HTTP/1.1 400 Invalid HTTP Request
+``Connection: close
+``Server: ATS/``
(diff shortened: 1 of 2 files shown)


shinrich:
[approve ci autest]

maskit (approved):

shinrich:
[approve ci autest]

shinrich:
[approve ci autest]

shinrich:
Kicked the CI a few times.  The good_request_after_bad test succeeded every time.

PR #8033, opened 2021-07-02
Treat Sphinx doc build warnings as errors.

bneradt:
It's easy to introduce warnings in the doc build without noticing them
due to the amount of output from the build. This changes the
sphinx-build invocation to treat any warnings we introduce as errors so
that such problems will get caught.

It also fixes a handful of warnings that got introduced recently as
well. With this patch, the build runs cleanly without warnings.

### DIFF
--- a/doc/developer-guide/api/functions/TSSslSecret.en.rst
+++ b/doc/developer-guide/api/functions/TSSslSecret.en.rst
@@ -34,7 +34,7 @@ Description
 
 :func:`TSSslSecretSet` updates the current secret map. Generally the secret name corresponds to the name of a certificate or a key.
 Future creation of SSL_CTX objects that use the secret will use the newly specified data. It can be useful to call this function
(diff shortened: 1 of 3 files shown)


doc/developer-guide/api/functions/TSSslSecret.en.rst
```diff
@@ -67,7 +67,7 @@ Synopsis
 
     #include <ts/ts.h>
 
-.. function:: TSReturnCode TSSslSecretGet(const char * secret_name, int secret_name_length)
+.. function:: TSReturnCode TSSslSecretUpdate(const char * secret_name, int secret_name_length)
```
shinrich, line 70: Ah, good catch of a cut-n-paste error.

shinrich:
Looks good to me!  Thanks for cleaning it up.

randall (approved):

PR #8035, opened 2021-07-02
Do not merge: Test CI doc build

bneradt:
Intentionally introducing a warning back in to verify that our PR CI doc
build fails if warnings are introduced.


bneradt:
Cool: the docs build failed due to the Sphinx warning introduced in this patch, as desired.

PR #8040, opened 2021-07-03
Fix uri_signing unit test for out of source builds

bneradt:
The URI signing plugin unit test accesses a config file in the source
directory. This works fine for in-source builds, but did not for out of
source. This updates the Makefile and unit test so that the file is
accessible for out of source builds.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1410,7 +1410,7 @@ TS_CHECK_BROTLI
 TS_CHECK_LUAJIT
 
 #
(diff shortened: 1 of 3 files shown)


bneradt:
This is a build fix for a unit test and is therefore safe for all branches. And since this fixes out of source builds, we should port this to all our supported branches for users who want or require out of source builds.

randall (approved):

zwoop:

PR #8041, opened 2021-07-03
Fix uri_signing unit test for out of source builds

bneradt:
The URI signing plugin unit test accesses a config file in the source
directory. This works fine for in-source builds, but did not for out of
source. This updates the Makefile and unit test so that the file is
accessible for out of source builds.


ywkaras:
[approve ci Fedora]

ezelkow1 (approved):

PR #8058, opened 2021-07-13
Replace assert with warning in error event processing

shinrich:
Fixes an issue introduced in PR #7809.

Replaced the ink_release_assert with a warning so we can gain information about this situation.


shinrich:
We have been running this in our production.  Vinith looked up the messages we received.  50 over our footprint in a 4 hour period.  Most of them are just EPOLLERR (0x8).  A least one also had EPOLLHUP (0x18).  May be ok to just ignore them.  Or to pass them along via a write event as this PR does.

jrushford (commented):
Looks good to me, I'm going to deploy this change in our 9.1.x testing environment where we've been hitting the assertion that is being replaced here.

jrushford (approved):
Looks good.

zwoop:

PR #8064, opened 2021-07-14
Remove check_system_constants(), which is unused

nekomachi-touge:
`check_system_constants()` on `src/traffic_server/traffic_server.cc` does nothing and is referenced only by `main()` .
I found nothing on GitHub/JIRA/web, so I believe the original intention of this function is lost and the code is safe to remove.


randall:
[approve ci]

bryancall (approved):

nekomachi-touge:
Thank you for merging this so quickly!

PR #8074, opened 2021-07-17
Make the InkAPI mHandleAllocator be a Proxy Allocator

zwoop:
Running a test with xdebug.so and header_rewrite.so with rules to both trigger xdebug and read and update a few headers via header_rewrite, I see about a 30% increase on RPS with this, and a 50% reduction in response latency. It's also using a noticeably less amount of CPU even with those RPS increases (likely we can see even more RPS increase if I had clients that could push the servers).

### DIFF
--- a/src/traffic_server/InkAPI.cc
+++ b/src/traffic_server/InkAPI.cc
@@ -773,7 +773,7 @@ isWriteable(TSMBuffer bufp)
 static MIMEFieldSDKHandle *
 sdk_alloc_field_handle(TSMBuffer /* bufp ATS_UNUSED */, MIMEHdrImpl *mh)
 {
-  MIMEFieldSDKHandle *handle = mHandleAllocator.alloc();
+  MIMEFieldSDKHandle *handle = THREAD_ALLOC(mHandleAllocator, this_thread());
 
   // TODO: Should remove this when memory allocation can't fail.
   sdk_assert(sdk_sanity_check_null_ptr((void *)handle) == TS_SUCCESS);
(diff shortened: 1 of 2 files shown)


src/traffic_server/InkAPI.cc
```diff
@@ -773,7 +773,7 @@ isWriteable(TSMBuffer bufp)
 static MIMEFieldSDKHandle *
 sdk_alloc_field_handle(TSMBuffer /* bufp ATS_UNUSED */, MIMEHdrImpl *mh)
 {
-  MIMEFieldSDKHandle *handle = mHandleAllocator.alloc();
+  MIMEFieldSDKHandle *handle = THREAD_ALLOC(mHandleAllocator, this_thread());
```
masaori335, line 776: Do we need to get rid of `ClassAllocator<MIMEFieldSDKHandle> mHandleAllocator`?
https://github.com/apache/trafficserver/blob/c4b7eaac6ec3e98702b205895f914f2dcec76369/src/traffic_server/InkAPI.cc#L395

masaori335: NVM, they need to be the same name.

https://github.com/apache/trafficserver/blob/c4b7eaac6ec3e98702b205895f914f2dcec76369/iocore/eventsystem/I_ProxyAllocator.h#L74-L75

masaori335:
@c-taylor also measured this in the same way as #8052. This improves 50 - 60 %!

Test | 8.1.2 | #8052 | #8074
----|----|----|----
KA|	     	112400 | 184900 | 172740
KA-tls|		114700 | 196700 | 184671
KA-tls-h2|	116300 | 178000 | 174117

The size of this change is obviously small and no confusion around TS APIs. IMO, we should land and backport this change regardless of what we'll do for TS APIs (#8052, #8069, or whatever).

shinrich (approved):

zwoop:
Cherry-picked to v9.1.x branch after review with Bryan.

PR #8075, opened 2021-07-17
Add plugin parent_select reloading

rob05c:
Adds the parent_select remap plugin calling `TSMgmtConfigFileAdd` with the strategy file name, to cause remap.config to be reloaded when the strategy file has changed.


bryancall:
@ywkaras is going to look at this

ywkaras (approved):

PR #8082, opened 2021-07-19
Changes HttpSM to be Proxy Allocated

zwoop:
A few comments on this:

1. I couldn't get the proxy allocation macros to work inside the HttpSM::allocate. But, this patch looks more like what we do for other allocators, making it explicit where we need to allocate a new HttpSM. If someone feels strongly that HttpSM::allocate has to survive, I can see if there's a way to make it work.
2. Bryan requested that I verify that the allocation of the HttpSM happens on the ET_NET thread even with accept threads, and I did verify that and it does. So the HttpSM gets allocated and released on the same ET_NET thread.

### DIFF
--- a/iocore/eventsystem/I_Thread.h
+++ b/iocore/eventsystem/I_Thread.h
@@ -122,6 +122,7 @@ public:
   ProxyAllocator http1ClientSessionAllocator;
   ProxyAllocator http2ClientSessionAllocator;
   ProxyAllocator http2StreamAllocator;
+  ProxyAllocator httpSMAllocator;
   ProxyAllocator quicClientSessionAllocator;
   ProxyAllocator httpServerSessionAllocator;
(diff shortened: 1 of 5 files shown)


bryancall (approved):

bryancall:
@zwoop can you post the performance numbers of having a proxy allocator for the HttpSM vs not?

zwoop:
I've tried this on master, 9.1.x and 9.0.x branches, and on the 9.0.x branch, it will assert always. 9.1.x or master does not have this problem, I think something changed around either the class allocators or the constructor / destructors for the HttpSM.

zwoop:
The performance increase is not tremendous, about 5%. It's possible it make more of a difference on other hardware platforms, where the costs of the atomic operations on the class allocator causes more contention.

zwoop:

PR #8083, opened 2021-07-19
Cancel active timeout when releasing session

### LINKED ISSUE #8079: Assert due to active timeout on server session approaching SessionManager
In our environment, we started seeing crashes with this stack.

```
#0  0x00002af8ef890387 in __GI_raise (sig=sig@entry=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:55
#1  0x00002af8ef891a78 in __GI_abort () at abort.c:90
#2  0x00002af8ecdeb63b in ink_abort (message_format=message_format@entry=0x2af8ece0dc47 "%s:%d: failed assertion `%s`") at ../../../../../../_scm/trafficserver9.1/src/tscore/ink_error.cc:99
#3  0x00002af8ecde7bb5 in _ink_assert (expression=expression@entry=0x7e4b6a "0", file=file@entry=0x7f9c50 "../../../../../../_scm/trafficserver9.1/proxy/http/HttpSM.cc", line=line@entry=1418)
    at ../../../../../../_scm/trafficserver9.1/src/tscore/ink_assert.cc:37
#4  0x000000000056d96d in HttpSM::state_common_wait_for_transform_read(HttpTransformInfo*, int (HttpSM::*)(int, void*), int, void*) () at ../../../../../../_scm/trafficserver9.1/proxy/http/HttpSM.cc:1418
#5  0x000000000056da74 in HttpSM::state_response_wait_for_transform_read(int, void*) () at ../../../../../../_scm/trafficserver9.1/proxy/http/HttpSM.cc:1329
#6  0x000000000057e73b in HttpSM::main_handler(int, void*) () at ../../../../../../_scm/trafficserver9.1/proxy/http/HttpSM.cc:2742
#7  0x000000000079840e in read_signal_and_update(int, UnixNetVConnection*) () at ../../../../../../_scm/trafficserver9.1/iocore/net/UnixNetVConnection.cc:83
#8  0x00000000007988bb in UnixNetVConnection::mainEvent(int, Event*) () at ../../../../../../_scm/trafficserver9.1/iocore/net/UnixNetVConnection.cc:1160
#9  0x000000000078d47b in InactivityCop::check_inactivity (this=0x2af90a8cb000, event=<optimized out>, e=<optimized out>) at ../../../../../../_scm/trafficserver9.1/iocore/net/UnixNet.cc:100
#10 0x00000000007da0dd in handleEvent (data=0x2af90a800060, event=2, this=0x2af90a8cb000) at ../../../../../../_scm/trafficserver9.1/iocore/eventsystem/I_Continuation.h:219
#11 handleEvent (data=0x2af90a800060, event=2, this=0x2af90a8cb000) at ../../../../../../_scm/trafficserver9.1/iocore/eventsystem/I_Continuation.h:215
#12 EThread::process_event(Event*, int) () at ../../../../../../_scm/trafficserver9.1/iocore/eventsystem/UnixEThread.cc:164
#13 0x00000000007dae04 in EThread::execute_regular (this=this@entry=0x2af8f53b0e00) at ../../../../../../_scm/trafficserver9.1/iocore/eventsystem/UnixEThread.cc:273
#14 0x00000000007db182 in execute (this=0x2af8f53b0e00) at ../../../../../../_scm/trafficserver9.1/iocore/eventsystem/UnixEThread.cc:364
#15 EThread::execute (this=0x2af8f53b0e00) at ../../../../../../_scm/trafficserver9.1/iocore/eventsystem/UnixEThread.cc:342
#16 0x00000000007d97ca in spawn_thread_internal (a=0x2af8f08e0d40) at ../../../../../../_scm/trafficserver9.1/iocore/eventsystem/Thread.cc:92
#17 0x00002af8eec22ea5 in start_thread (arg=0x2af8fc703700) at pthread_create.c:307
#18 0x00002af8ef9588dd in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111
```

The end of the HttpSM history contains the following
```
{location = {file = 0x7f9c50 "../../../../../../_scm/trafficserver9.1/proxy/http/HttpSM.cc", 
        func = 0x7fd680 <HttpSM::setup_server_transfer_to_transform()::__FUNCTION__> "setup_server_transfer_to_transform", line = 6666}, event = 34463, reentrancy = 6}, {location = {
        file = 0x7f9c50 "../../../../../../_scm/trafficserver9.1/proxy/http/HttpSM.cc", 
        func = 0x7fdaf0 <HttpSM::tunnel_handler_transform_write(int, HttpTunnelConsumer*)::__FUNCTION__> "tunnel_handler_transform_write", line = 3954}, event = 103, reentrancy = 6}, {location = {
        file = 0x7f9c50 "../../../../../../_scm/trafficserver9.1/proxy/http/HttpSM.cc", func = 0x7fdc80 <HttpSM::tunnel_handler_server(int, HttpTunnelProducer*)::__FUNCTION__> "tunnel_handler_server", 
        line = 3063}, event = 102, reentrancy = 6}, {location = {file = 0x7f9c50 "../../../../../../_scm/trafficserver9.1/proxy/http/HttpSM.cc", 
        func = 0x7fe000 <HttpSM::state_response_wait_for_transform_read(int, void*)::__FUNCTION__> "state_response_wait_for_transform_read", line = 1314}, event = 2301, reentrancy = 7}, {location = {
        file = 0x7f9c50 "../../../../../../_scm/trafficserver9.1/proxy/http/HttpSM.cc", 
        func = 0x7fdfc0 <HttpSM::state_common_wait_for_transform_read(HttpTransformInfo*, int (HttpSM::*)(int, void*), int, void*)::__FUNCTION__> "state_common_wait_for_transform_read", line = 1344}, 
      event = 2301, reentrancy = 7}, {location = {file = 0x7f9c50 "../../../../../../_scm/trafficserver9.1/proxy/http/HttpSM.cc", 
        func = 0x7fe000 <HttpSM::state_response_wait_for_transform_read(int, void*)::__FUNCTION__> "state_response_wait_for_transform_read", line = 1314}, event = 106, reentrancy = 1}, {location = {
        file = 0x7f9c50 "../../../../../../_scm/trafficserver9.1/proxy/http/HttpSM.cc", 
        func = 0x7fdfc0 <HttpSM::state_common_wait_for_transform_read(HttpTransformInfo*, int (HttpSM::*)(int, void*), int, void*)::__FUNCTION__> "state_common_wait_for_transform_read", line = 1344}, 
      event = 106, reentrancy = 1}, 
```

You see that the tunnel was set up.  The consumer got a write complete, the producer got a read complete and tunnel done was sent to the final handler.  Then inactivity timeout (106) is sent to the default handler.  In our environment the outbound active timeout is set to 14 minutes.

Digging into the core a bit more, we see that the problem requests were GET requests.  The respond body is 0 bytes long (correctly reported by the content-length).  In our environment a transform plugin (for most of the cases this was txn_box) was installed to adjust the response being sent to the user agent.  But from the history, the TRANSFORM_READ_READY is never sent.  Instead things timeout and the HttpSM is in a bad state for the particularly handler left.  (I will file a separate issue about not handling a 0 length body for a write transform).

@SolidWallOfCode was able to reproduce this in a txn_box autest.  We will try to get a test together to exercise this scenario using one of the transforms in the trafficserver source tree (null or buffererd null).

We were able to get around the assert crash by clearing the active timeout when ProxyTransaction::release is called on server_txn.  I think this issue arose when PR #7849.  We are running with that change in our 9.1.x branch.  With that change, the server_txn is not immediately delivered to the HttpSessionManager when release() is called.  Rather the actual transfer of control is delayed until the HttpSM is deleted.

The change is to apply the keepalive timeout and clear the active timeout when release() is called.  The active timeout makes no sense when the server session is in the keep alive pool.

shinrich:
More details are in the issue.  This fix cancels the active timeout when the server session is released.  It may be a while still before the HttpSM is killed and the server session is actually put in the server session pool as discussed in the issue.

This PR also partially undoes the change in PR #7618.  I actually reviewed and approved that one, but on looking at it more closely while debugging this issue, I think the change is faulty since it will lose any overriden value associated with the Transaction, making the overridable nature of the keep alive timeout ineffective.

This change fixed our TxnBox-based reproducing test.  I am getting ready to set this up for testing in our 9.1 based production build.

I'll try to set up a reproducing test in another PR that uses a transform plugin that is already in the trafficserver code base.

This closes #8079

### DIFF
--- a/proxy/http/Http1ServerSession.cc
+++ b/proxy/http/Http1ServerSession.cc
@@ -146,6 +146,11 @@ Http1ServerSession::release(ProxyTransaction *trans)
   Debug("http_ss", "[%" PRId64 "] Releasing session, private_session=%d, sharing_match=%d", con_id, this->is_private(),
         sharing_match);
   if (state == SSN_IN_USE) {
+    // The caller should have already set the inactive timeout to the keep alive timeout
+    // Unfortunately, we do not have access to that value from here.
+    // However we can clear the active timeout here.  The active timeout makes no sense
+    // in the keep alive state
+    cancel_active_timeout();
     state = SSN_TO_RELEASE;
     return;
   }
--- a/proxy/http/HttpSessionManager.cc
+++ b/proxy/http/HttpSessionManager.cc
@@ -211,11 +211,9 @@ ServerSessionPool::releaseSession(PoolableSession *ss)
   // Transfer control of the write side as well
   ss->do_io_write(this, 0, nullptr);
 
-  // we probably don't need the active timeout set, but will leave it for now
-  HttpConfigParams *http_config_params = HttpConfig::acquire();
-  ss->set_inactivity_timeout(HRTIME_SECONDS(http_config_params->oride.keep_alive_no_activity_timeout_out));
-  ss->set_active_timeout(HRTIME_SECONDS(http_config_params->oride.keep_alive_no_activity_timeout_out));
-  HttpConfig::release(http_config_params);
+  ss->set_inactivity_timeout(ss->get_netvc()->get_inactivity_timeout());
+  ss->cancel_active_timeout();
+
   // put it in the pools.
   this->addSession(ss);
 
@@ -251,9 +249,8 @@ ServerSessionPool::eventHandler(int event, void *data)
     return 0;
   }
 
-  sockaddr const *addr                 = net_vc->get_remote_addr();
-  HttpConfigParams *http_config_params = HttpConfig::acquire();
-  bool found                           = false;
+  sockaddr const *addr = net_vc->get_remote_addr();
+  bool found           = false;
 
   for (auto spot = m_ip_pool.find(addr); spot != m_ip_pool.end() && spot->_ip_link.equal(addr, spot); ++spot) {
     if ((s = spot)->get_netvc() == net_vc) {
@@ -271,8 +268,7 @@ ServerSessionPool::eventHandler(int event, void *data)
                 "[%" PRId64 "] [session_bucket] session received io notice [%s], "
                 "resetting timeout to maintain minimum number of connections",
                 s->connection_id(), HttpDebugNames::get_event_name(event));
-          s->get_netvc()->set_inactivity_timeout(HRTIME_SECONDS(http_config_params->oride.keep_alive_no_activity_timeout_out));
-          s->get_netvc()->set_active_timeout(HRTIME_SECONDS(http_config_params->oride.keep_alive_no_activity_timeout_out));
+          s->get_netvc()->set_inactivity_timeout(s->get_netvc()->get_inactivity_timeout());
           found = true;
           break;
         }
@@ -292,7 +288,6 @@ ServerSessionPool::eventHandler(int event, void *data)
     }
   }
 
-  HttpConfig::release(http_config_params);
   if (!found) {
     // We failed to find our session.  This can only be the result of a programming flaw. Since we only ever keep
     // UnixNetVConnections and SSLNetVConnections in the session pool, the dynamic cast won't fail.


bryancall:
@vmamidi is going to take a look

maskit:
I don't understand why overridden timeout values have to be retained when a session is returned to the session pool. Retaining the values means you will acquire a session that has unpredictable timeout values which depend on previous transactions.

maskit:
And like I said on #7618, a session could be kept in the pool forever if inactive timeout was set to 0.

shinrich:
This issue is for proxy.config.http.keep_alive_no_activity_timeout_out.  The documentation identifies this setting as overridable.  So we need to set the timeout to that transaction-specific value before it is added to the session pool.  Otherwise the setting is effectively not overrideable, and we should update the documentation.

https://docs.trafficserver.apache.org/en/latest/admin-guide/files/records.config.en.html?highlight=keepalive#proxy-config-http-keep-alive-no-activity-timeout-out

shinrich:
The inactivity timeout is set before returning to the pool.  Within the return to pool logic, we could check if the inactivity timeout is 0 and set it from the global setting.  This would protect us from future errors where the caller does not set the inactivity timeout.

maskit:
I guess this confusion comes from having multiple values for the timeout. We have a) the original value in records.config, b) new value set by a transaction, and c) the effective value in a session instance. And c) can be set to 0 if you call cancel_timeout() or something.

Before my change on #7618, the code seemed like to reset c) with c). It didn't make sense to me, so I changed the code to reset c) with a). If we want to override the timeout, what we really need to do is probably to reset c) with b). Does it make sense?

shinrich:
Right.  Before your change, the code was setting C back to C.  I assume in an attempt to ensure the full inactivity timeout was applied during the keep-alive period.  But before calling ServerSessionPool::releaseSession, the caller had set the inactivity timeout to B (the override version of the keepalive timeout).

Unfortunately, by the time we get to ServerSessionPool::releaseSession, the code does not have direct access to B.  No access to the state machine.  We could update the API, so B is passed along the release call chain, so it can be set in ServerSessionPool::releaseSession.

Ultimately, the keep-alive timeout should really be overridable at the session level.  But unfortunately, our existing override mechanism is centered around the HttpSM.  We have talked in various meetings about making session more first class objects, but I'm afraid that is a bigger decision effort that this PR.

maskit:
Fair enough. I still think resetting the value with C is incorrect, but resetting the value with A is also incorrect and sounds like C has a bit more practical value.

maskit (approved):

PR #8084, opened 2021-07-19
Defer canceling UA inactivity timeout for chunked bodies too

shinrich:
I noticed this in the code when working on PR #8083.  When deciding when to cancel the user agent inactivity timeout, the code was using request_content_length >= 0 to determine whether there was a body associated with the request.  This logic will miss the cases when the body is chunked encoded.  So for the chunked encoded body the UA inactivity timeout will be cancelled too soon.


bryancall:
@randall volunteered to look at this

randall (approved):

Issue #7196, opened 2020-09-16
stms log field can have bad values.


traeak:
Observed in ats80x and ats81x, master.

We tried to enable the stms field in our logs.  We started observing invalid values in some circumstances.

When the parent isn't visited the value is -1.  This isn't a valid value for a time field, it should be '0' since '0' time was spent contacting the origin.  stmsf in this case is set to '-.001', sts to '0' and stmsh to 'ffff...'

In the case an intercept plugin is in play (like the stats_over_http or slice plugins), the stms value is set to utc milliseconds (ie: 1600263600889).  I speculate that the stms field is initialized to some start time but never resolved with an end time later in the transaction.

mtorluemke:
+1 for getting this fix in -- seeing this issue in production as well.

PR #7259, opened 2020-10-08
detect and fix stms log field having epoch time ms for intercept plugins.

traeak:
For intercept plugins like stats_over_http and slice the stms logging field may end up being set to epoch time in ms.
This fix detects that condition inside HttpSM by only allowing TS_MILESTONE_SERVER_CLOSE to be set if 
TS_MILESTONE_SERVER_CONNECT was already set.

LogAccess::marshal_server_resp_time_ms uses this code to calculate this stms field.
```
m_http_sm->milestones.difference_msec(TS_MILESTONE_SERVER_CONNECT, TS_MILESTONE_SERVER_CLOSE)
```

This is a mitigation in the code.

resolves #7196

PR #8086, opened 2021-07-19
Set milestone for first connect before returning in the PluginVC case.

elsloo:
* Relocates the server first connect milestone to set it prior to connecting the plugin VC tunnel and returning
* Appears to be the fundamental cause for closed issue #7196
* This should negate the need for PR #7259, however, we may want to leave it in place for defensive purposes


shinrich (approved):

ezelkow1:

PR #8094, opened 2021-07-20
Remove unused OOB network logic

shinrich:
Another observation from the code coverage report.  This PR removes code that is only within methods with OOB in the name with two exceptions.  The free methods for SSLNetVConnection and UnixNetVConnection call cancelOOB.  But I saw nowhere that started an OOB operation.  Even one of the comments on send_OOB questioned the need for its existence.


randall (approved):

bneradt (commented):
Looks good.

PR #8096, opened 2021-07-20
Fail requests with incomplete line ends

shinrich:
To be RFC compliant, trafficserver should reject requests that use \n instead of \r\n.  This required updating a number of our test files.


shinrich:
[approve ci]

bneradt:
[approve ci clang-format]

bneradt:
[approve ci clang-format]

shinrich:
Lots of files changed, but only one file is core. One file for unit tests.  And the rest are updates where are autests were relying on newline only line endings.

maskit:
This may be slightly different from the original issue, but would you add these tests and update the parser while you are at it?

```
{"GET /index.html HTTP/1.0\r\na\rb\n", PARSE_RESULT_ERROR, 30},
{"GET /index.html HTTP/1.0\r\nUser-Agent: foobar\r \n", PARSE_RESULT_ERROR, 47},
```

Also, I noticed that the test case below is untestable because it causes a crash, regardless of this change. It's probably because the request string is static. I couldn't reproduce the crash if I send the request over network (to traffic_server by nc command).
```
{"GET /index.html HTTP/1.0\r\nUser-Agent: foobar\r\n \r\n", PARSE_RESULT_ERROR, 49},
```

shinrich:
Pushed new commit adding the tests @maskit.  Plus fixing the issues these tests introduced.

maskit:
[approve ci Clang-Analyzer]
[approve ci autest]

maskit:
It seems like Autest fails due to file name mismatch. "test-0rtt-s_client.py" probably need to be updated.

shinrich:
[approve ci autest]

maskit (approved):

zwoop:
I'm going to remove this from 9.1.x project, since we have a better patch soon to replace this.

PR #8098, opened 2021-07-20
Fixes an issue in ParentSelection and NextHop strategies where a down parent may not retried

jrushford:
Fixes and issue in ParentSelection and Nexthop strategies introduced with #7744 where a down parent may not be retried due to retriers limiting.

### DIFF
--- a/proxy/http/remap/NextHopConsistentHash.cc
+++ b/proxy/http/remap/NextHopConsistentHash.cc
@@ -326,6 +326,7 @@ NextHopConsistentHash::findNextHop(TSHttpTxn txnp, void *ih, time_t now)
         // check if the host is retryable.  It's retryable if the retry window has elapsed
         _now == 0 ? _now = time(nullptr) : _now = now;
         if ((pRec->failedAt.load() + retry_time) < static_cast<unsigned>(_now)) {
+          NH_Debug(NH_DEBUG_TAG, "[%" PRIu64 "] next hop %s, retriers: %d", sm_id, pRec->hostname.c_str(), pRec->retriers.load());
           if (pRec->retriers.fetch_add(1, std::memory_order_relaxed) < max_retriers) {
             nextHopRetry        = true;
             result->last_parent = pRec->host_index;
--- a/proxy/http/remap/NextHopHealthStatus.cc
+++ b/proxy/http/remap/NextHopHealthStatus.cc
@@ -149,3 +149,29 @@ NextHopHealthStatus::markNextHop(TSHttpTxn txn, const char *hostname, const int 
     break;
   }
 }
+
+void
+NextHopHealthStatus::retryComplete(TSHttpTxn txn, const char *hostname, const int port)
+{
+  HttpSM *sm          = reinterpret_cast<HttpSM *>(txn);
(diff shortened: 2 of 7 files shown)


proxy/http/remap/NextHopHealthStatus.cc
```diff
@@ -149,3 +149,29 @@ NextHopHealthStatus::markNextHop(TSHttpTxn txn, const char *hostname, const int
...
+  }
+
+  const std::string host_port = HostRecord::makeHostPort(hostname, port);
+  auto iter                   = host_map.find(host_port);
+  if (iter == host_map.end()) {
+    NH_Debug(NH_DEBUG_TAG, "[%" PRId64 "] no host named %s found in host_map", sm_id, host_port.c_str());
+    return;
+  }
+
+  std::shared_ptr h = iter->second;
```
ywkaras, line 172: Nitpick:  I would make this `auto &retriers = iter->second->retriers;` to avoid creating a duplicate shared pointer and atomic increment/decrement of the reference count.

ywkaras:
I'm thinking we should make sure there are no race conditions when accessing the retriers count first:  https://github.com/apache/trafficserver/pull/8103

proxy/http/remap/NextHopConsistentHash.cc
```diff
@@ -326,6 +326,7 @@ NextHopConsistentHash::findNextHop(TSHttpTxn txnp, void *ih, time_t now)
         // check if the host is retryable.  It's retryable if the retry window has elapsed
         _now == 0 ? _now = time(nullptr) : _now = now;
         if ((pRec->failedAt.load() + retry_time) < static_cast<unsigned>(_now)) {
+          NH_Debug(NH_DEBUG_TAG, "[%" PRIu64 "] next hop %s, retriers: %d", sm_id, pRec->hostname.c_str(), pRec->retriers.load());
```
ywkaras, line 329: For example suppose retriers == max_retriers - 1.  Then:
1.  one core executes line 330, making retriers == max_retriers, so it goes to the else clause.
2. another core executes line 330, making retriers == max_retriers + 1, so it also goes to the else clause.
Both cores will then execute line 341, making retriers == max_retriers - 1 again.  But neither one acted as a retrier.

jrushford: I think that was the idea with the else clause @ywkaras, to negate the increment on line 330 if they aren't retriers.  Am I missing something here?

jrushford: With that said, I like how you've abstracted the retriers in the other PR.

jrushford: That was the idea of the else clause, to back out the increment if the transaction on that core wasn't going to be a retrier.  Am I missing something @ywkaras?  I do like how you've abstracted this in the other PR though.

ywkaras: Is there some way you guarantee the thread interleaving scenario I describe above can not happen?

jrushford:
Closing this PR in favor of #8159

PR #8106, opened 2021-07-21
Make the rest of InkAPI allocators Proxy Allocated

zwoop:


randall:
[approve ci autest]

zwoop:
[approve ci autest]

ywkaras:
The test_hooks Au test is causing a segment violation in TSContDestroy():  https://ci.trafficserver.apache.org/autest/ausb-8106.16267/test_hooks/_output/test_hooks-ts/stream.stderr.txt

I think what's happening is that INKContAllocator or it's proxy allocator is getting destroyed before `cleanup`:
https://github.com/apache/trafficserver/blob/c214c9f40096afbe580c0ee90647933ad8b21bcd/tests/tools/plugins/test_hooks.cc#L371

I think you should just delete the lines https://github.com/apache/trafficserver/blob/c214c9f40096afbe580c0ee90647933ad8b21bcd/tests/tools/plugins/test_hooks.cc#L348-L372 .

There are tricky ways to control the order of destruction at program exit, but I don't think it's worth it.  The only problem is that it will probably look like a memory leak to ASAN.

ywkaras (changes requested):
Delete the indicated lines from test_tsapi.cc.

bryancall:
[approve ci autest]

ywkaras:
If this ever passes Au test regression without changing test_tsapi.cc, it will be a fluke, it will be flakey forever more.

ywkaras:
[approve ci autest]

ywkaras (approved):

zwoop:

PR #8159, opened 2021-07-22
Fixes an issue in ParentSelection and NextHop strategies where a down parent may not be retried

ywkaras:
Alternative to https://github.com/apache/trafficserver/pull/8098

Fixes and issue in ParentSelection and Nexthop strategies introduced with #7744 where a down parent may not be retried due to retriers limiting.

Avoids race conditions in incrementing, decrementing and clearing retriers count.


jrushford (approved):
Looks good @ywkaras.  I ran this patch with production traffic and found no issues.

PR #8162, opened 2021-07-22
HostDBProcessor: ensure hostname_offset is initialized to '0' to indicate null hostname

traeak:
In the new CI there is a periodic crash while running the HostDBProcessor regression test.

Running valgrind I see a complaints about
"Conditional jump or move depends on uninitialised value" during the regression test run.

One run pointed out
hostname_offset being uninitialized in the following call for the hostname_offset:

```
char *
HostDBInfo::perm_hostname() const
{
  if (hostname_offset == 0) {
    return nullptr;
  }

  return (char *)this + hostname_offset;
}
```

Setting initializer values for these data members in HostDBProcessor cleans up the valgrind messages.

### DIFF
--- a/iocore/hostdb/HostDB.cc
+++ b/iocore/hostdb/HostDB.cc
@@ -2257,7 +2257,11 @@ struct HostDBRegressionContinuation : public Continuation {
       // since this is a lookup done, data is either hostdbInfo or nullptr
       if (r) {
         rprintf(test, "hostdbinfo r=%x\n", r);
-        rprintf(test, "hostdbinfo hostname=%s\n", r->perm_hostname());
+        char const *hname = r->perm_hostname();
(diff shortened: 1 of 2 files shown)


shinrich (approved):
Seems reasonable

zwoop:
Hmmm, this may be an 8.x and 9.0.x candidate too? Non-terminated strings sounds scary :).

zwoop:

traeak:
> Hmmm, this may be an 8.x and 9.0.x candidate too? Non-terminated strings sounds scary :).

Yes it looks like it should be.

zwoop:
@traeak This has a merge conflict in 9.0.x, can you please make a back port PR for it please?

ezelkow1:

PR #8165, opened 2021-07-22
WIP Adds new API: TSSslSNIGet

randall:
converts rate_limit to use this API instead of calling OpenSSL APIs,
allowing this plugin to work with BoringSSL

### DIFF
--- a/src/traffic_server/InkAPI.cc
+++ b/src/traffic_server/InkAPI.cc
@@ -9306,6 +9306,26 @@ TSVConnSslConnectionGet(TSVConn sslp)
   return ssl;
 }
 
+const char *
+TSSslSNIGet(TSVConn sslp, int *length)
+{
+  char const *server_name   = nullptr;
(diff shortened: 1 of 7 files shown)


src/traffic_server/InkAPI.cc
```diff
@@ -9306,6 +9306,26 @@ TSVConnSslConnectionGet(TSVConn sslp)
   return ssl;
 }
 
+const char *
+TSSslSNIGet(TSVConn sslp, int *length)
+{
+  char const *server_name   = nullptr;
+  NetVConnection *vc        = reinterpret_cast<NetVConnection *>(sslp);
+  SSLNetVConnection *ssl_vc = dynamic_cast<SSLNetVConnection *>(vc);
```
maskit, line 9314: We don't need SSLNetVC here because `get_server_name()` is declared in `NetVConnection`.

PR #8178, opened 2021-07-26
Abstract adding Connection: close header to avoid triggering H2 draining logic

shinrich:
Ran across this while production testing H2 to origin.  Details of that investigation in #7622.

The upshot is that the HTTP/2 logic does pay attention to the Connection header if the value is "closed".  In the HTTP core logic, it regularly sets the "Connection: closed" header in the case of an error to ensure that in the HTTP/1.x case, the client or server connection is closed so any junk left on the socket will not be interpreted as a new transaction.  I had been assuming that adding this header for HTTP/2 ua_txn's was a no-op since HTTP/2 doesn't use the Connection header.

But our implementation will take the presence of a Connection:closed header as an indication that it should start draining the HTTP/2 session.  This PR replaces the logic that directly sets the Connection:closed header with a call to a virtual method.   The HTTP/1.x version will set the header, and the default version does nothing.

https://docs.trafficserver.apache.org/en/latest/admin-guide/plugins/header_rewrite.en.html?highlight=connection%20close%20drain#close-connections-for-draining

With this change adding a header-rewrite rule to add the Connection:closed header will still work to trigger the draining logic.  But core logic to clean up HTTP/1.x connections will not trigger the draining logic.  In theory this change should increase the lifetime of HTTP/2 Sessions.

### DIFF
--- a/proxy/ProxyTransaction.cc
+++ b/proxy/ProxyTransaction.cc
@@ -240,3 +240,11 @@ ProxyTransaction::allow_half_open() const
 {
   return false;
 }
+
+// Most protocols will not want to set the Connection: header
+// For H2 it will initiate the drain logic.  So we make do nothing
+// the default action.
+void
+ProxyTransaction::set_close_connection(HTTPHdr &hdr) const
+{
+}
--- a/proxy/http/HttpTransact.cc
+++ b/proxy/http/HttpTransact.cc
@@ -7052,7 +7052,7 @@ HttpTransact::handle_response_keep_alive_headers(State *s, HTTPVersion ver, HTTP
     if (s->client_info.keep_alive != HTTP_NO_KEEPALIVE || (ver == HTTP_1_1)) {
       if (s->client_info.proxy_connect_hdr) {
         heads->value_set(c_hdr_field_str, c_hdr_field_len, "close", 5);
-      } else {
+      } else if (s->state_machine->ua_txn != nullptr) {
         s->state_machine->ua_txn->set_close_connection(*heads);
       }
       s->client_info.keep_alive = HTTP_NO_KEEPALIVE;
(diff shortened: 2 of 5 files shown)


shinrich:
[approve ci clang-analyzer]

shinrich:
[approve ci autest]

proxy/http/HttpTransact.cc
```diff
@@ -7050,7 +7053,11 @@ HttpTransact::handle_response_keep_alive_headers(State *s, HTTPVersion ver, HTTP
   case KA_CLOSE:
   case KA_DISABLED:
     if (s->client_info.keep_alive != HTTP_NO_KEEPALIVE || (ver == HTTP_1_1)) {
-      heads->value_set(c_hdr_field_str, c_hdr_field_len, "close", 5);
+      if (s->client_info.proxy_connect_hdr) {
```
masaori335, line 7056: Why do we need to take care of the Proxy-Connection header here?

proxy/ProxyTransaction.cc
```diff
@@ -240,3 +240,11 @@ ProxyTransaction::allow_half_open() const
 {
   return false;
 }
+
+// Most protocols will not want to set the Connection: header
+// For H2 it will initiate the drain logic.  So we make do nothing
+// the default action.
```
masaori335, line 246: For H2, I agree with we should not start the graceful shutdown. I'm a bit wondering about we should close the stream (send RST_STREAM frame to the peer).

shinrich: With the normal processing the stream will shutdown.  With HTTP/1.1 the response is sent as normal even it if has Connection: close.  With HTTP/2 after the response header and body are sent, the EOS bit will be sent on the last frame which will cause the stream to shutdown on the other side.

proxy/http/HttpTransact.cc
shinrich: This is to deal with the c_hdr_field_str.  If you look a few lines up, it is either set to Proxy-connection or Connection depending on whether the parent selection path was taken or not.  For H2 I assume we do want to set the proxy-connection (or at least that doesn't trigger the drain logic), so I added a test here to only call the new method in the Connection case.

proxy/ProxyTransaction.cc
masaori335: OK, then, do nothing :+1:

proxy/http/HttpTransact.cc
masaori335: Thanks for the explanation, makes sense.

PR #7674, opened 2021-04-03
Reduce overhead incurred by enabling debug output.

ywkaras:
Setting proxy.config.diags.debug.enabled to 3 enables most core debug output (whose tag matches the debug
regex), and plugin debug output from the new API macro TSDbg().  Setting it to 1 enables all (matching) debug
output, but with the previous high overhead.

Note that enabling a large qualtity of debug output will still have a big impact on performance.

PR #8179, opened 2021-07-27
Add a new --enable-event-tracker configure option

masaori335:
# Feature
When this option is enabled, events track who scheduled them.

# Motivation

This would give us some clues when we face some crashes on the event handler. Especially, the Continuation doesn't have history.

# Usage

To get symbol, convert the address by `addr2line(1)` (Linux) or `atos(1)` (BSD) ( assuming ATS has debug info) like below.

```
» lldb /opt/ats/bin/traffic_server
(lldb) target create "/opt/ats/bin/traffic_server"
Current executable set to '/opt/ats/bin/traffic_server' (x86_64).
...
traffic_server was compiled with optimization - stepping may behave oddly; variables may not be available.
frame #6: 0x00000001002e2135 traffic_server`EThread::process_event(this=0x0000000018038000, e=0x0000000000830b80, calling_code=2) at UnixEThread.cc:164 [opt]
   161       // Restore the client IP debugging flags
   162       set_cont_flags(e->continuation->control_flags);
   163
-> 164       e->continuation->handleEvent(calling_code, e);
   165       ink_assert(!e->in_the_priority_queue);
   166       ink_assert(c_temp == e->continuation);
   167       MUTEX_RELEASE(lock);
(lldb) p e->_location
(const void *) $0 = 0x00000001001a94ac
----
» atos -o /opt/ats/bin/traffic_server 0x00000001001a94ac
HostDBProcessor::start(int, unsigned long) (in traffic_server) (HostDB.cc:425)
```

### DIFF
--- a/iocore/eventsystem/UnixEvent.cc
+++ b/iocore/eventsystem/UnixEvent.cc
@@ -29,6 +29,8 @@
 *****************************************************************************/
 #include "P_EventSystem.h"
 
+#include "tscore/ink_stack_trace.h"
+
 ClassAllocator<Event> eventAllocator("eventAllocator", 256);
 
 void
@@ -104,3 +106,19 @@ Event::schedule_every(ink_hrtime aperiod, int acallback_event)
     ethread->EventQueueExternal.enqueue_local(this);
   }
 }
+
+#ifdef ENABLE_EVENT_TRACKER
+
+void
+Event::set_location()
+{
+  _location = ink_backtrace(3);
+}
+
+const void *
+Event::get_location() const
+{
+  return _location;
+}
+
+#endif
--- a/configure.ac
+++ b/configure.ac
@@ -175,6 +175,13 @@ AC_ARG_ENABLE([all-asserts],
 )
 AC_MSG_RESULT([$enable_all_asserts])
 
+AC_MSG_CHECKING([whether to enable event tracker])
+AC_ARG_ENABLE([event-tracker],
+  [AS_HELP_STRING([--enable-event-tracker],[turn on event tracker])],
+  [],
+  [enable_event_tracker=no]
+)
+AC_MSG_RESULT([$enable_event_tracker])
 
 # Enable code coverage instrumentation only if requested by the user.
 AC_MSG_CHECKING([whether to code coverage])
@@ -973,8 +980,13 @@ fi
 if test "x${enable_mime_sanity_check}" = "xyes"; then
   TS_ADDTO(AM_CPPFLAGS, [-DENABLE_MIME_SANITY_CHECK])
 fi
+
 if test "x${enable_all_asserts}" = "xyes"; then
   TS_ADDTO(AM_CPPFLAGS, [-DENABLE_ALL_ASSERTS])
+fi
+
+if test "x${enable_event_tracker}" = "xyes"; then
+  TS_ADDTO(AM_CPPFLAGS, [-DENABLE_EVENT_TRACKER])
 fi
 
 # Flags for ASAN
--- a/include/tscore/ink_stack_trace.h
+++ b/include/tscore/ink_stack_trace.h
@@ -29,8 +29,15 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
+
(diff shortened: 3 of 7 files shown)


ywkaras:
When pigs fly, and https://github.com/apache/trafficserver/pull/7674 gets merged, seems like it might combine nicely with this, to allow use in production:
```
DbgCtl event_problem_xyz_dbg_ctl("event.problem.xyz");
...
if (is_dbg_ctl_enabled(event_problem_xyz_dbg_ctl)) {
  e->set_location();
}
e->scheule_imm(cont, EVENT_XYZ, nullptr);
...
DbgCtl event_problem_xyz_abc_dbg_ctl("event.problem.xyz.abc");
...
case EVENT_XYZ:
  ...
  if (bad_thing_abc_happened && is_dbg_cntl_enabled(event_problem_xyz_abc_dbg_ctl)) {
    DbgPrint(event_problem_xyz_abc_dbg_ctl, "Oh my ABC happened!!!");
    auto loc = e->get_location();
    // DbgPrint() calls to output location.
    ...
  }
```

iocore/eventsystem/UnixEvent.cc
```diff
@@ -104,3 +106,19 @@ Event::schedule_every(ink_hrtime aperiod, int acallback_event)
     ethread->EventQueueExternal.enqueue_local(this);
   }
 }
+
+#ifdef ENABLE_EVENT_TRACKER
+
+void
+Event::set_location()
+{
+  _location = ink_backtrace(3);
```
masaori335, line 115: This is doing pretty tricky. Assuming this is called by `schedule_*` functions and records where they're called (the 3rd frame of the call stack).

bryancall:
[approve ci]

masaori335:
[approve ci]

iocore/eventsystem/UnixEvent.cc
rob05c: Any idea how performant this is? I.e. is it going to significantly impact how much traffic a production cache can take, and/or TTMS? Or is it very small, could we just leave this on in Production?

iocore/eventsystem/UnixEvent.cc
```diff
@@ -104,3 +106,19 @@ Event::schedule_every(ink_hrtime aperiod, int acallback_event)
...
+}
+
+const void *
+Event::get_location() const
```
rob05c, line 119: I don't see anything calling `get_location`. Does anything use the location yet? Or is the plan to do that in a future PR?

iocore/eventsystem/UnixEvent.cc
rob05c: > the 3rd frame of the call stack
Yeah, the whole "3" thing is awkward, but that's pretty common when looking back on the stack. It seems fragile, but it's also unlikely to change. Meh.

rob05c (approved):
LGTM. 

I'm not super-familiar with this area of the code, but the code looks good, and I built and ran, everything worked as-expected.

IMO this is a great feature, I can see it being useful for lots of things, logging, maybe exposing it to HttpTransact::State and/or plugins. And so far it's pretty small and non-invasive.

iocore/eventsystem/UnixEvent.cc
masaori335: Right. It's just the buddy of `set_location` for now. 
A possible use case is printing the location on debug log instead of using `add2line` command.

iocore/eventsystem/UnixEvent.cc
masaori335: I though `backtrace(3)` doesn't have noticeable overhead. Because it just refers address of current call stack. I'll measure on/off at production.
https://man7.org/linux/man-pages/man3/backtrace.3.html

rob05c: Ah, excellent. I was just curious. It's still very useful in debug mode, but there's so much more we can do if we can just leave it on all the time.

zwoop:
Cherry-picked to v9.2.x

PR #8181, opened 2021-07-28
Various maxmind_acl fixes

ezelkow1:
- Add a check for non-nullptr client addr
 - Add null termination for iso_codes. They are stored unterminated in the database so to properly
   lookup in the map we need to terminate ourselves

These should probably also be backported to 9.1 since the plugin also exists there

### DIFF
--- a/plugins/experimental/maxmind_acl/mmdb.cc
+++ b/plugins/experimental/maxmind_acl/mmdb.cc
@@ -406,6 +406,15 @@ Acl::eval(TSRemapRequestInfo *rri, TSHttpTxn txnp)
 {
   bool ret = default_allow;
   int mmdb_error;
+
+  auto sockaddr = TSHttpTxnClientAddrGet(txnp);
+
+  if (sockaddr == nullptr) {
+    TSDebug(PLUGIN_NAME, "Err during TsHttpClientAddrGet, nullptr returned");
+    ret = false;
+    return ret;
+  }
+
   MMDB_lookup_result_s result = MMDB_lookup_sockaddr(&_mmdb, TSHttpTxnClientAddrGet(txnp), &mmdb_error);
 
   if (MMDB_SUCCESS != mmdb_error) {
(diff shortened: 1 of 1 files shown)


plugins/experimental/maxmind_acl/mmdb.cc
```diff
@@ -498,8 +507,11 @@ Acl::eval_country(MMDB_entry_data_s *entry_data, const char *path, int path_len)
   bool ret     = false;
   bool allow   = default_allow;
   char *output = nullptr;
-  output       = static_cast<char *>(malloc((sizeof(char) * entry_data->data_size)));
+
+  // We need to null terminate the iso_code ourselves, they are unterminated in the DBs
+  output = (char *)malloc((sizeof(char) * (entry_data->data_size + 1)));
```
randall, line 512: can this be changed back to the static_cast?

plugins/experimental/maxmind_acl/mmdb.cc
```diff
@@ -406,6 +406,15 @@ Acl::eval(TSRemapRequestInfo *rri, TSHttpTxn txnp)
...
   int mmdb_error;
+
+  auto sockaddr = TSHttpTxnClientAddrGet(txnp);
+
+  if (sockaddr == nullptr) {
+    TSDebug(PLUGIN_NAME, "Err during TsHttpClientAddrGet, nullptr returned");
+    ret = false;
+    return ret;
+  }
+
```
ywkaras, line 417: I think you meant to change the next line to use sockaddr instead of calling TSHttpTxnClientAddrGet() again.

plugins/experimental/maxmind_acl/mmdb.cc
ezelkow1: yea, these are just things im porting over from our 8.1.x branch for this as we've fixed things, and I think my editor may have stripped that out, Ill add it back

plugins/experimental/maxmind_acl/mmdb.cc
ezelkow1: ahh, yes, thanks, was a quick copy pasta over from 8x

ywkaras (approved):

PR #8182, opened 2021-07-28
Revert "Remove UnixNetVConnection::startEvent - not actually called."

tomoatan:
Reverts #7596

In #7596, UnixNetVConnection::startEvent was removed but it's necessary in UnixNetProcessor::connect_re_internal.
In the following code, UnixNetVConnection is scheduled if some mutexes can't be locked.
https://github.com/apache/trafficserver/blob/84cf02e6d87377d68cce8f687f6c001bc43aea47/iocore/net/UnixNetProcessor.cc#L240-L254

In this case, UnixNetVConnection::startEvent is appropriate as a handler
because UnixNetVConnection is scheduled to execute UnixNetVConnection::connectUp.
Before #7596, UnixNetVConnection::startEvent was set as a handler in constructor, so I revert it.

### DIFF
--- a/iocore/net/P_QUICNetVConnection.h
+++ b/iocore/net/P_QUICNetVConnection.h
@@ -168,6 +168,7 @@ public:
   int connectUp(EThread *t, int fd) override;
 
   // QUICNetVConnection
+  int startEvent(int event, Event *e);
   int state_pre_handshake(int event, Event *data);
   int state_handshake(int event, Event *data);
   int state_connection_established(int event, Event *data);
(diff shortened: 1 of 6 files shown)


randall:
[approve ci]

bneradt:
I’m on PTO and away from my computer right now. But I see that clang analyzer fails with this. @SolidWallOfCode  and I looked at the report a few months ago and that’s why we initially removed this code (to address it). We convinced ourselves that it was not called.  But apparently this code can’t be removed. We’ll  have to figure out how to fix the analyzer issue another way apparently.

masaori335:
#8184 has the commit from @tomoatan and the clang-analyzer fix. Thanks!

PR #7885, opened 2021-05-26
Close connection after every bad request for HTTP/1.1

shinrich:

PR #8189, opened 2021-07-29
Minor updates to HTTP version validation

bryancall:
Renamed the functions to be more explicit about only supporting HTTP/1.x
Changed the version check to be only a logic statement

This is a update to #7875

I would like to pull #7875 into the 9.1.x branch because we right now have an autest for testing invalid versions and it is failing right now (because of #7885).  However, @zwoop and I didn't like the function names.

### DIFF
--- a/proxy/hdrs/HTTP.cc
+++ b/proxy/hdrs/HTTP.cc
@@ -633,26 +633,23 @@ http_hdr_type_set(HTTPHdrImpl *hh, HTTPType type)
   -------------------------------------------------------------------------*/
 
 bool
-is_version_supported(const uint8_t major, const uint8_t minor)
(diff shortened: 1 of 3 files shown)


shinrich:
Seems ok.  Since the HttpSM currently only manipulates HTTP/1.x headers the original more generic name made some sense.  Currently for HTTP/2 (and I assume HTTP/3) the headers are translated to HTTP/1.x before they are passed from the ProxyTransaction to the HttpSM. Hopefully that translation disappears at some point.  Then the naming will need to be reconsidered (and the function will need to be expanded).

masaori335 (approved):
Looks good. As @shinrich pointed, we'll need to revisit in the future. But it's good for now.

bryancall:
Cherry picked to 9.1.x

PR #8191, opened 2021-07-30
Fix error connection logging crash

shinrich:
If unmapped_url is not set, the code that logs a connection failure will crash.  We had this reported by a customer using a plugin which meant no remap rule was traversed.  This fix addressed their issue.

This code was brought in PR #7757, so any branch with that change also needs this change.


bneradt:
Looks good to me.

randall (approved):

PR #8201, opened 2021-08-03
Fix H2 logic when setting EOS flag on DATA frame

shinrich:
Noticed this while debugging HTTP2 to origin.  stream->update_sent_count() updates the write_vio.ndone.  But in the original code this update is done after the stream->is_write_vio_done() check is made to determine whether the EOS flag should be set.  The EOS flag will only be set if the payload is 0.

I would think this would have caused problems quite a lot of the time.  Generally showing up as timeouts as the client waits forever for the EOS flag after receiving all the data.

### DIFF
--- a/proxy/http2/Http2ConnectionState.cc
+++ b/proxy/http2/Http2ConnectionState.cc
@@ -1677,6 +1677,8 @@ Http2ConnectionState::send_a_data_frame(Http2Stream *stream, size_t &payload_len
     payload_length = 0;
   }
 
+  stream->update_sent_count(payload_length);
+
   // Are we at the end?
   // If we return here, we never send the END_STREAM in the case of a early terminating OS.
   // OK if there is no body yet. Otherwise continue on to send a DATA frame and delete the stream
@@ -1700,8 +1702,6 @@ Http2ConnectionState::send_a_data_frame(Http2Stream *stream, size_t &payload_len
 
   Http2DataFrame data(stream->get_id(), flags, resp_reader, payload_length);
   this->ua_session->xmit(data, flags & HTTP2_FLAGS_DATA_END_STREAM);
-
-  stream->update_sent_count(payload_length);
 
   if (flags & HTTP2_FLAGS_DATA_END_STREAM) {
     Http2StreamDebug(ua_session, stream->get_id(), "END_STREAM");
--- a/proxy/http2/Http2ConnectionState.cc
+++ b/proxy/http2/Http2ConnectionState.cc
@@ -1688,7 +1688,7 @@ Http2ConnectionState::send_a_data_frame(Http2Stream *stream, size_t &payload_len
     return Http2SendDataFrameResult::NO_PAYLOAD;
   }
 
-  if (stream->is_write_vio_done() && !resp_reader->is_read_avail_more_than(0)) {
+  if (stream->is_write_vio_done()) {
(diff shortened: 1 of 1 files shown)


shinrich:
[ci approve clang-analyzer]

shinrich:
[approve ci clang-analyzer]

shinrich:
[approve ci clang-analyzer]

bryancall:
@masaori335 is going to look at it

proxy/http2/Http2ConnectionState.cc
```diff
@@ -1677,6 +1677,8 @@ Http2ConnectionState::send_a_data_frame(Http2Stream *stream, size_t &payload_len
     payload_length = 0;
   }
 
+  stream->update_sent_count(payload_length);
```
masaori335, line 1680: The condition of setting END_STREAM flag is `stream->is_write_vio_done() && !resp_reader->is_read_avail_more_than(0)`  (line 1691).

This change affects the first condition but has no effect on the second condition.

masaori335:
> I would think this would have caused problems quite a lot of the time. Generally showing up as timeouts as the client waits forever for the EOS flag after receiving all the data.

I think we're not seeing this issue because of sending the pure END_STREAM DATA frame (length 0).

proxy/http2/Http2ConnectionState.cc
shinrich: Correct.  Assuming there is no more data to read, then promoting the update_sent_count() will cause both cases on the test on line 1691 to be true.

shinrich:
True, the empty EOS data frames are probably saving us.  However, it seems better to sent the EOS on the last data frame rather than sending an empty data frame just for the EOS flag.

proxy/http2/Http2ConnectionState.cc
masaori335: I'm +1 to set the EOS flag when it's possible. But I can't imagine there're no more data to read but `payload_length` is not 0 because of this logic.

```
1663   if (resp_reader->is_read_avail_more_than(0)) {
...
1676   } else {
1677     payload_length = 0;
1678   }
```

shinrich: Hmm, you are right the extra check on  !resp_reader->is_read_avail_more_than(0) before setting the EOS byte means nothing will really change.

Seems like we shouldn't need that check either.  Let me work with that a bit more and close or update as necessary.

shinrich:
Removed the !resp_reader->is_read_avail_more_than(0) test that was guarding the setting of the EOS bit.

Either the write_vio nbytes is set to a specific nbytes and the payload_length got us there.  Or it didn't . The read_avail will still be > 0 until we get to the xmit() call and that is ok.  Should still set the EOS flag, if the data still in the buffer is the last set we will send.

Ran it in my production test with a Warning message for the EOS and payload > 0 case.  We do run into that case.  The overall numbers look fine.  Ran a few of the h2 autests and they seem fine.  Will see how the PR autest goes..

masaori335:
I was wondering this can work well on chunked response, but AuTest says it's working. Let's ship this.

masaori335 (approved):

shinrich:
In the chunked case, we may well be falling back to the empty data frame EOS case.  In the chunked processing, we eventually adjust the nbytes to the actual number of bytes processed, but that doesn't happen until a READ_COMPLETE is processed.

PR #8202, opened 2021-08-03
Reduce possibility for H2 client count mismatch

### LINKED ISSUE #8200: Count of concurrent connections diverges between ATS H2 client and ATS H2 server
This is an issue I saw while debugging the HTTP/2 to origin branch between two ATS servers.  But the issue could in theory be seen between another busy H2 client and ATS acting as the H2 server.

For ATS, Http2ConnectionState has the client_streams_in_count to track how many streams the session has active.  It checks this value against the the session's max_concurrent_stream setting to make sure creating a new stream does not violate the session's max_concurrent_stream policy.  Similarly, the H2 client (ATS for my branch) must be keeping similar counts to track the peer's current stream count.  It should not send a new HEADER frame to the peer if that would push the peer over it's max_concurrent_stream limit.  Therefore, client_streams_in_count for each machine should be kept consistent.

This consistency was not always maintained, because on the client_streams_in_count was not being decremented until the HttpSM was destroyed. Depending on caching and other operations this may be some time after the frame with the EOS was sent or received.

So the client could dutifully be decrementing the stream count on receiving the EOS frame, and then have a lower stream count than the peer and send too many HEADER frames.  While this is only a STREAM error, it is likely that a DATA frame would be sent soon after resulting in a BAD stream ID error which is a connection error.  This would take down the session and all active streams (which could be quite a few streams on a busy session).

In my test branch I pulled the client_streams_in_count into the Http2SStream::change_state method and Http2Stream::do_io_close and Http2Stream::initiating_close for streams that get shutdown from an odd state.

shinrich:
Details in issue #8200.  The code change pulls up the decrementing logic for the client stream count, so we can be more likely to be consistent with the H2 peers counting.

This closes #8200

### DIFF
--- a/proxy/http2/Http2Stream.cc
+++ b/proxy/http2/Http2Stream.cc
@@ -273,103 +273,114 @@ Http2Stream::send_request(Http2ConnectionState &cstate)
 bool
 Http2Stream::change_state(uint8_t type, uint8_t flags)
 {
-  switch (_state) {
-  case Http2StreamState::HTTP2_STREAM_STATE_IDLE:
-    if (type == HTTP2_FRAME_TYPE_HEADERS) {
-      if (recv_end_stream) {
-        _state = Http2StreamState::HTTP2_STREAM_STATE_HALF_CLOSED_REMOTE;
-      } else if (send_end_stream) {
-        _state = Http2StreamState::HTTP2_STREAM_STATE_HALF_CLOSED_LOCAL;
+  Http2StreamState original_state = _state;
+  bool retval                     = true;
+  if (type == HTTP2_FRAME_TYPE_RST_STREAM) {
+    _state = Http2StreamState::HTTP2_STREAM_STATE_CLOSED;
+  } else {
+    switch (_state) {
+    case Http2StreamState::HTTP2_STREAM_STATE_IDLE:
+      if (type == HTTP2_FRAME_TYPE_HEADERS) {
+        if (recv_end_stream) {
+          _state = Http2StreamState::HTTP2_STREAM_STATE_HALF_CLOSED_REMOTE;
+        } else if (send_end_stream) {
+          _state = Http2StreamState::HTTP2_STREAM_STATE_HALF_CLOSED_LOCAL;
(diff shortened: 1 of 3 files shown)


shinrich:
[approve ci clang-analyzer]

shinrich:
[approve ci autest]

bryancall:
@masaori335 is going to look at it.

proxy/http2/Http2Stream.cc
```diff
@@ -273,103 +273,111 @@ Http2Stream::send_request(Http2ConnectionState &cstate)
...
-    return true;
+  }
 
-  default:
-    return false;
+  // Did we just transition to closed or remote closed? Decrement stream count
+  if (_state == Http2StreamState::HTTP2_STREAM_STATE_CLOSED && original_state != Http2StreamState::HTTP2_STREAM_STATE_CLOSED) {
+    Http2ClientSession *h2_proxy_ssn = static_cast<Http2ClientSession *>(this->_proxy_ssn);
+    SCOPED_MUTEX_LOCK(lock, h2_proxy_ssn->connection_state.mutex, this_ethread());
+    h2_proxy_ssn->connection_state.decrement_stream_count(this->_id);
```
masaori335, line 375: I can understand we can check state changes here, but I'm not a big fan of giving another task to the `Http2Stream::change_state()`.

bryancall:
[approve ci]

maskit (commented):
Decrementing stream count before removing a stream from `stream_list` sounds dangerous. That could allow keeping a lot of stream instances in the list. I think we should not do it separately at different places.

If I understand the point of this PR correctly, it's decrementing the count as soon as possible when it gets obvious to close a stream. I agree that it would reduce possibility for the count mismatch a little, but I'm not sure if the change helps a lot. I'd try to delete a stream asap if the delay matters.

And it seems like the root cause of the mismatch is a bug in ATS's client code, according to the last comment on #8200. I suggest closing this PR without landing.

ezelkow1:
[approve ci rocky]

PR #8203, opened 2021-08-03
Add setting to allow different windows for stream and session

### LINKED ISSUE #8199: Should have separate values for stream and session windows in HTTP/2
Noticed this while debugging HTTP/2 to origin.  Traffic Server provides proxy.config.http2.initial_window_size_in to change the initial window size from the default of 65536.  However, this value is used for both the stream window and the session window.  So any single stream can overrun the shared session window.

Looking at window values from our clients, it seems that many clients will have a higher session window than the stream window.

For my testing, I just multiplied the session window by 50.  That plus greatly increasing the stream window size, got my test scenario working (one ATS level using a H2 connection to another ATS level to interleave streams from many client requests).

I propose that we add a second setting to control the session window.  It could be a multiplier, e.g proxy.config.http2.session_window_multiplier_in, or just an explicit session window setting, e.g., proxy.config.http2.session initial_window_size_in.  If the session version of the setting is not set, the stream version will be used as it is in the current code.

I would appreciate your thoughts on this.

shinrich:
Issue #8199 outlines the issue.  This PR adds a new setting to allow for setting the session window separately from the stream window.

This closes #8199

### DIFF
--- a/doc/admin-guide/files/records.config.en.rst
+++ b/doc/admin-guide/files/records.config.en.rst
@@ -3952,7 +3952,17 @@ HTTP/2 Configuration
 .. ts:cv:: CONFIG proxy.config.http2.initial_window_size_in INT 65535
    :reloadable:
 
-   The initial window size for inbound connections.
+   The initial window size for inbound connection streams.
+
+.. ts:cv:: CONFIG proxy.config.http2.session_initial_window_size_in INT 0
+   :reloadable:
+
+   The initial window size for inbound connection session.  HTTP/2 provides both
+   a per stream window and a session wide window.  Each data byte exchanged decrements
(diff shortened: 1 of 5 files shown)


shinrich:
[approve ci clang-analyzer]

bryancall:
@masaori335 or @maskit can you please take a look at this PR

maskit (commented):
As I commented on the issue, I don't think this is a good solution. -0 from me.

bneradt:
Let's re-open this and pull it in separately from the new HTTP/2 to origin PR (#8963).

bneradt:
Since this has conflicts, I opened #9085 to take over from this PR.

PR #8208, opened 2021-08-04
Make sni.yaml errors cause an unrecoverable TS crash.

ywkaras:


brbzull0 (dismissed):

iocore/net/SSLSNIConfig.cc
```diff
@@ -170,7 +170,7 @@ SNIConfigParams::Initialize()
   if (!zret.isOK()) {
     std::stringstream errMsg;
     errMsg << zret;
-    Error("%s failed to load: %s", sni_filename, errMsg.str().c_str());
+    Emergency("%s failed to load: %s", sni_filename, errMsg.str().c_str());
```
randall, line 173: Is this just the initial load(on startup) of the the sni.yaml?

ywkaras: Yes

ywkaras:
[approve ci Clang-Analyzer]

iocore/net/SSLSNIConfig.cc
masaori335: It looks like reloading the sni.yaml comes here for me...what I'm missing?
https://github.com/apache/trafficserver/blob/11f987e3a84df907582ac6ea8150b414d15fc15e/iocore/net/SSLSNIConfig.cc#L196-L201

ywkaras: @rob05c did you want it to die hard on a reload as well?

rob05c: Yeah, I guess that's fine. On startup, you're not breaking any running traffic. Reload is a harder decision: do we kill the app and break running traffic in the face of a potential security/access concern?

I guess not breaking running traffic is an acceptable choice here. @alficles ^

I tried to see what other configs like ip_allow.yaml do. It throws an exception and crashes ATS. But I'm not sure if that's intentional or desirable.

I also feel like a running config changing to be malformed is at least a little less likely. It seems highly likely to me that a user might create a file wrong and never notice the log error; it seems slightly less likely that a user would create a correct file, and later malform it.

iocore/net/SSLSNIConfig.cc
```diff
@@ -166,13 +167,19 @@ SNIConfigParams::Initialize()
...
+  ts::Errata zret = Y_sni_tmp.loader(sni_filename);
   if (!zret.isOK()) {
     std::stringstream errMsg;
     errMsg << zret;
-    Error("%s failed to load: %s", sni_filename, errMsg.str().c_str());
+    if (TSSystemState::is_initializing()) {
+      Emergency("%s failed to load: %s", sni_filename, errMsg.str().c_str());
+    } else {
+      Error("%s failed to load: %s", sni_filename, errMsg.str().c_str());
+    }
```
ywkaras, line 179: This logic may be common enough that we should add a new diag output call, perhaps InitEmergency(), that blocks  startup if initializing, but simply logs an error otherwise.

iocore/net/SSLSNIConfig.cc
```diff
@@ -166,13 +167,19 @@ SNIConfigParams::Initialize()
...
     return 1;
   }
+  Y_sni = std::move(Y_sni_tmp);
```
ywkaras, line 182: I added this logic to make sure the static data remained the coherent, old version if an invalid sni.yaml was reloaded.

bryancall:
[approve ci]

zwoop:
Cherry-picked to v9.2.x

PR #8212, opened 2021-08-05
WIP - First view of the records.yaml file -  prototype 

brbzull0:
**WIP Draft PR** that shows a first view of the converted `records.config` into a `yaml `structure.  The whole idea is to show how this looks, start the conversation and collect feedback. This does not involve any core changes, it's just a view of the yaml file.

----------
As a part of this draft I have added a new  enable feature flag:  `--enable-records-yaml-gen` to let the build to process the existing `records.config.default.in` and convert it into a `YAML` file, with a proper structured objects. The script will basically drop the `proxy.config` and `proxy.local `and work from there. Main level structure will be `ts`. You can try this on with your own record.config and see how it looks, just check the parameters of the script(schema validation may fail)
 
#### Example
 
Just a small snap of a config file.
```
CONFIG proxy.config.exec_thread.autoconfig INT 1
CONFIG proxy.config.exec_thread.autoconfig.scale FLOAT 1.0
CONFIG proxy.config.exec_thread.limit INT 2
CONFIG proxy.config.exec_thread.affinity INT 1
CONFIG proxy.config.accept_threads INT 1
CONFIG proxy.config.task_threads INT 2
CONFIG proxy.config.cache.threads_per_disk INT 8
CONFIG proxy.config.net.sock_option_flag_in INT 0x5
CONFIG proxy.config.diags.debug.tags STRING http|dns
CONFIG proxy.config.dump_mem_info_frequency INT 0
CONFIG proxy.config.http.slow.log.threshold INT 0
```
 
will generate:
```yaml
ts:
  exec_thread:
    autoconfig:
      value: 1
      scale: 1.0
    limit: 2
    affinity: 1
  accept_threads: 1
  task_threads: 2
  cache:
    threads_per_disk: 8
  net:
    sock_option_flag_in: '0x5'
  diags:
    debug:
      tags: http|dns
  dump_mem_info_frequency: 0
  http:
    slow:
      log:
        threshold: 0
 
```
 
Note that for configs like this:
 
```yaml
CONFIG proxy.config.exec_thread.autoconfig INT 1
CONFIG proxy.config.exec_thread.autoconfig.scale FLOAT 1.0
```
We changed  the approach and added a 'value' field into the key(autoconfig) and made the main key a structure. So the above example will look like this:
```yaml
  exec_thread:
    autoconfig:
      value: 1
      scale: 1.0
```
If you want to see how it looks like the entire generated file: [check here](https://gist.github.com/brbzull0/e286e64e74f71061a6664cc961da6e51)
 
 
####  Schema
 
This draft also has a `records.schema.json` that is used to validate the generated records.yaml output, this is optional in the gen script.
 
The schema file is very basic and in its early stage, booleans, some enums and common types should be defined, this is WIP.
 
---- 
Python script is not yet a PR quality, so you can ignore it.
 
Feedback will be appreciated.

### DIFF
--- a/configs/Makefile.am
+++ b/configs/Makefile.am
@@ -41,6 +41,24 @@ dist_sysconf_DATA =	\
 	strategies.yaml.default \
 	volume.config.default
 
+if GENERATE_RECORDS_YAML
+all-am: gen-rec-yaml
+endif
+
+gen-rec-yaml:
+# generate records.yaml base on the current records.config, also run a schema validation with it.
+	@python3 $(top_builddir)/contrib/python/records_config_to_yaml.py \
+		--file records.config.default.in \
+		--schema records.schema.json \
+		--yaml \
+		--save2 records.yaml; \
+	if [ $$? -ne 0 ] ; then \
+		echo records.yaml file not generated.; \
+		exit 1; \
+	else \
+		echo records.yaml file generated.; \
+	fi
+
 install-exec-hook:
 	for dfltcfgfile in $(dist_sysconf_DATA) $(nodist_sysconf_DATA) ; \
 	do \
--- a/configs/records.schema.json
+++ b/configs/records.schema.json
@@ -1,0 +1,2457 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Rercods config schema",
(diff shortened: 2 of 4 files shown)


bryancall:
[approve ci]

PR #8217, opened 2021-08-06
uri_signing: fix for copying too many bytes and compiler warnings

traeak:
Fixes a memory fault discovered by asan (off by one error).
Also switches to use memset to initialize cjose_err and jwt_error_t types to remove compiler warnings.

### DIFF
--- a/plugins/experimental/uri_signing/config.c
+++ b/plugins/experimental/uri_signing/config.c
@@ -298,7 +298,8 @@ read_config_from_json(json_t *const issuer_json)
     }
 
     json_t *jwk_obj;
-    cjose_err jwk_err = {0};
+    cjose_err jwk_err;
+    memset(&jwk_err, 0, sizeof(cjose_err));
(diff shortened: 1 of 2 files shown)


plugins/experimental/uri_signing/config.c
```diff
@@ -298,7 +298,8 @@ read_config_from_json(json_t *const issuer_json)
     }
 
     json_t *jwk_obj;
-    cjose_err jwk_err = {0};
+    cjose_err jwk_err;
+    memset(&jwk_err, 0, sizeof(cjose_err));
```
ywkaras, line 302: I don't think these changes are necessary (but harmless).  https://godbolt.org/z/xceT5f5z8

ywkaras (approved):

plugins/experimental/uri_signing/config.c
traeak: True.  I wouldn't have changed these if clang wasn't complaining about those {0} initializers.  This was from centos llvm toolset 7.

zwoop:

zwoop:

PR #8225, opened 2021-08-09
Disambiguate overloads of Machine::is_self.

ywkaras:


iocore/utils/I_Machine.h
```diff
@@ -77,6 +77,11 @@ struct Machine {
   /// @return The global instance of this class.
   static self_type *instance();
   bool is_self(std::string_view name);
+  bool
+  is_self(char const *name)
+  {
+    return this->is_self(std::string_view(name));
+  }
```
bneradt, line 84: Probably best to keep the definition in `iocore/utils/Machine.cc` like the rest of the functions here.

bneradt (changes requested):
Thanks for fixing this Walt!

iocore/utils/I_Machine.h
ywkaras: OK

ywkaras: Not good for optimal inlining, but not called in any high-rep loop I don't think.

bneradt (commented):
Looks good to me.

ywkaras:
> Looks good to me.

You've still got it blocked.

randall (approved):

bneradt (approved):
Thanks Walt!

PR #8226, opened 2021-08-09
Add normalize_ae option 3

ezelkow1:
This normalizes when both `br` and `gzip` are present to `br, gzip`. In the other cases the AE gets stripped down a single value, eliminating options for origin responses. In this case the origin still has the option to
send back br or gzip while normalizing the cached AE header to still keep alternates to a minimum and still have compression. If both are not present then it acts like the existing option `2` if br is present, then option
`1` if `br` is not present but `gzip` is


ezelkow1:
[approve ci autest]

ywkaras (approved):

PR #8227, opened 2021-08-09
Fix the skipping logic for autest and docs

bryancall:
Look at the difference in the main branch instead of looking at the last
commit on the branch is being merged in

Turned on debugging for the autest and docs build scripts

Here is an example of what was happening on our CI boxes:

After the merge the tree looked like this:
```
*   b05db4c2e - (HEAD, origin/pr/7925/merge) Merge 2e22854328145c4dbd79d286c684a6dca7cc22a7 into f15a71955830726a4f71cb5b47fd0bad1292dd18 (5 hours ago) <Walt Karas>
|\
| * 2e2285432 - (origin/pr/7925/head) Clarify when the keyword 'self' should be used in strategies.yaml. (5 hours ago) <Walter Karas>
| * b406fa47a - Fix clang format error. (6 hours ago) <Walter Karas>
| * abfaa1071 - Disambiguate overloads of Machine::is_self. (6 hours ago) <Walter Karas>
| * 3cc2bbc2a - For peering ring, make upstream group of hosts optional. (6 hours ago) <Walter Karas>
* | f15a71955 - Add the origin IP to the error message for invalid server response in OSDNSLookup (#8031) (6 hours ago) <Bryan Call>
|/
* 440382cdd - Fix heap-use-after-free of ts-lua plugin (#8215) (3 days ago) <Masaori Koshiba>
```

And the command CI was running was:
`++ git diff '2e22854328145c4dbd79d286c684a6dca7cc22a7^...2e22854328145c4dbd79d286c684a6dca7cc22a7' --name-only`

This would only pick up the change in the last commit on the branch that is being merged.

### DIFF
--- a/ci/jenkins/bin/autest.sh
+++ b/ci/jenkins/bin/autest.sh
@@ -16,14 +16,14 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-set +x
+set -x
 
 cd src
 sleep 30
 
 git branch --contains ${ghprbActualCommit} > /dev/null
 if [ $? = 0 -a ! -z "$ghprbActualCommit" ]; then
-    git diff ${ghprbActualCommit}^...${ghprbActualCommit} --name-only | egrep -E '^(build|iocore|proxy|tests|include|mgmt|plugins|proxy|src)/' > /dev/null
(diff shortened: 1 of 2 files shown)


bneradt (commented):
I see. So `HEAD~1` means to compare with the "first" parent, which will mean the receiving branch of the PR, like master.

Makes sense to me.

masaori335 (approved):

randall (approved):

PR #8231, opened 2021-08-10
Upgrade Catch.hpp to v2.13.7

randall:


bneradt (approved):
Thanks for keeping this updated.

bneradt:
To fix the unit tests in 9.1.x for newer compilers, we will need this as well.

PR #8236, opened 2021-08-12
Fix crash in lua host lookup

shukitchan:
I think I miss reading the documentation - https://docs.trafficserver.apache.org/en/latest/developer-guide/api/functions/TSHostLookup.en.html

I was not aware of the case when the call can be synchronous

Also we need to yield the luaVM when we need to wait for an answer. And we should not yield if the call is synchronous. And then in the handler, we need to see if we have yielded or not. So we can resume only if we have yielded.

### DIFF
--- a/plugins/lua/ts_lua_misc.c
+++ b/plugins/lua/ts_lua_misc.c
@@ -444,13 +444,17 @@ ts_lua_host_lookup(lua_State *L)
 
   host = luaL_checklstring(L, 1, &host_len);
 
-  contp  = TSContCreate(ts_lua_host_lookup_handler, ci->mutex);
-  action = TSHostLookup(contp, host, host_len);
+  contp = TSContCreate(ts_lua_host_lookup_handler, ci->mutex);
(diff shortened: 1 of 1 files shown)


bneradt (commented):
Makes sense to me.

plugins/lua/ts_lua_misc.c
```diff
@@ -444,30 +444,42 @@ ts_lua_host_lookup(lua_State *L)
...
   lua_State *L;
   ts_lua_coroutine *crt;
+  unsigned int resume;
 
   ai  = TSContDataGet(contp);
   ci  = ai->cinfo;
   crt = &ci->routine;
   L   = crt->lua;
 
+  // find out if need to resumt luaVM before async item cleanup
```
randall, line 475: typo (resumt)

shukitchan: fixed

randall (approved):

zwoop:

zwoop:

ezelkow1:
Do we still want this in 8.x? If so can we get a backport?

PR #8237, opened 2021-08-12
Cleanup url_*_get/set functions

maskit:
An alternative to #8170. Instead of removing unused free functions, this converts the free functions to methods of URLImpl class and call those from URL class.

This also changes names of type_get/set to type_code_get/set, because there are "type" and "type code" in URLImpl and type_get/set were used for "type code". And I redefined type_get/set for "type".

I could add a constructor and make other functions to methods as well, and member variables of URLImpl should be private, but I didn't make those changes for now because this is enough to save the unused functions.


bryancall:
@shinrich can you please take a look at this.

SolidWallOfCode (approved):

PR #8245, opened 2021-08-13
Upgrades to latest fastlz version

randall:
Move to lib/fastlz to remind people it exists
    
Removed reference to lib/wcpp and lib/tsconfig from NOTICE. These were
dropped in ATS 9.x


ywkaras:
+1 your intentions.  But when I've tried to do cleanup PRs, no matter how big the mess it's cleaning up, people seem to feel that, if it seems to be working, don't even breath on it.

src/tscore/fastlz.c
```diff
@@ -24,512 +21,615 @@
...
+    memmove(dest, src, count);
+  } else {
+    switch (count) {
+    default:
+      do {
+        *dest++ = *src++;
+      } while (--count);
+      break;
+    case 3:
+      *dest++ = *src++;
```
ywkaras, line 98: https://en.cppreference.com/w/cpp/language/attributes/fallthrough

ywkaras: Oh wait it's C.  Maybe just change the extension?  Or else:  https://stackoverflow.com/questions/45349079/how-to-use-attribute-fallthrough-correctly-in-gcc

src/tscore/fastlz.cc
```diff
@@ -0,0 +1,635 @@
...
+    memmove(dest, src, count);
+  } else {
+    switch (count) {
+    default:
+      do {
+        *dest++ = *src++;
+      } while (--count);
+      break;
+    case 3:
+      *dest++ = *src++;
```
ywkaras, line 98: You still need to add `[[fallthrough]]` here and after lines 100 and 102.

masaori335: `[[fallthrough]]` is C23. I think we need to move this to `/lib/` and compile with cc.
https://en.cppreference.com/w/c/language/attributes/fallthrough

randall: `[[fallthrough]]` added. Moved to lib/fastlz and renamed to .cc

randall:
[approve ci autest]

ywkaras (approved):

randall:
I almost gave up here. No good deed goes unpunished.

PR #8246, opened 2021-08-13
Fix log filteris for IP conditions.

### LINKED ISSUE #6405: LogFilterIP appears to be broken
Using a simple log filter condition like `%<chi> MATCH 127.0.0.1` doesn't ever match, so LogFilterIP::toss_entry always tosses the log line as it's being reported by the debug log:
```
Feb 07 13:28:13 traffic-cache-atstext traffic_manager[8927]: [Feb  7 13:28:13.466] {0x2adf1ea94700} DEBUG: <LogObject.cc:555 (log)> (log) entry filtered, skipping ...
```

it looks like `LogFilterIP::is_match` https://github.com/apache/trafficserver/blob/master/proxy/logging/LogFilter.cc#L831-L844 needs to be fixed.

bneradt:
Our log filter mechanism for IP addresses was broken. This fixes the
matching mechanism so it compares IP addresses with log fields
correctly.

Fixes #6405

---

The LogFilterIP::is_match mechanism stopped working at some point in the past. That is what the linked issue observes.  I fixed this by addressing the warning in the comment that was in that function:

```
    // This is bad, we abuse the fact that the initial layout of LogFieldIpStorage and IpAddr
    // are identical. We should look at converting the log stuff to use IpAddr directly.
```

Those structures have diverged such that they do not match between each other anymore. I thus converted the use of LogFieldIp* types to IpAddr. @SolidWallOfCode : you wrote that code (and the comment) initially. Can you please verify that what I'm doing here is what you had in mind?

### DIFF
--- a/proxy/logging/LogAccess.cc
+++ b/proxy/logging/LogAccess.cc
@@ -387,26 +387,8 @@ LogAccess::marshal_mem(char *dest, const char *source, int actual_len, int padde
 int
 LogAccess::marshal_ip(char *dest, sockaddr const *ip)
 {
-  LogFieldIpStorage data;
-  int len = sizeof(data._ip);
-  if (nullptr == ip) {
-    data._ip._family = AF_UNSPEC;
-  } else if (ats_is_ip4(ip)) {
-    if (dest) {
-      data._ip4._family = AF_INET;
-      data._ip4._addr   = ats_ip4_addr_cast(ip);
-    }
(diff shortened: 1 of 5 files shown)


bneradt:
This is a bit ambitious for fixing this bug. Converting the logging code to IpAddr can be a separate PR. I'll close this PR for now and simply convert to IpAddr in the LogFilterIP code.

PR #8247, opened 2021-08-13
Traffic Dump: dump response bodies and filter by client IP

bneradt:
This adds the -b option to traffic dump to dump server response body
data. It also adds the -4 and -6 options so that the client can filter
what is dumped based upon a client IP address.


bneradt:
This requires https://github.com/apache/trafficserver/pull/8204 and thus cannot go to any releases that does not have that fix.

bryancall:
@randall can you please review this

bneradt:
[approve ci freebsd]

randall (approved):

PR #8256, opened 2021-08-17
Bubble up ssl_multicert.config load failures

randall:


ywkaras (dismissed):

ywkaras:
This just seems to be adding sensible error checking.  Profoundly scary it's making half the Au test cases fail.  What terrible thing did Randall do in a previous life time to get stuck with this problem?

randall:
[approve ci autest]

iocore/net/SSLUtils.cc
```diff
@@ -1953,8 +1953,9 @@ SSLMultiCertConfigLoader::load(SSLCertLookup *lookup)
   if (ec) {
     switch (ec.value()) {
     case ENOENT:
+      // missing config file is an acceptable runtime state
```
shinrich, line 1956: We don't what to fail is ssl_multicert.config is unavailable?

randall: autest fails all over the place without allowing this. I think we generally try to strive to allow for missing config files (at least since I've been involved in the project)

mlibbey: A server may not have any https remaps ... can imagine that requiring a blank ssl_multicert would be confusing.

shinrich: Ok.  Good point.

shinrich (approved):

zwoop:

PR #8257, opened 2021-08-18
Allow for graceful handling of ip_allow.yaml format errors

randall:
Also, when a bad config is attempted to be reloaded, we won't overwrite a good working config.


shinrich (approved):

zwoop:

PR #8260, opened 2021-08-18
Fix Http2Stream leaks

duke8253:
We have been observing high memory utilization in production after upgrading to ats9.1. JeMalloc profiling shows the problem is related to `Http2Stream`.

### DIFF
--- a/proxy/http2/Http2Stream.cc
+++ b/proxy/http2/Http2Stream.cc
@@ -154,10 +154,12 @@ Http2Stream::main_event_handler(int event, void *edata)
   if (e == _read_vio_event) {
     _read_vio_event = nullptr;
     this->signal_read_event(e->callback_event);
+    reentrancy_count--;
     return 0;
   } else if (e == _write_vio_event) {
     _write_vio_event = nullptr;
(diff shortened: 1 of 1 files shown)


shinrich:
Setting zombie event, we found a session that had the http2 inactivity timeout trigger.  Http2Stream::initiating_close had been called on the stream, but it had not been cleaned up in the 5 minutes of the zombie timeout.  We think the problem is that the write_vio had nbytes and ndone set to 0.  So a WRITE_COMPLETE was sent to the state machine to clean it up.  But there was no real write and the watch for client abort handler really needs an EOS to clean up.  This change should case an EOS to be sent in this scenario.

shinrich:
Note also that we are only seeing the memory growth on our deployments that have the follow redirect enabled.

masaori335 (approved):
I'd like to see which commit made the leak, but checking `write_vio.nbytes > 0` looks harmless. 

Probably, 9.1 starts calling something like `do_io_write(this, 0, nullptr)` from HttpSM in some cases.

zwoop:

PR #8264, opened 2021-08-19
Prefetch plugin: add  --fetch-query option

knutsel:
Add the ability to hint the next to-be-prefetched url in a query string.


knutsel:
[approve ci clang-format]

traeak (changes requested):
plugin.cc.orig shouldn't be here.

knutsel:
> plugin.cc.orig shouldn't be here.

Sorry! Pushed a commit to rm.

traeak (approved):
The additions are fine.  In general the plugin docs could probably be improved with things like better examples.  Also adding autests for this plugin should happen.

PR #8266, opened 2021-08-19
Do not turn off cache for internal requests

vmamidi:
As of now, auth proxy disables cache for all the internal requests. Auth proxy disables the cache for any plugin that generates internal requests and goes through the same remap rule. For example, background fetch plugin which generates a internal request doesn't cache internal requests.

If there is a requirement to disable caching, it can be achieved by using the conf_remap plugin.  

Update: retained the existing behavior and added a variable to enable caching


zwoop (changes requested):
How does this work to make sure it doesn’t serve the auth request via a cached response?

vmamidi:
> How does this work to make sure it doesn’t serve the auth request via a cached response?

The plugin doesn't make sure that auth requests are not served from the cache. Disabling cache in auth proxy for existing remap( existing behavior) also doesn't guarantee that as auth requests may go to a different remap rule.

bryancall:
[approve ci]

vmamidi:
> > How does this work to make sure it doesn’t serve the auth request via a cached response?
> 
> The plugin doesn't make sure that auth requests are not served from the cache. Disabling cache in auth proxy for existing remap( existing behavior) also doesn't guarantee that as auth requests may go to a different remap rule.

Added configuration to enable/disable caching.

vmamidi:
[approve ci centos}

vmamidi:
[approve ci centos]

vmamidi:
[approve ci]

vmamidi:
[approve ci]

bryancall (changes requested):
please update the docs for the PR

vmamidi:
> please update the docs for the PR

updated.

vmamidi:
[approve ci]

bryancall (approved):

zwoop:
Cherry-picked to v9.2.x

Issue #8265, opened 2021-08-19
Memory leaks in Http2Stream


duke8253:
As described in #8260 , we've been observing high memory utilization in production after upgrading to ats9.1. JeMalloc profiling shows the problem is related to Http2Stream.

Here's what I've collected using jemalloc, the graph ignores the memory allocated to iobuffers. After running for 3 hours or so, ats9.1's memory usage is about 6-7GB more than what ats9.0 is using, and this continues to grow till the host eventually runs out of memory.

ats9.0
[n7.pdf](https://github.com/apache/trafficserver/files/7017105/n7.pdf)

ats9.1
[n7.pdf](https://github.com/apache/trafficserver/files/7017104/n7.pdf)

As @shinrich mentioned in the PR:

> Setting zombie event, we found a session that had the http2 inactivity timeout trigger. Http2Stream::initiating_close had been called on the stream, but it had not been cleaned up in the 5 minutes of the zombie timeout. We think the problem is that the write_vio had nbytes and ndone set to 0. So a WRITE_COMPLETE was sent to the state machine to clean it up. But there was no real write and the watch for client abort handler really needs an EOS to clean up. This change should case an EOS to be sent in this scenario.

This is what the memory utilization looks like after the fix:

ats9.1 with fix
[n7.pdf](https://github.com/apache/trafficserver/files/7017137/n7.pdf)

duke8253:
Here's the allocator stats, collected a few minutes apart.

```
     Allocated      |        In-Use      | Type Size  |   Free List Name
--------------------|--------------------|------------|----------------------------------
          128188416 |          121688064 |       2048 | memory/hdrStrHeap
          127139840 |          121810944 |       2048 | memory/hdrHeap
             131072 |              75008 |        256 | memory/httpCacheAltAllocator
           44498944 |           44340640 |       3104 | memory/http2StreamAllocator
           38819840 |           38809136 |       3568 | memory/http2ClientSessionAllocator
                  0 |                  0 |        144 | memory/RemapPluginsAlloc
           16625664 |           16154688 |       1056 | memory/httpServerSessionAllocator
           48281600 |           40424800 |       9200 | memory/httpSMAllocator
            1081344 |             759264 |       1056 | memory/http1ClientSessionAllocator
                  0 |                  0 |        160 | memory/socksProxyAllocator
              45056 |              19776 |         32 | memory/MIMEFieldSDKHandle
                  0 |                  0 |        320 | memory/INKVConnAllocator
[...]

PR #8268, opened 2021-08-20
Avoid chunking in the post redirect case

duke8253:
After fixing #8265 we started seeing some crashes happening, this fixes the crash.

### DIFF
--- a/proxy/http/HttpSM.cc
+++ b/proxy/http/HttpSM.cc
@@ -5962,6 +5962,7 @@ HttpSM::do_setup_post_tunnel(HttpVC_t to_vc_type)
     // When redirect in process is true and redirection is enabled
     // add http server as the consumer
     if (post_redirect) {
+      chunked = false;
       HTTP_SM_SET_DEFAULT_HANDLER(&HttpSM::tunnel_handler_for_partial_post);
       tunnel.add_consumer(server_entry->vc, HTTP_TUNNEL_STATIC_PRODUCER, &HttpSM::tunnel_handler_post_server, HT_HTTP_SERVER,
                           "redirect http server post");


shinrich:
The problem is we were marking that the post body should be chunked.  But in the post redirect case, the data is being replayed from a static buffer.  There is no real VC to be doing the chunked calculation against.  So p->do_chunking is not set.  So chunked_buffer_start is not set.  So the call to clone that nullptr causes the crash.

```
[ 00 ] traffic_server      HttpTunnel::producer_run(HttpTunnelProducer*)                ( P_IOBuffer.h:753 )
[ 01 ] traffic_server      HttpTunnel::tunnel_run(HttpTunnelProducer*)                  ( HttpTunnel.cc:720 )
[ 02 ] traffic_server      HttpSM::do_setup_post_tunnel(HttpVC_t)                       ( HttpSM.cc:5923 )
[ 03 ] traffic_server      HttpSM::state_send_server_request_header(int, void*)         ( HttpSM.cc:2197 )
[ 04 ] traffic_server      HttpSM::state_read_server_response_header(int, void*)        ( HttpSM.cc:2004 )
[ 05 ] traffic_server      HttpSM::main_handler(int, void*)                             ( HttpSM.cc:2701 )
[ 06 ] traffic_server      write_signal_and_update(int, UnixNetVConnection*)            ( UnixNetVConnection.cc:117 )
[ 07 ] traffic_server      write_to_net_io(NetHandler*, UnixNetVConnection*, EThread*)  ( UnixNetVConnection.cc:162 )
[ 08 ] traffic_server      NetHandler::process_ready_list()                             ( UnixNet.cc:432 )
[ 09 ] traffic_server      NetHandler::waitForActivity(long)                            ( UnixNet.cc:552 )
[ 10 ] traffic_server      EThread::execute_regular()                                   ( UnixEThread.cc:303 )
[ 11 ] traffic_server      execute                                                      ( UnixEThread.cc:364 )
[ 12 ] traffic_server      EThread::execute()                                           ( UnixEThread.cc:342 )
[ 13 ] traffic_server      spawn_thread_internal                                        ( Thread.cc:92 )
[ 14 ] libpthread-2.17.so  start_thread                                                 ( pthread_create.c:307 )
```

I think in the past (9.0), we were placing transfer-encoding headers on the H2 client requests.  And chunkifying the H2 request data before it came to the HttpSM.  With 9.1, we cleaned up that logic.

masaori335:
Any idea of why this AuTest doesn't cover this case? Can we make it to cover?
https://github.com/apache/trafficserver/blob/edef25cc5ddf1355009bdabcdf68b59c88218833/tests/gold_tests/h2/http2.test.py#L184-L192

randall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8269, opened 2021-08-20
Fix leaks in ConfigManager::configName

bneradt:
This fixes an ASan reported leak of ConfigManager::configName. It used
to be strdup'd but not freed in the destructor. This simply changes it
to a std::string.

ASan also reported a leak in AddConfigFilesHere which is fixed with an
ats_free as well.


brbzull0:
Looks good!

brbzull0 (approved):

bneradt:
[approve ci clang-format]

brbzull0 (approved):

traeak (approved):

PR #8271, opened 2021-08-23
Add warning message about why enhance your calm is being sent

shinrich:
Noticed this while debugging H2 to origin.  The warning message was being printed that an ENHANCE_YOUR_CALM error was being returned, but no message about why as is normal for most of the connection/stream errors.


bneradt (commented):
Looks good.

masaori335 (approved):
Looks reasonable.

PR #8272, opened 2021-08-24
Adjust calls to eliminate selecting already used port in tls_hooks test

shinrich:
Looks like tls_hooks11 sometimes picks an already used port for ATS and fails.


randall (approved):

shinrich:
[approve ci autest]

bneradt:
Thanks for looking into these failures. I don't think these changes should make a functional difference however. When present, `StillRunningAfter` should simply cause the test to fail if the processes stop for some reason before the test run completes.

I think we'll rather need to revisit the get_port selection logic.

shinrich:
Agreed.  I don't think these changes are really relevant.  Put this up more to exercise the problem, which of course does not show up locally.

shinrich:
[approve ci autest]

randall (approved):

PR #8274, opened 2021-08-24
Use better done file locations for some AuTests

bneradt:
Some of the AuTests used a done file for AuTest process Ready
conditions. These files got left around in the test directory. This
change puts these done files in the sandbox instead.

Fixes #5546

Issue #6909, opened 2020-06-17
Autest job fails to find the latest commit


maskit:
It seems like autest job can't find the latest commit after force pushes. https://ci.trafficserver.apache.org/job/autest-github/13549/console

```
error: no such commit 36a479333ec3217e4d8ca9f338bee56bb0d6cf8a
```

ywkaras:
It also seems to sometimes use the branch tip from before the forced push.  I think I have seen this happen with other CI jobs besides autest.

shinrich:
I have noticed this too.  Didn't associate it with force pushes, but that is probably when I'm seeing it too.

bneradt:
I've investigated this a bit in the past. To restate the context: when someone submits a PR for review, the desire is for CI to rebase or merge the commits in the PR on top of the latest associated branch before the CI runs. Our Jenkins CI does not implement this by hand. Rather, this mechanism is done via a Jenkins plugin. As many of us have experienced, this plugin seems to have a problem when the commits in the PR change. When these commits change, Jenkins appropriately re-runs the CI, but the plugin seems to have an off-by-one issue in which it runs the CI against the previous commit instead of the latest one. I've experienced this, as @maskit mentioned in the initial commit, with force pushes on the PR's branch. I'm not sure whether it is limited to force pushes. For instance, if I push a second commit on the branch for the PR, does this off-by-one issue happen? I'm not sure since I personally generally force push, keeping the number of my commits to one for a PR, rather than adding commits on top of the initial PR.

I expect that our new CI environment will resolve this. It implements merging onto the latest branch differently. I've explicitly tested this by force pushing commits in our test github environment and observed that it did in fact run with the new commit content, so that's promising.  I'll leave this bug open for now until we transition to the new CI.

PR #8277, opened 2021-08-25
Revert "Use better done file locations for some AuTests"

bneradt:
Reverts apache/trafficserver#8274

---

I got bit by this:

https://github.com/apache/trafficserver/issues/6909

My latest force push on #8274 passed locally and in the PR run (because it ran against the wrong commit) but seems to fail in CI after its merged in. I'll circle back on this PR, but in the meantime I'll unblock PRs by reverting this commit.

### DIFF
--- a/tests/.gitignore
+++ b/tests/.gitignore
@@ -1,1 +1,2 @@
 _sandbox/
+done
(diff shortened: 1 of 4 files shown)


bneradt:
The problem turned out easy enough to fix so I created a PR for it:
https://github.com/apache/trafficserver/pull/8278

PR #8278, opened 2021-08-25
AuTests: Avoid capture_output subprocess command

bneradt:
The subprocess capture_output is a Python 3.9 parameter. This changes
the AuTests that used it to instead use the older stdout/stderr
parameters because that's compatible with Python 3.9.


masaori335 (approved):
I'm not familiar with Subprocess, but this is following the docs.

https://docs.python.org/3/library/subprocess.html
>  If you wish to capture and combine both streams into one, use stdout=PIPE and stderr=STDOUT instead of capture_output.

PR #8281, opened 2021-08-25
In preparation for H2 outbound, split out common Session

shinrich:
Should not change existing functionality, but if this can land soon, I worry less about divergence between 9.2/master and the 10-dev branch which will include H2 to origin logic.

This branch splits out Http2CommonSession from Http2ClientSession.  On the H2 to origin branch, the Http2CommonSession class is also used by the Http2ServerSession.  One of my big concerns about rebasing the H2 to origin branch is that I will lose changes in Http2ClientSession that were moved to Http2CommonSession.

### DIFF
--- a/proxy/http2/Http2ClientSession.cc
+++ b/proxy/http2/Http2ClientSession.cc
@@ -1,6 +1,6 @@
 /** @file
 
-  Http2ClientSession.
+  Http2ClientSession.cc
 
   @section license License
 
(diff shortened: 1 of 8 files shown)


maskit:
Ah, PoolableSession. I feel like having PoolableSession as mix-in makes more sense, but we'll see when we support H3 outbound.

maskit (approved):

PR #8284, opened 2021-08-26
autest: log port # when port is recycled

randall:


randall:
I think someone will need to login to the autest box and add the `--debug` `--verbose` (not `--trace`) to the autest script or the script that invokes autest.sh

bneradt:
[approve ci autest]

bneradt (approved):

PR #8285, opened 2021-08-26
Cleanup generated LDFLAGS for jemalloc

randall:
Use the already expanded jemalloc_ldflags instead of overwriting what
was previously specified earlier in the file.

Also, drop use --add-needed,--no-as-needed

### DIFF
--- a/build/jemalloc.m4
+++ b/build/jemalloc.m4
@@ -55,8 +55,7 @@ if test "$enable_jemalloc" != "no"; then
   jemalloc_have_libs=0
   if test "$jemalloc_base_dir" != "/usr"; then
     TS_ADDTO(CPPFLAGS, [-I${jemalloc_include}])
-    TS_ADDTO(LDFLAGS, [-L${jemalloc_ldflags}])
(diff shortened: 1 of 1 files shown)


bneradt:
[approve ci autest]

randall:
[approve ci autest]

build/jemalloc.m4
```diff
@@ -56,7 +56,7 @@ if test "$enable_jemalloc" != "no"; then
   if test "$jemalloc_base_dir" != "/usr"; then
     TS_ADDTO(CPPFLAGS, [-I${jemalloc_include}])
     TS_ADDTO(LDFLAGS, [-L${jemalloc_ldflags}])
```
duke8253, line 58: Should we remove this line then? Looks like a duplicate

duke8253 (changes requested):

build/jemalloc.m4
randall: good catch!

duke8253 (approved):

zwoop:

PR #8292, opened 2021-08-30
autest: double the initial port pool for get_port

bneradt:
We're running into issues with port selection with our tests. This seems
to be an issue with recycled ports. This temporarily expands the number
of ports in the port pool to avert this problem while we investigate
this issue further.

### DIFF
--- a/tests/gold_tests/autest-site/ports.py
+++ b/tests/gold_tests/autest-site/ports.py
@@ -117,7 +117,7 @@ def _get_available_port(queue):
     return port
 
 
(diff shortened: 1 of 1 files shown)


randall (approved):
Good call!

bneradt:
> ... This temporarily expands the number
> of ports in the port pool to avert this problem while we investigate
> this issue further.

Let's revert this temporary stop gap measure once #8296 is merged in.

bneradt:
Reverted via #8300.

PR #8296, opened 2021-08-30
Make port queue an ordered set

bneradt:
Our port queue wound up having duplicates in it. We never want
duplicates in our port queue as this results in multiple processes
trying to use the same port leading to EADDRINUSE errors. Since unique
port entries is a requirement, this changes the port queue to be an
ordered set queue which, by design, will only contain unique port
values.

### DIFF
--- a/tests/gold_tests/autest-site/ports.py
+++ b/tests/gold_tests/autest-site/ports.py
@@ -23,10 +23,8 @@ import platform
 
 import hosts.output as host
 
-try:
-    import queue as Queue
-except ImportError:
-    import Queue
+from ordered_set_queue import OrderedSetQueue
+
 
 g_ports = None  # ports we can use
 
@@ -126,7 +124,7 @@ def _setup_port_queue(amount=2000):
         host.WriteDebug(
             '_setup_port_queue',
             "Populating the port queue.")
-        g_ports = Queue.Queue()
+        g_ports = OrderedSetQueue()
     else:
         # The queue has already been populated.
         host.WriteDebug(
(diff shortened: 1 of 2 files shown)


tests/gold_tests/autest-site/ports.py
```diff
@@ -117,7 +115,7 @@ def _get_available_port(queue):
     return port
 
 
-def _setup_port_queue(amount=2000):
+def _setup_port_queue(amount=1000):
```
bneradt, line 118: This we will not commit here. This is only for testing that this change helps the test run better.

bneradt:
[approve ci autest]

bneradt:
[approve ci autest]

bneradt:
[approve ci autest]

bneradt:
### Testing Notes:
I initially created this PR which changes the port queue to be a unique set as a draft PR. I included with that draft a reversion of the expansion of the port queue size (see #8292). Thus the initial port queue size was 1000 instead of 2000, as it was when we were seeing test failures. As a draft PR, I ran the autests for this PR four times and the tests passed all four times. I conclude that making the port container a unique set of ports fixes the issue with the recycled ports.

I'll now update this PR from a draft to a regular PR. In doing so, I'll remove the commit that reverted #8292 (the expansion of the initial port queue size) and make reverting the initial port queue size a separate PR.

randall (approved):

bneradt:
[approve ci autest]

shinrich:
[approve ci autest]

bneradt:
This should port everywhere cleanly and will alleviate issues with port recycling for autest. Therefore I'm marking it for our supported releases.

PR #8301, opened 2021-09-01
Make separate read and write vc_handlers

shinrich:
I have noticed occasional crashes in working with the H2 to origin branch and in our 9.1 with producer_run().  Both branches include PR #7976.  With this change, we need to be ready to respond to both read and write events on the server_entry->vc (read for responses and write for post bodies).  Currently we do a vio check in both state_read_server_response_header and state_send_server_request_header to hand the incorrectly passed read/write to othe other side.

In general with the vc_entry we have parallel entries for read and write operations, except for the vc_handler.  This does result in some write events unexpectedly showing up in a read handler and visa versa which ends up with confusing code.

This PR, splits the vc_handler into vc_read_handler and vc_write_handler.  That should hopefully simplify some of handler logic moving forward.

### DIFF
--- a/proxy/http/HttpSM.cc
+++ b/proxy/http/HttpSM.cc
@@ -1998,8 +1998,6 @@ HttpSM::state_read_server_response_header(int event, void *data)
   // a WRITE event appear after receiving EOS from the server connection
   if (server_entry->eos) {
     return 0;
-  } else if (data == server_entry->write_vio) {
-    return this->state_send_server_request_header(event, data);
   }
 
   ink_assert(server_entry->eos != true);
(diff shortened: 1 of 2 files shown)


bryancall:
@bneradt is going to take a look to see if we are running this in prod and will review

bneradt (approved):
Looks fine. I've run with this on a prod box for a day and it ran well. We are currently deploying it across 5 hosts.

zwoop:
Reverted from 9.2.x

bneradt:
I notice this crash on 9.2.x even though it has the patch for this PR locally. I don't think this PR is the cause of that crash therefore

I don't mind that this got removed from 9.2.x, but we need to be aware of a couple things:

1. This crash happens without this PR so this PR is likely not the cause of this crash.
2. HTTP/2 to origin relies upon the functionality added by this PR. So we cannot remove this from 10.x. We'll have to fix this crash without simply reverting this patch.

Here's a branch with an autest that reproduces this crash on 9.2.x even though this PR is reverted:
https://github.com/bneradt/trafficserver/tree/9229_reproduction

PR #8303, opened 2021-09-01
Truly treat multiple certs with the same SAN as a warning

randall:


randall:
This code path leads to a warning https://github.com/apache/trafficserver/pull/8303/files#diff-cb865c0bc65fb8ef103a206282b78a50e0c1c9e93ca6713322ba9df9d921e6c9R1709 . I just want to ensure that the warning is treated as such. Otherwise, we should make that log line an error.

shinrich (approved):

randall:
@zwoop #8308 is the cherry-pick for 9.1.x

PR #8306, opened 2021-09-02
Fix crash during raw connect failures

bneradt:
In the case of raw open, the server_txn will generally be a nullptr.  We
tried to use server_txn when calling set_connect_fail, which resulted in
a crash. This change simply removes the call. Note that a generic call
to set_connect_fail was already called previous to calling this
function.

This error was introduced with PR #7849


shinrich:
Looks good to me.  Might be nice to add back the particulars of the errno at some point, but for the raw open the data argument may be the netvc structure or the lerrno directly.

randall (approved):

PR #8307, opened 2021-09-02
statichit: allow exact path matching to be optional

randall:


bneradt (approved):
I haven't used this plugin, but what you're doing makes sense.

zwoop:

PR #8309, opened 2021-09-03
Fix a compile error in SSLCertLookup

maskit:
```
  CXX      SSLCertLookup.o
SSLCertLookup.cc: In member function ?int SSLCertLookup::insert(const char*, const SSLCertContext&)?:
SSLCertLookup.cc:364:3: error: control reaches end of non-void function [-Werror=return-type]
  364 |   default:
      |   ^~~~~~~
```
Found on Fedora 34 with gcc-11 and BoringSSL.

The code was introduce by #8014. There's no need to backport this to 9.1.x.

### DIFF
--- a/iocore/net/SSLCertLookup.cc
+++ b/iocore/net/SSLCertLookup.cc
@@ -363,6 +363,7 @@ SSLCertLookup::insert(const char *name, SSLCertContext const &cc)
     return this->ec_storage->insert(name, cc);
   default:
     ink_assert(false);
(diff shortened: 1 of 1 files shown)


shinrich (approved):

bneradt (approved):

zwoop:
Possibly this depends on #8309 and or #8014. We intentionally do not want #8014 into 9.1.x, being a significant new feature.

PR #8312, opened 2021-09-03
Fix corner cases in rate limiting plugin

elsloo:
* Fixes an issue that led to `release()` being called too frequently

* Fixes a crash that occurs when the SNI string is `nullptr` and length is non-zero


zwoop (approved):

zwoop:

PR #8313, opened 2021-09-03
Adds new API: TSSslSNIGet

randall:
converts rate_limit to use this API instead of calling OpenSSL APIs,
allowing this plugin to work with BoringSSL


randall:
[approve ci autest]

zwoop:
Cherry-picked to v9.2.x

Issue #8253, opened 2021-08-16
trafficserver 9.1.0 does not build with yaml-cpp 0.7.0


chenrui333:
👋  it like the trafficserver does not build with yaml-cpp 0.7.0

relates to https://github.com/Homebrew/homebrew-core/pull/82097

bryancall:
@chenrui333 Can you please put in the issue the commands you are running for the build and what the output error is?

chenrui333:
Here are the commands:
- `./configure --prefix=/usr/local/Cellar/trafficserver/9.1.0_1 --mandir=/usr/local/Cellar/trafficserver/9.1.0_1/share/man --localstatedir=/usr/local/var --sysconfdir=/usr/local/etc/trafficserver --with-openssl=/usr/local/opt/openssl@1.1 --with-yaml-cpp=/usr/local/opt/yaml-cpp --with-group=admin --disable-silent-rules --enable-experimental-plugins`
- `make install`

configure option summary

```
configure: Build option summary:
    CC:                 clang
    CXX:                clang++
    CPP:                clang -E
    CFLAGS:             
    CXXFLAGS:           
    CPPFLAGS:           -D_GNU_SOURCE -I/usr/local/opt/openssl@1.1/include -DOPENSSL_NO_SSL_INTERN -I/usr/local/Cellar/pcre/8.45/include
    LDFLAGS:            -L/usr/local/opt/openssl@1.1/lib -L/usr/local/Cellar/pcre/8.45/lib
    AM_CFLAGS:          -std=gnu99 -g -pipe -Wall -Wno-deprecated-declarations -Qunused-arguments -Wextra -Wno-ignored-qualifiers -Wno-unused-parameter -O3 -fno-strict-aliasing
    AM_CXXFLAGS:        -std=c++17 -g -pipe -Wall -Wno-deprecated-declarations -Qunused-arguments -Wextra -Wno-ignored-qualifiers -Wno-unused-parameter -O3 -fno-strict-aliasing -Wno-invalid-offsetof
    AM_CPPFLAGS:        -Ddarwin -D_LARGEFILE64_SOURCE=1 -D_COMPILE64BIT_SOURCE=1 -D_REENTRANT -D__STDC_LIMIT_MACROS=1 -D__STDC_FORMAT_MACROS=1
[...]

PR #7948, opened 2021-06-16
Improve parsing error messages for strategies.yaml.

ywkaras:
This PR primarily flags errors where an unrecognized key (perhaps due to a misspelling) appears in a YAML.

Error reporting could be further improved by including line numbers with errors whenever possible.
Unfortunately, the yamlcpp library seems to have a bug where it doesn't count comment lines when giving
a line number.

PR #8319, opened 2021-09-10
Fix yamlcpp include folder by using the YAMLCPP_INCLUDE  variable

brbzull0:
Fix yamlcpp include folder by using the YAMLCPP_INCLUDE  variable,
so if we configure our own version of yamlcpp then the right include files will be picked up.
This will avoid mixin up the internal and the configured yamlcpp library.

### DIFF
--- a/mgmt/Makefile.am
+++ b/mgmt/Makefile.am
@@ -34,7 +34,7 @@ AM_CPPFLAGS += \
 	-I$(abs_top_srcdir)/proxy \
 	-I$(abs_top_srcdir)/proxy/http \
 	-I$(abs_top_srcdir)/proxy/hdrs \
-	-I$(abs_top_srcdir)/lib/yamlcpp/include \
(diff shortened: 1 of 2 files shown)


brbzull0:
@bryancall I think you are looking at this https://github.com/apache/trafficserver/issues/8253 .  I had this same issues locally in my Linux box and this fixed it. Haven't try on mac but I am pretty confident that this will fix the issue.

randall (approved):

bryancall (approved):

bryancall:
This changed fixes issue #8253

zwoop:
Do we really need this in 9.1.x ? I think this fixes something that was introduced in #7948, which is not in 9.1.x (but will be in 9.2.x)

randall:
One half the patch is from #7948, but the other half is from 9db871a399 , so it will conflict. But I think this should go in since it blocks people from using newer out of tree yaml-cpp in 9.1.x

PR #8323, opened 2021-09-13
Add current active SNI Routing Tunnel stats

masaori335:
# Motivation

Part of #7241. All [SNI Routing](https://docs.trafficserver.apache.org/en/9.0.x/admin-guide/layer-4-routing.en.html) tunnels are counted as "active" connections because we don't know the state of the protocol on the tunnel. However, these tunnels are actually "inactive" in some cases and we want to count only "active" tunnels.

# New stats
```
proxy.process.tunnel.current_active_connections
```

A gauge of current active SNI Routing Tunnels.

# New config
```
proxy.config.tunnel.activity_check_period
```

Frequency of checking the activity of SNI Routing Tunnel.


bneradt (changes requested):
Looks good @masaori335 . I just have a few minor suggestions.

masaori335:
@bneradt Thanks for the review. I followed your suggestions. Please take another look.

masaori335:
[approve ci autest]

bneradt (approved):
Looks great. Thank you Masaori.

PR #8327, opened 2021-09-15
Make fqdn of sni.yaml case insensitive.

bneradt:
SNI and fqdn of sni.yaml should be matched case insensitvely. This
changes the matching mechanism for these to be case insensitive.


randall:
[approve ci autest]

bneradt:
I realized there's another code path I need to cover here. I made the SNI fqdn pcre lower case, but the incoming servername that matches it must be made lowercase too. I'll work on a tweak to this and update this PR.

PR #7302, opened 2020-10-28
Remove the queuing option from proxy.config.http.per_server.connection

shinrich:
The PR removes unused (or lightly used) queuing feature for the proxy.config.http.per_server.connection settings.   The queueing feature lets you queue of a fixed number of requests to origin once the max has been met.  Not clear at all how valuable this feature is.  I think it was added because the previous mechanism, would send all excess requests into the future in a similar kind of delay but hopefully things are better in the future approach.

Remove this feature, cleans up some logic and removes some configuration knobs that are probably just confusing.

Noticed this too while working through the outbound connection logic refactoring.

PR #8328, opened 2021-09-16
Fix missing outbound_conntrack config assigment(from master config to global parameter config)

brbzull0:
While doing some tests with `proxy.config.http.per_server.connection.queue_size` I've noticed that the `outbound_conntrack` never gets updated with the configured records config values, without this change all the following records wont never reflect the configured values(in the `records.config`) and the defaults values will be used instead.  

```
proxy.config.http.per_server.connection.queue_size
proxy.config.http.per_server.connection.alert_delay
proxy.config.http.per_server.connection.queue_delay
```

Using the default queue_size will make us block even if expressed otherwise(by configuring the above values):
https://github.com/apache/trafficserver/blob/68926aeeec230ecb4b8408bcefabfcf9a95b3bca/proxy/http/HttpSM.cc#L5233-L5255

Now, I think this will be removed in 10(https://github.com/apache/trafficserver/pull/7302), but at last it should work in 9. 

Closes https://github.com/apache/trafficserver/issues/8340

### DIFF
--- a/proxy/http/HttpConfig.cc
+++ b/proxy/http/HttpConfig.cc
@@ -1437,6 +1437,8 @@ HttpConfig::reconfigure()
   params->server_max_connections    = m_master.server_max_connections;
   params->max_websocket_connections = m_master.max_websocket_connections;
   params->oride.outbound_conntrack  = m_master.oride.outbound_conntrack;
+  params->outbound_conntrack        = m_master.outbound_conntrack;
+
   // If queuing for outbound connection tracking is enabled without enabling max connections, it is meaningless, so we'll warn
   if (params->outbound_conntrack.queue_size > 0 &&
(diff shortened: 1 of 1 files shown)


zwoop:
I'm somewhat confused here, because the code already does this, but in the correct overridable portion of the HttpConfig:

```
  params->oride.outbound_conntrack  = m_master.oride.outbound_conntrack;
```

I'm wondering if there's something else broken with this outbound_conntrack making it not behave? Your patch would "copy" the config, but it copies it to the wrong section (not making it overridable). @SolidWallOfCode  ?

zwoop (changes requested):
I'm gonna mark this "needs changes" for now, because something is fishy here, and I don't want us to accidentally merge this.

zwoop (approved):

PR #8331, opened 2021-09-17
Change the function signature of safe_[get|set]sockopt

maskit:
The type of `optval` argument for [get|set]sockopt is `void*`, but not `char*`.
Some values don't fit into a `char`, and casting the pointer to `char*` causes compile error.

```
       int getsockopt(int sockfd, int level, int optname,
                      void *optval, socklen_t *optlen);
       int setsockopt(int sockfd, int level, int optname,
                      const void *optval, socklen_t optlen);
```

### DIFF
--- a/include/tscore/ink_sock.h
+++ b/include/tscore/ink_sock.h
@@ -34,7 +34,7 @@
 
 #include "tscore/ink_apidefs.h"
 
-int safe_setsockopt(int s, int level, int optname, void *optval, int optlevel);
+int safe_setsockopt(int s, int level, int optname, const void *optval, int optlevel);
 int safe_getsockopt(int s, int level, int optname, void *optval, int *optlevel);
 int safe_bind(int s, struct sockaddr const *name, int namelen);
 int safe_listen(int s, int backlog);
--- a/include/tscore/ink_sock.h
+++ b/include/tscore/ink_sock.h
@@ -34,8 +34,8 @@
 
 #include "tscore/ink_apidefs.h"
 
-int safe_setsockopt(int s, int level, int optname, char *optval, int optlevel);
-int safe_getsockopt(int s, int level, int optname, char *optval, int *optlevel);
+int safe_setsockopt(int s, int level, int optname, void *optval, int optlevel);
+int safe_getsockopt(int s, int level, int optname, void *optval, int *optlevel);
 int safe_bind(int s, struct sockaddr const *name, int namelen);
 int safe_listen(int s, int backlog);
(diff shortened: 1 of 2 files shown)


bneradt (approved):

include/tscore/ink_sock.h
```diff
@@ -34,8 +34,8 @@
 
 #include "tscore/ink_apidefs.h"
 
-int safe_setsockopt(int s, int level, int optname, char *optval, int optlevel);
-int safe_getsockopt(int s, int level, int optname, char *optval, int *optlevel);
+int safe_setsockopt(int s, int level, int optname, void *optval, int optlevel);
+int safe_getsockopt(int s, int level, int optname, void *optval, int *optlevel);
```
bryancall, line 38: If you are going to mirror the setsockopt API then shouldn't it be const void * and socklen_t for the length?

bryancall (changes requested):

include/tscore/ink_sock.h
maskit: I kept my change minimal.

If I change `int *` to `socklen_t *` I have to deal with the error below at all places that call `safe_getsockopt`. I could just reinterpret_cast at all the places but I guess some of the places can declare variables as `unsigned int` and don't need any casts. I don't want to check and fix that.
```
/Users/mkitajo/src/github.com/trafficserver/iocore/eventsystem/P_UnixSocketManager.h:482:11: error: no matching function for call to 'safe_getsockopt'
  r     = safe_getsockopt(s, SOL_SOCKET, SO_SNDBUF, (char *)&bsz, &bszsz);
          ^~~~~~~~~~~~~~~
../../include/tscore/ink_sock.h:38:5: note: candidate function not viable: no known conversion from 'int *' to 'socklen_t *' (aka 'unsigned int *') for 5th argument
int safe_getsockopt(int s, int level, int optname, void *optval, socklen_t *optlevel);
    ^
```

I added `const` for `optval` for now.

bneradt (approved):
Sounds good to me. This is an improvement.

randall:
[approve ci]

maskit:
@bryancall Can you take another look? Changing `char *` to `void *` solves an issue, but I don't see any real issue on the `int *` type.

maskit:
@bryancall I left this open for you because your comment was a change request, but I'm going to merge this with Brian's approval next week if there's no response.

zwoop:
Cherry-picked to v9.2.x

PR #8332, opened 2021-09-17
docs: Update host_sni_policy for some behaviors

bneradt:
This expands the documentation for a few behaviors of the
`host_sni_policy` feature that can be unexpected for users.  Namely:

* `host_sni_policy` is keyed off of the Host header field instead of the
SNI value in the handshake.
* The records.config global `host_sni_policy` only applies when there is
a relevant security policy set for the host in the sni.yaml file but no
specific `host_sni_policy` is set there.


zwoop:
[approve ci].

randall (approved):

PR #8334, opened 2021-09-20
Cleanup VSCode include directories

zwoop:
This also adds support for a ${workspace}/BUILD directory, which should be
a symlink to an out-of-tree build tree. This helps VSCode finding those
generated include files it will need for symbol resolution to function.

Note that you don't have to have a BUILD symlink, if you build in-tree.


zwoop:
Closing, something goes horribly wrong when trying to land this config change as-is.

PR #8344, opened 2021-09-22
[doc] Add a note for TSLifecycleHookAdd.

brbzull0:
Warn users that a `contp` could eventually be executed in a `ET_NET` when it was originally scheduled in the `ET_TASK`.

### DIFF
--- a/doc/developer-guide/api/functions/TSContScheduleOnPool.en.rst
+++ b/doc/developer-guide/api/functions/TSContScheduleOnPool.en.rst
@@ -57,6 +57,11 @@ impact system performance by blocking other continuations on the threads.
 Note that the TSContSchedule() family of API shall only be called from an ATS EThread.
 Calling it from raw non-EThreads can result in unpredictable behavior.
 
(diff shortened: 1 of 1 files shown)


maskit:
Although having the note is definitely good, shouldn't we prevent the scheduling if task threads are not ready?

maskit (approved):
Behavior change may be considered as an incompatible change and we'd have to wait for 10.0. The note is necessary for 9.x in any cases.

brbzull0:
> Although having the note is definitely good, shouldn't we (the API) prevent the scheduling if task threads are not ready?

Yes, this is something that we have talked with @SolidWallOfCode , maybe havig a "pre-queue" so if a contp gets scheduled on that thread pool, it does get queued if the threads aren't running. This should probably need another thought but in the meantime In any case as you also said, I think a note will help(I hope) any distracted dev to wait till the lifecycle hook is pop up.
Thanks for the review!

maskit:
Yeah, having a pre-queue is another option. I'm fine with it as well. Maybe it's more user friendly than returning an error, but would have a bit of more internal complexity.

zwoop:

Issue #8342, opened 2021-09-21
8.1.2 don't build on Fedora 34 / GCC 11


thelounge-zz:
the only package which can't be rebuilt on F34 and no i don't want TS9 for now :-(

In file included from ../../include/tscore/ink_inet.h:34,
                 from ../../include/tscore/Diags.h:41,
                 from P_RecUtils.h:26,
                 from RecConfigParse.cc:34:
../../include/tscore/BufferWriterForward.h:63:28: error: 'numeric_limits' is not a member of 'std'
   63 |   unsigned int _max = std::numeric_limits<unsigned int>::max(); ///< Maxium width
      |                            ^~~~~~~~~~~~~~
../../include/tscore/BufferWriterForward.h:63:43: error: expected primary-expression before 'unsigned'
   63 |   unsigned int _max = std::numeric_limits<unsigned int>::max(); ///< Maxium width
      |                                           ^~~~~~~~
../../include/tscore/BufferWriterForward.h: In constructor 'constexpr ts::BWFSpec::BWFSpec()':
../../include/tscore/BufferWriterForward.h:44:24: error: member 'ts::BWFSpec::_max' must be initialized by mem-initializer in 'constexpr' constructor
   44 |   constexpr BWFSpec() {}
      |                        ^
../../include/tscore/BufferWriterForward.h:63:16: note: declared here
   63 |   unsigned int _max = std::numeric_limits<unsigned int>::max(); ///< Maxium width
      |                ^~~~
make[2]: *** [Makefile:719: RecConfigParse.o] Error 1
make[2]: *** Waiting for unfinished jobs....
In file included from ../../include/tscore/ink_inet.h:34,
                 from ../../include/tscore/Diags.h:41,
                 from P_RecUtils.h:26,
                 from RecFile.cc:28:
../../include/tscore/BufferWriterForward.h:63:28: error: 'numeric_limits' is not a member of 'std'
   63 |   unsigned int _max = std::numeric_limits<unsigned int>::max(); ///< Maxium width
      |                            ^~~~~~~~~~~~~~
../../include/tscore/BufferWriterForward.h:63:43: error: expected primary-expression before 'unsigned'
   63 |   unsigned int _max = std::numeric_limits<unsigned int>::max(); ///< Maxium width
      |                                           ^~~~~~~~
../../include/tscore/BufferWriterForward.h: In constructor 'constexpr ts::BWFSpec::BWFSpec()':
[...]

PR #8345, opened 2021-09-22
Add <limits> include to BufferWriterForward.h

bneradt:
Adding the include of <limits> in BufferWriterForward.h because it uses
numeric_limits. Certain compilers require this.

Fixes: #8342

---

I reproduced the #8342 build failure locally on the 8.1.x branch without this change, and verified that this patch fixes Fedora 34 GCC 11 builds.

### DIFF
--- a/include/tscore/BufferWriterForward.h
+++ b/include/tscore/BufferWriterForward.h
@@ -24,6 +24,7 @@
 #pragma once
 
 #include <cstdlib>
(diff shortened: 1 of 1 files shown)


randall (approved):

bneradt:
Created https://github.com/apache/trafficserver/pull/8347 for 8.1.x

PR #8346, opened 2021-09-22
unit_tests/test_MemArena: fix memory leak on arena.make tests

traeak:
Found using --enable-lsan on make check


bneradt:
Doesn't the arena release the memory itself on destruction?

PR #8350, opened 2021-09-23
slice plugin to remove X-Slicer-Info header after it has been detected

traeak:
This removes the slice plugin special header so that it is not passed upstream.

This needs to account for both the loopback to same remap rule and also for the split remap rule case.


randall:
[approve ci clang-format]

PR #8353, opened 2021-09-23
remove unused RecConfigFileEntry struct

randall:
Follow up to e52009cee1ccccff154b2dc3955697d2adb6bb05


traeak (approved):
looks good thanks for catching that.

PR #8354, opened 2021-09-23
Adds support for TCP_NOTSENT_LOWAT sockopt

randall:


randall:
[approve ci]

randall:
[approve ci docs]

mgmt/RecordsConfig.cc
```diff
@@ -786,6 +786,8 @@ static const RecordElement RecordsConfig[] =
   ,
   {RECT_CONFIG, "proxy.config.net.sock_mss_in", RECD_INT, "0", RECU_RESTART_TS, RR_NULL, RECC_NULL, nullptr, RECA_NULL}
   ,
+  {RECT_CONFIG, "proxy.config.net.sock_notsent_lowat", RECD_INT, "16384", RECU_DYNAMIC, RR_NULL, RECC_NULL, nullptr, RECA_NULL}
```
zwoop, line 789: Should this really be the default? I'm thinking, for now at least, shouldn't this be default "off" ?

randall: It's off by default (aka don't do setsockopt), but if you do enable via the masks (adding 64 to `proxy.config.net.sock_option_flag_in` record), it would enable it with the default value from here (or whatever is set in `records.config`)

zwoop: Ah, duh, thanks! Wonder if the default should be 32KB to fit in with some other "defaults" that we have at 32KB?

randall: done!

bneradt (approved):
Looks good to me.

PR #8355, opened 2021-09-24
remap_stats: ensure hostname is deleted in txn_close

traeak:
The logic left a hole where hostname might not be deallocated.  Closes this hole.
Also reenables the autest by using a polling shell script on the traffic_ctl metrics match call.


zwoop:
Please remember to remove Projects and Milestones when closing a PR.

PR #8357, opened 2021-09-27
test_MMH: fix memory leak in unit test

traeak:
Discovered while running valgrind on test_tscore


zwoop:
Cherry-picked to v9.2.x

PR #8360, opened 2021-09-27
Stabilize regex_revalidate Au test.

ywkaras:
Delay sufficiently between writes of config file so the differences in the write times will be greater
than the granualirity of the timestamp used by the plugin.

(cherry picked from commit 9031965889ab430bd182dd5485ecda451ecb6a8a)


ywkaras:
This only changes an Au test.  This change has already been made in Yahoo internal 9.1.  The problem it fixes was seen frequently in Yahoo Au test regression runs.  I don't remember seeing it in Apache, but make regression more robust, and helps Yahoo to minimize deviation from open source version.

randall (approved):

zwoop:

PR #8362, opened 2021-09-28
remap_stats: convert to using TSHttpTxnPristineUrlGet and TSHttpTxnClienReqGet

traeak:
remap_stats: convert to using TSHttpTxnPristineUrlGet and TSHttpTxnClienReqGet only during TXN_CLOSE hook.
remove pre remap continuation
post remap continuation now only sets a value.
also fixes and adds an autest for --post-remap-host
Includes memory leak fix for hostname when the old post_remap hook wasn't hit.

Regression tested for some days using the default (post remap) configuration with cloned prod traffic.  Also flipped to using -P (pristine) and values seem to match.

### DIFF
--- a/plugins/experimental/remap_stats/remap_stats.cc
+++ b/plugins/experimental/remap_stats/remap_stats.cc
@@ -34,11 +34,16 @@
 
 #define MAX_STAT_LENGTH (1 << 8)
 
+enum UriType {
+  REMAP,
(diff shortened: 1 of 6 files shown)


randall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8363, opened 2021-09-29
Added support for promoting internal (plugin-initiated) requests.

elsloo:
* Allows an operator to specify whether or not internal requests are considered for promotion through a configuration option


bneradt (approved):
Looks reasonable to me. As a disclaimer, though, we do not use this plugin.

PR #8364, opened 2021-09-29
Added support for verifying cacheability before attempting to force an object into cache.

elsloo:
* Intended to be used in conjunction with other plugins, such as cache promote (see #8363).


bneradt (approved):
Looks reasonable.

zwoop:

Issue #8329, opened 2021-09-16
Crash while using next hop strategy


meeramn:
Strategies yaml config got loaded but seeing crash after sending curl request. 

Please find the configuration below.

**cat remap.config**
map /nexthop-test/ http://httpbin.org/ @strategy=nexthop-test-strategy

**cat strategies.yaml**
hosts:
  - &nexthop-test-host
    host: test-host.xyz.com
    protocol:
      - scheme: http
        port: 80
      - scheme: https
        port: 443

groups:
  - &nexthop-test-group
    - <<: *nexthop-test-host
      weight: 1.5

strategies:
  - strategy: 'nexthop-test-strategy'
    policy: consistent_hash
    hash_key: path
    go_direct: false
    groups:
      - *nexthop-test-group
    scheme: http
    failover:
      ring_mode: alternate_ring
      response_codes:
        - 404
      markdown_codes:
        - 503
      health_check:
        - passive

**Curl** 
curl -vvv -o /dev/null http://localhost/nexthop-test/get
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0* About to connect() to localhost port 80 (#0)
*   Trying ::1...
* Connected to localhost (::1) port 80 (#0)
> GET /nexthop-test/get HTTP/1.1
> User-Agent: curl/7.29.0
> Host: localhost
> Accept: */*
>
  0     0    0     0    0     0      0      0 --:--:--  0:00:34 --:--:--     0* Empty reply from server
  0     0    0     0    0     0      0      0 --:--:--  0:00:34 --:--:--     0
* Connection #0 to host localhost left intact
curl: (52) Empty reply from server

**Stack trace**

traffic_server: received signal 11 (Segmentation fault)
traffic_server - STACK TRACE:
/opt/bazinga/bin/traffic_server(_Z19crash_logger_invokeiP9siginfo_tPv+0x94)[0x55edd0cc3524]
/lib64/libpthread.so.0(+0xf630)[0x7f8c4352f630]
[...]

PR #8365, opened 2021-09-29
Fixes issue #8329 crash in NextHopConsistentHash where when the only

jrushford:
Fixes issue #8329 crash in NextHopConsistentHash where when the only
host in a strategy is unavailable due to DNS lookup failures.

- refactored NextHopConsistentHash::findNextHop() to fix the crash
  and simplify it.
- fixed the unit tests in test_NextHopConsistentHash.cc so that test
  failure checks using strcmp() do not crash with a nullptr hostname
  result.


proxy/http/remap/NextHopConsistentHash.cc
```diff
@@ -214,244 +234,230 @@ NextHopConsistentHash::findNextHop(TSHttpTxn txnp, void *ih, time_t now)
...
+  // parent on the consistent hash ring.  If not first call, then the transaction was unable to use the parent
+  // returned from the "firstcall" due to some error so, subsequent calls will not search using a hash but,
+  // will instead just increment the hash table iterator to find the next parent on the ring
   if (firstcall) {
-    NH_Debug(NH_DEBUG_TAG, "[%" PRIu64 "] firstcall, line_number: %d, result: %s", sm_id, result->line_number,
-             ParentResultStr[result->result]);
-    result->line_number = distance;
-    cur_ring            = 0;
+    NH_Debug(NH_DEBUG_TAG, "[%" PRIu64 "] firstcall, line_number: %d, result: %s", sm_id, result.line_number,
+             ParentResultStr[result.result]);
```
ywkaras, line 268: Because of this change we have to change the 31 lines line this one:  https://github.com/apache/trafficserver/blob/44d1be681245f73e82eee59699ad0660dea91290/tests/gold_tests/next_hop/zzz_strategies_peer2/trace.gold#L3 in the gold file from result->result to result.result .

ywkaras: Also have to change the 48 line like this one:  https://github.com/apache/trafficserver/blob/44d1be681245f73e82eee59699ad0660dea91290/tests/gold_tests/next_hop/zzz_strategies_peer/trace.gold#L3

jrushford: @ywkaras I changed the text of the Debug at line 448 back so that it logs 'result->result'

PR #6782, opened 2020-05-14
NextHop Strategy Refactor and Fixes

rob05c:
This is a collaboration between @jrushford and me. This:
- Refactors the NextHopSelectionStrategy interface to simplify and prepare for Plugins (John's work)
- Refactors Strategy Markdown logic, to simplify, encapsulate, and also prepare for Plugin usage
- Fixes strategies.yaml parent errors with 4xx to not mark down and 5xx to mark down, matching parent.config behavior
- Fixes strategies.yaml markdown to include the port, to not mark down parents with the same hostname and different ports, matching parent.config behavior
- Fixes strategies.yaml ignore_self_detect 
- Fixes  strategies.yaml hash_string override

Includes unit tests for the above fixes.

PR #8367, opened 2021-09-30
Fixes issue #8366, assertion failure in HttpTransact.cc::is_response_valid()

### LINKED ISSUE #8366: Assertion failure in HttpTransact::is_response_valid() when a parent proxy is down.
When testing on a debug master build, a failure to connect to an upstream parent proxy that is down, an assertion crash occurs at HttpTransact::is_response_valid():334.  s->current.state is set to BAD_INCOMING_RESPONSE which is not in the list of current.state values accepted in the ink_assert().  

In 9.1.x, when a parent is down, the current.state comes in as CONNECTION_CLOSED but in master it's now BAD_COMING_RESPONSE.

jrushford:
Fixes #8366 An assertion failure occurs at HttpTransact::is_response_valid() when the current.state is BAD_INCOMING_RESPONSE.  I ran into this while testing Nexthop strategies and parent selection with a down unreachable parent.

This ink_assert() check was added with PR #6782 - NextHopStrategy Refactor and Fixes.


jrushford:
needs more investigation.

PR #8368, opened 2021-10-01
TSSslSecretSet: Update SSL_CTX TLS Secrets

bneradt:
TSSslSecretSet previously updated the in-memory certificate data stored
in SSLSecret, but did not update the stored SSL_CTX certificates. As a
result, any newly created SSL contexts would get the updated secret
while the previously stored contexts would still use the old secrets.
This patch updates TSSslSecretSet to also update the secrets for stored
SSL_CTX objects.

This also fixes the related SSLSecret debug logging which had formatting
issues, rendering their output unreadable.

### DIFF
--- a/iocore/net/SSLConfig.cc
+++ b/iocore/net/SSLConfig.cc
@@ -573,6 +573,9 @@ SSLCertificateConfig::acquire()
 void
 SSLCertificateConfig::release(SSLCertLookup *lookup)
 {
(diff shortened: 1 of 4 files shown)


bneradt:
[approve ci autest]

randall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8373, opened 2021-10-04
Removed references to the throttle option from the slice plugin.

elsloo:
* Removes `--throttle` option from docs
* Removes unused `-o` option from the call to `getopt_long()` in the plugin's configuration code


traeak (approved):
Thanks for catching and cleaning up!

elsloo:
[approve ci autest]

PR #8374, opened 2021-10-04
9.2.x autest fixes

ezelkow1:
When we went to bump our 9.1 build to 9.2 we hit 4 breaking autests

Fixing them in here as I can


bneradt (approved):
Looks good.

As a community (myself included), let's keep our eye out for the over-specification of this Server header.

randall (approved):

PR #8381, opened 2021-10-06
Fix a potential H2 stall

maskit:
I was looking into an H2 stall issue on production, and the cause was in code that hasn't been publicly available yet, but realized that this code has the same issue. Like "no window" case, we need to call `flush()` here, otherwise data can stay in write_buffer forever.


bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8388, opened 2021-10-07
body factory does not respect runroot

dragon512:
Current logic with body_factory is that it will use the prefix value and the compile-time logic for the config dir. When a run_root.yaml file uses a different path structure for the config dir such as config/trafficserver instead of build time etc/trafficserver, Trafficserver will fail to load the body_factory templates. This patch addresses this issues


zwoop:
Cherry-picked to v9.2.x

PR #8395, opened 2021-10-08
Added metrics to the rate limit plugin and document the new options

elsloo:


zwoop:
Cherry-picked to v9.2.x

PR #8429, opened 2021-10-18
Remove empty tests/unit_tests directory+makefile

randall:
This came into being in 79a4045749 but the referenced main.cpp wasn't checked in, nor was it ever hooked into the tests/ Makefile


zwoop:
Cherry-picked to v9.2.x

PR #8430, opened 2021-10-18
tests: Add shbang to python scripts with a main

randall:


zwoop:
Cherry-picked to v9.2.x

Issue #8432, opened 2021-10-19
Compile warnings by LLVM 13


masaori335:
Some warnings show up when I tried the llvm-13.0.0 against the latest master.
```
RecRawStats.cc:580:122: warning: performing pointer subtraction with a null pointer may have undefined behavior [-Wnull-pointer-subtraction]
                                                 (reinterpret_cast<char *>(&reinterpret_cast<RecRecord *>(0)->stat_meta) -
                                                                                                                         ^
1 warning generated.
Makefile:2601: warning: overriding recipe for target 'CompileParseRules'
Makefile:1191: warning: ignoring old recipe for target 'CompileParseRules'
Makefile:2601: warning: overriding recipe for target 'CompileParseRules'
Makefile:1191: warning: ignoring old recipe for target 'CompileParseRules'
In file included from MemArena.cc:27:
In file included from ../../include/tscore/MemArena.h:30:
../../include/tscpp/util/MemSpan.h:239:13: warning: definition of implicit copy assignment operator for 'MemSpan<void>' is deprecated because it has a user-declared copy constructor [-Wdeprecated-copy]
  constexpr MemSpan(self_type const &that) = default;
            ^
MemArena.cc:77:13: note: in implicit copy assignment operator for 'ts::MemSpan<void>' first required here
    zret    = _active->alloc(n);
            ^
1 warning generated.
UnixUDPNet.cc:966:7: warning: variable 'real_len' set but not used [-Wunused-but-set-variable]
  int real_len = 0;
      ^
1 warning generated.
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: libinknet.a(BoringSSLUtils.o) has no symbols
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: libinknet.a(SSLInternal.o) has no symbols
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: libinknet.a(BoringSSLUtils.o) has no symbols
/Library/Developer/CommandLineTools/usr/bin/ranlib: file: libinknet.a(SSLInternal.o) has no symbols
In file included from HTTP.cc:30:
In file included from ./HTTP.h:29:
../../include/tscore/HTTPVersion.h:30:3: warning: definition of implicit copy assignment operator for 'HTTPVersion' is deprecated because it has a user-declared copy constructor [-Wdeprecated-copy]
  HTTPVersion(HTTPVersion const &that) = default;
  ^
[...]

PR #8435, opened 2021-10-19
Fix unused-but-set-variable warnings

masaori335:
Part of #8432


zwoop:
Cherry-picked to v9.2.x

zwoop:
This has a conflict on 9.1.x, @masaori335 if we want this in 9.1.x, can you make a back port PR please?

masaori335:
Roger, #8528 is it.

PR #8438, opened 2021-10-19
Fix traffic_manager build when mime-sanity-check is enabled

jbfavre:
When building ATS 9.1.0 on Debian with `--enable-mime-sanity-check` option, build fails with `undefined reference` error for `mime_hdr_sanity_check` method.


randall:
[approve ci]

zwoop:
Cherry-picked to v9.2.x

PR #8439, opened 2021-10-19
Fix Multiplexer POST Body Handling

bneradt:
The previous Multiplexer Plugin test only verified that the plugin would
load without error. This updates the test to verify it can multiplex to
multiple hosts, exercising both HTTP and HTTPS connections to the
multiplexed hosts.
    
While doing this, I added POST request method tests. This showed that
the plugin could not handle POST bodies and simply failed a
misconfigured assertion when the request body was processed. This change
also fixes that bug in the plugin.


zwoop:
Cherry-picked to v9.2.x

PR #8444, opened 2021-10-21
Exports all symbols for missing_mangled_definition.so test plugin

randall:
This fixes the tests verify_global_plugin and verify_remap_plugin when
using lld and LTO


zwoop:
Cherry-picked to v9.2.x

PR #8445, opened 2021-10-22
Include CPPFLAGS when building CompileParseRules

randall:
This is a follow on to 852d90d3


zwoop:
Cherry-picked to v9.2.x

PR #8450, opened 2021-10-26
Document proxy.config.memory.max_usage

ema:
Throttling based on resident memory usage was implemented in cbd5cc1f.
This commit documents the feature.


zwoop:
Cherry-picked to v9.2.x

PR #8451, opened 2021-10-26
Reject Transfer-Encoding in pre-HTTP/1.1 requests

bneradt:
Per spec, Transfer-Encoding is only supported in HTTP/1.1. For earlier
versions, we must reject Transfer-Encoding rather than interpret it
since downstream proxies may ignore the chunk header and rely upon the
Content-Length, or interpret the body some other way.  These differences
in interpretation may open up the door to compatibility issues. To
protect against this, we reply with a 4xx if the client uses
Transfer-Encoding with HTTP versions that do not support it.


zwoop:
I don't know what's going on here, but the commit ID above is "empty", whereas 6e5070118a does have the changes here. So I've cherry-picked that to 9.2.x.

PR #8452, opened 2021-10-26
Detect and handle chunk header size truncation

bneradt:
This detects if a chunk header size is too large and, if so, closes the
connection.


zwoop:
Cherry-picked to v9.2.x

PR #8453, opened 2021-10-26
Ignore ECONNABORTED on blocking accept

masaori335:
Ignore the error instead of adding delay.


zwoop:
Cherry-picked to v9.2.x

PR #8305, opened 2021-09-02
Revert "Fail requests with incomplete line ends (#8096)"

shinrich:
This reverts commit 2c8bb98b2cb364b1a820236dc83271af082e7434.

PR #8455, opened 2021-10-26
Fix output '\n' HTTP field line endings

bneradt:
This is another attempt to fix what was initially addressed in #8096 but
got backed out via #8305. That more extensive patch was considered too
invasive and potentially risky.  This more targeted patch will fix
clients that only send the \n endings but it will force the \r\n line
ending on output.

This was mostly in place except for header lines that get
m_n_v_raw_printable set, which seems to be most header lines. The
addition checks to see if the header line ends in \r\n. If it does not
the m_n_v_raw_printable flag gets cleared and the logic that explicitly
adds the line endings while be invoked on output.


zwoop:
Cherry-picked to v9.2.x

PR #8462, opened 2021-10-27
ssl_verify_test: clang-analyzer fix to account for null sni_name

traeak:
clang-analyzer was upset due to a debug statement using a char* that could potentially be null.  This should resolve the error.


zwoop:
Cherry-picked to v9.2.x

zwoop:
Thanks for the fix, please remember to mark for the appropriate release branches (10-Dev is not relevant here, since this is not a compatibility issue)

PR #8463, opened 2021-10-27
Propagate proxy.config.net.sock_option_flag_in to newly accepted connections

randall:
This also updates the default `proxy.config.net.sock_option_flag_in` to not include setting the SO_LINGER socket option as it's not a good default (and should only be enabled after careful consideration).

### DIFF
--- a/doc/admin-guide/files/records.config.en.rst
+++ b/doc/admin-guide/files/records.config.en.rst
@@ -4594,7 +4594,7 @@ Sockets
 
    Sets the receive buffer size for connections from the client to |TS|.
 
-.. ts:cv:: CONFIG proxy.config.net.sock_option_flag_in INT 0x5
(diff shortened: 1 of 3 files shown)


bryancall:
I will take this

maskit:
👍 I'm not sure about SO_LINGER, but I needed the propagation too when I tried zero copy.

shinrich:
The propagation seems good.  Does changing the default value make this a breaking change?

shinrich (approved):

randall:
@shinrich Good point; I'll remove from 9.2 consideration.

bryancall:
An email should go to the mailing list and then, Leif and I will talk if we should backport the entire change (default change and propagation) or just the propagation.

zwoop:
Cherry-picked to v9.2.x

PR #8469, opened 2021-10-28
Make factory.response_suppression_mode an overridable config

zwoop:
This changes proxy.config.body_factory.response_suppression_mode to be
an overridable configuration in the Http "state". This can be used to avoid
having body factory populating a body when it's undeseriable.

A typical use case would be e.g. this configuration for HRW:

```
cond %{REMAP_PSEUDO_HOOK}
    set-redirect 302 http://www.example.com
    set-config proxy.config.body_factory.response_suppression_mode 1
```


sudheerv (approved):
LGTM 👍 

Thanks for fixing the regression I introduced!

ezelkow1 (approved):

zwoop:
@sudheerv  No, thank you for fixing the bug :-).

zwoop:
Cherry-picked to v9.2.x

PR #8248, opened 2021-08-13
After parents are exhausted, serve stale content from child

serrislew:
After parents are exhausted, child server calls `handle_parent_died` to build the error response message. Changes made will first check if the stale cache response is returnable. If not, the server will handle the dead server/parent appropriately and build the error response message as previously explained.

Note: this only applies to parent/hierarchical caching. The changes do not affect servers who route directly to audience server

PR #8473, opened 2021-10-28
Serve stale content from child if parent returns invalid/marked down

serrislew:
For hierarchical caching, when a child proxy needs to revalidate its stale content with the parent proxy but the parent returns an invalid response or is marked "unavailable", the child proxy should still return the stale content with a warning header. This is the same behavior without hierarchical caching, when the origin server is down and a server tries to revalidate its stale content and when it fails to, it returns back a 200 OK response with a warning header. 

These changes still respect the `max_stale_age` configurations and do not serve stale content pass the expiration date. 

Changes made:
- when child receives invalid response from parent, it first checks if the stale cache response is returnable. if not, then it calls `handle_parent_died` to build the error response message 
- when the parent is marked down or unavailable (without making a request), it first checks if the stale cache response is returnable before building the error response message
   - the parent can be marked unavailable when the number of retries to the parent exceeds `proxy.config.http.parent_proxy.fail_threshold` within `proxy.config.http.parent_proxy.retry_time` 
   - the `is_server_negative_cached` is removed as the conditional checks with configs only applicable to origin server. parent server status and selection (whether its marking a parent down or deciding the next available parent) are already decided in `ParentSelectionStrategy`

Autest checks the following using a PUSH workaround to mock parent going down:
- Child returns stale content with warning header when object's age is less than `max_stale_age` and parent fail count is below `parent_proxy.fail_threshold`
- Child returns stale content with warning header when object's age is less than `max_stale_age` and parent fail count is passed `parent_proxy.fail_threshold` and parent is marked unavailable
- Child does not return stale content when object's age exceeds `max_stale_age`

Note: this is an updated PR from #8248

### DIFF
--- a/proxy/http/HttpTransact.cc
+++ b/proxy/http/HttpTransact.cc
@@ -2936,7 +2936,7 @@ HttpTransact::HandleCacheOpenReadHit(State *s)
     // a parent lookup could come back as PARENT_FAIL if in parent.config, go_direct == false and
     // there are no available parents (all down).
     else if (s->current.request_to == HOST_NONE && s->parent_result.result == PARENT_FAIL) {
-      if (is_server_negative_cached(s) && response_returnable == true && is_stale_cache_response_returnable(s) == true) {
+      if (response_returnable == true && is_stale_cache_response_returnable(s) == true) {
         server_up = false;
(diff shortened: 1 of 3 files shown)


jrushford (approved):
@serrislew looks good to me, passes tests

zwoop:
Cherry-picked to v9.2.x

PR #8475, opened 2021-10-28
Check length before search accept-encoding header

masaori335:
Fix a bug of the stats-over-http plugin.


zwoop:
Cherry-picked to v9.2.x

PR #8479, opened 2021-10-29
UnixNetVConnection: add check for nh in fail block

traeak:
Ran into this core dump while performing network stress testing.

In UnixNetVConnection::connectUp
if this fails:
```
  if ((res = get_NetHandler(t)->startIO(this)) < 0) {
    goto fail;
  }
```
this->nh is never set. In the fail block check this->nh.

### DIFF
--- a/iocore/net/UnixNetVConnection.cc
+++ b/iocore/net/UnixNetVConnection.cc
@@ -678,7 +678,7 @@ UnixNetVConnection::do_io_close(int alerrno /* = -1 */)
     if (nh) {
       nh->free_netevent(this);
     } else {
-      free(t);
+      this->free(t);
     }
   }
 }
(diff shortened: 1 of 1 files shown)


masaori335 (approved):
Look reasonable. If `NetHandler::startIO` fails, the `nh` isn't set.

traeak:
In tracing through the code the one thing possibly not handled is netevent->free(ethread).

So the code could look like:
```
  if (nullptr != nh) {
    nh->free_netevent(this);
  } else {
    this->free(t);
  }
```

traeak:
Updated to add the "free(t)" (ethread) call in case nh is nullptr.

masaori335 (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8483, opened 2021-11-01
ssl_secret debug printing: print only the first 50 bytes

bneradt:
The TLS secrets are sensitive. Print only the first 50 bytes, even with
the debug tag for the print statements on.

---

Depends upon: #8368


randall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8484, opened 2021-11-01
Serve stale content when DNS lookup fails

serrislew:
Proxies should serve stale content even if DNS lookup to OS is unsuccessful. 
- For hierarchical caching, if a child proxy matches successfully to a parent proxy and child has requested stale object in its cache, it should not need to make a DNS request for the origin.
   - It can validate with the parent and return the object as expected. 
- Without hierarchical caching, if stale content is being requested from a proxy and the DNS lookup to the OS is unsuccessful, it should also serve the stale object
   - this behavior is similar to when the origin server is down and the proxy serves stale content with a warning header

These changes still respect the `max_stale_age` configurations and do not serve stale content pass the expiration date.

Autest tests the following using PUSH command to store stale content in proxy's cache: 
- With child mapping to an origin server without a DNS entry and child matching successfully to parent, a request from child after content becomes stale returns `200`
   - After `max_stale_age` has passed, request from child returns `500` as stale content is no longer returnable
- Without hierarchical caching and proxy mapping to an origin server without a DNS entry, a request from proxy after content becomes stale returns `200`
   - After `max_stale_age` has passed, request from proxy returns `500` as stale content is no longer returnable

### DIFF
--- a/tests/gold_tests/proxy_protocol/proxy_serve_stale_dns_fail.test.py
+++ b/tests/gold_tests/proxy_protocol/proxy_serve_stale_dns_fail.test.py
@@ -28,7 +28,7 @@ Test.testName = "STALE"
 # Config child proxy to route to parent proxy
 ts_child.Disk.records_config.update({
     'proxy.config.url_remap.pristine_host_hdr': 1,
-    'proxy.config.http.cache.max_stale_age': 30,
+    'proxy.config.http.cache.max_stale_age': 10,
     'proxy.config.http.parent_proxy.self_detect': 0,
 })
 ts_child.Disk.parent_config.AddLine(
@@ -41,7 +41,7 @@ ts_child.Disk.remap_config.AddLine(
 # Configure parent proxy
 ts_parent.Disk.records_config.update({
     'proxy.config.url_remap.pristine_host_hdr': 1,
-    'proxy.config.http.cache.max_stale_age': 30,
+    'proxy.config.http.cache.max_stale_age': 10,
 })
 ts_parent.Disk.remap_config.AddLine(
     f'map http://localhost:{ts_parent.Variables.port} {server_name}'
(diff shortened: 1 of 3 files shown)


tests/gold_tests/proxy_protocol/proxy_serve_stale_dns_fail.test.py
```diff
@@ -0,0 +1,79 @@
...
+ts_parent.Disk.remap_config.AddLine(
+    f'map http://localhost:{ts_parent.Variables.port} {server_name}'
+)
+ts_parent.Disk.remap_config.AddLine(
+    f'map {server_name} {server_name}'
+)
+
+# Object to push to proxies
+stale_5 = "HTTP/1.1 200 OK\nServer: ATS/10.0.0\nAccept-Ranges: bytes\nContent-Length: 6\nCache-Control: public, max-age=5\n\nCACHED"
+stale_30 = "HTTP/1.1 200 OK\nServer: ATS/10.0.0\nAccept-Ranges: bytes\nContent-Length: 6\nCache-Control: public, max-age=30\n\nCACHED"
```
ywkaras, line 55: Could this be changed to max-age=10 (or other value less than 30)?  @zwoop may only give you a fruit cake for X-mas if you make Au regression test runs take 30 seconds longer.

tests/gold_tests/proxy_protocol/proxy_serve_stale_dns_fail.test.py
```diff
@@ -0,0 +1,79 @@
...
+# Configure parent proxy
+ts_parent.Disk.records_config.update({
+    'proxy.config.url_remap.pristine_host_hdr': 1,
+    'proxy.config.http.cache.max_stale_age': 10,
+})
+ts_parent.Disk.remap_config.AddLine(
+    f'map http://localhost:{ts_parent.Variables.port} {server_name}'
+)
+ts_parent.Disk.remap_config.AddLine(
+    f'map {server_name} {server_name}'
```
ywkaras, line 50: Why do you need this?

ywkaras:
I forget now, did you say Apple is already running this on production proxies?

tests/gold_tests/proxy_protocol/proxy_serve_stale_dns_fail.test.py
```diff
@@ -0,0 +1,79 @@
...
+Test.testName = "STALE"
+
+# Config child proxy to route to parent proxy
+ts_child.Disk.records_config.update({
+    'proxy.config.url_remap.pristine_host_hdr': 1,
+    'proxy.config.http.cache.max_stale_age': 10,
+    'proxy.config.http.parent_proxy.self_detect': 0,
+})
+ts_child.Disk.parent_config.AddLine(
+    f'dest_domain=. parent=localhost:{ts_parent.Variables.port} round_robin=consistent_hash go_direct=false'
```
ywkaras, line 35: So even with this configured, child TS will still do a DNS lookup on unkown.domain.com ?  Did you see the DNS lookup in the http debug trace output?

tests/gold_tests/proxy_protocol/proxy_serve_stale_dns_fail.test.py
serrislew: I am testing for non-hierarchical scenarios as well so the `ts_parent` server is mapped to the OS and gets called directly

tests/gold_tests/proxy_protocol/proxy_serve_stale_dns_fail.test.py
serrislew: The http debug trace output shows the child doing PPDNS lookup to the parent and the parent doing a DNS lookup to unknown.domain.com (OS). After the DNS lookup fails for the parent and the parent sends a 500 response to the child, the child will first check if the stale object is returnable before sending it back to the client. Let me know if you are seeing something different in the debug trace output!

tests/gold_tests/proxy_protocol/proxy_serve_stale_dns_fail.test.py
```diff
@@ -0,0 +1,79 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+Test.ContinueOnFail = True
+# Set up hierarchical caching processes
+ts_child = Test.MakeATSProcess("ts_child")
+ts_parent = Test.MakeATSProcess("ts_parent")
+server_name = "http://unknown.domain.com/"
+
+Test.testName = "STALE"
```
ywkaras, line 26: Most of our Au tests don't assigned testName, I'm not sure what it's for:
```
wkaras ~/REPOS/TS3/tests/gold_tests
M$ find . -name \*.test.py | wc -l
     239
wkaras ~/REPOS/TS3/tests/gold_tests
M$ fgrep Test.testName $( find . -name \*.test.py ) | wc -l
      44
wkaras ~/REPOS/TS3/tests/gold_tests
M$
```

tests/gold_tests/proxy_protocol/proxy_serve_stale_dns_fail.test.py
ywkaras: Lately we've been including DNS servers in our Au tests:  https://github.com/apache/trafficserver/blob/57749218b297849e26a00ba80ed0f26358190929/tests/gold_tests/next_hop/strategies_ch/strategies_ch.test.py#L71

tests/gold_tests/proxy_protocol/proxy_serve_stale_dns_fail.test.py
ywkaras: Alternatively you can also do:  https://github.com/apache/trafficserver/blob/57749218b297849e26a00ba80ed0f26358190929/tests/gold_tests/pluginTest/test_hooks/test_hooks.test.py#L50

But the way you have it is fine.

tests/gold_tests/proxy_protocol/proxy_serve_stale_dns_fail.test.py
ywkaras: I don't understand why ATS would do a DNS lookup on the request host when it's not using the request host as the next hop.  But I suppose that's beyond the scope of this PR.

ywkaras:
We discussed this in today's meeting, it turns out microdns is trickier to use than I thought, and not well-documented, so approving this as is.

ywkaras (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8489, opened 2021-11-02
Better TLS Secrets Truncation.

bneradt:
This improves upon #8483 with an observation from @ywkaras that since
the data is stored in a std::string it will always be null terminated.
Thus these debug logs can be simplified to just use the "%.50s" print
format.


bneradt:
[approve ci]

bneradt:
[approve ci autest]

randall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8492, opened 2021-11-02
When checking for SSL_set1_verify_cert_store, save/restore LIBS before manipulating LIBS

### LINKED ISSUE #8490: Conflict between --with-openssl and --with-jemalloc build options
While building with `--with-jemalloc`, I keep getting the error message `configure: error: jemalloc has bogus version`. After some digging I found that this started happening after commit https://github.com/apache/trafficserver/commit/3412c4f69552b10779cfe0ce1bc6a71971734491

After that commit, if I specify both `--with-openssl` and `--with-jemalloc` options, I get the error message, if I only specify `--with-jemalloc` everything works as expected.

randall:
Closes #8490


duke8253 (approved):
Yep, this fixes it. Thanks!

zwoop:
Cherry-picked to v9.2.x

PR #8501, opened 2021-11-08
Improve option processing in cache promote

elsloo:
* Fixes segfault scenario when unknown option(s) are encountered

* Changes behavior to ignore unknown options rather than rejecting the entire config and refusing to load

### DIFF
--- a/plugins/cache_promote/configs.cc
+++ b/plugins/cache_promote/configs.cc
@@ -42,7 +42,9 @@ static const struct option longopt[] = {
 // The destructor is responsible for returning the policy to the PolicyManager.
 PromotionConfig::~PromotionConfig()
 {
-  _manager->releasePolicy(_policy);
+  if (_policy) {
+    _manager->releasePolicy(_policy);
+  }
 }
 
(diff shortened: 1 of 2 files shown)


plugins/cache_promote/configs.cc
```diff
@@ -99,6 +99,10 @@ PromotionConfig::factory(int argc, char *argv[])
     }
   }
 
+  if (!_policy) {
```
traeak, line 102: compare to nullptr instead of just '!'

bneradt (dismissed):

plugins/cache_promote/configs.cc
bneradt: Yes, we generally do compare pointers to `!= nullptr`

elsloo: This was the second of two `nullptr` checks written this way. I just pushed a commit that changes this occurrence and the inverse check in the destructor at the top of this file. Please take a look again.

bneradt (approved):
Looks good to me.

bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8505, opened 2021-11-09
url_sig: fix bad cast to pointer of different type

traeak:
This affects the TSBase64Decode call which can result in a valid url_sig being marked as bad.


bneradt (approved):

ezelkow1:

zwoop:
Cherry-picked to v9.2.x

PR #8506, opened 2021-11-10
Add proxy.config.http2.stream_error_sampling_threshold

masaori335:
We faced a situation to increase the `proxy.config.http2.stream_error_rate_threshold` (default `0.1`) and noticed that increasing the rate reduces the sampling number in the `Http2ConnectionState::get_stream_error_rate()`.
- e.g. Imagine an error happens on the first stream, with 0.1 we would kill the conn after another 9 successful streams but at 0.3 we would kill the connection after just another 3 streams I think ( from Leif )

I'd like to introduce a config to set the minimum threshold of the sampling number.

----

Co-authored-by: Leif Hedstrom <zwoop@apache.org>

### DIFF
--- a/doc/admin-guide/files/records.config.en.rst
+++ b/doc/admin-guide/files/records.config.en.rst
@@ -4157,6 +4157,11 @@ HTTP/2 Configuration
    |TS| gracefully closes connections that have stream error rates above this
    setting by sending GOAWAY frames.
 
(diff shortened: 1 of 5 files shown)


masaori335:
[approve ci]

bneradt:
@bneradt will review this.

bneradt (approved):
Makes sense.

zwoop:
Cherry-picked to v9.2.x

PR #8512, opened 2021-11-11
AuTest: address various permissions issues

randall:
By default, the Traffic Server process runs as user nobody. This often isn't a
problem with the AuTest runs, but in certain circumstances this process will
not be able to either read or write from certain locations. For instance, by
default, shells in docker containers run as root and the home directory is
/root with permissions set to 550. There's nothing fundamental about AuTest
that prevents this, but since the ATS process is unprivileged it will not be
able to read or write to anything in /root.  There can be other manifestations
of permissions limitations like this.

The AuTest Traffic Server extension already sets up various directories in the
sandbox chowned to nobody for the ATS process to access. This tweaks a few of
the tests that assumed ATS access outside of these directories to use these
designated ones. This allows the test to run in more environments.

(cherry picked from commit 36440ef175d08f33887551f7dc721c68c6d97147)

Conflicts:
    tests/gold_tests/tls/tls_client_verify3.test.py

### DIFF
--- a/plugins/esi/combo_handler.cc
+++ b/plugins/esi/combo_handler.cc
@@ -1121,7 +1121,7 @@ ContentTypeHandler::loadAllowList(std::string const &file_spec)
   bool extra_junk_on_line{false};
   int line_num = 0;
 
-  fs.open(file_spec);
+  fs.open(file_spec, std::ios_base::in);
   if (fs.good()) {
     for (;;) {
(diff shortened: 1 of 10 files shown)


randall:
Backport of #7668 to 9.1.x

bneradt (approved):
Looks good. Thanks for backporting this to 9.1.x.

PR #8526, opened 2021-11-17
Add a knob to enable kTLS 

maskit:
This adds `proxy.config.ssl.ktls.enabled` to enable kTLS. With the setting enabled, ATS passes `SSL_OP_ENABLE_KTLS` to OpenSSL.

### DIFF
--- a/iocore/net/SSLConfig.cc
+++ b/iocore/net/SSLConfig.cc
@@ -435,6 +435,11 @@ SSLConfigParams::initialize()
   }
 
   REC_ReadConfigInt32(ssl_ktls_enabled, "proxy.config.ssl.ktls.enabled");
+#ifndef SSL_OP_ENABLE_KTLS
+  if (ssl_ktls_enabled) {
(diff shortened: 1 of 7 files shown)


iocore/net/SSLConfig.cc
```diff
@@ -433,6 +434,8 @@ SSLConfigParams::initialize()
     TLSKeyLogger::enable_keylogging(keylog_file);
   }
 
+  REC_ReadConfigInt32(ssl_ktls_enabled, "proxy.config.ssl.ktls.enabled");
```
ywkaras, line 437: I suggest adding:
```
#ifdef SSL_OP_ENABLE_KTLS
  if (ssl_ktls_enabled) {
    Error("kTLS configured but not supported by OpenSSL library");
  }
#endif
```

maskit: Good suggestion. I added the code.

ywkaras (approved):

PR #8515, opened 2021-11-12
Prevent bad ssl_multicert.config load from being swapped in

randall:
retStatus contains the load status of the config file, but wasn't used
to check if it successful before being swapped in

This is a follow up to #8256

PR #8532, opened 2021-11-19
Adds autest to test error cases loading ssl_multicert.config

randall:
With this test in place, we might've caught #8515 and #8256 earlier

### DIFF
--- a/tests/gold_tests/tls/ssl_multicert_loader.test.py
+++ b/tests/gold_tests/tls/ssl_multicert_loader.test.py
@@ -1,0 +1,113 @@
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+Test.Summary = '''
+Test reloading ssl_multicert.config with errors and keeping around the old ssl config structure
+'''
+
+sni_domain = 'example.com'
+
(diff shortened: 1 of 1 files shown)


bneradt:
[approve ci]

tests/gold_tests/tls/ssl_multicert_loader.test.py
```diff
@@ -0,0 +1,118 @@
...
+ts.Disk.records_config.update({'proxy.config.ssl.server.cert.path': f'{ts.Variables.SSLDir}',
+                               'proxy.config.ssl.server.private_key.path': f'{ts.Variables.SSLDir}',
+                               })
+
+tr = Test.AddTestRun("ensure we can connect for SNI $sni_domain")
+tr.Processes.Default.StartBefore(Test.Processes.ts)
+tr.Processes.Default.StartBefore(server)
+tr.StillRunningAfter = ts
+tr.StillRunningAfter = server
+tr.Processes.Default.Command = f"curl -q -s -v -k --resolve '{sni_domain}:{ts.Variables.ssl_port}:127.0.0.1' https://{sni_domain}:{ts.Variables.ssl_port}"
```
mlibbey, line 54: What would you think about adding a comment here about why -k is needed for our future selves. I presume its something about a self signed cert / chain doesn't validate?

randall: Almost all the tests that use curl to do https use `-k` or `--insecure` and the ones that don't, specify `--cacert`.

It's all copy paste!

tests/gold_tests/tls/ssl_multicert_loader.test.py
```diff
@@ -0,0 +1,118 @@
...
+    'ssl_cert_name=server_does_not_exist.pem ssl_key_name=server_does_not_exist.key',
+    'dest_ip=* ssl_cert_name=server.pem_doesnotexist ssl_key_name=server.key',
+])
+tr2.StillRunningAfter = ts
+tr2.StillRunningAfter = server
+tr2.Processes.Default.Command = 'echo Updated configs'
+tr2.Processes.Default.Env = ts.Env
+tr2.Processes.Default.ReturnCode = 0
+
+ts.Disk.diags_log.Content = Testers.ContainsExpression('ERROR: ', 'ERROR')
```
bneradt, line 75: Maybe this should be added as a TestRun (tr2reload) file like you did for the multicert_config above? I think that way it will check during the execution of the reload and not globally after all the tests complete.

bneradt (changes requested):
I have a few minor comments, but overall this looks good. Thanks for adding coverage for this!

bneradt (approved):
Looks good!

zwoop:
Cherry-picked to v9.2.x

PR #8534, opened 2021-11-21
TS_USE_MMAP: mmap+memcopy instead pread+pwrite

cukiernik:
apt install libaio-dev
./configure --enable-mmap --enable-experimental-linux-native-aio
./configure --enable-mmap

### DIFF
--- a/iocore/aio/I_AIO.h
+++ b/iocore/aio/I_AIO.h
@@ -82,7 +82,21 @@ typedef struct io_event ink_io_event_t;
 
 struct AIOCallback;
 struct ink_aiocb {
-  int aio_fildes    = -1;      /* file descriptor or status: AIO_NOT_IN_PROGRESS */
+#if TS_USE_MMAP
+  struct aio_mmap {
+    void *first = MAP_FAILED; /* file descriptor or status: AIO_NOT_IN_PROGRESS */
+    void *last  = nullptr;
+    operator char *() const { return (char *)first; }
+    void
+    operator=(void *p)
+    {
+      first = p;
+      last  = nullptr;
+    }
+  } aio_fildes;
+#else
+  int aio_fildes = -1; /* file descriptor or status: AIO_NOT_IN_PROGRESS */
+#endif
   void *aio_buf     = nullptr; /* buffer location */
   size_t aio_nbytes = 0;       /* length of transfer */
   off_t aio_offset  = 0;       /* file offset */
@@ -96,13 +110,23 @@ struct ink_aiocb {
 
 struct ink_aiocb {
 #if TS_USE_MMAP
-  void *aio_fildes  = MAP_FAILED; /* file descriptor or status: AIO_NOT_IN_PROGRESS */
+  struct aio_mmap {
+    void *first = MAP_FAILED; /* file descriptor or status: AIO_NOT_IN_PROGRESS */
+    void *last  = nullptr;
+    operator char *() const { return (char *)first; }
+    void
+    operator=(void *p)
+    {
+      first = p;
+      last  = nullptr;
+    }
+  } aio_fildes;
 #else
   int aio_fildes = -1; /* file descriptor or status: AIO_NOT_IN_PROGRESS */
 #endif
-  void *aio_buf     = nullptr;    /* buffer location */
-  size_t aio_nbytes = 0;          /* length of transfer */
-  off_t aio_offset  = 0;          /* file offset */
+  void *aio_buf     = nullptr; /* buffer location */
+  size_t aio_nbytes = 0;       /* length of transfer */
+  off_t aio_offset  = 0;       /* file offset */
 
   int aio_lio_opcode = 0; /* listio operation */
   int aio_state      = 0; /* state flag for List I/O */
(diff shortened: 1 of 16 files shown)


bryancall:
@cukiernik Is this supposed to be faster than pread/pwrite?  Do you have any benchmarks.

cukiernik:
The improvement is not significant, single percentages. Speed up is due to reduce a system calls, but at the cost of increased page faults.
The reading from the file is done only on the missing page (instead of a system call each time). Memory writes to a file when the system is low on free pages. So this performance of IO thread's is very close to that of a swap file.
Further acceleration will be achieved by optimizing the agg_copy. Mmap allows aggregation into target memory, which will eliminate memcpy.

cukiernik:
Should I redo "push" to start "**expected checks**" processes?

ezelkow1:
[approve ci]

ywkaras:
@cukiernik Have you run all the Au tests on a build with this configured?

Are you currently at Intel, can you talk about the applications you all are using ATS for?

cukiernik:
@ywkaras : Yes, I ran a tests, but I don't know what's on your mind when you write _all_?

ywkaras:
To run all the Au tests, you cd to the root directory of your clone of the trafficserver repo, build it with your branch, install it, and then do:
```
cd tests
./autests.sh --ats-bin <install-root-directory>/bin
```
It will take 20 - 40 minutes to run.

keesspoelstra:
@cukiernik , A few percent sounds significant already 👍 this could be very useful. You have run it without RAM cache?

I think that for a write into a memory mapped file it has to page the data into memory first (a read), so that adds some overhead. We may be able to get some more performance here by writing with pwrite instead of memcpy. 
This also keeps write behaviour predictable, which is better if ATS crashes (which it never does of course). 
With memory mapped files it is not guaranteed that a valid index will have been written after a crash. A msync could solve this, but the order in which the msync writes the dirty pages is not guaranteed, so a bit more orchestration would be needed for msyncing the index + data. Worst case a lot of data will be written at a very inconvenient moment.
Linux provides a consistent view between pwrites and memory mapped files, for other systems we'd have to check the behaviour.

You might want to take a look at this:
https://github.com/apache/trafficserver/blob/3012ca0f94cf7202e75422757d7498bc7e4f9295/iocore/aio/AIO.cc#L598
I think we're hitting sync I/O here if the pages are not in memory, blocking the thread instead of using the original async nature of the function. I don't know if passing the read to a number of specialized threads would yield better performance, I suspect it will all depend on the workload. 
Mincore and msync can check whether pages are in memory, so that could be used to set up a fast (serve from memory) path, but the drawback would be the overhead of these functions.

keesspoelstra:
Just checked whether ink_aio_read and ink_aio_write are used by anything else. In the C API we have TSAIOWrite + TSAIORead. This feature will break plugins which expect to be able to write past the end of the file with TSAIOWrite.

As the mmap stuff is experimental it I guess it is not a problem now?

cukiernik:
Hello, after my long holidays.
@ywkaras: I didn't run the _au test_, because _pip_ and _PyPi_ seem to endanger my system security.
@keesspoelstra: I was also wondering whether to force _msync_ after each modyfication, using _MS_ASYNC_, a add dedicated thread to _MS_SYNC_. but I couldn't decide on any of the solutions, I will propose a solution without it. **It is not needet to a correcty working**. As I wrote above, I check it with a gigantic amount of RAM, and after modifying the whole cache, _msync MS_SYNC_ takes more than a few minutes.
In my task, I don't need to store content cache after _shutdown_. In this case, _MAP_ANONYMOUS_ works faster. It is also suggested @SolidWallOfCode, but with _MAP_SHARED_ it's easier to trace and debug for me.

bryancall:
@moonchen is going to review this

moonchen:
I'm still reading through all the changes.  One thing I've noticed is that there doesn't seem to be bounds check for the memory mapped I/O.  This may introduce crashes or even security issues.  Interested in hearing how others feel about this.

iocore/aio/I_AIO.h
```diff
@@ -33,6 +33,7 @@
 #include "tscore/ink_platform.h"
 #include "I_EventSystem.h"
 #include "records/I_RecProcess.h"
+#include <sys/mman.h>
```
moonchen, line 36: This should be behind #if TS_USE_MMAP

moonchen (changes requested):
Hi, sorry about taking a long time to review this.  I have some recommendations and two questions.

1. I understand that the cache code is the only caller and it should be doing bounds checking on its disk accesses, but I think AIO should still be verifying that accesses are within the bounds of the memory mapped device because:
   1. This is currently done by the other AIO modes, Linux native AIO, and threaded AIO using pread/pwrite.  Having the modes differ in whether they check bounds will be confusing.
   1. An error in the caller can cause arbitrary memory accesses, which is a serious security issue.
1. There are a lot of changes in the cache code because the type of file descriptor changes.  I suggest using a typedef to abstract the FD type, and using a macro for its printf string, so that you don't need to have #if TS_USE_MMAP in the cache code.  We should minimize the number of places where the cache code has to be aware of the AIO mode to maintain abstraction between the cache code and AIO code.
2. Why is the io.mutex needed when using mmap?
3. Since we already have AIO_MODE_NATIVE and AIO_MODE_THREAD, does it not make more sense to add an AIO_MODE_MMAP?

bryancall:
[approve ci autest]

ywkaras:
@cukiernik do we still need this?  If so, I can ask about it in the weekly meeting.

PR #8546, opened 2021-12-07
backout down parent retry limiting in parent selection and nexthop strategies

jrushford:
backs out the parent retry limiting in parent selection and next hop strategies due to issues found with this in production use.


bneradt:
@ywkaras volunteered to take a look at this.

ywkaras (approved):

PR #8548, opened 2021-12-08
Traffic Dump: Use the correct transaction user index

bneradt:
Traffic Dump has both session and transaction user data. In one of the
calls to TSUserArgGet, it accidentally used the session index to access
the transaction data. This could result in the corruption of another
plugin's session data. This patch fixes this so the correct index is
used.

---

In parallel I'm working on a core change that will detect incorrect indexing rather than allow the accidental corrupting of another plugin's data. See: #8550.

### DIFF
--- a/plugins/experimental/traffic_dump/transaction_data.cc
+++ b/plugins/experimental/traffic_dump/transaction_data.cc
@@ -479,7 +479,7 @@ TransactionData::global_transaction_handler(TSCont contp, TSEvent event, void *e
   }
 
   case TS_EVENT_HTTP_TXN_CLOSE: {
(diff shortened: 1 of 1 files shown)


brbzull0 (approved):
Good catch!

randall (approved):

ywkaras:
Does this need to go into 9.1.x and any 8.x maintenance releases?

bneradt:
> Does this need to go into 9.1.x and any 8.x maintenance releases?

I'll add it to 9.1.x. The 8.1.x traffic_dump plugin is pretty minimal and does not have this bug.

Thanks.

zwoop:
Cherry-picked to v9.2.x

PR #8550, opened 2021-12-08
TSUserArg: add value type checking

bneradt:
A plugin may have multiple user argument types. For instance, it may
have both TS_USER_ARGS_SSN and TS_USER_ARGS_TXN plugin data. Due to a
coding error, it is possible that it may accidentally use the wrong
index type while retrieving one or the other user data types. For
instance, if it uses its transaction index to retrieve its session data,
this may result in them getting the transaction data for another plugin.
In these situations, the plugin will either read what looks like garbage
data to it (because it is casting incorrect memory to its data type), or
it may write to that data and thus corrupt the other plugin's data, or
both.

This change updates the core to segment off the plugin user data ids by
category. This allows the core to verify that the plugin passes the
correct id type per the data type (session, transaction, etc.) and
fails an assertion if there is a mismatch.

---

Note: autests for this PR will fail until a couple plugins with indexing issues uncovered by this patch are merged in:

* #8548
* #8551

### DIFF
--- a/include/tscore/PluginUserArgs.h
+++ b/include/tscore/PluginUserArgs.h
@@ -35,6 +35,41 @@ static constexpr std::array<size_t, TS_USER_ARGS_COUNT> MAX_USER_ARGS = {{
   128 /* max number of user arguments for GLB */
 }};
 
+/** Stagger each user argument value so we can detect mismatched
+ * indices.
+ *
+ * For example, say a plugin associates data with both sessions and
+ * transactions and that its session index is 2 and its transaction index is 4.
+ * In this case, we'll hand back to the plugin 2002 for its session index and
+ * 1004 for its transaction index. If it then accidentally uses its session
+ * index to reference its transaction index, it will pass back 2002 instead of
+ * 1004, which we will identify as belonging to the wrong user argument type
+ * because it is in the 2000 session block rather than the expected 1000
+ * transaction block.
+ *
+ * Note that these higher value 1000 block indices are only used when
+ * interfacing with the plugin. Internally the lower valued index is used.
+ * That is, for a transaction, expect internally a value of 3 instead of 1003.
+ */
+static constexpr size_t
+get_user_arg_offset(TSUserArgType type)
+{
+  // TS_USER_ARGS_TXN indices begin at 1000, TS_USER_ARGS_SSN begin at 2000,
+  // etc.
+  return (static_cast<size_t>(type) + 1) * 1000;
+}
+
+/** Verify that the user passed in an index whose value corresponds with the
+ * type. See the comment above the declaration of get_user_arg_offset for the
+ * intention behind this.
+ */
+static constexpr inline bool
+SanityCheckUserIndex(TSUserArgType type, int idx)
+{
+  int const block_start = get_user_arg_offset(type);
+  return idx >= block_start && idx < block_start + 1000;
+}
+
 /**
   This is a mixin class (sort of), implementing the appropriate APIs and data storage for
   a particular user arg table. Used by VConn / Ssn / Txn user arg data.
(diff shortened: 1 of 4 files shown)


include/tscore/PluginUserArgs.h
```diff
@@ -35,6 +35,40 @@ static constexpr std::array<size_t, TS_USER_ARGS_COUNT> MAX_USER_ARGS = {{
...
+  4000  /* max number of user arguments for GLB */
+}};
+
+/** Verify that the user passed in an index whose value corresponds with the
+ * type. See the comment above the declaration of USER_ARGS_OFFSETS for the
+ * intention behind this.
+ */
+inline bool
+SanityCheckUserIndex(TSUserArgType type, int idx)
+{
```
ywkaras, line 67: I suggest adding:
```
if (static_cast<std::size_t>(type) >= MAX_USER_ARGS.size()) {
  return false;
}

File: include/tscore/PluginUserArgs.h [hunk 2]
```diff
@@ -35,6 +35,40 @@ static constexpr std::array<size_t, TS_USER_ARGS_COUNT> MAX_USER_ARGS = {{
...
+ * Note that these higher value 1000 block indices are only used when
+ * interfacing with the plugin. Internally the lower valued index is used.
+ * That is, for a transaction, expect internally a value of 3 instead of 1003.
+ */
+static constexpr std::array<size_t, TS_USER_ARGS_COUNT> USER_ARGS_OFFSETS = {{
+  1000, /* max number of user arguments for TXN */
+  2000, /* max number of user arguments for SSN */
+  3000, /* max number of user arguments for VCONN */
+  4000  /* max number of user arguments for GLB */
+}};
```
ywkaras, line 59: You could instead use:
```
static constexpr size_t USER_ARG_OFFSETS(TSUserArgType type)
{
  return (static_cast<size_t>(type) + 1) * 1000;
}
```

ywkaras:
Nitpick:  seems like static constexpr items should be private members of PluginUserArgsMixin.  But that's perhaps beyond the scope of this PR.

include/tscore/PluginUserArgs.h
bneradt: I like this idea.

bneradt:
> Nitpick: seems like static constexpr items should be private members of PluginUserArgsMixin. But that's perhaps beyond the scope of this PR.

I agree that it's good to keep stuff private if they can be, but these functions are both needed externally in the InkAPI.cc code. Here, for example:
https://github.com/apache/trafficserver/pull/8550/files#diff-e188d413739d07f1ad183ae5912ca435557809252154bdd5f2fcf625f95b9600R6331

ywkaras (approved):

bneradt:
[approve ci autest]

ywkaras:
Unfortunately I think this is probably detecting errors in core plugins, that you'll need to fix in this PR to get Au test regression to pass.  I suggest trying the failing cases in your own test environment, see what's going on.

duke8253:
I like this

bneradt:
> Unfortunately I think this is probably detecting errors in core plugins, that you'll need to fix in this PR to get Au test regression to pass. I suggest trying the failing cases in your own test environment, see what's going on.

Correct, there were issues with core plugins. But I think those are now both addressed and merged in via the PRs mentioned in the description. We'll see whether I missed any with this latest autest run.

ywkaras (approved):

bneradt:
@masaori335 has volunteered to review this.

masaori335:
[approve ci centos]

masaori335 (approved):

bneradt:
[approve ci centos]

bneradt:
[approve ci centos]

bneradt:
[approve ci centos]

bneradt:
[approve ci centos]

bneradt:
[approve ci centos]

zwoop:
I'd like to leave this out of 9.1.x if ok with you.

zwoop:
Cherry-picked to v9.2.x

bneradt:
> I'd like to leave this out of 9.1.x if ok with you.

Fine with me. I'll remove the 9.1.x project.

PR #8551, opened 2021-12-08
combo_handler: Initialize User Arg Index in TSRemapInit

bneradt:
The combo_handler plugin did not register its transaction argument index
in TSRemapInit, which resulted in using an unregistered id. This
generally meant it indexed 0, which could result in it accessing the
wrong plugin's transaction user argument data.


randall (approved):

zwoop:
Cherry-picked to v9.2.x

ezelkow1:

PR #8570, opened 2022-01-03
Port #6816 from core strategy to plugin

rob05c:
Ports #6816 fix for core strategies to the parent_select plugin.

Fixes a bug with the YAML loading calling opendir but not closedir,
which leaked the file descriptor.


bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

PR #7749, opened 2021-04-27
Fixes Issue #7739 - Next hop strategy with bad 'to' URL causes TS crash.

jrushford:
Fixes #7739 NextHopConsistentHash::findParent().  A crash occurs when searching for an available next hop results in a failure and alternate_ring mode is used.

PR #8571, opened 2022-01-03
Port #7749 from core strategy to plugin

rob05c:
Ports #7749 fix for core strategies to the parent_select plugin.

Fixes a crash when a strategy has a bad 'to' URL, which manifests as a
missing null pointer check.


bneradt (approved):

traeak (approved):
Probably should be using nullptr in compares (ie: (nullptr != pRec) instead of just (pRec)) but it is consistent with the rest of the code.

zwoop:
Cherry-picked to v9.2.x

PR #8573, opened 2022-01-04
Port #7837 from core strategy to plugin

rob05c:
Ports #7837 from core strategies to parent_select plugin.

Adds unavailable_server_retries and markdown to strategies.


ywkaras (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8576, opened 2022-01-05
LGTM: Fix constant comparison

masaori335:
https://lgtm.com/projects/g/apache/trafficserver/?mode=tree&ruleFocus=2154840804

I might be misunderstanding, but many of the alerts seem false positive.

### DIFF
--- a/iocore/net/UnixUDPNet.cc
+++ b/iocore/net/UnixUDPNet.cc
@@ -534,7 +534,7 @@ UDPNetProcessor::recvfrom_re(Continuation *cont, void *token, int fd, struct soc
     cont->handleEvent(NET_EVENT_DATAGRAM_READ_COMPLETE, event);
     completionUtil::destroy(event);
     return ACTION_RESULT_DONE;
-  } else if (actual == 0 || (actual < 0 && actual == -EAGAIN)) {
+  } else if (actual == 0 || actual == -EAGAIN) {
     UDPReadContinuation *c = udpReadContAllocator.alloc();
     c->init_token(event);
     c->init_read(fd, buf, len, fromaddr, fromaddrlen);
@@ -989,7 +989,7 @@ UDPQueue::SendUDPPacket(UDPPacketInternal *p, int32_t /* pktLen ATS_UNUSED */)
   while (true) {
     // stupid Linux problem: sendmsg can return EAGAIN
     n = ::sendmsg(p->conn->getFd(), &msg, 0);
-    if ((n >= 0) || ((n < 0) && (errno != EAGAIN))) {
+    if ((n >= 0) || (errno != EAGAIN)) {
       // send succeeded or some random error happened.
       if (n < 0) {
         Debug("udp-send", "Error: %s (%d)", strerror(errno), errno);
--- a/plugins/experimental/memcache/tsmemcache.cc
+++ b/plugins/experimental/memcache/tsmemcache.cc
@@ -498,45 +498,16 @@ int
 MC::binary_get_event(int event, void *data)
 {
   ink_assert(!"EVENT_ITEM_GOT is incorrect here");
-  if (event != TSMEMCACHE_EVENT_GOT_ITEM) {
-    CHECK_READ_AVAIL(binary_header.request.keylen, &MC::binary_get);
-    key         = binary_get_key(this);
-    header.nkey = binary_header.request.keylen;
-    return get_item();
-  } else if (event == CACHE_EVENT_OPEN_READ_FAILED) {
-    if (f.noreply) {
-      return read_from_client();
-    }
-    if (binary_header.request.opcode == PROTOCOL_BINARY_CMD_GETK) {
-      add_binary_header(PROTOCOL_BINARY_RESPONSE_KEY_ENOENT, 0, header.nkey, header.nkey);
-      wbuf->write(key, header.nkey);
-      return write_then_read_from_client();
-    } else {
-      return write_binary_error(PROTOCOL_BINARY_RESPONSE_KEY_ENOENT, 0);
-    }
-  } else if (event == CACHE_EVENT_OPEN_READ) {
-    protocol_binary_response_get *rsp = &res.get;
-    uint16_t keylen                   = 0;
-    uint32_t bodylen                  = sizeof(rsp->message.body) + (rcache_header->nbytes - 2);
-    bool getk =
-      (binary_header.request.opcode == PROTOCOL_BINARY_CMD_GETK || binary_header.request.opcode == PROTOCOL_BINARY_CMD_GETKQ);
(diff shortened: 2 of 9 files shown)


iocore/net/UnixUDPNet.cc
```diff
@@ -534,7 +534,7 @@ UDPNetProcessor::recvfrom_re(Continuation *cont, void *token, int fd, struct soc
     cont->handleEvent(NET_EVENT_DATAGRAM_READ_COMPLETE, event);
     completionUtil::destroy(event);
     return ACTION_RESULT_DONE;
-  } else if (actual == 0 || (actual < 0 && actual == -EAGAIN)) {
+  } else if (actual == 0 || (actual < 0 && actual == -EAGAIN)) { // lgtm[cpp/constant-comparison]
```
ywkaras, line 537: Seems like this should simply be:
```
  } else if (actual == 0 || actual == -EAGAIN) {
```

iocore/net/UnixUDPNet.cc
```diff
@@ -989,7 +989,7 @@ UDPQueue::SendUDPPacket(UDPPacketInternal *p, int32_t /* pktLen ATS_UNUSED */)
   while (true) {
     // stupid Linux problem: sendmsg can return EAGAIN
     n = ::sendmsg(p->conn->getFd(), &msg, 0);
-    if ((n >= 0) || ((n < 0) && (errno != EAGAIN))) {
+    if ((n >= 0) || ((n < 0) && (errno != EAGAIN))) { // lgtm[cpp/constant-comparison]
```
ywkaras, line 992:
```
A  B  A || B  A || (!A && B)
-  -  ------  --------------
F  F  F       F
F  T  T       T
T  F  T       T
T  T  T       T
```
So the above is equivalent to:
```
    if ((n >= 0) || ((errno != EAGAIN)) {
```

plugins/experimental/memcache/tsmemcache.cc
```diff
@@ -503,7 +503,7 @@ MC::binary_get_event(int event, void *data)
     key         = binary_get_key(this);
     header.nkey = binary_header.request.keylen;
     return get_item();
-  } else if (event == CACHE_EVENT_OPEN_READ_FAILED) {
+  } else if (event == CACHE_EVENT_OPEN_READ_FAILED) { // lgtm[cpp/constant-comparison]
```
ywkaras, line 506: Assuming `CACHE_EVENT_OPEN_READ_FAILED != TSMEMCACHE_EVENT_GOT_ITEM`, this code is unreachable.

plugins/experimental/memcache/tsmemcache.cc
```diff
@@ -514,7 +514,7 @@ MC::binary_get_event(int event, void *data)
     } else {
       return write_binary_error(PROTOCOL_BINARY_RESPONSE_KEY_ENOENT, 0);
     }
-  } else if (event == CACHE_EVENT_OPEN_READ) {
+  } else if (event == CACHE_EVENT_OPEN_READ) { // lgtm[cpp/constant-comparison]
```
ywkaras, line 517: Assuming `CACHE_EVENT_OPEN_READ != TSMEMCACHE_EVENT_GOT_ITEM`, this code is unreachable.

src/traffic_layout/engine.cc
```diff
@@ -227,7 +227,7 @@ LayoutEngine::create_runroot()
     empty_check_directory = ts_runroot;
     if (ftw(ts_runroot.c_str(), check_directory_empty, OPEN_MAX_FILE) != 0) {
       // if the directory is not empty, check for valid Y/N
-      for (int i = 0; i < 3; i++) {
+      for (int i = 0; i < 3; i++) { // lgtm[cpp/constant-comparison]
```
ywkaras, line 230: This can just be:
```
     for (int i = 0; true; i++) {
```
because the if statement at line 240 guarantees that i will never be 3.

src/traffic_logstats/logstats.cc
```diff
@@ -1108,7 +1108,7 @@ update_codes(OriginStats *stat, int code, int size)
     update_counter(stat->codes.c_4xx, size);
   } else if (code >= 300) {
     update_counter(stat->codes.c_3xx, size);
-  } else if (code >= 200) {
+  } else if (code >= 200) { // lgtm[cpp/constant-comparison]
```
ywkaras, line 1111: This could be:
```
} else /* 200 <= code < 300 */ {
```

src/traffic_manager/traffic_manager.cc
```diff
@@ -179,7 +179,7 @@ is_server_idle_from_new_connection()
 
   Debug("lm", "%" PRId64 " active clients, threshold is %" PRId64, active, threshold);
 
-  return active <= threshold;
+  return true;
 }
```
ywkaras, line 183: I suggest:
```
static bool
is_server_idle_from_new_connection()
{
  RecInt threshold = 0;

  Debug("lm", "%" PRId64 " active clients, threshold is %" PRId64, active, threshold);

#if 0
  // TODO implement with the right metric
  RecInt active    = 0;
  return active <= threshold;
#else
  return true;
#endif
}
```
This is clearer, and will avoid any potential unused local variable warnings.

iocore/net/UnixUDPNet.cc
masaori335: Ah, right. Now I got what the alert says.

plugins/experimental/memcache/tsmemcache.cc
masaori335: That's right! These `else if` blocks are dead code.

src/traffic_layout/engine.cc
masaori335: Right. This is just checking 'Y/N'...

masaori335:
Addressed @ywkaras pointed out. Obviously, I was still in vacation mode.

plugins/experimental/memcache/tsmemcache.cc
```diff
@@ -498,45 +498,16 @@ int
 MC::binary_get_event(int event, void *data)
 {
   ink_assert(!"EVENT_ITEM_GOT is incorrect here");
-  if (event != TSMEMCACHE_EVENT_GOT_ITEM) {
-    CHECK_READ_AVAIL(binary_header.request.keylen, &MC::binary_get);
-    key         = binary_get_key(this);
-    header.nkey = binary_header.request.keylen;
-    return get_item();
-  } else if (event == CACHE_EVENT_OPEN_READ_FAILED) {
```
masaori335, line 506: As @ywkaras pointed out, these `else if` blocks are unreachable.

bneradt (approved):

PR #8578, opened 2022-01-05
Add SSLSessionDup for older OpenSSL and BoringSSL

maskit:
PR #8498 uses `SSL_SESSION_dup` which is available since OpenSSL 1.1.1. I only confirmed BoringSSL build is broken, but it probably doesn't build with older OpenSSL as well.

This PR introduces a wrapper function for `SSL_SESSION_dup`, and it mimics the behavior of `SSL_SESSION_dup` if the function is unavailable. I didn't test the implementation very seriously, but it should work if I understand the code before #8498.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1311,6 +1311,7 @@ AC_CHECK_FUNCS([ \
   X509_get0_signature \
   ERR_get_error_all \
   SHA1 \
+  SSL_SESSION_dup \
 ])
 
 AC_CHECK_FUNC([ASN1_STRING_get0_data], [],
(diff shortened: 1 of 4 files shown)


duke8253 (approved):
Looks good to me.

bneradt (approved):

traeak (approved):
Looks good.  I had updated CI to ensure we were using openssl 1.1.1l due to the security issues with previous releases but had forgotten about boringssl.

zwoop:
Cherry-picked to v9.2.x

PR #8579, opened 2022-01-05
Fixes compacting spaces in S3 auth plugin

### LINKED ISSUE #7888: S3 auth plugin configuration parsing does not account for inner spaces
The S3 plugin currently only strips leading and trailing whitespace from the entire configuration line before determining which field is currently being parsed. Once the line is stripped, a string comparison is performed to identify keys, which includes the equals symbol, and if that matches, we advance the pointer based on the key's length and set the value for the key with the value immediately following the equals symbol.

Leading and trailing whitespace around the equals symbol is not handled, for example, if someone provides `secret_key = foo` or `access_key= bar`.

Because we are already stripping the entire string, it would be desirable if the value was checked for leading whitespace before its setter is called. Leading whitespace could be stripped, or logged as a warning if found, but as written, any value set could contain leading whitespace, which may lead to unexpected and difficult to diagnose failures, in particular around the secret and access keys, and session token.

Similarly, a space after the key name but before the equals will cause all of the string matches to fail, and the `else` case is not implemented, and instead contains: `// ToDo: warnings?`. Ideally a minimal fix for this issue would implement a warning if the key in the line being processed did not match any of the string comparisons, or a more robust solution could be a mechanism to strip inner spaces on either side of the equals symbol.

Area in question: https://github.com/apache/trafficserver/blob/master/plugins/s3_auth/s3_auth.cc#L557

hejcz:
This fixes #7888

Hi, 

Here's a change that I believe fixes the https://github.com/apache/trafficserver/issues/7888. Could you review it?

### DIFF
--- a/plugins/s3_auth/aws_auth_v4.cc
+++ b/plugins/s3_auth/aws_auth_v4.cc
@@ -168,18 +168,19 @@ canonicalEncode(const String &in, bool isObjectName)
 }
 
 /**
- * @brief trim the white-space character from the beginning and the end of the string ("in-place", just moving pointers around)
+ * @brief Trim white spaces from beginning and end. Squeeze consecutive spaces from middle.
+ *
+ * @see AWS spec: https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html
  *
  * @param in ptr to an input string
  * @param inLen input character count
- * @param newLen trimmed string character count.
  * @return pointer to the trimmed string.
  */
-const char *
-trimWhiteSpaces(const char *in, size_t inLen, size_t &newLen)
+std::string
+trimWhiteSpacesAndSqueezeInnerSpaces(const char *in, size_t inLen)
 {
   if (nullptr == in || inLen == 0) {
-    return in;
+    return std::string(in, inLen);
   }
 
   const char *first = in;
@@ -192,8 +193,22 @@ trimWhiteSpaces(const char *in, size_t inLen, size_t &newLen)
     last--;
   }
 
-  newLen = last - first + 1;
-  return first;
+  std::stringstream result;
+  int consecutiveSpaces = 0;
+  while (first <= last) {
+    if (*first == ' ') {
+      consecutiveSpaces++;
+    } else {
+      if (consecutiveSpaces > 0) {
+        result << ' ';
+      }
+      consecutiveSpaces = 0;
+      result << *first;
+    }
+    first++;
+  }
+
+  return result.str();
 }
 
 /**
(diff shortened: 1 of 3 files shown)


bneradt:
[approve ci]

duke8253:
[approve ci ubuntu]

plugins/s3_auth/aws_auth_v4.cc
```diff
@@ -192,8 +192,22 @@ trimWhiteSpaces(const char *in, size_t inLen, size_t &newLen)
     last--;
   }
 
-  newLen = last - first + 1;
-  return first;
+  std::stringstream result;
+  int consecutiveSpaces = 0;
+  while (first <= last) {
+    if (*first == ' ') {
```
duke8253, line 198: Not sure this will work. Correct me if I'm wrong, but to me it looks like you're replacing consecutive spaces with one space, and putting it back in the string. So `access_key=    foo` will become `access_key= foo`. But from what I understand, #7888 wants all space removed. Also, I believe `isspace` checks for all white-space characters, not just space.

Btw, if we're removing all white-spaces in the string, why have separate loops to remove the leading and trailing spaces, this could all be done in one go.

duke8253 (changes requested):

plugins/s3_auth/aws_auth_v4.cc
hejcz: You're right. I didn't read that issue carefully. I opened a new issue describing what I'm actually trying to fix https://github.com/apache/trafficserver/issues/8583

hejcz: Regarding removing leading and trailing whitespaces separately it was done this way before. What's the idea to remove leading and trailing spaces in one loop? Going from begging to the end how do you know that trailing spaces started? In the issue I linked how to create canonical headers and from what I understand you need to remove leading and trailing whitespaces but you need to compact ordinary spaces only. If we were removing only spaces or all whitespaces it could be done in one iteration.

duke8253: Ok, nvm the "in one go" part, I thought the value can't have spaces in it, but looks like the value itself may contain spaces in between. 

But I would still suggest you fix both #7888 and #8583 at the same time, you will only need to add a check for `=`, should be pretty easy to do.

duke8253: Also, you need to fix your code, the Ubuntu build is failing at `test_aws_auth_v4.cc:239`.

hejcz: @duke8253 I managed to fix tests and add some new. I can fix #7888 in another pr as it affects different code (and file) and doesn't seem to be related to the change I proposed in this pr.

duke8253: Looks mostly good, could you change your condition check on line 199 to use `isspace()`? Keep it consistent as the trimming part of the code, and it checks for all white space characters, not just spaces.

duke8253: Hmm, ok, missed that part.

duke8253 (approved):

hejcz:
@duke8253 so what's gonna happen next? Do we wait for another review or is the change ready to merge?

PR #8582, opened 2022-01-06
Add TLSEarlyDataSupport

maskit:
Like other `TLS*Support`, this defines an interface for TLS Early Data.

Since we already use the same interface for the both Early Data and Non-Early Data in terms of data I/O, `TLSEarlyDataSupport` only provides an auxiliary function. It doesn't really clean up code, but removes a dependency for `SSLNetVC` from proxy sessions, and clarify how to access necessary information.

For QUIC, `TLSEarlyDataSupport` will also be implemented by `QUICStream` that is in between `QUICNetVC` and HTTP/3 proxy session, in addition to `QUICNetVC`. Then we will be able to use the same interface from proxy layer in spite of difference between non-multiplexed connections (regular TLS) and multiplexed connections (QUIC).

### DIFF
--- a/iocore/net/Makefile.am
+++ b/iocore/net/Makefile.am
@@ -191,6 +191,7 @@ libinknet_a_SOURCES = \
 	SSLUtils.cc \
 	OCSPStapling.cc \
 	TLSBasicSupport.cc \
+	TLSEarlyDataSupport.cc \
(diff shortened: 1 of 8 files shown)


bryancall:
@duke8253 Can you please take a look at this one

duke8253 (approved):
Looks good to me.

PR #8592, opened 2022-01-10
Fix outbound transparent HTTP/2 and DNS issues

### LINKED ISSUE #8587: HTTP/2 does not work with outbound transparent
Once I fixed issue #7525, I could pass traffic for HTTP and HTTPS HTTP/1.1 with the certifier plugin.  But for HTTP/2 traffic I would get a 500 response.  Looking at the debug messages, it appeared that the HttpSM was very insistent about doing a DNS lookup on the domain.

Eventually tracked it down to the fact that the accept object transparent flags were not being propagated properly to the Http2Streams.  I am putting up a PR that fixes the issue, but perhaps in not the most beautiful way.

This is observed working on the 9.1.1 codebase.

shinrich:
Fixes issues with propagating the outbound transparent flag.  Would like @SolidWallOfCode to review. This fix removes the ability to override outbound transparent on a per transaction basis, but I fail to see how that would have ever been useful.

Also adds a check to avoid exercising the DNS lookup if outbound transparent is enabled.

This closes #8587

### DIFF
--- a/proxy/ProxySession.cc
+++ b/proxy/ProxySession.cc
@@ -312,3 +312,9 @@ ProxySession::support_sni() const
 {
   return _vc ? _vc->support_sni() : false;
 }
+
+bool
+ProxySession::is_outbound_transparent() const
+{
+  return accept_options->f_outbound_transparent;
+}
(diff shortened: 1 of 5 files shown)


shinrich:
Will create a new PR based on @SolidWallOfCode's comments.

PR #8594, opened 2022-01-10
Propagate accept options to HTTP/2

### LINKED ISSUE #8587: HTTP/2 does not work with outbound transparent

shinrich:
Propagates the outbound transparent field so HTTP/2 requests pass over the transparent session. Details in issue.

This closes #8587

### DIFF
--- a/proxy/http2/Http2Stream.cc
+++ b/proxy/http2/Http2Stream.cc
@@ -46,11 +46,12 @@ Http2Stream::Http2Stream(ProxySession *session, Http2StreamId sid, ssize_t initi
 
   this->mark_milestone(Http2StreamMilestone::OPEN);
 
-  this->_sm          = nullptr;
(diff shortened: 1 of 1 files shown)


masaori335 (approved):
Looks good for a workaround. As @shinrich commented on #8587, the session-based hook solution seems better.

PR #8595, opened 2022-01-10
Adds two overridable config variables to control parent mark downs.

jrushford:
This PR adds two overridable config variables to control parent mark downs when using
parent selection or NextHop Strategies. These configs may be overridden by the 
header_rewrite plugin on a per remap basis.

- proxy.config.http.parent_proxy.enable_parent_timeout_markdowns is
  used to enable parent markdowns when an inactivity timeout triggers
  on a parent proxy.  The default is disabled or '0'.

- proxy.config.http.parent_proxy.disable_parent_markdowns  is used to
  disable all parent proxy markdowns.  The default is disabled or '0'.

Currently when using either parent selection or next hop strategies, parents are marked down only
when a connection cannot be established, CONNECTION_ERROR, or when there is a DNS resolution
failure of the selected parent.  We have use cases where we need to mark down a parent due to
INACTIVITY_TIMEOUT for a particular remap.  Also, we have use cases where the parent endpoint
is a VIP and we wish not to mark down the VIP endpoint at all.  These two configuration parameters
will allow us to control that by using the header_rewrite plugin on a per re-map basis.


bneradt:
[approve ci]

doc/admin-guide/files/records.config.en.rst
```diff
@@ -1347,6 +1347,25 @@ Parent Proxy Configuration
    ``2`` Mark the host down. This is the default.
    ===== ======================================================================
 
+.. ts:cv:: CONFIG proxy.config.http.parent_proxy.enable_parent_timeout_markdowns INT 0
+   :reloadable:
+   :overridable:
+
+   Enables (``1``) or disables (``0``) parent proxy mark downs due to inactivity
+   timeouts.  The default is disabled (``0``).  This should be used for specific
+   remap entries using ``header_rewrite`` to enable inactivity timeout markdowns rather
```
ywkaras, line 1356: Did you mean conf_remap, not header_rewrite?

doc/admin-guide/files/records.config.en.rst
```diff
@@ -1347,6 +1347,25 @@ Parent Proxy Configuration
...
+
+.. ts:cv:: CONFIG proxy.config.http.parent_proxy.disable_parent_markdowns INT 0
+   :reloadable:
+   :overridable:
+
+   Enables (``1``) or disables (``0``) parent proxy markdowns.  This is useful
+   if a parent entries in a parent.config line are VIP's and one doesn't wish
+   to mark down a VIP which may have several origin or parent proxies behind
+   this load balancer.  This should be used for specific remap entries using
+   ``header_rewrite`` rather than disabling parent mark downs globally.
```
ywkaras, line 1367: Would it be cleaner to just have disable_parent_markdowns:
0 - parent markdowns enabled.
1 - (default) parent markdowns disabled due to timeouts.
2 - all parent markdowns disabled.

doc/admin-guide/files/records.config.en.rst
jrushford: @ywkaras I mean using the header_rewrite plugin to override the default value of the config so as to enable marking down on timeouts or by disabling markdowns altogether for a particular remap.  Should I change header_rewrite to header_rewrite plugin?

doc/admin-guide/files/records.config.en.rst
jrushford: @ywkaras We want to do this using overridable configuration parameters.  One wouldn't override both enabling timeout markdowns and disabling markdowns at the same time.  I can add this to the documentation if it's not clear that this would not work.  I prefer it as it is.  I'll update the documentation though if you think it's necessary.

doc/admin-guide/files/records.config.en.rst
ywkaras: So, I thought we normally used this plugin to change config values per remap rule:  https://docs.trafficserver.apache.org/admin-guide/plugins/conf_remap.en.html

ywkaras: Oh OK, you can do it with both:  https://docs.trafficserver.apache.org/admin-guide/plugins/header_rewrite.en.html#set-config .  Is there reason to think these would only ever be overridden by header_rewrite?  Maybe it's sufficient that they are marked as overridable?

jrushford: @ywkaras I've updated the documentation.  Please take a look.

jrushford: @ywkaras thanks for the link.  I've never used the conf_remap plugin.  We've always used header_rewrite plugin for that purpose.  I'll change the docs again :)

jrushford: @ywkaras the docs now mention both plugins.

doc/admin-guide/files/records.config.en.rst
```diff
@@ -1347,6 +1347,32 @@ Parent Proxy Configuration
...
+.. ts:cv:: CONFIG proxy.config.http.parent_proxy.enable_parent_timeout_markdowns INT 0
+   :reloadable:
+   :overridable:
+
+   Enables (``1``) or disables (``0``) parent proxy mark downs due to inactivity
+   timeouts.  By default parent proxies are not marked down due to inactivity
+   timeouts, the transaction will retry using another parent instead.  The
+   default for this configuration keeps this behavior and is disabled (``0``).
+   This setting is overridable using one of the two plugins ``header_rewrite``
+   or ``conf_remap`` to enable inactivity timeout markdowns and should be done
```
ywkaras, line 1359: Why do you feel like we should mention these particular plugins as the ones overriding the configs?  Any plugin that uses these TS API calls can override them:  https://github.com/apache/trafficserver/blob/4e6dc02f7ef7ae15ebda6c3bdf54e80f1ec55ba5/include/ts/ts.h#L2459 .  The lua plugin could also override them.

proxy/http/HttpTransact.cc
```diff
@@ -214,6 +214,18 @@ markParentDown(HttpTransact::State *s)
   HTTP_INCREMENT_DYN_STAT(http_total_parent_marked_down_count);
   url_mapping *mp = s->url_map.getMapping();
 
+  TxnDebug("http_trans", "sm_id[%" PRId64 "] enable_parent_timeout_markdowns: %d, disable_parent_markdowns: %d",
+           s->state_machine->sm_id, s->txn_conf->enable_parent_timeout_markdowns, s->txn_conf->disable_parent_markdowns);
+
+  if (s->txn_conf->disable_parent_markdowns == 1) {
+    TxnDebug("http_trans", "parent markdowns are disabled for this request");
+    return;
+  }
```
ywkaras, line 223: If disable_parent_markdowns is 1, we return on line 222, so the value of enable_parent_timeout_markdowns is moot.  That is what makes me think it would be clearer to combine the two settings into a single one.

jrushford: Ok, I’ll make the changes

ywkaras (approved):
I gave Yahoo Prod a chance to whine about this but they didn't, so I'll leave it up to John if it's better to keep the two config variables or consolidate into one.

zwoop:
Cherry-picked to v9.2.x

PR #8600, opened 2022-01-12
Prevent calling SSL_set_session in the middle of handshake

maskit:
ATS crashes if origin session cache is used with BoringSSL, because BoringSSL doesn't allow you to call SSL_set_session after starting handshake.

The crash is not 100% reproducible and I was only able to reproduce it on 9.2.x, but the change should make sense on master as well.

https://www.openssl.org/docs/man1.1.1/man3/SSL_set_session.html
> If there is already a session set inside ssl (because it was set with SSL_set_session() before or because the same ssl was already used for a connection), SSL_SESSION_free() will be called for that session. If that old session is still open, it is considered bad and will be removed from the session cache (if used). A session is considered open, if SSL_shutdown(3) was not called for the connection (or at least SSL_set_shutdown(3) was used to set the SSL_SENT_SHUTDOWN state).

https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_set_session
> It is an error to call this function after the handshake has begun.

The change is very small if you hide whitespace (indent) changes.

### DIFF
--- a/iocore/net/P_SSLNetVConnection.h
+++ b/iocore/net/P_SSLNetVConnection.h
@@ -459,6 +459,7 @@ private:
 
   enum SSLHandshakeStatus sslHandshakeStatus = SSL_HANDSHAKE_ONGOING;
   bool sslClientRenegotiationAbort           = false;
+  bool first_ssl_connect                     = true;
   MIOBuffer *handShakeBuffer                 = nullptr;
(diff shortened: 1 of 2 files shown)


bryancall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8590, opened 2022-01-07
    Port #7925 and #8365 from core strategies to parent_select plugin

rob05c:
Ports #7925 and #8365 from core strategies to parent_select plugin.
    
PR #7925 adds go_direct support to peering_ring strategies.
PR #8365 refactors strategy consistent hashing to fix a bug.
    
They both essentially completely rewrite the strategy consistent
hashing, so it was safer and easier to port them both at the same
time.

This PR is based on https://github.com/apache/trafficserver/pull/8580 - recommend merging that first.

PR #8602, opened 2022-01-12
Add parent_select plugin autests

rob05c:
Tests are mostly copied from the core strategies tests, which makes
sense since the plugin strategies should behave the same as core.


randall:
[approve ci debian]

rob05c:
The contents of this PR have been added to https://github.com/apache/trafficserver/pull/8590 as requested.

PR #8606, opened 2022-01-14
Fixed a scoping bug that leads to "sticky" parent weights

elsloo:
* Due to scoping, the last user-provided weight seen in `parent.config` will be applied to all subsequent parents that lack a weight

* Added a constant to specify the default weight

* Resets the weight to default prior to parsing each parent's configuration to ensure defaults are used when no weight exists


jrushford (approved):
Good catch @elsloo, LGTM

zwoop:
Cherry-picked to v9.2.x

PR #8608, opened 2022-01-16
Update to Proxy Verifier version v2.3.0

bneradt:
This updates to the Proxy Verifier v2.3.0 release. This release updates
to the new libswoc::Errata and comes with some logging improvements.
Since the logs change, the test expectations for Proxy Verifier output
is updated with this commit as well.


bneradt:
This is a test-only change that comes with logging improvements for Proxy Verifier. I'm marking this for 9.2.x because I think it will be easier for us to have 9.2.x and 10.x have the same test log expectations rather than differences which might introduce merge conflicts. 

This currently cherry-picks cleanly to 9.2.x. Let me know if there are conflicts at the time of the backport and I can create a separate PR for 9.2.x.

randall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8610, opened 2022-01-17
Make TLS Early Data available with BoringSSL

### LINKED ISSUE #7503: Early Data is not available with BoringSSL
Current code only works with OpenSSL. The code is ifdef-ed so there's no build issue with BoringSSL, but autest (tls_0rtt_server) fails.

maskit:
Added code to support TLS Early Data with BoringSSL.
https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#Early-data

I reused code for OpenSSL as much as possible, and there should be no logic change on code for OpenSSL. I tested the code for BoringSSL with the existing autest, and also ran the test commands (openssl s_client) by hand and checked outputs on the both ATS side and the client side. All the outputs seemed fine to me, though we can't check those on our CI because we don't have BoringSSL build job.

This closes #7503.

### DIFF
--- a/iocore/net/SSLNetVConnection.cc
+++ b/iocore/net/SSLNetVConnection.cc
@@ -2041,6 +2041,8 @@ SSLNetVConnection::_ssl_accept()
         finished_reading_early_data = true;
       }
 #else
+      // If SSL_read_early_data is unavailable, it's probably BoringSSL,
+      // and SSL_in_early_data should be available.
       ret = SSL_accept(ssl);
       if (ret <= 0) {
         had_error_on_reading_early_data = true;
@@ -2191,6 +2193,10 @@ SSLNetVConnection::_ssl_write_buffer(const void *buf, int64_t nbytes, int64_t &n
   ERR_clear_error();
 
   int ret;
+  // If SSL_write_early_data is available, it's probably OpenSSL,
+  // and SSL_is_init_finished should be available.
+  // If SSL_write_early_data is unavailable, its' probably BoringSSL,
+  // and we can use SSL_write to send early data.
 #if TS_HAS_TLS_EARLY_DATA
   if (SSL_version(ssl) >= TLS1_3_VERSION) {
 #ifdef HAVE_SSL_WRITE_EARLY_DATA
@@ -2282,6 +2288,8 @@ SSLNetVConnection::_ssl_read_buffer(void *buf, int64_t nbytes, int64_t &nread)
         finished_reading_early_data = true;
       }
 #else
+      // If SSL_read_early_data is unavailable, it's probably OpenSSL,
+      // and SSL_in_early_data should be available.
       if (SSL_in_early_data(ssl)) {
         ret = SSL_read(ssl, buf, nbytes);
         finished_reading_early_data = !SSL_in_early_data(ssl);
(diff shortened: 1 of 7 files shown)


iocore/net/SSLNetVConnection.cc
```diff
@@ -2014,15 +2021,61 @@ SSLNetVConnection::_ssl_accept()
...
+#if HAVE_SSL_READ_EARLY_DATA
+      ret = SSL_read_early_data(ssl, block->buf(), index_to_buffer_size(BUFFER_SIZE_INDEX_16K), &nread);
       if (ret == SSL_READ_EARLY_DATA_ERROR) {
-        Debug("ssl_early_data", "SSL_READ_EARLY_DATA_ERROR");
+        had_error_on_reading_early_data = true;
+      } else if (ret == SSL_READ_EARLY_DATA_FINISH) {
+        finished_reading_early_data = true;
+      }
+#else
+      ret = SSL_accept(ssl);
```
bneradt, line 2044: Let's comment the `#else` block indicating that it is the BoringSSL implementation. (You've done that elsewhere in this patch, I think it would be good here as well.)

iocore/net/SSLNetVConnection.cc
```diff
@@ -2205,15 +2262,45 @@ SSLNetVConnection::_ssl_read_buffer(void *buf, int64_t nbytes, int64_t &nread)
...
       if (ret == SSL_READ_EARLY_DATA_ERROR) {
-        Debug("ssl_early_data", "SSL_READ_EARLY_DATA_ERROR");
-        ssl_error = SSL_get_error(ssl, ret);
+        had_error_on_reading_early_data = true;
+        ssl_error                       = SSL_get_error(ssl, ret);
+      } else if (ret == SSL_READ_EARLY_DATA_FINISH) {
+        finished_reading_early_data = true;
+      }
+#else
+      if (SSL_in_early_data(ssl)) {
```
bneradt, line 2285: Let's also comment this as the BoringSSL version.

bneradt (changes requested):
Looks good. Just a few suggestions for comments we can add.

Maybe we can set up BoringSSL to run for one of the PR jobs in CI? Perhaps we can use the Fedora run, for instance, to run with BoringSSL.

bneradt:
> Looks good. Just a few suggestions for comments we can add.
> 
> Maybe we can set up BoringSSL to run for one of the PR jobs in CI? Perhaps we can use the Fedora run, for instance, to run with BoringSSL.

For the record, this seems to do the trick:

```
docker pull controller.trafficserver.org/ats/fedora:33
docker run -it controller.trafficserver.org/ats/fedora:33 /bin/bash

# Now in the docker container, running as root.
yum install -y cmake go
cd /opt
git clone https://boringssl.googlesource.com/boringssl
cd boringssl
mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE=Release ..
make -j $(nproc)
cd ../
mkdir lib
cd lib
ln -s ../build/ssl/libssl.a
ln -s ../build/crypto/libcrypto.a
```

We can add that to the controller.trafficserver.org/ats/fedora:33, making it a part of its image.

duke8253:
Looks good to me, let's fix @bneradt 's comments though.

duke8253 (approved):
Looks good.

bneradt (approved):
Looks good.

PR #8611, opened 2022-01-18
Removes hard dependency on having perl installed to build

randall:
If perl is not detected, the Perl module (`Apache::TS`) will not be built.


bneradt:
The patch is straightforward enough. What doesn't get build if you don't have Perl on the system?

randall:
> The patch is straightforward enough. What doesn't get build if you don't have Perl on the system?

The Apache::TS perl module, which already had the conditional logic to disable building if no perl.

bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8612, opened 2022-01-18
Introduce TLSTunnelSupport interface

maskit:
This introduces a mix-in class for TLS tunneling stuff.

The goal of this PR is to remove `dynamic_cast<SSLNetVConnection *>` from `HttpSM` and cleanup SSLNetVC a bit. Although some code that can be moved to the mix-in class is remained, code for tunneling is spread over multiple modules and I'd like to hand over that part to someone who is familiar with tunneling stuff.

On this PR, although there are two small changes, I simply moved functions that are just for tunneling into the mix-in class, so the logic isn't changed. Please see inline comments for the two changes

### DIFF
--- a/iocore/net/SSLNetVConnection.cc
+++ b/iocore/net/SSLNetVConnection.cc
@@ -219,6 +219,7 @@ SSLNetVConnection::_bindSSLObject()
   TLSSessionResumptionSupport::bind(this->ssl, this);
   TLSSNISupport::bind(this->ssl, this);
   TLSEarlyDataSupport::bind(this->ssl, this);
+  TLSTunnelSupport::bind(this->ssl, this);
 }
 
 void
@@ -230,6 +231,7 @@ SSLNetVConnection::_unbindSSLObject()
   TLSSessionResumptionSupport::unbind(this->ssl);
   TLSSNISupport::unbind(this->ssl);
   TLSEarlyDataSupport::unbind(this->ssl);
+  TLSTunnelSupport::unbind(this->ssl);
 }
 
 static void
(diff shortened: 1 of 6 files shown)


iocore/net/SSLNetVConnection.cc
```diff
@@ -991,8 +994,6 @@ SSLNetVConnection::free(EThread *t)
   }
   con.close();
 
-  ats_free(tunnel_host);
```
maskit, line 994: This part was moved into `TLSTunnelSupport::_clear()`. It should be ok as we call the function in `clear()` on line 1009.

proxy/http/HttpSM.cc
```diff
@@ -655,16 +655,16 @@ HttpSM::setup_blind_tunnel_port()
...
-        if (ssl_vc->get_tunnel_port() > 0) {
-          t_state.hdr_info.client_request.url_get()->port_set(ssl_vc->get_tunnel_port());
+        if (tts->get_tunnel_port() > 0) {
+          t_state.hdr_info.client_request.url_get()->port_set(tts->get_tunnel_port());
         } else {
           t_state.hdr_info.client_request.url_get()->port_set(netvc->get_local_port());
         }
       } else {
-        t_state.hdr_info.client_request.url_get()->host_set(ssl_vc->get_server_name(), strlen(ssl_vc->get_server_name()));
+        t_state.hdr_info.client_request.url_get()->host_set(netvc->get_server_name(), strlen(netvc->get_server_name()));
```
maskit, line 667: I don't think we need to use `ssl_vc` because `get_server_name` is a virtual function.

iocore/net/SSLNetVConnection.cc
duke8253: agreed.

proxy/http/HttpSM.cc
duke8253: agreed.

duke8253 (approved):
Looks reasonable, and a lot cleaner.

PR #8613, opened 2022-01-18
fix for collapsed forwarding ink_abort for CacheHitFresh fail

traeak:
This fixes an abort when
```CONFIG proxy.config.http.cache.open_write_fail_action INT 5```
is set which rewinds the state machine to before cache read/write.

This was tested by simulating a manifest that updates every 10s with max-age=1 cache control, hitting that asset with a thundering herd.

The failure was happening with a cache hit on a stale manifest.

While this "fix" allows ATS to to run without aborting, under testing 1 or more transactions end up with a client timeout with no bytes transferred.

### DIFF
--- a/proxy/http/HttpSM.cc
+++ b/proxy/http/HttpSM.cc
@@ -4743,6 +4743,9 @@ HttpSM::do_cache_lookup_and_read()
   // ink_assert(server_txn == NULL);
   ink_assert(pending_action.empty());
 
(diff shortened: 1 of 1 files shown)


traeak:
[approve ci debian]

ywkaras (approved):

zwoop:
Cherry-picked to v9.2.x

PR #7368, opened 2020-12-03
Cleanup: Remove SSL Wire Trace releated code in UnixNetVConnection

masaori335:
These `origin_trace` code are introduced as part of SSL Wire Trace feature by 6e3ce5022d67a9c4098595d3d45e24a3c697c80c. But the feature is removed by 6ee5c616d165b969b42a86c4eb2ca0b1be86b881 and these remaining code is not used anywhere.

Part of #4652.

Issue #4652, opened 2018-11-27
Should we get rid of the ssl.wire_trace feature?


shinrich:
We had used it for a while in an attempt to debug TLS issues and do some packet capture.  However, we have moved on to use other mechanisms.  If others are not using this feature, we should probably remove it.  I fear that it will fail with other TLS reworks if no one is watching it.

zwoop:
I agree.

bryancall:
+1 we should remove it.

bryancall:
Blocks #5457

shinrich:
Looks like it is now gone.

bryancall:
It was removed in this commit: 6ee5c616d165b969b42a86c4eb2ca0b1be86b881

PR #8619, opened 2022-01-19
Remove leftovers from SSL wire tracing

maskit:
Code for SSL wire tracing was partially removed by #4677 and #7368 but there are still some leftovers. This is an additional cleanup for #4652.


bneradt:
[approve ci autest]

bneradt (approved):

PR #8620, opened 2022-01-19
is_stale_cache_response_returnable: verify HTTP method

bneradt:
This adds the method verification check currently in
is_cache_response_returnable to is_stale_cache_response_returnable. That
is, if an object is being requested by the client and the object is in
our cache, we can only reply with the cached object if the method of the
incoming request is the same as the method of the request for the cached
object.

---

This PR came out of a discussion with @serrislew about the check here:
https://github.com/apache/trafficserver/blob/master/proxy/http/HttpTransact.cc#L1958

We shouldn't build and reply with that response if the method of the incoming request doesn't match the method of the request that resulted in the cached response.

### DIFF
--- a/proxy/http/HttpTransact.cc
+++ b/proxy/http/HttpTransact.cc
@@ -6079,6 +6079,19 @@ HttpTransact::is_stale_cache_response_returnable(State *s)
     return false;
   }
 
+  // We may be caching responses to methods other than GET, such as POST. Make
+  // sure that our cached resource has a method that matches the incoming
+  // requests's method. If not, then we cannot reply with the cached resource.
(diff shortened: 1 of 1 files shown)


vmamidi (approved):

mlibbey:
If the stale object method is GET, will we still be able to serve HEAD from cache?

bneradt:
[approve ci]

bneradt:
[approve ci]

bneradt:
[approve ci fedora]

bneradt:
[approve ci fedora]

bneradt:
[approve ci fedora]

bneradt:
[approve ci rockylinux]

PR #8621, opened 2022-01-19
Commenting TSHttpTxnCacheLookupStatusGet need_to_revalidate

bneradt:
Adding a comment explaining why we return a MISS when need_to_revalidate
fails as it would be natural to consider returning STALE in such
circumstances.


randall (approved):

zwoop:
Cherry-picked to v9.2.x

bneradt:
Adding the 9.2.x project as this will be reverted in 9.2.x when this is cherry-picked back:
https://github.com/apache/trafficserver/pull/8637

PR #8625, opened 2022-01-21
Rate Limit Plugin: Re-enable VConnection when SNI is empty

masaori335:
When the rate limit plugin is used as a global plugin, the VConnection without SNI should be re-enabled immediately.


bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8626, opened 2022-01-21
Fix 32bit build failure on Odroid Xu-4

AD5GB:
It's and oddball device but ATS now builds and runs well on it.
Also built and tested on OpenSuSE  Tumbleweed, Fedora-35 and Centos-8 aarch64 and x86_64 platforms.

Signed-off-by: Randy DuCharme <radio.ad5gb@gmail.com>

### DIFF
--- a/proxy/logging/LogUtils.cc
+++ b/proxy/logging/LogUtils.cc
@@ -599,7 +599,7 @@ LogUtils::file_is_writeable(const char *full_filename, off_t *size_bytes, bool *
     if (e < 0) {
       ret_val = -1;
     } else {
-      if (limit_data.rlim_cur != static_cast<rlim_t> RLIM_INFINITY) {
+      if (limit_data.rlim_cur != static_cast<rlim_t>(RLIM_INFINITY)) {
         if (has_size_limit) {
           *has_size_limit = true;
         }


masaori335:
[approve ci]

bneradt:
[approve ci]

bneradt (approved):

proxy/logging/LogUtils.cc
```diff
@@ -599,7 +599,7 @@ LogUtils::file_is_writeable(const char *full_filename, off_t *size_bytes, bool *
     if (e < 0) {
       ret_val = -1;
     } else {
-      if (limit_data.rlim_cur != static_cast<rlim_t> RLIM_INFINITY) {
+      if (limit_data.rlim_cur != static_cast<rlim_t>(RLIM_INFINITY)) {
```
ywkaras, line 602: Thanks, this is good code cleanup in general.  I think the only reason it hasn't broken for x86 build is that RLIM_INFINITY is implemented as macro with parentheses enclosing the value:
```
# define RLIM_INFINITY ((__rlim_t) -1)
```

ywkaras (approved):

proxy/logging/LogUtils.cc
bneradt: Ah, thanks @ywkaras. I was wondering how this built before.

zwoop:
Cherry-picked to v9.2.x

PR #8632, opened 2022-01-27
Default proxy.config.http.strict_uri_parsing to "2"

zwoop:
Following up on Susan's #8012, this patch makes the default parsing
option "2", which enforces no white spaces in URL components.


bryancall (approved):

ywkaras (approved):

maskit (approved):
I think release note should mention this change since it changes the default behavior.

zwoop:
Cherry-picked to v9.2.x

ezelkow1:
Can we get a backport to 8.1.x for this?

PR #8640, opened 2022-02-02
Adding clangd language server files to .gitignore

bneradt:
Adding some ignore entries for files generated by the clangd language
server for editors that use it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,9 +18,15 @@ Makefile.in
 
 .dirstamp
 
+# IDE generated files.
 .project
 .cproject
 .idea
+
+# clangd language server files.
+/compile_commands.json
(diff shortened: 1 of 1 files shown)


bneradt:
Marking for the various supported releases since this is could be helpful for anyone editing files in those releases while it is low risk since it's not a product-impacting change.

masaori335 (approved):
Actually, I have this in my local git config :)

ezelkow1:

bneradt:
> If we want this for 8.x can we get a backport?

Sure thing:
https://github.com/apache/trafficserver/pull/8652

zwoop:
Cherry-picked to v9.2.x

PR #8641, opened 2022-02-03
Test: Make trafficserver ext test use runroot file by default.

brbzull0:
The idea of this change is to let  `traffic_server` and `traffic_ctl` use  the `runroot` file by defaut in the tests, this can be disable when calling `MakeATSProcess` with `dump_runroot=False`.
This change also:
- Removes the previous and ambiguos way to set up the runroot, which it was introduced in 10-Dev only.
- Updates docs.

Note:
I am expecting the regex_revalidate_miss test to fail, it seems to be a transient error when updating the  config file which is not picked up properly by the config reloader as the two versions shows the same timestamp. There is another [PR](https://github.com/apache/trafficserver/pull/8644) to fix this.

### DIFF
--- a/doc/appendices/command-line/traffic_ctl_jsonrpc.en.rst
+++ b/doc/appendices/command-line/traffic_ctl_jsonrpc.en.rst
@@ -570,38 +570,9 @@ Configure Traffic Server to insert ``Via`` header in the response to the client
 Autest
 ======
 
-If you want to interact with |TS| under a unit test, then a few things need to be considered.
-
-- Runroot needs to be configured in order to  let `traffic_ctl` knows where to find the socket.
-
-   There are currently two ways to do this:
-
-   1. Using `run-root` param.
-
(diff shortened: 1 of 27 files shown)


brbzull0:
[approve ci autest]

brbzull0:
```
[Feb  4 13:14:22.315] [ET_TASK 1] DIAG: (regex_revalidate) In config Handler
[Feb  4 13:14:22.315] [ET_TASK 1] DIAG: (regex_revalidate) File mod time is not newer: 1643980462 >= 1643980462
[Feb  4 13:14:22.315] [ET_TASK 1] DIAG: (regex_revalidate) No Changes
```

https://github.com/apache/trafficserver/pull/8644 should fix this. I can rebase this PR after the other is merged in.

bneradt (approved):

randall:
[approve ci autest]

PR #8644, opened 2022-02-04
Autest: Add delay before updating a config file. (regex_revalidate_miss)

brbzull0:
Adding a delay before updating the regex revalidate config file so the reload can catch up the diff between config files timestamps.


bneradt (approved):
This test was inconsistent before. With this patch, it now passes consistently.

Thanks @brbzull0 !

PR #8645, opened 2022-02-04
Make TsSharedMutex.h compile on MacOS.

ywkaras:


zwoop:
@ywkaras Do we need this for 9.1.x too ?

cmcfarlen (approved):
This PR worked for me compiling on mac.

zwoop (approved):

bneradt (approved):

ywkaras:
@zwoop I think we can live without it in 9.1.  Worst case, without it, we'd see a memory corruption crash when we were already executing a Fatal() or TSFatal().

zwoop:
Cherry-picked to v9.2.x

PR #8649, opened 2022-02-07
Fix plugin parent_select missing hostname len

rob05c:
I'm not sure if this was causing any issues. Most of core assumes the hostname is null-terminated. I only noticed it because some debug prints were broken. But it's definitely a bug, whether or not it has immediate symptoms.


ywkaras (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8650, opened 2022-02-07
Fix strategies to initialize scheme

rob05c:
This makes strategies (both core and plugin) default initialize the scheme to NONE.

Without this, omitting the scheme in strategies.yaml causes an assertion failure during a runtime request.

With this, a strategy with no scheme specified will match a host with no scheme, allowing users to omit the scheme.

This "scheme" is only ever used to match strategies to hosts, the actual scheme requested of the parent is the remap target.

This also updates the docs to note the scheme is only used to match, not request, and that it's optional.


masaori335 (approved):
Looks reasonable.

zwoop:
Cherry-picked to v9.2.x

PR #8661, opened 2022-02-10
Making 9.2.x backwards compatible with 9.1.x

bneradt:
In order to make 9.2.x backwards compatible against plugins compiled
with 9.1.x, this PR does the following:

1. Adds back exposed enum values removed in 9.2.x
2. Any added enum values are put at the end of their enumeration lists.
3. Adds back the TSHttpTxnCntl function


bneradt:
We'll want this for the 9.x trains (currently 9.2.x and master), but not for the 10.x train. That experimental `TSHttpTxnCntl` should not exist in 10.x. Thus after this is merged, I'll merge up 10-Dev and then revert this patch from there.

bneradt:
[approve ci centos]

bneradt:
[approve ci debian]

bneradt:
[approve ci fedora]

bneradt:
[approve ci ubuntu]

bneradt:
[approve ci centos]

bneradt:
[approve ci debian]

bneradt:
[approve ci ubuntu]

zwoop:
Are you sure we really need that experimental crud? Is someone using it? :-). The Lua feature has been updated already to use the new API and should be in 9.2.x.

bneradt:
> Are you sure we really need that experimental crud? Is someone using it? :-). The Lua feature has been updated already to use the new API and should be in 9.2.x.

I vote to remove it in 10.x, but keep it for 9.2.x. Yahoo has an internal plugin that uses the `TSHttpTxnCntl` function.

bneradt:
Oh, I see. The regression tests try to use the `proxy.client.verify.server` config via the `TS_CONFIG_SSL_CLIENT_VERIFY_SERVER` which I put back in to preserve the ABI, but it fails because none of that is implemented now.

@zwoop / @randall : I think we'll have to back out #7639 from 9.2.x. 10.x will be fine, for sure, but I don't think we can remove `porxy.client.verify.server` and be compatible with the previous 9.x branches.

bneradt:
> Oh, I see. The regression tests try to use the `proxy.client.verify.server` config via the `TS_CONFIG_SSL_CLIENT_VERIFY_SERVER` which I put back in to preserve the ABI, but it fails because none of that is implemented now.
> 
> @zwoop / @randall : I think we'll have to back out #7639 from 9.2.x. 10.x will be fine, for sure, but I don't think we can remove `porxy.client.verify.server` and be compatible with the previous 9.x branches.

As an alternative to this, I'm updating the InkAPITest.cc to handle the defunct `proxy.client.verify.server` config. If this seems OK to @randall and @zwoop, then we can move forward with that.

bneradt:
[approve ci]

zwoop (approved):

bryancall:
I am still seeing some issue with data types:
https://trafficserver.apache.org/compat_reports/traffic_server/9.1.x_to_master/compat_report.html

bneradt:
> I am still seeing some issue with data types:
> https://trafficserver.apache.org/compat_reports/traffic_server/9.1.x_to_master/compat_report.html

We talked about this offline, but just to record it here: this will happen on master because the configure.ac defines `TS_VERSION_MAJOR` as 10, so some enum values and definitions are removed in master. This shouldn't be a problem once this patch is picked back to 9.2.x.

zwoop:
Cherry-picked to v9.2.x

PR #8665, opened 2022-02-11
Fixed issues when compiling with -Og

bryancall:
abi-dumper likes to have binaries compiled with -Og for better testing of binary compatibility.

https://github.com/lvc/abi-dumper


masaori335 (approved):
Looks harmless & reasonable.

proxy/http2/HTTP2.cc
```diff
@@ -606,7 +606,7 @@ http2_convert_header_from_1_1_to_2(HTTPHdr *headers)
 
     // :path
     if (MIMEField *field = headers->field_find(HTTP2_VALUE_PATH, HTTP2_LEN_PATH); field != nullptr) {
-      int value_len;
+      int value_len     = 0;
```
masaori335, line 609: It's pretty interesting that line 589 above is almost the same situation but doesn't need initialization.

PR #8666, opened 2022-02-11
slice and cache_range_requests: allow header override

traeak:
This adds the ability to override the internal slice skip header (was hard coded to X-Slicer-Info) and the cache_range_requests ims header (was hard coded to X-Crr-Ims).

### DIFF
--- a/plugins/cache_range_requests/cache_range_requests.cc
+++ b/plugins/cache_range_requests/cache_range_requests.cc
@@ -47,11 +47,14 @@ using parent_select_mode_t = enum parent_select_mode {
   PS_CACHEKEY_URL, // Set parent selection url to cache_key url
 };
 
+constexpr std::string_view DefaultImsHeader = {"X-Crr-Ims"};
+
 struct pluginconfig {
   parent_select_mode_t ps_mode{PS_DEFAULT};
   bool consider_ims_header{false};
   bool modify_cache_key{true};
   bool verify_cacheability{false};
+  std::string ims_header;
 };
 
 struct txndata {
@@ -60,9 +63,6 @@ struct txndata {
   time_t ims_time{0};
   bool verify_cacheability{false};
 };
-
-// Header for optional revalidation
-constexpr std::string_view X_IMS_HEADER = {"X-Crr-Ims"};
 
 // pluginconfig struct (global plugin only)
 pluginconfig *gPluginConfig = {nullptr};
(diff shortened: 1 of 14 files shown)


bryancall:
@elsloo Can you please review this?

plugins/cache_range_requests/cache_range_requests.cc
```diff
@@ -111,24 +112,29 @@ create_pluginconfig(int argc, char *const argv[])
...
     }
 
     switch (opt) {
-    case 'p': {
-      DEBUG_LOG("Plugin modifies parent selection key");
-      pc->ps_mode = PS_CACHEKEY_URL;
-    } break;
     case 'c': {
-      DEBUG_LOG("Plugin considers the '%.*s' header", (int)X_IMS_HEADER.size(), X_IMS_HEADER.data());
+      DEBUG_LOG("Plugin considers the CRR ims header");
```
elsloo, line 122: Can this be reverted such that the default header is printed out when a custom header is not in use? Specifically, print the value of `DefaultImsHeader` here.

plugins/experimental/slice/slice.h
```diff
@@ -32,8 +32,6 @@
 #define PLUGIN_NAME "slice"
 #endif
 
-constexpr std::string_view X_CRR_IMS_HEADER = {"X-Crr-Ims"};
```
elsloo, line 35: This constant and all references to `X_CRR_IMS_HEADER` were removed, but `SLICER_MIME_FIELD_INFO` still exists in `HttpHeader.h` (line 39). Consequently there are two areas that still reference the old default `X-Slicer-Info` header name using the constant, and it is injected into requests despite having a "custom" header configuration.

plugins/experimental/slice/client.cc
```diff
@@ -74,7 +75,7 @@ handle_client_req(TSCont contp, TSEvent event, Data *const data)
         data->m_statustype = TS_HTTP_STATUS_REQUESTED_RANGE_NOT_SATISFIABLE;
 
         // First block will give Content-Length
-        rangebe = Range(0, data->m_config->m_blockbytes);
+        rangebe = Range(0, conf->m_blockbytes);
       }
     } else {
       DEBUG_LOG("%p Full content request", data);
```
elsloo, line 81: Please mitigate the reference to `SLICER_MIME_FIELD_INFO` on line 84 below.

plugins/experimental/slice/server.cc
```diff
@@ -268,8 +268,8 @@ logSliceError(char const *const message, Data const *const data, HttpHeader cons
   crange.toStringClosed(normstr, &normlen);
```
elsloo, line 268: Please mitigate the reference to `SLICER_MIME_FIELD_INFO` on line 262 above.

elsloo (changes requested):
Please remove the `SLICER_MIME_FIELD_INFO` constant and update the two dependent lines of code to use the new member variables.

plugins/cache_range_requests/cache_range_requests.cc
traeak: The other 2 messages in the same function should cover it (based on override being used or not):

DEBUG_LOG("Plugin uses custom CRR ims header: %s", optarg);
DEBUG_LOG("Plugin uses default ims header: %s", pc->ims_header.c_str())

plugins/experimental/slice/client.cc
traeak: stupid missed this.

plugins/experimental/slice/server.cc
traeak: missed this sry

ezelkow1:
[approve ci rocky]

elsloo (approved):
I think this looks good now. Thanks for addressing the feedback @traeak!

zwoop:
This was back ported to 9.2.x in #8898.

PR #8667, opened 2022-02-14
TLS Session Resumption: fix timed out session

moonchen:
A recent change to use shared_ptr for sessions change modified the behavior of timed out sessions in `getOriginSession()`.  This change reverts the behavior of `getOriginSession()` for a timed out session to the previous behavior, where it returns a null session pointer.


duke8253 (approved):
Looks like I missed this one when cleaning things up.

zwoop:
Cherry-picked to v9.2.x

zwoop:
@moonchen We don't need this for 9.1.x do we ?

moonchen:
> @moonchen We don't need this for 9.1.x do we ?

I don't think so.  The shared_ptr code isn't in 9.1.x.

PR #8669, opened 2022-02-14
Fix a libswoc build error on macOS

moonchen:
This change fixes a build failure on macOS by including `<array>` in `swoc_ip.h`.

### DIFF
--- a/lib/swoc/include/swoc/swoc_ip.h
+++ b/lib/swoc/include/swoc/swoc_ip.h
@@ -10,6 +10,7 @@
 #include <netinet/in.h>
 #include <string_view>
 #include <variant>
+#include <array>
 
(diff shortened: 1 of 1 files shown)


bneradt:
The CentOS build failed due to a temporary issue. I'm using it to try to triage regression tests failures. Let me mark it as offline for now. After the rest of the tests run, we'll rerun the centos run and it should grab a new machine.

bneradt:
[approve ci centos]

bneradt (approved):

bneradt:
I'll merge this in so that we can resolve this build issue quickly for those who need it, but @moonchen agreed to file this as well in libswoc to make sure we don't lose it from there.

PR #8673, opened 2022-02-16
Fix to allow running autopep8 from outside top_srcdir

cmcfarlen:
Sometimes I like to build in a subdirectory but this script is not happy with that.


bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8675, opened 2022-02-17
Make sure each remap rule gets its own continuation

zwoop:
This problem means that statichit can only safely be used by one remap rule, otherwise they will stomp over each other on the single cont-data member.


bryancall (approved):

zwoop:

PR #8676, opened 2022-02-17
Fix plugin parent_select failover

rob05c:
This fixes the parent_select plugin failover. 

Failing over to the next parent wasn't working right in a number of scenarios. 

I've tested these changes up to the 3rd parent, for DNS failures, connection failures, and strategy-based code failures.

They all seem to work as-expected, take the same amount of time as parent.config, and inspecting debug logs seems to confirm it's working as-expected, parents aren't being tried repeatedly when they shouldn't, marked down correctly per the strategy, etc.


ezelkow1:
[approve ci rocky]

ywkaras (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8679, opened 2022-02-18
Update asf.yaml to require Rocky over CentOS build

ezelkow1:
We have changed the CI to use rocky now that centos is defunct, need to update the asf to match


bneradt (approved):

ezelkow1:
[approve ci]

PR #8683, opened 2022-02-19
Upgrade Catch to v2.13.8

bneradt:
This upgrades our Catch unit test framework to version v2.13.8 from
v2.13.7.


bneradt:
[approve ci autest]

randall (approved):

bneradt:
Marking this for the 9.1.x and 9.2.x branches since this will allow Catch to be built with newer, more strict warning mechanisms of recent compilers. Note that this is a test-only change.

zwoop:
Cherry-picked to v9.2.x

PR #8684, opened 2022-02-19
Fix warnings from GCC 12.0.1

bneradt:
This patch fixes warnings generated by GCC 12.0.1. The warnings involved the
use of the deprecated std::binary_function, std::unary_function, and
std::iterator interfaces. The use of these was considered more confusing than
explicitly declaring the associated types. Therefore this patch replaces the
use of these deprecated interfaces with the declaration of the associated
types.

### DIFF
--- a/example/plugins/cpp-api/boom/boom.cc
+++ b/example/plugins/cpp-api/boom/boom.cc
@@ -98,13 +98,16 @@ GlobalPlugin *plugin;
 // Functor that decides whether the HTTP error can be rewritten or not.
 // Rewritable codes are: 2xx, 3xx, 4xx, 5xx and 6xx.
 // 1xx is NOT rewritable!
-class IsRewritableCode : public std::unary_function<std::string, bool>
+class IsRewritableCode
 { // could probably be replaced with mem_ptr_fun()..
 private:
   int current_code_;
   std::string current_code_string_;
 
 public:
+  using argument_type = std::string;
+  using result_type   = bool;
+
   explicit IsRewritableCode(int current_code) : current_code_(current_code)
   {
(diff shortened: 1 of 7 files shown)


masaori335 (approved):
Looks good. This is following the guidance, getting rid of the inheritance and defining aliases explicitly.

masaori335:
It looks like LLVM-15 removed the function. We need to backport this for release branches.

```
../../include/tscore/IntrusivePtr.h:333:29: error: no template named 'binary_function' in namespace 'std'; did you mean '__binary_function'?
  class Order : public std::binary_function<IntrusivePtr<T>, IntrusivePtr<T>, bool>
                       ~~~~~^~~~~~~~~~~~~~~
                            __binary_function
/usr/local/opt/llvm/include/c++/v1/__functional/binary_function.h:49:1: note: '__binary_function' declared here
using __binary_function = __binary_function_keep_layout_base<_Arg1, _Arg2, _Result>;
^
```

BTW, std::binay_function was "deprecated in C++11" and "removed in C++17".
https://en.cppreference.com/w/cpp/utility/functional/binary_function

bneradt:
> It looks like LLVM-15 removed the function (or my env is broken). We need to backport this for release branches.

Just to document things: I have already created separate 8.x and 9.2.x PRs for this. This is only missing from 9.1.x, and our current intention is to release 9.2.x rather than creating another 9.1.x release. If we wind up creating a 9.1.x release, we can work on a PR for this. This kind of patch usually doesn't backport cleanly because each branch has its own set of compilation issues and we should build it using the latest compilers and fix the issues seen by them in a specific PR.

PR #6996, opened 2020-07-11
Make the setting of the continuation handler safer.

ywkaras:
These three code examples only differ in the first two lines:
https://godbolt.org/z/Gq38j7
https://godbolt.org/z/hGb9PT
https://godbolt.org/z/6MdKjn

They illustrate how using a C-style cast to convert a derived
class member function pointer to a base class member function
pointer bypasses a compile time check that is derived class is
actually derived from the base class.

PR #8689, opened 2022-02-21
Hoststatus remove stats creation.

jrushford:
Modifications to the traffic_ctl host status command and HostStatus

- modifies traffic_ctl to use JSON-RPC to query host statuses directly from the HostStatus singleton
- removes the persistent stats creation in HostStatus
- adds in a new persistent mechanism for HostStatus

Host status will no-longer create persistent stats so now when a host is marked up or down using **traffic_ctl,** the entire set of HostStatus records will be dumped from memory to var/trafficserver/host_records.yaml so that when ATS restarts, the records are re-loaded.  stats_over_http will no-longer show **"proxy.process.host_status"** stats and are no-longer seen with **traffic_ctl metric match**.

**traffic_ctl host status** without any other arguments will list all host status records replacing the metric match usage to list all host records.


bneradt:
[approve ci rocky]

bneradt:
[approve ci rocky]

brbzull0:
https://github.com/apache/trafficserver/pull/8694 should fix the autest. Once it's merged into 10-Dev you can rebase.

jrushford:
[approve ci autest]

bneradt:
[approve ci autest]

bneradt:
[approve ci fedora]

bneradt:
[approve ci autest]

src/traffic_server/traffic_server.cc
```diff
@@ -2162,7 +2162,6 @@ main(int /* argc ATS_UNUSED */, const char **argv)
     RecProcessStart();
     initCacheControl();
     IpAllow::startup();
-    HostStatus::instance().loadHostStatusFromStats();
```
brbzull0, line 2165: This was key I think, it was creating the first  `HostStatus` instance, without this the ctor will not be called so no handlers registered.  You should probably need a `init()` function to register the handlers.

jrushford:
[approve ci Fedora]

jrushford:
[approve ci fedora]

jrushford:
@brbzull0 I've added in the persistence.  You'll see that I added back in the HostStatus call in traffic_server.cc with the new HostStatus::loadFromPersistentStore() function

jrushford:
[approve ci clang-analyzer

jrushford:
[approve ci clang-analyzer]

jrushford:
[approve ci fedora]

brbzull0:
@jrushford this looks good, just have a question.
Do you think you will be injectingthe `host_records.yaml` into traffic_server through the JSON-RPC node?(by using traffic_ctl or any other tool). If this is the plan at some point maybe the field names in the rpc request and the dump `host_records.yaml` file should be the same in order to re-use them.

jrushford:
> @jrushford this looks good, just have a question.
> Do you think you will be injectingthe `host_records.yaml` into traffic_server through the JSON-RPC node?(by using traffic_ctl or any other tool). If this is the plan at some point maybe the field names in the rpc request and the dump `host_records.yaml` file should be the same in order to re-use them.

@brbzull0 I didn't have any plans to use the JSON-RPC node to inject them but, that could be useful. In this PR, the host_records.yaml is loaded at startup by trafficserver.  However, I'll take a look and make sure that the names are the same, someone might have a use case for the injection.

proxy/HostStatus.h
```diff
@@ -168,6 +175,35 @@ struct HostStatRec {
   }
 };
 
+struct HostStatusSync;
+typedef int (HostStatusSync::*HostStatusSyncHandler)(int, void *);
```
ywkaras, line 179: Don't need this.

proxy/HostStatus.h
```diff
@@ -168,6 +175,35 @@ struct HostStatRec {
...
+    std::string rundir(RecConfigReadRuntimeDir());
+    hostRecordsFile = Layout::relative_to(rundir, ts::filename::HOST_RECORDS);
+  }
+
+public:
+  HostStatusSync()
+  {
+    getHostStatusPersistentFilePath();
+
+    SET_HANDLER((HostStatusSyncHandler)&HostStatusSync::mainEvent);
```
ywkaras, line 196: You should not need to cast the type of the event function anymore because of https://github.com/apache/trafficserver/pull/6996 .

It is best to not use C casts in new code.  A C cast is equivalent to reinterpret_cast and is hard to find with grep.

proxy/HostStatus.h
```diff
@@ -182,11 +218,16 @@ struct HostStatus {
...
   void setHostStatus(const std::string_view name, const TSHostStatus status, const unsigned int down_time,
                      const unsigned int reason);
-  HostStatRec *getHostStatus(const std::string_view name);
-  void createHostStat(const std::string_view name, const char *data = nullptr);
-  void loadHostStatusFromStats();
+  void loadFromPersistentStore();
   void loadRecord(std::string_view name, HostStatRec &h);
-  RecErrT getHostStat(std::string &stat_name, char *buf, unsigned int buf_len);
+  HostStatRec *getHostStatus(const std::string_view name);
+  void getAllHostStatuses(std::vector<HostStatuses> &hosts);
```
ywkaras, line 224: You could return a vector, it should be efficient due to Return Value Optimization.

src/traffic_ctl_jsonrpc/jsonrpc/CtrlRPCRequests.h
```diff
@@ -129,10 +139,16 @@ struct HostSetStatusRequest : shared::rpc::ClientRequest {
...
-  using super                         = shared::rpc::RecordLookupRequest;
-  HostGetStatusRequest() : super() {}
+struct HostGetStatusRequest : shared::rpc::ClientRequest {
+  using super  = shared::rpc::ClientRequest;
+  using Params = std::vector<std::string>;
+  HostGetStatusRequest(Params p) { super::params = std::move(p); }
+
+  std::string
+  get_method() const
+  {
```
ywkaras, line 149: @brbzull0 it's not clear why this even needs to be a virtual function:  https://github.com/apache/trafficserver/blob/68e69e27b87576abfaa7a1bbeb8394795c522e82/include/shared/rpc/RPCRequests.h#L45

Should derived classes just set the "method" member string?

src/traffic_server/HostStatus.cc
```diff
@@ -20,85 +20,32 @@
...
-    Debug("host_statuses", "marking down server %s", name);
-    HostStatus &hs = HostStatus::instance();
-    if (hs.getHostStat(stat_name, buf, 1024) == REC_ERR_FAIL) {
-      hs.createHostStat(name);
-    }
-    hs.setHostStatus(name, TSHostStatus::TS_HOST_STATUS_DOWN, down_time, reason);
-  }
-}
+ts::Rv<YAML::Node> server_get_status(std::string_view const &id, YAML::Node const &params);
+ts::Rv<YAML::Node> server_set_status(std::string_view const &id, YAML::Node const &params);
```
ywkaras, line 48: Why reference to string_view ?

doc/appendices/command-line/traffic_ctl_jsonrpc.en.rst
```diff
@@ -366,7 +365,8 @@ endpoint.
    :ref:`admin_lookup_records`
 
    Get the current status of the specified hosts with respect to their use as targets for parent
-   selection. This returns the same information as the per host stat.
+   selection. This returns the same information as the per host record. If the HOSTNAME arguments
```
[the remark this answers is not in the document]
ywkaras: But the status subcommand is explained in line 328.  It seems like what this is saying is that the host record returns the same information as the host record, so maybe we can just delete that sentence.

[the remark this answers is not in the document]
jrushford: ahh ok, I'll do that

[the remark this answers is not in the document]
jrushford: @ywkaras ok, it's removed.

proxy/HostStatus.h
jrushford: @ywkaras ok, it's removed.

src/traffic_server/HostStatus.cc
jrushford: @ywkaras okay, will pass by value.

src/traffic_server/HostStatus.cc
```diff
@@ -44,8 +44,8 @@ struct HostCmdInfo {
 
 } // namespace
 
-ts::Rv<YAML::Node> server_get_status(std::string_view const &id, YAML::Node const &params);
-ts::Rv<YAML::Node> server_set_status(std::string_view const &id, YAML::Node const &params);
+ts::Rv<YAML::Node> server_get_status(std::string_view const id, YAML::Node const &params);
+ts::Rv<YAML::Node> server_set_status(std::string_view const id, YAML::Node const &params);
```
ywkaras, line 48: It's redundant to use 'string_view const' instead of just plain 'string_view' but harmless.

proxy/HostStatus.h
ywkaras: We should let the people who went to grad school tell us if it's wrong.

ywkaras:
Looks good to me, let's see what Damian says.

src/traffic_server/HostStatus.cc
jrushford: @ywkaras I'll change it.

jrushford:
[approve ci fedora]

jrushford:
[approve ci fedora]

jrushford:
[approve ci]

src/traffic_ctl_jsonrpc/jsonrpc/CtrlRPCRequests.h
brbzull0: > it's not clear why this even needs to be a virtual function

Because of this https://github.com/apache/trafficserver/blob/68e69e27b87576abfaa7a1bbeb8394795c522e82/include/shared/rpc/yaml_codecs.h#L211
 
>Should derived classes just set the "method" member string?

yes they can, in which case `JSONRPCRequest::get_method` will return it.

I wasn't very concern about the virtual thing in here as this is used by `traffic_ctl` and `traffic_top`.

src/traffic_server/HostStatus.cc
brbzull0: This is the RPC handler API:

https://github.com/apache/trafficserver/blob/68e69e27b87576abfaa7a1bbeb8394795c522e82/mgmt2/rpc/jsonrpc/JsonRPCManager.h#L64

The reason is `const&` is becase I wanted  to make it clear that this should not mutate despite the fact that you cannot change it if it was only a `std::string_view`. I always thought that this is ambiguous and if someone try to change it well, good luck. Now kind of I am inclined to remove the `const &`

src/traffic_server/HostStatus.cc
brbzull0: > It's redundant to use 'string_view const' instead of just plain 'string_view' but harmless.

Totally, but the reason is this:

https://github.com/apache/trafficserver/blob/68e69e27b87576abfaa7a1bbeb8394795c522e82/mgmt2/rpc/jsonrpc/JsonRPCManager.h#L64

It is expected to be like this.

--
The reason is `const&` is becase I wanted  to make it clear that this should not mutate despite the fact that you cannot change it if it was only a `std::string_view`. I always thought that this is ambiguous and if someone try to change it well, good luck. Now kind of I am inclined to remove the `const &`

jrushford:
@brbzull0 and @ywkaras any more changes?  Are you ok with the `const &` removal?

proxy/HostStatus.h
jrushford: changed

ywkaras (approved):

brbzull0:
Looks good to me.

PR #8691, opened 2022-02-21
9.2.x: Fix test_QUIC unit test builds. (#8678)

bneradt:
The test_QUIC unit tests were failing to build because they didn't link
against a file with the TLSKeyLogger definition. This fixes the
undefined references by breaking out TLSKeyLogger into a separate object
that the unit tests can link with.

(cherry picked from commit 2d70a007933176c80b09ddeaf164206794671442)


masaori335:
[approve ci fedora rocky autest]

PR #8699, opened 2022-02-24
8.1.x: GCC 12: Address compiler warnings

bneradt:
This patch fixes warnings generated by GCC 12.0.1. The warnings involved the
use of the deprecated std::binary_function, std::unary_function, and
std::iterator interfaces. The use of these was considered more confusing than
explicitly declaring the associated types. Therefore this patch replaces the
use of these deprecated interfaces with the declaration of the associated
types.

This also addresses a couple uninitialized buffer warnings.


bneradt:
[approve ci fedora]

bneradt:
[approve ci fedora]

ezelkow1 (approved):
Approving to get in for other changes

PR #8704, opened 2022-02-25
traffic_ctl - Use better format output options.

brbzull0:
Use better output format options to display traffic_ctl JSONRPC messages. 
`all,req,resp` are now grouped into `rpc`.
```bash
$ traffic_ctl config get variable --format rpc
 --> {request}
 <-- {response}
 variable 123
```


bneradt:
[approve ci autest]

bryancall:
We had a discussion in the weekly PR/issue review meeting and most people thought `--output` would mean sending it to a file.  Leaving it as `--format` or `--output-format` might be better.

jrushford (approved):
looks good and tests ok.

PR #8705, opened 2022-02-27
Add back "DNS: Fix lack of nameserver failover in low use circumstances."

bneradt:
This adds back the fix for nameserver failover in low use circumstances
and addresses the regression test instability caused when the test box
has connectivity issues with the DNS server.

---

### For Review:
This is the exact change that was introduced via #7843, but it tweaks the regression test's configuration to not treat the DNS server for the test box as "dead." With #7843, the regression tests started failing when run on certain boxes (most problematically, CI's boxes). This makes those tests more robust against DNS nameserver issues.

### DIFF
--- a/iocore/dns/DNS.cc
+++ b/iocore/dns/DNS.cc
@@ -524,12 +524,12 @@ DNSHandler::open_con(sockaddr const *target, bool failed, int icon, bool over_tc
     }
     return false;
   } else {
-    ns_down[icon] = 0;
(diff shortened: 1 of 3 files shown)


bneradt:
[approve ci autest]

bneradt:
[approve ci autest]

SolidWallOfCode (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8707, opened 2022-03-02
cache-request-method autest: Extending IO delay

bneradt:
This tweaks a delay that waits upon IO to finish in the
cache-request-method autest which has frequently failed because the item
is not yet cached.


bneradt:
[approve ci fedora]

randall (approved):

bneradt:
Marking this for 9.2.x: it's a test only timing tweak that should make the test more stable.

zwoop:
Cherry-picked to v9.2.x

PR #8715, opened 2022-03-07
Fixes build when hwloc is not installed

randall:
Follow up to 08d35060

Co-authored-by: Chris McFarlen <cmcfarlen@apple.com>


bneradt (approved):

zwoop:
ink_hw.cc doesn't exist in 9.2.x I think, is there another file we should apply this change to ?

zwoop:
ah, it depended on another PR. 

Cherry-picked to v9.2.x

PR #8719, opened 2022-03-09
Fixed memory leaks with CryptoContext

bryancall:
This fixes a memory leak when switching over to the evp functions in OpenSSL.  I also removed support for the MMH hash for caches that are older than ATS 4.0.0, it was easier to remove the support.  The evp functions require malloc() for the calls vs the old OpenSSL APIs that have been deprecated in OpenSSL 3.  I think it is still possible to use the old APIs if you #define something.

Benchmarks of various hash implementations on an AMD Ryzen 9 3900X:
```
-----------------------------------------------------------------
Benchmark                       Time             CPU   Iterations
-----------------------------------------------------------------
openssl_old_api_md5          1312 ns         1310 ns       536272
openssl_evp_md5              1413 ns         1411 ns       497822
cryptopp_md5                 1448 ns         1445 ns       485112
openssl_old_api_sha256        586 ns          585 ns      1184803
openssl_evp_sha256            681 ns          680 ns      1035870
cryptopp_sha256               606 ns          605 ns      1159814
```

### DIFF
--- a/include/tscore/CryptoHash.h
+++ b/include/tscore/CryptoHash.h
@@ -23,6 +23,7 @@
 #pragma once
 
 #include "tscore/BufferWriter.h"
+#include <openssl/evp.h>
 #include <string_view>
 
 /// Apache Traffic Server commons.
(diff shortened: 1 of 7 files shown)


zwoop (approved):
Fixes the leak, increases CPU usage a little bit :).

PR #8736, opened 2022-03-16
[ink_base64] Fix buffer size computation

moonchen:
These buffer size computation macros are meant to be used for callers of
ats_base64_encode/decode.  They were not taking into account a null
terminator, which is always written by those functions.  This causes a
buffer overflow when encode or decode are called on buffers of certain
sizes, e.g. decode on an input buffer of size 4.

This change makes these buffer size calculations match the functions.
It also updates unit tests for the access_control plugin, which uses
these functions.

Since this is a macro change, plugins which use encode/decode will need to
be recompiled.

### DIFF
--- a/include/tscore/ink_base64.h
+++ b/include/tscore/ink_base64.h
@@ -42,5 +42,7 @@ bool ats_base64_encode(const unsigned char *inBuffer, size_t inBufferSize, char 
 bool ats_base64_decode(const char *inBuffer, size_t inBufferSize, unsigned char *outBuffer, size_t outBufSize, size_t *length);
 
 // Little helper functions to calculate minimum required output buffer for encoding/decoding.
-#define ATS_BASE64_ENCODE_DSTLEN(_length) ((_length * 8) / 6 + 4)
-#define ATS_BASE64_DECODE_DSTLEN(_length) (((_length + 3) / 4) * 3)
+// These sizes include one byte for null termination, because ats_base64_encode and ats_base64_decode will always write a null
+// terminator.
+#define ATS_BASE64_ENCODE_DSTLEN(_length) (((_length + 2) / 3) * 4 + 1)
+#define ATS_BASE64_DECODE_DSTLEN(_length) (((_length + 3) / 4) * 3 + 1)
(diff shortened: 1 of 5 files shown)


ywkaras:
[approve ci fedora]

ywkaras:
[approve ci Fedora]

ezelkow1:
[approve ci fedora]

include/tscore/ink_base64.h
```diff
@@ -42,5 +42,16 @@ bool ats_base64_encode(const unsigned char *inBuffer, size_t inBufferSize, char
...
+ats_base64_encode_dstlen(size_t length)
+{
+  return ((length + 2) / 3) * 4 + 1;
+}
+
+inline constexpr size_t
+ats_base64_decode_dstlen(size_t length)
+{
+  return ((length + 3) / 4) * 3 + 1;
+}
```
ywkaras, line 57: Why not?
```
inline constexpr size_t
ats_base64_encode_dstlen(size_t length, bool pad = true)
{
  size_t num_bit_pairs = 4 * length;
  return (pad ? (((num_bit_pairs + 11) / 12) * 4) : ((num_bit_pairs + 2) / 3)) + 1;
}

inline constexpr size_t
ats_base64_unpadded_decode_dstlen(size_t length)
{
  return (((length / 4) * 3) + (length % 4)) + 1;
}

inline size_t
ats_base64_padded_decode_dstlen(char const *data, size_t length)
{
  ink_assert(!(length % 4));

  if (length && ('=' == data[length - 1])) {
    --length;
  }
  if (length && ('=' == data[length - 1])) {
    --length;
  }

  return ats_base64_unpadded_decode_dstlen(length);
}
```

moonchen: The length calculation is meant for `ats_base64_encode()`, so the goal is to just give the length of the buffer needed to store the encoded output.  Both `ats_base64_encode()` and the [RFC](https://datatracker.ietf.org/doc/html/rfc4648#section-4) specify padding that is non-optional.

ywkaras: Then why not?
```
inline size_t
ats_base64_decode_dstlen(char const *data, size_t length)
{
  ink_assert(!(length % 4));

  size_t adjust = 0;
  if (length && ('=' == data[length - 1])) {
    adjust = ('=' == data[length - 2])) ? 2 : 1;
  }
  return (((length / 4) * 3) - adjust + 1;
}
```

moonchen: If I'm reading this right, you're trying to give a buffer size that is as small as possible, accounting for the possibility of padding characters at the end.  While I appreciate being frugal with allocation size, I think that the size savings of possibly one or two bytes may not warrant such an increase in complexity when computing the buffer size.

ywkaras (approved):

PR #8737, opened 2022-03-17
Require use of 'override' keyword when valid.

ywkaras:


ywkaras:
[approve ci fedora]

ywkaras:
https://www.fluentcpp.com/2020/02/21/virtual-final-and-override-in-cpp/

ywkaras:
Hmm looks like clang doesn't support -Wsuggest-override .

ywkaras:
https://reviews.llvm.org/D82728

masaori335 (approved):
Looks reasonable.

Issue #7824, opened 2021-05-12
The strategies.yaml parser can incorrectly interpret the YAML elements in the host protocol.


jrushford:
A user inadvertently created an invalid host definition in the strategies.yaml.  When configuring the **scheme** section of a **host** the user added the **port** element as an array field.  The parser allowed this but at runtime, the next hop lookup failed to get the proper port number which caused a runtime assertion failure.

Incorrect YAML:
`
     protocol:
        - scheme: http
        - port:  80
`

Correct YAML:
`
     protocol:
        - scheme: http
           port
`

jrushford:
After parsing the YAML, I think a check for all required settings should catch this problem.

ywkaras:
It seems like we should have seen this note in diags.log:

https://github.com/apache/trafficserver/blob/8a948afc6caafdbd4d28e63b377610f66ee8ea5b/proxy/http/remap/NextHopSelectionStrategy.cc#L76

Seems like we should either default the scheme to NH_SCHEME_HTTP, or call Emergency() rather than Note(), so TS doesn't continue running in unpredictable ways.

jrushford:
Fixed with PR #8742

jrushford:
Fixed by PR #87442

jrushford:
Fixed by PR #8742

PR #8742, opened 2022-03-18
Fixes Issue #7824 - The strategies.yaml parser can incorrectly interpret

jrushford:
Fixes Issue #7824 - The strategies.yaml parser can incorrectly interpret the YAML elements in the host protocol in both core and the parent_select plugin.


ywkaras (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8745, opened 2022-03-21
Add log format for whether origin TLS connection resumed an existing TLS session

### LINKED ISSUE #7850: Should add log format for whether origin TLS connection resumed an existing TLS session
With PR #7479 ATS support session reuse for connections to the origin.  For the client-side request we can log `%<cqssr>` which indicates whether the user-agent connect was started by resuming an existing TLS connection.

We should have a similar entry for the connection from ATS to origin.  Maybe `%<pqssr>`

https://docs.trafficserver.apache.org/en/latest/admin-guide/logging/formatting.en.html?highlight=logging%20format#ssl-encryption

duke8253:
This resolves #7850


SolidWallOfCode (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8753, opened 2022-03-23
Update to proxy-verifier-v2.3.1

bneradt:
Proxy Verifier version 2.3.1 actively closes the connection when the
proxy (ATS) returns a response with a "Connection: close" header. This
improves the stability of the cache-request-method AuTest, wherein ATS,
dependent upon packet timing, would sometimes return a "Connection:
close" and follow it with a TCP FIN. The FIN would then be processed by
the Verifier client after request headers were sent, which would
interrupt the traffic flow and cause the test to fail. With Verifier
version 2.3.1, this is avoided by the client actively closing the
connection when ATS returns the intermittent "Connection: close".


bneradt:
Marking this for 9.2.x because it is a test-only change that improves test stability.

randall (approved):

ezelkow1 (approved):

zwoop:
@bneradt Do we need this for 9.1.x as well ?

zwoop:
Cherry-picked to v9.2.x

bneradt:
> @bneradt Do we need this for 9.1.x as well ?

Yes, that's probably a good idea. That would help the AuTests to run more reliably there.

PR #8750, opened 2022-03-22
Use Sphinx 4.x

bneradt:
Sphinx has fixed the following issue that was blocking us from moving to
Sphinx 4.x:
https://github.com/sphinx-doc/sphinx/issues/9322

With that fixed, we can unpin from 3.x, which this patch does.

Fixes: #7941

PR #8754, opened 2022-03-23
Revert "Use Sphinx 4.x"

bneradt:
Reverts apache/trafficserver#8750

Reverting because this breaks the rendering of our docs. The build succeeds, but the css seems broken. This will need to be investigated further. For now I'll simply revert this for later investigation.


bneradt:
[approve ci fedora]

masaori335 (approved):

PR #8762, opened 2022-03-28
Avoid allocation when matching hosts for vol lookup

cmcfarlen:
This avoids a malloc, free, memcpy and a scan for end of string (for string_view construction).

This call to `CacheHostMatcher::Match` could happen multiple times per request when a hosting.config is used.


bryancall:
[approve ci autest]

SolidWallOfCode (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8765, opened 2022-03-29
Fixing the httpbin AuTests by pinning Werkzeug

bneradt:
The AuTests using httpbin all started failing today when new AuTest
pipenv environments started using the 2.1.0 version of Werkzeug. This
new Werkzeug release removes its BaseResponse API because it has been
deprecated for a few releases. httpbin has a dependency upon
BaseResponse, however, and thus it currently fails with this latest
Werkzeug release. This patch temporarily pins Werkzeug until httpbin is
updated so that our httpbin AuTests can run.

### DIFF
--- a/tests/Pipfile
+++ b/tests/Pipfile
@@ -43,5 +43,11 @@ microserver = ">=1.0.6"
 jsonschema = "*"
 python-jose = "*"
 
+# The latest version of Werkzeug (2.1.0) breaks httpbin because it removes
+# deprecated function `BaseResponse` that httpbin directly or indirectly
+# depends upon. Pinning Wekrzeug for now until httpbin or its dependencies is
+# updated for the changed API.
+Werkzeug = "==2.0.3"
+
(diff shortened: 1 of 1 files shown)


bneradt:
This needs to go to all our supported branches so that our httpbin AuTests can run. I'll make a separate 8.1.x.

randall (approved):

ezelkow1 (approved):

ezelkow1:
[approve ci fedora]

zwoop:
Cherry-picked to v9.2.x

zwoop:
@bneradt Do we need this for the 9.1.x branch too ?

bneradt:
> @bneradt Do we need this for the 9.1.x branch too ?

Yep: we'll need this in 9.1.x since the autests there will use httpbin which requires Werkzeug to be pinned. Otherwise all those httpbin tests fail.

PR #8757, opened 2022-03-24
Use Sphinx 4.x (second attempt)

bneradt:
This is a second attempt to upgrade to Sphinx 4.x. The first was #8750.
It was noticed, however, that the new Sphinx didn't automatically
deliver the theme.css file. This updates the conf.py file to make the
addition of theme.css explicit.

PR #8771, opened 2022-03-29
Revert "Use Sphinx 4.x (second attempt) (#8757)"

bneradt:
This reverts commit 4dbf9147c639afdc2d6e21b98218240563ea91d2.

In the previous attempt to use Sphinx 4.x, I tried to fix our CSS 4.x
rendering by explicitly adding theme.css to the css_files html_context.
That did indeed add the needed theme.css rendering to the various docs,
but there are still some artifacts that don't render correctly (such as
ts:cv parameters). Reverting 4.x for now since 3.x works fine and moving
to 4.x is turning out to be a bigger project than I expected.

### DIFF
--- a/doc/Pipfile
+++ b/doc/Pipfile
@@ -23,7 +23,12 @@ verify_ssl = true
 
 [packages]
 
-sphinx = "*"
+# The latest 4.x sphinx release has issues with style rendering. The 3.x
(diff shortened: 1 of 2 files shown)


bneradt:
[approve ci]

traeak (approved):

PR #8772, opened 2022-03-29
Add metrics for loop detection.

### LINKED ISSUE #7900: Add metric for loop detection
We should have a metric that increments when a proxy loop is detected. Likely here:

https://github.com/apache/trafficserver/blob/f36cf6a6e5b1372916170541bec681a80b34c46f/proxy/http/HttpTransact.cc#L6577

cmcfarlen:
Fixes #7900 

ATS detect loops and multi-hop loops.  I added separate metrics for these, but I can combine them if that is preferred.


ywkaras:
Don't you need to add these to the documention?

ywkaras (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8773, opened 2022-03-29
Pin Jinja2 for doc builds

bneradt:
The latest jinja2 breaks Sphinx 3.x builds. For now, pinning jinja2 to
an earlier release for doc builds.


randall (approved):

bneradt:
Marking this for all our supported branches because pinning jinja2 will be needed for all Sphinx 3.x builds, which is currently what is used for all our supported trains.

PR #8776, opened 2022-03-31
Fix incrementation and logging of server connection attempts.

ywkaras:
For transaction.

### DIFF
--- a/proxy/http/HttpTransact.h
+++ b/proxy/http/HttpTransact.h
@@ -575,8 +575,7 @@ public:
     class Attempts
     {
     public:
-      // Must only be constructed within a CurrenntInfo instance within a State Instance.
-      Attempts(State const *state_i_am_in);
+      Attempts()                 = default;
       Attempts(Attempts const &) = delete;
 
       unsigned
@@ -586,13 +585,13 @@ public:
       }
 
       void
-      maximize()
+      maximize(MgmtInt configured_connect_attempts_max_retries)
       {
-        ink_assert(_v <= configured_connect_attempts_max_retries());
-        if (_v < configured_connect_attempts_max_retries()) {
+        ink_assert(_v <= configured_connect_attempts_max_retries);
+        if (_v < configured_connect_attempts_max_retries) {
           ink_assert(0 == _saved_v);
           _saved_v = _v;
-          _v       = configured_connect_attempts_max_retries();
+          _v       = configured_connect_attempts_max_retries;
         }
       }
 
@@ -604,10 +603,10 @@ public:
       }
 
       void
-      increment()
+      increment(MgmtInt configured_connect_attempts_max_retries)
       {
         ++_v;
-        ink_assert(_v <= configured_connect_attempts_max_retries());
+        ink_assert(_v <= configured_connect_attempts_max_retries);
       }
 
       unsigned
@@ -618,15 +617,13 @@ public:
 
     private:
       unsigned _v{0}, _saved_v{0};
-
-      MgmtInt configured_connect_attempts_max_retries() const;
     };
     Attempts attempts;
     unsigned simple_retry_attempts             = 0;
     unsigned unavailable_server_retry_attempts = 0;
     ParentRetry_t retry_type                   = PARENT_RETRY_NONE;
 
-    _CurrentInfo(State const *state_i_am_in) : attempts(state_i_am_in) {}
+    _CurrentInfo()                     = default;
     _CurrentInfo(_CurrentInfo const &) = delete;
 
   } CurrentInfo;
@@ -829,7 +826,7 @@ public:
     }
 
     // Constructor
-    State() : current(this)
+    State()
     {
       int i;
       char *via_ptr = via_string;
@@ -1092,19 +1089,6 @@ public:
   static void delete_warning_value(HTTPHdr *to_warn, HTTPWarningCode warning_code);
   static bool is_connection_collapse_checks_success(State *s); // YTS Team, yamsat
 };
-
-inline MgmtInt
-HttpTransact::CurrentInfo::Attempts::configured_connect_attempts_max_retries() const
-{
-  return reinterpret_cast<const HttpTransact::State *>(reinterpret_cast<const char *>(this) -
-                                                       offsetof(HttpTransact::State, current.attempts))
-    ->configured_connect_attempts_max_retries();
-}
-
-inline HttpTransact::CurrentInfo::Attempts::Attempts(HttpTransact::State const *state_i_am_in)
-{
-  ink_assert(this == &(state_i_am_in->current.attempts));
-}
 
 typedef void (*TransactEntryFunc_t)(HttpTransact::State *s);
 
--- a/proxy/http/HttpTransact.h
+++ b/proxy/http/HttpTransact.h
@@ -23,6 +23,8 @@
 
 #pragma once
 
+#include <cstddef>
+
 #include "tscore/ink_assert.h"
 #include "tscore/ink_platform.h"
 #include "I_HostDB.h"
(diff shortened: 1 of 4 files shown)


ywkaras:
I'm going to try to get Yahoo Prod to try out this fix.  If it looks good, I'll de-draft this.

ywkaras:
This has soaked for many weeks in Yahoo prod, with no issues.

ywkaras:
[approve ci autest]

proxy/http/HttpTransact.h
```diff
@@ -1034,6 +1090,19 @@ class HttpTransact
   static bool is_connection_collapse_checks_success(State *s); // YTS Team, yamsat
 };
 
+inline MgmtInt
+HttpTransact::CurrentInfo::Attempts::configured_connect_attempts_max_retries() const
+{
+  return reinterpret_cast<const HttpTransact::State *>(reinterpret_cast<const char *>(this) -
+                                                       offsetof(HttpTransact::State, current.attempts))
+    ->configured_connect_attempts_max_retries();
```
cmcfarlen, line 1098: Couldn't you just store the max retries in the Attempts class and not have to do all of this magical stuff?

ywkaras: It's not so simple because connect_attempts_max_retries is overridable, this is a simple way to see any overrides.  And what if we want to fit ATS into 640K so we can port it to MS-DOS?

cmcfarlen: Do you need to call it every update? Could you just call it from the Attempts constructor?

maskit: I'd simply keep `state_i_am_in`, which would have one less arithmetic calculation 😉

ywkaras: > Do you need to call it every update? Could you just call it from the Attempts constructor?

I don't understand, that sounds like the same question rephrased.  How to you pick up the overrides doing that?

ywkaras: > I'd simply keep `state_i_am_in`, that would have ~one~ two less arithmetic calculations wink

Not necessarily faster.  Instruction prefetching is more effective than data prefetching, leading to fewer execution pipeline stalls.

ywkaras: Note that this sort of offsetting is very common in C++.  It's just that the compiler generates it for you.  https://godbolt.org/z/cz115c3br

cmcfarlen: My question is: Do you need to call the override repeatedly within a transaction? Or, another way, will it ever be the case that the return value of the virtual function will be different within a transaction? And if it could be different, would this be desirable/safe?

ywkaras: There's no virtual function involved.  Only two inline functions, small enough so we can be confident they will be inlined in the object code.  The changes you're asking for would probably be a pessimization.  Making an unnecessary, possibly stale copy that may be no faster to access.

cmcfarlen: Since you said "see any overrides" and "How to you pick up the overrides doing that?" I assumed that's what the point of this was.  Storing a value near other, related, values will most likely be faster than accessing that value through two pointer de-references plus some casts and pointer arithmetic, even if you do inline the functions.  If you are saying you want this value to change *during* a transaction, then I'm not sure I agree with that.  Why is that desirable?

ywkaras: I mean config overrides:  https://docs.trafficserver.apache.org/developer-guide/api/functions/TSHttpOverridableConfig.en.html .

ywkaras: In the assembly code for f2() here, https://godbolt.org/z/cz115c3br , you can see that the compiler combines subtracting the offset to b in C2 and adding the offset to k into a single offset.  I don't see any significant way to optimize the two-instruction function f2().  What do you see as the significant difference between this small example, and the use of the same technique in this PR?

cmcfarlen: It seems unnecessary.  My original question wasn't about the perf of this complication (I don't think this is a critical section of code that needs optimizing), but about storing the value vs dereferencing it(where it might change during connect attempts).  I still would like to understand why this value needs to be "overridden" during the course of a transaction.  It seems like the intent of the fix here is to get accurate retry counts in the logs, right?

ywkaras: The value may be changed by a plugin via the code in this file:  https://github.com/apache/trafficserver/blob/46c1a0adf450ac6c27d3eb62ebc55bba0a626b81/src/traffic_server/InkAPI.cc#L8839 .  I think the value may change after the Attempts object is constructed, which makes it tricky to keep a copy in synch.

ywkaras: I have proposed in the past to add a utility to make offsetting like this prettier:
```
#define TS_MEM_TO_CLASS_PTR(CLASS_NAME, MEMBER_SPEC, MEMBER_PTR) \
(reinterpret_cast<CLASS_NAME *>(reinterpret_cast<char *>(MEMBER_PTR) - 
                                offsetof(CLASS_NAME, MEMBER_SPEC)))
```
but it was shot down.

cmcfarlen: Ah, ok. Now it makes sense.  A plugin could change this between the time Attempts is constructed and its first use.  Since the reinterpret_cast offset this pattern is unpopular, perhaps this could be simplified by moving the functions to update and save attempts into State itself and not have the Attempts class at all?  This avoids both nutty pointer arithmetic and storing an extra pointer!

ywkaras: What's nutty about it?  It's crucial to how inheritance works in C++.  There seem to be 1400+ lines of code in TS using reinterpret_cast, which sounds reasonably popular.  This is prod-tested code.  It's going to be a very open-ended process for me to try to rewrite it to conform to some set of aesthetics I don't understand.  And what is the extra pointer?

cmcfarlen: The extra pointer was @maskit 's idea to just store the State pointer in the Attempts class.

ywkaras: OK at least the extra pointer thing is straight-forward.  Like B2 in C3 here:  https://godbolt.org/z/h64jTjE84 .

maskit: My comment about arithmetic was just a joke. I'm not really worried about the speed nor memory footprint on this change.

I'm just uncomfortable with using `offsetof()` for two reasons. 1) It's very C++ and not portable to other object oriented languages. This makes me feel like there are design issues from perspective of OOP and it's trying to cover that by magic. 2) The behavior seems like relatively unstable, because it requires compiler support (according to cppreferrence.com). We've been troubled by compiler and platform differences occasionally.

The second reason is just from my experience so maybe you can resolve my concern, but the first one is a fact that the design doesn't follow regular OOP patterns and principals. Although I'm not going to say we shouldn't use C++ specific stuff, my preference is to not make code too C++ whenever it's possible and reasonable.

At least two people here are not very happy with the proposed change, and two alternative ways are suggested. I don't remember about TS_MEM_TO_CLASS_PTR, but if it's not accepted, I guess the reviewer didn't like the idea either.

I didn't read the entire change so I don't understand why you wanted to add Attempt class, but @cmcfarlen's suggestion also sounds good to me.

cmcfarlen: Sorry @ywkaras, poor choice of words.  I should have said `complicated`.  You have intertwined the Attempts class with CurrentInfo and State so it cannot be reused and you used knowledge of how C++ lays out data to work around the type system to read mutable state in a scope that would otherwise be invisible.  I don't see this as good practice when there are more straight forward methods available.  Here are three options (in order of personal preference):

  
 * Pass the max value in a parameter when calling the maximize and increment functions, leaving resolving the max value to the caller and allowing Attempts to be a standalone class. (unit-testable!)
 * Move the attempts calculations into State where all of the information is in scope and remove the Attempts class.
 * Store a pointer to State in Attempts so you can directly access the data.

ywkaras: This is an actual bug, this isn't just refactoring.  So, seems that there's two realistic alternatives:
1.  I add the superfluous pointer.
2. Chris writes it following the aesthetics that I don't understand.
Which is preferred?

proxy/http/HttpTransact.h
```diff
@@ -570,12 +572,60 @@ class HttpTransact
...
+      unsigned
+      get() const
+      {
+        return _v;
+      }
+
+      void
+      maximize()
+      {
+        ink_assert(0 == _saved_v);
```
bneradt, line 591: I believe this assertion, while generally silenced for release builds, is failing.

bneradt: After more investigation, I see that this assertion is only failing when this patch is applied on top of our HTTP/2 to origin branch.

proxy/http/HttpTransact.h
ywkaras: OK Chris, uncle, you got your first choice.

cmcfarlen (approved):
lgtm

vmamidi (approved):
Approving for Chris McFarlen

PR #8784, opened 2022-04-08
Propagate proxy.config.net.sock_option_flag_in to newly accepted connections

randall:
* Propagate proxy.config.net.sock_option_flag_in to newly accepted connections

* For client connections, disable setting SO_LINGER by default

This was originally approved under #8463 and was reverted in #8622

### DIFF
--- a/doc/admin-guide/files/records.config.en.rst
+++ b/doc/admin-guide/files/records.config.en.rst
@@ -4647,7 +4647,7 @@ Sockets
 
    Sets the receive buffer size for connections from the client to |TS|.
 
-.. ts:cv:: CONFIG proxy.config.net.sock_option_flag_in INT 0x5
+.. ts:cv:: CONFIG proxy.config.net.sock_option_flag_in INT 0x1
 
    Turns different options "on" for the socket handling client connections:::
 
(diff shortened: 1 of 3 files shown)


randall:
This is known to fix`set-conn-dscp` which broke sometime in 9.1+

ezelkow1 (approved):

pbchou:
@randall -- I am not sure that I understand the history behind this PR being merged, reverted, and now merged again. I believe that we need this for setting a custom TOS towards the client in our use case. Is this safe to back-port to 9.1.2?

zwoop:
Cherry-picked to v9.2.x

randall:
@pbThe reason for reverting was because there was the possibility the `master` branch could become 9.3. So @zwoop had asked that anything that was backwards incompatible (like this change) not to be merged into master. But it's been more or less decided that 9.2.x will be the last of the 9.x and now we're free to break compatibility again on master (which will become 10.x at some point).

It's probably safe to back port to 9.1, but again, this change backwards incompatible (see the long thread in slack in the last week or 2) and that's for the release manager to decide if they want to break things for a patch release that has already had 2 patches.

PR #8787, opened 2022-04-11
Move error logging to use diags.log instead of error.log

bryancall:
Consistently using Error() instead of sometimes using Log:error()

### DIFF
--- a/proxy/http/HttpSM.cc
+++ b/proxy/http/HttpSM.cc
@@ -1277,7 +1277,7 @@ HttpSM::state_request_wait_for_transform_read(int event, void *data)
       //  sending a request message body.  Change the event
       //  to an error and fall through
       event = VC_EVENT_ERROR;
-      Log::error("Request transformation failed to set content length");
+      Error("Request transformation failed to set content length");
     }
   // FALLTHROUGH
   default:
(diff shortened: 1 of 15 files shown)


proxy/http/HttpSM.cc
```diff
@@ -4175,28 +4175,28 @@ HttpSM::check_sni_host()
...
-                                 .extend(1)
-                                 .write('\0')
-                                 .data());
+              Error("%s", lbw()
+                            .clip(1)
+                            .print("No SNI for TLS request: connecting to {} for host='{}', returning a 403",
+                                   t_state.client_info.dst_addr, std::string_view{host_name, static_cast<size_t>(host_len)})
+                            .extend(1)
+                            .write('\0')
+                            .data());
```
bneradt, line 4184: Maybe keep these connection-oriented messages in error.log? That's what I expect in error.log anyway: connection issues.

bneradt (changes requested):
Thanks. I personally feel that error.log should just have connection (TCP or TLS) oriented issues. That's what I look to error.log for. Thus removing the "failed to log transaction..." message from error.log makes sense to me. But should the other connection related issues stay in error.log?

bneradt:
[approve ci fedora]

bryancall:
[approve ci autest]

maskit:
I think we should mark this as incompatible and land this on 10-Dev branch first. Some people may watch error.log for error logs but not diags.log.

PR #3609, opened 2018-05-08
Remove traffic_cop

shinrich:
Getting rid of traffic_cop as per the decision of the ATS Euro Tour 2018.
Removed traffic cop, related config settings, state machine tests, and documentation.

Should be able to remove the health checkports too out of traffic_server and trafffic_manager.

PR #8793, opened 2022-04-13
Removes remaining vestiges to the backdoor port

elsloo:
* Majority of code was removed in #4866

* `--proxyBackDoor` argument removed from `traffic_manager` in #3609

### DIFF
--- a/doc/appendices/command-line/traffic_manager.en.rst
+++ b/doc/appendices/command-line/traffic_manager.en.rst
@@ -34,7 +34,6 @@ Description
 .. option:: --help
 .. option:: --nosyslog
 .. option:: --path FILE
-.. option:: --proxyBackDoor PORT
 .. option:: --proxyOff
 .. option:: --listenOff
 .. option:: --proxyPort PORT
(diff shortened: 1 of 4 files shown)


elsloo:
I'm actually +1 on adding the backdoor back into ATS because having an avenue for tools to avoid the throttle in high load scenarios is advantageous for a number of reasons. However, these remaining vestiges to the feature creates an inconsistency that can be misleading for those reading the code and docs.

A future PR may restore this functionality, but because the code was removed in the 8.0.x timeframe, we may need further changes to restore the prior feature.

bneradt (approved):

elsloo:
We briefly discussed the desired mechanism to leverage for a reimplementation of the backdoor in the future during our weekly bug scrub today. We agreed that we would ideally utilize the existing paradigm of tagging ports with a desired property in `proxy.config.http.server_ports` (docs reference [here](https://docs.trafficserver.apache.org/admin-guide/files/records.config.en.html#proxy-config-http-server-ports)). If we added a new keyword, for example `backdoor`, to the list of options, we will clearly know which port is used as a backdoor (if any), and the implementation will in theory be simpler than what we had previously due to not being tied to `traffic_manager`.

zwoop:
Cherry-picked to v9.2.x

PR #8798, opened 2022-04-18
Create clean exit path for traffic_manager on SIGTERM

jeredfloyd:
Since the removal of traffic_cop, it appears that there is no way to cause traffic_manager to exit cleanly.  SIGTERM causes the cleanup of child processes and a graceful shutdown, however still exits with code 15.  We can ignore this in the systemd unit file, but it's cleaner to have some path to a clean exit of traffic_manager.

### DIFF
--- a/src/traffic_manager/traffic_manager.cc
+++ b/src/traffic_manager/traffic_manager.cc
@@ -982,6 +982,8 @@ SignalHandler(int sig)
   case SIGXCPU:
   case SIGXFSZ:
     abort();
+  case SIGTERM:
+    ::exit(0);
   default:
     fprintf(stderr, "[TrafficManager] ==> signal #%d\n", sig);
     mgmt_log("[TrafficManager] ==> signal #%d\n", sig);


jeredfloyd:
It might also be desirable to add a "--manager" option to `traffic_ctl server stop`, to be parallel with `restart`?  I'm not entirely clear on the intended operation, however both the init.d file and the systemd unit use signals today, so a clean way to exit traffic_manager via signal is always valuable.

randall:
[approve ci fedora]

bryancall (approved):

bryancall:
@jeredfloyd We do plan to get rid of traffic_manager in the next major release (10.0.0).

bneradt:
[approve ci fedora]

jeredfloyd:
@bryancall Thanks! Once I have 9.x in Fedora I will try to follow future plans a little more closely.

zwoop:
Cherry-picked to v9.2.x

PR #8809, opened 2022-04-27
Change HttpSM.cc from using Debug() to new interface, Dbg().

ywkaras:


proxy/http/HttpSM.cc
```diff
@@ -260,7 +260,104 @@ HttpVCTable::cleanup_all()
   }
 }
 
-#define SMDebug(tag, fmt, ...) SpecificDebug(debug_on, tag, "[%" PRId64 "] " fmt, sm_id, ##__VA_ARGS__)
+#if defined(Dbg) // Optimized debug available.
```
ywkaras, line 263: Use of conditional compilation here means that we could backport this PR to a 9.x release without having to also backport https://github.com/apache/trafficserver/pull/8581 .  Could be useful if we need to backport later PRs that also change HttpSM.cc extensively.

proxy/http/HttpSM.cc
```diff
@@ -260,7 +260,104 @@ HttpVCTable::cleanup_all()
...
+
+#define is_debug_tag_http_set() is_dbg_ctl_enabled(dbgCtlHttp)
+#define is_debug_tag_http_tproxy_set() is_dbg_ctl_enabled(dbgCtlHttpTproxy)
+#define is_debug_tag_http_timeout_set() is_dbg_ctl_enabled(dbgCtlHttpTimeout)
+#define is_debug_tag_dns_set() is_dbg_ctl_enabled(dbgCtlDns)
+#define is_debug_tag_ip_allow_set() is_dbg_ctl_enabled(dbgCtlIpAllow)
+
+#else
+
+#define DebugHttpRange(fmt, ...) Debug("htto_range", fmt, ##__VA_ARGS__)
```
maskit, line 327: typo: htto

proxy/http/HttpSM.cc
```diff
@@ -273,7 +370,7 @@ HttpVCTable::cleanup_all()
 #define STATE_ENTER(state_name, event)                                                 \
   {                                                                                    \
     /*ink_assert (magic == HTTP_SM_MAGIC_ALIVE); */ REMEMBER(event, reentrancy_count); \
-    SMDebug("http", "[%s, %s]", #state_name, HttpDebugNames::get_event_name(event));   \
+    SMDebugHttp("[%s, %s]", #state_name, HttpDebugNames::get_event_name(event));       \
```
maskit, line 373: It's sad that I can't see the actual tag name, although that might have already been discussed when the new interface was proposed. This one is ok, but "ip-allow" is tricky because it uses "-" where most other tags use "_".

proxy/http/HttpSM.cc
ywkaras: thanks

proxy/http/HttpSM.cc
ywkaras: @SolidWallOfCode did you pick the tag name "ip-allow"?  Do you care if I change it to "ip_allow"?

I could also change to a naming pattern where, for example, SMDebug_http_trans (rather than SMDebugHttpTrans) corresponds to the tag "http_trans".

ywkaras: If all debug tags were valid C++ identifiers, then I wouldn't need the SMDebugXxx() macros.  This:
```
SMDebug("http_trans", "Unable to accelerate range request, fallback to transform");
```
would instead become this:
```
SMDebug(http_trans, "Unable to accelerate range request, fallback to transform");
```
The implementation would use either the concatenation or stringization macro operators:
```
#if defined(Dbg)

namespace
{
DbgCtl dbgCtl_http_trans;
// ...
}

#define SMDebug(tag, fmt, ...) SpecificDbg(debug_on, dbgCtl_ # tag, "[%" PRId64 "] " fmt, sm_id, ##__VA_ARGS__)

#else

#define SMDebug(tag, fmt, ...) SpecificDbg(debug_on, #tag, "[%" PRId64 "] " fmt, sm_id, ##__VA_ARGS__)

#endif
```

ywkaras:
[approve ci ubuntu]

ywkaras:
[approve ci fedora]

proxy/http/HttpSM.cc
maskit: > If all debug tags were valid C++ identifiers...

I like the concatenation/stringization macro idea. It would be confusing a little since old Debug() requires quotations though, I prefer it rather than other naming patterns.

Also, somebody told me that he prefer to have "Debug" at the end of macro names because that enables you to find (grep) Debug lines easily, when I added macros for HTTP/2 or 3.

ywkaras:
I marked this for backport to 9.2, since it seems likely there will be later changes to HttpSM.cc that will have to be backported to 9.2.

maskit:
If we change "ip-allow" to "ip_allow", we should change ones in IPAllow.cc, HttpTransact.cc, and ip_allow.test.py as well.

> I marked this for backport to 9.2, since it seems likely there will be later changes to HttpSM.cc that will have to be backported to 9.2.

I understand that would make backporting other changes easy but isn't the change for "ip-allow" technically an incompatible change? I personally don't mind changing debug tag names on a minor release though.

In either case, we should probably make the change for "ip-allow" separately on another PR for change logs and release notes. We can discuss about the incompatibility on the PR, and that would tell whether we can backport this to 9.x.

ywkaras:
https://github.com/apache/trafficserver/pull/8830 should be merged prior to merging this PR.

ywkaras:
The Au test run finished successfully, even though it is reported above as "pending":  https://ci.trafficserver.apache.org/job/Github_Builds/job/autest/968/consoleFull
(See the line 'Commit message: "Merge commit '654417e6ab6fce629d50976c71dac90a32a4b348' into HEAD"'.

bneradt:
[approve ci autest]

maskit:
I think we should land this on 10-Dev branch if I understand why the branch exists correctly.

maskit:
It isn't limited to large features. Incompatible changes should not be merged on master.

I thought we agreed that the ip_allow change is incompatible because you unmarked 9.2.

ywkaras:
There would probably be many merge conflicts in HttpSM.cc every time we merged master to 10-Dev.

In any case, more urgent work has come up, so I must put this aside for now.

ywkaras:
The practical way to do this is to merge it on master right after release 9.2 comes out (and make sure it's the first commit in any hypothetical 9.3).

maskit:
Strictly speaking, this change itself is not incompatible, but #8830 is, IMO. The problem (I think) is that #8830 is already on master. #8830 should be reverted on master and the change should be made on 10-Dev first, if we agree on the change is incompatible.

PR #8814, opened 2022-04-29
9.1.x: Unit test compiler and library run time fixes

bneradt:
This fixes a signed type warning and a library runtime warning in the unit
(Catch) tests.


bneradt:
The CI here will require the Catch tests to be updated in 9.1.x before they will pass:

* #8231
* #8683

bneradt:
Oh, the autests on 9.1.x are failing because we need this as well in 9.1.x to fix httpbin:

#8765

randall (approved):

PR #8815, opened 2022-05-02
Fedora/RHEL path layout for config.layout

jeredfloyd:
Fedora/RHEL prefer files in slightly different places, and there aren't configure options to override everything in config.layout, so a new layout entry makes sense.


bryancall:
I will review it

bryancall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8488, opened 2021-11-02
cache_range_requests plugin: hard set reason string for handled 206 response

traeak:
The cache_range_requests plugin expects the parent to have the reason set to "Partial Content" (case insensiitve) in addition to the expecting a 206 status.

This PR will hard set the Reason to "Partial Content" if not already set.

closes #8481

PR #8816, opened 2022-05-02
Add support for caching complete responses to the cache range requests plugin

elsloo:
* Adds support for caching full object responses.

* Refactored logic to be more efficient.

* Added new plugin parameter to relevant docs.

### DIFF
--- a/doc/admin-guide/plugins/cache_range_requests.en.rst
+++ b/doc/admin-guide/plugins/cache_range_requests.en.rst
@@ -185,6 +185,15 @@ status code is reset back to 206, which leads to the object not being cached.
 
 This option is useful when used with other plugins, such as Cache Promote.
 
+Cache Complete Responses
+------------------------
+
+.. option:: --cache-complete-responses
+.. option:: -r
+
+This option causes the plugin to cache complete responses (200 OK). By default,
+only 206 Partial Content responses are cached by this plugin; without this flag,
+any 200 OK observed will be marked as not cacheable.
 
 Configuration examples
 ======================
--- a/plugins/cache_range_requests/README.md
+++ b/plugins/cache_range_requests/README.md
@@ -79,12 +79,13 @@ X-CRR-IMS header support
     Consider using the header_rewrite plugin to protect the parent
 		from using this option as an attack vector against an origin.
 
-Object Cacheability:
+Object Cacheability
+
     Normally objects are forced into the cache by changing the status code in the
     response from the upstream host from 206 to 200. The default behavior is to
     perform this operation blindly without checking cacheability. Add the `-v`
     flag to cause the plugin to ensure the object is cacheable; when it is not,
-    the 206 status code is restored and the object will not be cached.
+    the status code is not changed to 200 and the object will not be cached.
 
     Global Plugin (plugin.config):
 
@@ -95,3 +96,19 @@ Object Cacheability:
 
       <from-url> <to-url> @plugin=cache_range_requests.so @pparam=--verify-cacheability
       <from-url> <to-url> @plugin=cache_range_requests.so @pparam=-v
+
+Caching Complete Responses
+
+    To enable caching of complete responses, that is, a 200 OK instead of a 206 Partial
+    Content response, add the `-r` flag to the plugin parameters. By default, complete
+    responses are marked as uncacheable.
+
+    Global Plugin (plugin.config):
+
+      cache_range_requests.so --cache-complete-responses
+      cache_range_requests.so -r
+
+    Remap Plugin (remap.config):
+
+      <from-url> <to-url> @plugin=cache_range_requests.so @pparam=--cache-complete-responses
+      <from-url> <to-url> @plugin=cache_range_requests.so @pparam=-r
--- a/plugins/cache_range_requests/cache_range_requests.cc
+++ b/plugins/cache_range_requests/cache_range_requests.cc
@@ -54,6 +54,7 @@ struct pluginconfig {
   bool consider_ims_header{false};
   bool modify_cache_key{true};
   bool verify_cacheability{false};
+  bool cache_complete_responses{false};
   std::string ims_header;
 };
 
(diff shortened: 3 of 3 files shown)


elsloo:
[approve ci fedora]

elsloo:
[approve ci rocky]

elsloo:
[approve ci autest]

elsloo:
[approve ci clang-analyzer]

traeak:
When running a test on this first time around I got a 200 response with a modified cache key. 

```
< HTTP/1.1 200 OK
< Accept-Ranges: bytes
< Cache-Control: max-age=123
< Content-Length: 5242880
< Content-Type: text/plain
< Last-Modified: Thu, 05 May 2022 12:36:40 GMT
< Date: Thu, 05 May 2022 12:38:43 GMT
< Age: 0
< Proxy-Connection: keep-alive
< Via: http/1.1 ipcdn-cache-58.cdnlab.comcast.net (ApacheTrafficServer/10.0.0 [uScMsSfWpSeN:t cCMpSs ])
< Server: ATS/10.0.0
< X-Cache-Key: http://localhost/~p.txt/~s.5M/~ui.200/a-bytes=0-100
```

Second request I got back the whole asset but with a 206 response:

```
< HTTP/1.1 206 OK
< Accept-Ranges: bytes
< Cache-Control: max-age=123
< Content-Length: 5242880
< Content-Type: text/plain
< Last-Modified: Thu, 05 May 2022 12:36:40 GMT
< Date: Thu, 05 May 2022 12:38:43 GMT
< Age: 59
< Proxy-Connection: keep-alive
< Via: http/1.1 ipcdn-cache-58.cdnlab.comcast.net (ApacheTrafficServer/10.0.0 [uScHs f p eN:t cCHp s ])
< Server: ATS/10.0.0
< X-Cache-Key: http://localhost/~p.txt/~s.5M/~ui.200/a-bytes=0-100
```

elsloo:
> A 200 is being cached with a cachekey that's been modified by the CRR plugin. Any origin request that returns 200 responses for a range request could be exploited to overload the cache with multiple full asset. **Docs should have warning signs all over this if the CRR plugin's endpoint is exposed.**

When run as a standalone plugin, and when allowing the cache range requests plugin to modify the cache key, this statement is true. However, when this modification is used with the cache key plugin, the slice plugin, and with the cache key manipulation within cache range requests plugin disabled, the statement is no longer true. In my testing, the cache key becomes anchored to the slice reference block, and consequently, that is the only cache key used for the object.

I understand your concern, but in my opinion, having this plugin directly exposed to the open internet without the slice plugin or a normalization tier is already a huge concern because of the damage a user can do by polluting the cache with random ranges on any object. That is basically the same "exploit," just with fewer bytes per cache key. Nevertheless, I believe the possibility still exists with the default cache range requests plugin, regardless of whether this new plugin parameter is used. The only actual difference is the rate at which you can inflict damage.

I can add a comment to recommend only using this when using this style of configuration if that will help reduce the risk of accidental use.

> But there's also a bug uncovered in testing:
> 
> Here's the request I was using. Request strips the range request header at the synthetic origin:
> 
> ```
> curl -x localhost:18080 "http://crr/~p.txt/~s.5M/~ui.200/a" -r 0-100 -H "X-Dtp: ~rmhdrs.Range"  -H "x-debug: x-cache-key" -Lv
> ```
> 
> When running a test on this first time around I got a 200 response with a modified cache key.
> 
> ```
> < HTTP/1.1 200 OK
> < Accept-Ranges: bytes
> < Cache-Control: max-age=123
> < Content-Length: 5242880
> < Content-Type: text/plain
> < Last-Modified: Thu, 05 May 2022 12:36:40 GMT
> < Date: Thu, 05 May 2022 12:38:43 GMT
> < Age: 0
> < Proxy-Connection: keep-alive
> < Via: http/1.1 ipcdn-cache-58.cdnlab.comcast.net (ApacheTrafficServer/10.0.0 [uScMsSfWpSeN:t cCMpSs ])
> < Server: ATS/10.0.0
> < X-Cache-Key: http://localhost/~p.txt/~s.5M/~ui.200/a-bytes=0-100
> ```
> 
> Second request I got back the whole asset but with a 206 response:
> 
> ```
> < HTTP/1.1 206 OK
> < Accept-Ranges: bytes
> < Cache-Control: max-age=123
> < Content-Length: 5242880
> < Content-Type: text/plain
> < Last-Modified: Thu, 05 May 2022 12:36:40 GMT
> < Date: Thu, 05 May 2022 12:38:43 GMT
> < Age: 59
> < Proxy-Connection: keep-alive
> < Via: http/1.1 ipcdn-cache-58.cdnlab.comcast.net (ApacheTrafficServer/10.0.0 [uScHs f p eN:t cCHp s ])
> < Server: ATS/10.0.0
> < X-Cache-Key: http://localhost/~p.txt/~s.5M/~ui.200/a-bytes=0-100
> ```

This should be relatively easy to fix, but I have not observed this issue due to not running this as a standalone plugin. In my case, I was getting 200s consistently for any object that was smaller than the slice block size.

elsloo:
@traeak I was able to reproduce this issue and in the process discovered that the content revalidation case is not being handled correctly, in addition to what you observed. This is due to changes introduced in PR #8488 and instead of pushing the fix into this PR, I wanted to separate the fix from this feature and instead opened #8828.

Please retest, possibly with both PRs. If necessary I can push the fix into this PR to ease the testing burden.

traeak (dismissed):
Changes look good and test fine.

traeak:
If possible adding an autest would be nice.

elsloo:
> If possible adding an autest would be nice.

I will add an AuTest and update the docs to address your initial feedback about the risk before merging this.

elsloo:
@traeak I just pushed two commits that add an AuTest and updates the docs with a warning and detailed description of the intended use case for this feature. Please take a look at both additions and provide any necessary feedback.

traeak (approved):
Looks and tests fine.

zwoop:
Cherry-picked to v9.2.x

PR #8818, opened 2022-05-04
Expose setting some HTTP/2 tunables via sni.yaml

randall:
proxy.config.http2.default_buffer_water_mark can now be set on a
per-domain basis.

### DIFF
--- a/iocore/net/I_NetVConnection.h
+++ b/iocore/net/I_NetVConnection.h
@@ -228,8 +228,6 @@ struct NetVCOptions {
 
   bool tls_upstream = false;
 
-  /// Reset all values to defaults.
-
   /**
    * Set to DISABLED, PERFMISSIVE, or ENFORCED
    * Controls how the server certificate verification is handled
@@ -241,6 +239,8 @@ struct NetVCOptions {
    * Currently SIGNATURE and NAME
    */
   YamlSNIConfig::Property verifyServerProperties = YamlSNIConfig::Property::NONE;
+
+  /// Reset all values to defaults.
   void reset();
 
   void set_sock_param(int _recv_bufsize, int _send_bufsize, unsigned long _opt_flags, unsigned long _packet_mark = 0,
--- a/iocore/net/P_SNIActionPerformer.h
+++ b/iocore/net/P_SNIActionPerformer.h
@@ -98,6 +98,26 @@ private:
   bool enable_h2 = false;
 };
 
+class HTTP2BufferWaterMark : public ActionItem
+{
+public:
+  HTTP2BufferWaterMark(int value) : value(value) {}
+  ~HTTP2BufferWaterMark() override {}
+
+  int
+  SNIAction(TLSSNISupport *snis, const Context &ctx) const override
+  {
+    auto ssl_vc = dynamic_cast<SSLNetVConnection *>(snis);
+    if (ssl_vc) {
+      ssl_vc->hints_from_sni.http2_buffer_water_mark = value;
+    }
+    return SSL_TLSEXT_ERR_OK;
+  }
+
+private:
+  int value = -1;
+};
+
 class TunnelDestination : public ActionItem
 {
 public:
--- a/proxy/http2/Http2ClientSession.cc
+++ b/proxy/http2/Http2ClientSession.cc
@@ -25,6 +25,7 @@
 #include "HttpDebugNames.h"
 #include "tscore/ink_base64.h"
 #include "Http2CommonSessionInternal.h"
+#include "TLSSNISupport.h"
 #include "TLSEarlyDataSupport.h"
 
 ClassAllocator<Http2ClientSession, true> http2ClientSessionAllocator("http2ClientSessionAllocator");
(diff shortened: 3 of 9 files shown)


iocore/net/I_NetVConnection.h
```diff
@@ -187,6 +187,8 @@ struct NetVCOptions {
   uint32_t packet_tos;
   uint32_t packet_notsent_lowat;
 
+  uint32_t http2_buffer_water_mark;
```
maskit, line 190: I'm ok with exposing a setting tunable via sni.yaml and I'm not going to block this PR because we can change the internal implementation later without breaking compatibility, but I don't like to have this option here because it's for a specific application protocol (not for generic connections).

I'd put it like:

```
class TLSSNISupport {
  ...
  struct HintsFromSNI { // NEW
    uint32_t http2_buffer_water_mark;
  } hints_from_sni;
};
```

It'd still have an application specific setting in one of TLS stuff, but it's better than contaminating NetVCOptions, IMO. And it also suggests the value is just a hint, and you can ignore it if appropriate (remember connection coalescing?).

iocore/net/P_SNIActionPerformer.h
```diff
@@ -98,6 +98,30 @@ class ControlH2 : public ActionItem
...
+class HTTP2BufferWaterMark : public ActionItem
+{
+public:
+  HTTP2BufferWaterMark(int value) : value(value) {}
+  ~HTTP2BufferWaterMark() override {}
+
+  int
+  SNIAction(TLSSNISupport *snis, const Context &ctx) const override
+  {
+    if (!value) {
```
bneradt, line 110: For non-bools, I suggest explicit compares:

```
if (value == 0) {
```

proxy/http2/Http2ClientSession.cc
```diff
@@ -116,9 +116,16 @@ Http2ClientSession::new_connection(NetVConnection *new_vc, MIOBuffer *iobuf, IOB
...
 
   // This block size is the buffer size that we pass to SSLWriteBuffer
-  auto buffer_block_size_index   = iobuffer_size_to_index(Http2::write_buffer_block_size, MAX_BUFFER_SIZE_INDEX);
-  this->write_buffer             = new_MIOBuffer(buffer_block_size_index);
-  this->write_buffer->water_mark = Http2::buffer_water_mark;
+  auto buffer_block_size_index = iobuffer_size_to_index(Http2::write_buffer_block_size, MAX_BUFFER_SIZE_INDEX);
+  this->write_buffer           = new_MIOBuffer(buffer_block_size_index);
+
+  uint32_t buffer_water_mark;
+  if (this->_vc->options.http2_buffer_water_mark != 0) {
```
bneradt, line 123: Can you help me think through this: is this OK if the value is the default of `-1`? Or is `_vc->options.http2_buffer_water_mark` guaranteed to not be `-1` here? I might be missing something...

bneradt (commented):
Looks fine in general. I need some help thinking through the logic and default values in `proxy/http2/Http2ClientSession.cc`.

iocore/net/I_NetVConnection.h
randall: That sounds reasonable. 

And yes, I remember connection coalescing. I work for a CDN that tends to keep SANs for certs separated by SNI. ;) How this works today, is perfect for our use case. But I understand others probably do not work in the same way.

proxy/http2/Http2ClientSession.cc
randall: That's a good callout. I might've posted this PR in mid transition from the PR this required.

The desire is to only apply the change if the override(via `sni.yaml`)has been explicitly been set. Or if it's not explicitly set , to use the default from `records.config`. 

Maybe I should use `std::optional` here as well instead of trying to track/remember sentinel values.

iocore/net/I_NetVConnection.h
```diff
@@ -187,6 +188,8 @@ struct NetVCOptions {
   uint32_t packet_tos;
   uint32_t packet_notsent_lowat;
 
+  TLSSNISupport::HintsFromSNI hintsFromSNI;
```
maskit, line 191: Do you still need this? Since you added `hints_from_sni` to `SNISupport`, you should be able to access the member like `vc->hints_from_sni`.

randall: ah. Now I see the relationship here. Corrected this.

proxy/http2/Http2ClientSession.cc
```diff
@@ -116,9 +117,17 @@ Http2ClientSession::new_connection(NetVConnection *new_vc, MIOBuffer *iobuf, IOB
...
 
   // This block size is the buffer size that we pass to SSLWriteBuffer
-  auto buffer_block_size_index   = iobuffer_size_to_index(Http2::write_buffer_block_size, MAX_BUFFER_SIZE_INDEX);
-  this->write_buffer             = new_MIOBuffer(buffer_block_size_index);
-  this->write_buffer->water_mark = Http2::buffer_water_mark;
+  auto buffer_block_size_index = iobuffer_size_to_index(Http2::write_buffer_block_size, MAX_BUFFER_SIZE_INDEX);
+  this->write_buffer           = new_MIOBuffer(buffer_block_size_index);
+
+  uint32_t buffer_water_mark;
+  SSLNetVConnection *ssl_vc = dynamic_cast<SSLNetVConnection *>(this->_vc);
```
maskit, line 124: Sorry for bugging you on this, but the point of TLSSomethingSupport is to remove dependencies for SSLNetVC. Including P_SSLNetVConnection.h brings a lot of dependencies unnecessarily (I haven't completely removed the dependency so there's no difference at the moment though).

You could include "TLSSNISupport.h" instead, and access the member like:
```
TLSSNISupport *snis = dynamic_cast<TLSSNISupport *>(this->_vc);
if (snis && snis->hints_from_sni....
```

randall: I think I finally understand what you're getting at here. Take another look, please. 

I'm sorry for being slow here. I thought it was weird there was no other casts to SSLNetVC.

maskit:
I approved this PR, but someone else need to approve it to merge.

bneradt (approved):

randall:
For back porting to 9.2.x: https://github.com/apache/trafficserver/pull/8875

PR #8820, opened 2022-05-04
Fixes leak in SNIAction name globbing

randall:
pcre_compile allocated object is never pcre_free-ed


randall:
This goes back to at least ATS 9.

This probably also effects whatever is using `NextHopItem` as it is derived from `namedElement`

randall:
[approve ci autest]

randall:
[approve ci autest]

brbzull0:
[approve ci autest]

PR #8822, opened 2022-05-05
Fixes leak of ssl_ocsp_response_path_only on reload

randall:


randall:
This may be occurring in 9.0.x as well.

zwoop:
Cherry-picked to v9.2.x

PR #8828, opened 2022-05-05
Fixes issues with the CRR plugin introduced in #8488

elsloo:
* Fixes an issue that leads to an incorrect assumption about the origin status code on cache hit

* Fixes the content revalidation case, as original implementation did not recognize the 304


elsloo:
[approve ci autest]

bryancall:
#8488

PR #8829, opened 2022-05-06
Fix a compile error

maskit:
The error was introduced by #8716. Probably need to be backported to 9.2.x.

```
Making all in cache
  CXX      CacheWrite.o
CacheWrite.cc:293:95: error: format specifies type 'long' but the argument has type 'long long' [-Werror,-Wformat]
    DDebug("cache_evac", "dir_evac_bucket out of bounds, skipping evacuate: %ld(%d), %d, %d", bucket, evacuate_size,
                                                                            ~~~               ^~~~~~
                                                                            %lld
```


bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8830, opened 2022-05-06
Change ip-allow debug tag to ip_allow.

ywkaras:


ywkaras:
This change is to prepare for https://github.com/apache/trafficserver/pull/8830 , which required debug tags in HttpSM to be valid suffixes of C/C++ identifiers.

maskit:
I'm fine with making this change on 10.0, but have a little concern on backporting it to 9.x. Although it's just a debug tag name change, I think it's technically an incompatible change. I personally don't mind, but thought it's better to make a consensus.

I found  several other tag names that don't follow the new rule. I guess those have to be changed as well when we use the new debug interface. If we are going to change ip-allow we should probably change the others.

Here's the incomplete list:
- log-file
- log-filter-compare
- log-agg
- log-logbuffer
- log-config
- log-config-transfer
- log-api-mutex
- Trie::Insert

ywkaras:
We would have to backport this if we want to backport #8830 .  The question we must consider is if there will be further changes to HttpSM.cc that will be backported to 9.x .  I'm worried about difficult merge conflicts, if we must backport changes to HttpSM.cc that follow #8830 , if #8830 is not backported.

SolidWallOfCode (approved):

PR #8831, opened 2022-05-06
Fix parent_select optional scheme

rob05c:
Fixes core strategies and parent_select plugin scheme to be optional, and adds a test.

Backstory: some time ago I noticed the strategies.yaml scheme was optional, and a strategy without a scheme would match a host without a scheme. This behavior seemed to be incidental, but it made it much easier for me to generate config, it would have been difficult to get the scheme in the parent generation, all that logic lived in a separate area of our code for remap generation. And the "scheme" isn't actually used by the strategy for anything except matching hosts; the actual scheme used for upstream requests is always the remap target scheme.

So I documented the behavior:
https://docs.trafficserver.apache.org/en/latest/admin-guide/files/strategies.yaml.en.html
> scheme: Indicates which scheme the strategy supports, http or https. Note this is only used to match hosts to strategies; the actual scheme used in the upstream request will be the scheme of the remap.config target, regardless of this value. The scheme is optional, and strategies without a scheme will match hosts without a scheme. This allows for omitting the scheme if hosts are not shared between strategies, or if all strategies using a given host use the same scheme.

This was broken recently, and changed to reject strategies.yaml files without schemes (as part of a larger validation PR). Which broke all our generated strategies.yaml files.

This PR fixes it to match the docs, and adds an autest to ensure it doesn't break again, and also comments in the code to make the intent clear.

### DIFF
--- a/plugins/experimental/parent_select/strategy.cc
+++ b/plugins/experimental/parent_select/strategy.cc
@@ -71,6 +71,7 @@ PLNextHopSelectionStrategy::Init(const YAML::Node &n)
   bool self_host_used = false;
 
   try {
+    // scheme is optional, and strategies with no scheme will match hosts with no scheme
     if (n["scheme"]) {
       auto scheme_val = n["scheme"].Scalar();
       if (scheme_val == "http") {
@@ -78,9 +79,7 @@ PLNextHopSelectionStrategy::Init(const YAML::Node &n)
       } else if (scheme_val == "https") {
         scheme = PL_NH_SCHEME_HTTPS;
       } else {
-        scheme = PL_NH_SCHEME_NONE;
-        PL_NH_Note("Invalid 'scheme' value, '%s', for the strategy named '%s', setting to PL_NH_SCHEME_NONE", scheme_val.c_str(),
-                   strategy_name.c_str());
+        PL_NH_Note("Invalid scheme '%s' for strategy '%s', setting to NONE", scheme_val.c_str(), strategy_name.c_str());
       }
     }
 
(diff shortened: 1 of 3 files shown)


plugins/experimental/parent_select/strategy.cc
```diff
@@ -71,16 +71,15 @@ PLNextHopSelectionStrategy::Init(const YAML::Node &n)
...
       auto scheme_val = n["scheme"].Scalar();
       if (scheme_val == "http") {
         scheme = PL_NH_SCHEME_HTTP;
       } else if (scheme_val == "https") {
         scheme = PL_NH_SCHEME_HTTPS;
       } else {
-        scheme = PL_NH_SCHEME_NONE;
-        PL_NH_Note("Invalid 'scheme' value, '%s', for the strategy named '%s', setting to PL_NH_SCHEME_NONE", scheme_val.c_str(),
-                   strategy_name.c_str());
+        PL_NH_Note("Invalid scheme '%s' for strategy '%s', setting to NONE", scheme_val.c_str(), strategy_name.c_str());
```
ywkaras, line 82: Since this isn't a constructor, seem slightly fragile to assume data member contain values set at construction.  Hopefully no one ever assumes this function can re-init as well as init.

plugins/experimental/parent_select/strategy.cc
```diff
@@ -395,16 +394,16 @@ template <> struct convert<PLNHProtocol> {
...
-      if (node["scheme"].Scalar() == "http") {
+      const auto scheme_val = node["scheme"].Scalar();
+      if (scheme_val == "http") {
         nh.scheme = PL_NH_SCHEME_HTTP;
-      } else if (node["scheme"].Scalar() == "https") {
+      } else if (scheme_val == "https") {
         nh.scheme = PL_NH_SCHEME_HTTPS;
       } else {
-        throw YAML::ParserException(node["scheme"].Mark(), "no valid scheme defined, valid schemes are http or https");
+        PL_NH_Note("Invalid scheme '%s' for protocol, setting to NONE", scheme_val.c_str());
```
ywkaras, line 405: ditto

ywkaras (approved):

plugins/experimental/parent_select/strategy.cc
rob05c: I didn't write the func, it's inherited from core. But it has to exist because the object is constructed by the YAML library.

I think `Init` is effectively like a constructor, because we can't construct, since we can't prevent the YAML library from doing the construction.

But, I think `Init` does make that assumption. There are probably other data it doesn't reset either. I don't object to making Init safe in case someone re-init's an object, and resetting all members to their defaults, but that's probably a larger change and might be better as its own PR

zwoop:
Cherry-picked to v9.2.x

PR #8832, opened 2022-05-06
Make post-early-return Au test more robust.

ywkaras:
Adds a delay between starting the origin server (an nc -l command)
for each test run and sending the POST request.


ywkaras:
This test seems stable in CI checks for PRs, but (without this change) fails frequently in my local test environment (a podman container on RHEL8).  Seems prudent to make it more robust.

zwoop:
Cherry-picked to v9.2.x

PR #8835, opened 2022-05-09
Remove incorrect comment from base64 functions

### LINKED ISSUE #7108: Comment for ats_base64_ functions seems wrong
The comment says functions return pointers for malloc-ed buffers, but the implementations don't seem like so.
https://github.com/apache/trafficserver/blob/c196db5f0a5b957983351ad51ef360bfc68fbd66/include/tscore/ink_base64.h#L34-L35

moonchen:
These functions have been changed to use user-supplied string buffers instead of allocating.  The comment is incorrect and confusing.

This fixes #7108.

### DIFF
--- a/include/tscore/ink_base64.h
+++ b/include/tscore/ink_base64.h
@@ -31,9 +31,6 @@
  * authentication scheme does not require them.  This implementation is
  * intended for web-related use, and line breaks are not implemented.
  *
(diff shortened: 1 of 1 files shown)


randall:
[approve ci autest]

zwoop:
Cherry-picked to v9.2.x

Issue #7341, opened 2020-11-20
☂️ATS is incompatible with OpenSSL 3.0 (unreleased)


maskit:
Several functions are going to be deprecated since OpenSSL 3.0.

- [ ] SHA256_Init/Update/Final
- [ ] ERR_get_error_line_data
- [ ] ENGINE API (e.g. ENGINE_get_default_RSA)

maskit:
> We are expecting 3.0 actual to be released in Q3 of this year.

https://mta.openssl.org/pipermail/openssl-users/2021-June/013900.html

bneradt:
Openssl 3.0 has been released now. @maskit : can you please update this issue with anything relevant now that it is released?

maskit:
I updated the list. The biggest change since the last check is that `HMAC()` function (one-shot HMAC function) was un-deprecated. HMAC low level functions were kept deprecated.

Also, this is not new info though, I realized that some deprecated functions are actually faster than the current equivalent functions (see #7447). It means we should not simply replace the deprecated functions with new ones.

jeredfloyd:
Fedora 36 has switched to OpenSSL 3.0.2 by default.  ATS builds successfully, but the tscore HKDF tests fail.

src/tscore/HKDF_openssl.cc:extract/expand() fail with EVP_PKEY_derive() returning failure.  These interfaces are deprecated in favor of the EVP_KDF_* interfaces, but I suspect this is an upstream bug that this simply fails.

I can build against the openssl1.1 compat packages in Fedora 36 for now so further investigation is lower on my priority list, but I welcome insight from OpenSSL experts!

jeredfloyd:
RHEL 9 does not ship with OpenSSL 1.1 compatibility libraries, so I'm looking to tackle the HKDF failures. Before I dig in, has anyone here already investigated?

```
-------------------------------------------------------------------------------
HKDF tests
  Basic test case with SHA-256
-------------------------------------------------------------------------------
unit_tests/test_HKDF.cc:30
...............................................................................

unit_tests/test_HKDF.cc:60: FAILED:
  CHECK( hkdf.extract(prk, &prk_len, salt, sizeof(salt), ikm, sizeof(ikm)) == 1 )
with expansion:
  -6 == 1
```

jeredfloyd:
Actual error message from OpenSSL is:
[...]

PR #8837, opened 2022-05-11
Add compatibility define when building with OpenSSL3

bryancall:
This is required for now when building with OpenSSL 3 until we switch to the new APIs and stop using the deprecated OpenSSL APIs.

### DIFF
--- a/build/crypto.m4
+++ b/build/crypto.m4
@@ -60,6 +60,25 @@ int main() {
 ])
 
 dnl
+dnl Check OpenSSL Version 3
+dnl
+AC_DEFUN([TS_CHECK_OPENSSL3], [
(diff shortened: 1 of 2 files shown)


bryancall:
[approve ci fedora]

maskit:
Linking to #7341 which has a list of the deprecated functions ATS uses.

maskit (approved):
I'm fine with this change for now but hope 10.0 release won't need it.

zwoop:
Are we really ok for master to use OpenSSL APIs for hashing etc. that calls malloc() on every lookup? This is in reference to 10.0.0 and the changes that were made there.

zwoop:
Cherry-picked to v9.2.x

bryancall:
Cherry picked to 9.1.x

PR #8838, opened 2022-05-11
mkdfa.c is not being used and doesn't compile with gcc 12.1.1

bryancall:


bryancall:
[approve ci debian]

zwoop:
Shouldn't we have nuked the mkdfa.c from the source ?

zwoop:
Cherry-picked to v9.2.x

bryancall:
Cherry picked to 9.1.x

PR #8846, opened 2022-05-12
Add `#pragma once` for PendingAction.h

bneradt:
PendingAction.h needs a `#pragma once` to protect it from double
inclusion. This patch adds this. This has not caused problems yet
because it is currently only used once in HttpSM.h, but as it gets more
use this will be required.


rob05c (approved):
LGTM. Obviously correct

zwoop:
Cherry-picked to v9.2.x

PR #8847, opened 2022-05-12
Restore down nameservers after they come back online

bneradt:
This patch ensures that a previously down DNS will be used again by ATS
after connectivity to it is restored. Before this patch, we would try to
reconnect to it after one attempt at DNS_PRIMARY_RETRY_PERIOD, but not
again after.

---

## For review

Calling `DNSHandler::mainEvent` will attempt to reconnect to the name server if it had been previously marked as down. Before this patch, calling this was done via a set of `schedule_in` calls which didn't result in it being called regularly. It was called once, but then not again after that. This simplifies the logic by periodically calling `mainEvent` every 5 seconds over the lifetime of ATS which will, if needed, retry any down name servers. This is the purpose of the new `schedule_every` being  added in `startEvent` which replaces the various `schedule_in` calls in other code paths. This guarantees that we will re-inspect any down name servers every 5 seconds.

### DIFF
--- a/iocore/dns/DNS.cc
+++ b/iocore/dns/DNS.cc
@@ -609,6 +609,11 @@ DNSHandler::startEvent(int /* event ATS_UNUSED */, Event *e)
       n_con = 1;
     }
 
+    // Retrying the name servers is something done periodically over the
+    // lifetime of the handler. This ensures that we don't miss retrying if it
+    // is necessary.
+    this->_dns_retry_event = this_ethread()->schedule_every(this, DNS_PRIMARY_RETRY_PERIOD);
+
     return EVENT_CONT;
   } else {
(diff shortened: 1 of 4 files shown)


bneradt:
[approve ci fedora]

bneradt:
[approve ci fedora]

bneradt:
[approve ci fedora]

SolidWallOfCode (approved):

bneradt:
Note that this PR fixes an issue that was introduced by:
https://github.com/apache/trafficserver/pull/7843

Since that latter PR only went into 9.2.x and beyond, this PR should only go into (and should only help issues with) 9.2.x and beyond as well.

zwoop:
Cherry-picked to v9.2.x

PR #8849, opened 2022-05-13
Jsonrpc socket credentials work( and related work to support this sort of checks )

brbzull0:
**I will split this PR into two separade ones, 1 with the Context change and another one for the credentials work, so it will be easy to takle on a review**

When moved from `traffic_manger` admin socket to the new `JSONRPC` api the management api [restrictions](https://docs.trafficserver.apache.org/en/latest/admin-guide/files/records.config.en.html?highlight=restricted#proxy.config.admin.api.restricted) were left out, the only restriction in place was related to the creation of the jsonrpc unix socket(either `700` or `777`) just at user level, before we have the capability to check the socket’s peers credentials and allow or deny certain operations, like read only(record lookup) vs operation that can change the state of ATS(like a record set). This was a useful method to allow some non-root users to perform some actions while we leave root users to have full access to the API. 
This change brings a similar base approach,  and allows the RPC API to let the developers specify this restriction when registering the jsonrpc node handler.

As(of course) the message parsing is done earlier than the actual rpc handler callback is made and there was no connection between them, so  I had to implement a sort of late check **(this logic is most of this change)** that can be set by the network part and be executed later on right before the actual call is made, now a Context object travels from the server to the actual rpc manager which can be populated with specific checks which are executed after the rpc handler was found valid.

All this permissions checks are handled by a new error code (`Unauthorized Action`) which is the high level error, depending of the particular check that failed to pass the sub error section will be set, leaving the client an error like this:

```
   {
      "jsonrpc": "2.0",
      "error":{
         "code": 10,
         "message": "Unauthorized action",
         "data":[
            {
               "code": 2,
               "message":"Denied privileged API access for uid=XXX gid=XXX"
            }
         ]
      },
      "id":"5e273ec0-3e3b-4a81-90ec-aeee3d38073f"
   }
```
This change also includes:
- Documentation support 
- Some clean up in the code.

### DIFF
--- a/doc/admin-guide/files/jsonrpc.yaml.en.rst
+++ b/doc/admin-guide/files/jsonrpc.yaml.en.rst
@@ -56,7 +56,7 @@ In order for programs to communicate with |TS|, the server exposes a
 will find some of the configurable arguments that can be changed.
 
 
-.. _admnin-jsonrpc-configuration:
+.. _admin-jsonrpc-configuration:
 
 Configuration
 =============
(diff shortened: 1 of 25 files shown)


brbzull0:
I am creating a new one with separated commits for an easy review.

PR #8850, opened 2022-05-14
Fix Rocky Linux 8 arm64 GCC Compiler Warnings

bneradt:
This patch addresses the compiler warnings raised by the GCC 11.2.1
compiler on Rocky Linux 8 arm64:

-Werror=shift-negative-value in I_Continuation.h:

```
In file included from src/ts_asf_master_test_build/iocore/eventsystem/I_Action.h:29,
                 from src/ts_asf_master_test_build/iocore/eventsystem/I_EventSystem.h:32,
                 from src/ts_asf_master_test_build/iocore/eventsystem/I_VConnection.h:29,
                 from QUICStreamVCAdapter.cc:24:
src/ts_asf_master_test_build/iocore/eventsystem/I_Continuation.h: In instantiation of _constexpr int (Continuation::* continuation_handler_void_ptr(int (C::*)(int, T*)))(int, void*) [with C = QUICStreamVCAdapter; T = void; ContinuationHandler = int (Continuation:
:*)(int, void*)]_:
QUICStreamVCAdapter.cc:29:3:   required from here
src/ts_asf_master_test_build/iocore/eventsystem/I_Continuation.h:74:10: error: left shift of negative value [-Werror=shift-negative-value]
   74 |   return static_cast<ContinuationHandler>(fp2);
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

-Werror=sign-compare in fastlz.cc:

```
fastlz.cc: In function _void flz_copy64(uint8_t*, const uint8_t*, uint32_t)_:
fastlz.cc:209:17: error: comparison of integer expressions of different signedness: _int_ and _uint32_t_ {aka _unsigned int_} [-Werror=sign-compare]
  209 |   for (c = 0; c < count * 8; ++c) {
      |               ~~^~~~~~~~~~~
```

-Werror=maybe-uninitialized in header_rewrite:

```
header_rewrite/conditions.cc: In static member function _static void ConditionInbound::append_value(std::string&, const Resources&, NetworkSessionQualifiers)_:
header_rewrite/conditions.cc:1131:56: error: _tags_ may be used uninitialized [-Werror=maybe-uninitialized]
 1131 |     TSHttpTxnClientProtocolStackGet(res.txnp, tags.size(), tags.data(), &count);
      |                                               ~~~~~~~~~^~
In file included from header_rewrite/conditions.cc:29:
/opt/rh/gcc-toolset-11/root/usr/include/c++/11/array:176:7: note: by argument 1 of type _const std::array<const char*, 8>*_ to _constexpr std::array<_Tp, _Nm>::size_type std::array<_Tp, _Nm>::size() const [with _Tp = const char*; long unsigned int _Nm = 8]_ declared here
  176 |       size() const noexcept { return _Nm; }
      |       ^~~~
header_rewrite/conditions.cc:1128:33: note: _tags_ declared here
 1128 |     std::array<char const *, 8> tags;
      |                                 ^~~~
```

-Werror=type-limits in traffic_dump:

```
experimental/traffic_dump/json_utils.cc: In function _int {anonymous}::esc_json_out(const char*, int64_t, std::ostream&)_:
experimental/traffic_dump/json_utils.cc:117:18: error: comparison is always true due to limited range of data type [-Werror=type-limits]
  117 |       if ('\x00' <= c && c <= '\x1f') {
      |           ~~~~~~~^~~~
```

### DIFF
--- a/iocore/eventsystem/I_Continuation.h
+++ b/iocore/eventsystem/I_Continuation.h
@@ -71,6 +71,14 @@ constexpr ContinuationHandler
 continuation_handler_void_ptr(int (C::*fp)(int, T *))
 {
   auto fp2 = reinterpret_cast<int (C::*)(int, void *)>(fp);
+
+  // We keep this a static_cast for added static type analysis from the
+  // compiler. If a compiler warning is generated for the following line of
+  // code of the type "-Werror=shift-negative-value", then this may be an issue
+  // with multiple inheritance of the C templated type. Make sure that for type
+  // C the Continuation parent is listed first (either directly or indirectly
+  // via the inheritance tree) before any other parent in the multiple class
+  // heirarchy of C.
   return static_cast<ContinuationHandler>(fp2);
 }
 
--- a/iocore/net/quic/QUICStreamVCAdapter.cc
+++ b/iocore/net/quic/QUICStreamVCAdapter.cc
@@ -24,7 +24,7 @@
 #include "I_VConnection.h"
 #include "QUICStreamVCAdapter.h"
 
-QUICStreamVCAdapter::QUICStreamVCAdapter(QUICStream &stream) : QUICStreamAdapter(stream), VConnection(new_ProxyMutex())
+QUICStreamVCAdapter::QUICStreamVCAdapter(QUICStream &stream) : VConnection(new_ProxyMutex()), QUICStreamAdapter(stream)
 {
   SET_HANDLER(&QUICStreamVCAdapter::state_stream_open);
 }
--- a/iocore/net/quic/QUICStreamVCAdapter.h
+++ b/iocore/net/quic/QUICStreamVCAdapter.h
@@ -25,7 +25,7 @@
 #include "QUICStreamAdapter.h"
 #include "I_IOBuffer.h"
 
-class QUICStreamVCAdapter : public QUICStreamAdapter, public VConnection
+class QUICStreamVCAdapter : public VConnection, public QUICStreamAdapter
 {
 public:
   class IOInfo;
--- a/lib/fastlz/fastlz.cc
+++ b/lib/fastlz/fastlz.cc
@@ -205,7 +205,7 @@ flz_copy64(uint8_t *dest, const uint8_t *src, uint32_t count)
 {
   const uint8_t *p = (const uint8_t *)src;
   uint8_t *q       = (uint8_t *)dest;
-  int c;
+  unsigned int c;
   for (c = 0; c < count * 8; ++c) {
     *q++ = *p++;
   }
--- a/plugins/experimental/traffic_dump/json_utils.cc
+++ b/plugins/experimental/traffic_dump/json_utils.cc
@@ -114,7 +114,7 @@ esc_json_out(const char *buf, int64_t len, std::ostream &jsonfile)
       break;
     }
     default: {
-      if ('\x00' <= c && c <= '\x1f') {
+      if (c <= '\x1f') {
         write_buffered_context(buf, prevIdx, idx, jsonfile);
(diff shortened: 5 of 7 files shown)


iocore/eventsystem/I_Continuation.h
```diff
@@ -71,7 +71,7 @@ constexpr ContinuationHandler
 continuation_handler_void_ptr(int (C::*fp)(int, T *))
 {
   auto fp2 = reinterpret_cast<int (C::*)(int, void *)>(fp);
-  return static_cast<ContinuationHandler>(fp2);
+  return reinterpret_cast<ContinuationHandler>(fp2);
```
bneradt, line 74: @ywkaras : Let me know whether you have thoughts about what I'm doing here. This patch is modifying code you updated to get better compile time checks (#6996). The compiler warning I'm addressing is the rather unobvious:

```
     In file included from src/ts_asf_master_test_build/iocore/eventsystem/I_Action.h:29,
                      from src/ts_asf_master_test_build/iocore/eventsystem/I_EventSystem.h:32,
                      from src/ts_asf_master_test_build/iocore/eventsystem/I_VConnection.h:29,
                      from QUICStreamVCAdapter.cc:24:
     src/ts_asf_master_test_build/iocore/eventsystem/I_Continuation.h: In instantiation of _constexpr int (Continuation::* continuation_handler_void_ptr(int (C::*)(int, T*)))(int, void*) [with C = QUICStreamVCAdapter; T = void; ContinuationHandler = int (Continuation:
     :*)(int, void*)]_:
     QUICStreamVCAdapter.cc:29:3:   required from here
     src/ts_asf_master_test_build/iocore/eventsystem/I_Continuation.h:74:10: error: left shift of negative value [-Werror=shift-negative-value]
        74 |   return static_cast<ContinuationHandler>(fp2);
           |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Changing to a reinterpret_cast from static_cast makes it happy, but if you have other ideas about how to address this, let me know.

randall (dismissed):

iocore/eventsystem/I_Continuation.h
ywkaras: If we change static_cast to reinterpret_cast, there will no longer be a compile-time check that C is derived from Continuation.  If C isn't we will just see undefined behavior at runtime.  This seem clearly to be a bizarre bug in gcc on Rocky Linux, as there is no shift operator involved.  Can we change it only on Rocky Linux using conditional compilation?

ywkaras: If someone could provide the contents of /etc/system-release on Rocky Linux, we could probably write an auto config macro to add '#define TS_ON_ROCKY_LINUX 0|1' in ink_config.h.

ywkaras (changes requested):
Use of reinterpret_cast may result in a wild **this** pointer, and it bypasses some compile-time checking.

ywkaras (approved):

iocore/eventsystem/I_Continuation.h
maskit: I happened to find this conversation (not watching changes to QUIC/H3 files 😎 ). I'm totally fine with switching the order of the parents, I wonder what would be the runtime difference though.

ywkaras: > I wonder what would be the runtime difference though.
It would be none, unless the compiler optimizer is able to detect cases at compile time where the address offset in the member function pointer is zero, and omit code to add it.  But, in that case, the change makes it faster, because it's causing the offset to be zero 😎 .  (Unless the the other base class besides VConnection has virtual functions, then it might be slower, but not by much.)

randall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8852, opened 2022-05-16
traffic_ctl - JSONRPC: make sure we display the error regardless of the formatting type.

brbzull0:
While working on a different pr I found out that some error weren't displayed by `traffic_ctl` as it was before the jsonrpc change, this change fixes this and also changes the output style for a "better" quick read, more like it was before jsonrpc.

Now this prints the main error and in a second level the more specifics. There could be more than one second level error.
```
$ traffic_ctl config set invalid.record val
Server Error found:
[9] Error during execution
- [2000] Record not found.
```

### DIFF
--- a/include/shared/rpc/RPCRequests.h
+++ b/include/shared/rpc/RPCRequests.h
@@ -203,24 +203,10 @@ operator<<(std::ostream &os, const RecordLookUpResponse::RecordError &re)
 inline std::ostream &
 operator<<(std::ostream &os, const JSONRPCError &err)
 {
-  os << "Error found.\n";
-  os << "code: " << err.code << '\n';
-  os << "message: " << err.message << '\n';
-  if (err.data.size() > 0) {
-    os << "---\nAdditional error information found:\n";
-    auto my_print = [&](auto const &e) {
-      os << "+ code: " << e.first << '\n';
-      os << "+ message: " << e.second << '\n';
-    };
-
-    auto iter = std::begin(err.data);
-
-    my_print(*iter);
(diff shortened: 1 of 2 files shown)


SolidWallOfCode (approved):
What happened to pretty printing? Is that always the case now?

include/shared/rpc/RPCRequests.h
```diff
@@ -203,23 +203,15 @@ operator<<(std::ostream &os, const RecordLookUpResponse::RecordError &re)
...
-
     auto iter = std::begin(err.data);
 
-    my_print(*iter);
+    os << "- [" << iter->first << "] " << iter->second << '\n';
     ++iter;
     for (; iter != std::end(err.data); ++iter) {
-      os << "---\n";
-      my_print(*iter);
+      os << "- [" << iter->first << "] " << iter->second << '\n';
```
cmcfarlen, line 214: Since the interleaved `---` is no longer being printed, this could just be a simple `for (auto it: err.data) ...` loop.

brbzull0:
> What happened to pretty printing? Is that always the case now?

No, it still work for  cases  that it's implemented and you request it, but the pretty printing is more for message ouitput(`jsonrpc.result`) and not for the error. This is for  general error, so it will be printed anyway  regardless your `--format` choice.

include/shared/rpc/RPCRequests.h
brbzull0: True.  Missed that, thanks a lot, fixed now.

PR #8857, opened 2022-05-20
Add ts::overloaded as a type matching helper of std::visit

masaori335:
This is pretty useful when we use `std::variant` and `std::visit`. An example is in the comment. 
More details are in https://en.cppreference.com/w/cpp/utility/variant/visit.

### DIFF
--- a/include/tscpp/util/Overloaded.h
+++ b/include/tscpp/util/Overloaded.h
@@ -1,0 +1,52 @@
+/** @file
+
+   A brief file description
+
+   @section license License
+
+   Licensed to the Apache Software Foundation (ASF) under one
+   or more contributor license agreements.  See the NOTICE file
+   distributed with this work for additional information
+   regarding copyright ownership.  The ASF licenses this file
+   to you under the Apache License, Version 2.0 (the
+   "License"); you may not use this file except in compliance
+   with the License.  You may obtain a copy of the License at
+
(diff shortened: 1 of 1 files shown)


brbzull0:
Very useful, I had this defined in [10-Dev](https://github.com/apache/trafficserver/blob/d1313eab797c2183b1c13653fbb50d6a9266384a/include/tscpp/util/ts_meta.h#L98-L103) already and now that we have included swoc, it's also defined [here](https://github.com/apache/trafficserver/blob/ef917f9277e9cc3d254d6f458a98703a767201f3/lib/swoc/include/swoc/swoc_meta.h#L115-L133)

ywkaras:
I feel hesitant about heavy use of std::visit.  In many cases, you can use if-else-if instead:  https://godbolt.org/z/bsqhf1dq5 .  It's more straight-forward, and not significantly more verbose.  It looks like the one case where std::visit has a clear advantage is when one of the lambda functions uses an auto reference, and can handle multiple fields in the variant.

In (out of date) gcc 8.3 (the version we use at Yahoo), the optimizer doesn't optimize std::visit well (gcc 12 does).  But I couldn't blame the ATS community if they don't have great sympathy for our self-inflicted wounds.

masaori335:
@brbzull0 Cool, I haven't noticed we already have this. Obviously, we don't need three places.

Issue #7296, opened 2020-10-25
Benchmarking BufferWriter


bryancall:
I am seeing BufferWriter slower than snprinf and fmtlib.  Here are are the benchmark numbers:
```
buffer_writer_bigger   685922313 ns    682309686 ns            1
fmtlib_compile_bigger   17078534 ns     16988623 ns           40
fmtlib_bigger          121250664 ns    120652887 ns            7
snprintf_bigger        144683098 ns    143956151 ns            5
```

Here is the code I am running for the benchmark:
https://github.com/bryancall/format-benchmark/tree/mo_benchmarking

Here is a discussion I was having today with the author of fmtlib:
https://github.com/fmtlib/format-benchmark/issues/20

zwoop:
Seems we should pull in Fmtlib ? What's the license ?

bryancall:
@zwoop it is a pretty open license: https://github.com/fmtlib/fmt/blob/master/LICENSE.rst

giorgiozoppi:
and moving at c++20?

bryancall:
Closing this as obsolete. This was a 2020 performance investigation comparing the in-tree tscore BufferWriter against snprintf and fmtlib. Since then the in-tree BufferWriter and its formatting code have been removed, and Traffic Server now uses the external libswoc library (bundled under lib/swoc/) for BufferWriter and its formatting, maintained upstream by the same author referenced in the original fmtlib discussion. The specific implementation that was benchmarked here no longer exists in the tree, so this investigation no longer tracks actionable work. Reopen or file a fresh issue if there is a current libswoc BufferWriter performance concern.

PR #8867, opened 2022-05-25
Add utility for printing string_view with printf-like functions.

ywkaras:


maskit:
I guess it'd be slower than current because of memcpy, right? I'm not keen on using the non-standard utility just to shorten lines calling printf-like functions a little. I hope a new format specifier for string_view will be introduced in the future though. Any pros I'm missing?

ywkaras:
C++ supports printf() only as  legacy function, so I'd be surprised if a specifier were added in the standard lib.

If we add a string_view format specifier in our functions, we might not be able to declare them "printf-like" using the GCC extension, and we can't loose that compile time checking.

ywkaras:
Me personally, I still prefer:  https://github.com/apache/trafficserver/pull/7711

maskit:
I'd say #7711 is better, but still not my cup of tea.

ywkaras:
> What do you think about using fmtlib? Seems like it supports string_view.
> 
> #7296

Looks good except for one apparent, huge flaw, when you print to a buffer, you don't pass in the buffer length it seems:  https://github.com/bryancall/format-benchmark/commit/0fc372058e7db96de3423a989173a9bc0fe68533#diff-a7cd0eba35fcc2bae863d4093c81bac869812eb9017cbf152308481c80af9716R430

So presumably that means you can corrupt memory potentially.

rob05c:
> Looks good except for one apparent, huge flaw, when you print to a buffer, you don't pass in the buffer length it seems: > https://github.com/bryancall/format-benchmark/commit/0fc372058e7db96de3423a989173a9bc0fe68533#diff-a7cd0eba35fcc2bae863d4093c81bac869812eb9017cbf152308481c80af9716R430

> So presumably that means you can corrupt memory potentially.

You can pass the buffer length with `format_to_n`

https://en.cppreference.com/w/cpp/utility/format/format_to_n
https://fmt.dev/latest/api.html#_CPPv4I0DpEN3fmt11format_to_nE18format_to_n_resultI8OutputItE8OutputIt6size_t13format_stringIDp1TEDpRR1T

PR #8868, opened 2022-05-26
Update to Proxy Verifier v2.3.3

bneradt:
Proxy Verifier v2.3.3 adds a chunk verification check that makes the
bad_chunked_encoding AuTest more resilient to timing issues.

---

Yahoo's internal CI had some timing issues related to bad_chunked_encoding.test.py which made it fail frequently in our internal CI. This was caused when the chunk body of some of the bad chunk responses came in separately from the headers. The headers would come through before the body, which was not what the test expected. It expected ATS to always close the connection abruptly without headers. Proxy Verifier v2.3.3 added a chunk strictness check that the test can now use to verify ATS closed the connection more reliably, even if the headers come first.


randall (approved):

zwoop:
Cherry-picked to v9.2.x

Issue #8860, opened 2022-05-23
AuTest CI: inconsistent remap_reload test results


bneradt:
I've noticed that the new `remap_reload` test is a bit inconsistent. Out of the last 24 autest CI runs, it has failed 4 times:

* https://ci.trafficserver.apache.org/view/Github/job/Github_Builds/job/autest/996/console
* https://ci.trafficserver.apache.org/view/Github/job/Github_Builds/job/autest/997/console
* https://ci.trafficserver.apache.org/view/Github/job/Github_Builds/job/autest/1002/console
* https://ci.trafficserver.apache.org/view/Github/job/Github_Builds/job/autest/1007/console

PR #8869, opened 2022-05-26
Autest: Increase the delay time after a config reload rpc all was done.

brbzull0:
This particular test `remap_reload` seems to be failing randomly, it could be the delay time was too short, setting it to 10s now.

Let's see if this fixes https://github.com/apache/trafficserver/issues/8860

If not, I'll be happy to revert this back.

### DIFF
--- a/tests/gold_tests/remap/remap_reload.test.py
+++ b/tests/gold_tests/remap/remap_reload.test.py
@@ -56,28 +56,24 @@ tr_2.Disk.File(remap_cfg_path, typename="ats:config").AddLines([
     f"map http://alpha.ex http://alpha.ex:{pv_port}", f"map http://bravo.ex http://bravo.ex:{pv_port}"
 ])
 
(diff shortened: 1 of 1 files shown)


brbzull0:
> Thank you for working on this. Let's run the autest in here a few times (maybe 4?). If the test passes each of those times, let's merge it in.

sound good to me! thanks.

bneradt:
[approve ci autest]

bneradt:
[approve ci autest]

bneradt:
[approve ci autest]

brbzull0:
[approve ci autest]

bneradt:
Reading through your patch in this PR made me realize that rather than waiting a certain amount of time for the reload to finish, we can instead await a log message that indicates that the reload has finished. Playing around with it a bit, I came up with this patch:

https://github.com/apache/trafficserver/pull/8872

Any thoughts?

PR #8871, opened 2022-05-26
Handle hit as miss when cach obj is invalid

serrislew:
When cache object is invalid but has a hit status, it should be handled as a miss so no assumptions or invalid access to the object will be made

### DIFF
--- a/proxy/http/HttpTransact.cc
+++ b/proxy/http/HttpTransact.cc
@@ -2798,6 +2798,11 @@ HttpTransact::HandleCacheOpenReadHit(State *s)
     obj = s->cache_info.object_read;
   }
 
+  if (obj == nullptr || !obj->valid()) {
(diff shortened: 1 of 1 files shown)


bneradt (approved):

bneradt:
@serrislew: thank you for the fix.  I've marked this just for 10.x for now. Do you think this should go to 9.2.x? If so, please add a comment with your reasons and I can mark as such.

serrislew:
Yes 9.2.x would be better. We encountered a crash where the cached object was a hit and suspect a plugin marked it as null before utilizing it, ultimately handling an invalid object. It relates to the "serve stale when DNS is down" PR since all hit status objects go to `HandleCacheOpenReadHit` to revalidate before determining how to serve the cached object

zwoop:
Cherry-picked to v9.2.x

PR #8872, opened 2022-05-27
remap_reload AuTest: await config reload messages

bneradt:
Change the remap_reload AuTest from waiting an arbitrary amount of time
(5 seconds) for the config reload to finish to instead waiting upon a
log message to indicate the reload has finished. This resolves some
flaky test results from the test when the previous 5 second sleep wasn't
enough time for the reload to finish.

Fixes: #8860


masaori335 (approved):
Looks reasonable. Thanks for making it robust.

brbzull0 (approved):
Better this way.

As a note I think 30s is a lot,  a reload shouldn't take more than 10s to be done.

PR #8873, opened 2022-05-29
Add proxy.process.hostdb.total_serve_stale

bneradt:
Add a HostDB statistic for the number of times we serve a stale DNS
response from the HostDB cache while a DNS refresh fetch is taking
place.


randall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8879, opened 2022-06-01
Adds a new member for TSRemapInterface tracking the plugin loaded

zwoop:
This allows for a plugin to introspect itself when TSRemapInit() is
being called. It's currently only implementing TSRemapDLHandleGet(),
which returns the handle to the .so as returned by dlopen(). With
this handle, a plugin can use e.g. dlsym() and dlinfo() to introspect
the plugin that it's loaded with.

Note that I've put this on 10-Dev. It's perhaps not ABI incompatible,
but seeing that it changes ts/remap.h internals, better to be safe
then sorry.

### DIFF
--- a/doc/developer-guide/api/functions/TSTypes.en.rst
+++ b/doc/developer-guide/api/functions/TSTypes.en.rst
@@ -196,6 +196,16 @@ more widely. Those are described on this page.
       The API version of the C API. The lower 16 bits are the minor version, and the upper bits
       the major version.
 
+   .. member:: TSRemapPluginInfo plugin_info
+
+      An opaque object, which holds various details about the currently loaded and running
+      plugin. You can't access the details directly, but rather must call an acess function
+      such as :func:`TSRemapDLHandleGet`.
+
+.. type:: TSRemapPluginInfo
+
+   Opaque data passed to a remap plugin via :func:`TSRemapInit` in the `plugin_info` member above.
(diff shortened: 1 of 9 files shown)


zwoop:
This also fixes an unrelated clang-format issue that snuck into the 10-Dev branch.

shukitchan:
can we add an example and documentation? That will be really helpful.

brbzull0:
[approve ci autest]

randall:
[approve ci autest]

zwoop:
Added a doc section, but not an example (the API is pretty clear IMO, and none of the existing plugins needs it).

doc/developer-guide/api/functions/TSTypes.en.rst
```diff
@@ -196,6 +196,16 @@ more widely. Those are described on this page.
       The API version of the C API. The lower 16 bits are the minor version, and the upper bits
       the major version.
 
+   .. member:: TSRemapPluginInfo plugin_info
+
+      An opaque object, which holds various details about the currently loaded and running
+      plugin. You can't access the details directly, but rather must call an acess function
```
randall, line 202: access

bneradt:
[approve ci autest]

zwoop:
I fixed the typo Randall pointed out, so need another +1 :).

PR #8882, opened 2022-06-02
Make ats_scoped_obj a synonym for std::unique_ptr .

ywkaras:


maskit:
> Could we just delete ats_scoped_obj entirely and change all the usages to std::unique_ptr?

+1

randall (approved):

PR #8886, opened 2022-06-04
Add escape json for logging

hnakamur:
This pull requests adds an optional `escape` attribute in `format` entries in `logging.yaml`.
The possible values are `json` and `none`.

An example `logging.yaml`

```
logging:
  formats:
    - name: "mycustom"
      escape: "json"
      format: '{"chi":"%<chi>","caun":"%<caun>","cqtn":"%<cqtn>","cqtx":"%<cqtx>","pssc":"%<pssc>","pscl":"%<pscl>","sssc":"%<sssc>","sscl":"%<sscl>","cqcl":"%<cqcl>","pqcl":"%<pqcl>","cqhl":"%<cqhl>","pshl":"%<pshl>","pqhl":"%<pqhl>","sshl":"%<sshl>","tts":"%<tts>","phr":"%<phr>","cfsc":"%<cfsc>","pfsc":"%<pfsc>","crc":"%<crc>","my_header":"%<{My-Header}cqh>","my_header_slice":"%<{My-Header}cqh[2:-3]>"}'

  logs:
    - filename: custom
      format: mycustom
      mode: ascii
```

An example request:

```
curl -v -H $'My-Header: a\tbcde\x1dfghijkl\x1em' 'localhost:8080?a=c\td'
```

An example log output:

```
{"chi":"127.0.0.1","caun":"-","cqtn":"04/Jun/2022:10:46:55 +0900","cqtx":"GET http://127.0.0.1/?a=c\\td HTTP/1.1","pssc":"200","pscl":"34","sssc":"200","sscl":"34","cqcl":"0","pqcl":"0","cqhl":"125","pshl":"242","pqhl":"256","sshl":"236","tts":"0","phr":"DIRECT","cfsc":"FIN","pfsc":"FIN","crc":"TCP_MISS","my_header":"a\tbcde\u001dfghijkl\u001em","my_header_slice":"bcde\u001df"}
```

### DIFF
--- a/proxy/logging/LogAccess.cc
+++ b/proxy/logging/LogAccess.cc
@@ -643,6 +643,126 @@ LogAccess::unmarshal_str(char **buf, char *dest, int len, LogSlice *slice)
   return -1;
 }
 
+static int
+escape_json(char *dest, const char *buf, int len)
+{
+  ink_assert(buf != nullptr);
+
+  int escaped_len = 0;
+
+  for (int i = 0; i < len; i++) {
+    char c = buf[i];
+    switch (c) {
+    case '"':
+    case '\\': {
+      if (dest) {
+        *dest++ = '\\';
+        *dest++ = c;
+      }
+      escaped_len += 2;
+      break;
+    }
+    case '\b': {
+      if (dest) {
+        *dest++ = '\\';
+        *dest++ = 'b';
+      }
+      escaped_len += 2;
+      break;
+    }
+    case '\f': {
+      if (dest) {
+        *dest++ = '\\';
+        *dest++ = 'f';
+      }
+      escaped_len += 2;
+      break;
+    }
+    case '\n': {
+      if (dest) {
+        *dest++ = '\\';
+        *dest++ = 'n';
+      }
+      escaped_len += 2;
+      break;
+    }
+    case '\r': {
+      if (dest) {
+        *dest++ = '\\';
+        *dest++ = 'r';
+      }
+      escaped_len += 2;
+      break;
+    }
+    case '\t': {
+      if (dest) {
+        *dest++ = '\\';
+        *dest++ = 't';
+      }
+      escaped_len += 2;
+      break;
+    }
+    default: {
+      if (c <= '\x1f') {
+        if (dest) {
+          *dest++ = '\\';
+          *dest++ = 'u';
+          *dest++ = '0';
+          *dest++ = '0';
+          *dest++ = '0' + (c >> 4);
+          int d   = c & 0x0f;
+          *dest++ = (d >= 10 ? 'a' + (d - 10) : '0' + d);
+        }
+        escaped_len += 6;
+      } else {
+        if (dest) {
+          *dest++ = c;
+        }
+        escaped_len++;
+      }
+      break;
+    }
+    }
+  }
+  return escaped_len;
+}
+
+int
+LogAccess::unmarshal_str_json(char **buf, char *dest, int len, LogSlice *slice)
+{
+  ink_assert(buf != nullptr);
+  ink_assert(*buf != nullptr);
+  ink_assert(dest != nullptr);
+
+  char *val_buf   = *buf;
+  int val_len     = static_cast<int>(::strlen(val_buf));
+  int escaped_len = escape_json(nullptr, val_buf, val_len);
+
+  *buf += LogAccess::strlen(val_buf); // this is how it was stored
+
+  if (slice && slice->m_enable) {
+    int offset, n;
+
+    n = slice->toStrOffset(val_len, &offset);
+    if (n <= 0) {
+      return 0;
+    }
+
+    if (n >= len) {
+      return -1;
+    }
+
+    escape_json(dest, (val_buf + offset), n);
+    return n;
+  }
+
+  if (escaped_len < len) {
+    escape_json(dest, val_buf, val_len);
+    return escaped_len;
+  }
+  return -1;
+}
+
 int
 LogAccess::unmarshal_ttmsf(char **buf, char *dest, int len)
 {
@@ -801,6 +921,35 @@ LogAccess::unmarshal_http_text(char **buf, char *dest, int len, LogSlice *slice)
   p += res1;
   *p++     = ' ';
   int res2 = unmarshal_str(buf, p, len - res1 - 1, slice);
+  if (res2 < 0) {
+    return -1;
+  }
+  p += res2;
+  *p++     = ' ';
+  int res3 = unmarshal_http_version(buf, p, len - res1 - res2 - 2);
+  if (res3 < 0) {
+    return -1;
+  }
+  return res1 + res2 + res3 + 2;
+}
+
+int
+LogAccess::unmarshal_http_text_json(char **buf, char *dest, int len, LogSlice *slice)
+{
+  ink_assert(buf != nullptr);
+  ink_assert(*buf != nullptr);
+  ink_assert(dest != nullptr);
+
+  char *p = dest;
+
+  //    int res1 = unmarshal_http_method (buf, p, len);
+  int res1 = unmarshal_str_json(buf, p, len);
+  if (res1 < 0) {
+    return -1;
+  }
+  p += res1;
+  *p++     = ' ';
+  int res2 = unmarshal_str_json(buf, p, len - res1 - 1, slice);
   if (res2 < 0) {
     return -1;
   }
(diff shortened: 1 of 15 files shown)


masaori335:
[approve ci]

bryancall:
[approve ci]

ywkaras:
We had a fairly long conversation about this in the weekly Monday meeting.  We talked about whether there might be a more general solution, but couldn't immediately think of one.

proxy/logging/LogAccess.cc
```diff
@@ -643,6 +643,126 @@ LogAccess::unmarshal_str(char **buf, char *dest, int len, LogSlice *slice)
...
+static int
+escape_json(char *dest, const char *buf, int len)
+{
+  ink_assert(buf != nullptr);
+
+  int escaped_len = 0;
+
+  for (int i = 0; i < len; i++) {
+    char c = buf[i];
+    switch (c) {
```
ywkaras, line 655: Here are two alternatives to a switch statement:  https://godbolt.org/z/G3fz3469v

hnakamur: Thanks for your suggestion.
I tried to modify `escape_json` using bitset but I found that I need the mapping like `\b` to `b`.

I made a modified version at
https://github.com/hnakamur/json-escape-cpp-experiment/blob/main/test.cpp

This version properly escapes the DEL `0x7f` character too.

Can I replace `escape_json` with this one?

hnakamur:
Thanks for taking time to discuss log escape in the meeting.
`escape_json` function in this pull request was based on [esc_json_out](https://github.com/apache/trafficserver/blob/69ebe48fe7e9c927e6b876ca4b7c7a60e99868cd/plugins/experimental/traffic_dump/json_utils.cc#L73-L134) in the traffic_dump plugin.

> Does the escaping assume the message has no deviations from the IETF standards?

Could you tell me what do you mean by "the IETF standards"?
I would like to do the string escape defined in https://www.json.org/json-en.html

proxy/logging/LogAccess.cc
ywkaras: How about this?
```
namespace
{
class EscLookup
{
public:
  static const char NO_ESCAPE{'\0'};
  static const char LONG_ESCAPE{'\x01'};

  static char result(char c)
  {
    return _lu.table[static_cast<unsigned char>(c)];
  }

private:
  struct _LUT
  {
    _LUT();

    unsigned short table[0x7f];
  };

  inline static _LUT const _lu;
};

EscLookup::_LUT::_LUT()
{
  for (unsigned i = 0; i < ' '; ++i) {
    table[i] = LONG_ESCAPE;
  }
  for (unsigned i = '\x7f'; i < sizeof(table); ++i) {
    table[i] = LONG_ESCAPE;
  }

  // Short escapes.
  //
  table['\b'] = 'b';
  table['\t'] = 't';
  table['\n'] = 'n';
  table['\f'] = 'f';
  table['\r'] = 'r';
  table['\\'] = '\\';
  table['\"'] = '\\';
}

char nibble(int nib)
{
  return nib >= 0xa ? 'a' + (nib - 0xa) : '0' + nib;
}

} // end anonymous namespace

static int
escape_json(char *dest, const char *buf, int len)
{
  int escaped_len = 0;

  for (int i = 0; i < len; i++) {
    char c = buf[i];
    char ec = EscLookup::result(c);
    if (__builtin_expect(EscLookup::NO_ESCAPE == ec, 1)) {
      if (dest) {
        *dest++ = c;
      }
      escaped_len++;

    } else if (EscLookup::LONG_ESCAPE == ec) {
      if (dest) {
        *dest++ = '\\';
        *dest++ = 'u';
        *dest++ = '0';
        *dest++ = '0';
        *dest++ = nibble(static_cast<unsigned char>(c) >> 4);
        *dest++ = nibble(c & 0x0f);
      }
      escaped_len += 6;

    } else { // Short escape.
      if (dest) {
        *dest++ = '\\';
        *dest++ = ec;
      }
      escaped_len += 2;
    }
  } // end for
  return escaped_len;
}
```

hnakamur: Thanks! I added commit https://github.com/apache/trafficserver/pull/8886/commits/ab9772927c23505b0b8b7e247428e0b5048364b4 before noticing your code.
I'll update it later.

Also I noticed slicing is done after escape now and produces incorrect strings like `\u0` or so.
I'll looking into it later this too.

ywkaras:
That page seems to say / should also be escaped.  It doesn't say that " should  be escaped as \\ .  Seems like that would confuse parses of the log, if \ and " both are mapped to \\.

I looked through the IETF standards a little, they are not very clear.  I think the function should handle bytes in the 0x80 to 0xff .

traffic_dump is only uses once in a while.  It would not do harm to make the function more efficient, since logging is often done for each and every transaction.

hnakamur:
* " must be escaped as \", not as \\
* It seems ok / not escaped

[ECMA-404 The JSON Data Interchange Standard](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf) (linked from https://www.json.org/json-en.html) p.11
> All code points may be placed within the quotation marks except for the code points that must be escaped: quotation mark
(U+0022), reverse solidus (U+005C), and the control characters U+0000 to U+001F.

proxy/logging/LogAccess.cc
```diff
@@ -813,6 +903,35 @@ LogAccess::unmarshal_http_text(char **buf, char *dest, int len, LogSlice *slice)
...
+int
+LogAccess::unmarshal_http_text_json(char **buf, char *dest, int len, LogSlice *slice)
+{
+  ink_assert(buf != nullptr);
+  ink_assert(*buf != nullptr);
+  ink_assert(dest != nullptr);
+
+  char *p = dest;
+
+  //    int res1 = unmarshal_http_method (buf, p, len);
```
ywkaras, line 915: Why not remove this?

ywkaras:
As well as fixing the compiler warnings, I think we need some Au test coverage of this new feature before it can be merged.

ywkaras:
> * `"` must be escaped as `\"`, not as `\\`

Sorry I read the code wrong.

> * It seems ok `/` not escaped

The goal is that someone could write a script to parse a log without having to read the source code to find the details of how the escaping works.  If the escaping standard has different versions and options, we should say in the ATS documentation which version and options ATS implements.

proxy/logging/LogAccess.cc
hnakamur: Copied the code at https://github.com/apache/trafficserver/pull/8886#discussion_r892896951 to https://github.com/apache/trafficserver/pull/8886/commits/a65b8b4b6284e049e860d85119ec311ce03a5ccf with specifying the author [ywkaras](https://github.com/ywkaras). Is it OK with you?

hnakamur:
Changed to escape forward slash at https://github.com/apache/trafficserver/pull/8886/commits/110ce548ea53ac5c8ad0d6049c81c6e084319ccd

The tasks left:

* [ ] Fix slicing
* [ ] Add tests

proxy/logging/LogAccess.cc
ywkaras: > Copied the code at [#8886 (comment)](https://github.com/apache/trafficserver/pull/8886#discussion_r892896951) to [a65b8b4](https://github.com/apache/trafficserver/commit/a65b8b4b6284e049e860d85119ec311ce03a5ccf) with specifying the author [ywkaras](https://github.com/ywkaras). Is it OK with you?

Yes OK.

proxy/logging/LogAccess.cc
```diff
@@ -643,6 +643,141 @@ LogAccess::unmarshal_str(char **buf, char *dest, int len, LogSlice *slice)
...
+  for (unsigned i = 0; i < ' '; ++i) {
+    table[i] = LONG_ESCAPE;
+  }
+  for (unsigned i = '\x7f'; i < sizeof(table); ++i) {
+    table[i] = LONG_ESCAPE;
+  }
+
+  // Short escapes.
+  //
+  table['\b'] = 'b';
```
ywkaras, line 681: Our PR checks are now tricky to use.  After you click on details, you must then again find the URL for the check in the list of URLs, and click on it.  Then you can click Console Output.

With the change above, you can change these to be like:
```
table('\\b') = 'b';
```
to get rid of the warning that is treated as an error.

ywkaras (approved):

hnakamur:
Thanks for your view!

zwoop:
Cherry-picked to v9.2.x

PR #8889, opened 2022-06-07
remap_reload test: Spacing out file modification and reload

bneradt:
The remap_reload.test.py AuTest is still flaky despite the last effort
to make it reliable. It looks like this might be because the
modification of the remap file and the reload is racing with each other.
This puts them in separate TestRuns and spaces them out by a few seconds
to see whether this makes the test more reliable.

---

## For Review

Let's run this a few times in CI to see if this improves reliability. Running this in a loop locally failed to reproduce this locally for me.

### DIFF
--- a/tests/gold_tests/remap/remap_reload.test.py
+++ b/tests/gold_tests/remap/remap_reload.test.py
@@ -42,46 +42,53 @@ tm.Disk.records_config.update({
     'proxy.config.dns.nameservers': f"127.0.0.1:{nameserver.Variables.Port}", 'proxy.config.dns.resolv_conf': 'NULL'
 })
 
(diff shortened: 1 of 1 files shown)


bneradt:
The first remap_reload autest passed. Running again.

bneradt:
[approve ci autest]

bneradt:
[approve ci autest]

masaori335 (approved):
Looks good. Please merge when your retry test on the ci box is done.

zwoop:
This file doesn't exist in 9.2.x

bneradt:
> This file doesn't exist in 9.2.x

Ah, right. Good catch. I forgot that we didn't put this work in 9.2.x. I'll remove that project.

Thanks!

PR #8890, opened 2022-06-08
Update slice to only prefetch when first block is miss/hit-stale

serrislew:
Version 2 of prefetch feature in slice plugin. 

Prefetch in background if first block is cacheable and has miss or hit-stale cache status (similar behavior as prefetch plugin). Checking cache lookup and fill status can prevent unnecessary background fetches if blocks are already fresh in the cache. 
- Docs updated
- Autest updated to check logs for background fetches

Functional and performance testing complete. Internal experiments show improvement in latency


serrislew:
Removed plugin dependency of `xdebug` and `header_rewrite`
- `xdebug` is still used in autest for testing purposes with logs to ensure blocks were cached with hit-fresh status when expected

slice prefetching only depends on crr plugin to cache blocks
- headers are sent between slice and crr to indicate whether cache lookup is needed from crr and if needed, crr will send header back to slice to indicate block is cacheable with a hit-stale or miss status. this will inform slice that the successive blocks are prefetchable 
- request header from slice to crr will only be sent for the first block and if prefetching is enabled 
- headers are stripped before sending, to prevent user sent headers

traeak (approved):
Changes look fine and test as expected.

zwoop:
Cherry-picked to v9.2.x

PR #8893, opened 2022-06-08
Fix %<chi> with PROXY Protocol

### LINKED ISSUE #8544: Feature: Add "logging field" to identify proxy protocol client's IP address (previous proxy hop)
Relevant fields for this discussion (from https://docs.trafficserver.apache.org/en/latest/admin-guide/logging/formatting.en.html#admin-logging-fields-network) 

| Field | Source | Description |
| --- | --- | --- |
| chi | Client | IP address of the client’s host |
| pps | Proxy Protocol Source IP | Source IP received via Proxy Protocol context from the LB to the Traffic Server |

With only these relevant fields available, if a `proxy-protocol` client connects to ATS while passing the right `proxy-protocol` headers, both fields `chi` and `pps` are updated with the `true client IP address`, ie, the IP address of the user-agent. (meaning a typical user on the internet)

This means we lose the ability to identify the `IP address` of the previous `hop` when proxy protocol is used. Is there a need for adding a new field that signals the IP address of the `previous hop` when proxy protocol is used?

masaori335:
Fix #8544.

Prior to this change, the source address in the PROXY Protocol message was set as the remote address when it was parsed.
Now, `%<chi>` always represents the previous hop.


proxy/ProtocolProbeSessionAccept.cc
```diff
@@ -123,7 +123,6 @@ struct ProtocolProbeTrampoline : public Continuation, public ProtocolProbeSessio
 
       if (netvc->has_proxy_protocol(reader)) {
         Debug("proxyprotocol", "ioCompletionEvent: http has proxy protocol header");
-        netvc->set_remote_addr(netvc->get_proxy_protocol_src_addr());
```
masaori335, line 126: This is the main change.

iocore/net/SSLNetVConnection.cc
```diff
@@ -455,7 +454,6 @@ SSLNetVConnection::read_raw_data()
 
     if (this->has_proxy_protocol(buffer, &r)) {
       Debug("proxyprotocol", "ssl has proxy protocol header");
-      set_remote_addr(get_proxy_protocol_src_addr());
```
masaori335, line 458: This is the main change.

masaori335: I'm wondering if we can unify these code in the `SSLNetVConnection` and `ProtocolProbeSessionAccept`, but it's out of scope from this PR.

masaori335:
The doc of `%<chi>` is updated.

ywkaras (approved):

PR #8900, opened 2022-06-09
Eliminate potentially error-prone function pointer casting in logging code.

ywkaras:


proxy/logging/LogField.h
```diff
@@ -82,6 +82,12 @@ class LogField
   typedef int (*UnmarshalFuncWithMap)(char **buf, char *dest, int len, const Ptr<LogFieldAliasMap> &map);
   typedef void (LogAccess::*SetFunc)(char *buf, int len);
 
+  static UnmarshalFunc
+  castToUnmarshalFunc(UnmarshalFuncWithSlice func)
+  {
+    return reinterpret_cast<UnmarshalFunc>(func);
+  }
+
```
ywkaras, line 90: Perhaps it would be better to consolidate m_unmarshal_func and m_unmarshal_func_map into an std::variant and also add an UnmarshalFuncWithSlice variant.

https://github.com/apache/trafficserver/blob/12801c966e5710d818287e79a7a977e837bfddc6/proxy/logging/LogField.h#L192

PR #8902, opened 2022-06-09
Fix plugin stats_over_http OK reason phrase

rob05c:
Fix plugin stats_over_http 200 Reason Phrase to "OK" from "Ok."

> A client SHOULD ignore the reason-phrase content

So this is a minor fix. But it is case-sensitive, and I've personally seen clients in the wild break from reason phrases they didn't like.
 
And it was making me twitch.

References:
https://datatracker.ietf.org/doc/html/rfc7230#section-3.1.2
https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.1
https://datatracker.ietf.org/doc/html/rfc9110#section-15.3.1
https://datatracker.ietf.org/doc/html/rfc9112#section-4


randall (dismissed):
LGTM

Any reason this shouldn't go to 9.2?

rob05c:
> Any reason this shouldn't go to 9.2?
It's probably technically incompatible. Do we backport incompatible bug fixes?

bneradt:
In case it's helpful, it looks like at least one gold file needs to be updated for the new case:

https://ci.trafficserver.apache.org/job/Github_Builds/job/autest/1102/consoleText

brbzull0:
@rob05c Should this be merge to `10-Dev` or `master`? Asking as it was assigned to `10-Dev` project

rob05c:
> @rob05c Is this meant to be merge into master or 10-Dev? Asking as it is assigned to 10-Dev project.

I'm not sure. I thought all new things were supposed to go into `10-Dev`.

I don't really understand the difference in `10-Dev` and `master`? This is pretty minor, it's not like a feature we need in 10 or anything. But it'd be nice if it didn't have to wait for 11.

rob05c:
> My understanding is that all Incompatibles changes are supposed to land in 10-Dev branch which will eventually become 10.

Ah, thanks!

I asked around and got an answer to my other question too: we do generally backport incompatible bug fixes. This is also a very minor incompatibility. So I put it in 9.2.

randall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8906, opened 2022-06-13
Fix autest uses of File exists parameter

bneradt:
The AuTest File exists parameter can be configured in one of three ways:

1. If True, then AuTest will fail unless the file exists.
2. If False, then AuTest will fail unless the file does not exist.
3. If None (the default) then no existence check is done.

Many of our tests configured exists=False assuming it would behave like
the default None configuration above. That is, they used exists=False to
turn off verification. That is in fact how AuTest has behaved until now,
but that is a bug in the AuTest framework. The next release of AuTest
will have a fix such that exists=False will now verify that a file does
not in fact exist which will cause some of our tests to fail. This patch
anticipates that fix to AuTest and fixes our use of exists accordingly.


randall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8899, opened 2022-06-09
Do not modify Transfer-Encoding header on retry

masaori335:
Fix #8801.

Issue #8801, opened 2022-04-20
Transfer-Encoding: chunked, chunked, chunked, chunked


masaori335:
When ATS retries the chunked requests, the value of the `Transfer-Encoding` header gets odd. 
It becomes `Transfer-Encoding: chunked, chunked, chunked, chunked` at 3rd retry.

# Repro
##  ATS
- remap.config
```
# force to retry. nothing is running on port 9999
map / http://127.0.0.1:9999/ 
```

- records.config
```
# default setting
CONFIG proxy.config.http.connect_attempts_max_retries INT 3
```

## Client
```
» curl --http1.1 -sk http://127.0.0.1:8080/ -o /dev/null -H 'Transfer-Encoding: chunked' -d 'hello: world'
```

## Debug log
```
+++++++++ Proxy's Request after hooks +++++++++
-- State Machine Id: 0
POST / HTTP/1.1
Host: localhost:9999
User-Agent: curl/7.79.1
Accept: */*
Content-Type: application/x-www-form-urlencoded
Client-ip: 127.0.0.1
X-Forwarded-For: 127.0.0.1
Via: http/1.1 traffic_server[7a7d1c9b-cbcf-4e17-9664-db213a704bc4] (ApacheTrafficServer/10.0.0)
Transfer-Encoding: chunked
...
[Apr 20 10:57:38.798] [ET_NET 8] DEBUG: <HttpTransact.cc:3806 (handle_response_from_server)> (http_trans) [0] max_connect_retries: 3 s->current.attempts: 0
...
+++++++++ Proxy's Request after hooks +++++++++
-- State Machine Id: 0
POST / HTTP/1.1
Host: localhost:9999
User-Agent: curl/7.79.1
Accept: */*
Content-Type: application/x-www-form-urlencoded
Client-ip: 127.0.0.1
X-Forwarded-For: 127.0.0.1
Via: http/1.1 traffic_server[7a7d1c9b-cbcf-4e17-9664-db213a704bc4] (ApacheTrafficServer/10.0.0)
Transfer-Encoding: chunked, chunked
...
[Apr 20 10:57:38.799] [ET_NET 8] DEBUG: <HttpTransact.cc:3806 (handle_response_from_server)> (http_trans) [0] max_connect_retries: 3 s->current.attempts: 1
...
+++++++++ Proxy's Request after hooks +++++++++
-- State Machine Id: 0
POST / HTTP/1.1
Host: localhost:9999
User-Agent: curl/7.79.1
[...]

PR #8908, opened 2022-06-14
Add autest for connect_attempts_max_retries

masaori335:
Basic AuTest of [proxy.config.http.connect_attempts_max_retries](https://docs.trafficserver.apache.org/admin-guide/files/records.config.en.html#proxy-config-http-connect-attempts-max-retries) for #8801. ( #8899 fixes the issue )


masaori335:
[approve ci autest]

masaori335:
This only works on macOS ( and probably BSD ). We need to find a way to test the behavior of `connect_attempts_max_retries` on linux.

PR #8911, opened 2022-06-14
Override LogField constructor to avoid reinterpret_cast at call site

cmcfarlen:
Alternative to #8900

This adds a LogField constructor override that accepts the `unmarshal` function with the additional slice parameter and handles the reinterpret cast.  This lets the type system handle the decision to cast and removes that requirement of the caller.


ywkaras (approved):

zwoop:
This isn't necessary for 9.2.x, for now at least. If we start getting merge conflicts, we will revisit.

PR #8901, opened 2022-06-09
uri_signing plugin: Fix missing payload validation for the iss field.

brbzull0:
Should fix the crash for https://github.com/apache/trafficserver/issues/8874

Also adding some error messages that would eventually help for debugging.

PR #8918, opened 2022-06-17
uri_signing/parse: check for null iss when kid == NULL

traeak:
In the uri_signing plugin add a check for non null iss in the case where kid == NULL. Previous PR added the check for then kid != NULL


bneradt (approved):

traeak:
[approve ci autest]

plugins/experimental/uri_signing/parse.c
```diff
@@ -222,6 +222,10 @@ validate_jws(cjose_jws_t *jws, struct config *cfg, const char *uri, size_t uri_c
     }
     TimerDebug("checking crypto signature for jwt");
   } else {
+    if (!jwt->iss) {
```
brbzull0, line 225: maybe I am missing something:

I guess it does not harm but, If  the `jwt_validate` [up](https://github.com/apache/trafficserver/blob/8ca74ee27cc14f27a1480a88fc453fbabc954e4e/plugins/experimental/uri_signing/parse.c#L195) above fails because `iss`  is null then we should never get this far. Previous [PR](https://github.com/apache/trafficserver/pull/8901/) added a validation for this field in which case it should prevent us getting this far.

brbzull0:
[approve ci autest]

plugins/experimental/uri_signing/parse.c
traeak: You could be correct.  This path was core dumping with the autest after applying some internal patches.  I can dig a bit to see if I can move the check up higher.

brbzull0: I think if you chery-pick [this](https://github.com/apache/trafficserver/pull/8901/) change then this will not core dump at least because of the missing `iss`, it will instead fail the validation.

traeak: The jwt_validate should take care of this in the normal code.  I'll need to further review our internal patch.

traeak:
The current logic should make it so this is never hit.

zwoop:
Remember to remove projects and milestone when closing without merging.

PR #8924, opened 2022-06-23
Update location for core rule set

shukitchan:
The location of OWASP core rule set has been changed a while back.
Thus updating the documentation to reflect them.

Also find that I miss a rule that I want to remove since the lua example here does not support examining response body yet.

### DIFF
--- a/example/plugins/lua-api/modsecurity/README.md
+++ b/example/plugins/lua-api/modsecurity/README.md
@@ -1,7 +1,7 @@
 Integrating ATS with ModSecurity V3 using LuaJIT and FFI
 ====
 
-Opensource WAF for [Apache Traffic Server](http://trafficserver.apache.org/).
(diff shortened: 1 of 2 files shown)


example/plugins/lua-api/modsecurity/README.md
```diff
@@ -30,7 +30,7 @@ Contents/Rules inside example.conf
 
 Working with CRS
 ====
- - Go to [here](https://github.com/SpiderLabs/owasp-modsecurity-crs) and get release v3.2.0
+ - Go [here](https://github.com/coreruleset/coresuleset) and download release v3.3.2
```
randall, line 33: is this the right repo URL? I get a 404 visiting that.

shukitchan: fixed the typo

randall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8925, opened 2022-06-24
Cleanup. Remove unnecessary use of a memory arena when logging.

brbzull0:
Not very clear to me why it's being allocated into the arena, It is not necessary anyway, the data is already allocated.


maskit (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8926, opened 2022-06-24
Allows errors from plugin initialization to bubble up

randall:
Prior to this change, if plugin initialization fails, the error message from plugin init
would get cleared when trying to unload the DSO.


zwoop:
Cherry-picked to v9.2.x

PR #8931, opened 2022-06-29
Fix clang-format installation with multiple threads

bneradt:
The initial run of the clang-format target with multiple threads would
invoke multiple instances of tools/clang-format.sh at the same time.
This would result in multiple processes trying to install the
clang-format binary into .git/fmt, leading to a failure with a a "Text
file busy" error. Subsequent runs would succeed, however, because the
binary had already been installed.

This updates the tools/clang-format.sh script to take a `--install`
argument which just installs clang-format to .git/fmt, and makes the
existence of that directory a prerequisite for the clang-format target.
If it doesn't exist, then it is created with a run of
`tools/clang-format.sh --install`. Then the various `clang-format-<dir>`
targets run in parallel without each trying to install the tool.

This also adds, for convenience, a `format` target that runs both
`clang-format` and `autopep8`.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -111,6 +111,7 @@ endif
 rat:
 	java -jar $(top_srcdir)/ci/apache-rat-0.13-SNAPSHOT.jar -E $(top_srcdir)/ci/rat-regex.txt  -d $(top_srcdir)
 
+.PHONY: autopep8
 autopep8:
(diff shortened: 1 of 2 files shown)


masaori335 (approved):
Looks reasonable.

zwoop:
Cherry-picked to v9.2.x

PR #8943, opened 2022-07-07
Fix doc formatting for rate_limit plugin

rob05c:
The missing newlines were causing the doc generator to think the body text was part of the option name. See https://docs.trafficserver.apache.org/en/latest/admin-guide/plugins/rate_limit.en.html#cmdoption-rate-limit-iprep_buckets


randall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8935, opened 2022-06-30
Make clang-format not modify ink_autoconf.h.in and ink_autoconf.h

bneradt:
ink_autoconf.h.in and ink_autoconf.h are generated files created with
autoconf. After they are made, if the user builds with them and then
calls the clang-format target, clang-format will dutifully format these
files according to our .clang-format specification. This then results in
a future invocation of make rebuilding most of the files in the
repository because those file are fundamental dependencies in our
repository.

This patch addresses the issue by simply not calling clang-format on
these generated, not checked in, files.

This fixes #8934

PR #8944, opened 2022-07-07
Make the autopep8 clang-format targets quieter

bneradt:
This makes our clang-format and autopep8 execution quieter. They will
now only print any files that they modified.

---

While working on #8935 I realized that we all would have seen the issue of these files being modified had our clang-format output been quieter. This addresses that by only printing the files modified by clang-format and autopep8. I've tested this both on MacOS and Linux.

Here's example output with three files being formatted by the `clang-format` and `autopep8` targets:

```
$ make -j8 format
/Library/Developer/CommandLineTools/usr/bin/make clang-format-src clang-format-example clang-format-iocore clang-format-lib clang-format-mgmt clang-format-plugins clang-format-proxy clang-format-tools
./tests/gold_tests/proxy_protocol/proxy_serve_stale.test.py
./plugins/experimental/metalink/test/http09
./proxy/hdrs/URL.cc
```

### DIFF
--- a/tools/clang-format.sh
+++ b/tools/clang-format.sh
@@ -87,7 +87,17 @@ EOF
       exit 1
   else
       [ ${just_install} -eq 1 ] && return
-      for file in $(find $DIR -iname \*.[ch] -o -iname \*.cc -o -iname \*.h.in); do
+
+      # Efficiently retrieving modification timestamps in a platform
+      # independent way is challenging. We use find's -newer argument, which
+      # seems to be broadly supported. The following file is created and has a
+      # timestamp just before running clang-format. Any file with a timestamp
+      # after this we assume was modified by clang-format.
+      start_time_file=$(mktemp -t clang-format-start-time.XXXXXXXXXX)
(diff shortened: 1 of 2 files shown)


tools/clang-format.sh
```diff
@@ -87,17 +87,29 @@ EOF
...
   else
       [ ${just_install} -eq 1 ] && return
-      for file in $(find $DIR -iname \*.[ch] -o -iname \*.cc -o -iname \*.h.in); do
+
+      # Efficiently retrieving modification timestamps in a platform
+      # independent way is challenging. We use find's -newer argument, which
+      # seems to be broadly supported. The following file is created and has a
+      # timestamp just before running clang-format. Any file with a timestamp
+      # after this we assume was modified by clang-format.
+      start_time_file=/tmp/format_start.$$
```
randall, line 96: This could be insecure. Probably better to use mktemp (like the autopep does)

bneradt:
[approve ci autest]

randall (approved):

PR #8949, opened 2022-07-08
Do NOT setup internal transfer if ua_entry is nullptr

### LINKED ISSUE #8948: Crash on HttpTunnel::producer_run
```
[ 00 ] libpthread-2.17.so           waitpid                                                          
[ 01 ] traffic_server               crash_logger_invoke(int, siginfo_t*, void*)                      ( Crash.cc:168 )
[ 02 ] libpthread-2.17.so                                                                            
[ 03 ] traffic_server               HttpTunnel::producer_run(HttpTunnelProducer*)                    ( HttpTunnel.cc:988 )
[ 04 ] traffic_server               HttpTunnel::tunnel_run(HttpTunnelProducer*)                      ( HttpTunnel.cc )
[ 05 ] traffic_server               HttpSM::setup_internal_transfer(int (HttpSM::*)(int, void*))     ( HttpSM.cc:6714 )
[ 06 ] traffic_server               HttpSM::state_api_callout(int, void*)                            ( HttpSM.cc )
[ 07 ] traffic_server               HttpSM::state_api_callback(int, void*)                           ( HttpSM.cc:1466 )
[ 08 ] traffic_server               TSHttpTxnReenable                                                ( InkAPI.cc:6304 )
```

We're facing this crash with 9.2.x.

masaori335:
Fix #8948.

It looks like HttpSM don't need to setup HttpTunnel if `ua_entry` or `ua_entry->vc` is `nullptr`

### DIFF
--- a/proxy/http/HttpSM.cc
+++ b/proxy/http/HttpSM.cc
@@ -6673,18 +6673,22 @@ HttpSM::setup_internal_transfer(HttpSMHandler handler_arg)
 
   HTTP_SM_SET_DEFAULT_HANDLER(handler_arg);
 
-  // Clear the decks before we setup the new producers
-  // As things stand, we cannot have two static producers operating at
-  // once
-  tunnel.reset();
+  if (ua_entry && ua_entry->vc) {
+    // Clear the decks before we setup the new producers
+    // As things stand, we cannot have two static producers operating at
+    // once
+    tunnel.reset();
 
-  // Setup the tunnel to the client
-  HttpTunnelProducer *p =
-    tunnel.add_producer(HTTP_TUNNEL_STATIC_PRODUCER, nbytes, buf_start, (HttpProducerHandler) nullptr, HT_STATIC, "internal msg");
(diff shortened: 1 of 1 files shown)


proxy/http/HttpSM.cc
```diff
@@ -6597,6 +6597,11 @@ HttpSM::setup_error_transfer()
 void
 HttpSM::setup_internal_transfer(HttpSMHandler handler_arg)
 {
+  if (ua_entry == nullptr || ua_entry->vc == nullptr) {
```
maskit, line 6600: Don't we need to call `HTTP_SM_SET_DEFAULT_HANDLER(handler_arg)` even if a client is already gone?
https://github.com/apache/trafficserver/blob/15bea4dd946c8cb6fc2000a4b31cf4f2f261b29d/proxy/http/HttpSM.cc#L1769-L1771

masaori335: Ah, in that case, it looks like `HttpSM::setup_server_transfer_to_cache_only();` should be called eventually.
OK, I'll go with another approach.

maskit: Yeah, that's what I thought.

bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8950, opened 2022-07-08
.fit/fmt/.clang-format-installed prerequisite

bneradt:
Before this patch, .git/fmt was used as the prerequisite to determine
whether clang-format was installed. But this directory was also created
by the autopep8 target, so if someone ran autopep8 before clang-format,
Make would think that clang-format was installed already. This adds a
hidden sentinel file to the clang-format install process that can be
used as a specific target in our clang-format Makefile.am.

---

## For Review

Use the ignore whitespace option to simplify the diff. I had to change indentation.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -133,11 +133,12 @@ CLANG_FORMAT_DIR_TARGETS = \
 
 .PHONY:	$(CLANG_FORMAT_DIR_TARGETS)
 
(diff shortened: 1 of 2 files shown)


randall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8954, opened 2022-07-12
Use std::unique_ptr for X509 and BIO scoped heap objects.

ywkaras:


maskit (approved):
👍

zwoop:
Cherry-picked to v9.2.x

PR #8962, opened 2022-07-13
Revert "File change monitoring on s3_auth (#8905)"

bneradt:
This reverts commit e93063e26172eb9f7ed92ff9627b0a807479ba59.

PR #8905 was accidentally filed against 9.2.x instead of master.
@moonchen filed a new PR for master via #8960, so the work will continue
with that PR. Since this patch is not intended for 9.2.x, this PR
reverts the commit.


bneradt:
@zwoop took care of reverting this.

PR #7473, opened 2021-02-02
Adjust so transfer-encoding header can be treated hop-by-hop

shinrich:
This is additional fixes to PR #6908 proposed by @ema.  That PR was good, but the http2 test was failing.  While working through the HTTP/2 logic, I noticed that there are places where we are making assumptions about the presence of the Transfer-Encoding header and not reapplying it where appropriate.

This PR makes some additional changes in the HttpSM to more correctly identify transactions with bodies and places the Transfer-Encoding header appropriately.

I also had to make a few small changes to some test output expectations since the Transfer-Encoding header is not blindly copied along.

This closes issue #6907

PR #8963, opened 2022-07-14
Http2 to origin

bneradt:
This implements the HTTP/2 to origin feature.


maskit:
Since H2 to Origin has been unmarged for very long time, I'm slightly lowering a bar for merging this.

I still think issue https://github.com/apache/trafficserver/issues/5230 should be fixed eventually (not just for performance but also privacy), but I'm going to let you off the hook. I actually have an incomplete patch (which works for simple use case at minimum) to pass an `HttpHdr` for request to `HttpSM` without marshaling and unmarshaling. Maybe we can add `sensitive` flag later to keep HPACK header representation. For now, I'd like to see some warning or note on the doc that says header representation won't be kept and sensitive headers can be stored into HPACK dynamic table.

doc/admin-guide/files/records.config.en.rst
```diff
@@ -3956,6 +3956,13 @@ Client-Related Configuration
 
    Enables (``1``) or disables (``0``) TLSv1_3 in the ATS client context. If not specified, enabled by default
 
+.. ts:cv:: CONFIG proxy.config.ssl.client.alpn_protocol STRING ""
```
maskit, line 3959: I pointed out the same thing on the previous PR, but I'm not sure if users  want to set "ALPN string".

The ALPN protocol IDs are inconsistent; The ID for HTTP/1.1 is "http/1.1" but not "h1.1" where ones for HTTP/2 and 3 are "h2" and "h3". I don't think these IDs are user friendly.

Even if we use those raw IDs, the doc should say what IDs are available. Obviously we can't set "spdy/1", although it's a registered valid ID. We shouldn't send "h3" on a regular TLS connection, and similarly we shouldn't send "h2" on a QUIC connection.

Also, the order of IDs is unclear if a user is unfamiliar with ALPN.

I don't have a good idea right now, but I think we need discussion about this setting at minimum. Changing this setting would be an incompatible change and we wouldn't be able to change it until 11.0 once we make 10.0 release with this setting.

maskit: I just remembered I made this comment as well on the previous PR.

> Let's separate out the outbound ALPN part. We still need to discuss how to enable protocols, but the feature itself should work without H2 (e.g. use http/1.1 but not 1.0), and it should be testable.

proxy/http2/Http2Stream.h
```diff
@@ -164,17 +196,22 @@ class Http2Stream : public ProxyTransaction
...
 
-  bool trailing_header = false;
-  bool has_body        = false;
+  bool _possible_trailing_header = false;
+  bool _expect_send_trailer      = false;
+  bool _expect_receive_trailer   = false;
+
+  bool has_body = false;
+
+  bool _outbound_flag     = false;
```
maskit, line 213: This name is confusing a little, because a stream can be initiated by a server on an inbound connection (e.g. Server Push). Since a stream (ProxyTransaction) knows its owner session, we should be able to ask the session whether this stream is on an inbound session or an outbound session.

I'd remove this variable, but please rename it accordingly if you want to keep it.

proxy/http2/HTTP2.h
```diff
@@ -247,8 +252,8 @@ enum Http2SettingsIdentifier {
   HTTP2_SETTINGS_INITIAL_WINDOW_SIZE    = 4,
   HTTP2_SETTINGS_MAX_FRAME_SIZE         = 5,
   HTTP2_SETTINGS_MAX_HEADER_LIST_SIZE   = 6,
-
-  HTTP2_SETTINGS_MAX
+  HTTP2_SETTINGS_MAX, // Really just the max of the "densely numbered" core id's
+  HTTP2_SETTINGS_GRPC_ALLOW_TRUE_BINARY_METADATA = 0xfe03,
```
maskit, line 256: This doesn't seem like to be used.

proxy/http2/Http2ConnectionState.cc
```diff
@@ -1545,10 +1739,14 @@ Http2ConnectionState::release_stream()
         // Can't do this because we just destroyed right here ^,
         // or we can use a local variable to do it.
         // session = nullptr;
+      } else if (shutdown_state == HTTP2_SHUTDOWN_IN_PROGRESS && fini_event == nullptr) {
```
maskit, line 1742: This is concerning because it reverts a fix.
See https://github.com/apache/trafficserver/pull/7267

proxy/http2/Http2ConnectionState.cc
```diff
@@ -1545,10 +1739,14 @@ Http2ConnectionState::release_stream()
...
+        session->do_clear_session_active();
+        fini_event = this_ethread()->schedule_imm_local((Continuation *)this, HTTP2_SESSION_EVENT_FINI);
       } else if (session->get_proxy_session()->is_active()) {
         // If the number of clients is 0, HTTP2_SESSION_EVENT_FINI is not received or sent, and session is active,
         // then mark the connection as inactive
         session->do_clear_session_active();
+        session->set_no_activity_timeout();
```
maskit, line 1749: This should be safe, but I'm wondering if inactivity timeout currently works for incoming H2 sessions... Even if it works, the timeout might be based on another setting.

proxy/logging/LogAccess.cc
```diff
@@ -1412,7 +1412,7 @@ LogAccess::marshal_client_provided_cert(char *buf)
 {
   int provided_cert = 0;
   if (m_http_sm) {
-    auto txn = m_http_sm->get_ua_txn();
+    auto txn = m_http_sm->ua_txn;
```
maskit, line 1415: I don't think this is an intentional change, but let's use the getter.

proxy/hdrs/HdrToken.cc
```diff
@@ -230,7 +230,10 @@ static HdrTokenFieldInfo _hdrtoken_strs_field_initializers[] = {
   {"Strict-Transport-Security", MIME_SLOTID_NONE, MIME_PRESENCE_NONE, (HTIF_MULTVALS)},
   {"Subject", MIME_SLOTID_NONE, MIME_PRESENCE_SUBJECT, HTIF_NONE},
   {"Summary", MIME_SLOTID_NONE, MIME_PRESENCE_SUMMARY, HTIF_NONE},
-  {"TE", MIME_SLOTID_TE, MIME_PRESENCE_TE, (HTIF_COMMAS | HTIF_MULTVALS | HTIF_HOPBYHOP)},
+  // Need to figure out why this cannot be handled as hop by hop.  If it is hop-by-hop
+  // the information is not propagated for gRPC
+  //{"TE", MIME_SLOTID_TE, MIME_PRESENCE_TE, (HTIF_COMMAS | HTIF_MULTVALS | HTIF_HOPBYHOP)},
+  {"TE", MIME_SLOTID_TE, MIME_PRESENCE_TE, (HTIF_COMMAS | HTIF_MULTVALS)},
```
maskit, line 236: This partially(?) reverts #7473, and it doesn't seem like a right direction. The question should be figured out before merge.

proxy/http/HttpProxyServerMain.cc
```diff
@@ -50,6 +50,9 @@
 
 HttpSessionAccept *plugin_http_accept             = nullptr;
 HttpSessionAccept *plugin_http_transparent_accept = nullptr;
+extern std::function<PoolableSession *()> Create_h1_server_session;
```
maskit, line 53: Why don't you make the first lowercase? These are not types but variables.

proxy/http/HttpSM.cc
```diff
@@ -3405,6 +3657,66 @@ HttpSM::tunnel_handler_ua(int event, HttpTunnelConsumer *c)
   return 0;
 }
 
+int
+HttpSM::tunnel_handler_trailer_ua(int event, HttpTunnelConsumer *c)
```
maskit, line 3661: Is it possible to add support for trailer before or after H2 to Origin separately? I guess it was added for gRPC, but I assume it'll be used for H1 as well. And if we support trailer header, that should be explicitly stated on release note.

proxy/http/HttpTransact.h
```diff
@@ -571,12 +573,64 @@ class HttpTransact
     ConnectionAttributes *server                 = nullptr;
     ink_time_t now                               = 0;
     ServerState_t state                          = STATE_UNDEFINED;
-    unsigned attempts                            = 0;
-    unsigned simple_retry_attempts               = 0;
-    unsigned unavailable_server_retry_attempts   = 0;
-    ParentRetry_t retry_type                     = PARENT_RETRY_NONE;
+    class Attempts
```
maskit, line 576: Would you make this change separately?

proxy/http2/Http2ConnectionState.cc
```diff
@@ -1756,16 +1980,26 @@ Http2ConnectionState::send_headers_frame(Http2Stream *stream)
...
+
+  HTTPHdr *send_hdr = stream->get_send_header();
+  if (stream->expect_send_trailer()) {
+    // Which is a no-op conversion
+  } else {
+    http2_convert_header_from_1_1_to_2(send_hdr);
+  }
 
-  uint32_t buf_len = resp_hdr->length_get() * 2; // Make it double just in case
+  uint32_t buf_len = send_hdr->length_get() * 2; // Make it double just in case
```
maskit, line 1996: Would make another PR for this kind of name changes (e.g. send_hdr, peer_settings, local_settings, etc.)? That would reduce the volume of this PR (which is great for reviewers so they can focus on the logical changes for H2 to Origin), and would also ease maintaining this branch.

proxy/ProxySession.cc
```diff
@@ -313,3 +315,21 @@ ProxySession::support_sni() const
 {
   return _vc ? _vc->support_sni() : false;
 }
+
+PoolableSession *
+ProxySession::protocol_creation(NetVConnection *netvc)
+{
+  // Figure out what protocol was negotiated
+  int proto_index             = SessionProtocolNameRegistry::INVALID;
+  SSLNetVConnection *sslnetvc = dynamic_cast<SSLNetVConnection *>(netvc);
```
maskit, line 324: Let's not depend on SSLNetVC. `ALPNSupport` should be enough here.

maskit: Actually, I don't see any reason to receive a NetVC here. Just a protocol id is enough. Let's deal with NetVC on HttpSM side.

How about this? I want function names to have verbs.
```
PoolableSession *
ProxySession::create_outbound_sesion(int protocol_id) {
  auto iter = ProtocolSessionCreateMap.find(protocol_id);
   ink_release_assert(iter != ProtocolSessionCreateMap.end());
   return iter->second();
}
```

proxy/http2/Http2ServerSession.cc
```diff
@@ -0,0 +1,420 @@
...
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+
+#include "Http2ServerSession.h"
+#include "HttpDebugNames.h"
+#include "tscore/ink_base64.h"
+#include "Http2CommonSessionInternal.h"
+#include "HttpSessionManager.h"
+#include "P_SSLNetVConnection.h"
```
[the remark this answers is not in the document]
maskit: But this session behaves as a client and `TLSEarlyDataSupport` only supports server side at the moment. We can simply get rid of it for now.

proxy/http2/Http2ServerSession.cc
```diff
@@ -0,0 +1,420 @@
...
+  this->_vc    = new_vc;
+  _vc->set_inactivity_timeout(HRTIME_SECONDS(Http2::accept_no_activity_timeout));
+  this->schedule_event = nullptr;
+  this->mutex          = new_vc->mutex;
+
+  this->connection_state.mutex = this->mutex;
+
+  TLSEarlyDataSupport *eds = dynamic_cast<TLSEarlyDataSupport *>(new_vc);
+  if (eds != nullptr) {
+    this->read_from_early_data = eds->get_early_data_len();
```
maskit, line 116: A client never receives early data. We don't need this block.

proxy/http2/Http2ConnectionState.cc
```diff
@@ -1278,8 +1333,115 @@ Http2ConnectionState::state_closed(int event, void *edata)
...
+{
+  if (stream->get_transaction_id() < 0) {
+    Http2StreamId stream_id = (latest_streamid_in == 0) ? 3 : latest_streamid_in + 2;
+    stream->set_transaction_id(stream_id);
+    latest_streamid_in = stream_id;
+  }
+}
+
+Http2Stream *
+Http2ConnectionState::create_initiating_stream(bool client_streamid, Http2Error &error)
```
maskit, line 1359: It seems like `client_streamid` is always `true`, and `error` is not used on the caller side.

proxy/http2/Http2CommonSession.cc
```diff
@@ -172,6 +171,7 @@ void
 Http2CommonSession::flush()
 {
   if (this->_pending_sending_data_size > 0) {
+    total_write_len += this->_pending_sending_data_size;
```
maskit, line 174: `total_write_len` doesn't seem to be used.

proxy/http/HttpTunnel.cc
```diff
@@ -984,11 +986,14 @@ HttpTunnel::producer_run(HttpTunnelProducer *p)
           p->handler_state = HTTP_SM_POST_SUCCESS;
         }
       }
+      Debug("http_tunnel", "Start write vio %ld bytes", c_write);
       // Start the writes now that we know we will consume all the initial data
       c->write_vio = c->vc->do_io_write(this, c_write, c->buffer_reader);
       ink_assert(c_write > 0);
       if (c->write_vio == nullptr) {
         consumer_handler(VC_EVENT_ERROR, c);
+      } else if (c->write_vio->ntodo() == 0 && c->alive) {
```
maskit, line 995: Not just this line, but I don't get why these changes to HttpTunnel is needed for H2 to Origin. I'm wondering (and concerning) if these affect H1 and client transactions.

iocore/net/ALPNSupport.cc
```diff
@@ -144,3 +144,36 @@ ALPNSupport::registerNextProtocolSet(SSLNextProtocolSet *s, const SessionProtoco
   this->npnSet       = s;
   npnSet->create_npn_advertisement(protoenabled, &npn, &npnsz);
 }
+
+bool
+ALPNSupport::process_alpn_protocols(const std::string_view protocols, unsigned char *client_alpn_protocols, int &alpn_array_len)
```
maskit, line 149: It would be nice if you could put this into some module that parses setting values. Maybe RecHttp.cc, which parses `server_ports`.

bneradt:
I appreciate the thoughtful review, @maskit. And thank you, too, for repeating some of your comments from the previous PR. I'm working through your observations now and will spin off other PRs per your suggestions.

proxy/http/HttpTransact.h
maskit: I just realized this change is similar to #8776. I guess merging it helps minimizing this PR? @ywkaras

ywkaras: This change is already in Yahoo TS.  I think this somehow just got picked up.

doc/admin-guide/files/records.config.en.rst
```diff
@@ -4096,7 +4103,17 @@ HTTP/2 Configuration
    :reloadable:
    :units: bytes
 
-   The initial window size for inbound connections.
+   The initial window size for inbound connection streams.
+
+.. ts:cv:: CONFIG proxy.config.http2.session_initial_window_size_in INT 0
```
maskit, line 4108: This is #8203, right? This is obviously for incoming connections and should be merged separately. I was -0 on the PR. Take a look at the discussion on #8199 as well.

mgmt/RecordsConfig.cc
```diff
@@ -1354,6 +1358,8 @@ static const RecordElement RecordsConfig[] =
   ,
   {RECT_CONFIG, "proxy.config.http2.no_activity_timeout_in", RECD_INT, "120", RECU_DYNAMIC, RR_NULL, RECC_STR, "^[0-9]+$", RECA_NULL}
   ,
+  {RECT_CONFIG, "proxy.config.http2.no_activity_timeout_out", RECD_INT, "120", RECU_DYNAMIC, RR_NULL, RECC_STR, "^[0-9]+$", RECA_NULL}
```
maskit, line 1361: Need documentation.

proxy/http2/HTTP2.cc
```diff
@@ -599,6 +610,9 @@ http2_convert_header_from_1_1_to_2(HTTPHdr *headers)
       } else {
         field->value_set(headers->m_heap, headers->m_mime, value, value_len);
       }
+      // Remove the host header field, redundant to the authority field
+      // For istio/envoy, having both was causing 404 responses
```
maskit, line 614: This is interesting, because RFC 9113 says:

> An intermediary that forwards a request over HTTP/2 MAY retain any
      Host header field.

Removing `Host` header should be fine, but I wonder why it causes 404. It could be because of mismatch between the generated `:authority` pseudo header and the retained `Host` header.

The RFC also says:

> Clients MUST NOT generate a request with a Host header field that
      differs from the ":authority" pseudo-header field.  A server
      SHOULD treat a request as malformed if it contains a Host header
      field that identifies an entity that differs from the entity in
      the ":authority" pseudo-header field.

> An intermediary that forwards a request over HTTP/2 MUST construct
      an ":authority" pseudo-header field using the authority
      information from the control data of the original request, unless
      the original request's target URI does not contain authority
      information (in which case it MUST NOT generate ":authority").
      Note that the Host header field is not the sole source of this
      information; see Section 7.2 of [[HTTP](https://datatracker.ietf.org/doc/html/rfc9113#ref-HTTP)].

and ATS does gather the information from the request line if available. We might want to check generated `:authority` is really appropriate.

I don't remember why we didn't remove `Host` header. There may be no reason, but it might be because removing it is unnecessary/optional.

proxy/http2/HTTP2.cc
```diff
@@ -446,10 +451,16 @@ http2_convert_header_from_2_to_1_1(HTTPHdr *headers)
     if (MIMEField *field = headers->field_find(HTTP2_VALUE_AUTHORITY, HTTP2_LEN_AUTHORITY);
         field != nullptr && field->value_is_valid()) {
       int authority_len;
-      const char *authority = field->value_get(&authority_len);
 
+      // Set the host header field
+      MIMEField *host = headers->field_find(MIME_FIELD_HOST, MIME_LEN_HOST);
+      if (host == nullptr) {
+        host = headers->field_create(MIME_FIELD_HOST, MIME_LEN_HOST);
+        headers->field_attach(host);
```
maskit, line 459: Okay, `Host` header was unnecessary for ATS internal process if the request URL is absolute form, but adding it makes sense.

This block is fine at all and no change is needed, but I realized one thing while I'm checking RFC 9110. According to the RFC, `Host` header should be at the beginning of a header field list.

> A user agent that sends Host SHOULD send it as the first
   field in the header section of a request.

That makes sense and we do that for pseudo headers though, that's not easy to do with our `MIMEHdr` structure, because it doesn't allow us to insert/prepend headers.

I also checked RFC 9113 (HTTP/2) and it says:

> All pseudo-header fields MUST appear in a field block before all
   regular field lines.  Any request or response that contains a pseudo-
   header field that appears in a field block after a regular field line
   MUST be treated as malformed ([Section 8.1.1](https://datatracker.ietf.org/doc/html/rfc9113#section-8.1.1)).

So it's MUST for pseudo headers, but it's SHOULD for `Host` header. It would be nice if we could have `Host` header at the beginning but I don't think it has to be done on this PR.

proxy/http2/HTTP2.cc
maskit: I found a TODO comment at right after your changes (on line 651). We can remove the comment.

Also, I finished reviewing changes in `http2_convert_header_from_2_to_1_1` and `http2_convert_header_from_1_1_to_2`. No additional change is needed except the TODO comment I mentioned above. If you make a separate PR for these changes, I can approve it immediately.

proxy/http2/Http2ServerSession.cc
```diff
@@ -0,0 +1,420 @@
...
+
+  if (!this->is_draining() && this->connection_state.get_shutdown_reason() == Http2ErrorCode::HTTP2_ERROR_MAX) {
+    this->connection_state.set_shutdown_state(HTTP2_SHUTDOWN_NONE);
+  }
+
+  if (this->connection_state.get_shutdown_state() == HTTP2_SHUTDOWN_NONE) {
+    if (this->is_draining()) { // For a case we already checked Connection header and it didn't exist
+      Http2SsnDebug("Preparing for graceful shutdown because of draining state");
+      this->connection_state.set_shutdown_state(HTTP2_SHUTDOWN_NOT_INITIATED);
+    } /*else if (this->connection_state.get_stream_error_rate() >
```
maskit, line 229: I don't think this part should be completely removed. If ATS is used as a forward proxy, it could be attacked by malicious/misbehaving origin servers.

There's no graceful shutdown mechanism defined on the spec, but we could stop making new requests on the connection that has high error rate to make the connection inactive and close it sooner rather than later.

We could also mark the origin server as a bad H2 server and use H1 on future connections. It should work as fallback mechanism. I think we need this kind of marking in the future anyway to choose a transport protocol (TCP or QUIC) for an origin server.

maskit: But if we are going to say H2 to Origin is experimental support, like QUIC, we can ignore security concerns for now.

proxy/http2/Http2Stream.cc
```diff
@@ -435,23 +477,22 @@ Http2Stream::do_io_close(int /* flags */)
...
 
-    if (_proxy_ssn && this->is_client_state_writeable()) {
-      // Make sure any trailing end of stream frames are sent
-      // We will be removed at send_data_frames or closing connection phase
-      Http2ClientSession *h2_proxy_ssn = static_cast<Http2ClientSession *>(this->_proxy_ssn);
-      SCOPED_MUTEX_LOCK(lock, h2_proxy_ssn->mutex, this_ethread());
-      h2_proxy_ssn->connection_state.send_data_frames(this);
-    }
+    // Adjust state, so we don't process any more data
+    _state = Http2StreamState::HTTP2_STREAM_STATE_CLOSED;
```
maskit, line 490: Not a blocker since there is already one place that does similar thing, but I think we should not change `_state` like this without actual activity on H2 protocol, because we won't be able to know the last state of a stream, which should be CLOSED state on a clean close even without this line.

proxy/http/HttpSM.cc
```diff
@@ -974,19 +973,34 @@ HttpSM::state_watch_for_client_abort(int event, void *data)
...
+      NetVConnection *netvc = ua_txn->get_netvc();
+      /** TODO: we want to put tunnel.has_consumer_besides_client back in here.
+       *
+       * Currently, our HTTP/2 to origin branch does not work with background
+       * fill.  Trying to keep the tunnel around to fill the cache after the
+       * client aborts currently results in a connection leak. This has to be
+       * fixed, but for now we're avoiding this relatively new feature (Spring
+       * 2021) to stop the leak.
+       *
+       * Uncomment this code to reinstate background fill once it is fixed.
```
maskit, line 986: It's been over a year since this TODO comment. Any progress?

proxy/http/ConnectingEntry.cc
```diff
@@ -0,0 +1,157 @@
...
+{
+  Debug("http_connect", "entered inside ConnectingEntry::state_http_server_open");
+
+  switch (event) {
+  case NET_EVENT_OPEN: {
+    _netvc                 = static_cast<NetVConnection *>(data);
+    UnixNetVConnection *vc = static_cast<UnixNetVConnection *>(_netvc);
+    ink_release_assert(_pending_action == nullptr || _pending_action->continuation == vc->get_action()->continuation);
+    _pending_action = nullptr;
+    Debug("http_connect", "ConnectingEntrysetting handler for TCP handshake");
```
maskit, line 47: The message implies code around this line is for TCP connections. Is it true?

I'm not going to say make this work with QUIC connections right now but I'm very sad if QUIC is completely out of your scope and we have to redesign it when we support QUIC to Origin. Wording that is not protocol-specific would help understanding what condition/state is expected.

More specifically, I'm wondering if TLS handshake has to be undone, because QUIC can't do TLS handshake separately.

Anyways, I don't want to make this change with H2-to-Origin at once. Connection pool should work without H2.

proxy/http/HttpSM.cc
bneradt: @duke8253 is currently working on fixing this. Our hope/intention is to fix this as an additional patch on this PR before merging.

proxy/http/HttpTransact.h
bneradt: Sorry for the delay. Yes, I picked this change into HTTP/2 to origin because we had it internally and it fixes a problem we saw in the connect attempts calculation.

bneradt: Just to follow up on this: I just rebased this branch on top of latest 10-Dev. 10-Dev now has Walt's `Attempts` work on it, so that is no longer in this PR.

doc/admin-guide/files/records.config.en.rst
bneradt: This work got separated out into PR https://github.com/apache/trafficserver/pull/8972 and is now merged.

iocore/net/ALPNSupport.cc
bneradt: Yes, this is good feedback. We addressed this in the separate ALPN PR: https://github.com/apache/trafficserver/pull/8972.

proxy/http/HttpSM.cc
bneradt: By the way, CI is currently failing the autest for background_fill precisely because of this issue. So, to be clear, I don't intend to move forward with this PR until we get this feature working.

proxy/http2/Http2Stream.h
bneradt: Yes, good observation. Thank you. I updated the function to query the session whether it is outbound or not.

proxy/http/HttpSM.cc
bneradt: @duke8253 provided a fix for background fill which I have now incorporated into this PR.

proxy/http2/Http2ConnectionState.cc
bneradt: Thanks. Good catch. The merge is pretty large and I suppose I accidentally pulled this in thinking that I wanted the additional code, not the removal of it.

proxy/http/ConnectingEntry.cc
shinrich: That is just a working thing.  Can remove TCP.  Will need to test this in QUIC case.  There is a specific sequency of events that get sent as the TCP/TLS handshake works out.  Will probably be slightly different in the QUIC case.

In the meantime, this should be changed to more neutral language like "connection handshake" instead of "TCP handshake"

proxy/hdrs/HdrToken.cc
shinrich: True.  I'd leave the comment, so we have it when debugging gRPC.  But leave the code along.  There must be another way to deal with the gRPC case. Or perhaps the other changes made for #7473 fixed the gRPC case as well.

proxy/http/ConnectingEntry.cc
shinrich: Noticed that some Debug messages in HttpSM also refer to TCP.  They should also be cleaned up.

proxy/http/HttpSM.cc
shinrich: Could separate it out.  What is present now has not been tested for a while (AFAIK).  I did very basic tests early on with gRPC.

The goal for the support is gRPC which is only needed for HTTP/2.  We will need that support very soon after the base H2 to origin is committed.  

I agree that it would be nice to support for H1 as well, but do we have a use case for trailers with H1?

proxy/http2/Http2ConnectionState.cc
shinrich: I recall running into idle H2 sessions timing out.  Definitely for outbound connections. I don't recall if I saw it for inbound or not. I assume this would be an issue for incoming H2 sessions as well.  But probably more apparent in the case of origin sessions sitting in a reuse pool.

proxy/http2/Http2ServerSession.cc
shinrich: Could add this back in.  As I dimly recall in my testing, the shutdown due to spurious errors seemed high. But could have been due to other reasons.

proxy/hdrs/HdrToken.cc
bneradt: Thank you Susan for the history. I added `HTIF_MULTVALS` back in and added a TODO comment.

proxy/http2/Http2ConnectionState.cc
bneradt: Sounds good: #9084.

doc/admin-guide/files/records.config.en.rst
bneradt: Yes, correct. This PR pulls in that change. I agree that it's a good idea to keep it separate. I'll re-open #8203.

proxy/http/HttpSM.cc
bneradt: Just to wrap up this thread, we can work on trailer/gRPC further in a separate PR. Thank you, @maskit and @shinrich, for clarifying this.

doc/admin-guide/files/records.config.en.rst
bneradt: Since #8203 has conflicts, I opened a new PR wherein I can resolve the conflicts: #9085.

proxy/http2/Http2ConnectionState.cc
```diff
@@ -1275,6 +1331,99 @@ Http2ConnectionState::state_closed(int event, void *edata)
   return 0;
 }
 
+bool
+Http2ConnectionState::is_peer_concurrent_stream_ub() const
+{
+  return client_streams_in_count >= (peer_settings.get(HTTP2_SETTINGS_MAX_CONCURRENT_STREAMS)) * 0.9;
```
maskit, line 1337: I don't understand why these magic thresholds are needed. Although an implementation must not exceed the limit on purpose, it doesn't need to consider a possibility that a limit change is on wire.

> The values in the SETTINGS frame MUST be processed in the order they
   appear, with no other frame processing between values.  Unsupported
   settings MUST be ignored.  Once all values have been processed, the
   recipient MUST immediately emit a SETTINGS frame with the ACK flag
   set.  Upon receiving a SETTINGS frame with the ACK flag set, the
   sender of the altered settings can rely on the values from the oldest
   unacknowledged SETTINGS frame having been applied.

>   If the sender of a SETTINGS frame does not receive an acknowledgment
   within a reasonable amount of time, it MAY issue a connection error
   ([Section 5.4.1](https://datatracker.ietf.org/doc/html/rfc9113#section-5.4.1)) of type SETTINGS_TIMEOUT.  In setting a timeout, some
   allowance needs to be made for processing delays at the peer; a
   timeout that is solely based on the round-trip time between endpoints
   might result in spurious errors.

The RFC also says:
> An endpoint that wishes to reduce the value of
   SETTINGS_MAX_CONCURRENT_STREAMS to a value that is below the current
   number of open streams can either close streams that exceed the new
   value or allow streams to complete.

It's not a something that need to be taken care by stream initiator

maskit:
If we don't support Server Push on origin side for now, we need to indicate it by sending SETTINGS_ENABLE_PUSH:0.

> SETTINGS_ENABLE_PUSH (0x02):  This setting can be used to enable or
      disable server push.  A server MUST NOT send a PUSH_PROMISE frame
      if it receives this parameter set to a value of 0; see
      [Section 8.4](https://datatracker.ietf.org/doc/html/rfc9113#section-8.4).  A client that has both set this parameter to 0 and
      had it acknowledged MUST treat the receipt of a PUSH_PROMISE frame
      as a connection error ([Section 5.4.1](https://datatracker.ietf.org/doc/html/rfc9113#section-5.4.1)) of type PROTOCOL_ERROR.

>  The initial value of SETTINGS_ENABLE_PUSH is 1.  For a client,
      this value indicates that it is willing to receive PUSH_PROMISE
      frames.

bneradt:
> If we don't support Server Push on origin side for now, we need to indicate it by sending SETTINGS_ENABLE_PUSH:0.

Yes, this is a good observation. I'll work on this when #9085 is finished.

bneradt:
> > If we don't support Server Push on origin side for now, we need to indicate it by sending SETTINGS_ENABLE_PUSH:0.
> 
> Yes, this is a good observation. I'll work on this when #9085 is finished.

Updated to set `SETTINGS_ENABLE_PUSH` to 0. Good observation.

bneradt:
[approve ci autest]

bneradt:
[approve ci autest]

bneradt:
The `http2_to_origin` AuTest will fail until the new Proxy Verifier `await` directive is merged in:
https://github.com/yahoo/proxy-verifier/pull/228

maskit:
@bneradt Is it possible to separate out the change for ConnectionPool and ConnectingEntry?

Since this PR is for H2 to Origin, the code on this PR should not do anything bad unless H2 to Origin is enabled by having `h2`  in `proxy.config.ssl.client.alpn_protocols`. I understand nobody can guarantee that, but ConnectionPool and ConnectingEntry are new and those may affect H1 because the code is used for the both protocols.

As you can see, people who are unfamiliar with H2 haven't been reviewing this PR at all, but the change can bother them. I think you should gather feedback from them as well. Otherwise, we would be blamed like "H2 to Origin made this mess where we don't need it!".

bneradt:
> The `http2_to_origin` AuTest will fail until the new Proxy Verifier `await` directive is merged in:
> [yahoo/proxy-verifier#228](https://github.com/yahoo/proxy-verifier/pull/228)

https://github.com/apache/trafficserver/pull/9221

bneradt:
> As you can see, people who are unfamiliar with H2 haven't been reviewing this PR at all, but the change can bother them. I think you should gather feedback from them as well. Otherwise, we would be blamed like "H2 to Origin made this mess where we don't need it!".

Just to follow up on this, after mentioning this suggestion in the PR/issue call this Monday, it was suggested that @keesspoelstra would be a good candidate to review this PR as well. He mentioned that he already has been following it and would spend some time looking at that patch.

Thank you again, @maskit , for the time you have spent reviewing this.

bneradt:
> This is an approval for changes around H2.
> 
> I'm not sure about connection management part. I still think we should make the change separately, but if that's not an option, please wait for another approval from someone else.

Thank you @maskit . Again, I appreciate all the feedback you've provided on this feature.

@keesspoelstra is carefully reviewing and testing this in the meantime too. I'll wait upon at least his review.

bneradt:
Going to close this and re-open on master now that 10-Dev is master.

PR #8965, opened 2022-07-20
Proxy Verifier: Update to version 2.4.1

bneradt:
This updates the Proxy Verifier version used by our AuTests to version
v2.4.1. This Proxy Verifier version prints the ALPN used by the proxy
(ATS in our case) to the server. This will allow us to verify ALPN
behavior for an upcoming PR for HTTP/2 to origin.


maskit (approved):

zwoop:
Cherry-picked to v9.2.x

Issue #8951, opened 2022-07-09
Request body isn't read after a cache hit for a conditional GET


mrdg:
When a client sends a conditonal GET request with a body and the server responds with a cache hit, the request body isn't drained. This results in the next request on that connection getting a 400 response. It's arguably weird to send a conditional GET with a body, but we're seeing requests like this pretty regularly. It's especially problematic with a proxy in front of ATS, because some other client might see the 400 if its request was sent over the same connection.

The program below shows the issue. I tested it against master.

```go
package main

import (
	"io"
	"io/ioutil"
	"log"
	"net/http"
	"strings"
)

func main() {
	tr := http.Transport{
		MaxConnsPerHost: 1,
	}

	// First request puts object in cache
	req, _ := http.NewRequest("GET", "http://localhost:8080/cached", nil)
	rsp, err := tr.RoundTrip(req)
	if err != nil {
		log.Fatal(err)
	}
	io.Copy(ioutil.Discard, rsp.Body)
	log.Printf("%d", rsp.StatusCode)

	// Conditional GET request with a body
	req, _ = http.NewRequest("GET", "http://localhost:8080/cached", strings.NewReader("\n"))
	req.Header.Add("If-None-Match", rsp.Header.Get("ETag"))
	rsp, err = tr.RoundTrip(req)
	if err != nil {
		log.Fatal(err)
	}
	log.Printf("%d", rsp.StatusCode)

	// Server reads the body from the previous request as part of the next request and returns
	// a 400
	req, _ = http.NewRequest("GET", "http://localhost:8080/cached", nil)
	rsp, err = tr.RoundTrip(req)
	if err != nil {
		log.Fatal(err)
	}
	body, _ := ioutil.ReadAll(rsp.Body)
	log.Printf("%d - %s", rsp.StatusCode, string(body))
}
```

This prints the following:

```
2022/07/09 13:03:52 200
2022/07/09 13:03:52 304
2022/07/09 13:03:52 400 - <HTML>
<HEAD>
<TITLE>Bad Request</TITLE>
</HEAD>

<BODY BGCOLOR="white" FGCOLOR="black">
<H1>Bad Request</H1>
<HR>

<FONT FACE="Helvetica,Arial"><B>
Description: Could not process this request.
</B></FONT>
<HR>
</BODY>
```

--------------------------------------------------

[...]

PR #8967, opened 2022-07-20
Fixes issue #8951 Client Request Body processing

kanseiishikawa:
Fixes[https://github.com/apache/trafficserver/issues/8951](https://github.com/apache/trafficserver/issues/8951).
It's okay call to `do_drain_request_body` if `t_state.next_action` is `SM_ACTION_INTERNAL_CACHE_NOOP`.
Calling `do_drain_request_body` to process unwanted Body.

### DIFF
--- a/proxy/http/HttpSM.cc
+++ b/proxy/http/HttpSM.cc
@@ -7727,6 +7727,7 @@ HttpSM::set_next_state()
     if (t_state.api_next_action == HttpTransact::SM_ACTION_API_SEND_RESPONSE_HDR) {
       handle_api_return();
     } else {
+      do_drain_request_body(t_state.hdr_info.client_response);
       t_state.api_next_action = HttpTransact::SM_ACTION_API_SEND_RESPONSE_HDR;
       do_api_callout();
     }


proxy/http/HttpSM.cc
```diff
@@ -7727,6 +7727,7 @@ HttpSM::set_next_state()
     if (t_state.api_next_action == HttpTransact::SM_ACTION_API_SEND_RESPONSE_HDR) {
       handle_api_return();
     } else {
+      do_drain_request_body(t_state.hdr_info.client_response);
```
bryancall, line 7730: Is there a reason you are doing this within the conditional and not always draining?

kanseiishikawa: When I reproduced it in the same way as  [https://github.com/apache/trafficserver/issues/8951](https://github.com/apache/trafficserver/issues/8951) , I did a drain here and the problem was solved.
I don't have a reason not always drain.

bneradt:
[approve ci]

bneradt:
[approve ci]

bryancall:
[approve ci autest]

PR #8968, opened 2022-07-20
 make doc corrupt for content length mismatch

vmamidi:


cmcfarlen:
see #8976 for replacement for this PR

PR #8970, opened 2022-07-20
add a metric to track how often the range seek bug is detected

cmcfarlen:
This add a metric to track how often objects are deleted due to incorrect fragments or fragment table issues.  There is a comment that this occasionally happens in production, so I would like to be able to understand how often.


bryancall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8971, opened 2022-07-20
Fixes issue with file size calculation for existing logs

elsloo:
* Issue arises with existing log files at startup

* Because the existing bytes are not accounted for, log rolling does not occur at the correct time

* Existing code can lead to logging being suspended indefinitely without manual intervention if thresholds are exceeded and no rolled log files can be deleted

* Corner case more evident when other data not rolled by ATS is present in the logging directory

### DIFF
--- a/src/tscore/BaseLogFile.cc
+++ b/src/tscore/BaseLogFile.cc
@@ -333,7 +333,8 @@ BaseLogFile::open_file(int perm)
   }
 
   // set m_bytes_written to force the rolling based on file size.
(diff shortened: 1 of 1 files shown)


bneradt (approved):
Makes sense.

maskit (commented):
I think this bug should be fixed on 9.2.0 but 9.2.1 is also acceptable.

zwoop:
Cherry-picked to v9.2.x

PR #8972, opened 2022-07-21
Adding origin-side ALPN configuration.

bneradt:
Adding the ability for ATS to specify the ALPN string it sends in the
TLS ClientHello handshake.


proxy/ProxySession.cc
```diff
@@ -313,3 +315,21 @@ ProxySession::support_sni() const
 {
   return _vc ? _vc->support_sni() : false;
 }
+
+PoolableSession *
+ProxySession::protocol_creation(NetVConnection *netvc)
+{
+  // Figure out what protocol was negotiated
+  int proto_index             = SessionProtocolNameRegistry::INVALID;
+  SSLNetVConnection *sslnetvc = dynamic_cast<SSLNetVConnection *>(netvc);
```
maskit, line 324: I don't see any reason to receive a NetVC here. Just a protocol id is enough. Let's deal with NetVC on HttpSM side.

How about this? I want function names to have verbs.
```
PoolableSession *
ProxySession::create_outbound_sesion(int protocol_id) {
  auto iter = ProtocolSessionCreateMap.find(protocol_id);
   ink_release_assert(iter != ProtocolSessionCreateMap.end());
   return iter->second();
}
```

iocore/net/ALPNSupport.cc
```diff
@@ -144,3 +144,36 @@ ALPNSupport::registerNextProtocolSet(SSLNextProtocolSet *s, const SessionProtoco
   this->npnSet       = s;
   npnSet->create_npn_advertisement(protoenabled, &npn, &npnsz);
 }
+
+bool
+ALPNSupport::process_alpn_protocols(std::string_view protocols, unsigned char *client_alpn_protocols, int &alpn_array_len)
```
maskit, line 149: It would be nice if you could put this into some module that parses setting values. Maybe RecHttp.cc, which parses server_ports.

maskit (changes requested):
Thank you for making this PR, and adding the autest. I repeated my comments on #8963 so we can discuss here.

I think none of them is a blocker, but I'd like to hear opinions from others about the alpn_protocols setting. The documentation would need to be updated accordingly (it should tell which protocols are available anyway).

maskit:
I'd like to hear how you want to configure which protocols to use/offer for origin connection.

Examples:
- h2,http/1.1 -- ALPN protocol IDs
- http2,http1 -- Aliases
- all -- Another alias to use all available protocols, and prioritize newer one

maskit:
For `server_ports`, we no longer need to specify `http/2` explicitly though, these protocol names (not IDs) are available.
https://docs.trafficserver.apache.org/admin-guide/configuration/session-protocol.en.html

shinrich:
I'm confused why are we bringing this forward to 10-dev, but not just bringing in the Http2 to origin branch?  Does this logic come from there?

bneradt:
> I'm confused why are we bringing this forward to 10-dev, but not just bringing in the Http2 to origin branch?  Does this logic come from there?

Correct @shinrich. This comes from the HTTP/2 to origin PR (#8963) from which this PR was requested.

maskit:
As for target branch, I'm fine with either 10-dev or master. I requested making this PR to have more eyes.

proxy/ProxySession.cc
bneradt: Good suggestion.

iocore/net/ALPNSupport.cc
bneradt: RecHttp.cc works well. I put it there and added unit tests for it as well.

proxy/http/HttpSM.cc
```diff
@@ -1795,9 +1795,20 @@ HttpSM::handle_api_return()
 PoolableSession *
 HttpSM::create_server_session(NetVConnection *netvc)
 {
-  HttpTransact::State &s  = this->t_state;
-  PoolableSession *retval = httpServerSessionAllocator.alloc();
+  // Figure out what protocol was negotiated
+  int proto_index             = SessionProtocolNameRegistry::INVALID;
+  SSLNetVConnection *sslnetvc = dynamic_cast<SSLNetVConnection *>(netvc);
```
maskit, line 1800: `ALPNSupport *` should be enough.

doc/admin-guide/files/records.config.en.rst
```diff
@@ -3956,6 +3956,12 @@ Client-Related Configuration
 
    Enables (``1``) or disables (``0``) TLSv1_3 in the ATS client context. If not specified, enabled by default
 
+.. ts:cv:: CONFIG proxy.config.ssl.client.alpn_protocols STRING ""
```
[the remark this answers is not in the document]
bneradt: These are all really good points. I think I've addressed each of your thoughts:

1. I've updated the ALPN parsing logic to enforce that the user only passes expected protocols, currently `http/1.1` and `http/1.0`. With HTTP/2 to origin we'll add `h2`. If any other protocol strings are passed, a WARNING is emitted and no ALPN extension is sent (the default behavior). This makes the configuration "fool-proof" for the user in that they can't break anything by configuring it, and a WARNING is emitted if something is wrong with their configured ALPN.
2. I expanded our documentation of this feature, pointing to the ALPN RFC and the official IANA list of registered ALPN protocol strings. I also explicitly documented the ALPN protocols ATS supports. (Again, we'll add `h2` with the HTTP/2 to origin PR.)

Thank you for your thoughtful feedback on this, @maskit. It really improved this feature.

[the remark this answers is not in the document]
maskit: Thanks for updating the documentation. At minimum it works for me, but I can't tell if users want to read the external documents. The example configurations are nice. I guess most users would just see the examples, and copy a setting value from there. In that sense, whether name or ID is not really important, although I think using names (or aliases) requires less knowledge.

I'm not going to judge if the setting format is user friendly, but if I made a Web UI for a CDN service that uses ATS, it'd have a reorderable list for this setting and the items would have protocol names but not IDs. It wouldn't even tell it's a setting for ALPN.

And very strictly speaking, "," is a valid character for an ALPN protocol ID. I hope future protocol IDs (which we want to support) won't use it.

maskit (approved):
I can't tell whether users like the setting format, but it works for me at minimum.

Issue #8979, opened 2022-07-26
Crash on RangeTransform::handle_event


masaori335:
We faced the below crash on a 9.2.0 testing box. There must be some conditions like a slow disk to trigger the crash, but we haven't found it.

```
[ 00 ] libpthread-2.17.so  waitpid                                                              
[ 01 ] traffic_server      crash_logger_invoke(int, siginfo_t*, void*)                          ( Crash.cc:168 )
[ 02 ] libpthread-2.17.so                                                                       
[ 03 ] traffic_server      RangeTransform::handle_event(int, void*)                             ( iocore/eventsystem/I_Lock.h:263 )
[ 04 ] traffic_server      EThread::process_event(Event*, int)                                  ( I_Continuation.h:227 )
[ 05 ] traffic_server      EThread::process_queue(Queue<Event, Event::Link_link>*, int*, int*)  ( UnixEThread.cc:188 )
[ 06 ] traffic_server      EThread::execute_regular()                                           ( UnixEThread.cc:248 )
[ 07 ] traffic_server      EThread::execute()                                                   ( UnixEThread.cc:353 )
[ 08 ] traffic_server      spawn_thread_internal(void*) [clone .llvm.12268301668034307738]      ( Thread.cc:79 )
[ 09 ] libpthread-2.17.so  start_thread                                                         
```

masaori335:
```
(gdb) f 3
#3  RangeTransform::handle_event(int, void*) (this=0x7f86751c0000, event=<optimized out>, edata=<optimized out>) at Transform.cc:817
817          MUTEX_TRY_LOCK(trylock, m_write_vio.mutex, this_ethread());
(gdb) p this->m_write_vio
$1 = {
  cont = 0x0, 
  nbytes = 0, 
  ndone = 0, 
  op = 0, 
  buffer = {
    mbuf = 0x0, 
    entry = 0x0
  }, 
  vc_server = 0x0, 
  mutex = {
    m_ptr = 0x0
  }, 
  _disabled = false
}
```

masaori335:
I tried the `--enable-event-tracker` build option to see where the event is scheduled. However, it points odd line, unfortunately.

```
(gdb) p e->_location
$3 = (const void *) 0x5654998b4f10 <HttpTunnel::consumer_handler(int, HttpTunnelConsumer*)+320>
(gdb) info line *0x5654998b4f10
[...]

PR #8980, opened 2022-07-26
Add RangeTransform::m_write_vio state checks

masaori335:
Workaround fix for #8979. It looks like events are fired to the RangeTransform::handle_event regardless of the `m_write_vio` state.


ywkaras (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8982, opened 2022-07-26
Add anonymous mmdb support

ezelkow1:
Adds a new anonymous field for when you are using mmdbs with anonymous IP support

### DIFF
--- a/doc/admin-guide/plugins/maxmind_acl.en.rst
+++ b/doc/admin-guide/plugins/maxmind_acl.en.rst
@@ -78,3 +78,32 @@ Optional
 ========
 
 There is an optional ``html`` field which takes a html file that will be used as the body of the response for any denied requests if you wish to use a custom one.
+
+Anonymous
+=========
+
+There is also an optional ``anonymous`` field. This allows you to use the maxmind anonymous IP database and reference it's optional fields. These are ``ip``, ``vpn``, 
+``hosting``, ``public``, ``tor``, and ``residential``. Currently anonymous blocking cannot be combined with GeoIP blocking since it is considered a separate database.
+However if you custom generate your own database that includes anonymous data then you could use them at the same time. Another solution is to have two instances
+of the maxmind_acl plugin on a remap, one to handle geo blocking and another for anonymous blocking. A setting of ``true`` will cause that type to be blocked, ``false``
+will not block and is effectively a no-op.
+
+An example configuration ::
+
+   maxmind:
+    database: GeoIP2-Anonymous-IP.mmdb
+    anonymous:
+     hosting: false
+     ip: true
+     vpn: true
+     tor: false
+    allow:
+     ip:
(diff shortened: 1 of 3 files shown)


doc/admin-guide/plugins/maxmind_acl.en.rst
```diff
@@ -78,3 +78,32 @@ Optional
 ========
 
 There is an optional ``html`` field which takes a html file that will be used as the body of the response for any denied requests if you wish to use a custom one.
+
+Anonymous
+=========
+
+There is also an optional ``anonymous`` field. This allows you to use the maxmind anonymous IP database and reference it's optional fields. These are ``ip``, ``vpn``,
```
randall, line 85: MaxMind GeoIP2 Anonymous IP database?

ezelkow1: yes, meant to be used with, https://dev.maxmind.com/geoip/docs/databases/anonymous-ip?lang=en#binary-database, the maxmind anonymous IP database. It's the same format but more barebones than the normal geo since its just a handful of fields returned with no nested structure to it

plugins/experimental/maxmind_acl/mmdb.cc
```diff
@@ -497,6 +577,94 @@ Acl::eval(TSRemapRequestInfo *rri, TSHttpTxn txnp)
...
+  if (_anonymous_ip) {
+    status = MMDB_get_value(entry, &entry_data, "is_anonymous", NULL);
+    if ((MMDB_SUCCESS == status) && (entry_data.has_data)) {
+      if (entry_data.type == MMDB_DATA_TYPE_BOOLEAN) {
+        if (entry_data.boolean == true) {
+          TSDebug(PLUGIN_NAME, "saw is_anonymous set to true bool");
+          return false;
+        }
+      }
+    }
```
brbzull0, line 602: not an issue.
Maybe you can have an lambda function in here to perform this field get and call it for every field, code seems similar except for the field name.

doc/admin-guide/plugins/maxmind_acl.en.rst
randall: Sorry, I meant to say replace "maxmind anonymous IP database" with "MaxMind GeoIP2 Anonymous IP database". (adding capitalization to the company and product names).

plugins/experimental/maxmind_acl/mmdb.cc
ezelkow1: yea, eventually Im planning on breaking it out more like that to make it more generic (we have some internal ideas around custom maxmind databases so the hope would be to basically have this be based just on a field name in the config and could be completely arbitrary)

just wanted to get this out there for now in a working state

brbzull0 (approved):
Looks good to me! Thanks.

PR #8985, opened 2022-07-28
Fix autest: Add missing config value for origin's dead policy.

brbzull0:
This seems to be in [master](https://github.com/apache/trafficserver/blob/80a0ff9e9ccbdfa1667a836291d9fd18f3937762/tests/gold_tests/tls/tls_verify_override_base.test.py#L118) but missing in [10-Dev](https://github.com/apache/trafficserver/blob/1b982db067c9285b8c65b2a0ed5a30c73d3a5789/tests/gold_tests/tls/tls_verify_override_base.test.py#L117)


bneradt (commented):
Thank you @brbzull0 for fixing autest on 10-Dev!

PR #8987, opened 2022-07-29
Cleanup: Don't refer to P_*.h from header files in proxy dir

maskit:
Including `P_*.h` is too much and that brings unnecessary dependencies. The source files (`*.cc`) in proxy directory still include `P_*.h` but that is ok for now because those files don't pollute other files by being included.

### DIFF
--- a/iocore/dns/I_DNSProcessor.h
+++ b/iocore/dns/I_DNSProcessor.h
@@ -24,7 +24,11 @@
 #pragma once
 
 #include "tscore/ink_resolver.h"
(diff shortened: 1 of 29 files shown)


maskit:
The same error (undefined symbol on runtime) happened on Debian 11 as well.

To not bring unnecessary dependencies, I changed Makefile to link individual object files that are really tested instead of libhttp2.a.

bneradt (approved):
Looks great. Glad you were able to work out  the undefined symbol issues with he unit tests.

PR #8988, opened 2022-07-29
Change H2 frame handlers to member functions

### LINKED ISSUE #8183: Should rework frame_handlers table in Http2ConnectionState to reference methods instead of functions
This is how the handler dispatch is handled in HttpSM::main_handler and in the handleEvent function for Continuations.

This way the rcv_* methods in Http2Connection could be methods instead of free functions, and we wouldn't have to pass the Http2ConnectionState object in as the first argument.  More C++ consistent.  And the rcv_* methods could then access private data about the Http2ConnectionState object.

maskit:
This closes #8183

### DIFF
--- a/proxy/http2/Http2ConnectionState.cc
+++ b/proxy/http2/Http2ConnectionState.cc
@@ -49,8 +49,6 @@
   SsnDebug(session->get_proxy_session(), "http2_con", "[%" PRId64 "] [%u] " fmt, session->get_connection_id(), stream_id, \
            ##__VA_ARGS__);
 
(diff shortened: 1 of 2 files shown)


brbzull0:
Nice change. To follow the same goal of this change I am wondering if most of the public members of `Http2ConnectionState` can now be also private as we aren't passing the state object around free functions?

maskit:
Yes, we should be able to make some of the members private, but not all of them. `Http2ConnectionState` is still used by `Http2Session` and `Http2Stream`.

I'm +1 on making those members private, but I'd like to make changes around H2 slowly step by step so the PR for H2-to-Orign can catch up changes on master easily.

brbzull0 (approved):
Looks good. Thanks.

PR #8990, opened 2022-08-01
Add HTTP header version converter

maskit:
This removes dependency for http2 form http3 by moving their common code into `proxy/hdrs/`.

The logic is basically unchanged. Only trick I added is a switch to create (reserve) pseudo header fields on HTTPHdr initialization. If you pass `HTTP_2_0` or `HTTP_3_0` to the init function, `MIMEField`s for pseudo header fields will be added automatically to ensure those fields will be placed at the beginning of the header field list.

### DIFF
--- a/proxy/hdrs/VersionConverter.cc
+++ b/proxy/hdrs/VersionConverter.cc
@@ -1,0 +1,274 @@
+/** @file
+
+  HTTP header version converter
+
+  @section license License
+
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+
+#include "VersionConverter.h"
+#include "tscpp/util/LocalBuffer.h"
+
+int
+VersionConverter::convert(HTTPHdr &header, int from, int to) const
+{
+  int type;
+
+  switch (http_hdr_type_get(header.m_http)) {
+  case HTTP_TYPE_REQUEST:
+    type = 0;
+    break;
+  case HTTP_TYPE_RESPONSE:
+    type = 1;
+    break;
+  default:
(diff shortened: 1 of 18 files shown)


proxy/hdrs/VersionConverter.h
```diff
@@ -0,0 +1,96 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+#include <string_view>
+#include "HTTP.h"
+
+#pragma once
```
bneradt, line 27: Probably best to put `#pragma once` before any source code, including includes.

proxy/hdrs/VersionConverter.h
```diff
@@ -0,0 +1,96 @@
...
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+#include <string_view>
+#include "HTTP.h"
```
bneradt, line 25: I suggest removing the include of `HTTP.h` from this header file, forward declaring `HTTPHdr` (since it is only used as a reference in the header), and adding the include of `HTTP.h` to the `VersionConverter.cc` file.

bneradt (changes requested):
Looks good.

proxy/hdrs/VersionConverter.cc
```diff
@@ -0,0 +1,265 @@
...
+  int type;
+
+  switch (http_hdr_type_get(header.m_http)) {
+  case HTTP_TYPE_REQUEST:
+    type = 0;
+    break;
+  case HTTP_TYPE_RESPONSE:
+    type = 1;
+    break;
+  default:
```
bneradt, line 39: I suggest changing from `default:` to `case HTTP_TYPE_UNKNOWN:`. That way if, for some reason, another enumeration is added, the compiler will warn about an unhandled case here which will remind the person changing the enum to check the logic in this switch.

proxy/hdrs/VersionConverter.h
```diff
@@ -0,0 +1,96 @@
...
+
+#pragma once
+
+class VersionConverter
```
bneradt, line 29: Adding a doxygen comments for this class would be nice.

bneradt (changes requested):

maskit:
@bneradt Thank you for your feedback. I updated the code and added oxygen comments.

bneradt (dismissed):
Looks good to me.

maskit:
@bneradt I rebased the branch on top of the latest master to resolve conflicts. PTAL.

bneradt (approved):

bneradt:
Looks good.

PR #8991, opened 2022-08-01
Initial Checkin for Wasm plugin

shukitchan:


shukitchan:
cc @bryancall 

as discussed, this is the wasm plugin. Please take a look at how the library license is included.

shukitchan:
Please note that the contents inside /lib directory are directly coming from the proxy-wasm repo. 
And I accidentally run clang-format on them so i need to undo that.

ywkaras:
Will there be Au tests for this feature?

shukitchan:
Probably we can add au tests later. 
For now, it is just the initial checkin

plugins/experimental/wasm/ats_wasm.h
```diff
@@ -0,0 +1,68 @@
...
+*/
+
+#pragma once
+
+#include "include/proxy-wasm/wasm.h"
+#include "ats_context.h"
+
+#include "ts/ts.h"
+
+namespace proxy_wasm
```
cmcfarlen, line 26: This namespace name should probably be ats specific right?  Maybe `ats_wasm`?

plugins/experimental/wasm/ats_wasm.cc
```diff
@@ -0,0 +1,149 @@
...
+    if (c->scheduler_cont() == nullptr) {
+      c->initialize(contp);
+    } else {
+      // keep the old continuation
+      TSContDestroy(contp);
+    }
+    c->onStart(plugin);
+    return c;
+  }
+  auto context     = std::unique_ptr<Context>((Context *)createRootContext(plugin));
```
cmcfarlen, line 81: This is a strange use of unique_ptr.  context is moved to root_contexts_ and also returned as the pointer.  Perhaps this should be shared (or weak_shared)?  Since `createRootContext` just calls new anyway, this code could be cleaner.

plugins/experimental/wasm/ats_wasm.cc
```diff
@@ -0,0 +1,149 @@
...
+
+ContextBase *
+Wasm::createContext(const std::shared_ptr<PluginBase> &plugin)
+{
+  return new Context(this, plugin);
+}
+
+// Function to start a new root context
+Context *
+Wasm::start(std::shared_ptr<PluginBase> plugin, TSCont contp)
```
cmcfarlen, line 67: pass plugin by const ref

cmcfarlen (changes requested):
There are a lot of clang-tidy errors that seem important, maybe go through those and see if some stuff could be cleaned up.

moonchen (changes requested):

plugins/experimental/wasm/ats_wasm.cc
ywkaras: Also preferable to avoid C-style casting.

plugins/experimental/wasm/ats_context.h
```diff
@@ -0,0 +1,207 @@
...
+#include "ts/experimental.h"
+
+#define WASM_DEBUG_TAG "wasm"
+
+namespace proxy_wasm
+{
+class Wasm;
+
+// constants for property names
+static std::string p_request_path("request\0path\0", 13);
```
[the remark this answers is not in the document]
ywkaras: Why the embedded null characters.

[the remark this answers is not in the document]
shukitchan: That is for interopration with envoy wasm filter. Their property names are defined in the same way.

plugins/experimental/wasm/ats_wasm.cc
shukitchan: root_contexts_ from the base class needs a unique_ptr as the value

bryancall:
[approve ci autest]

shukitchan:
[approve ci fedora]

moonchen (approved):

cmcfarlen (approved):

bryancall (approved):
@moonchen and @cmcfarlen approved this

PR #8992, opened 2022-08-01
Experimental io_uring AIO implementation for disk IO

cmcfarlen:
This PR adds support for using io_uring for AIO on Linux hosts.

For this implementation, each net thread and the main thread would have its own io_uring command queue for doing disk io.  The io_uring queue processing is integrated into the net thread's epoll loop via eventfd.  In this approach, there are no AIO specific threads and any request specifying AIO thread completion will be handled in the the net/main thread.


bryancall:
[approve ci]

maskit:
I wonder if we can also have a run-time setting (i.e. one for records.config) to enable/disable io_uring. It would be useful for testing/trouble shooting. And it would be also nice to be able to enable io_uring for disk IO and socket IO separately (e.g. 0: disabled, 1: just disk IO, 2: just socket IO, 3: the both).

bryancall (approved):

bryancall (approved):

cmcfarlen:
[approve ci]

cmcfarlen:
[approve ci]

Issue #6245, opened 2019-12-06
Remove INKUDPXX api


scw00:
If nobody use the these APIs, we can remove this or update it to TS API.

https://github.com/apache/trafficserver/blob/080889f68701c994826e2e5781d2564b4ed6733d/include/ts/InkAPIPrivateIOCore.h#L138-L198

sudheerv:
Worth sending an email to dev@

alokrahul513:
Hi, I want to work on this issue. How to get started?

abhi-jeet-kumar:
has this been fixed?

bryancall:
All the INKUDP* APIs have been removed for the 10.0.0 release

PR #8993, opened 2022-08-02
chore: refactored InkAPIPrivateIOCore.h

shikharvashistha:
This commit refactors the header to remove API's mentioned in #6245

Signed-off-by: Shikhar Vashistha <51105234+shikharvashistha@users.noreply.github.com>

### DIFF
--- a/proxy/http2/Http2ConnectionState.cc
+++ b/proxy/http2/Http2ConnectionState.cc
@@ -49,8 +49,6 @@
   SsnDebug(session->get_proxy_session(), "http2_con", "[%" PRId64 "] [%u] " fmt, session->get_connection_id(), stream_id, \
            ##__VA_ARGS__);
 
-using http2_frame_dispatch = Http2Error (*)(Http2ConnectionState &, const Http2Frame &);
-
 static const int buffer_size_index[HTTP2_FRAME_TYPE_MAX] = {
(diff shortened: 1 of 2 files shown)


brbzull0:
Hello, Thanks for working on this.
few things to mention:
- Check [here](https://github.com/apache/trafficserver/blob/86790c7a3158898219f807e04c284b172f9a7cfc/src/traffic_server/InkIOCoreAPI.cc#L436) as this needs to be removed too if we are going ahead with this cleanup.
- Haven't check but on the issue it was mentioned that it would be good to run this by sending an email to the dev@ list. You can include the link to this PR .

Thanks again for working on this.

bryancall:
[approve ci]

include/ts/InkAPIPrivateIOCore.h
```diff
@@ -138,64 +138,6 @@ tsapi void TSIOBufferBlockDestroy(TSIOBufferBlock blockp);
 typedef void *INKUDPPacket;
 typedef void *INKUDPacketQueue;
 typedef void *INKUDPConn;
```
ywkaras, line 140: Why not also delete these three INKUDPxxx typedefs?

shikharvashistha: I think they are being utilized

PR #8996, opened 2022-08-03
Add stack guard pages

moonchen:
Use r/w protected pages on top of stacks to guard against stack
overflow.  The number of VM pages to use for guarding can be set
via the config option proxy.config.thread.default.stackguard_pages.

### DIFF
--- a/iocore/eventsystem/UnixEventProcessor.cc
+++ b/iocore/eventsystem/UnixEventProcessor.cc
@@ -47,7 +47,7 @@ public:
   void init();
   /// Set the affinity for the current thread.
   int set_affinity(int, Event *);
-  /// Allocate a stack.
+  /// Allocate a stack and set guard pages.
   /// @internal This is the external entry point and is different depending on
   /// whether HWLOC is enabled.
   void *alloc_stack(EThread *t, size_t stacksize);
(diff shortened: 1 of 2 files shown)


moonchen:
Frequently asked questions:

**Q: How much overhead does this add?**

**A:** The VM mappings cost space in the page table.  It also adds fragmentation to the page table since adjacent mappings can no longer be coalesced due to having different protection flags.

**Q: Do the stack guard pages map to physical memory?**

**A:** Not when I tested on Linux.  The pages are not marked as present initially.  I believe they're mapped on demand.

**Q: Why did you change all stack allocations to use `mmap()`?**

**A:** `mprotect()` requires that the memory be first allocated with `mmap()`.

**Q: Why do you allow for more than one page of stack guard?**

**A:** Stack hopping can happen where stack allocations of larger than one page in size can skip the guard page.  Having more pages of guard reduces the likelihood of this.  See [this article](https://lwn.net/Articles/725832/
) for more details.

ywkaras (approved):

masaori335:
I want this on the 9.2.x branch because this will detect the stack overflow instead of the random crash.

@zwoop What do you think about this should go to 9.2.0 or 9.2.1?

bneradt:
> I want this on the 9.2.x branch because this will detect the stack overflow instead of the random crash.
> 
> @zwoop What do you think about this should go to 9.2.0 or 9.2.1?

Just to document this: we need to make sure that wherever this is ported, the Apple Silicon build fix goes with it:
https://github.com/apache/trafficserver/pull/8999

PR #8999, opened 2022-08-04
Fix compile on M1 Mac

moonchen:
Add arm64 to the list of known stack growth directions.


bneradt (approved):

bneradt:
This patch is tied to #8996. Since that latter PR is marked for 9.2.x, then this PR should go with it. Therefore I'm marking this PR with the 9.2.x project.

zwoop:
Cherry-picked to v9.2.x

PR #9004, opened 2022-08-08
Clean up some errors in output of ASAN build.

ywkaras:
The "leak" of the DbgCtl tags is actually benign, since these are not dereferenced until TS process exit.

### DIFF
--- a/proxy/logging/LogConfig.cc
+++ b/proxy/logging/LogConfig.cc
@@ -162,9 +162,13 @@ LogConfig::read_configuration_variables()
   ats_free(ptr);
 
   ptr = REC_ConfigReadString("proxy.config.log.hostname");
-  if (ptr != nullptr && std::string_view(ptr) != "localhost") {
-    ats_free(hostname);
-    hostname = ptr;
+  if (ptr != nullptr) {
+    if (std::string_view(ptr) != "localhost") {
+      ats_free(hostname);
+      hostname = ptr;
+    } else {
+      ats_free(ptr);
+    }
   }
 
   ptr = REC_ConfigReadString("proxy.config.error.logfile.filename");
--- a/src/traffic_server/traffic_server.cc
+++ b/src/traffic_server/traffic_server.cc
@@ -1917,6 +1917,7 @@ main(int /* argc ATS_UNUSED */, const char **argv)
   char *hostname = REC_ConfigReadString("proxy.config.log.hostname");
   if (hostname != nullptr && std::string_view(hostname) == "localhost") {
     // The default value was used. Let Machine::init derive the hostname.
+    ats_free(hostname);
     hostname = nullptr;
   }
   Machine::init(hostname, &machine_addr.sa);
--- a/src/tscore/DbgCtl.cc
+++ b/src/tscore/DbgCtl.cc
@@ -26,6 +26,7 @@
 #include <set>
 #include <cstring>
 
+#include <tscore/ink_assert.h>
 #include <tscore/Diags.h>
 
 // The resistry of fast debug controllers has a ugly implementation to handle the whole-program initialization
@@ -50,6 +51,13 @@ public:
   struct Data {
     std::mutex mtx;
     Set set;
+
+    ~Data()
+    {
+      for (auto &ctl : set) {
+        delete[] ctl.tag;
+      }
+    }
   };
 
   static Data &
@@ -85,13 +93,15 @@ DbgCtl::_get_ptr(char const *tag)
   ink_assert(sz > 0);
 
   {
-    char *t = new char[sz + 1]; // Deleted implicitly by program termination.
+    char *t = new char[sz + 1]; // Deleted by ~Data().
     std::memcpy(t, tag, sz + 1);
     ctl.tag = t;
   }
   ctl.on = diags() && diags()->tag_activated(tag, DiagsTagType_Debug);
 
   auto res = d.set.insert(ctl);
+
+  ink_assert(res.second);
 
   return &*res.first;
 }


configure.ac
```diff
@@ -996,14 +996,14 @@ if test "x${enable_asan}" = "xyes"; then
   if test "x${enable_tsan}" = "xyes" -o "x${enable_tsan}" = "xstatic"; then
     AC_MSG_ERROR(Cannot have ASAN and TSAN options at the same time, pick one)
   fi
-  TS_ADDTO(AM_CFLAGS, [-fno-omit-frame-pointer -fsanitize=address])
-  TS_ADDTO(AM_CXXFLAGS, [-fno-omit-frame-pointer -fsanitize=address])
+  TS_ADDTO(AM_CFLAGS, [-fno-omit-frame-pointer -fsanitize=address -DIS_ASAN_BUILD])
```
ywkaras, line 999: I originally wanted to add an "is ASAN build" flag to ink_config.h, but I couldn't figure out how to do it.

bryancall:
[approve ci autest]

ywkaras:
[approve ci autest]

ywkaras:
Ideally, all Au tests should run with a ASAN build with no errors.  This PR doesn't full achieve that goal, but it does make ASAN much more usable.  It eliminates a lot of benign leak errors from the ASAN output.

proxy/logging/LogConfig.cc
```diff
@@ -156,9 +156,13 @@ LogConfig::read_configuration_variables()
...
-  if (ptr != nullptr && std::string_view(ptr) != "localhost") {
-    ats_free(hostname);
-    hostname = ptr;
+  if (ptr != nullptr) {
+    if (std::string_view(ptr) != "localhost") {
+      ats_free(hostname);
+      hostname = ptr;
+    } else {
+      ats_free(ptr);
+    }
```
ywkaras, line 165: This fixes a small actual (non-spurious) leak.

src/traffic_server/traffic_server.cc
```diff
@@ -1917,6 +1917,7 @@ main(int /* argc ATS_UNUSED */, const char **argv)
   char *hostname = REC_ConfigReadString("proxy.config.log.hostname");
   if (hostname != nullptr && std::string_view(hostname) == "localhost") {
     // The default value was used. Let Machine::init derive the hostname.
+    ats_free(hostname);
```
ywkaras, line 1920: This fixes a small actual (non-spurious) leak.

cmcfarlen (changes requested):
This pointer registry is used to keep track of allocations in two poorly designed places in the code.  I think we could refactor those and until then just tell the leak sanitizer to ignore the two functions where the allocations occur.  One place is because of a global un-owned pointer(a big no-no in modern C++) and the other is a naked union(another no-no).  Both situations are described in the C++ core guidelines (this is my appeal to authority, such that the core guidelines are). Having this registry condones bad coding practices and I don't think we should include it.

moonchen:
There is an documented way in LeakSanitizer to ignore leaks: https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer#suppressions.  We should use that method if we want to ignore leaks.

If we want to change the code to remove the leak, we should do it in a way that removes it in both sanitizer and non-sanitizer builds.  Having different code for sanitizer vs non-sanitizer should be a last resort since it defeats the verification function of the sanitizers.

ywkaras:
The intent of this PR is to make ASAN more usable.  Much of ATS could politely be called "pre-Modern", but there doesn't seem to be consensus to make addressing that a priority.  There has to be consensus on strategic goals.  If we all start blocking PRs to try to enforce what we individually think strategic goals should be, that's a recipe for gridlock.

The registry only identifies and filters "leaks" that have been determined to be benign.  It doesn't defeat the purpose of the sanitizer.  To me, it seems more maintainable than a single file with a vague(pattern-based) list of all the benign leaks.

src/records/RecUtils.cc
```diff
@@ -129,10 +131,12 @@ RecDataSet(RecDataT data_type, RecData *data_dst, RecData *data_src)
       }
     } else if (((data_dst->rec_string) && (strcmp(data_dst->rec_string, data_src->rec_string) != 0)) ||
                ((data_dst->rec_string == nullptr) && (data_src->rec_string != nullptr))) {
+      OwningPtrRegistry::rm(&(data_dst->rec_string), data_dst->rec_string);
       ats_free(data_dst->rec_string);
 
       data_dst->rec_string = ats_strdup(data_src->rec_string);
-      rec_set              = true;
+      OwningPtrRegistry::add_raw(&(data_dst->rec_string), data_dst->rec_string);
```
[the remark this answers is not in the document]
ywkaras: Generally one should look at change raw unions to std::variants.  But that's beyond the scope of this PR.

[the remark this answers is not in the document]
cmcfarlen: Agreed.  A std::variant would be better!

cmcfarlen:
I prefer the file as its easier to reason about and it doesn't come with incidental code.

ywkaras:
> I prefer the file as its easier to reason about and it doesn't come with incidental code.

I don't understand "reason about".  To me, when something is added to the registry, it clarifies the intent to allow process exit to clean up the allocation.  Without having to go look in a separate file, remember the precise format of the ASAN output.  Doesn't look like the meaning of "pattern" is well documented, whether it's a full regex, etc.  What would the leak suppressor do if a pattern matched no leak?  The registry can handle it if whether or not something is deallocated by process exit depends on runtime circumstances.

cmcfarlen:
I just mean that when you are doing "ASAN" things and looking for leaks, all of the excluded ones are all in one place.  This way managed with code is incidental to anything else.  Also, we don't need to invent a new method to mark "benign" leaks when one is already provided.

ywkaras:
When something is provided, one also has to consider how well documented and therefore usable it is.

ywkaras:
[approve ci autest]

ywkaras:
OK the registry is gone.

cmcfarlen (approved):

PR #9005, opened 2022-08-08
Add back validatation that the scheme matches the wire protocol

bneradt:
This adds back in the scheme and wire protocol check (see #8465) along
with a configuration to be able to disable the check if the verification
is not desired.


moonchen (approved):

masaori335 (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9008, opened 2022-08-08
Pin flask to version 2.1.3

bneradt:
This resolves an AuTest Pipenv package dependency conflict for Werkzeug,
which is used by httpbin. Latest versions of flask require newer
versions of flask which conflicts with our pin to keep httpbin working.


randall (approved):

zwoop:
Cherry-picked to v9.2.x

Issue #8196, opened 2021-08-03
Format codes for url log fields are confusing


shinrich:
@maskit brought this up for PR #8015 which addresses issue #4684.  The format codes start start with c, but most/all of them really are logging information about the proxy request url.

For the PR @maskit suggests the following which I agree with.

Log field names for content length is very clear. We should have followed this manner.

cqcl | Client Request | Client request content length, in bytes.
pqcl | Proxy Request | Content body length of the Traffic Server proxy request to the origin server.
pscl | Proxy Response | Content body length of the Traffic Server proxy response.
sscl | Origin Response | Content body length of the origin server response to Traffic Server.

We merged the PR since it improved the documentation to reflect reality.  Filing this issue to address fixing the underlying implementation to improve a future reality.

bryancall:
This was addressed for 10.0.0 by #9019, which added clearer Proxy Request URL log field codes (`pqu`, `pquc`, `pqup`, `pqus`) and deprecated the confusingly-named `cqu`/`cquc`/`cqup`/`cqus`. The docs now mark the old codes deprecated in favor of the `pqu*` set. Closing as done.

PR #9019, opened 2022-08-10
Add log field names pqu* and deprecates cqu*

maskit:
This is for #8196, and the target branch is 10-Dev because it deprecates a couple of log fields.

#8015 corrected errors on the documentation, but the field names were kept as they are for compatibility. This PR deprecates the incorrectly named fields, `cpu`, `cquc`, `cqup`, and `cqus`.  And introduces `pqu`, `pquc`, `pqup`, and `pqus` as replacements.

I know variable names used in Log.cc still has "client_request" in their names and those should be renamed accordingly, but I wanted to make this field name change before it's too late since we can make this change only on a new major version release.


maskit:
I had to update a binary test data, and I made the change like this.
```
sed -e 's/cquc/pquc/' src/traffic_logstats/tests/logstats.blog.orig > src/traffic_logstats/tests/logstats.blog
```

bneradt (approved):

PR #9021, opened 2022-08-10
aio_thread_main does not have to be static

cukiernik:


moonchen:
Is it possible to go all the way, remove `thr_info`, and use `this` instead?

cukiernik:
@moonchen This commit is separated (as independent part of) from the major change in this part of the classes.

bryancall:
[approve ci autest]

cmcfarlen (approved):

PR #7925, opened 2021-06-07
For strategies.yaml, add peering_ring with go_direct, and Au tests for peering_ring.

ywkaras:
Caching is disabled when the next hop is a peer (in the first host group).

PR #9023, opened 2022-08-10
Remove plugin parent_select unused symbols

rob05c:
These `exclude_hostname` and `exclude_port` args were needed due to limitations on events, the parent_select plugin needed to get the subsequent parent to try, if necessary, because no event would fire. So it needed to say, "give me the next parent *if* this one were marked down, even though it isn't."

Their usage got lost in the parent_select consistent hash refactor (#8590), which I thought would break failover.

But I just extensively tested failover, for DNS, network, and internal code failures, and everything worked up to 3 tries. So it seems either #8590 (#7925 and #8365) or possibly the recent HostDB refactor (#8953) changed something to not make it necessary.

So this PR cleans up the parameters and variables, and just removes them entirely.

This is purely cleanup, doesn't affect behavior. But I recommend backporting to 9.2.x just to reduce future merge conflicts.

### DIFF
--- a/plugins/experimental/parent_select/consistenthash.cc
+++ b/plugins/experimental/parent_select/consistenthash.cc
@@ -288,9 +288,8 @@ PLNextHopConsistentHash::deleteTxn(void *txn)
 }
 
 void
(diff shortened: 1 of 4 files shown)


bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9027, opened 2022-08-11
Change linux/fs include to build with glibc 2.36

jeredfloyd:
glibc 2.36 includes changes (documented at https://sourceware.org/glibc/wiki/Release/2.36#Usage_of_.3Clinux.2Fmount.h.3E_and_.3Csys.2Fmount.h.3E) that create build errors when including linux/fs.h for mount configuration.  sys/mount.h is a more appropriate include here, which works with older glibc as well.


bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9037, opened 2022-08-16
Add support for sendmmsg and UDP GSO

maskit:
This PR is basically to send multiple UDP packets by calling send(m)msg as less as possible. It improved QUIC throughput by 13%.

The configuration script detects the availability of send*m*msg. It'll be always used if available.

The code tries to use GSO if `SOL_UDP` is defined (and ATS is configured to use it), but it is not always available even if Kernel supports GSO because it can be administratively disabled. In case of a try to use GSO failed, it fallbacks to simple send(m)msg call.

- sendmsg (MacOS)
- sendmmsg (FreeBSD and Linux)
- sendmmsg + GSO (Linux)

Settings:
`proxy.config.udp.threads` has been there for years, but was not documented.
`proxy.config.udp.enable_gso` is a new setting that enables use of UDP GSO on ATS.

This is not incompatible change, but I thought it'd be nice to land on 10-Dev first because it's not small and it's only needed by QUIC.

### DIFF
--- a/doc/admin-guide/files/records.config.en.rst
+++ b/doc/admin-guide/files/records.config.en.rst
@@ -4536,6 +4536,18 @@ removed in the future without prior notice.
    This is just for debugging. Do not change it from the default value unless
    you really understand what this is.
 
+UDP Configuration
+=====================
+
+.. ts:cv:: CONFIG proxy.config.udp.threads INT 0
+
+   Specifies the number of UDP threads to run.
+
+.. ts:cv:: CONFIG proxy.config.udp.enable_gso INT 0
+
+   Enables (``1``) or disables (``0``) UDP GSO. When enabled, |TS| tries to use UDP GSO,
+   and disables it automatically if it causes send errors.
+
 Plug-in Configuration
 =====================
 
--- a/doc/admin-guide/files/records.config.en.rst
+++ b/doc/admin-guide/files/records.config.en.rst
@@ -4541,7 +4541,8 @@ UDP Configuration
 
 .. ts:cv:: CONFIG proxy.config.udp.threads INT 0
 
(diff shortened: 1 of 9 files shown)


iocore/net/UnixUDPNet.cc
```diff
@@ -860,7 +869,16 @@ UDPNetProcessor::UDPBind(Continuation *cont, sockaddr const *addr, int fd, int s
...
 
 // send out all packets that need to be sent out as of time=now
-UDPQueue::UDPQueue() {}
+#ifdef SOL_UDP
+UDPQueue::UDPQueue(bool enable_gso) : use_udp_gso(enable_gso) {}
+#else
+UDPQueue::UDPQueue(bool enable_gso)
+{
+  if (enable_gso) {
+    Warning("UDP GSO is unavailable");
```
bneradt, line 878: I suggest::

```
    Warning("Attempted to use UDP GSO per configuration, but it is unavailable");
```

doc/admin-guide/files/records.config.en.rst
```diff
@@ -4536,6 +4536,18 @@ removed in the future without prior notice.
    This is just for debugging. Do not change it from the default value unless
    you really understand what this is.
 
+UDP Configuration
+=====================
+
+.. ts:cv:: CONFIG proxy.config.udp.threads INT 0
+
+   Specifies the number of UDP threads to run.
```
bneradt, line 4544: Probably good to add:

```
By default 0 threads are dedicated to UDP, which results in effectively disabling UDP support.
```

bneradt (approved):
Looks good. Just a few comments.

maskit:
Thank you for the suggestions. I updated the message and documentation as suggested.

PR #9040, opened 2022-08-16
ts::shared_mutex - Fix lock counting for debug

moonchen:
Lock count variable was left uninitialized, causing assertions to fail.
This change initializes the debug data so they work properly.  Also add
a similar check to the locking side.

### DIFF
--- a/include/tscpp/util/TsSharedMutex.h
+++ b/include/tscpp/util/TsSharedMutex.h
@@ -107,7 +107,9 @@ public:
     if (error != 0) {
       _call_fatal("pthread_rwlock_rdlock", &_lock, error);
     }
+    X(TSAssert(_shared >= 0);)
     X(++_shared;)
+    X(TSAssert(_shared > 0);)
   }
 
   bool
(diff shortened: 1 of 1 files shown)


include/tscpp/util/TsSharedMutex.h
```diff
@@ -184,8 +186,8 @@ class shared_mutex
 
   // In debug builds, make sure shared vs. exlusive locks and unlocks are properly paired.
   //
-  X(std::atomic<bool> _exclusive;)
-  X(std::atomic<int> _shared;)
+  X(std::atomic<bool> _exclusive{false};)
+  X(std::atomic<int> _shared{0};)
```
ywkaras, line 190: This isn't necessary.

ywkaras (approved):

include/tscpp/util/TsSharedMutex.h
moonchen: According to this: https://en.cppreference.com/w/cpp/atomic/atomic/atomic

> The default constructor is trivial: no initialization takes place other than [zero initialization](https://en.cppreference.com/w/cpp/language/zero_initialization) of static and thread-local objects. [std::atomic_init](https://en.cppreference.com/w/cpp/atomic/atomic_init) may be used to complete initialization.

I also ran into an uninitialized value during local testing, which drove this change.

ywkaras: Oops, sorry, I thought the std::atomic constructor always zero initialized, even for stack instances.

moonchen: No problem!

zwoop:
Cherry-picked to v9.2.x

PR #9041, opened 2022-08-16
Fix reload memory leak

rob05c:
Fixes a memory leak when reloading hosting.config.

I manually tested and ran, modified hosting.config and reloaded. Appears to work, requests and things continue to work, debug prints in the constructor and destructor show a new `CacheHostTableConfig` being allocated on reload, and the old one being deleted once and only once.

Recommend backporting to All The Things. Appears to have been a bug forever.


masaori335 (approved):
Looks reasonable. 

The `CacheHostTableConfig` is allocated on scheduling the event and nobody owns the pointer.
https://github.com/apache/trafficserver/blob/2e31ef8a5a0a4327a17084c2ae51c0325d1feb61/iocore/cache/CacheHosting.cc#L228-L235

zwoop:
Cherry-picked to v9.2.x

PR #9046, opened 2022-08-18
Fix hosting.config reload

### LINKED ISSUE #7220: hosting.config requires restart, but docs say reloadable

rob05c:
This is an alternative to https://github.com/apache/trafficserver/pull/8664. That PR is fine in-itself and fixes a bug around the the atomic swap not actually reloading. But fixing that bug so that reloading hosting.config works exposes another deeper bug: the reloaded object has no synchronization around it. 

Not only is the pointer itself never read with synchronization, even if it were, the object itself is never synchronized. So even if the pointer were atomic, the object reads could get different variables from different objects, when callers that need multiple variables really need them to come from a single object. Worse, if a request takes longer than the arbitrary delete timer, it could access freed memory.

This fixes the object to be synchronized, and all access behind a mutex. The reload swap and bad drive removal acquire a write lock, and all other accesses (none of which modify the object) acquire a read lock.

I really don't like adding a mutex to the request path. But the only other options I see are an incredibly complex and bug-prone lock-free solution, or not making it possible to reload hosting.config. As much as I dislike it, the mutex should never have contention except on reload or drive failure. So it's probably actually faster and fewer atomic instructions than even a lock-free solution.

This also adds some helper classes, to make it impossible to accidentally forget to acquire or release the mutex, or modify the object with a readlock, via RAII idioms.

Recommend backporting to 9.2.

Fixes #7220

### DIFF
--- a/iocore/cache/P_CacheHosting.h
+++ b/iocore/cache/P_CacheHosting.h
@@ -22,6 +22,8 @@
  */
 
 #pragma once
+#include <memory>
+#include <shared_mutex>
 #include "P_Cache.h"
 #include "tscore/MatcherUtils.h"
 #include "tscore/HostLookup.h"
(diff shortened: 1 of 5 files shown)


traeak:
[approve ci docs]

bryancall:
[approve ci autest]

SolidWallOfCode (approved):
I wonder about line 284, though - there doesn't seem to be a provision for a failure on a reload.

iocore/cache/P_CacheHosting.h
```diff
@@ -152,23 +261,34 @@ class CacheHostTable
...
+
+    CacheType type = CACHE_HTTP_TYPE;
+    Cache *cache   = nullptr;
+    {
+      ReplaceablePtr<CacheHostTable>::ScopedReader hosttable(ppt);
+      type  = hosttable->type;
+      cache = hosttable->cache;
+    }
+    ppt->reset(new CacheHostTable(cache, type));
+    // new_Deleter(old, CACHE_MEM_FREE_TIMEOUT);
```
randall, line 285: remove commented out code.

rob05c: Removed

randall (approved):

rob05c:
[approve ci autest]

zwoop:
Cherry-picked to v9.2.x

PR #9049, opened 2022-08-19
Fast log buffering mode

cmcfarlen:
This adds a configurable `fast` mode to log objects that use per-thread log buffers rather than having all threads contend for per-object buffers.  While this is much faster, it does not support ordered log entries.  This makes it situationally useful, but if a user has binary logs or doesn't need ordered log output, this could allow them to log many more transactions.  Its configurable either globally via records.config or per-log via logging.yaml.

### DIFF
--- a/proxy/logging/unit-tests/benchmark_LogObject.cc
+++ b/proxy/logging/unit-tests/benchmark_LogObject.cc
@@ -1,0 +1,144 @@
+/** @file
+
+Simple benchmark for LogObject
+
+@section license License
+
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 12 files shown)


proxy/logging/unit-tests/benchmark_LogObject.cc
```diff
@@ -0,0 +1,169 @@
...
+  auto diagsConfig = new DiagsConfig("Server", "diags.log", "", "", false);
+  // unused, but constructor must be called for side effects.
+  (void)diagsConfig;
+  diags()->set_std_output(StdStream::STDOUT, bind_stdout);
+  diags()->set_std_output(StdStream::STDERR, bind_stderr);
+
+  if (is_debug_tag_set("diags")) {
+    diags()->dump();
+  }
+  Layout::create("/opt/ats");
```
maskit, line 81: The directory doesn't need to be exist, right? It's concerning if it does. No other tests pass a path.

cmcfarlen: This is a  benchmark so you have to go out of your way a bit to run it, but you are correct it shouldn't assume where to put files.  I'll clean this up.  Thanks!

cmcfarlen: After more looking this doesn't look like It changes the filesystem, this just needs to be called so the layout exists.  Also, the files for the benchmark are written to `/tmp`.

PR #9052, opened 2022-08-23
Fix the connection limit crash while using parents (#7602)

ezelkow1:
(cherry picked from commit e8db7c495127915c0fd3d7eb4d9daa6f1b2fa67d)

The original fix for this on master was in #7604 however I think a lot of that code got reverted on master during prep for 9.2. In the mean time a fix went into 9.0/9.1 in #7602 . That fix was then never brought into master with the reverted code or on 9.2. So we need this both in master and picked back to 9.2

### DIFF
--- a/proxy/http/HttpSM.cc
+++ b/proxy/http/HttpSM.cc
@@ -5214,11 +5214,10 @@ HttpSM::do_http_server_open(bool raw)
       } else { // queue size is 0, always block.
         ct_state.blocked();
         HTTP_INCREMENT_DYN_STAT(http_origin_connections_throttled_stat);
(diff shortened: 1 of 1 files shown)


ezelkow1:
Also we are currently putting out builds to prod with this picked in (9.2 based builds) so we can hold off until we get some test time in production if we want more concrete evidence. The push behind this for us was our original 9.2 production boxes were getting sporadic crashes exactly like those reported in the original issue which led me down this path

vmamidi (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9066, opened 2022-08-26
Allow HEAD req to be served from cached GET

serrislew:
- Allows an exception to returning cached response with mismatched methods: replying a HEAD request with a cached GET response
  - neither are unsafe methods
  - behavior remains the same for other unsafe methods (POST, PUT, etc)
- Added autest for HEAD and GET behavior


zwoop:
Cherry-picked to v9.2.x

PR #8956, opened 2022-07-12
Use quiche as QUIC implementation

maskit:
This PR adds a build option (`--with-quiche`) and code to use quiche as QUIC implementation instead of the original our own QUIC implementation.

The code for the original implementation is kept for later use in case we want more flexibility. If you use BoringSSL or OpenSSL from quictls, the original one will be compiled but it won't be enabled until you explicitly configure ATS to use QUIC.

Also, this PR adds support for QUICv1 (0x00000001) and "h3", the official versions, and drops the support for draft-27. Support for draft-29 is kept for now like other applications do.

The code is compatible with quiche-0.14.0.

PR #9069, opened 2022-08-29
Disable the QUIC TestRun in active_timeout AuTest

bneradt:
We have added QUIC version 0x1 support to our 10-Dev branch via #8956,
but that has not been backported to 9.x and we don't plan to backport
that at this time. ASF CI will be updated with the latest version of
curl which uses 0x1 and would therefore fail. This change disables the
QUIC TestRun of active_timeout so it doesn't run on 9.x since ATS
doesn't support the latest QUIC test tools on that branch. This patch,
therefore, should not go into 10-Dev.

Note that this patch can be reverted in the future if we decide to
backport QUIC 0x1 support to 9.x.


maskit (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9072, opened 2022-08-29
Update expected ALPN output for recent versions of curl

bneradt:
Recent versions of curl have updated their ALPN negotiation logging.
This updates a gold file to accommodate that new output. The update
should be backwards compatible.


maskit (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9074, opened 2022-08-31
Updates to origin side ALPN configuration implementation

bneradt:
ATS will now emit an ERROR instead of a WARNING if there are issues with
the user configured ALPN string. Also, the ALPN parsing is updated to
use the more capable TextView API.

### DIFF
--- a/include/records/I_RecHttp.h
+++ b/include/records/I_RecHttp.h
@@ -542,5 +542,9 @@ extern void ts_session_protocol_well_known_name_indices_init();
  * @return True if the conversion was successful, false otherwise. Note that
  * the wire format does not support an empty protocol list, therefore this
  * function returns false if @a protocols is an empty string.
+ *
+ * TODO: ideally this would take a ts::TextView for @a protocols, but currently
+ * ts::TextView does not have a char* constructor while std::string_view does.
+ * Once that is added, this can be seemlessly switched to a ts::TextView.
  */
 bool convert_alpn_to_wire_format(std::string_view protocols, unsigned char *wire_format_buffer, int &wire_format_buffer_len);
--- a/src/records/RecHttp.cc
+++ b/src/records/RecHttp.cc
@@ -33,6 +33,8 @@
 #include <unordered_set>
 #include <tscore/IpMapConf.h>
 
+using ts::TextView;
+
 SessionProtocolNameRegistry globalSessionProtocolNameRegistry;
 
 /* Protocol session well-known protocol names.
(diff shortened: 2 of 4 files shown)


bneradt:
[approve ci debian]

include/records/I_RecHttp.h
```diff
@@ -542,5 +542,9 @@ extern void ts_session_protocol_well_known_name_indices_init();
  * @return True if the conversion was successful, false otherwise. Note that
  * the wire format does not support an empty protocol list, therefore this
  * function returns false if @a protocols is an empty string.
+ *
+ * TODO: ideally this would take a ts::TextView for @a protocols, but currently
+ * ts::TextView does not have a char* constructor while std::string_view does.
+ * Once that is added, this can be seemlessly switched to a ts::TextView.
```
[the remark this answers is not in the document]
bneradt: True, I considered that: we can explicitly convert at the call sites to this function. Either way we are stuck with something that is possibly temporary:

* Change the call sites to this function to explicitly convert to ts::TextView and remove them later when `char *` is added, or
* Leave this parameter interface as std::string_view and swap it for ts::TextView when a `char *` constructor is added.

Leaving this as `std::string_view` creates a smaller patch both here and when the updated ts::TextView constructor is added. Also it lets the call sites remain as is with less cruft.

src/records/RecHttp.cc
```diff
@@ -859,60 +865,58 @@ convert_alpn_to_wire_format(std::string_view protocols, unsigned char *wire_form
...
-      protocol = protocols.substr(pos);
-      pos      = protocols.size();
-    } else {
-      protocol = protocols.substr(pos, next_pos - pos);
-      pos      = next_pos + 1;
-    }
+
+  // Make sure there are no whitespace characters in the protocols string.
+  if (std::any_of(protocols.begin(), protocols.end(), [](char const c) { return isspace(c); })) {
+    Error("The configured ALPN string contains whitespace characters: %.*s", static_cast<int>(protocols.size()), protocols.data());
```
maskit, line 873: Current implementation is more accurate actually. Strictly speaking, a protocol id can have spaces and also commas.

bneradt: > Current implementation is more accurate actually. Strictly speaking, a protocol id can have spaces and also commas.

True, but the ones we support do not.

This comes from an internal review comment from @ywkaras . The advantage of this check is that it will give an explicit error message to someone who might try this:

```
CONFIG proxy.config.ssl.client.alpn_protocols STRING "h2, http/1.1"
```

Note the space after the comma. Without the space check from this patch, the user will get a generic `Unknown protocol name in configured ALPN list:  http/1.1`. This Error message will explicitly point out that a space is the issue.

maskit: I hope we won't need to support ids with spaces. This is one of reasons why I didn't want to use protocol ids.

It's easy to prohibit spaces, but if users might try that format, wouldn't it be more user friendly if we ignore spaces?

bneradt: Yeah, that's a good point. Let me try with trimming. TextView has some nice APIs to make that easy.

bneradt: I updated the parsing logic to allow the protocols to be padded with white space.

src/records/unit_tests/test_RecHttp.cc
```diff
@@ -110,6 +110,7 @@ struct ConvertAlpnToWireFormatTestCase {
 
 // clang-format off
 std::vector<ConvertAlpnToWireFormatTestCase> convertAlpnToWireFormatTestCases = {
+#if 0
```
maskit, line 113: Leftover?

maskit (approved):

ywkaras (approved):

PR #9079, opened 2022-09-05
Fix ATS original QUIC impl

maskit:
The change to use Quiche had a couple of changes in common part which is used by ATS original QUIC impl as well, and it broke the original impl.

The first commits is to apply the same change for UDP socket initialization to the original impl. 99% of the code is copy-and-paste.

The second commit is to deal with use-after-free caused by events for closed stream. The commit for quiche support had a fix for memory leak, and the original impl didn't handle closed stream appropriately.

The forth commit is to update unit tests.

And the last commit adds back a test for H3, which was temporally commented out because of the issues above.

### DIFF
--- a/iocore/net/QUICNetProcessor.cc
+++ b/iocore/net/QUICNetProcessor.cc
@@ -220,8 +220,5 @@ QUICNetProcessor::main_accept(Continuation *cont, SOCKET fd, AcceptOptions const
   na->action_->server = &na->server;
   na->init_accept();
 
-  SCOPED_MUTEX_LOCK(lock, na->mutex, this_ethread());
-  udpNet.UDPBind((Continuation *)na, &na->server.accept_addr.sa, fd, 1048576, 1048576);
-
   return na->action_.get();
 }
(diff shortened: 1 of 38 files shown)


maskit:
Confirmed the four dots.
> Running Test active_timeout:.... Passed

maskit:
> Can you explain how the Fix a use-after-free issue in H3 native impl change works? While the change is large, it looks like almost all of it is adding the owner parameter to a handful of functions. Where is owner used though?

It's used by LossDetector. LossDetector tells frame owners/generators whether frames were acked or lost so that owners can retransmit (regenerate) frames. And originally QUICStream was an owner of a couple of types of frames. It worked because we didn't delete closed streams before the change for Quiche, but now the streams are deleted, so LossDetector can't tell closed streams whether frames are acked/lost (it crashes when LossDetector tries to access deleted streams). The fix changes the owner of the frames that are generated by QUICStream to QUICStreamManager. QUICStreamManager checks if a stream still exists first, and then let the stream know whether frames are acked/lost only if the stream is available.

Similar things happen inside a QUICStream as well actually. QUICFlowController in QUICStream generates frames, and it was originally a frame owner, but, like streams, it gets unavailable when a stream is closed.

https://github.com/apache/trafficserver/blob/9a319bd4162758174b5af214818bc58cf462abdb/iocore/net/quic/QUICLossDetector.cc#L296-L310

PR #9086, opened 2022-09-12
Add gdb helpers script to help debugging

cmcfarlen:
This script adds a data command to gdb so you can print out HttpSM data.  This is a sort-of proof of concept for writing helper commands to help dig into ATS data structures.  We can add more if we like this approach.  I thought about using pretty-printers but those might be problematic so I went the custom command route so it doesn't otherwise interfere with debugging.

This requires at least gdb 10

### DIFF
--- a/tools/gdb-helpers.py
+++ b/tools/gdb-helpers.py
@@ -39,7 +39,11 @@ from curses.ascii import isgraph
 def ats_str(addr, addr_len):
     #print("addr {} len {}".format(addr, addr_len))
     inferior = gdb.selected_inferior()
-    return inferior.read_memory(addr, addr_len)
+    buff = inferior.read_memory(addr, addr_len)
+    if buff[0] == '\x00':
+        return 'null'
+    else:
+        return buff
 
 
 def hdrtoken(idx):
--- a/tools/gdb-helpers.py
+++ b/tools/gdb-helpers.py
@@ -226,7 +226,7 @@ commands = [
 
 
 def usage():
-    print("Usage: ats <command> <expr>")
+    print("Usage: atspr <command> <expr>")
     print("commands:")
     for cmd, f, desc in commands:
         print("  {}: {}".format(cmd, desc))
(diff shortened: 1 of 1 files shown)


cmcfarlen:
```
(gdb) atspr sm this
smid = 229501
---- request ----
GET http://127.0.0.1/path/to/thing
Host: 127.0.0.1
Accept: text/plain, text/*, */*
Accept-Encoding: gzip
Connection: close
---- response ----
status: 403
Content-Type: text/plain
Cache-Control: no-cache
Content-Length: 0
Date: Mon, 12 Sep 2022 16:19:42 GMT
Age: 0
Connection: close
X-Cache: none
---- server_request  ----
GET :///path/to/thing
Host: 127.0.0.1
Accept: text/plain, text/*, */*
Accept-Encoding: gzip
Connection: close
X-Forwarded-For: 127.0.0.1
Connection: close
---- server_response  ----
status: 403
Content-Type: text/plain
Cache-Control: no-cache
Content-Length: 0
Date: Mon, 12 Sep 2022 16:19:42 GMT
---- transform_response  ----
invalid
---- cache_response  ----
invalid
---- client_info  ----
HTTP 1.1 (127.0.0.1 -> 127.0.0.1)
---- server_info  ----
HTTP 1.1 (127.0.0.1 -> 127.0.0.1)
---- parent_info  ----
HTTP 1.1 (127.0.0.1 -> 127.0.0.1)
```

bneradt (approved):
Looks helpful. Thanks for starting this.

tools/gdb-helpers.py
```diff
@@ -0,0 +1,259 @@
...
+
+
+commands = [
+    ("sm", sm_command, "Print HttpSM details (type HttpSM*)"),
+    ("hdrs", hdrs_command, "Print HttpHdr details (type HTTPHdr)"),
+]
+
+
+def usage():
+    print("Usage: ats <command> <expr>")
```
shukitchan, line 229: atspr instead of ats ?

bneradt (approved):

PR #9096, opened 2022-09-15
Fast log buffer

cmcfarlen:
This adds a configurable fast mode to log objects that use per-thread log buffers rather than having all threads contend for per-object buffers. While this is much faster, it does not support ordered log entries. This makes it situationally useful, but if a user has binary logs or doesn't need ordered log output, this could allow them to log many more transactions. Its configurable either globally via records.config or per-log via logging.yaml.

### DIFF
--- a/doc/admin-guide/logging/understanding.en.rst
+++ b/doc/admin-guide/logging/understanding.en.rst
@@ -157,3 +157,19 @@ aforementioned aggregate functions and the specification of an interval, as so:
 The interval itself is given with *n* as the number of seconds for each period
 of aggregation. There is no default value.
 
+Logging Performance
+-------------------
+
+In normal operations, log entries are strictly ordered in the output file.
+This serialization of entries comes at a cost as multiple threads potentially
+contend for log access.  For binary logs or when order does not matter, ATS supports
+faster logging where each thread can buffer its own entries.  In this mode, log
+parsers will need to expect out of order entries, but ATS can log much larger
+transaction rates.
+
+.. code:: yaml
+
+   formats:
+   - name: mysummary
+     format: '%<operator(field)> , %<operator(field)>'
+     fast: true
--- a/doc/admin-guide/logging/understanding.en.rst
+++ b/doc/admin-guide/logging/understanding.en.rst
@@ -165,7 +165,8 @@ This serialization of entries comes at a cost as multiple threads potentially
 contend for log access.  For binary logs or when order does not matter, ATS supports
 faster logging where each thread can buffer its own entries.  In this mode, log
 parsers will need to expect out of order entries, but ATS can log much larger
-transaction rates.
+transaction rates. Consider adding fields to the log format that include the timestamp so
+entries can be reordered if necessary (see :ref:`admin-logging-fields-time`)
 
 .. code:: yaml
 
--- a/proxy/logging/LogBuffer.cc
+++ b/proxy/logging/LogBuffer.cc
@@ -192,6 +192,32 @@ LogBuffer::~LogBuffer()
   -------------------------------------------------------------------------*/
 
 LogBuffer::LB_ResultCode
+LogBuffer::add_entry(size_t *write_offset, size_t write_size)
+{
+  LB_ResultCode ret_val    = LB_OK;
+  size_t actual_write_size = INK_ALIGN(write_size + sizeof(LogEntryHeader), m_write_align);
+  if (m_state.s.offset + actual_write_size <= m_size) {
+    size_t offset = m_state.s.offset;
+
+    LogEntryHeader *entry_header = reinterpret_cast<LogEntryHeader *>(&m_buffer[offset]);
+    struct timeval tp            = ink_hrtime_to_timeval(Thread::get_hrtime());
+
+    entry_header->timestamp      = tp.tv_sec;
+    entry_header->timestamp_usec = tp.tv_usec;
+    entry_header->entry_len      = actual_write_size;
+
+    *write_offset = offset + sizeof(LogEntryHeader);
+
+    m_state.s.offset += actual_write_size;
+    m_state.s.num_entries++;
+  } else {
+    ret_val = LB_FULL_NO_WRITERS;
+  }
+
+  return ret_val;
+}
+
+LogBuffer::LB_ResultCode
 LogBuffer::checkout_write(size_t *write_offset, size_t write_size)
 {
   // checkout_write should not be called if m_unaligned_buffer was
@@ -321,7 +347,6 @@ LogBuffer::checkin_write(size_t write_offset)
     if (--new_s.s.num_writers == 0) {
       ret_val = (old_s.s.full ? LB_ALL_WRITERS_DONE : LB_OK);
     }
-
   } while (!switch_state(old_s, new_s));
 
   //    Debug("log-logbuffer","[%p] checkin_write for buffer %u (%s) "
--- a/proxy/logging/LogBuffer.cc
+++ b/proxy/logging/LogBuffer.cc
@@ -188,11 +188,14 @@ LogBuffer::~LogBuffer()
 }
 
 /*-------------------------------------------------------------------------
-  LogBuffer::checkout_write
+  LogBuffer::fast_write
+
+  This function is used for the thread-local 'fast' log buffer mode and is not thread-safe.
+  In 'fast' mode, each thread owns its own LogBuffer rather than multiple threads contending over one.
   -------------------------------------------------------------------------*/
 
 LogBuffer::LB_ResultCode
-LogBuffer::add_entry(size_t *write_offset, size_t write_size)
+LogBuffer::fast_write(size_t *write_offset, size_t write_size)
 {
   LB_ResultCode ret_val    = LB_OK;
   size_t actual_write_size = INK_ALIGN(write_size + sizeof(LogEntryHeader), m_write_align);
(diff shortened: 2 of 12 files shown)


proxy/logging/unit-tests/benchmark_LogObject.cc
```diff
@@ -0,0 +1,169 @@
...
+{
+  ink_freelist_init_ops(true, true);
+  init_buffer_allocators(0);
+
+  Thread *main_thread = new EThread;
+  main_thread->set_specific();
+
+  auto diagsConfig = new DiagsConfig("Server", "diags.log", "", "", false);
+  // unused, but constructor must be called for side effects.
+  (void)diagsConfig;
```
ywkaras, line 74: Why not just:
```
new DiagsConfig("Server", "diags.log", "", "", false);
```
?

proxy/logging/unit-tests/benchmark_LogObject.cc
```diff
@@ -0,0 +1,169 @@
...
+
+#include <thread>
+#include <condition_variable>
+#include <chrono>
+
+AppVersionInfo appVersionInfo;
+static char bind_stdout[512] = "";
+static char bind_stderr[512] = "";
+
+namespace notstd
```
ywkaras, line 42: Why not just use an anonymous namespace?

ywkaras:
What performance difference does your benchmark show?  How about a cut-n-paste of the output?

cmcfarlen:
```
benchmark name                       samples       iterations    estimated
logobject fast                                 100             1     50.7197 s
                                        497.339 ms    482.904 ms    502.998 ms
                                        43.2047 ms    20.2149 ms    91.1906 ms

logobject slow                                 100             1     11.9278 m
                                         5.55437 s     5.43973 s     5.68358 s
                                        619.759 ms    508.693 ms    748.277 ms
```

Here is the benchmark output.  The fast log buffers are about 10x faster.  The performance improvement greatly depends on the performance of the CAS instruction, which on intel is abysmal under high contention.  For AMD the perf difference is far less dramatic.  As an anecdotal difference, we were able to run full logging (without sampling) without a noticeable drop in requests per second and fairly instantly filled a 10gig RAM disk with "fast" log files.  Before this change, enabling logging would dramatically reduce our requests/second.

Also, I have to modify other parts of the code to run the benchmark since the JemallocNodumpAllocator creates (and leaks) thread local arenas, it cannot support as many thread starts as the benchmark spawns, but using the ink_freelist_* functions just adds another CAS bottleneck which ruins everything.  I've included a commit that replaces all of that with a straight malloc Allocator class in case you want to try to reproduce this, but I was going to add that as an optional allocator in a separate PR.  I'd like to remove  that from this PR before any potential merging of this code (assuming it's acceptable).

build/jemalloc.m4
```diff
@@ -58,8 +58,8 @@ if test "$has_jemalloc" != "no"; then
   if test "$jemalloc_base_dir" != "/usr"; then
     TS_ADDTO(CPPFLAGS, [-I${jemalloc_include}])
     TS_ADDTO(LDFLAGS, [-L${jemalloc_ldflags}])
-    TS_ADDTO(LDFLAGS, [-Wl,--add-needed -L${jemalloc_ldflags} -Wl,-rpath,${jemalloc_ldflags} -Wl,--no-as-needed])
-    TS_ADDTO_RPATH(${jemalloc_ldflags})
+  dnl  TS_ADDTO(LDFLAGS, [-Wl,--add-needed -L${jemalloc_ldflags} -Wl,-rpath,${jemalloc_ldflags} -Wl,--no-as-needed])
+  dnl  TS_ADDTO_RPATH(${jemalloc_ldflags})
```
ywkaras, line 62: Did you mean to include this, was it for testing only?

cmcfarlen: Oh, just testing so I could use a debug jemalloc.  I'll fix it. Thanks!

doc/admin-guide/logging/understanding.en.rst
```diff
@@ -157,3 +157,19 @@ aforementioned aggregate functions and the specification of an interval, as so:
...
+faster logging where each thread can buffer its own entries.  In this mode, log
+parsers will need to expect out of order entries, but ATS can log much larger
+transaction rates.
+
+.. code:: yaml
+
+   formats:
+   - name: mysummary
+     format: '%<operator(field)> , %<operator(field)>'
+     fast: true
```
ywkaras, line 175: Should we link to this section?  https://docs.trafficserver.apache.org/admin-guide/logging/formatting.en.html#timestamps-and-durations .  Since timestamping would be even more important if log record order corresponded less to request order.

proxy/logging/unit-tests/benchmark_LogObject.cc
```diff
@@ -0,0 +1,169 @@
+/** @file
```
ywkaras, line 1: How does this get built into an executable?  I can't see it in any makefile changes.

proxy/logging/LogBuffer.cc
```diff
@@ -191,6 +191,32 @@ LogBuffer::~LogBuffer()
   LogBuffer::checkout_write
   -------------------------------------------------------------------------*/
 
+LogBuffer::LB_ResultCode
+LogBuffer::add_entry(size_t *write_offset, size_t write_size)
```
ywkaras, line 195: Maybe call this fast_write()?  Add a comment that it's not thread safe and/or only for fast mode?

proxy/logging/LogObject.cc
```diff
@@ -477,6 +481,89 @@ LogObject::log(LogAccess *lad, const char *text_entry)
...
+      current_buffers[o] = buffer;
+      if (buffer->add_entry(offset, bytes_needed) != LogBuffer::LB_OK) {
+        return nullptr;
+      }
+    };
+    return buffer;
+  }
+
+private:
+  std::map<LogObject *, LogBuffer *> current_buffers;
```
ywkaras, line 544: Seems like this could be an unordered map.

ywkaras:
Has this been tested in production at all?

proxy/logging/Makefile.am
```diff
@@ -101,7 +101,9 @@ test_RolledLogDeleter_SOURCES = \
 test_RolledLogDeleter_LDADD = \
 	$(top_builddir)/src/tscore/libtscore.la \
 	$(top_builddir)/src/tscpp/util/libtscpputil.la \
-	$(top_builddir)/iocore/eventsystem/libinkevent.a
+	$(top_builddir)/iocore/eventsystem/libinkevent.a \
+	@HWLOC_LIBS@
```
ywkaras, line 105: Why do you need this?

proxy/logging/unit-tests/benchmark_LogObject.cc
cmcfarlen: I took it out of the build since I couldn't figure out how to make it not break the ci build.

doc/admin-guide/logging/understanding.en.rst
ywkaras: I'm wondering if we should mention that increasing this, https://docs.trafficserver.apache.org/admin-guide/files/records.config.en.html?highlight=records%20config#proxy-config-log-log-buffer-size, will increase the speedup seen from using `fast`.

proxy/logging/Makefile.am
cmcfarlen: Unintentionally added.  I'll fix it.

cmcfarlen:
This has only been tested in lab environments

proxy/logging/LogObject.cc
```diff
@@ -477,6 +481,89 @@ LogObject::log(LogAccess *lad, const char *text_entry)
...
+/*
+ * This will return a LogBuffer object that is per-LogObject and per-Thread.  This function will handle all of
+ * the details around flushing buffers to the preproc threads and periodically checking for idle buffers.
+ */
+LogBuffer *
+thread_local_buffer(LogObject *o, size_t *offset, size_t bytes_needed)
+{
+  thread_local ThreadLocalLogBufferManager manager;
+  return manager.current_buffer(o, offset, bytes_needed);
+}
```
ywkaras, line 556: I'm unapologetically anal about narrow scoping, so I suggest making this a public static member of ThreadLocalBufferManager, so that all other members can be private.

ywkaras:
Looks good except for nitpicks.

proxy/logging/unit-tests/benchmark_LogObject.cc
ywkaras: I concur, there's not much value in having benchmarks in the CI runs.  Can you add a short comment that explains how to build it as a stand-alone executable?

proxy/logging/LogObject.cc
ywkaras: Why is it better as an ordered map?

ywkaras: Since logging is per-transaction, seems like it's worth making the most optimal choice

ywkaras: My other comments are about judgement calls, and I don't use reviews to second guess, but I think it's worth knowing your reasons here.

cmcfarlen: LogObjects are created at startup per log, and then more won't be created.  Therefore, this map will just grow to a small fixed size and then only the buffer pointer will change as log buffers are filled up.  My presumption is the overhead of finding the entry will be negligible with either tree or table storage. I briefly considered a vector to store them since the lookup would be practically constant time.

cmcfarlen: If you prefer an unordered_map, I don't mind changing it to that.

ywkaras (dismissed):

proxy/logging/LogObject.cc
ywkaras: Sounds reasonable, I'll approve as soon as the docs problem is resolved.

ywkaras:
[approve ci docs]

ywkaras:
[approve ci]

ywkaras (approved):

PR #9098, opened 2022-09-16
Add URLImpl::get_scheme

maskit:
`URLImpl::get_scheme` doesn't exist somehow where there are accessors for other parts, and `URL` has to touch member variables in `URLImpl` directly because of that.

This add `URLImpl::get_scheme` and move the current implantation in `URL` into `URLImpl`. There should be no logic/behavior change.


bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

Issue #9094, opened 2022-09-15
Unexpected behaviour when parsing requests without a leading /


jvgutierrez:
We are experiencing an unexpected behavior when requests reach ATS without a leading /. An easy PoC: `curl --http1.1 http://127.0.0.1:3128 --request-target favicon.ico`. Per my current understanding of [RFC 9112 Section 3.2 Request Target](https://www.rfc-editor.org/rfc/rfc9112.html#name-request-target), this should result in a 400 response cause request-target doesn't comply with any of the expected forms:
```
  request-target = origin-form
                 / absolute-form
                 / authority-form
                 / asterisk-form
```
origin-form is defined like this:
```
origin-form    = absolute-path [ "?" query ] # https://www.rfc-editor.org/rfc/rfc9112.html#name-origin-form
absolute-path  = 1*( "/" segment ) # https://www.rfc-editor.org/rfc/rfc9110.html#name-uri-references
```
and the RFC explicitly mentions `When making a request directly to an origin server, other than a CONNECT or server-wide OPTIONS request (as detailed below), a client MUST send only the absolute path and query components of the target URI as the request-target.`

The other kind of request targets are defined like this:
```
  absolute-form   = absolute-URI # https://www.rfc-editor.org/rfc/rfc9112.html#name-absolute-form
  absolute-URI    = scheme ":" hier-part [ "?" query ] # https://www.rfc-editor.org/rfc/rfc3986.html#section-4.3
  authority-form  = uri-host ":" port # https://www.rfc-editor.org/rfc/rfc9112.html#name-authority-form
  asterisk-form   = "*" # https://www.rfc-editor.org/rfc/rfc9112.html#name-asterisk-form
```

What we are seeing with both ATS 9.1.3 and ATS 8.0.8 is that ATS takes this request as "absolute-form" (thanks @maskit for pointing it out) and as defined in [RFC 9112 Section 3.2.2 absolute-form ](https://www.rfc-editor.org/rfc/rfc9112.html#name-absolute-form)ATS ignores the received Host header and replaces it with the host information of the request target.

This results in the following request reaching the origin server:
```
GET / HTTP/1.1
Host: favicon.ico
User-Agent: curl/7.64.0
Accept: */*
```

I've tested at least two other major HTTP proxy implementations and both of them generate the following request for the origin server:
```
GET favicon.ico HTTP/1.1
Host: 127.0.0.1:3128
User-Agent: curl/7.64.0
Accept: */*
```

[...]

PR #9100, opened 2022-09-20
Validate request-target

maskit:
This should fixes https://github.com/apache/trafficserver/issues/9094.

The validation is not very strict. It still allows "authority-form" that doesn't have a port number, but I think it's enough for #9094 and I didn't want to make it overly strict.

### DIFF
--- a/proxy/hdrs/HTTP.cc
+++ b/proxy/hdrs/HTTP.cc
@@ -1000,6 +1000,9 @@ http_parser_parse_req(HTTPParser *parser, HdrHeap *heap, HTTPHdrImpl *hh, const 
                                           false, max_hdr_field_size);
       // If we're done with the main parse do some validation
       if (ret == PARSE_RESULT_DONE) {
+        ret = validate_hdr_request_target(HTTP_WKSIDX_GET, url);
+      }
+      if (ret == PARSE_RESULT_DONE) {
         ret = validate_hdr_host(hh); // check HOST header
       }
       if (ret == PARSE_RESULT_DONE) {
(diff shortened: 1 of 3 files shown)


proxy/hdrs/HTTP.cc
```diff
@@ -1163,6 +1169,37 @@ http_parser_parse_req(HTTPParser *parser, HdrHeap *heap, HTTPHdrImpl *hh, const
...
+  url->get_host(&host_len);
+  int path_len;
+  const char *path = url->get_path(&path_len);
+  int scheme_len;
+  url->get_scheme(&scheme_len);
+
+  if (host_len == 0) {
+    if (path_len == 1 && path[0] == '*') { // asterisk-form
+      // Skip this check for now because URLImpl can't distinguish '*' and '/*'
+      // if (method_wk_idx != HTTP_WKSIDX_OPTIONS) {
```
maskit, line 1186: I have an additional patch to enable this check, but I excluded it because some people may not like it.

proxy/hdrs/HTTP.cc
```diff
@@ -1163,6 +1169,37 @@ http_parser_parse_req(HTTPParser *parser, HdrHeap *heap, HTTPHdrImpl *hh, const
...
+      // }
+    } else { // origin-form
+      // Nothing to check here
+    }
+  } else if (scheme_len == 0 && host_len != 0) { // authority-form
+    if (method_wk_idx != HTTP_WKSIDX_CONNECT) {
+      ret = PARSE_RESULT_ERROR;
+    }
+  } else { // absolute-form
+    // Nothing to check here
```
jvgutierrez, line 1197: maybe it's redundant but could we double check here that `path[0] == '/'` as it's the very defining characteristic of the absolute-form per RFC 9112?

jvgutierrez: oh, understood!

jvgutierrez (approved):
looks good, tested in our environment after backporting this PR and some dependencies to 9.1.3

bryancall:
Cherry picked to 9.1.x

zwoop:
Cherry-picked to v9.2.x

PR #9102, opened 2022-09-22
Move plugin_init for verify cmd

ezelkow1:
plugin_init calls api_init which will intialize all the global hooks needed for lifecycle and other plugins. Without doing this those hook objects are empty so if a plugin (such as lua) makes a call to add a lifecycle hook on remapinit, this will crash since they haven't been initialized. This occurs during the urlRewriteVerify step of verification where it tries to load plugins or at least initialize them, these will crash

### DIFF
--- a/src/traffic_server/traffic_server.cc
+++ b/src/traffic_server/traffic_server.cc
@@ -895,6 +895,13 @@ cmd_verify(char * /* cmd ATS_UNUSED */)
     Layout::get()->update_sysconfdir(conf_dir);
   }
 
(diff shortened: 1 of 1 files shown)


ezelkow1:
[approve ci]

ezelkow1:
[approve ci]

ezelkow1:
[approve ci]

randall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9107, opened 2022-09-26
Adds support for serving statichit content out of a directory

zwoop:
This does some fairly expensive computations when producing the directory paths. This is intentional,
to make sure requests aren't escaping the specified directory and out elsewhere. This cleans out abusive
requests such as get `/../../etc/hosts` etc.

The idea is that rather than having to specify every individual file that the server should service, you can specify a
directory and every file (and subdirectory thereunder) can now be served.

### DIFF
--- a/plugins/experimental/statichit/statichit.cc
+++ b/plugins/experimental/statichit/statichit.cc
@@ -90,8 +90,8 @@ struct StaticHitConfig {
 
   ~StaticHitConfig() { TSContDestroy(cont); }
 
-  std::string
-  makePath(TSHttpTxn txnp) const
+  std::string_view
+  makePath(TSHttpTxn txnp, std::string &output) const
   {
     if (!dirPath.empty()) {
       TSMBuffer reqp;
@@ -110,19 +110,20 @@ struct StaticHitConfig {
 
           if (std::equal(dirPath.begin(), dirPath.end(), requested_file_path.begin()) &&
               std::filesystem::is_regular_file(requested_file_path)) {
-            return requested_file_path;
+            output = requested_file_path.string();
+            return {output.c_str(), output.size()};
           } else {
-            return "";
+            return {};
           }
         } else {
           TSHandleMLocRelease(reqp, TS_NULL_MLOC, hdr_loc);
-          return "";
+          return {};
         }
       } else {
-        return "";
+        return {};
       }
     } else {
-      return filePath;
+      return {filePath};
     }
   }
 
@@ -229,12 +230,12 @@ struct StaticHitRequest {
   {
     StaticHitRequest *shr = new StaticHitRequest;
     std::ifstream ifstr;
+    std::string output;
+    std::string_view filePath = tc->makePath(txn, output);
 
-    auto filePath = tc->makePath(txn);
+    VDEBUG("Requested file path: %s", filePath.data());
 
-    VDEBUG("Requested file path: %s", filePath.c_str());
-
-    ifstr.open(filePath);
+    ifstr.open(filePath.data());
     if (!ifstr) {
       shr->statusCode = tc->failureCode;
       return shr;
--- a/plugins/experimental/statichit/statichit.cc
+++ b/plugins/experimental/statichit/statichit.cc
@@ -65,26 +65,21 @@ static int StaticHitTxnHook(TSCont contp, TSEvent event, void *edata);
 struct StaticHitConfig {
   explicit StaticHitConfig(const std::string &fileOrDir, const std::string &mimeType, bool exact) : mimeType(mimeType)
   {
-    std::filesystem::path base_path;
+    std::filesystem::path base_path{fileOrDir};
 
-    if (fileOrDir.find('/') != 0) {
-      base_path = std::filesystem::weakly_canonical(std::string(TSConfigDirGet()) + "/" + fileOrDir);
-    } else {
-      base_path = std::filesystem::weakly_canonical(fileOrDir);
+    if (!base_path.is_absolute()) {
+      base_path = std::filesystem::path(TSConfigDirGet()) / base_path;
     }
+    base_path = std::filesystem::weakly_canonical(base_path);
 
     if (std::filesystem::is_directory(base_path)) {
       dirPath      = base_path;
       filePath     = "";
       disableExact = true;
-    } else if (std::filesystem::is_regular_file(base_path)) {
+    } else {
       dirPath      = "";
       filePath     = base_path;
       disableExact = exact;
-    } else {
-      VERROR("Invalid file path: %s", filePath.c_str());
-      filePath = "";
-      dirPath  = "";
     }
   }
 
@@ -93,6 +88,8 @@ struct StaticHitConfig {
   std::string_view
   makePath(TSHttpTxn txnp, std::string &output) const
   {
+    std::string_view ret = {};
+
     if (!dirPath.empty()) {
       TSMBuffer reqp;
       TSMLoc hdr_loc = nullptr, url_loc = nullptr;
@@ -111,20 +108,17 @@ struct StaticHitConfig {
           if (std::equal(dirPath.begin(), dirPath.end(), requested_file_path.begin()) &&
               std::filesystem::is_regular_file(requested_file_path)) {
             output = requested_file_path.string();
-            return {output.c_str(), output.size()};
-          } else {
-            return {};
+            ret    = {output.c_str(), output.size()};
           }
         } else {
           TSHandleMLocRelease(reqp, TS_NULL_MLOC, hdr_loc);
-          return {};
         }
-      } else {
-        return {};
       }
     } else {
-      return {filePath};
+      ret = {filePath};
     }
+
+    return ret;
   }
 
   std::filesystem::path dirPath;
--- a/plugins/experimental/statichit/statichit.cc
+++ b/plugins/experimental/statichit/statichit.cc
@@ -33,6 +33,7 @@
 #include <sstream>
 
 #include <string>
+#include <filesystem>
 #include <getopt.h>
 
 #include <netinet/in.h>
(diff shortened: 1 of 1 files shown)


randall:
[approve ci]

bryancall:
@zwoop You should really add labels to the PRs.

zwoop:
As a side note for the curious reader, if this is acceptable by the community, I'd like to add support for the HTTPD `mime.types` file, allowing the mime type to be defined by the suffix of the filename.

plugins/experimental/statichit/statichit.cc
```diff
@@ -62,21 +63,80 @@ static int StaticHitInterceptHook(TSCont contp, TSEvent event, void *edata);
...
+      VERROR("Invalid file path: %s", filePath.c_str());
+      filePath = "";
+      dirPath  = "";
+    }
   }
 
   ~StaticHitConfig() { TSContDestroy(cont); }
 
+  std::string_view
+  makePath(TSHttpTxn txnp, std::string &output) const
```
ywkaras, line 94: Why not just have this return std::string and eliminate the output parameter?  C++98 is so five minutes ago.

plugins/experimental/statichit/statichit.cc
```diff
@@ -62,21 +63,80 @@ static int StaticHitInterceptHook(TSCont contp, TSEvent event, void *edata);
...
+  }
+
+  std::filesystem::path dirPath;
   std::string filePath;
-  std::string mimeType;
 
-  int successCode = 200;
-  int failureCode = 404;
-  int maxAge      = 0;
+  std::string mimeType = "";
```
ywkaras, line 133: Doesn't look like this is used.

plugins/experimental/statichit/statichit.cc
```diff
@@ -62,21 +63,80 @@ static int StaticHitInterceptHook(TSCont contp, TSEvent event, void *edata);
 static int StaticHitTxnHook(TSCont contp, TSEvent event, void *edata);
 
 struct StaticHitConfig {
-  explicit StaticHitConfig(const std::string &filePath, const std::string &mimeType, bool disableExact)
-    : filePath(filePath), mimeType(mimeType), disableExact(disableExact)
+  explicit StaticHitConfig(const std::string &fileOrDir, const std::string &mimeType, bool exact) : mimeType(mimeType)
```
SolidWallOfCode, line 66: Use `explicit` only on constructors that have exactly one parameter.

plugins/experimental/statichit/statichit.cc
```diff
@@ -62,21 +63,80 @@ static int StaticHitInterceptHook(TSCont contp, TSEvent event, void *edata);
...
   {
+    std::filesystem::path base_path;
+
+    if (fileOrDir.find('/') != 0) {
+      base_path = std::filesystem::weakly_canonical(std::string(TSConfigDirGet()) + "/" + fileOrDir);
```
SolidWallOfCode, line 71: No - use `std::filesystem::operator /`. E.g.
```
base_path = std::filesystem::weakly_canonical(std::filesystem::path(TSConfigDirGet()) / fileOrDir);
```
That operator takes care of the little details like trailing / leading separators, etc.

plugins/experimental/statichit/statichit.cc
```diff
@@ -62,21 +63,80 @@ static int StaticHitInterceptHook(TSCont contp, TSEvent event, void *edata);
...
+      disableExact = true;
+    } else if (std::filesystem::is_regular_file(base_path)) {
+      dirPath      = "";
+      filePath     = base_path;
+      disableExact = exact;
+    } else {
+      VERROR("Invalid file path: %s", filePath.c_str());
+      filePath = "";
+      dirPath  = "";
+    }
```
SolidWallOfCode, line 88: If you're going to do multiple checks, it's better to grab the status and check that.
```
std::error_code ec;
auto stat = std::filesystem::status(base_path, ec);
if (ec) {
  return swoc::Errata("{} is not a valid path - {}", base_path, ec);
} else if (std::filesystem::is_directory(stat)) {
  // ...
} else if (std::filesystem::is_regular_file(stat)) {
 // ...
}

File: plugins/experimental/statichit/statichit.cc (hunk 1 above)
Comment (SolidWallOfCode) [reply plugins/experimental/statichit/statichit.cc:94]: To avoid a copy when it's not needed.

File: plugins/experimental/statichit/statichit.cc [hunk 6]
```diff
@@ -62,21 +63,80 @@ static int StaticHitInterceptHook(TSCont contp, TSEvent event, void *edata);
...
   }
 
   ~StaticHitConfig() { TSContDestroy(cont); }
 
+  std::string_view
+  makePath(TSHttpTxn txnp, std::string &output) const
+  {
+    if (!dirPath.empty()) {
```
SolidWallOfCode, line 96: If this is a member variable, you need to either use leading underscores or `this->dirPath` so it's clear where the value is from.

plugins/experimental/statichit/statichit.cc
```diff
@@ -62,21 +63,80 @@ static int StaticHitInterceptHook(TSCont contp, TSEvent event, void *edata);
...
+          int path_len = 0;
+          auto path    = TSUrlPathGet(reqp, url_loc, &path_len);
+
+          std::filesystem::path requested_file_path(
+            std::filesystem::weakly_canonical(dirPath / std::string_view{path, static_cast<size_t>(path_len)}));
+
+          TSHandleMLocRelease(reqp, hdr_loc, url_loc);
+          TSHandleMLocRelease(reqp, TS_NULL_MLOC, hdr_loc);
+
+          if (std::equal(dirPath.begin(), dirPath.end(), requested_file_path.begin()) &&
```
SolidWallOfCode, line 111: `std::string::starts_with` would be clearer.
https://en.cppreference.com/w/cpp/string/basic_string/starts_with

plugins/experimental/statichit/statichit.cc
zwoop: Changed.

plugins/experimental/statichit/statichit.cc
zwoop: I didn't change this ....

plugins/experimental/statichit/statichit.cc
```diff
@@ -62,21 +63,80 @@ static int StaticHitInterceptHook(TSCont contp, TSEvent event, void *edata);
...
-    : filePath(filePath), mimeType(mimeType), disableExact(disableExact)
+  explicit StaticHitConfig(const std::string &fileOrDir, const std::string &mimeType, bool exact) : mimeType(mimeType)
   {
+    std::filesystem::path base_path;
+
+    if (fileOrDir.find('/') != 0) {
+      base_path = std::filesystem::weakly_canonical(std::string(TSConfigDirGet()) + "/" + fileOrDir);
+    } else {
+      base_path = std::filesystem::weakly_canonical(fileOrDir);
+    }
```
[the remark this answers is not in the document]
zwoop: But I don't think that's the same. The directory can be e.g. foo/bar, which is not absolute.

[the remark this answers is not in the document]
zwoop: Fine ...

plugins/experimental/statichit/statichit.cc
zwoop: Meh

plugins/experimental/statichit/statichit.cc
zwoop: Just staying with the way the original plugin was written, where public member variables are not underscored, and we do not explicitly specify this->

plugins/experimental/statichit/statichit.cc
```diff
@@ -62,21 +63,80 @@ static int StaticHitInterceptHook(TSCont contp, TSEvent event, void *edata);
...
+
+          TSHandleMLocRelease(reqp, hdr_loc, url_loc);
+          TSHandleMLocRelease(reqp, TS_NULL_MLOC, hdr_loc);
+
+          if (std::equal(dirPath.begin(), dirPath.end(), requested_file_path.begin()) &&
+              std::filesystem::is_regular_file(requested_file_path)) {
+            output = requested_file_path.string();
+            return {output.c_str(), output.size()};
+          } else {
+            return {};
```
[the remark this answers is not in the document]
zwoop: Fixed.

plugins/experimental/statichit/statichit.cc
zwoop: That's C++20 only for std::string

plugins/experimental/statichit/statichit.cc
zwoop: Also, this was changed in C++11.

plugins/experimental/statichit/statichit.cc
```diff
@@ -62,21 +63,74 @@ static int StaticHitInterceptHook(TSCont contp, TSEvent event, void *edata);
...
+              std::filesystem::is_regular_file(requested_file_path)) {
+            output = requested_file_path.string();
+            ret    = {output.c_str(), output.size()};
+          }
+        } else {
+          TSHandleMLocRelease(reqp, TS_NULL_MLOC, hdr_loc);
+        }
+      }
+    } else {
+      ret = {filePath};
```
ywkaras, line 118: I guess clang doesn't like the braces cause you're using a conversion operator not a constructor:  https://en.cppreference.com/w/cpp/string/basic_string/operator_basic_string_view

zwoop: Gah, thanks. I don't think it's a "clang" issue per se, it obviously worked with some clang compilers. Will look again when I get back.

zwoop:
[approve ci ubuntu]

plugins/experimental/statichit/statichit.cc
```diff
@@ -62,21 +63,74 @@ static int StaticHitInterceptHook(TSCont contp, TSEvent event, void *edata);
...
+          int path_len = 0;
+          auto path    = TSUrlPathGet(reqp, url_loc, &path_len);
+
+          std::filesystem::path requested_file_path(
+            std::filesystem::weakly_canonical(dirPath / std::string_view{path, static_cast<size_t>(path_len)}));
+
+          TSHandleMLocRelease(reqp, hdr_loc, url_loc);
+          TSHandleMLocRelease(reqp, TS_NULL_MLOC, hdr_loc);
+
+          if (std::equal(dirPath.begin(), dirPath.end(), requested_file_path.begin()) &&
```
ywkaras, line 108: Do we need a case insensitive comparison for OSX?

plugins/experimental/statichit/statichit.cc
```diff
@@ -62,21 +63,74 @@ static int StaticHitInterceptHook(TSCont contp, TSEvent event, void *edata);
...
+  }
+
+  std::filesystem::path dirPath;
   std::string filePath;
-  std::string mimeType;
 
-  int successCode = 200;
-  int failureCode = 404;
-  int maxAge      = 0;
+  std::string mimeType = "";
```
ywkaras, line 127: "" is the default value for std::string anyway.

plugins/experimental/statichit/statichit.cc
```diff
@@ -166,12 +220,16 @@ struct StaticHitRequest {
...
-  createStaticHitRequest(StaticHitConfig *tc)
+  createStaticHitRequest(StaticHitConfig *tc, TSHttpTxn txn)
   {
     StaticHitRequest *shr = new StaticHitRequest;
     std::ifstream ifstr;
+    std::string output;
+    std::string_view filePath = tc->makePath(txn, output);
 
-    ifstr.open(tc->filePath);
+    VDEBUG("Requested file path: %s", filePath.data());
```
ywkaras, line 230: You may spend centuries in Purgatory for assuming an std::string_view is null terminated.  Especially when you could just have makePath() return `std::string const &`.

ywkaras (approved):
Please never tell Bjarne Stroustrup I approved this.

zwoop:
Cherry-picked to v9.2.x

PR #9108, opened 2022-09-27
Allocator that just uses malloc and bypasses custom jemalloc/freelist

cmcfarlen:
This PR reimplements Allocator to just use malloc/free and completely ignores the ink_freelist code. It doesn't disable ProxyAllocator freelists. This means the -F (proxy allocator disable) works, but -f (disable freelist) does nothing.

This differs from the -f option (when jemalloc is enabled) in the following ways:

 - honors advice and calls (jemalloc's) madvise (for nodump support on io buffers).
 - Does not use the jemallocnodumallocator for aligned allocation, but uses (jemalloc's) aligned_alloc.
 - does not use posix_madvise, or posix_memalign in any case

To use this allocator, one must pass `--enable-malloc-allocator` to `configure` otherwise it will behave as it always has.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1825,6 +1825,17 @@ TS_CHECK_JEMALLOC
 TS_CHECK_MIMALLOC
 
 #
+# Check whether to use malloc or freelist allocator
+AC_MSG_CHECKING([Use malloc allocator])
+AC_ARG_ENABLE([malloc-allocator],
+  [AS_HELP_STRING([--enable-malloc-allocator],[Allocator uses malloc])],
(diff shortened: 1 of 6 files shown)


cmcfarlen:
Yes, this shows a modest performance increase over `-f` but a huge increase over using ink_freelist* yet still honors madvise.  Basically it is totally relying on the malloc implementation to do something performant and allows full tuning to that end.

Since this just uses the standard C memory calls, it would work with mimalloc as well but might have to include the overrides header.

duke8253 (approved):
looks good to me.

PR #9109, opened 2022-09-27
Change use of std::shared_mutex to ts::shared_mutex.

ywkaras:
Typically when shared_mutex is used, the ratio of shared_locks versus unique_locks is high, so writer starvation is a risk.  On Linux, ts::shared_mutex avoids writer starvation (std::shared_mutex does not).

### DIFF
--- a/iocore/net/SSLSessionCache.cc
+++ b/iocore/net/SSLSessionCache.cc
@@ -25,6 +25,7 @@
 
 #include <cstring>
 #include <memory>
+#include <shared_mutex>
 
 #define SSLSESSIONCACHE_STRINGIFY0(x) #x
 #define SSLSESSIONCACHE_STRINGIFY(x) SSLSESSIONCACHE_STRINGIFY0(x)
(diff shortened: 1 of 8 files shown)


iocore/net/SSLSessionCache.cc
```diff
@@ -25,6 +25,7 @@
 
 #include <cstring>
 #include <memory>
+#include <shared_mutex>
```
masaori335, line 28: `<tscpp/util/TsSharedMutex.h>`?

iocore/net/TLSKeyLogger.cc
```diff
@@ -23,6 +23,7 @@
 #include "tscore/Diags.h"
 
 #include <cstring>
+#include <shared_mutex>
```
masaori335, line 26: Why do we need this? This PR is replacing all `std::shared_mutex`, right?

masaori335: nvm, TLSKeyLogger used `std::shared_lock`.

iocore/net/SSLSessionCache.cc
masaori335: nvm. SSLSessionCache.cc has `std::shared_lock`.

masaori335 (approved):
Looks reasonable.

PR #9117, opened 2022-10-01
Fix HTTP/2 session receive window handling for small sizes

### LINKED ISSUE #9115: Improper HTTP/2 enforcement of a reduced session window size
While adding test coverage for #9085, I noticed that we have issues handling multiplexed HTTP/2 DATA frames on the client side as we handle translating the transactions toward the origin. I forced the client to send DATA continuation frames by reducing the stream window to 10 bytes. The client (Proxy Verifier) multiplexes three transactions per this window. The client and ATS handle this fine. Toward the origin, I notice that ATS sends the first request headers and the first 10 byte body as a chunked body. This 10 byte body corresponds to the first DATA frame content. The problem is that ATS does not continue to stream this first request until chunk termination. Instead, ATS then sends the headers for the second request. This is, naturally, not acceptable for HTTP/1 since there is no concept of multiplexing like that for HTTP/1.

I have an AuTest that reproduces this. I'll attach to this issue:

1. The AuTest.
2. The Proxy Verifier replay file used by the test to generate the HTTP traffic.
3. A tcpdump I captured while running the test. Port 2007 is client-side HTTP/2, port 2002 is the client side HTTP/1.1 traffic.
4. A TLS keylog file to decrypt the tcpdump on the client side (the origin side is plaintext HTTP/1.1).

Looking at the pcap file:
* Note that the transaction has an initial GET request for path `/zero-request`. Ignore this.
* Frame 55 contains the headers for the first relevant POST request. Wireshark doesn't parse it because the stream is ultimately corrupted per the bug recorded in this issue.
* Frame 57 contains the 10 byte chunk for the first request. It correctly has no chunk terminator because further bytes should be forthcoming.
* Frame 61 incorrectly contains the header bytes for the second POST request. This is malformed HTTP/1.1 as an HTTP/1.1 parser would be expecting more chunk bytes, not header bytes of another request.

[tls_session_keys.txt](https://github.com/apache/trafficserver/files/9679637/tls_session_keys.txt)
[h2_to_h1_malformed.pcap.gz](https://github.com/apache/trafficserver/files/9679643/h2_to_h1_malformed.pcap.gz)
[http2_flow_control.replay.yaml.gz](https://github.com/apache/trafficserver/files/9679648/http2_flow_control.replay.yaml.gz)
[http2_flow_control.test.py.gz](https://github.com/apache/trafficserver/files/9679649/http2_flow_control.test.py.gz)

bneradt:
This fixes the way we handle session window sizes if the user configures proxy.config.http2.initial_window_size_in to be smaller than the HTTP/2 protocol's default value of 65,535. For streams, we can communicate to the client a smaller stream size via a SETTINGS frame, but not so for smaller session window sizes. The way HTTP/2 hosts are expected to reduce their receive session window size is by receiving DATA frames without replying with WINDOW_UPDATE frames until the window falls to the desired value. ATS already handled this latter behavior correctly (letting the window size fall without sending WINDOW_UPDATE frames for the session until it reached the desired smaller value). However, it incorrectly reported an error and sent a GOAWAY frame when the user sent DATA content above the reduced window size before ATS gave the client a chance to reduce it via DATA frames. This fixes the receive session window check to allow for a shrinking session window.

Fixes #9115


proxy/http2/Http2ConnectionState.cc
```diff
@@ -1039,8 +1039,18 @@ Http2ConnectionState::Http2ConnectionState() : stream_list()
...
 {
-  session            = ssn;
-  this->_server_rwnd = Http2::initial_window_size;
+  session = ssn;
+
+  if (Http2::initial_window_size < HTTP2_INITIAL_WINDOW_SIZE) {
+    // There is no HTTP/2 specified way to shrink the connection window size
+    // other than to receive data and not send WINDOW_UPDATE frames for a
+    // while.
+    this->_server_rwnd              = HTTP2_INITIAL_WINDOW_SIZE;
```
maskit, line 1048: Why do we have the lower limit here? The point of having `_server_rwnd_is_shrinking` is to allow setting small session window size, right?

bneradt: `HTTP2_INITIAL_WINDOW_SIZE` is actually the 65,535 value, so this isn't the lower but rather the higher value. So here's the function:

```
 if (Http2::initial_window_size < HTTP2_INITIAL_WINDOW_SIZE) {
     // There is no HTTP/2 specified way to shrink the connection window size
     // other than to receive data and not send WINDOW_UPDATE frames for a
     // while.
     this->_server_rwnd              = HTTP2_INITIAL_WINDOW_SIZE;
     this->_server_rwnd_is_shrinking = true;
   } else {
     this->_server_rwnd              = Http2::initial_window_size;
     this->_server_rwnd_is_shrinking = false;
   }
```

* If, per the first branch, the user configured a _smaller_ value than the spec's default session window size (`HTTP2_INITIAL_WINDOW_SIZE`, 65,535), then we initialize the window to that higher value and indicate that we are actively shrinking the window to the desired state. `restart_receiving` already handles this appropriately without any change in this patch...it sends `WINDOW_UPDATE` frames when the receive session window drops below the configured session window (not this default window).
* If the user configures a session window _larger_ than the default window size of 65,535 (the else case above), then we can simply set the `_server_rwnd` to the larger value and, later in the function, we send a `WINDOW_UPDATE` frame to tell the peer to increase the session window size as desired.

bneradt:
[approve ci autest]

proxy/http2/Http2ConnectionState.cc
bneradt: Thank you for the review. Let me know whether I misunderstood your question.

maskit: Ah, I got it now. You're right. `_server_rwnd` is *current* window size and it has to be 64KB at the beginning by the spec. I was confused and thought it's the desired window size.

maskit (approved):

maskit:
We should probably back port this to 9.2.x.

zwoop:
@masaori335 This has some merge conflicts, possibly with something else you may have fixed which didn't go into 9.2.x? Can you take a look please?

PR #9121, opened 2022-10-04
Reply with a 503 response when per_server.connection.max is exceeded

### LINKED ISSUE #9120: 10-Dev: proxy.config.http.per_server.connection.max not returning with 503
The per_server_connection_max.test.py is failing on CI for 10-Dev. This is a new test which was added on master and was recently merged into 10-Dev. It turns out that ATS is not replying with a 503 when the maximum number of connections are reached.

bneradt:
When the number of origin connections to a server exceeds proxy.config.http.per_server.connection.max, we should reply with a 503. This is done on master, but after a refactoring in 10-Dev, this was missed. This adds the call back in that replies with a 503 when the max is reached.

Fixes #9120

---

Note, this is only needed for, and only applies to, 10-Dev and should not be backported.


shinrich (approved):
LGTM

bneradt:
[approve ci autest]

bneradt:
[approve ci autest]

PR #9122, opened 2022-10-05
Fix HTTP/2 session receive window handling for small sizes (#9117)

### LINKED ISSUE #9115: Improper HTTP/2 enforcement of a reduced session window size

maskit:
Back porting  #9117 to 9.2.x. I had to make `server_rwnd_is_shrinking` public member to resolve conflicts.

----

This fixes the way we handle session window sizes if the user configures proxy.config.http2.initial_window_size_in to be smaller than the HTTP/2 protocol's default value of 65,535. For streams, we can communicate to the client a smaller stream size via a SETTINGS frame, but not so for smaller session window sizes. The way HTTP/2 hosts are expected to reduce their receive session window size is by receiving DATA frames without replying with WINDOW_UPDATE frames until the window falls to the desired value. ATS already handled this latter behavior correctly (letting the window size fall without sending WINDOW_UPDATE frames for the session until it reached the desired smaller value). However, it incorrectly reported an error and sent a GOAWAY frame when the user sent DATA content above the reduced window size before ATS gave the client a chance to reduce it via DATA frames. This fixes the receive session window check to allow for a shrinking session window.

Fixes #9115

(cherry picked from commit 4fcb9b47dbe549901141ebeedd09f389f55be6e1)

 Conflicts:
	proxy/http2/Http2ConnectionState.cc

### DIFF
--- a/proxy/http2/Http2ConnectionState.cc
+++ b/proxy/http2/Http2ConnectionState.cc
@@ -158,7 +158,7 @@ rcv_data_frame(Http2ConnectionState &cstate, const Http2Frame &frame)
   }
 
   // Check whether Window Size is acceptable
-  if (cstate.server_rwnd() < payload_length) {
+  if (!cstate.server_rwnd_is_shrinking && cstate.server_rwnd() < payload_length) {
     return Http2Error(Http2ErrorClass::HTTP2_ERROR_CLASS_CONNECTION, Http2ErrorCode::HTTP2_ERROR_FLOW_CONTROL_ERROR,
                       "recv data cstate.server_rwnd < payload_length");
   }
@@ -1054,8 +1054,18 @@ Http2ConnectionState::Http2ConnectionState() : stream_list()
 void
 Http2ConnectionState::init(Http2CommonSession *ssn)
 {
-  session            = ssn;
-  this->_server_rwnd = Http2::initial_window_size;
+  session = ssn;
+
+  if (Http2::initial_window_size < HTTP2_INITIAL_WINDOW_SIZE) {
+    // There is no HTTP/2 specified way to shrink the connection window size
+    // other than to receive data and not send WINDOW_UPDATE frames for a
+    // while.
+    this->_server_rwnd             = HTTP2_INITIAL_WINDOW_SIZE;
+    this->server_rwnd_is_shrinking = true;
+  } else {
+    this->_server_rwnd             = Http2::initial_window_size;
+    this->server_rwnd_is_shrinking = false;
+  }
 
   local_hpack_handle  = new HpackHandle(HTTP2_HEADER_TABLE_SIZE);
   remote_hpack_handle = new HpackHandle(HTTP2_HEADER_TABLE_SIZE);
@@ -1439,6 +1449,7 @@ Http2ConnectionState::restart_receiving(Http2Stream *stream)
   if (this->server_rwnd() < min_rwnd) {
     Http2WindowSize diff_size = initial_rwnd - this->server_rwnd();
     this->increment_server_rwnd(diff_size);
+    this->server_rwnd_is_shrinking = false;
     this->send_window_update_frame(0, diff_size);
   }
 
--- a/proxy/http2/Http2ConnectionState.h
+++ b/proxy/http2/Http2ConnectionState.h
@@ -86,6 +86,18 @@ public:
   DependencyTree *dependency_tree  = nullptr;
   ActivityCop<Http2Stream> _cop;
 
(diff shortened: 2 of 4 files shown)


bneradt (approved):

zwoop:
[approve ci autest]

zwoop:
[approve ci autest]

ezelkow1:
[approve ci autest]

bneradt:
Looks like the s3_auth test needs to be updated in 9.2.x:

```
1 Tests were skipped:
--------------------------------------------------------------------------------
 Test "conn_timeout" Skipped
    File: conn_timeout.test.py
    Directory: /home/jenkins/workspace/Github_Builds/autest/src/tests/gold_tests/timeout
  Reason: Test requires privilege

 Test: s3_auth_config: Exception
    File: s3_auth_config.test.py
    Directory: /home/jenkins/workspace/Github_Builds/autest/src/tests/gold_tests/pluginTest/s3_auth
     Reason: Traceback (most recent call last):
       File "/home/jenkins/.local/share/virtualenvs/tests-iiJvYDvR/lib/python3.8/site-packages/autest/core/runtesttask.py", line 34, in __call__
         tl = self.__logic.Run(self.__test)
       File "/home/jenkins/.local/share/virtualenvs/tests-iiJvYDvR/lib/python3.8/site-packages/autest/runlogic/runlogic.py", line 18, in Run
         if not tmp.Start(obj):
       File "/home/jenkins/.local/share/virtualenvs/tests-iiJvYDvR/lib/python3.8/site-packages/autest/runlogic/test.py", line 90, in Start
         loadTest(self.__test)
       File "/home/jenkins/.local/share/virtualenvs/tests-iiJvYDvR/lib/python3.8/site-packages/autest/core/test.py", line 292, in loadTest
         execFile(fileName, locals, locals)
       File "/home/jenkins/.local/share/virtualenvs/tests-iiJvYDvR/lib/python3.8/site-packages/autest/common/execfile.py", line 16, in execFile
         exec(safeCompile(f.read(), fname), globals, locals)
       File "/home/jenkins/workspace/Github_Builds/autest/src/tests/gold_tests/pluginTest/s3_auth/s3_auth_config.test.py", line 65, in <module>
         ts.Disk.traffic_out.Content = "gold/s3_auth_parsing_ts.gold"
     AttributeError: 'Disk' object has no attribute 'traffic_out'
```

I can create a 9.2.x PR to fix that. Then hopefully the tests should work better.

zwoop (approved):

PR #9127, opened 2022-10-06
In NetHandler, don't use a union to do a ratchet conversion to base class.

ywkaras:


ywkaras:
[approve ci autest]

cmcfarlen (approved):
ratchet indeed.

iocore/net/P_UnixNet.h
```diff
@@ -90,7 +90,7 @@ struct EventIO {
   bool syscall         = true;    ///< if false, disable all functionality (for QUIC)
   int type             = 0;       ///< class identifier of union data.
   union {
-    Continuation *c;
```
ywkaras, line 93: This was only used to do the ratchet conversion from pointer to NetAccept to Continuation.

PR #9130, opened 2022-10-07
Enable slice prefetching for range requests

serrislew:
Currently slice's prefetching feature relies only on the reference block (block 0 of the full object) to determine whether the successive blocks should be prefetched. This new update uses the first block of the client request as opposed to the first block of the full object (bytes=0-)
- Doesn't change functionality for full get request
- Doesn't change how reference block is used, only adds prefetching to be enabled when conditions are met on the first block of the client request (cacheable, hit-stale/miss status)


traeak (approved):
They look good and work under testing.

zwoop:
Cherry-picked to v9.2.x

PR #9133, opened 2022-10-11
fix contradicting documentation and say a bit about the resident size…

### LINKED ISSUE #9111: Confused about memory allocation
I want to know about memory allocation.
I think there are 2 parts of memory allocation, both are disk storage relevant.
1. RAM Cache . 
2.  Cache directory is used as a memory resident structure

# RAM Cache
There are 2 statements, but contradictory.
1. 10MB to 1GB. https://docs.trafficserver.apache.org/en/latest/admin-guide/performance/index.en.html?highlight=ssd#memory-allocation
2. 1MB to 1GB.  https://docs.trafficserver.apache.org/en/latest/admin-guide/storage/index.en.html#changing-the-size-of-the-ram-cache

Q: what is real?

# Cache directory is used as a memory resident structure
In my opinion, Disk cache directory has indexed in memory. As HDD space grows, so does memory usage.
Q: So, for example, how many memory the indexes are used when for 1TB hard disk?

cmcfarlen:
… of a volume directory

fixes #9111

### DIFF
--- a/doc/admin-guide/storage/index.en.rst
+++ b/doc/admin-guide/storage/index.en.rst
@@ -130,7 +130,7 @@ To change the RAM cache size:
 
 #. Set the variable :ts:cv:`proxy.config.cache.ram_cache.size`
    to specify the size of the RAM cache. The default value of ``-1`` means
-   that the RAM cache is automatically sized at approximately 1MB per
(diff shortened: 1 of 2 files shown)


ezelkow1 (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9134, opened 2022-10-11
Adding regex capability based on pristine host name

ezelkow1:
Adds a new `regex_host` field to the maxmind plugin and also renames `regex` to `regex_path`. This allows doing regex based allow/deny based on the pristine hostname that the request came in on which can be useful for a variety of scenarios

### DIFF
--- a/plugins/experimental/maxmind_acl/mmdb.cc
+++ b/plugins/experimental/maxmind_acl/mmdb.cc
@@ -551,7 +551,7 @@ Acl::eval(TSRemapRequestInfo *rri, TSHttpTxn txnp)
 #endif
 
       MMDB_entry_data_s entry_data;
-      int path_len, host_len = 0;
+      int path_len = 0, host_len = 0;
       const char *path = nullptr;
       const char *host = nullptr;
       if (!allow_regex_path.empty() || !deny_regex_path.empty()) {
--- a/plugins/experimental/maxmind_acl/mmdb.cc
+++ b/plugins/experimental/maxmind_acl/mmdb.cc
@@ -81,7 +81,6 @@ Acl::init(char const *filename)
 
   // Find our database name and convert to full path as needed
   status = loaddb(maxmind["database"]);
-
   if (!status) {
     TSDebug(PLUGIN_NAME, "Failed to load MaxMind Database");
     return status;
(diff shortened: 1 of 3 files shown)


plugins/experimental/maxmind_acl/mmdb.cc
```diff
@@ -694,19 +740,39 @@ Acl::eval_country(MMDB_entry_data_s *entry_data, const char *path, int path_len)
...
       }
     }
-    if (!deny_regex[output].empty()) {
-      for (auto &i : deny_regex[output]) {
+    if (!deny_regex_path[output].empty()) {
+      for (auto &i : deny_regex_path[output]) {
         if (PCRE_ERROR_NOMATCH != pcre_exec(i._rex, i._extra, path, path_len, 0, PCRE_NOTEMPTY, nullptr, 0)) {
-          TSDebug(PLUGIN_NAME, "Got a regex deny hit on regex: %s, country: %s", i._regex_s.c_str(), output);
+          TSDebug(PLUGIN_NAME, "Got a regex deny hit on regex path: %s, country: %s", i._regex_s.c_str(), output);
+          ret = false;
```
cmcfarlen, line 757: If this is denied due to regex_path, should this skip the allow_regex_host check below?  There needs to be some precedence here since there are now two ways to both allow and deny.  Whatever is decided should be documented as well.

cmcfarlen (changes requested):
As is, the user could deny by path, but allow by host and it might be astonishing that it was allowed.  I suggest at least documenting this, but perhaps skip the ability for the regex_host to flip back to allow if the regex_path had a deny hit.

plugins/experimental/maxmind_acl/mmdb.cc
ezelkow1: true, thanks, yea the host section for the deny should probably check if it is already set to false and if it is leave it as false. Though the plugin can run in sort of a 'reverse' mode, where if you only set allow fields then by default it will deny everything else that isnt explicitly allowed.

So let me think about this, see if there is a way to logic a way out based on if its running in deny-everything mode plus there has been an allow or the default mode and you've already gotten a denial based on path (this was always the most brain hurting part of this plugin)

ezelkow1:
Closing, changing the regex to just be the full URL which should hopefully be a much smaller change

PR #9136, opened 2022-10-11
Change the path for keylog_file on http2_flow_control autest

maskit:
Our test environment doesn't allow ATS to make files under `Test.RunDirectory`. This PR changes the path to `/tmp` for now, although I think Autest should provide `Test.TempDirectory`.

> $ git grep tls_session_keys.txt
tests/gold_tests/h2/http2_flow_control.test.py:            'proxy.config.ssl.keylog_file': os.path.join(Test.RunDirectory, 'tls_session_keys.txt'),
tests/gold_tests/pluginTest/multiplexer/multiplexer.test.py:            'proxy.config.ssl.keylog_file': '/tmp/tls_session_keys.txt',

### DIFF
--- a/tests/gold_tests/h2/http2_flow_control.test.py
+++ b/tests/gold_tests/h2/http2_flow_control.test.py
@@ -107,7 +107,7 @@ class Http2FlowControlTest:
             'proxy.config.ssl.server.private_key.path': f'{ts.Variables.SSLDir}',
             'proxy.config.ssl.client.verify.server.policy': 'PERMISSIVE',
             'proxy.config.dns.nameservers': '127.0.0.1:{0}'.format(self._dns.Variables.Port),
-            'proxy.config.ssl.keylog_file': '/tmp/tls_session_keys.txt',
+            'proxy.config.ssl.keylog_file': '{0}/tls_session_keys.txt'.format(ts.Variables.LOGDIR),
 
             'proxy.config.diags.debug.enabled': 3,
             'proxy.config.diags.debug.tags': 'http',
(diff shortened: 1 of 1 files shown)


tests/gold_tests/h2/http2_flow_control.test.py
```diff
@@ -107,7 +107,7 @@ def _configure_trafficserver(self):
             'proxy.config.ssl.server.private_key.path': f'{ts.Variables.SSLDir}',
             'proxy.config.ssl.client.verify.server.policy': 'PERMISSIVE',
             'proxy.config.dns.nameservers': '127.0.0.1:{0}'.format(self._dns.Variables.Port),
-            'proxy.config.ssl.keylog_file': os.path.join(Test.RunDirectory, 'tls_session_keys.txt'),
+            'proxy.config.ssl.keylog_file': '/tmp/tls_session_keys.txt',
```
bneradt, line 110: A downside to this is that it won't be in the sandbox then. CI, for instance, won't have this in the artifacts directory on failure.  Can we instead put this in Traffic Server's log directory? (Like where diags.log is.)

bneradt (changes requested):

maskit:
[approve ci autest]

tests/gold_tests/h2/http2_flow_control.test.py
bneradt: After thinking about the problem you faced some, I think it could be helpful to just have the extension create the keylog file for all TLS traffic in a consistent way. This will avoid the situation where test writers choose random locations for the keylog file (like I've clearly done so far) and sometimes choosing a location which doesn't work in some CI environments. What do you think of this?

https://github.com/apache/trafficserver/pull/9137

maskit: Sounds good to me. The key logging is not a heavy task and the data size is small. I think we can enable it on all tests that use TLS/QUIC.

maskit:
The issue was resolved by #9137.

PR #9138, opened 2022-10-12
MaxMind change regex to operate on the full pristine URL and not just the path

ezelkow1:
This changes the regex field in the MaxMind plugin to operate on the full original pristine URL (without query strings). This way users can construct their own regexes that can operate on the hostname or the path or both

### DIFF
--- a/plugins/experimental/maxmind_acl/mmdb.cc
+++ b/plugins/experimental/maxmind_acl/mmdb.cc
@@ -517,10 +517,22 @@ Acl::eval(TSRemapRequestInfo *rri, TSHttpTxn txnp)
 #endif
 
       MMDB_entry_data_s entry_data;
-      int path_len     = 0;
-      const char *path = nullptr;
+      std::string url;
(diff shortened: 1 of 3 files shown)


ezelkow1:
[approve ci autest]

plugins/experimental/maxmind_acl/mmdb.cc
```diff
@@ -670,7 +682,7 @@ Acl::eval_anonymous(MMDB_entry_s *entry)
 // allowable country code from our map.
 // False otherwise
 bool
-Acl::eval_country(MMDB_entry_data_s *entry_data, const char *path, int path_len)
+Acl::eval_country(MMDB_entry_data_s *entry_data, std::string url)
```
cmcfarlen, line 685: `const std::string&` here?

ezelkow1:
[approve ci autest]

cmcfarlen (approved):

PR #9139, opened 2022-10-12
Traffic Dump: fix YAML format for CONNECT requests

bneradt:
For CONNECT requests, Traffic Dump will not have a server-side protocol stack because we populate that on response from the server. Since the connection is left open for tunneling, it never gets populated. This resulted in mal-formed YAML because we had an empty string followed by a ','.


bneradt:
[approve ci autest]

randall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #8533, opened 2021-11-19
Partial of revert "Cleanup generated LDFLAGS for jemalloc (#8285)"

randall:
This reverts commit 9b3cde501d9f72d250a34f6d4742c816e1e7797b

Restoring --add-needed, but keeping the corrected path

Closes #8529

Issue #8529, opened 2021-11-18
Link error under RHEL7 with ATS 9.1.1 and jemalloc.


pbchou:
@randall -- We had an issue with the linker when building ATS 9.1.1 under RHEL7. I had no problems with my development box, which is Ubuntu 20.04 LTS, but our build environment is still RHEL7 with the newer compiler installed. The problem was fixed when I reverted #8285 in our internal branch. We had previously built 9.0.2 and 9.1.0 with no issues as well. Any ideas?

```
/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/ld: /root/trafficserver-9.1.1/src/tscore/.libs/libtscore.so: undefined reference to symbol '_ZdlPv'
/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/ld: /opt/jemalloc521/lib/libjemalloc.so: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

randall:
Can you see if https://github.com/apache/trafficserver/pull/8492 fixes things for you?

pbchou:
No, I cannot due to merge conflicts. This is because #8492 depends upon #8048, which is only in the 9.2.x branch or later. The possibility of #8492 was mentioned to my colleague on the slack channel earlier, but it seems to be a different issue since that issue was introduced in 9.2.x according to that issue's reporter in #8490.

randall:
Sorry about that. I'll rollback the change.

randall:
@pbchou When you get a chance, can you try #8533 ?

pbchou:
@randall -- We tested your PR #8533 against 9.1.x in our RHEL7 build environment. It was able to be built successfully.

@bryancall  -- By the way, do we still have RHEL7 in our CI environment? I understand it is getting old.

PR #9141, opened 2022-10-13
Remove deprecated ld option (--add-needed)

masaori335:
While I'm trying mimalloc (and jemalloc) with llvm toolchain (`lld`), I got the below error.

```
ld.lld: error: unknown argument '--add-needed'
```

It looks like `lld` doesn't support this option. Also `ld` deprecated this option.

> --add-needed
--no-add-needed
These two options have been deprecated because of the similarity of their names to the --as-needed and --no-as-needed options. They have been replaced by --copy-dt-needed-entries and --no-copy-dt-needed-entries.

https://sourceware.org/binutils/docs/ld/Options.html

### DIFF
--- a/build/jemalloc.m4
+++ b/build/jemalloc.m4
@@ -58,7 +58,7 @@ if test "$has_jemalloc" != "no"; then
   if test "$jemalloc_base_dir" != "/usr"; then
     TS_ADDTO(CPPFLAGS, [-I${jemalloc_include}])
     TS_ADDTO(LDFLAGS, [-L${jemalloc_ldflags}])
-    TS_ADDTO(LDFLAGS, [-Wl,--add-needed -L${jemalloc_ldflags} -Wl,-rpath,${jemalloc_ldflags} -Wl,--no-as-needed])
+    TS_ADDTO(LDFLAGS, [-Wl,--as-needed -L${jemalloc_ldflags} -Wl,-rpath,${jemalloc_ldflags} -Wl,--no-as-needed])
     TS_ADDTO_RPATH(${jemalloc_ldflags})
   fi
   # On Darwin, jemalloc symbols are prefixed with je_. Search for that first, then fall back
--- a/build/mimalloc.m4
+++ b/build/mimalloc.m4
@@ -62,7 +62,7 @@ if test "$has_mimalloc" != "no"; then
   if test "$mimalloc_base_dir" != "/usr"; then
     TS_ADDTO(CPPFLAGS, [-I${mimalloc_include}])
     TS_ADDTO(LDFLAGS, [-L${mimalloc_ldflags}])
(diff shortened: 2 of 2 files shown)


build/jemalloc.m4
```diff
@@ -58,7 +58,7 @@ if test "$has_jemalloc" != "no"; then
   if test "$jemalloc_base_dir" != "/usr"; then
     TS_ADDTO(CPPFLAGS, [-I${jemalloc_include}])
     TS_ADDTO(LDFLAGS, [-L${jemalloc_ldflags}])
-    TS_ADDTO(LDFLAGS, [-Wl,--add-needed -L${jemalloc_ldflags} -Wl,-rpath,${jemalloc_ldflags} -Wl,--no-as-needed])
```
masaori335, line 61: @duke8253 is this typo of `--as-needed` option? Because this line has `--no-as-needed` option at the end and it restores the default behavior.

If we really need `--add-needed` option here, `--copy-dt-needed-entries` is the replacement, but `lld` doesn't have it either.

masaori335:
[approve ci fedora]

randall:
I'm +1 on this(and had tried to do it myself), but reverted it in #8533

masaori335:
I checked with the devtoolset-9 on centos7 (similar env of #8529), it looks like this change works with `ld`.

```
configure:30542: cc -o conftest  -D_GNU_SOURCE -DOPENSSL_NO_SSL_INTERN -I/opt/jemalloc/include   -L/opt/jemalloc/lib -Wl,--as-needed -Wl,-rpath,/opt/jemalloc/lib -Wl,--no-as-needed conftest.c -ljemalloc  -lpthread -ldl  >&5
configure:30542: $? = 0
configure:30559: result: -ljemalloc
configure:30573: checking jemalloc/jemalloc.h usability
configure:30573: cc -c  -D_GNU_SOURCE -DOPENSSL_NO_SSL_INTERN -I/opt/jemalloc/include conftest.c >&5
configure:30573: $? = 0
configure:30573: result: yes
...
LD='/opt/rh/devtoolset-9/root/usr/libexec/gcc/x86_64-redhat-linux/9/ld -m elf_x86_64'
```

build/jemalloc.m4
duke8253: I don't think it's a typo because I remember coming back to this and thought it was a typo, but changing it broke stuff. Though I can't remember exactly what was it now, vaguely recall that something about the order of specifying those options made a difference.

duke8253:
Yeah, at the time there were some weird build errors on our centos test boxes. I'll find some time tomorrow to build it on there with the changes to see if things are working.

duke8253 (approved):
Works fine in my testings.

zwoop:
Cherry-picked to v9.2.x

PR #9147, opened 2022-10-15
Fix transparent tr-pass

### LINKED ISSUE #9146: tr-pass does not work with transparent mode
It may not work in regular mode either, but I was testing with 

CONFIG proxy.config.http.server_ports STRING 8443:ssl:tr-full:tr-pass

I was running 9.1.3, but I think this has been broken for a while and is likely a problem in master and 10-dev. 

When sending http traffic to the ssl port, the client would get errors about the host not being set.

When sending non-http traffic over tls, the client would get error that the scheme is not set or invalid.

I have a fix for setup_blind_tunnel_port that fixes both of these issues.

shinrich:
I plan on writing an autest for this, which will require privilege to run to exercise the transparent case.

This fixes #9146

### DIFF
--- a/proxy/http/HttpSM.cc
+++ b/proxy/http/HttpSM.cc
@@ -651,17 +651,22 @@ HttpSM::setup_blind_tunnel_port()
   NetVConnection *netvc = ua_txn->get_netvc();
   ink_release_assert(netvc);
   int host_len;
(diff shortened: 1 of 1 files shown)


shinrich:
Oops didn't push to fork.

PR #9153, opened 2022-10-18
Cleanup: Remove dead code (CacheLink)

masaori335:
It looks like below functions are dead code.
```
CacheVC::linkWrite(int event, Event *e);
CacheVC::derefRead(int event, Event *e);
Cache::link(Continuation *cont, const CacheKey *from, const CacheKey *to, CacheFragType type, const char *hostname, int host_len);
Cache::deref(Continuation *cont, const CacheKey *key, CacheFragType type, const char *hostname, int host_len);
```


brbzull0 (approved):
Looks good.

PR #9154, opened 2022-10-20
Add TsScalableSharedMutex.h: shared mutex minimizing thread memory cache contention.

ywkaras:
Ran the performance test on an AMD EPYC 7742 64-Core Processor (128 logical cores).  Performance was about 30 times faster with 256 threads (shared locking the mutex).  However, performance was about 5 times slower with 16 threads.

If you compile this perf test with gcc version 8.3.1 , it will seg fault.  It runs with gcc 9.2.1 and clang 10.0.0 .

### DIFF
--- a/include/tscpp/util/Makefile.am
+++ b/include/tscpp/util/Makefile.am
@@ -25,4 +25,5 @@ library_include_HEADERS = \
 	Strerror.h \
 	string_view_util.h \
 	TextView.h \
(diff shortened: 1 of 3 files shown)


ywkaras:
I have to put this way on the back burner, as Yahoo is not prioritizing performance on hosts with a high core count.  If anyone else wants to work on it, feel free.

PR #9155, opened 2022-10-20
More HTTP/2 local and server renaming

bneradt:
This continues the work of #9084 with more renaming of HTTP/2 connection and stream variables and functions like so:

* client -> peer
* server -> local

### DIFF
--- a/proxy/http2/Http2ConnectionState.cc
+++ b/proxy/http2/Http2ConnectionState.cc
@@ -158,24 +158,24 @@ Http2ConnectionState::rcv_data_frame(const Http2Frame &frame)
   }
 
   // Check whether Window Size is acceptable
-  if (!this->_server_rwnd_is_shrinking && this->server_rwnd() < payload_length) {
+  if (!this->_local_rwnd_is_shrinking_in && this->get_local_rwnd_in() < payload_length) {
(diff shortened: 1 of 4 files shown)


maskit (approved):
I feel like local / remote is a more natural combination, but I'm not a best person to choose English words. I believe your choice.

I approve this PR for now, but we have `remote_hpack_handle` in Http2ConnectionState.  Please rename it as well if you think renaming it makes sense.

bneradt:
Thanks for the review.

> I feel like local / remote is a more natural combination, but I'm not a best person to choose English words. I believe your choice.

That's true. Often local and remote do go together. But I think in this case peer is better than remote for what these variables reference. The HTTP/2 RFC generally uses the term "peer" when referring to the other host:
https://www.rfc-editor.org/rfc/rfc9113.html

The term "remote" is used in the RFC, but generally to qualify peer (i.e., it says "remote peer" sometimes), or in reference in a few places to the name of a state. But in general, when discussing the other host, it calls it a "peer". 

Also, #9084 already started the trend of renaming the client to peer rather than remote. So if we did choose remote instead of peer, we would have to change those places too.
 
> I approve this PR for now, but we have `remote_hpack_handle` in Http2ConnectionState. Please rename it as well if you think renaming it makes sense.

Good catch. Let's rename that too.

maskit:
Okay, let's stick with the terms that are used in RFC.
The new commit didn't stale my review somehow. Please go ahead and merge this PR.

PR #9174, opened 2022-11-02
[autest] Make s3_auth_config less strict

moonchen:
Don't check header values that are not relevant to the test.


moonchen:
@bneradt beat me to it!

PR #9175, opened 2022-11-02
s3_auth_parsing.gold: Make Age Less Specified

bneradt:
The s3_auth_parsing.gold file overspecified the value of Age. We aren't testing that value in the test and due to timing issues the value can be 1 or 0. Making it allow other values.


moonchen (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9177, opened 2022-11-04
Fix the Loading of Client Certificate Chains

moonchen:
Additional certificates in a chain should be loaded with the `SSL_CTX_add_extra_chain_cert` API.

### DIFF
--- a/iocore/net/SSLConfig.cc
+++ b/iocore/net/SSLConfig.cc
@@ -834,7 +834,7 @@ SSLConfigParams::getCTX(const std::string &client_cert, const std::string &key_f
       // Continue to fetch certs to associate intermediate certificates
       cert = PEM_read_bio_X509(biop, nullptr, nullptr, nullptr);
       while (cert) {
-        if (!SSL_CTX_use_certificate(client_ctx.get(), cert)) {
+        if (!SSL_CTX_add_extra_chain_cert(client_ctx.get(), cert)) {
           SSLError("failed to attach client chain certificate from %s", client_cert.c_str());
           goto fail;
         }


moonchen:
[approve ci autest]

moonchen:
[approve ci autest]

moonchen:
[approve ci autest]

duke8253 (commented):
I think this is a good find. But the OpenSSL docs isn't very specific on whether the first cert needs to be added using `SSL_CTX_use_certificate`. Since it says (https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_use_certificate.html) "The rest of the certificates needed to form the complete certificate chain can be specified using the `SSL_CTX_add_extra_chain_cert` function". Could you confirm that all certs can be added to the chain using `SSL_CTX_add_extra_chain_cert`, including the first one?

moonchen:
> I think this is a good find. But the OpenSSL docs isn't very specific on whether the first cert needs to be added using `SSL_CTX_use_certificate`. Since it says (https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_use_certificate.html) "The rest of the certificates needed to form the complete certificate chain can be specified using the `SSL_CTX_add_extra_chain_cert` function". Could you confirm that all certs can be added to the chain using `SSL_CTX_add_extra_chain_cert`, including the first one?

The first cert is still added using `SSL_CTX_use_certificate` on line 828, so this only changes the loading of the second certificate and subsequent certs.

Before this change, loading a certificate chain and a private key will cause OpenSSL to give an error that the cert and key don't match.  I can confirm that loading the chain this way gets OpenSSL to pass the `SSL_CTX_check_private_key`.

zwoop:
@bryancall Approves this, but  he's lost his Github access because his mobile provider is terrible.

zwoop (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9181, opened 2022-11-05
Adding back set_connect_fail for generic I/O error

serrislew:
Without setting the generic I/O error, the server doesn't know it has failed and `t_state.current.server->had_connect_fail()` in `HttpSM::track_connect_fail()` always returns false. As a result, when the server has exhausted its attempts, it fails to reach `mark_host_failure` and does not update that HostDBInfo entry's last_failure timeout field

I am not sure why it was deleted before and do not see its replacement where the server's `connect_result` field is updated

### DIFF
--- a/tests/gold_tests/dns/dns_host_down.test.py
+++ b/tests/gold_tests/dns/dns_host_down.test.py
@@ -47,9 +47,10 @@ class DownCachedOriginServerTest:
             'proxy.config.hostdb.fail.timeout': 10,
             'proxy.config.dns.resolv_conf': 'NULL',
             'proxy.config.hostdb.ttl_mode': 1,
-            'proxy.config.hostdb.timeout': 10,
+            'proxy.config.hostdb.timeout': 2,
             'proxy.config.hostdb.lookup_timeout': 2,
             'proxy.config.http.transaction_no_activity_timeout_in': 2,
+            'proxy.config.http.connect_attempts_timeout': 2,
             'proxy.config.hostdb.host_file.interval': 1,
             'proxy.config.hostdb.host_file.path': os.path.join(Test.TestDirectory, "hosts_file"),
         })
--- a/tests/gold_tests/dns/dns_host_down.test.py
+++ b/tests/gold_tests/dns/dns_host_down.test.py
@@ -1,0 +1,93 @@
+'''
+Verify ATS handles down origin servers with domain cached correctly.
+'''
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+from ports import get_port
+import os
+
+Test.Summary = '''
+Verify ATS handles down origin servers with cached domain correctly.
+'''
+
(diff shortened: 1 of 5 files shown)


bneradt (commented):
Thanks for the PR. A few quick questions:

* Can we add an autest for this?
* You mention in the description that this was removed in a previous patch. Which PR took this out? I didn't find it when scanning patches to this file, so I'm probably not looking in the right place.
* Just checking: I notice this is marked for 9.2.x. Was the issue you're fixing introduced with the hostdb restructure in #8953? If so, that restructure only went into master/10-Dev.

serrislew:
Yes, I will definitely add an autest. The PR is #7580. I spoke to Susan about this and it was initially deleted since NET_EVENT_OPEN should have preemptively set the server connect failure code (EIO) but the failed case I've been testing doesn't reach this event. Instead of setting the EIO as I have it now (since it could possibly override other more specific failures), I will set it before NET_EVENT_OPEN so it applies to all cases.

serrislew:
[approve ci]

tests/gold_tests/dns/replay/server_down.replay.yaml
```diff
@@ -0,0 +1,54 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+meta:
+  version: "1.0"
+
+sessions:
+- transactions:
+  - client-request:
+      delay: 2s
```
bneradt, line 23: Let's add comments explaining the reasoning for the delays in the replay file. In particular I'm currently trying to think through the relationship between this first delay and the DelayStart directives in the test.py file. Shouldn't only one of those be needed?

tests/gold_tests/dns/dns_host_down.test.py
```diff
@@ -0,0 +1,96 @@
...
+            'proxy.config.hostdb.host_file.interval': 1,
+            'proxy.config.hostdb.host_file.path': os.path.join(Test.TestDirectory, "hosts_file"),
+        })
+
+    # Even when the origin server is down, SM will return a hit-fresh domain from HostDB.
+    # After request has failed, SM should mark the IP as down
+    def _test_host_mark_down(self):
+        tr = Test.AddTestRun()
+        self._server = tr.AddVerifierServerProcess("server", DownCachedOriginServerTest.replay_file)
+        self._configure_trafficserver()
```
bneradt, line 63: I suggest moving this to the `__init__` function since it's a Test-level rather than a TestRun-level process.

tests/gold_tests/dns/dns_host_down.test.py
```diff
@@ -0,0 +1,96 @@
...
+        tr.AddVerifierClientProcess(
+            "client-1",
+            DownCachedOriginServerTest.replay_file,
+            http_ports=[self._ts.Variables.port],
+            other_args='--keys 1')
+
+        tr.NotRunningAfter = self._server
+
+    # After host has been marked down from previous test, HostDB should not return
+    # the host as available and DNS lookup should fail.
```
bneradt, line 78: Any diags.log entries we should verify with the test via ContainsExpression? (traffic.out is debug-level, and it's generally good to avoid depending upon those since they may change more regularly.)

tests/gold_tests/dns/replay/server_down.replay.yaml
```diff
@@ -0,0 +1,54 @@
...
+        fields:
+        - [ Host, example.com ]
+        - [ X-Request, request ]
+        - [ uuid, 1 ]
+
+    server-response:
+      status: 200
+
+    proxy-response:
+      status: 502
```
bneradt, line 37: Comments explaining the reasoning of these expected return codes would be helpful too.

bneradt (changes requested):
Looks good. Thanks for adding the test. I just have a few minor comments.

tests/gold_tests/dns/replay/server_down.replay.yaml
serrislew: The client request delay is to allow the host file to sync so HostDB will include entries and the DelayStart is specific to the test run. I don't think the test run actually needs a delay so I will remove that one and also add comments to the replay file!

bneradt (approved):
Looks great.

zwoop:
Cherry-picked to v9.2.x

PR #9168, opened 2022-11-01
TSan: Fix data race of updating current time

masaori335:
Fix below TSan report.
```
WARNING: ThreadSanitizer: data race (pid=74346)
  Write of size 8 at 0x0000041330f8 by main thread (mutexes: write M0, write M1, write M2, write M3, write M4, write M5, write M6, write M7, write M8, write M9, write M10, write M11, write M12, write M13):
    #0 Thread::get_hrtime_updated() I_Thread.h:194 (traffic_server:x86_64+0x80513a) (BuildId: c245d2fe2dd53a87af825d77f24750ba32000000200000000100000000000c00)
    #1 PriorityEventQueue::PriorityEventQueue() PQ-List.cc:28 (traffic_server:x86_64+0x849536) (BuildId: c245d2fe2dd53a87af825d77f24750ba32000000200000000100000000000c00)
    #2 PriorityEventQueue::PriorityEventQueue() PQ-List.cc:27 (traffic_server:x86_64+0x8495e5) (BuildId: c245d2fe2dd53a87af825d77f24750ba32000000200000000100000000000c00)
    #3 EThread::EThread(ThreadType, Event*) UnixEThread.cc:107 (traffic_server:x86_64+0x84d7cb) (BuildId: c245d2fe2dd53a87af825d77f24750ba32000000200000000100000000000c00)
    #4 EThread::EThread(ThreadType, Event*) UnixEThread.cc:108 (traffic_server:x86_64+0x84da71) (BuildId: c245d2fe2dd53a87af825d77f24750ba32000000200000000100000000000c00)
    #5 EventProcessor::spawn_thread(Continuation*, char const*, unsigned long) UnixEventProcessor.cc:534 (traffic_server:x86_64+0x856566) (BuildId: c245d2fe2dd53a87af825d77f24750ba32000000200000000100000000000c00)
    #6 NetAccept::init_accept_loop() UnixNetAccept.cc:168 (traffic_server:x86_64+0x7dc6bc) (BuildId: c245d2fe2dd53a87af825d77f24750ba32000000200000000100000000000c00)
    #7 UnixNetProcessor::accept_internal(Continuation*, int, NetProcessor::AcceptOptions const&) UnixNetProcessor.cc:146 (traffic_server:x86_64+0x7e61e6) (BuildId: c245d2fe2dd53a87af825d77f24750ba32000000200000000100000000000c00)
[...]

PR #9184, opened 2022-11-08
TSan: Make Thread::cur_time thread local

masaori335:
The alternative (original) approach of #9168 to fix the data race.

### DIFF
--- a/iocore/eventsystem/I_Thread.h
+++ b/iocore/eventsystem/I_Thread.h
@@ -156,8 +156,8 @@ public:
       This gets a cached copy of the time so it is very fast and reasonably accurate.
       The cached time is updated every time the actual operating system time is fetched which is
       at least every 10ms and generally more frequently.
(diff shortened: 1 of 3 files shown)


masaori335:
[approve ci fedora]

SolidWallOfCode (approved):

zwoop:
We'd like to see this in the 9.2.x tree, but my recommendation is that we wait for v9.2.2, since we need to make a safe, stable v9.2.1 soon.

masaori335:
We haven't seen any issue with this change so far, but I'm +1 for v9.2.2 for the stability of v9.2.1 release.

A note is this change made 18% improvement on the high rps benchmark. I'm surprised by that, but it makes sense because the `Thread::cur_time` shows up on perf top without the change.

zwoop:
Cherry-picked to v9.2.x

PR #9186, opened 2022-11-08
Traffic Dump: Allow unlimited disk utilization

bneradt:
The traffic_dump.so plugin has a --limit argument that specifies how much disk space the plugin will use before stopping writing out new sessions to replay files. This patch allows the user to configure unrestricted disk utilization by making --limit optional and making the absence of `--limit` imply no disk utilization.

### DIFF
--- a/doc/admin-guide/plugins/traffic_dump.en.rst
+++ b/doc/admin-guide/plugins/traffic_dump.en.rst
@@ -46,7 +46,7 @@ Plugin Configuration
 
    .. option:: --limit <N>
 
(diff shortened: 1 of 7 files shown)


moonchen:
A couple of suggestions:
1. We could make it such that `--limit` is optional, and not specifying `--limit` means unlimited.
2. I think the implementation would be simpler if the "unlimited" value was INT64_MAX.

bneradt:
> A couple of suggestions:
> 
> 1. We could make it such that `--limit` is optional, and not specifying `--limit` means unlimited.
> 2. I think the implementation would be simpler if the "unlimited" value was INT64_MAX.

Yes, good points. I'll just make the `--limit` optional. Thanks @moonchen .

moonchen (approved):
Looks good.

PR #9189, opened 2022-11-10
Removed most sprintf and ignore sprintf warnings in bind code

bryancall:
When compiling on the mac I am getting a bunch of compiler errors for sprintf.  I converted all of them to snprintf, except for the bind code that we copied over.  For that I am ignoring the errors.

@bneradt I changed the way I am filling the `resize()`  in `TransactionData::initialize_default_sensitive_field`.  Hopefully this won't cause any issues. My worry is that we are writing 9 bytes including the null into data.  From my understanding it is not guaranteed to have space for that.

### DIFF
--- a/plugins/experimental/ja3_fingerprint/ja3_fingerprint.cc
+++ b/plugins/experimental/ja3_fingerprint/ja3_fingerprint.cc
@@ -319,7 +319,7 @@ client_hello_ja3_handler(TSCont contp, TSEvent event, void *edata)
     MD5((unsigned char *)data->ja3_string.c_str(), data->ja3_string.length(), digest);
 
     for (int i = 0; i < 16; i++) {
-      sprintf(&(data->md5_string[i * 2]), "%02x", static_cast<unsigned int>(digest[i]));
+      snprintf(&(data->md5_string[i * 2]), 3, "%02x", static_cast<unsigned int>(digest[i]));
     }
     TSDebug(PLUGIN_NAME, "Fingerprint: %s", data->md5_string);
     break;
--- a/plugins/experimental/ja3_fingerprint/ja3_fingerprint.cc
+++ b/plugins/experimental/ja3_fingerprint/ja3_fingerprint.cc
@@ -318,8 +318,10 @@ client_hello_ja3_handler(TSCont contp, TSEvent event, void *edata)
     unsigned char digest[MD5_DIGEST_LENGTH];
     MD5((unsigned char *)data->ja3_string.c_str(), data->ja3_string.length(), digest);
 
+    // Validate that the buffer is the same size as we will be writing into (compile time)
+    static_assert((16 * 2 + 1) == sizeof(data->md5_string));
     for (int i = 0; i < 16; i++) {
-      snprintf(&(data->md5_string[i * 2]), 3, "%02x", static_cast<unsigned int>(digest[i]));
+      snprintf(&(data->md5_string[i * 2]), sizeof(data->md5_string) - (i * 2), "%02x", static_cast<unsigned int>(digest[i]));
     }
     TSDebug(PLUGIN_NAME, "Fingerprint: %s", data->md5_string);
     break;
--- a/plugins/prefetch/plugin.cc
+++ b/plugins/prefetch/plugin.cc
@@ -584,9 +584,9 @@ contHandleFetch(const TSCont contp, TSEvent event, void *edata)
         TSHttpHdrReasonSet(bufp, hdrLoc, reason, reasonLen);
         PrefetchDebug("set response: %d %.*s '%s'", data->_status, reasonLen, reason, data->_body.c_str());
 
-        int length = data->_body.length() + 1;
-        char *buf  = static_cast<char *>(TSmalloc(length));
-        snprintf(buf, length, "%s", data->_body.c_str());
+        size_t bufsize = data->_body.length() + 1;
+        char *buf      = static_cast<char *>(TSmalloc(bufsize));
+        snprintf(buf, bufsize, "%s", data->_body.c_str());
         TSHttpTxnErrorBodySet(txnp, buf, strlen(buf), nullptr);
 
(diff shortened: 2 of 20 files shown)


bryancall:
[approve ci clang-format]

bneradt (changes requested):
Thanks for updating these.

bneradt:
> @bneradt I changed the way I am filling the `resize()` in `TransactionData::initialize_default_sensitive_field`. Hopefully this won't cause any issues. My worry is that we are writing 9 bytes including the null into data. From my understanding it is not guaranteed to have space for that.

In case it's helpful or interesting, the original was safe. Since C++11, std::string is guaranteed to have a null terminator. To demonstrate this, let's think of the default string being a size of 8 characters instead of 128k to make it easier to think through. Then the function becomes:

```cpp
  // 128 KB is the maximum size supported for all headers, so this size should
  // be plenty large for our needs.
  //constexpr size_t default_field_size = 128 * 1024;
  constexpr size_t default_field_size = 8;  // <--- Changed to 8.
  default_sensitive_field_value.resize(default_field_size);

  char *field_buffer = default_sensitive_field_value.data();
  for (auto i = 0u; i < default_field_size; i += 8) {
    sprintf(field_buffer, "%07x ", i / 8);
    field_buffer += 8;
  }

```

Thus the internal buffer size has to be at least 9 (8 for the resize value for the number of characters, and then one more value for the null terminator). The for loop will execute exactly once and write 8 characters (7 `0` values and a space "` `" character) and then sprintf will null terminate the string at `default_sensitive_field_value[8]` which is already guaranteed to be null. Thus there is no buffer overrun in the original and its behavior should be correct.

That said, I prefer the ostringstream version I suggest in my comment over the sprintf.

Again, thank you for the PR. I'm happier with the update to the function.

tools/jtest/jtest.cc
```diff
@@ -3083,24 +3088,25 @@ build_request(int sock)
...
       nheaders--;
     }
     if (nheaders > 0) {
-      eh += sprintf(eh, "Accept: image/gif, image/x-xbitmap, image/jpeg, image/pjpeg, image/png, */*\r\n");
+      eh += snprintf(eh, sizeof(eheaders) - (eh - eheaders),
+                     "Accept: image/gif, image/x-xbitmap, image/jpeg, image/pjpeg, image/png, */*\r\n");
     }
     while (--nheaders > 0) {
-      eh += sprintf(eh, "Extra-Header%d: a lot of junk for header %d\r\n", nheaders, nheaders);
+      eh += snprintf(eh, sizeof(eheaders) - (eh - eheaders), "Extra-Header%d: a lot of junk for header %d\r\n", nheaders, nheaders);
```
bneradt, line 3099: I don't think this accumulating `eh` index will work out if there is an overflow, right? `snprintf` will return the number of bytes it would have written regardless of the buffer size parameter. Thus, if it only writes 50 bytes because that's the buffer size, but it would have written 105, it will return 105. Then what happens to the subsequent snprintf buffer calculations when the `eh` value is larger than `sizeof(eheaders)`?

bneradt (changes requested):
I read through each of these and added some minor comments below. Thanks for working on a PR that protects us from overruns.

bryancall:
[approve ci]

bryancall:
[approve ci]

plugins/experimental/ja3_fingerprint/ja3_fingerprint.cc
```diff
@@ -318,8 +318,10 @@ client_hello_ja3_handler(TSCont contp, TSEvent event, void *edata)
     unsigned char digest[MD5_DIGEST_LENGTH];
     MD5((unsigned char *)data->ja3_string.c_str(), data->ja3_string.length(), digest);
 
+    // Validate that the buffer is the same size as we will be writing into (compile time)
+    static_assert((16 * 2 + 1) == sizeof(data->md5_string));
```
bneradt, line 322: Good idea

bneradt (commented):
Looks good to me.

plugins/prefetch/plugin.cc
```diff
@@ -584,8 +584,9 @@ contHandleFetch(const TSCont contp, TSEvent event, void *edata)
         TSHttpHdrReasonSet(bufp, hdrLoc, reason, reasonLen);
         PrefetchDebug("set response: %d %.*s '%s'", data->_status, reasonLen, reason, data->_body.c_str());
 
-        char *buf = static_cast<char *>(TSmalloc(data->_body.length() + 1));
-        sprintf(buf, "%s", data->_body.c_str());
+        size_t bufsize = data->_body.length() + 1;
+        char *buf      = static_cast<char *>(TSmalloc(bufsize));
```
moonchen, line 588: I think that `memcpy(buf, data->_body.c_str(), bufsize)` would be faster and simpler here.

proxy/http/remap/unit-tests/plugin_testing_common.cc
```diff
@@ -46,7 +46,7 @@ getTemporaryDir()
   tmpDir /= "sandbox_XXXXXX";
 
   char dirNameTemplate[tmpDir.string().length() + 1];
-  sprintf(dirNameTemplate, "%s", tmpDir.c_str());
+  snprintf(dirNameTemplate, sizeof(dirNameTemplate), "%s", tmpDir.c_str());
```
moonchen, line 49: Likewise, I think this should be `memcpy(dirNameTemplate, tmpDir.c_str(), sizeof(dirNameTemplate));`

moonchen (changes requested):

proxy/http/remap/unit-tests/plugin_testing_common.cc
moonchen: A lot of nice changes.  Thanks!

moonchen:
A lot of nice changes.  Thanks!

PR #9194, opened 2022-11-14
Update .gitignore for overridable_txn_vars.cc in mgmt/

bneradt:
mgmt/rpc/overridable_txn_vars.cc is a generated file that our .gitignore should ignore. This used to be in mgmt2, but after #8633 it is now in the mgmt directory. This updates the .gitignore file for the new location.


brbzull0 (approved):
Thanks!

PR #9195, opened 2022-11-14
AuTest: make MakeATSProcess accessible for TestRuns

bneradt:
In the past there were limitations with the Traffic Server AuTest process extension which would prevent it from being associated with TestRun objects. That has since been addressed. This change makes it so that the ATS process for our AuTests can be associated with TestRun objects in addition to the larger and currently available Test scope.


serrislew (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9201, opened 2022-11-16
update gdb-helpers

cmcfarlen:


bneradt:
Here's a link to the autopep8 output for `./tools/gdb-helpers.py`:

https://ci.trafficserver.apache.org/job/Github_Builds/job/clang-format/447/console

tools/gdb-helpers.py
```diff
@@ -222,10 +233,14 @@ def sm_command(val):
 def hdrs_command(val):
     HTTPHdr(val).pr()
 
+def url_command(val):
+    print("{}".format(URL(val)))
```
bneradt, line 237: Doesn't look like this needs to be a formatted string:

```
    print(URL(val))
```

tools/gdb-helpers.py
```diff
@@ -37,19 +37,30 @@
...
-    else:
-        return buff
+    try:
+        buff = inferior.read_memory(addr, addr_len)
+        if buff[0] == '\x00':
+            return 'null'
+        else:
+            return buff
+    except:
+        return 'unreadable({:x})'.format(int(addr))
```
bneradt, line 51: Preferable to use a format string literal:

```
        return f'unreadable({int(addr):x}'

bneradt (changes requested):
Looks good. Thanks for starting and updating this.

tools/gdb-helpers.py
bneradt: If you want to stick with `.format` to be compatible with more versions of Python, that's fine with me.

bneradt (approved):

PR #9212, opened 2022-11-21
traffic_ctl -  Fix number of mandatory arguments when sending a plugin msg.

brbzull0:
Fix number of mandatory arguments when sending a plugin msg.
It looks like this got badly merged when the traffic_manager was
removed.


SolidWallOfCode (approved):

brbzull0:
> Actshually, if you're fiddling with that, requiring two arguments is annoying. See [here](http://docs.solidwallofcode.com/txn_box/install.en.html#reloading) for an example, where to get TxnBox to reload its configuration, you have to provide a second argument that is useless. If there's no second argument, then an empty (zero length) message should be sent.

yes, it supposed to be at least 1 arg, it was enforing 2. This fixes that.

bneradt (approved):

PR #9214, opened 2022-11-21
up limits of lua string length and url length

traeak:
Hard code update the max string length and max url length for lua scripts.


shukitchan (changes requested):
32768 is better, i think

shukitchan (approved):

PR #9217, opened 2022-11-23
Clean RefCountObj

cukiernik:


ywkaras:
Can you also do the same thing to the assignment operator?

cmcfarlen:
Should the copy constructor and assignment be marked `delete` instead? The copy implementation doesn't semantically copy its argument so perhaps `delete` is better.

bryancall (approved):

PR #853, opened 2016-08-11
TS-4619: intermediate chain loading can miss certificates.

shinrich:
Made the changes @jpeach suggested in the bug.  Tested with three deep chains for rsa and ec (cert and two signers).  Tested with both signers in the ssl_ca_name files.  Tested with all three certs in the ssl_cert_name file.

PR #9230, opened 2022-12-03
Fail out when intermediate certificate chain fails to load

serrislew:
ATS should immediately fail out when intermediate certificate chains fail to load. Currently only a debug log is included when this fails. It originally failed out (by calling `goto fail`) but was missed during refactoring in #853 
- This follows similar behavior when `SSL_CTX_add_extra_chain_cert_file` is called, which fails out when returning an error


iocore/net/SSLUtils.cc
```diff
@@ -2445,6 +2445,8 @@ SSLMultiCertConfigLoader::load_certs(SSL_CTX *ctx, const std::vector<std::string
     // Load up any additional chain certificates
     if (!SSL_CTX_add_extra_chain_cert_bio(ctx, bio.get())) {
       Debug("ssl", "couldn't add chain to %p", ctx);
+      SSLError("failed to load intermediate certificate chain from %s", cert_names_list[i].c_str());
+      return false;
```
ywkaras, line 2449: Clearly this should be an Error.  But, if you look at the call stack this function is at the bottom of, it's not clear whether or not it better to return false here.  I assume none of these errors cause Fatal() calls because of the reload case.  TS code tends avoid exiting on a reload error.  Without necessarily considering it may be worse to keep running in a corrupt state.

serrislew: This error handling replicates the behavior of all other errors (i.e. `secret_data.empty()` and when cert fails to load certificate chain). Similar to those, it should exit out of that function immediately

ywkaras: Have you seen this happen in production?

serrislew: No, not on this specific error. This was the only error case that didn't add a SSLError and exit with returning false

ywkaras: Hopefully someone who knows this code better than I do will take a quick look.

shinrich (approved):
I agree, it should fail in this case.

I looked through the code a bit to see about the returning 0 or not.  Someone should probably do some experiments here.  By my reading, the main (only) caller of load_certs is init_server_ssl_ctx which returns an array of SSLLoadingContext.  If there is a failure in load_certs, the return list will be empty.  Since the function doesn't distinguish between a failure and no items, I think we lose the error at that point.  

Sound probably do some experiments and refactor that path to propagates the errors too in a separate PR.

serrislew:
@shinrich I agree, I think the error handling for all of those errors should be done in a separate effort/PR

zwoop:
Cherry-picked to v9.2.x

PR #9225, opened 2022-12-01
Verbose SSL error logs

serrislew:
Adding more verbose information to SSL error and debug logs
- Explicitly stating (i.e. "[empty file]") for empty files to avoid confusion 

Including #9222 with slight modifications

PR #9236, opened 2022-12-09
Remove out of bound SSL log

serrislew:
index variable `i` reinstantiated in inner for loop making ssl debug log call cert name out of bound
- bug from #9225


bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9238, opened 2022-12-09
Add slice config to strip Range header for HEAD requests

serrislew:
Added config in slice plugin to remove Range header from HEAD requests. When enabled, HEAD requests going through slice will always return full object content length
- `@pparam=--strip-range-for-head`
- Docs updated and autest added

### DIFF
--- a/doc/admin-guide/plugins/slice.en.rst
+++ b/doc/admin-guide/plugins/slice.en.rst
@@ -135,6 +135,10 @@ The slice plugin supports the following options::
         can improve cache miss latency.
         -f for short
 
+    --enable-strip-range (optional)
+        Enable slice plugin to strip Range header for HEAD requests.
+        -h for short
+
 Examples::
 
     @plugin=slice.so @pparam=--blockbytes=1000000 @plugin=cache_range_requests.so
--- a/doc/admin-guide/plugins/slice.en.rst
+++ b/doc/admin-guide/plugins/slice.en.rst
@@ -135,7 +135,7 @@ The slice plugin supports the following options::
         can improve cache miss latency.
         -f for short
 
-    --enable-strip-range (optional)
+    --strip-range-for-head (optional)
         Enable slice plugin to strip Range header for HEAD requests.
         -h for short
 
--- a/tests/gold_tests/pluginTest/slice/replay/slice_range.replay.yaml
+++ b/tests/gold_tests/pluginTest/slice/replay/slice_range.replay.yaml
@@ -1,0 +1,82 @@
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+meta:
+  version: "1.0"
+
+sessions:
+- transactions:
+  - client-request:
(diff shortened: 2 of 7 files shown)


traeak (dismissed):
Tests run fine, looks okay.

doc/admin-guide/plugins/slice.en.rst
```diff
@@ -135,6 +135,10 @@ The slice plugin supports the following options::
         can improve cache miss latency.
         -f for short
 
+    --enable-strip-range (optional)
```
bryancall, line 138: Would be nice to have head in the option name.

bryancall (changes requested):

serrislew:
[approve ci autest]

tests/gold_tests/pluginTest/slice/replay/slice_range.replay.yaml
```diff
@@ -0,0 +1,82 @@
...
+        fields:
+        - [ Content-Length, 6 ]
+        - [ X-Response, response ]
+        - [ Content-Range, bytes 0-5/16 ]
+
+    proxy-response:
+      status: 206
+      headers:
+        fields:
+        - [ Content-Length, 6 ]
```
bneradt, line 81: Do we expect this to have a Content-Length header? Proxy Verifier does support proxy-request and proxy-response headers:fields nodes without verification directives, but those are generally useful for something like the traffic_dump plugin which records in real-time what those header fields were at the time the replay files were recorded. In those instances, they are not helpful for replay, per se, but are helpful for someone reading the file to see what headers the proxy actually sent (either toward the origin or toward the client). For an AuTest, proxy-* header fields will almost always have verification directives with them, and will simply be absent for the headers we don't care about verifying (usually Date, Via, etc.).

That said, we likely do want to verify Content-Range?

bneradt (changes requested):
Looks good to me. I just have a few minor test suggestions.

tests/gold_tests/pluginTest/slice/replay/slice_range.replay.yaml
serrislew: So to be clear, are you saying the Content-Length header doesn't have verification directives for replay as opposed to the Content-Range header that does? I don't think I quite understand what you mean by verification directives for these header fields.

bneradt: Thanks for asking. No:  both `Content-Length` and `Content-Range` fields can have verification directives applied to them. But I'm pointing out that this proxy-response in this replay file as written does not verify anything with respect to Content-Length because no verification directive is set for it.

For example, the following `proxy-response` node only _records_ a Content-Length field but does no _verification_ for it:

```
    proxy-response:
       headers:
         fields:
         - [ Content-Length, 6 ]
```

This may be valid. For instance, in a traffic dump this would record that the proxy (ATS) had a Content-Length header of value 6 for this transaction. But during replay of this transaction, this would functionally be a no-op. No verification would be done. That's why it would be very uncommon in an AuTest to have a field without a verification directive in a `proxy-*` node. That is, since AuTest replay files generally function to replay traffic and verify results, rather than to record traffic like in traffic_dump, I would normally expect there to be replay directives in any `proxy-*` header fields in AuTests.

In order to verify something about Content-Length, you would need some verification directive like you did elsewhere for other fields in this replay file. For example, this would verify the existence of the Content-Length field regardless of its value:


```
    proxy-response:
       headers:
         fields:
         - [ Content-Length, { as: present } ]
```

Not the `{ as: present }` directive. That tells the verifier client to verify that there is some Content-Length field.

bneradt: In case it's helpful, here's the docs on Proxy Verifier field verification directives:
https://github.com/yahoo/proxy-verifier#field-verification

serrislew: Oh I see, that makes more sense! Thank you for explaining. I will update the autest

bneradt (approved):
Looks good to me.

bneradt:
[approve ci autest]

bryancall (approved):

ezelkow1:
id be +1 on this for 9.2.1, it's an experimental plugin so not a large change to core code and its pretty isolated in the change as well

zwoop:
Cherry-picked to v9.2.x

PR #9239, opened 2022-12-13
Fix ordering Via response header values

hnakamur:
Currently, when upstream returns 304 response, the order of Via response header values is wrong.

```
Expected order: upstream's Via, our Via
Actual order: our Via, upstream's Via
```

This pull request fixes this.

### DIFF
--- a/proxy/http/HttpTransact.cc
+++ b/proxy/http/HttpTransact.cc
@@ -4614,11 +4614,17 @@ HttpTransact::handle_cache_operation_on_forward_server_response(State *s)
     //  the order of the fields
     MIMEField *resp_via = s->hdr_info.server_response.field_find(MIME_FIELD_VIA, MIME_LEN_VIA);
     if (resp_via) {
+      int saved_our_via_len = 0;
+      char saved_our_via[1024]; // 512-bytes for hostname+via string, 512-bytes for the debug info
       MIMEField *our_via;
       our_via = s->hdr_info.client_response.field_find(MIME_FIELD_VIA, MIME_LEN_VIA);
       if (our_via == nullptr) {
         our_via = s->hdr_info.client_response.field_create(MIME_FIELD_VIA, MIME_LEN_VIA);
         s->hdr_info.client_response.field_attach(our_via);
+      } else {
+        const char *src = our_via->value_get(&saved_our_via_len);
+        memcpy(saved_our_via, src, saved_our_via_len);
+        s->hdr_info.client_response.field_value_set(our_via, "", 0, true);
       }
       // HDR FIX ME - Multiple appends are VERY slow
       while (resp_via) {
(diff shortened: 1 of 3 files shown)


proxy/http/HttpTransact.cc
```diff
@@ -4614,11 +4614,17 @@ HttpTransact::handle_cache_operation_on_forward_server_response(State *s)
...
+      int saved_our_via_len = 0;
+      char saved_our_via[1024]; // 512-bytes for hostname+via string, 512-bytes for the debug info
       MIMEField *our_via;
       our_via = s->hdr_info.client_response.field_find(MIME_FIELD_VIA, MIME_LEN_VIA);
       if (our_via == nullptr) {
         our_via = s->hdr_info.client_response.field_create(MIME_FIELD_VIA, MIME_LEN_VIA);
         s->hdr_info.client_response.field_attach(our_via);
+      } else {
+        const char *src = our_via->value_get(&saved_our_via_len);
+        memcpy(saved_our_via, src, saved_our_via_len);
```
bneradt, line 4626: We need to be defensive against buffer overrun corruptions. Rather than relying on 1024 being more than enough bytes, let's replace `saved_our_via` with a `std::string`. The `field_value_append` can then use `saved_our_via.c_str()`.

bneradt (changes requested):
Thank you for contributing.

proxy/http/HttpTransact.cc
hnakamur: Thanks for your review.

I suppose we are safe here since the value of `our_via` is created in `HttpTransactHeaders::insert_via_header_in_response` and the size of `saved_our_via` is matched to the size of the local variable `new_via_string`.

https://github.com/apache/trafficserver/blob/e8ccdb21b95a3f370be5ed974cb86ea64c435662/proxy/http/HttpTransactHeaders.cc#L842-L845

I thought it would be nice to avoid a heap allocation.
Is it better to use `std::string` here even it causes a heap allocation?
Or keep it as is and add a comment that the size of `our_via` is matchedd to `new_via_string` in `HttpTransactHeaders::insert_via_header_in_response`?

hnakamur: I made a commit using `std::string` at https://github.com/hnakamur/trafficserver/commit/202816a4f9f4dd77b8507599a5eb5fc20c1aea94.
If you prefer this version, I am fine to update this pull request.

masaori335: No heap allocation is the best choice. In this case, `1024` looks good to me.

If the buffer size is dynamic, please consider using `ts::LocalBuffer` instead of `std::string`. It switches stack and heap depends on the size.

https://github.com/apache/trafficserver/blob/master/include/tscpp/util/LocalBuffer.h

bneradt: If we don't want standard string, that's fine. But we can't do an unguarded write on that buffer. We should check that the size is less than the `sizeof(saved_our_via)`.

hnakamur: Thanks for your comments. I added an commit using `ts::LocalBufferWriter`.

masaori335 (approved):
Looks good to me.

`FixedBufferWriter::write(const void *data, size_t length)` has the buffer size check that we talked about.
https://github.com/apache/trafficserver/blob/ac0c7640cc6fcc299f40a4fda3a3d207f95b50f0/include/tscore/BufferWriter.h#L254-L270

bneradt (approved):

PR #9245, opened 2022-12-13
Add enable option to xdebug plugin

maskit:
This adds an option to selectively enable debug features.

This is an incompatible change. Please do not back port to old versions.


maskit:
Closing because this is incompatible change and it should be landed on 10-Dev.

PR #9246, opened 2022-12-13
Change Content-Type to text/plain if xdebug modifies response body

maskit:
Since the response body is modified, the original content type doesn't make sense. The original content-length header value  is kept as x-original-content-length.


maskit:
For 9.1.x https://github.com/apache/trafficserver/pull/9248

bryancall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9250, opened 2022-12-14
Remove a left over directory from the new RPC impl.

brbzull0:
removing folder `mgmt2` which is no longer used.


brbzull0:
[approve ci autest]

randall (approved):

PR #9253, opened 2022-12-14
Add release procedures doc

ezelkow1:
Some basic steps for creating a new release, so I dont have to refer to old slack messages that may disappear

### DIFF
--- a/doc/release-notes/index.en.rst
+++ b/doc/release-notes/index.en.rst
@@ -26,3 +26,4 @@ Release Notes
   whats-new.en
   upgrading.en
   roadmap.en
+  release-procedures.en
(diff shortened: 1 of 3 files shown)


maskit (commented):
Looks good, but I leave just a few comments because I've never done a release.

We need to make an announce on the lists. Not only the dev@ and users@ but also announce@t.a.o and announce@a.o at last.

It might be nice to have an email template for the vote and the release announce.

ezelkow1:
Sure Im looking for input right now. This was just my initial list of steps that I had written down that are like the bare minimum procedures

bryancall (approved):
I am good with this, we can add stuff later

ezelkow1:
ok sounds good, will just merge this in now and we can keep adding to it

PR #9255, opened 2022-12-15
Make plugin DSO work on macOS

maskit:
`dlopen` doesn't return an error even if target file is invalid at all on macOS. We need to call `dlopen_preflight` to examine target file.
With this fix, test_PluginDso passes on macOS.


maskit:
We might want to have this on 9.2.x so we can run CI jobs on macOS, but there's no immediate need.

bneradt (approved):

PR #9256, opened 2022-12-15
Use snprintf instead of sprintf

maskit:
This is a following work of #9189. Not complete, but it makes my build happy.

### DIFF
--- a/example/plugins/c-api/denylist_1/denylist_1.c
+++ b/example/plugins/c-api/denylist_1/denylist_1.c
@@ -162,7 +162,7 @@ handle_response(TSHttpTxn txnp, TSCont contp ATS_UNUSED)
   buf = (char *)TSmalloc(4096);
 
   url_str = TSUrlStringGet(bufp, url_loc, &url_length);
-  sprintf(buf, "You are forbidden from accessing \"%s\"\n", url_str);
+  snprintf(buf, 4096, "You are forbidden from accessing \"%s\"\n", url_str);
   TSfree(url_str);
   TSHandleMLocRelease(bufp, hdr_loc, url_loc);
   TSHandleMLocRelease(bufp, TS_NULL_MLOC, hdr_loc);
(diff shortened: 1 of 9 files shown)


example/plugins/c-api/denylist_1/denylist_1.c
```diff
@@ -179,7 +179,7 @@ read_denylist(TSCont contp)
   char denylist_file[1024];
   TSFile file;
 
-  sprintf(denylist_file, "%s/denylist.txt", TSPluginDirGet());
+  snprintf(denylist_file, 1024, "%s/denylist.txt", TSPluginDirGet());
```
ywkaras, line 182: nitpick:  in this case you could use `sizeof(denylist_file)` so the size only needs to be changed in on place.

example/plugins/c-api/secure_link/secure_link.c
```diff
@@ -125,7 +125,7 @@ TSRemapDoRemap(void *ih, TSHttpTxn rh, TSRemapRequestInfo *rri)
   EVP_MD_CTX_destroy(ctx);
 #endif
   for (i = 0; i < MD5_DIGEST_LENGTH; i++) {
-    sprintf(&hash[i * 2], "%02x", md[i]);
+    snprintf(&hash[i * 2], 32, "%02x", md[i]);
```
ywkaras, line 128:
```
wkaras /tmp/x
O$ cat x.cc
char x[3][32];

void f(int i)
{
  static_assert(sizeof(x[i]) == 32);
}
wkaras /tmp/x
O$ gcc -c -pedantic -Wall -Wextra -std=c++17 x.cc
wkaras /tmp/x
O$
```
Strongly recommend using `sizeof(hash[i * 2])` instead of `32`.

maskit: I don't understand this one. Do we have another `hash`? I only find `char hash[32]`.

maskit: It was wrong in either way. Fixed.

ywkaras (approved):

PR #9257, opened 2022-12-16
Remove mysql_remap plugin

maskit:
As the documentation says, mysql_remap plugin is deprecated as of v9.0.0. And it uses sprintf. I'd like to remove the plugin without fixing the sprintf issue.

This also removes the check for mysql library in the configure script.

```
$ git grep -i mysql | wc
       0       0       0
```


shukitchan:
[approve ci fedora]

shukitchan (approved):

bneradt (approved):

PR #9258, opened 2022-12-16
Remove cqhv log field

maskit:
This removes `cqhv` log field. It was deprecated on 9.0.0.

https://lists.apache.org/thread/zqmo3sj1m1s6tv9b7zdo86q0qpf2mjow

Unfortunately, we can't remove the marshaling and unmarshaling functions. The marshaling function is also used for `cqtx`, which reproduces HTTP/1.x request line. The unmarshaling function is also used for server request.

### DIFF
--- a/proxy/logging/LogAccess.cc
+++ b/proxy/logging/LogAccess.cc
@@ -917,7 +917,7 @@ LogAccess::unmarshal_http_version(char **buf, char *dest, int len)
   LogAccess::unmarshal_http_text
 
   The http text is simply the fields http_method (cqhm) + url (pqu) +
-  http_version (cqhv), all right next to each other, in that order.
+  http_version (cqhv, which was removed on 10.0.0), all right next to each other, in that order.
   -------------------------------------------------------------------------*/
 
 int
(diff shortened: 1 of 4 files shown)


proxy/logging/LogAccess.cc
```diff
@@ -917,7 +917,7 @@ LogAccess::unmarshal_http_version(char **buf, char *dest, int len)
   LogAccess::unmarshal_http_text
 
   The http text is simply the fields http_method (cqhm) + url (pqu) +
-  http_version (cqhv), all right next to each other, in that order.
+  http_version (cqhv, which was removed on 10.0.0), all right next to each other, in that order.
```
zizhong, line 920: should we remove the mentioning of cqhv?

maskit: Yea, I thought about it, but I was too lazy. How about this:
> The http text is a reproduced HTTP/1.x request line. It's HTTP method (cqhm) + URL (pqu) + HTTP version. This doesn't support HTTP/2 and HTTP/3 since those don't have a request line.

zizhong: sounds good!

zizhong (approved):
LG!

bneradt (approved):

PR #9268, opened 2022-12-20
Make net.default_inactivity_timeout overridable

bneradt:
This makes the default_inactivity_timeout configuration an overridable configuration.

---

# Review Notes

Much of the churn in this patch is boilerplate for supporting the overridable feature for this change (doc changes, lua changes, override enum changes, etc.). It's valuable to skim those changes, but the interesting parts of this patch are:

* The InactivityCop changes in `iocore/net/UnixNet.cc`
*  The setting of the override in `proxy/http/HttpTransact.cc`


bneradt:
[approve ci]

iocore/net/UnixNet.cc
```diff
@@ -70,20 +71,30 @@ class InactivityCop : public Continuation
...
       // set a default inactivity timeout if one is not set
       // The event `EVENT_INACTIVITY_TIMEOUT` only be triggered if a read
       // or write I/O operation was set by `do_io_read()` or `do_io_write()`.
-      if (ne->next_inactivity_timeout_at == 0 && nh.config.default_inactivity_timeout > 0 &&
-          (ne->read.enabled || ne->write.enabled)) {
+      if (ne->next_inactivity_timeout_at == 0 && ne->default_inactivity_timeout_in > 0 && (ne->read.enabled || ne->write.enabled)) {
         Debug("inactivity_cop", "vc: %p inactivity timeout not set, setting a default of %d", ne,
               nh.config.default_inactivity_timeout);
-        ne->set_default_inactivity_timeout(HRTIME_SECONDS(nh.config.default_inactivity_timeout));
+        ne->use_default_inactivity_timeout = true;
```
serrislew, line 89: Don't we also want to set `inactivity_timeout_in = 0` since that gets removed from the `set_default_inactivity_timeout` function and needed to return true for `is_default_inactivity_timeout`

iocore/net/UnixNet.cc
```diff
@@ -70,20 +71,30 @@ class InactivityCop : public Continuation
         continue;
       }
 
+      if (ne->default_inactivity_timeout_in == -1) {
+        // If no context-specific default inactivity timeout has been set by an
+        // override plugin, then use the global default.
+        Debug("inactivity_cop",
+              "vc: %p setting a the global default inactivity timeout of %d, next_inactivity_timeout_at: %" PRId64, ne,
```
serrislew, line 78: this is nit-picky but I think you can remove the "a"

iocore/net/UnixNet.cc
```diff
@@ -70,20 +71,30 @@ class InactivityCop : public Continuation
...
+              nh.config.default_inactivity_timeout, ne->next_inactivity_timeout_at);
+        ne->set_default_inactivity_timeout(HRTIME_SECONDS(nh.config.default_inactivity_timeout));
```
serrislew, line 80: Not sure when `initialize_thread_for_net` is called relative to when configs get changed but why don't we use `global_config` var, which reads the config values as opposed to `nh->config` which is just a copy that can possibly have stale data? Is it guaranteed to be updated when configs change?

proxy/ProxySession.cc
```diff
@@ -227,6 +227,14 @@ ProxySession::set_inactivity_timeout(ink_hrtime timeout_in)
   }
 }
 
+void
+ProxySession::set_default_inactivity_timeout(ink_hrtime timeout_in)
+{
+  if (_vc) {
+    _vc->set_default_inactivity_timeout(timeout_in);
```
serrislew, line 234: to confirm, if _vc is of type UnixNetVConnection, would we not want it to update `next_inactivity_timeout_at` as that is no longer part of the function?

iocore/net/UnixNet.cc
bneradt: Good catch.

iocore/net/UnixNet.cc
bneradt: That's an interesting idea, but I think we better stick with reading the config from the nethandler here rather than from global_config. Currently global_config is populated into each handler with each handler registering the `update_nethandler_config` for updates that I presume gets processed in a timely fashion. All the configurations are looked at in this file from nethandler, not directly from global_config. That is, notice that global_config is only used for nethandler config initialization and then subsequent updates via the update handler, it's never used to access the runtime values. I'm not sure, but this may be due to threading coordination reasons.

In any case, I'd say we stick with what is done in the rest of this file for this PR. If we want to look into changing things in UnixNet.cc to access global_config directly, let's do that in a separate PR.

proxy/ProxySession.cc
bneradt: This is a good question.

Note that before this patch, `set_default_inactivity_timeout` was called exactly once in UnixNet.cc. That was called off of NetEvent objects which were only ever `UnixNetVConnection` or `QUICNetVConnection` concrete types, both of which took the `UnixNetVConnection::set_default_inactivity_timeout` definition. In the previous use of this function, `set_default_inactivity_timeout` meant: (1) cancel any current activity_timeout_in, (2) set the next_inactivity_timeout_at to the provided default timeout value, and (3) set a boolean indicating that the current inactivity timeout is a default one. In other words, `set_default_inactivity_timeout` didn't mean "set the configured default inactivity timeout to this new value", it meant, "make the next timeout the provided value and indicate that it was set from the default inactivity timeout value". It was an ad-hoc implementation for the inactivity cop's single use of it.

This patch changes the meaning of the function to something more simple: "set the configured default_inactivity_timeout value to this new value". And now it is only called in:

1. In HttpTransact to provide a transaction level override.
2. In InactivityCop if no override has been set to set the global value.

So, to answer your question directly, we no longer set `next_inactivity_timeout_at` because the meaning of `set_default_inactivity_timeout` has changed to mean "set the configured default_inactivity_timeout", not the previous meaning of "configure an inactivity timeout value from this default value".

bneradt:
[approve ci fedora]

bneradt:
[approve ci autest]

proxy/ProxySession.cc
serrislew: I see, thanks for the explanation!

serrislew (approved):

proxy/ProxySession.cc
bneradt: Oh and I should add, in case someone reads this in the future, actually "setting" things to use the default inactivity timeout is done via the cancel_inactivity_timeout function. That cancels the explicit inactivity timeout, which effectively leads to falling back on the default inactivity timeout.

PR #9271, opened 2022-12-21
Add param to forward headers from the auth server to the origin

Ftywan:
Added a header prefix field to the config. Headers from the authentication response with the prefix will be appended to the original request.

new parameter: `@pparam=--forward-header-prefix`

### DIFF
--- a/doc/admin-guide/plugins/authproxy.en.rst
+++ b/doc/admin-guide/plugins/authproxy.en.rst
@@ -80,6 +80,11 @@ Plugin Options
   The TCP port of the authorization host. This is only used by the
   ``redirect`` transform.
 
+--forward-header-prefix=PREFIX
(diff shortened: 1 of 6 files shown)


masaori335:
[approve ci]

mlibbey:
Can we add this option to the documentation? Will all headers that match the prefix that come back from the auth server be proxied to the origin? What's the format of the value (just text)? Is there just one prefix that can be configured? (think the doc is at doc/admin-guide/plugins/authproxy.en.rst)

Ftywan:
Hey @mlibbey, I've added the option usage to the documentation. 
Regarding the queries, only one prefix can be configured in the form of plain text. All matching headers from the authentication server will be proxied to the origin.

bryancall (approved):

PR #9272, opened 2022-12-21
Fix crashes at shutdown due to references to stale debug control registry.

ywkaras:
Adds reference count for references to registry.  The registry is deleted only when the reference count goes to zero.

### DIFF
--- a/src/tscore/DbgCtl.cc
+++ b/src/tscore/DbgCtl.cc
@@ -22,15 +22,15 @@
  */
 
 #include <mutex>
-#include <atomic>
 #include <set>
 #include <cstring>
+#include <atomic>
 
 #include <tscore/ink_assert.h>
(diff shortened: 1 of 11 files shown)


ywkaras:
I think I have to make changes to make sure the registry mutex isn't used after it's destroyed.

ywkaras:
[approve ci autest]

src/tscore/DbgCtl.cc
```diff
@@ -84,6 +132,8 @@ DbgCtl::_get_ptr(char const *tag)
 
   auto &d{ra.data()};
 
+  ++ra.registry_reference_count;
```
wkaras, line 135: Increment of reference count must come before line 133.

cmcfarlen (approved):
This PR looks good and it does prevent crashes at shutdown.

Issue #5876, opened 2019-08-26
not-cache control by HTTP headers are differences between on the actual behavior and on the document


nozomi1773:
traffic server control whether to cache or not by client's request headers or origin's response headers , and this control are differences between on the actual behavior and on the document.

For explaining about actual behavior, I built trafficserver from latest source code ( commit: [05a6f19f99cebc4ec3210467fb9a5527b21c7d8d](https://github.com/apache/trafficserver/commit/05a6f19f99cebc4ec3210467fb9a5527b21c7d8d) ) on master branch, and did gold_tests on my repo ( [nozomi1773/trafficserver-test-docker--headers-to-not-cache](https://github.com/nozomi1773/trafficserver-test-docker--headers-to-not-cache) ).
I entered detailed "via" informations in this [*.test.py files](https://github.com/nozomi1773/trafficserver-test-docker--headers-to-not-cache/tree/master/build_master/tests/gold_tests/headers-to-not-cache).  
Could you check this behavior?  
Is this behavior unexpected? Or is the description of the document wrong?
If the latter, I will make a PR to correct it.

#### client's request headers ( document )
[latest document](https://github.com/apache/trafficserver/blob/master/doc/admin-guide/configuration/cache-basics.en.rst) notes following :
> Client Directives  
> By default, Traffic Server does not cache objects with the following request headers:  
> ・Authorization  
> ・Cache-Control: no-store  
> ・Cache-Control: no-cache  
>   To configure Traffic Server to ignore this request header, refer to Configuring Traffic Server to Ignore Client no-cache Headers.  
> ・Cookie (for text objects)  
>   By default, Traffic Server caches objects served in response to requests that contain cookies (even if the object is text). You can configure Traffic Server to not cache cookied content of any type, cache all cookied content, or cache cookied content that is of image type only. For more information, refer to Caching Cookied Objects.  

and
`Configuring Traffic Server to Ignore Client no-cache Headers` notes configure Traffic Server to ignore Client no-cache Headers:
```
CONFIG proxy.config.http.cache.ignore_client_no_cache INT 1
```
but latest traffic server's default is already 1 .

#### client's request headers ( actual behavior on my gold_tests )  
I checked behavior when origin's response header have cacheable "Cache-Control: max-age=5" with proxy.config.http.cache.required_headers is default(2) .
  
| CONFIG OPTION | client's request header | origin's response header | origin's status code | behavior |
|:-----------|:------------|:------------|:------------|:------------|
[...]

PR #9282, opened 2023-01-03
Updated documentation and added test coverage for some caching behavior 

lzx404243:
#5876 points out some mismatches between the documentation and the current behavior of ATS caching. 

## added coverage
Added AuTest coverage for caching behavior in the following areas to verify behavior:
* no-store and no-cache cache-control directives in request/response
* authorization headers(specifically response containing WWW-Authenticate)
* cookie-related request/response(different configurations specifying which content type to cache)

## documentation
The documentation is updated with the following changes:
* Make it clear that ATS by default requires either `Expires` or `max-age` headers to cache
* Make it clear that ATS by default ignores client `no-cache` and `no-store` directives
* Removed `Authorization` header from the list that ATS by default doesn't cache(since it's cached per testing)
* Make it clear that when `proxy.config.http.cache.ignore_authentication` is set to `1`, the response containing the header would be cached
* Listed out explicitly the `pragma` and `cache-control` directives that ATS ignores when ignore_client_no_cache or ignore_server_no_cache is enabled

### AuTest failure
The newly added `ResponseCacheControlIgnoredTest` is failing, indicating that there are some issues with the `proxy.config.http.cache.ignore_server_no_cache` configuration. The test verifies that ATS ignores the `no-cache` and `no-store` in responses when asked to do so.


bneradt:
[approve ci]

bneradt (changes requested):
Looks great. I just have a few simple word-smithing suggestions for the docs.

Thanks for adding test coverage for these cases.

doc/admin-guide/configuration/cache-basics.en.rst
```diff
@@ -547,18 +551,23 @@ Caching Cookied Objects
...
 -  Not cache cookied content of any type.
 
+-  Cache all cookied content regardless of type.
+
 -  Cache cookied content that is of image type only.
 
--  Cache all cookied content regardless of type.
+-  Cache all cookied content except text type.
+
+   It may be approprite to configure Traffic Server to not cache cookied text
```
randall, line 566: approprite should be appropriate

lzx404243: good catch, thx. Fixed it.

bneradt:
[approve ci autest]

doc/admin-guide/files/records.config.en.rst
```diff
@@ -2207,8 +2207,8 @@ Cache Control
 .. ts:cv:: CONFIG proxy.config.http.cache.ignore_authentication INT 0
    :overridable:
 
-   When enabled (``1``), |TS| ignores ``WWW-Authentication`` headers in responses ``WWW-Authentication`` headers are removed and
-   not cached.
+   When enabled (``1``), |TS| ignores ``WWW-Authentication`` headers in
+   responses and the responses are cached.
```
bneradt, line 2211: We discussed this offline, but just to record it here: It would be explicitly list the pragma and cache-control directives that are ignored for `proxy.config.http.cache.ignore_client_no_cache` and `proxy.config.http.cache.ignore_server_no_cache` so it's clearer to users what is being ignored when those features are enabled.

bneradt (changes requested):

doc/admin-guide/files/records.config.en.rst
lzx404243: Updated the doc for these. Thanks!

bneradt: Looks good.

bneradt (approved):
Looks good.

bneradt:
Marking for 9.2.x since this is a helpful doc and test-only change.

randall (approved):

PR #9285, opened 2023-01-04
Do not track config file periodically

vmamidi:


mgmt/RecordsConfig.cc
```diff
@@ -263,6 +263,8 @@ static const RecordElement RecordsConfig[] =
   ,
   {RECT_CONFIG, "proxy.config.admin.api.restricted", RECD_INT, "0", RECU_RESTART_TM, RR_NULL, RECC_INT, "[0-1]", RECA_NULL}
   ,
+  {RECT_CONFIG, "proxy.config.periodic_track_config_files", RECD_INT, "0", RECU_RESTART_TM, RR_NULL, RECC_INT, "[0-1]", RECA_NULL}
```
maskit, line 266: I understand this is not going to be a small change, but I feel like `config_tracking_interval` that accepts 0 (no tracking) or N (second) is a more common way of configuring this kind of things on ATS.

vmamidi: +1

vmamidi: I will see if it is easy to do that.

vmamidi:
@maskit updated

vmamidi:
[approve ci autest]

serrislew:
[approve ci autest]

vmamidi:
[approve ci autest]

serrislew:
[approve ci autest]

doc/admin-guide/files/records.config.en.rst
```diff
@@ -601,6 +601,14 @@ Local Manager
    This setting is not reloadable, since it is must be applied when
    :program:`traffic_manager` initializes.
 
+.. ts:cv:: CONFIG proxy.config.track_config_files INT 1
+
+   Enables (``1``) or disables (``0``) tracking configuration file updates.
+   This setting is enabled by default. Disabling this setting disables the config status of traffic_ctl.
```
bneradt, line 607: Let's describe a motivation for this configuration. Here's a suggestion for the first paragraph:

```
Enables (``1``) or disables (``0``) tracking configuration file updates.
This setting is enabled by default, meaning that configuration files are monitored for changes.
Having tracking enabled is a dependency for :option:`traffic_ctl config status` to function. However,
tracking the files is implemented via a frequent call to ``stat()`` which may be problematic
in some deployments. If the call to ``stat()`` on configuration files causes problems, then
it can be avoided by setting this value to ``0`` at the cost of disabling the config status feature
for :program:`traffic_ctl`.
```

bneradt (changes requested):
Pretty straightforward. I just have a couple small suggestions.

src/records/RecLocal.cc
```diff
@@ -71,7 +71,9 @@ sync_thr(void *data)
     RecSyncStatsFile();
 
     // If we didn't successfully sync to disk, check whether we need to update ....
-    if (check) {
+    bool found;
+    int track_time = static_cast<int>(REC_readInteger("proxy.config.track_config_files", &found));
+    if (found && track_time > 0 && check) {
```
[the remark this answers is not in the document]
vmamidi: yeah, i thought so too

doc/admin-guide/files/records.config.en.rst
vmamidi: updated..

bneradt (approved):

bneradt:
@vmamidi : can you check the applicability of this to 10-Dev? The merge from master to 10-Dev was trivial, but it doesn't really apply, as such, because there is no traffic_manager. We may need to check whether this has an analogous change that is needed in 10-Dev.

Issue #9229, opened 2022-12-02
Assertion failure in HttpSM::tunnel_handler_post_ua


maskit:
We saw crashes (assertion failures) while we test 9.2.0 on our production and we think #8301 is the cause (or it made things worse at a minimum).

The change made it possible to handle POST data and response data in parallel, but it also introduced a race. As you can see on the error log, ATS setup a tunnel to send POST data after sending a 502 response to the client. Obviously the tunnel is unnecessary, and this out of order setup messes up internal state and causes the assertion failure.

I was able to reproduce the abort on 9.2.x and master, on my laptop with the setup below:

Origin server (this immediately sends a broken response right after connection establishment)
```
$ echo HTTP | openssl s_server -accept 12345 -cert server.crt -key server.key
```

Remap
```
map / https://127.0.0.1:12345/
```

Client (just sends a POST request)
```
$ curl -vk --limit-rate 1 --data-binary @file https://localhost:8443/
```

Backtrace
```
(lldb) bt
* thread #2, name = '[ET_NET 0]', stop reason = signal SIGABRT
  * frame #0: 0x000000018ea3ad98 libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x000000018ea6fee0 libsystem_pthread.dylib`pthread_kill + 288
    frame #2: 0x000000018e9aa340 libsystem_c.dylib`abort + 168
    frame #3: 0x000000010518a938 libtscore.10.dylib`ink_abort(message_format="%s:%d: failed assertion `%s`") at ink_error.cc:99:3
    frame #4: 0x000000010517fe4c libtscore.10.dylib`::_ink_assert(expression="ua_entry->in_tunnel == true", file="HttpSM.cc", line=3609) at ink_assert.cc:37:3
    frame #5: 0x00000001003024c4 traffic_server`HttpSM::tunnel_handler_post_ua(this=0x0000000113eec800, event=105, p=0x0000000113eedad8) at HttpSM.cc:3609:5
    frame #6: 0x00000001004e982c traffic_server`HttpTunnel::producer_handler(this=0x0000000113eed7b0, event=105, p=0x0000000113eedad8) at HttpTunnel.cc:1250:7
    frame #7: 0x00000001004df7f4 traffic_server`HttpTunnel::main_handler(this=0x0000000113eed7b0, event=105, data=0x0000000113d8aaf0) at HttpTunnel.cc:1664:19
[...]

PR #9287, opened 2023-01-06
Fix a crash from a server early abort

bneradt:
While adding test coverage around #9229, it was found that ATS will crash if the server both replies early to a request (i.e., starts replying before the body of the request is received), and then aborts early. This addresses that crash.

---

## For Review

Unfortunately, github isn't recognizing that I renamed the test utility slow_post_client.py to slow_post_clients.py, and that I then re-purposed a new slow_post_client.py for this test. Thus:

* slow_post_clients.py is an unmodified rename from slow_post_client.py. There's nothing really to review with the contents of that file.
* slow_post_client.py is a new file. Trying to read the diff from old to new is not helpful since the contents are brand new.

### DIFF
--- a/proxy/http/HttpSM.cc
+++ b/proxy/http/HttpSM.cc
@@ -5814,7 +5814,7 @@ HttpSM::handle_server_setup_error(int event, void *data)
     ink_release_assert(0);
   }
 
-  if (event == VC_EVENT_INACTIVITY_TIMEOUT || event == VC_EVENT_ERROR) {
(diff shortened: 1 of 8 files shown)


bneradt:
[approve ci autest]

bneradt:
[approve ci autest]

maskit (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9288, opened 2023-01-06
Updated logic for server no-cache and no-store

### LINKED ISSUE #9283: no-cache and no-store in responses aren't ignored when proxy.config.http.cache.ignore_server_no_cache is enabled
Per documentation, Traffic Server should ignore origin server requests to bypass the cache when `proxy.config.http.cache.ignore_server_no_cache` is enabled. However, in my testing even after setting`proxy.config.http.cache.ignore_server_no_cache = 1`, the `no-cache` and `no-store` in responses are still honored (not ignored) by ATS, and the responses are not cached as a result. 

It seems there are some issues with the `proxy.config.http.cache.ignore_server_no_cache` config.

lzx404243:
This PR fixes #9283. 

### issues prior to this change
* With `ignore_server_no_cache = 0`(honoring the directives), ATS would determine responses containing `no-cache` as non-cacheable, while those should be cacheable but just require revalidation on future requests.
* With `ignore_server_no_cache = 1`(ignoring the directives), ATS fails to ignore the 'no-cache' and `no-store` in responses while it is asked to. 

After this change, ATS handles `no-cache` and `no-store` in responses correctly, both when configured to honor and ignore them.

### DIFF
--- a/proxy/http/HttpTransactHeaders.cc
+++ b/proxy/http/HttpTransactHeaders.cc
@@ -421,14 +421,14 @@ HttpTransactHeaders::calculate_document_age(ink_time_t request_time, ink_time_t 
 }
 
 bool
-HttpTransactHeaders::does_server_allow_response_to_be_stored(HTTPHdr *resp)
+HttpTransactHeaders::does_server_allow_response_to_be_stored(HTTPHdr *resp, bool ignore_no_store_and_no_cache_directives)
 {
-  uint32_t cc_mask = (MIME_COOKED_MASK_CC_NO_CACHE | MIME_COOKED_MASK_CC_NO_STORE | MIME_COOKED_MASK_CC_PRIVATE);
+  uint32_t cc_mask = MIME_COOKED_MASK_CC_PRIVATE | (ignore_no_store_and_no_cache_directives ? 0 : MIME_COOKED_MASK_CC_NO_STORE);
 
   // According to https://www.rfc-editor.org/rfc/rfc7234#section-5.4
   // ... When the Cache-Control header field is also present and
   // understood in a request, Pragma is ignored.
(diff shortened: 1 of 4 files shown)


proxy/http/HttpTransactHeaders.cc
```diff
@@ -421,14 +421,14 @@ HttpTransactHeaders::calculate_document_age(ink_time_t request_time, ink_time_t
 }
 
 bool
-HttpTransactHeaders::does_server_allow_response_to_be_stored(HTTPHdr *resp)
+HttpTransactHeaders::does_server_allow_response_to_be_stored(HTTPHdr *resp, bool ignore_no_store_and_no_cache_directives)
 {
-  uint32_t cc_mask = (MIME_COOKED_MASK_CC_NO_CACHE | MIME_COOKED_MASK_CC_NO_STORE | MIME_COOKED_MASK_CC_PRIVATE);
+  uint32_t cc_mask = MIME_COOKED_MASK_CC_PRIVATE | (ignore_no_store_and_no_cache_directives ? 0 : MIME_COOKED_MASK_CC_NO_STORE);
```
bneradt, line 426: Note that removing `MIME_COOKED_MASK_CC_NO_CACHE` is intentional here because, per the RFC, `no-cache` doesn't mean that the proxy can't store the response (that's what `no-store` means), rather it means that future requests for the object has to revalidate it with the origin before sending the cached response to the client.

bneradt (approved):
Makes sense.

bneradt:
@zwoop and I discussed this and we decided to mark this for 9.2.x with the intention that it will go into 9.2.1.

cmcfarlen (approved):

PR #9297, opened 2023-01-10
Optimize HPACK string comparison

maskit:
When compare strings in HPACK, we don't care which string comes first on a dictionary. We just want to know the two strings are exactly the same.

This PR replaces memcmp(sv, sv) and strcasecmp(sv, sv) with local functions that immediately return `false` when the lengths don't match.

Before:
```
finished in 14.74s, 6784.85 req/s, 53.19MB/s
requests: 100000 total, 100000 started, 100000 done, 100000 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 100000 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 783.92MB (822000249) total, 976.78KB (1000219) headers (space savings 96.45%), 781.25MB (819200000) data
time for request:      860us      3.60ms      1.45ms       179us    89.26%
time for connect:     4.08ms      4.08ms      4.08ms         0us   100.00%
time to 1st byte:     7.58ms      7.58ms      7.58ms         0us   100.00%
req/s           :    6784.90     6784.90     6784.90        0.00   100.00%
```

After:
```
finished in 12.77s, 7829.25 req/s, 61.38MB/s
requests: 100000 total, 100000 started, 100000 done, 100000 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 100000 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 783.92MB (822000238) total, 976.77KB (1000208) headers (space savings 96.45%), 781.25MB (819200000) data
time for request:      797us      2.85ms      1.25ms       155us    89.15%
time for connect:     2.79ms      2.79ms      2.79ms         0us   100.00%
time to 1st byte:     5.49ms      5.49ms      5.49ms         0us   100.00%
req/s           :    7829.30     7829.30     7829.30        0.00   100.00%
```

Before:
image

After:
image

### DIFF
--- a/proxy/http2/HPACK.cc
+++ b/proxy/http2/HPACK.cc
@@ -191,6 +191,7 @@ hpack_field_is_literal(HpackField ftype)
 
 // Try not to use memcmp(sv, sv) and strncasecmp(sv, sv) because we don't care which value comes first on a dictionary.
 // Return immediately if the lengths of given strings don't match.
+// Also, we noticed with profiling that taking char* and int was more performant than taking std::string_view.
 static inline bool
 match(const char *s1, int s1_len, const char *s2, int s2_len)
 {
@@ -266,6 +267,7 @@ namespace HpackStaticTable
     HpackLookupResult result;
 
     for (unsigned int index = 1; index < TS_HPACK_STATIC_TABLE_ENTRY_NUM; ++index) {
+      // Profiling showed that use of const reference here is more performant than copying string_views.
       const std::string_view &name  = STATIC_TABLE[index].name;
       const std::string_view &value = STATIC_TABLE[index].value;
 
--- a/proxy/http2/HPACK.cc
+++ b/proxy/http2/HPACK.cc
@@ -189,6 +189,44 @@ hpack_field_is_literal(HpackField ftype)
   return ftype == HpackField::INDEXED_LITERAL || ftype == HpackField::NOINDEX_LITERAL || ftype == HpackField::NEVERINDEX_LITERAL;
 }
 
+// Try not to use memcmp(sv, sv) and strncasecmp(sv, sv) because we don't care which value comes first on a dictionary.
+// Return immediately if the lengths of given strings don't match.
+static inline bool
+match(const char *s1, int s1_len, const char *s2, int s2_len)
+{
+  if (s1_len != s2_len) {
+    return false;
+  }
+
+  if (s1 == s2) {
+    return true;
+  }
+
+  if (memcmp(s1, s2, s1_len) != 0) {
+    return false;
+  }
+
+  return true;
+}
+
+static inline bool
+match_ignore_case(const char *s1, int s1_len, const char *s2, int s2_len)
+{
+  if (s1_len != s2_len) {
+    return false;
+  }
+
+  if (s1 == s2) {
+    return true;
+  }
+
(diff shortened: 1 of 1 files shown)


proxy/http2/HPACK.cc
```diff
@@ -189,6 +189,44 @@ hpack_field_is_literal(HpackField ftype)
   return ftype == HpackField::INDEXED_LITERAL || ftype == HpackField::NOINDEX_LITERAL || ftype == HpackField::NEVERINDEX_LITERAL;
 }
 
+// Try not to use memcmp(sv, sv) and strncasecmp(sv, sv) because we don't care which value comes first on a dictionary.
+// Return immediately if the lengths of given strings don't match.
+static inline bool
+match(const char *s1, int s1_len, const char *s2, int s2_len)
```
bneradt, line 195: It seems like the caller of these functions would be better served with these functions taking std::string_view parameters.

proxy/http2/HPACK.cc
```diff
@@ -228,12 +266,12 @@ namespace HpackStaticTable
     HpackLookupResult result;
 
     for (unsigned int index = 1; index < TS_HPACK_STATIC_TABLE_ENTRY_NUM; ++index) {
-      std::string_view name  = STATIC_TABLE[index].name;
-      std::string_view value = STATIC_TABLE[index].value;
+      const std::string_view &name  = STATIC_TABLE[index].name;
+      const std::string_view &value = STATIC_TABLE[index].value;
```
bneradt, line 270: Is the const reference cheaper than copying the pointer and length values of original `name` and `value` types, as the original did? Each time these are accessed, it will now require a pointer dereference.

proxy/http2/HPACK.cc
maskit: I found use of string_view actually costs a little in some cases. It probably depends on how it'll be optimized, but I was able to prune away time spent for string_view::data() and string_view::length() by receiving unwrapped pointers and lengths. This function is really busy and tiny optimization works, so I'd like to ensure remove any possibility of unfortunate cases.

proxy/http2/HPACK.cc
maskit: This also depends on how it'll be optimized probably. I don't know why but I saw memcpy without this change. I presume, by having these as const references, they are handled as aliases (no copies but direct access to the name, and the value).

bneradt: That is really strange. I wouldn't expect the creation of a string view to have a memcpy...a string_view is simply nothing more than a pointer to a char* and a length (int type) value.

proxy/http2/HPACK.cc
bneradt: Sounds good. If you saw an improvement by taking char* and int, then that's good. Can you please add a comment explaining this though? Something like, `// We noticed with profiling that taking char* and int was more performant than taking std::string_view.` Otherwise someone might think to refactor by taking string_view parameters since that is what the users have.

proxy/http2/HPACK.cc
bneradt: In any case, if your profiling found that const references are more performant, let's run with that. Can you also add a comment here explaining that the const references were more performant than std::string_view copies?

maskit:
Thank you for the suggestions. I added the comments.

bneradt (approved):
Looks good.

PR #9306, opened 2023-01-13
Updated warning message for adding duplicated remap entries

lzx404243:
Current logs when the user specifies duplicated remap rules:
```
[Jan 10 23:34:36.437] traffic_server ERROR: Couldn't insert into trie!
[Jan 10 23:34:36.437] traffic_server WARNING: Could not insert new mapping
[Jan 10 23:34:36.437] traffic_server ERROR: [ReverseProxy] failed to add remap rule at /tmp/sb/duplicated_remap_entries/ts2/config/remap.config line 2: unable to add mapping rule to lookup table
[Jan 10 23:34:36.437] traffic_server WARNING: something failed during BuildTable() -- check your remap plugins!
[Jan 10 23:34:36.437] traffic_server EMERGENCY: remap.config failed to load
```
This PR changes the `Could not insert new mapping` to `Could not insert new mapping: duplicated entry exists.` to make it more explicit why the rule insertion fails.

### DIFF
--- a/proxy/http/remap/UrlRewrite.cc
+++ b/proxy/http/remap/UrlRewrite.cc
@@ -769,7 +769,7 @@ UrlRewrite::TableInsert(std::unique_ptr<URLTable> &h_table, url_mapping *mapping
     h_table->emplace(src_host, ht_contents);
   }
   if (!ht_contents->Insert(mapping)) {
-    Warning("Could not insert new mapping");
+    Warning("Could not insert new mapping: duplicated entry exists.");
     return false;
   }
   return true;
--- a/proxy/http/remap/UrlRewrite.cc
+++ b/proxy/http/remap/UrlRewrite.cc
@@ -769,7 +769,8 @@ UrlRewrite::TableInsert(std::unique_ptr<URLTable> &h_table, url_mapping *mapping
     h_table->emplace(src_host, ht_contents);
   }
   if (!ht_contents->Insert(mapping)) {
-    Warning("Could not insert new mapping: duplicated entry exists.");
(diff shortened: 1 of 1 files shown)


proxy/http/remap/UrlRewrite.cc
```diff
@@ -769,7 +769,7 @@ UrlRewrite::TableInsert(std::unique_ptr<URLTable> &h_table, url_mapping *mapping
     h_table->emplace(src_host, ht_contents);
   }
   if (!ht_contents->Insert(mapping)) {
-    Warning("Could not insert new mapping");
+    Warning("Could not insert new mapping: duplicated entry exists.");
```
bneradt, line 772: It's not immediately obvious here that the only reason that `Insert` would return false is that there is a duplicate. Looking at [the code](https://github.com/apache/trafficserver/blob/f1ece8382a54fcbc7e3416e427b26d25b3177986/include/tscore/Trie.h#L169), as I'm sure you did, that seems to be the case. Let's add a comment assuring the reader of the code that this is true:

```
// Trie::Insert only fails due to an attempt to add a duplicate entry.
```

bneradt (changes requested):
Looks good. This will be helpful to people tripping on this.

proxy/http/remap/UrlRewrite.cc
lzx404243: Yeah this can be helpful. Added it.

bneradt (approved):

maskit (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9311, opened 2023-01-17
Docs: Unpin Sphinx

bneradt:
When Sphinx 4.x was initially released there were rendering issues with it building the ATS docs. To address this we simply pinned to the 3.x releases. Currently the latest Sphinx release is v6.1.3 and I noticed that unpinning doesn't have issues with our code anymore. This therefore unpins Sphinx. This also addresses a deprecated use of set_class from traffic-server.py.

### DIFF
--- a/doc/Pipfile
+++ b/doc/Pipfile
@@ -23,18 +23,10 @@ verify_ssl = true
 
 [packages]
 
-# The latest 4.x sphinx release has issues with style rendering. The 3.x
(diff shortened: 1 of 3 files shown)


zwoop (dismissed):
This resolves the Python version issues, and the set_class() changes fixes the "make html" targets.

zwoop (approved):

zwoop:
Cherry-picked to v9.2.x

Issue #5506, opened 2019-05-15
QUIC: Assert when udp_thread is 0


scw00:
Otherwise coredump .

zwoop:
Closing — fixed by #9314, which registers `ET_UDP` unconditionally so `QUICPacketHandler` no longer dereferences a null event loop when `udp_thread` is 0. Please reopen if the assert still reproduces on a recent build.

PR #9314, opened 2023-01-17
Register ET_UDP thread type even if no UDP threads are requested

maskit:
ET_UDP has to have a valid value because it's used to check the number of threads for the type.
https://github.com/apache/trafficserver/blob/1a5853908ff9b5251c691a3efcd4765344491427/iocore/net/QUICPacketHandler.cc#L214-L220

The way of registering event type is inconsistent in ATS. This follows what we do for ET_TASK.

### DIFF
--- a/iocore/net/I_UDPNet.h
+++ b/iocore/net/I_UDPNet.h
@@ -44,6 +44,7 @@
 class UDPNetProcessor : public Processor
 {
 public:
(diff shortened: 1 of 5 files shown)


maskit:
This fixes #5506.

SolidWallOfCode (approved):

PR #9315, opened 2023-01-18
Added http connect Autest with proxy verifier

lzx404243:
Added an Autest with proxy verifier to generate traffic that executes the http `CONNECT` request.

### DIFF
--- a/tests/gold_tests/connect/connect.test.py
+++ b/tests/gold_tests/connect/connect.test.py
@@ -18,6 +18,7 @@
 
 from enum import Enum
 import os
+import re
 
 Test.Summary = 'Exercise HTTP CONNECT Method'
 Test.ContinueOnFail = True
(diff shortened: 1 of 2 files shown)


tests/gold_tests/connect/replays/connect.replay.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - client-request:
+          method: GET
+          version: "1.1"
+          url: /get
+          headers:
+            fields:
+              - [uuid, 2]
+              - [Host, www.example.com]
+        server-response:
+          status: 200
```
bneradt, line 53: Let's add some `X-Response` header for this proxied transaction and add a `proxy-response` verification that the client gets that back. The content will be opaque to ATS, but the proxy-verifier server should still function as normal and the client should get its response back.

tests/gold_tests/connect/connect.test.py
```diff
@@ -106,3 +107,61 @@ def run(self):
...
+            description="Verify the server gets the second request.")
+
+    def setupTS(self):
+        self.ts = Test.MakeATSProcess("connect-ts")
+
+        self.ts.Disk.records_config.update({
+            'proxy.config.diags.debug.enabled': 1,
+            'proxy.config.diags.debug.tags': 'http',
+            'proxy.config.http.server_ports': f"{self.ts.Variables.port}",
+            'proxy.config.http.connect_ports': f"{self.server.Variables.http_port}",
```
bneradt, line 140: I don't think these need to be strings. Can you try without making these string literals and see whether that works?

bneradt (changes requested):
Looks good. Thank you for adding this.

tests/gold_tests/connect/connect.test.py
lzx404243: Tried that but that didn't work out. I think they take string type to support specifying a range of ports.

tests/gold_tests/connect/replays/connect.replay.yaml
lzx404243: Good point! Added those

tests/gold_tests/connect/connect.test.py
bneradt: Gotcha. Thanks for trying it out.

bneradt (approved):
Looks good.

randall (approved):

PR #9322, opened 2023-01-19
Add TLSCertSwitchSupport

maskit:
This is preparation for enabling cert switching (or enabling SNI action) on QUIC connections. Like other TLSSomethingSupport, there should be no logic change.

~Code change around QUICNetVC will be made on 10-Dev separately.~ Changes for QUIC is on #9347

### DIFF
--- a/iocore/net/SSLUtils.cc
+++ b/iocore/net/SSLUtils.cc
@@ -261,118 +261,6 @@ ssl_rm_cached_session(SSL_CTX *ctx, SSL_SESSION *sess)
   session_cache->removeSession(sid);
 }
 
-static int
-set_context_cert(SSL *ssl, void *arg)
-{
-  shared_SSL_CTX ctx  = nullptr;
-  SSL_CTX *verify_ctx = nullptr;
-  SSLCertContext *cc  = nullptr;
-  SSLCertificateConfig::scoped_config lookup;
-
-  const char *servername     = SSL_get_servername(ssl, TLSEXT_NAMETYPE_host_name);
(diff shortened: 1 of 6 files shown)


iocore/net/SSLUtils.cc
```diff
@@ -418,17 +306,19 @@ ssl_client_hello_callback(SSL *s, int *al, void *arg)
...
   }
 
-  SSLNetVConnection *netvc = SSLNetVCAccess(s);
-  if (!netvc || netvc->ssl != s) {
-    Debug("ssl.error", "ssl_client_hello_callback call back on stale netvc");
-    return SSL_CLIENT_HELLO_ERROR;
-  }
-
-  bool reenabled = netvc->callHooks(TS_EVENT_SSL_CLIENT_HELLO);
+  SSLNetVConnection *netvc = dynamic_cast<SSLNetVConnection *>(snis);
```
bryancall, line 309: Looking over the PR - I am concerned that we are adding 3 dynamic casts.  Dynamic casts seem to be expensive and have shown up in perf top.

maskit: We talked about the use of dynamic_cast several months ago. I think Alan suggested having `T* NetVConnection::get_service<T>())`. The function would return a pointer for a specified service (SomethingSupport) if the netvc supports it. The downside is that we'd have to maintain the function in addition to having the name in class declaration. And I already stated I'm ok with having the function if dynamic_cast costs much. It just don't have enough priority on anybody's list.

bryancall (approved):

maskit:
We'll continue the discussion about dynamic_cast on #9482 and resolve performance issue on it at once.

PR #9323, opened 2023-01-20
Support config reloading using plugin msg for STEK Share Plugin

duke8253:


plugins/experimental/stek_share/stek_share.cc
```diff
@@ -57,11 +63,41 @@ shutdown_handler(TSCont contp, TSEvent event, void *edata)
...
+    TSDebug(PLUGIN_NAME, "Message to '%s' - %zu bytes of data", msg->tag, msg->data_size);
+    if ((strnlen(msg->tag, 20) == 10) && (strncmp(msg->tag, PLUGIN_NAME, 10) == 0)) { // Message is for us
+      if (msg->data_size) {
+        if (strncmp(reinterpret_cast<char *>(const_cast<void *>(msg->data)), "reload", msg->data_size) == 0) {
+          TSDebug(PLUGIN_NAME, "Reloading configurations...");
+          plugin_threads.terminate();
+          stek_share_server.launcher_.shutdown();
+          if (set_server_info() == 0 && init_raft(nuraft::cs_new<STEKShareSM>()) == 0) {
+            TSDebug(PLUGIN_NAME, "Server ID: %d, Endpoint: %s", stek_share_server.server_id_, stek_share_server.endpoint_.c_str());
+            plugin_threads.reset();
```
SolidWallOfCode, line 80: How confident are you the other threads have actually shut down by this time?

Does this mean if the updated file has an error, that causes a `TSEmergency`?

plugins/experimental/stek_share/stek_share.cc
```diff
@@ -57,11 +63,41 @@ shutdown_handler(TSCont contp, TSEvent event, void *edata)
   return 0;
 }
 
+static int
+message_handler(TSCont contp, TSEvent event, void *edata)
+{
+  if (event == TS_EVENT_LIFECYCLE_MSG) {
+    TSPluginMsg *msg = static_cast<TSPluginMsg *>(edata);
+    TSDebug(PLUGIN_NAME, "Message to '%s' - %zu bytes of data", msg->tag, msg->data_size);
+    if ((strnlen(msg->tag, 20) == 10) && (strncmp(msg->tag, PLUGIN_NAME, 10) == 0)) { // Message is for us
```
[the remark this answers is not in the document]
duke8253: Ahh yes, I'll change that.

plugins/experimental/stek_share/stek_share.cc
SolidWallOfCode: I think the "I shut down my entire CDN because I distributed a bad STEK config" style is not going be popular.

plugins/experimental/stek_share/stek_share.h
```diff
@@ -21,103 +21,139 @@ limitations under the License.
...
 public:
-  STEKShareServer() : server_id_(1), addr_("localhost"), port_(25000), sm_(nullptr), smgr_(nullptr), raft_instance_(nullptr)
+  PluginConfig()
   {
-    last_updated_    = 0;
-    current_log_idx_ = 0;
-
     // Default ASIO thread pool size: 4.
-    asio_thread_pool_size_ = 4;
+    asio_thread_pool_size = 4;
```
SolidWallOfCode, line 37: All of these should be inlined, not done in the constructor.

plugins/experimental/stek_share/stek_share.h
```diff
@@ -21,103 +21,139 @@ limitations under the License.
...
-  {
-    sm_.reset();
-    smgr_.reset();
-    raft_instance_.reset();
+    client_req_timeout = 3000;
   }
 
   // Server ID.
-  int server_id_;
+  int server_id;
```
SolidWallOfCode, line 57: This was previously initialized in the constructor, now it's not initialized at all.

plugins/experimental/stek_share/stek_share.h
```diff
@@ -21,103 +21,139 @@ limitations under the License.
...
+  std::string address;
 
   // Server port.
-  int port_;
+  int port;
 
-  // Endpoint: "<addr>:<port>".
-  std::string endpoint_;
+  // Endpoint: "<address>:<port>".
+  std::string endpoint;
```
SolidWallOfCode, line 66: I would strongly consider `swoc::IPSrv` or `swoc::IPEndPoint` instead.

plugins/experimental/stek_share/stek_share.h
```diff
@@ -21,103 +21,139 @@ limitations under the License.
...
 
-  // State machine.
-  nuraft::ptr<nuraft::state_machine> sm_;
+  size_t asio_thread_pool_size;
 
-  // State manager.
-  nuraft::ptr<nuraft::state_mgr> smgr_;
+  int heart_beat_interval;
```
SolidWallOfCode, line 70: Use `std::chrono` for time.

plugins/experimental/stek_share/stek_share.h
```diff
@@ -21,103 +21,139 @@ limitations under the License.
...
 
-  // Raft server instance.
-  nuraft::ptr<nuraft::raft_server> raft_instance_;
+  int reserved_log_items;
 
-  // List of servers to auto add.
-  std::map<int, std::string> server_list_;
+  int snapshot_distance;
+
+  int client_req_timeout;
```
SolidWallOfCode, line 79: Use `std::chrono` for time.

plugins/experimental/stek_share/stek_share.h
```diff
@@ -21,103 +21,139 @@ limitations under the License.
...
 
   // STEK update interval.
-  int key_update_interval_;
+  int key_update_interval;
```
SolidWallOfCode, line 82: Use `std::chrono` for time.

plugins/experimental/stek_share/stek_share.h
```diff
@@ -21,103 +21,139 @@ limitations under the License.
...
+  std::string server_cert_file;
+  std::string server_key_file;
+  std::string cert_verify_str;
+};
 
-  size_t asio_thread_pool_size_;
+class STEKShareServer
+{
+public:
+  STEKShareServer() : sm_instance(nullptr), smgr_instance(nullptr), raft_instance(nullptr)
```
SolidWallOfCode, line 97: Move these initialization values to inline.

plugins/experimental/stek_share/stek_share.h
```diff
@@ -21,103 +21,139 @@ limitations under the License.
...
+  {
+    last_updated     = 0;
```
SolidWallOfCode, line 99: Move this initialization to inline.

plugins/experimental/stek_share/stek_share.h
```diff
@@ -21,103 +21,139 @@ limitations under the License.
...
+  }
 
-  int election_timeout_lower_bound_;
-  int election_timeout_upper_bound_;
+  void
+  reset()
+  {
+    {
+      std::unique_lock lock(sm_mutex);
+      sm_instance.reset();
```
SolidWallOfCode, line 111: Don't these two lines do exactly the same thing?

plugins/experimental/stek_share/stek_share.h
```diff
@@ -21,103 +21,139 @@ limitations under the License.
...
+    {
+      std::unique_lock lock(raft_mutex);
+      raft_instance.reset();
+      raft_instance = nullptr;
+    }
+  }
 
-  int reserved_log_items_;
+  // Server ID.
+  int server_id;
```
SolidWallOfCode, line 129: Isn't this in `PluginConfig`?

SolidWallOfCode (commented):
The pointer must be read under a read lock (`std::shared_lock`). I don't see that.

plugins/experimental/stek_share/stek_share.h
duke8253: Yeah, it's just easier this way to use it during the server run, just needed bits of the config for printing and other stuff, but I can change that logic a bit.

plugins/experimental/stek_share/stek_share.h
duke8253: `nuraft::raft_params` expects int here

SolidWallOfCode: If you need a raw integer that is the number of milliseconds, use `std::chrono::duration_cast<std::chrono::miliseconds>(req_timeout).count()`. But you can do that later.

SolidWallOfCode (approved):

PR #9326, opened 2023-01-21
ppc64 stack grows down, so it can be re-enabled as a supported arch

jeredfloyd:
ATS 9.2.0 FTBFS on ppc64le due to the architecture not being included on the "stack grows down" list.  This architecture does, in fact, grow down.


randall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9330, opened 2023-01-23
Don't send response body on status 204 No Content

maskit:
It's a protocol violation.

> A 204 response is terminated by the end of the header section; it cannot contain content or trailers.

https://www.rfc-editor.org/rfc/rfc9110#section-15.3.5-5

PR #9333, opened 2023-01-25
Make 204 cacheable again

maskit:
#9330 made 204 uncacheable regardless of negative caching list configuration, because ATS caches empty documents only if the size, which is 0, is explicitly indicated by Content-Length.

Status 204 response does not allowed to have any content, and Content-Length header is not required. We should allow this case too.


ezelkow1 (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9334, opened 2023-01-26
Fix an error on SSL config reload (plus some cleanup).

ywkaras:
Fix an error on SSL config reload (plus some cleanup).
    
    This seems to have eliminated some ERROR diags we were seeing in Yahoo Prod when
    doing config reloads.
    
    The SSLSecret public functions no longer return pointers into the unorded_map of
    secrets, they return a copy of the secret data. This seemed thread unsafe. A
    periodic poll running in the background can update the secret data for an entry
    for a secret name in the map.
    
    To avoid exporting pointers, I had to change the prototype of TSSslSecretGet().
    Hopefully there are no existing plugins that are already using this TS API function,
    so breaking this rule will be moot.

### DIFF
--- a/iocore/net/SSLUtils.cc
+++ b/iocore/net/SSLUtils.cc
@@ -1053,8 +1053,8 @@ SSLPrivateKeyHandler(SSL_CTX *ctx, const SSLConfigParams *params, const char *ke
     scoped_BIO bio(BIO_new_mem_buf(secret_data, secret_data_len));
     pkey = PEM_read_bio_PrivateKey(bio.get(), nullptr, nullptr, nullptr);
     if (nullptr == pkey) {
-      Debug("ssl_load", "failed to load server private key from %s",
-            (!keyPath || keyPath[0] == '\0') ? "[empty key path]" : keyPath);
+      Debug("ssl_load", "failed to load server private key (%.*s) from %s", secret_data_len < 50 ? secret_data_len : 50,
+            secret_data, (!keyPath || keyPath[0] == '\0') ? "[empty key path]" : keyPath);
       return false;
     }
     if (!SSL_CTX_use_PrivateKey(ctx, pkey)) {
@@ -2285,8 +2285,8 @@ SSLMultiCertConfigLoader::load_certs_and_cross_reference_names(
   }
 
   for (size_t i = 0; i < data.cert_names_list.size(); i++) {
-    std::string_view secret_data;
-    std::string_view secret_key_data;
+    std::string secret_data;
+    std::string secret_key_data;
     params->secrets.getOrLoadSecret(data.cert_names_list[i], data.key_list.size() > i ? data.key_list[i] : "", secret_data,
                                     secret_key_data);
     if (secret_data.empty()) {
@@ -2445,8 +2445,8 @@ SSLMultiCertConfigLoader::load_certs(SSL_CTX *ctx, const std::vector<std::string
 
   for (size_t i = 0; i < cert_names_list.size(); i++) {
     std::string keyPath = (i < key_list.size()) ? key_list[i] : "";
-    std::string_view secret_data;
-    std::string_view secret_key_data;
+    std::string secret_data;
+    std::string secret_key_data;
     params->secrets.getOrLoadSecret(cert_names_list[i], keyPath, secret_data, secret_key_data);
     if (secret_data.empty()) {
       SSLError("failed to load certificate secret for %s with key path %s", cert_names_list[i].c_str(),
@@ -2482,6 +2482,7 @@ SSLMultiCertConfigLoader::load_certs(SSL_CTX *ctx, const std::vector<std::string
     }
 
     if (secret_key_data.empty()) {
+      Note("Empty private key for public key %.*s", int(secret_data.size()), secret_data.data());
       secret_key_data = secret_data;
     }
     if (!SSLPrivateKeyHandler(ctx, params, keyPath.c_str(), secret_key_data.data(), secret_key_data.size())) {
--- a/doc/developer-guide/api/functions/TSSslSecret.en.rst
+++ b/doc/developer-guide/api/functions/TSSslSecret.en.rst
@@ -11,6 +11,7 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
    implied.  See the License for the specific language governing
    permissions and limitations under the License.
+
 .. include:: /common.defs
 
 .. default-domain:: c
@@ -27,7 +28,7 @@ Synopsis
 
     #include <ts/ts.h>
 
-.. function:: TSReturnCode TSSslSecretSet(const char * secret_name, int secret_name_length, const char * secret_data, int secret_data_len)
+.. function:: TSReturnCode TSSslSecretSet(const char * secret_name, int secret_name_length, const char * secret_data, int secret_data_length)
 
 Description
 ===========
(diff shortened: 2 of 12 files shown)


ywkaras:
This PR replaces https://github.com/apache/trafficserver/pull/8790 , which was laid waste by zwoop's merciless purge of personal branches in the main repo.

The TSSslSecretGet() function is now compatible with the rest of the TS API (and bell bottoms and stagflation).

iocore/net/SSLUtils.cc
```diff
@@ -1053,8 +1053,8 @@ SSLPrivateKeyHandler(SSL_CTX *ctx, const SSLConfigParams *params, const char *ke
     scoped_BIO bio(BIO_new_mem_buf(secret_data, secret_data_len));
     pkey = PEM_read_bio_PrivateKey(bio.get(), nullptr, nullptr, nullptr);
     if (nullptr == pkey) {
-      Debug("ssl_load", "failed to load server private key from %s",
-            (!keyPath || keyPath[0] == '\0') ? "[empty key path]" : keyPath);
+      SSLError("failed to load server private key (%.*s) from %s", secret_data_len < 50 ? secret_data_len : 50, secret_data,
```
maskit, line 1056: I'm not sure when to use SSLError instead of Debug, but do the battle on another PR.

maskit:
Oh no, we released 9.2.0 without this fix. This means 9.2.0 has a problematic TS API and we can't fix it until 10.0.

maskit:
@zwoop What do you think about releasing 9.2.1 soon with this technically an incompatible change?

iocore/net/SSLUtils.cc
ywkaras: Errors go into diags.log, so it's things that the operations people should see.

maskit: I'm not joining the battle. I'm saying focus on fixing the API issue on this PR.

ywkaras: Yahoo is not currently using TSSslSecretGet().  For us, the main purpose of this change was to fix ERRORs in diags.log that we were seeing on config reload.  The change to the API was just part of the effort to prevent future problems as well as the one we were seeing.  Many problems happen first (or only) in prod, and will go unnoticed if they are only seen in debug trace rather than in diags.log.

maskit: I don't see a reason to do the two things at the same time on one PR.

This line was changed to Debug about two months ago for some reasons. Changing it back to SSLError would need a discussion. I don't want to keep the API problematic until the discussion ends.

ywkaras: Do you have a link to that discussion?  If bad configuration files are not an error, what is?'

This is a cleanup PR, so it does many, many things at once.  If you want me to break it out into a large number of small PRs, I'll have to do that in the future as a low-priority task.  Otherwise, Yahoo would fire me for having very very poor judgement about priorities ha ha.

maskit: The discussion is not started yet. You should start the discussion if you want to change this line back to SSLError. I don't think you should rewrite it as you want without understanding why it was changed to Debug. And the previous change was approved by you. #9225. What was it?

> This is a cleanup PR

I don't understand why you still call this just a clean up. You mean the API is fine as is?

ywkaras: @serrislew why did you change this to a debug trace?

ywkaras: It's a cleanup PR because it includes many relatively small changes.  One was fixing a potential bug if the TSSslSecretGet() API call was used.  The other changes fixed the config reload error Yahoo was seeing, but I can't identify which change or set of changes constituted the fix.

serrislew: It still calls a SSLError log https://github.com/apache/trafficserver/pull/9225/files#diff-cb865c0bc65fb8ef103a206282b78a50e0c1c9e93ca6713322ba9df9d921e6c9L2453

ywkaras: But that's a different line of code.

serrislew: But that additional SSLError log replaces all of the changed SSLError to Debug calls. `SSLPrivateKeyHandler` is only called once and it will reach that additional SSLError when it returns false. And now with the SSLError called after `SSLPrivateKeyHandler` return false, it can print the cert name that wasn't available inside the function

ywkaras: Thanks Serris.

tests/gold_tests/tls/tls_client_cert2_plugin.test.py
```diff
@@ -105,7 +105,7 @@
     '  client_cert: {0}/../signed-bar.pem'.format(ts.Variables.SSLDir),
     '  client_key: {0}/../signed-bar.key'.format(ts.Variables.SSLDir),
     '- fqdn: bob.*.com',
-    '  client_cert: {0}/../combo-signed-foo.pem'.format(ts.Variables.SSLDir),
+    '  client_cert: {0}/combo-signed-foo.pem'.format(ts.Variables.SSLDir),
```
maskit, line 108: This file is in the same directory as other files. Why just this one?

ywkaras: I don't know.  I made that change and the test passed.

ywkaras: It seems that combo-signed-foo.pem is the only cert or key listed in sni.yaml that is actually needed:
```
wkaras ~/REPOS/YTS/tests/_sandbox/tls_client_cert2_plugin/ts
O$ ls
bin  cache  config  log  plugin  runtime  ssl  storage
wkaras ~/REPOS/YTS/tests/_sandbox/tls_client_cert2_plugin/ts
O$ ls ssl
combo-signed-foo.pem  server.pem       signed2-foo.pem  signed-bar.pem  signed-foo.pem
server.key            signed2-bar.pem  signed-bar.key   signed-foo.key
wkaras ~/REPOS/YTS/tests/_sandbox/tls_client_cert2_plugin/ts
O$ cat config/sni.yaml 
sni:
- fqdn: bob.bar.com
  client_cert: /home/wkaras/REPOS/YTS/tests/_sandbox/tls_client_cert2_plugin/ts/ssl/../signed-bar.pem
  client_key: /home/wkaras/REPOS/YTS/tests/_sandbox/tls_client_cert2_plugin/ts/ssl/../signed-bar.key
- fqdn: bob.*.com
  client_cert: /home/wkaras/REPOS/YTS/tests/_sandbox/tls_client_cert2_plugin/ts/ssl/combo-signed-foo.pem
- fqdn: "*bar.com"
  client_cert: /home/wkaras/REPOS/YTS/tests/_sandbox/tls_client_cert2_plugin/ts/ssl/../signed2-bar.pem
  client_key: /home/wkaras/REPOS/YTS/tests/_sandbox/tls_client_cert2_plugin/ts/ssl/../signed-bar.key
- fqdn: "foo.com"
  client_cert: /home/wkaras/REPOS/YTS/tests/_sandbox/tls_client_cert2_plugin/ts/ssl/../signed2-foo.pem
  client_key: /home/wkaras/REPOS/YTS/tests/_sandbox/tls_client_cert2_plugin/ts/ssl/../signed-foo.key
wkaras ~/REPOS/YTS/tests/_sandbox/tls_client_cert2_plugin/ts
O$
```

ywkaras: OK, when I remove my change, the test still works.  I don't remember why I made it.

maskit (approved):

zwoop:
Is this something we want for 9.2.1 ?

ywkaras:
I would recommend backporting, I'll do if you want, in case there are any conflicts.

ywkaras:
It will change the prototype of one TS API function.  But the current prototype for that function (which returns a pointer to data that's modified by multiple threads) is inherently crashy to use (hopefully no one's using it).

maskit:
We talked about it on weekly bug scrub. The API was newly added on 9.2.0. It's now (before many people start using it) or 10.0. Fixing it on 9.2.2 doesn't make sense.

PR #9337, opened 2023-01-30
s3_auth autest: convert from gold file to file contains

traeak:
This makes the autest more resilient against different curl versions by checking just for return status and content length for test pass instead of using a gold file.


randall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9339, opened 2023-01-30
Revert PRs 7667 and 7807

vmamidi:
(cherry picked from commit 1d581cb7b0c1038cb831f715e1f5241c69da0505)


bryancall (approved):
I am ok with reverting these changes for 9.2.1, but we should get a better understanding on what to do for 10.0.0

zwoop (approved):

Issue #9335, opened 2023-01-27
Proxy Protocol strings have dest and src reversed


bneradt:
I noticed that when we configure ATS to send Proxy Protocol via [proxy.config.http.proxy_protocol_out](https://docs.trafficserver.apache.org/admin-guide/files/records.config.en.html#proxy-config-http-proxy-protocol-out), our Proxy Protocol strings have src and dest descriptions reversed:

https://github.com/apache/trafficserver/blob/3dcf2a8a0fae7d1ca2140af60b55ea50b8d2b4ec/proxy/http/HttpSM.cc#L132

We should pass `vc_in->get_remote_endpoint()` (which is the client's src address) before `local` (which is the client's dest address).

I have an internal Yahoo patch I'm working on for this with an AuTest that I plan to opensource soon.

PR #9341, opened 2023-01-31
Proxy protocol out fixes

bneradt:
This fixes the following 2 issues with proxy.config.http.proxy_protocol_out:

1. The Proxy Protocol string had src and dest reversed #9335 
2. Proxy Protocol strings weren't sent with a blind tunnel


masaori335 (approved):
Nice catch! AuTest for outbound proxy protocol seems great!

PR #9336, opened 2023-01-27
Fixed some memory leaks with http/3

bryancall:
These numbers are with using freelists:

Before:
`SUMMARY: AddressSanitizer: 9997094750 byte(s) leaked in 4587157 allocation(s).
`
After:
`SUMMARY: AddressSanitizer: 137185838 byte(s) leaked in 77301 allocation(s).
`

PR #9342, opened 2023-01-31
Fixed memory leaks with http/3

bryancall:
- Leak with not freeing packets
- Leak in buffers for QPACK
- Leak with the http/3 session

These number are with not using the freelist (-f -F):

Before with previous memory leak fixes (#9336):
`SUMMARY: AddressSanitizer: 3214562750 byte(s) leaked in 7288765 allocation(s).`

After:
`SUMMARY: AddressSanitizer: 24670 byte(s) leaked in 429 allocation(s).`

### DIFF
--- a/iocore/net/QUICNet.cc
+++ b/iocore/net/QUICNet.cc
@@ -77,6 +77,7 @@ QUICPollCont::_process_packet(QUICPollEvent *e, NetHandler *nh)
   }
 
   // Note: We should free QUICPollEvent here since vc could be freed from other thread.
+  p->free();
   e->free();
 }
 #else
--- a/iocore/net/QUICNetVConnection.cc
+++ b/iocore/net/QUICNetVConnection.cc
@@ -1649,7 +1649,6 @@ QUICNetVConnection::_packetize_frames(uint8_t *packet_buf, QUICEncryptionLevel l
   max_frame_size = std::min(max_frame_size, this->_maximum_stream_frame_data_size());
 
   bool probing                   = false;
-  int frame_count                = 0;
   size_t len                     = 0;
   Ptr<IOBufferBlock> first_block = make_ptr<IOBufferBlock>(new_IOBufferBlock());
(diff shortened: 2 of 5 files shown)


iocore/net/QUICNet.cc
```diff
@@ -77,6 +77,7 @@ QUICPollCont::_process_packet(QUICPollEvent *e, NetHandler *nh)
   }
 
   // Note: We should free QUICPollEvent here since vc could be freed from other thread.
+  p->free();
```
maskit, line 80: I don't think we can free the `p` here because QUICNetVC doesn't immediately process packets in `handle_receiv ed_packet()`. It should be freed on QUICNetVC side.

maskit: Actually the Quiche version may process packets immediately, but it'd break the original version.

bryancall: `quic_conn_recv()` get called right before this, does that process the packet?

`QUICPollCont::_process_packet() > QUICNetVConnection::handle_received_packet() > quiche_conn_recv()`

maskit: I found #ifdef. This code is just for quiche and no problem.

maskit (dismissed):

bryancall:
[approve ci Rocky]

bryancall:
[approve ci Fedora]

bryancall:
[approve ci]

bryancall:
[approve ci]

bryancall:
[approve ci]

bryancall:
[approve ci autest]

maskit:
Doesn't the second commit leak application instance on tests? What was the problem?

maskit:
Seems like changing base branch confuses GitHub. I see changes already on master.

maskit (approved):
Let's leave the ownership issue for now. The change looks fine. Go ahead and merge when autest passes.

maskit:
[approve ci autest]

PR #9352, opened 2023-02-01
Merge 10 dev into master

bneradt:
This is the trivial (i.e., non-conflicting) merge from 10-Dev into master. Master has been merged frequently into 10-Dev, as recently as this morning. Thus this merge brings master to be the same content as 10-Dev.

@zwoop points out that we probably want to "merge" this rather "squash and merge".


zwoop:
I converted this to a Draft, to make sure we don't merge it, yet.

brbzull0:
Just for the record, this does contains https://github.com/apache/trafficserver/pull/9264 records.yaml (already approved)

zwoop (approved):
I checked every single line, looks good.

PR #9355, opened 2023-02-01
Upgrades master branch to use clang-format v15.0.7

zwoop:
In addition to the upgrade to v15.0.7, with its new features and fixes, this also adds binaries for aarch64 platforms for both Linux and macOS.


zwoop:
[approve ci]

zwoop:
[approve ci]

maskit (commented):
👍 Tested the binaries on aarch64.

iocore/dns/SplitDNS.cc
```diff
@@ -407,7 +407,7 @@ SplitDNSRecord::ProcessDomainSrchList(char *val)
 
   for (int i = 0; i < numTok; i++) {
     const char *current = pTok[i];
-    int cnt             = sz += strlen(current);
+    int cnt = sz += strlen(current);
```
bneradt, line 410: This is an interesting change. Most of the other changes add alignment with the previous `=` or whatever, this removes it. (This isn't a problem, just something that caught my eye.)

bneradt (approved):
I skimmed through about half of these files, up until github was autohiding the files because of the volume. I then ignored white space changes, which highlighted the changes that added or removed newlines.

I didn't see any changes I didn't like. I think these changes are an improvement and will make things more readable.

Thanks for keeping clang-format updated.

bneradt:
And I do definitely like having the arm64 clang-format binaries. That makes running clang-format much quicker on my system.

iocore/dns/SplitDNS.cc
zwoop: Yeh, there are still imperfections :).

bryancall (approved):

PR #9357, opened 2023-02-01
Add dockerfile for building h3 tools for testing

cmcfarlen:


zwoop:
Wonder if it'd nicer to do something like

```
tools/http3/Dockerfile
```

zwoop:
@cmcfarlen remember to set labels, Milsetone etc.

bryancall:
The build_h3_tools.sh script uses tags for each of the tools.

zwoop:
[approve ci autest]

zwoop (changes requested):
Couple of suggestions:

1. As Bryan pointed out, we should "freeze" the versions of each of these. If you do, maybe also update the build_h3_tools.sh script accordingly so we have the same versions consistently?
2. Instead of -j20, maybe something like ```-j$(nproc)```  ?

cmcfarlen:
[approve ci autest]

zwoop (approved):

PR #9363, opened 2023-02-02
Fixes spelling across the tree

randall:


shukitchan:
everything under plugins/wasm/lib/ are actually copied from other project (i.e. proxy-wasm-cpp-host)
May be we should leave them alone?

doc/conf.py
```diff
@@ -486,7 +486,7 @@ def __init__(self, builder, *args, **kwds):
 # The format is a list of tuples containing the path and title.
 #epub_pre_files = []
 
-# HTML files shat should be inserted after the pages created by sphinx.
+# HTML files should should be inserted after the pages created by sphinx.
```
maskit, line 489: should should

brbzull0:
[approve ci autest]

zwoop (approved):

PR #9365, opened 2023-02-02
Fix the autest Proxy Verifier version check

bneradt:
The Proxy Verifier autest version check logic assumed that verifier-client was in the PATH. This is most often not the case, and even if it is it can be the wrong one since the test will by default use the unpacked Proxy Verifier from bintray. This fixes the check so that it explicitly refers to the verifier-client that the test will use.


bneradt:
[approve ci autest]

bneradt:
[approve ci autest]

bneradt:
[approve ci autest]

PR #9369, opened 2023-02-02
fatal.test.py autest: make it more reliable

bneradt:
The fatal.test.py test has been flaky because of a race condition between the process shutdown caused by TSFatal and the autest shutting down the process because the Default Process print statement finished. This patch adds a FileContaines ready condition for the TSFatal log message before ending the TestRun which makes it run correctly more reliably.


bneradt:
[approve ci autest]

bneradt:
On second thought, I'm just going to make one PR that fixes the related test inconsistency issues.

PR #9370, opened 2023-02-02
remap_load_empty_failure.test.py autest: make it more reliable

bneradt:
The remap_load_empty_failure.test.py had a race condition between the autest run finishing due to the Default echo process finishing and the traffic server instance writing the "remap.config failed to load" message to the diags.log. This updates the Ready condition for the traffic server process to wait until that log is printed before running the echo and ending the test.


bneradt:
I'll close this PR and file one that fixes the various test flakiness issues on master.

PR #9377, opened 2023-02-03
Still load legacy config if it exists instead of failing

cmcfarlen:
Everything is broken and I'm not prepared for records.yaml

### DIFF
--- a/tests/gold_tests/records/old_config_fail.test.py
+++ b/tests/gold_tests/records/old_config_fail.test.py
@@ -18,6 +18,8 @@
 
 Test.Summary = 'Test old config ATS startup'
 
+Test.SkipIf(Condition.true("Don't require records.yaml just yet")
+
 ts = Test.MakeATSProcess("ts")
 ts.Disk.MakeConfigFile('records.config').AddLine('dummy file!')
 
--- a/include/tscore/Filenames.h
+++ b/include/tscore/Filenames.h
@@ -25,10 +25,11 @@ namespace ts
 {
 namespace filename
 {
-  constexpr const char *STORAGE = "storage.config";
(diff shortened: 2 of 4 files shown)


DoingTheNeedful:
Thanks, @cmcfarlen - swapping from a format that's been in use for years to yaml necessitates a change of a ton of tooling.  The prior change would have been breaking for just about every org that uses ATS, with a significant investment in trying to change every part of build, validation, and config management.  Even if you believe that swapping to yaml would be better, you cannot expect that to be a minor change.

tests/gold_tests/records/old_config_fail.test.py
```diff
@@ -1,32 +0,0 @@
-'''
```
brbzull0, line 1: Why removing this?? In any case we can just:

```
Test.SkipIf(Condition.true("..."))
```

brbzull0:
We have already talked about this on slack but just for the record:
As discussed(dev and user mailing list) before, we agreed to have single parsing and fail if the legacy file is found. As also requested,  a tool is provided for the translation with a guide on how to move on. 
I have no issues with having it dual but just making sure that we are all on the same page.

If there is anything I can do to make the process easy just let me know and I'll do it.

brbzull0:
I've created an issue which I will be implementing to track a solution for this.

https://github.com/apache/trafficserver/issues/9384

bryancall:
We talked about this in the PR/issue scrub and would like to have a discussion at next weeks PR/issue scrub.  A final decision should go to the mailing list.   Marking it as a draft, so it doesn't get merged in before we discuss.

cmcfarlen:
I'm going to close this and we can work on the yaml support around tooling instead.

Issue #9376, opened 2023-02-03
Error message from convert2yaml is not very user friendly


maskit:
I faced this error when I converted my records.conf to yaml format.
```
Something went wrong: 'int' object does not support item assignment
Traceback (most recent call last):
  File "/Users/mkitajo/src/github.com/trafficserver/tools/records/convert2yaml.py", line 300, in <module>
    func(args)
  File "/Users/mkitajo/src/github.com/trafficserver/tools/records/convert2yaml.py", line 259, in handle_file_input
    add_object(config, name, value[:-1], type)
  File "/Users/mkitajo/src/github.com/trafficserver/tools/records/convert2yaml.py", line 203, in add_object
    add_object(config[key], var[index + 1:], value, type=type)
  File "/Users/mkitajo/src/github.com/trafficserver/tools/records/convert2yaml.py", line 203, in add_object
    add_object(config[key], var[index + 1:], value, type=type)
  File "/Users/mkitajo/src/github.com/trafficserver/tools/records/convert2yaml.py", line 203, in add_object
    add_object(config[key], var[index + 1:], value, type=type)
  [Previous line repeated 1 more time]
  File "/Users/mkitajo/src/github.com/trafficserver/tools/records/convert2yaml.py", line 197, in add_object
    config[var] = get_value(type, value)
TypeError: 'int' object does not support item assignment
```

The problem was that my records.config had both an old setting and new settings for `proxy.config.ssl.client.verify.server`.

Old
```
CONFIG proxy.config.ssl.client.verify.server INT 0
```

New
```
CONFIG proxy.config.ssl.client.verify.server.policy STRING DISABLED
CONFIG proxy.config.ssl.client.verify.server.properties STRING ALL
```

The old one made the convert script think it's an integer setting and then new ones tried to add a child to the integer item. It's not the script's fault, but it would be nice if we could show more user friendly error message.

PR #9378, opened 2023-02-03
Make sure we let the user know we are about to perform an invalid conversion.

brbzull0:
before failing with an ugly message.
Also make the traceback output optional, add (-e) option.

it may with https://github.com/apache/trafficserver/issues/9376


maskit (approved):
Looks good. Thank you.

PR #9380, opened 2023-02-03
Try shasum with algo option if sha256sum is unavailable

maskit:
`sha256sum` is unavailable on macOS but `shasum`.


zwoop (approved):
Cool. My mad had shasum256 but I didn’t check to see what provides it.

PR #9381, opened 2023-02-03
Fix stack overflow on sending UDP packets

maskit:
The check is simply wrong. It has to break out from the loop if `npackets` reaches the limit.

### DIFF
--- a/iocore/net/UnixUDPNet.cc
+++ b/iocore/net/UnixUDPNet.cc
@@ -979,7 +979,7 @@ sendPackets:
     bytesThisPipe       -= pktLen;
     packets[npackets++] = p;
   next_pkt:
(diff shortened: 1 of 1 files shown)


zwoop (approved):

bryancall (approved):

iocore/net/UnixUDPNet.cc
```diff
@@ -979,7 +979,7 @@ UDPQueue::SendPackets()
     bytesThisPipe       -= pktLen;
     packets[npackets++] = p;
   next_pkt:
-    if (bytesThisPipe < 0 && npackets == N_MAX_PACKETS) {
+    if (bytesThisPipe < 0 || npackets == N_MAX_PACKETS) {
```
bryancall, line 982: >= seems safer, but it really shouldn't matter

PR #9382, opened 2023-02-04
Use snprintf

maskit:
Another cleanup for deprecated sprintf. This makes `make` and `make check` happy with my configuration (there are still some sprintf left).

### DIFF
--- a/iocore/cache/test/main.cc
+++ b/iocore/cache/test/main.cc
@@ -116,11 +116,11 @@ build_hdrs(HTTPInfo &info, const char *url, const char *content_type)
 
   REQUIRE(url != nullptr);
 
-  p += snprintf(p, sizeof(buf), "GET %s HTTP/1.1\n", url);
-  p += snprintf(p, sizeof(buf), "User-Agent: curl/7.47.0\n");
-  p += snprintf(p, sizeof(buf), "Accept: %s\n", content_type);
-  p += snprintf(p, sizeof(buf), "Vary: Content-type\n");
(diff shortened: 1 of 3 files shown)


zwoop (approved):

bneradt (approved):

iocore/cache/test/main.cc
```diff
@@ -116,11 +116,11 @@ build_hdrs(HTTPInfo &info, const char *url, const char *content_type)
 
   REQUIRE(url != nullptr);
 
-  p += sprintf(p, "GET %s HTTP/1.1\n", url);
-  p += sprintf(p, "User-Agent: curl/7.47.0\n");
-  p += sprintf(p, "Accept: %s\n", content_type);
-  p += sprintf(p, "Vary: Content-type\n");
-  p += sprintf(p, "Proxy-Connection: Keep-Alive\n\n");
+  p += snprintf(p, sizeof(buf), "GET %s HTTP/1.1\n", url);
+  p += snprintf(p, sizeof(buf), "User-Agent: curl/7.47.0\n");
```
bneradt, line 120: Actually, I suppose we need to decrement the size of the buffer passed to snprintf with each write. Maybe something like:

```
  p += snprintf(p, sizeof(buf) - strnlen(p, sizeof(buf)), "User-Agent: curl/7.47.0\n");
```

bneradt (changes requested):

iocore/cache/test/main.cc
maskit: You're right. I updated the code.

bneradt (approved):
Looks good.

Issue #5974, opened 2019-10-01
Turning on PROXY Protocol forces clients to perform ALPN, otherwise connections are dropped


HT154:
ATS: 8.0.5
OpenSSL: 1.0.2r

```
CONFIG proxy.config.http.server_ports STRING 8080:pp:proto=http 8443:ssl:pp:proto=http
```

Running Nginx on localhost, stream proxying connections on 80/443 to 8080/8443 with PROXY protocol.

Working via cURL:
```bash
$ curl https://localhost -I
HTTP/1.1 200 OK
Server: ATS/8.0.5
Date: Tue, 01 Oct 2019 00:09:30 GMT
Content-Type: text/html
X-Artifactory-Id: 7ec3d2bf97259bfba06cad2cf4ef343cb4370a8f
X-Artifactory-Node-Id: art_mr2_3
Age: 0
Connection: keep-alive
```

Not working via Python `requests`:
```python
import requests
requests.get('https://localhost')
# <traceback snipped>
# requests.exceptions.SSLError: HTTPSConnectionPool(host='localhost', port=443): Max retries exceeded with url: /artifactory/api/pypi/pypi-virtual/simple/crude/ (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate in certificate chain (_ssl.c:1056)')))
```

Working via Python `requests` after forcing ALPN:
```python
from requests import Session
from requests.adapters import HTTPAdapter
from requests.packages.urllib3.util.ssl_ import create_urllib3_context

class HTTP11ALPNAdapter(HTTPAdapter):
    def init_poolmanager(self, *args, **kwargs):
        ctx = create_urllib3_context()
        ctx.set_alpn_protocols(['http/1.1'])
        kwargs['ssl_context'] = ctx
        return super().init_poolmanager(*args, **kwargs)

s = requests.Session()
s.mount('https://localhost', HTTP11ALPNAdapter())
s.get('https://localhost')
# <Response [200]>
```

@masaori335 was able to reproduce this and track down that it was the `pp` in the `server_ports` config line that was triggering this behavior. When `pp` was not specified, clients not performing ALPN worked as expected.

masaori335:
@reveller Could you take a look? It looks like the combination of `:pp` & `:ssl` goes something wrong. I tired some combinations with above settings (`8443:ssl:pp:proto=http`).
[...]

PR #9383, opened 2023-02-05
Proxy Protocol: Allow for no Proxy Protocol header

bneradt:
Prior to this patch, if the user specified that a listening port may have Proxy Protocol headers on connections via `pp` in the server_ports configuration, the connection would be dropped if such a header was not sent. This makes ATS flexible in such situations: if the Proxy Protocol port is there, then it is processed, otherwise the connection is processed according to the other specifications in the server_ports configuration for the port.


masaori335 (approved):
This makes ATS more flexible with `pp` on `server_port`. Also, this fixes #5974.

sfeltner-godaddy (approved):
This looks great.  Thanks for this update, @bneradt!

PR #9385, opened 2023-02-07
Use IOBufferBlock instead of alloc to avoid stackoverflow

maskit:
Use of `alloca` is inappropriate because buffer sizes requested here can be megabytes in total.


bryancall (approved):
Ran some benchmarks with 2M requests and it doesn't have memory problems anymore.

PR #9386, opened 2023-02-07
Enable UDP GSO by default

maskit:
There's no reason not to enable GSO by default. If it's enabled on a system that doesn't support GSO, ATS fallbacks to non GSO mode automatically.


zwoop:
[approve ci autest]

zwoop (approved):

bryancall (approved):

PR #9387, opened 2023-02-07
Update h2disable tests for new curl output

bneradt:
The h2disable and h2disable_no_accept_threads autests both had curl output expectations that have changed in the recent release of curl. Here is the old output:

```
Using HTTP2, server supports multiplexing
```

Here is the new output

```
using HTTP/2
```

This updates the ContainsExpression and ExcludesExpression strings to be able to work with both versions of curl.


randall (approved):

bneradt:
I'm suggesting we port this back to 9.2.x because it is a test only change that will be annoying if anyone updates their curl and runs autest for the older branch.

zwoop:
Cherry-picked to v9.2.x

PR #9388, opened 2023-02-07
Remove mgmt library

cmcfarlen:


SolidWallOfCode (approved):
Thanks for doing this and adding change the include paths so we don't have to have so many include directories.

PR #9401, opened 2023-02-12
Removes the UglyStub file

zwoop:
With all the cleanup, and refactoring etc., there are only a small number of things that needs the ugly stubs. Fixing those allowed us to eliminate the UglyLogStubs library completely.


zwoop:
Well, seems linux builds are different than OSX builds for UDP. Looking into the issues.

zwoop:
Changing this to a draft, while I iterate over fixing all the oddities.

zwoop:
I think with this landed, we may be able to disconnect the dependencies on the ```udpPacketAllocator```, but doing so requires us to move some "inline" functions into proper classes (as static). Can't quite get it to work though, but landing this will at least put us in a better position to clean up this mess with the UDP dependencies.

I also think that resolving this may help fixing "make check" problems.

cmcfarlen:
We could move `new_UDPPacket` and `UDPPacketInternal::free` into `UnixUDPNet.cc` and then remove the extern from the `P_UDPPacket.h` file.  The other `new_` functions could just call `new_UDPPacket` instead of the allocator directly.

cmcfarlen:
I really support this work as it's currently very complicated to add unit tests to much of the code.  Any little bit helps.

zwoop:
> We could move `new_UDPPacket` and `UDPPacketInternal::free` into `UnixUDPNet.cc` and then remove the extern from the `P_UDPPacket.h` file. The other `new_` functions could just call `new_UDPPacket` instead of the allocator directly.

Yeh, I was trying that last night, and yes, I think this is right path, but ran into other problems. I think we should land this first, and then continue from there. I also think fixing all this will likely fix the "make check" issues on macOS.

cmcfarlen (approved):

bneradt (approved):

Issue #9384, opened 2023-02-06
Allow traffic_ctl to modify records.yaml.


brbzull0:
The idea would be to use `traffic_ctl config set` to modify the `records.yaml`.  This will insert or change the existing field in the `records.yaml`.

PR #9404, opened 2023-02-13
records.yaml - Add support to load multiple YAML docs from the same file and let traffic_ctl to modify a records.yaml file

brbzull0:
This includes 2 changes.

1 - Support for loading multiple YAML documents from the same `records.yaml`  file. This is done basically to  keep the same behavior as we had before with `records.config` where the latest config set overwrites a previous set. So basically:

`records.yaml`:
```yaml
---
ts:
  diags:
    debug:
      enabled: 0
---
ts:
  diags:
    debug:
      enabled: 1
```
So, `enabled=1` will be the variable value after loading.

2 - Add support for modifying a `records.yaml` file  using `traffic_ctl`, this will let `traffic_ctl` to:

- Append  a new configuration field in a new YAML doc (same file)
```bash
traffic_ctl config set proxy.config.diags.debug.enabled 1 -c [file]
```
- Modify the latest(top to bottom) YAML doc without appending a new doc
```bash
traffic_ctl config set proxy.config.diags.debug.enabled 1 -u -c [file]
```
- Get a value from a `records.yaml` file
```bash
traffic_ctl config get proxy.config.diags.debug.enabled -c [file]
```

`traffic_ctl` docs are updated to reflect these changes.

https://github.com/apache/trafficserver/issues/9384

### DIFF
--- a/src/traffic_ctl/CtrlCommands.cc
+++ b/src/traffic_ctl/CtrlCommands.cc
@@ -24,11 +24,16 @@
 #include "CtrlCommands.h"
 #include "jsonrpc/CtrlRPCRequests.h"
 #include "jsonrpc/ctrl_yaml_codecs.h"
+#include "utils/yaml_utils.h"
+#include "swoc/TextView.h"
+#include "swoc/BufferWriter.h"
+#include "swoc/bwf_base.h"
+
 namespace
 {
 /// We use yamlcpp as codec implementation.
 using Codec = yamlcpp_json_emitter;
-
+} // namespace
 const std::unordered_map<std::string_view, BasePrinter::Options::OutputFormat> _Fmt_str_to_enum = {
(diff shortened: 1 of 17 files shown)


src/traffic_ctl/CtrlCommands.cc
```diff
@@ -59,8 +64,8 @@ parse_print_opts(ts::Arguments *args)
 {
   return {parse_format(args)};
 }
-} // namespace
-
+// } // namespace
+// using namespace print_utils;
```
cmcfarlen, line 68: remove comment

cmcfarlen (changes requested):
looks good.  just some comment cleanup.

cmcfarlen (approved):

cmcfarlen:
This will help to build tooling around records.yaml.  Thanks so much!

PR #9410, opened 2023-02-14
Fixes the compile to work with LLVM15

zwoop:


zwoop:
Unclear why clang v15 is so finicky here, but one of these sizes is a size_t, so heck, just cast to it and use the appropriate printf for it (%zu).

maskit (approved):
I'm fine with this.

PR #9412, opened 2023-02-15
QUIC: Remove hardcoded quiche set_initial_max_streams and use config values instead.

brbzull0:
Need to figure out how to test this.


maskit:
We might be able to use [h3spec](https://github.com/kazu-yamamoto/h3spec), but it's not updated for long time and I'm not sure if we can rely on it.

It'd be really nice to have tests for this but I'm fine with landing this without tests.

brbzull0:
> We might be able to use [h3spec](https://github.com/kazu-yamamoto/h3spec), but it's not updated for long time and I'm not sure if we can rely on it.
> 
> It'd be really nice to have tests for this but I'm fine with landing this without tests.

We can probably land this and create a  new issue to work on the tests.

maskit (approved):

PR #9413, opened 2023-02-15
Remove UDP_stubs.h

cmcfarlen:
Followup to #9401


doc/developer-guide/api/types/TSSslSession.en.rst
```diff
@@ -34,7 +34,6 @@ Synopsis
 
    .. member:: size_t len
 
-   .. member:: char bytes[TS_SSL_MAX_SSL_SESSION_ID_LENGTH]
```
bneradt, line 37: This seems unrelated. Did you have to do this to make docs happy for some reason? If so, we should make it a separate PR. I'd like to know what introduced the change in behavior...

bneradt (changes requested):

iocore/net/I_UDPPacket.h
```diff
@@ -63,48 +63,9 @@ class UDPPacket
...
-/**
-   Create a new packet to be sent over UDPConnection. This actually
-   copies data from a buffer.
 
-
-   @param to  address of where to send packet
-   @param when ink_hrtime relative to ink_get_hrtime_internal()
-   @param buf if !nullptr, then len bytes copied from buf and made into packet.
-   @param len # of bytes to copy from buf
- */
```
bneradt, line 77: Looks like these docs are being dropped in the diff.

bneradt (changes requested):
Looks great. Thanks for cleaning up the build.

doc/developer-guide/api/types/TSSslSession.en.rst
zwoop: Yeh, this is my bad, I gave him a bad patch. This should not be done.

zwoop:
Curious, does this fix the ```make check``` target on macOS?

maskit:
> does this fix the make check target on macOS?

It works for me even without this change.

doc/developer-guide/api/types/TSSslSession.en.rst
bneradt: Gotcha.

zwoop:
> > does this fix the make check target on macOS?
> 
> It works for me even without this change.

Weird, it fails on my mac and on the CI :/.

cmcfarlen:
I tested on Mac, Linux, and the ci docker image.  I think it's good.  I would like to cleanup more Inline.cc as I don't believe it buys us anything anymore.

bneradt (approved):
Looks good.

bneradt:
[approve ci]

iocore/net/I_UDPPacket.h
zwoop: Yeh, again my fault, and honestly I don't find those useful anyways. But yeh, we can ask Chris to restore them. One thing though, several of those "extern" declarations had no implementation anywhere, so they were there, documented, and not usable.

zwoop (changes requested):
I think we need to address those removals that accidentally snuck in.

cmcfarlen:
[approve ci]

cmcfarlen:
> I think we need to address those removals that accidentally snuck in.

I restored the doc line and the doxygen comments. Was there something else?

zwoop (approved):

PR #9418, opened 2023-02-16
Prevent object_read and cache_lookup_result diverge

maskit:
We've been occasionally seeing crashes caused by accessing invalid (nullptr) `object_read`. We could add null checks here and there but I don't want to repeat crash-then-bandaid process for it.

I think there is a seed that started causing the crashes somewhere and ideally we should find and remove the root cause, but that'd be difficult. This change does not fix the root cause, but should prevent accessing invalid `object_read` where `cache_lookup_result` suggests it's cache-hit and an object should be read and available, by adding accessor functions that restricts setting/getting unreasonable values. It might not be complete, but `object_read` is read at many places that assumes it's cache-hit and returning cache-miss when there's doubt avoids the problematic access.

The key change is in HttpTransaction.h. When ATS accesses `cache_lookup_result`, the setter function detects unreasonable change and try to keep it makes sense, and the getter detects outdated result value and return a safer value (i.e. cache miss).

### DIFF
--- a/iocore/cache/Cache.cc
+++ b/iocore/cache/Cache.cc
@@ -430,10 +430,10 @@ CacheVC::set_data(int /* i ATS_UNUSED */, void * /* data */)
   return true;
 }
 
-void
-CacheVC::get_http_info(CacheHTTPInfo **ainfo)
+CacheHTTPInfo *
+CacheVC::get_http_info()
 {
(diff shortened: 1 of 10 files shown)


maskit:
I haven't done any tests with this actually, but I'd like to gather feedback before I spend too much time on this.

Do we want to go this way?
Any better solution?

PR #9423, opened 2023-02-16
traffic_ctl: Add support to monitor metrics.

brbzull0:
Handles metric monitor in a simply way giving some data out as a summary.

monitor3

PS: I'll have a different PR to handle SIGINT to display the summary(as in `ping`).


cmcfarlen (approved):
Cool!

PR #9430, opened 2023-02-16
On arm64 macOS, do not use pagezero linker flag when using luajit

randall:
This flag isn't supported in the arm64 environment https://developer.apple.com/forums/thread/655950


zwoop (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9431, opened 2023-02-17
Spelling fixes

randall:


shukitchan (approved):

src/tscore/unit_tests/test_ArgParser.cc
```diff
@@ -59,10 +59,10 @@ TEST_CASE("Parsing test", "[parse]")
   REQUIRE(parsed_data.get("init").env() == "env_test2");
   REQUIRE(parsed_data.get("globalx_key") == true);
   REQUIRE(parsed_data.get("globalx_key").env() == "env_test");
-  REQUIRE(parsed_data.get("globaly") == true);
-  REQUIRE(parsed_data.get("globaly").size() == 2);
-  REQUIRE(parsed_data.get("globaly").value() == "default1");
-  REQUIRE(parsed_data.get("globaly").at(1) == "default2");
```
shukitchan, line 65: actually i think this change is not right

src/tscore/unit_tests/test_ArgParser.cc
```diff
@@ -79,10 +79,10 @@ TEST_CASE("Parsing test", "[parse]")
   parsed_data = parser.parse(argv2);
   REQUIRE(parsed_data.get("init") == true);
   REQUIRE(parsed_data.get("subinitopt") == true);
-  REQUIRE(parsed_data.get("globaly") == true);
-  REQUIRE(parsed_data.get("globaly").size() == 2);
-  REQUIRE(parsed_data.get("globaly")[0] == "y1");
-  REQUIRE(parsed_data.get("globaly")[1] == "y2");
```
shukitchan, line 85: i think this change may not be right

plugins/experimental/wasm/Makefile.inc
```diff
@@ -75,14 +75,14 @@ experimental_wasm_wasm_la_SOURCES = \
 if HAS_WASMEDGE
 experimental_wasm_wasm_la_SOURCES += \
   experimental/wasm/lib/src/wasmedge/wasmedge.cc \
-  experimetnal/wasm/lib/src/wasmedge/types.h \
+  experimental/wasm/lib/src/wasmedge/types.h \
```
shukitchan, line 78: thanks!

plugins/experimental/wasm/Makefile.inc
```diff
@@ -75,14 +75,14 @@ experimental_wasm_wasm_la_SOURCES = \
...
   experimental/wasm/lib/include/proxy-wasm/wasmedge.h
 endif
 
 if HAS_WAMR
 experimental_wasm_wasm_la_SOURCES += \
   experimental/wasm/lib/src/wamr/wamr.cc \
-  experimetnal/wasm/lib/src/wamr/types.h \
+  experimental/wasm/lib/src/wamr/types.h \
```
shukitchan, line 85: thanks!

src/tscore/unit_tests/test_ArgParser.cc
shukitchan: probably don't need to do this change?

src/tscore/unit_tests/test_ArgParser.cc
```diff
@@ -79,10 +79,10 @@ TEST_CASE("Parsing test", "[parse]")
...
+  REQUIRE(parsed_data.get("globally") == true);
+  REQUIRE(parsed_data.get("globally").size() == 2);
+  REQUIRE(parsed_data.get("globally")[0] == "y1");
+  REQUIRE(parsed_data.get("globally")[1] == "y2");
```
shukitchan, line 85: I think "globaly" is not a spelling mistake.

shukitchan (changes requested):
I think "globaly" is not a spelling mistake. Please undo that change. Thanks.

src/tscore/unit_tests/test_ArgParser.cc
mlibbey: globally means relating to the whole thing, and is in english dictionaries.
globaly isn't in english dictionaries -- what does it mean? relating to a glob? 
(or is your point that it needs fixing in more spots?)

maskit: It's actually Global Y. There are also globalx and globalz.

shukitchan: Yeah. It is a test for the argument parser. So the test defines a few string like "global x", "global y" and "global z" as tokens for the parser. So it is not a spelling mistake for "globaly". Rather it is not referring to one of the tokens to be found inside the string.

PR #9432, opened 2023-02-17
webp_transform plugin: Use Accept header from client request.

ywkaras:


maskit (dismissed):
TSDbg is unavailable on 9.x and we can't cherry pick this commit to fix the plugin.

ywkaras:
OK changes made.

maskit:
I'm not sure if we should make this change. Accept header on the origin side should be the same as the one on the client side unless something modifies it. If it's not the same, that means something modified it for some reason, and this change ignores it.

ywkaras:
Yahoo origin servers sanitize some responses, which includes not putting the Accept header field in the response.

What's a plausible scenario where the server would communicate something to the client by changing the accept header?

maskit:
I don't see the connection with sanitizing responses. Accept header field in the response? Can you tell me some more detail?

I have no idea in what case one would want to modify Accept header, but if you need this change, something on your ATS modifies it, right?

maskit:
I still don't understand the context.

Your origin server includes Accept header in HTTP response?
How using Accept header from a client helps something in the situation?

In any case, the behavior should be configurable at a minimum because it could affect somebody else's use case. And the default behavior should be current behavior unless there's some justification.

ywkaras:
@bryancall any thougths?  Our relevant Jira is YTSATS-3928.

bryancall:
I agree we should use the client Accept header.

plugins/experimental/webp_transform/ImageTransform.cc
```diff
@@ -183,7 +183,7 @@ class GlobalHookPlugin : public GlobalPlugin
 
     // If we might need to convert check to see if what the browser supports
     if (transaction_convert_to_webp == true || transaction_convert_to_jpeg == true) {
-      std::string accept  = transaction.getServerRequest().getHeaders().values("Accept");
+      std::string accept  = transaction.getClientRequest().getHeaders().values("Accept");
```
ywkaras, line 186: The next few lines of code decide whether and how to transform the body of the response to client to a different content type.  So, it make more sense to look at what content types the client will accept, not what types we told the server ATS would accept.

plugins/experimental/webp_transform/ImageTransform.cc
```diff
@@ -183,7 +183,7 @@ class GlobalHookPlugin : public GlobalPlugin
...
       bool webp_supported = accept.find("image/webp") != std::string::npos;
```
ywkaras, line 187: Seems wrong that we just assume the client will accept "image/jpeg" regardless of what's in the accept header the client sent.

plugins/experimental/webp_transform/ImageTransform.cc
maskit: True, if there is not another transform plugin, which supposedly modified Accept header, between this plugin and the client.

plugins/experimental/webp_transform/ImageTransform.cc
maskit: If something modifies Accept header, the something should be responsible for returning a content that the client accept.

bneradt:
[approve ci cmake]

ezelkow1:
[approve ci centos]

PR #9433, opened 2023-02-17
Ethread::process_event(): make sure event mutex is unlocked before freeing event.

ywkaras:
Also adds important information to docs about TSMutexDestroy().

### DIFF
--- a/iocore/eventsystem/UnixEThread.cc
+++ b/iocore/eventsystem/UnixEThread.cc
@@ -139,6 +139,7 @@ EThread::process_event(Event *e, int calling_code)
     EventQueueExternal.enqueue_local(e);
   } else {
     if (e->cancelled) {
+      MUTEX_RELEASE(lock);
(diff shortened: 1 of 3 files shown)


iocore/eventsystem/UnixEThread.cc
```diff
@@ -161,6 +162,7 @@ EThread::process_event(Event *e, int calling_code)
         EventQueueExternal.enqueue_local(e);
       }
     } else if (!e->in_the_prot_queue && !e->in_the_priority_queue) {
+      MUTEX_RELEASE(lock);
```
masaori335, line 165: It looks like the `lock` is already released in line 154 for this case.

ywkaras: Yes, I'll fix this.

masaori335 (approved):
Looks good.

PR #9439, opened 2023-02-17
Use TSDbg in webp_transform plugin

bneradt:
This updates the logging for the webp_transform plugin to use the TSDbg feature, allowing the more efficient value of 3 for proxy.config.diags.debug.enabled for the associated plugin's debug tag (`webp_transform`).


ywkaras:
Looks good to me.

maskit (approved):

PR #9442, opened 2023-02-17
Replace exclusive locks with rwlocks in hostdb

bneradt:
It was observed that changing to rwlocks significantly improved hostdb performance.

### DIFF
--- a/iocore/hostdb/HostDB.cc
+++ b/iocore/hostdb/HostDB.cc
@@ -222,9 +222,11 @@ hostdb_cont_free(HostDBContinuation *cont)
 {
   if (cont->pending_action) {
     cont->pending_action->cancel();
+    cont->pending_action = nullptr;
   }
   if (cont->timeout) {
     cont->timeout->cancel();
+    cont->timeout = nullptr;
   }
   cont->mutex        = nullptr;
   cont->action.mutex = nullptr;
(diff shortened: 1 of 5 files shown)


shinrich:
Especially for use cases where few names are used.

iocore/hostdb/P_RefCountCache.h
```diff
@@ -169,7 +170,7 @@ template <class C> class RefCountCachePartition
 
   hash_type &get_map();
 
-  Ptr<ProxyMutex> lock; // Lock
+  ink_rwlock lock = PTHREAD_RWLOCK_WRITER_NONRECURSIVE_INITIALIZER_NP;
```
masaori335, line 173: Pretty interesting change. Could you use `std::shared_mutex` here? 
We can do it later, but using `std::shared_mutex` will make it easier to apply `ts::bravo::shared_mutex` or other reader-writer lock improvements in the future, IMO.

bneradt: Maybe I need to since apparently freebsd/osx don't support PTHREAD_RWLOCK_WRITER_NONRECURSIVE_INITIALIZER_NP. :)

iocore/hostdb/HostDB.cc
```diff
@@ -625,29 +640,34 @@ probe(const Ptr<ProxyMutex> &mutex, HostDBHash const &hash, bool ignore_timeout)
   }
 
   // Otherwise HostDB is enabled, so we'll do our thing
-  ink_assert(this_ethread() == hostDB.refcountcache->lock_for_key(hash.hash.fold())->thread_holding);
-  uint64_t folded_hash = hash.hash.fold();
+  uint64_t folded_hash           = hash.hash.fold();
+  std::shared_mutex &bucket_lock = hostDB.refcountcache->lock_for_key(folded_hash);
```
ywkaras, line 644: Why not use ts::shared_mutex to avoid the risk of writer starvation?

bneradt: Switched to ts::shared_mutex.

bneradt:
[approve ci autest]

PR #9443, opened 2023-02-18
cleanup cache inlines to avoid duplicate symbol definitions

cmcfarlen:
Mac builds without debug enabled (-O3) where failing `make check` with duplicate symbol errors on the linker.  This is because the `Inline.cc` file in `iocore/cache` was including files from other libraries.  This PR cleans those up.


zwoop (approved):
This fixes my problems with "make check" on macOS as well.

SolidWallOfCode (approved):

PR #9444, opened 2023-02-18
Fix SSLSessionDup for old OpenSSL and BoringSSL

maskit:
The code added by #8578 had a bug. Confirmed the behavior with autest this time.


bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9445, opened 2023-02-18
Stabilize autest for traffic_dump

maskit:
Messages sent by traffic_ctl are processed periodically and may not be applied immediately.


bneradt (commented):
Make sense, but would one second be enough?

maskit:
I tried one sec too but it wasn't enough, it was better than no pause though.

bneradt:
> I tried one sec too but it wasn't enough, it was better than no pause though.

OK, thanks for trying. Definitely keeping it stable with 2 seconds is better than it failing some. Thanks for making it more reliable.

bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9450, opened 2023-02-21
Stabilize autest for tls_check_cert_selection_reload

maskit:
If the test runs really fast, `touch` on line 95 may not update mtime (it can be done within a second since the creation), and then the file won't be reloaded.


bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9451, opened 2023-02-22
Bumps to latest Werkzeug version

zwoop:
This is sketchy, since our Pipfile seemed to indicate issues. But, we keep getting warned about these security issues in Werkzeug, which won't affect us, but still annoying.

### DIFF
--- a/tests/Pipfile
+++ b/tests/Pipfile
@@ -44,11 +44,7 @@ jsonschema = "*"
 python-jose = "*"
 pyyaml ="*"
 
-# The latest version of Werkzeug (2.1.0) breaks httpbin because it removes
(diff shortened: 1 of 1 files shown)


bneradt (changes requested):
I understand the desire to unpin this, but we really cannot unpin Werkzeug because httpbin requires the older version:
https://github.com/apache/trafficserver/pull/9421#issuecomment-1433561879

You will notice all httpbin AuTests fail without this pin.

Since httpbin has been abandoned, we'll need to replace httpbin with something else in order to alleviate the pin. Proxy Verifier can do some of it, but I'm not sure all of its functionality can easily be put into Proxy Verifier in a short amount of time.

PR #9452, opened 2023-02-22
Eliminates the anonymous struct, which is not standard C++

zwoop:
The old code seemed to have worked, but it's technically not standard C++. And -pedantic would complain on this.


bneradt (approved):
Looks good.

maskit (approved):

PR #9454, opened 2023-02-23
Stabilize autest tls_hook18

maskit:
The test can crash in `ReenableSSL` because it can be called in 2 seconds regardless of the status of VC.


bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9456, opened 2023-02-23
Add clang-format-tests to clang-format target

maskit:
`clang-format-tests` isn't ran when `make clang-format`.

Is it actually intentional by any chance?


bneradt (commented):
Makes sense to add. We'll have to run the format target and commit that with the PR.

bneradt (approved):
Thanks for including this in the formatting.

PR #9464, opened 2023-02-26
Adds a --enable-lto option (Link Time Optimization)

zwoop:
In my tests (not particularly scientific), this is about 10% more RPS (faster), while at the same time over 40% lower latency on first response byte. Yes, this can be accomplished without a new configure option, I just find this more convenient.

I've only tested this with LLVM.

### DIFF
--- a/build/common.m4
+++ b/build/common.m4
@@ -37,6 +37,9 @@ EOF
   if test -n "$CXX"; then
     echo "CXX=\"$CXX\"; export CXX" >> $1
   fi
+  if test -n "$LD"; then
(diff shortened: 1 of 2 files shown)


bneradt (commented):
Looks good.

I assume this doesn't increase build times too much?

zwoop:
I've updated this with some input from smart people (thanks Randall and Craig).

bneradt (dismissed):

bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9465, opened 2023-02-27
Make header prefix matching case-insensitive in authProxy

Ftywan:
An update on #9271, making the matching of headers and prefix param case-insensitive.


zwoop:
Looks good. @ezelkow1 Can you please review?

ezelkow1 (approved):

PR #9469, opened 2023-02-28
Use deprecated OpenSSL APIs for MD5 and SHA256 if available

maskit:
The new API requires heap allocation. We want to stick with the old API that doesn't require it as long as it's available.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1307,6 +1307,8 @@ AC_CHECK_FUNCS([ \
   X509_get0_signature \
   ERR_get_error_all \
   SHA1 \
+  SHA256_Init \
+  MD5_Init \
   SSL_SESSION_dup \
 ])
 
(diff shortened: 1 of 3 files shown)


zwoop:
Do we need this for 9.2.x as well?

maskit:
On 9.2.x, only sha256 use non-deprecated API. We wouldn't be able to simply cherry-pick this change, but the change for 9.2.x should not be too difficult.

zwoop:
Even with jemalloc, this makes master ~15% more RPS and ~15% lower latency, running at close to 1M RPS. Without jemalloc, the improvements is even more. Lesson is: don't malloc :).

bryancall (approved):

maskit:
#9473 is for 9.2.x.

PR #9475, opened 2023-03-01
Replace httpbin with go-httpbin

bneradt:
The Python httpbin is no longer maintained and has dependency issues upon an old Werkzeug from which we would like to unpin. This converts our tests to using go-httpbin instead of the Python httpbin module.


brbzull0 (approved):
Thanks Brian

randall (approved):

PR #9477, opened 2023-03-01
QUIC: Make sure the some of the quic configs get set into the quiche impl.

brbzull0:
Just passing them into the quche lib.

```yaml
ts:
  quic:
    no_activity_timeout_in: 1000
    initial_max_data_in: 60000
    initial_max_stream_data_bidi_local_in: 1000
    initial_max_stream_data_bidi_remote_in: 1000
    initial_max_stream_data_uni_in: 4000
```

values are advertised correctly:
```
I00000018 0x65b8963682593300 cry remote transport_parameters initial_max_stream_data_bidi_local=1000
I00000018 0x65b8963682593300 cry remote transport_parameters initial_max_stream_data_bidi_remote=1000
I00000018 0x65b8963682593300 cry remote transport_parameters initial_max_stream_data_uni=4000
I00000018 0x65b8963682593300 cry remote transport_parameters max_idle_timeout=1000
I00000018 0x65b8963682593300 cry remote transport_parameters initial_max_data=60000
```
curl seems to be getting it:
```
...
* ngtcp2_conn_handle_expiry returned error: ERR_IDLE_CLOSE
* Connection #0 to host 127.0.0.1 left intact
...
```

### DIFF
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -4489,12 +4489,16 @@ removed in the future without prior notice.
 .. ts:cv:: CONFIG proxy.config.quic.no_activity_timeout_in INT 30000
    :reloadable:
 
+   The max idle timeout in milliseconds.
    This value will be advertised as ``idle_timeout`` Transport Parameter.
+   Transport Parameter.
 
 .. ts:cv:: CONFIG proxy.config.quic.no_activity_timeout_out INT 30000
    :reloadable:
 
+   The max idle timeout in milliseconds.
    This value will be advertised as  ``idle_timeout`` Transport Parameter.
+   Transport Parameter.
 
 .. ts:cv:: CONFIG proxy.config.quic.preferred_address_ipv4 STRING ""
    :reloadable:
@@ -4510,6 +4514,9 @@ removed in the future without prior notice.
 
 .. ts:cv:: CONFIG proxy.config.quic.initial_max_data_in INT 65536
    :reloadable:
+
+   Integer value that contains the initial value for the maximum amount of data
+   that can be sent on the connection.
 
    This value will be advertised as ``initial_max_data`` Transport Parameter.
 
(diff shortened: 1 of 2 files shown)


brbzull0:
@maskit  is `ts.quic.no_activity_timeout_in` the right field to be used for the `max_idle_timeout` set by `quiche_config_set_max_idle_timeout`?

brbzull0:
[approve ci autest]

brbzull0:
[approve ci autest]

maskit:
> is ts.quic.no_activity_timeout_in the right field to be used for the max_idle_timeout set by quiche_config_set_max_idle_timeout?

Yes, it is. I'm wondering if the setting should take milliseconds. It's millisecond on the spec but we use seconds for most of ATS settings. `proxy.config.http.accept_no_activity_timeout` for example.

brbzull0:
> > is ts.quic.no_activity_timeout_in the right field to be used for the max_idle_timeout set by quiche_config_set_max_idle_timeout?
> 
> Yes, it is. I'm wondering if the setting should take milliseconds. It's millisecond on the spec but we use seconds for most of ATS settings. `proxy.config.http.accept_no_activity_timeout` for example.

I think we can do something for this, and instead of be tight to a particular unit we can actually support multipliers like we do for [other](https://docs.trafficserver.apache.org/en/latest/admin-guide/files/records.yaml.en.html#values) units like M,K,G 

so eventually we can say:

```yaml
ts:
  quic:
    no_activity_timeout_in: 1s
```

brbzull0:
[approve ci autest]

maskit:
Hmm, I like the idea but I'm not sure if that works for timeouts. It's kinda obvious that we cannot set 500 milli-bytes nor 8.3 bytes as a file/buffer size. Its minimum unit is obviously byte. On the other hand, 125 milliseconds can make sense for some timeout, but users don't know the actual internal precision unless the documentation says it.

I found the setting currently takes millisecond. I guess I used the unit which the spec use without thinking about it much. Let's keep this PR simple and think about the unit specifier at somewhere else.

maskit (approved):

brbzull0:
> Let's keep this PR simple and think about the unit specifier at somewhere else.

sure.
Thanks.

brbzull0:
[approve ci clang-analyzer]

PR #9486, opened 2023-03-03
QUIC: Add support to configure UDP max payload limit. 

brbzull0:
This commit also makes this configuration available for the quiche quic implementation.

### DIFF
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -4667,6 +4667,23 @@ removed in the future without prior notice.
    that |TS| is willing to store. The value MUST be at least 2.
    Transport Parameter.
 
+.. ts:cv:: CONFIG proxy.config.quic.max_recv_udp_payload_size_in INT 65527
+
+   This value will be advertised as ``max_udp_payload_size`` Transport Parameter.
+
+.. ts:cv:: CONFIG proxy.config.quic.max_recv_udp_payload_size_out INT 65527
+
+   This value will be advertised as ``max_udp_payload_size`` Transport Parameter.
+
+.. ts:cv:: CONFIG proxy.config.quic.max_send_udp_payload_size_in INT 65527
+
+   Specified the maximum outgoing UDP payload size.
+
+.. ts:cv:: CONFIG proxy.config.quic.max_send_udp_payload_size_out INT 65527
+
+   Specified the maximum outgoing UDP payload size.
+
+
 UDP Configuration
 =====================
 
@@ -4679,6 +4696,7 @@ UDP Configuration
 
    Enables (``1``) or disables (``0``) UDP GSO. When enabled, |TS| tries to use UDP GSO,
    and disables it automatically if it causes send errors.
+
 
 Plug-in Configuration
 =====================
--- a/iocore/net/QUICNetProcessor_quiche.cc
+++ b/iocore/net/QUICNetProcessor_quiche.cc
@@ -90,8 +90,8 @@ QUICNetProcessor::start(int, size_t stacksize)
   }
 
   quiche_config_set_max_idle_timeout(this->_quiche_config, params->no_activity_timeout_in());
-  quiche_config_set_max_recv_udp_payload_size(this->_quiche_config, 16384);
-  quiche_config_set_max_send_udp_payload_size(this->_quiche_config, 16384);
+  quiche_config_set_max_recv_udp_payload_size(this->_quiche_config, params->get_max_recv_udp_payload_size_in());
+  quiche_config_set_max_send_udp_payload_size(this->_quiche_config, params->get_max_send_udp_payload_size_in());
   quiche_config_set_initial_max_data(this->_quiche_config, params->initial_max_data_in());
   quiche_config_set_initial_max_stream_data_bidi_local(this->_quiche_config, params->initial_max_stream_data_bidi_local_in());
   quiche_config_set_initial_max_stream_data_bidi_remote(this->_quiche_config, params->initial_max_stream_data_bidi_remote_in());
--- a/src/records/RecordsConfig.cc
+++ b/src/records/RecordsConfig.cc
@@ -1439,6 +1439,15 @@ static const RecordElement RecordsConfig[] =
   ,
   {RECT_CONFIG, "proxy.config.quic.disable_active_migration", RECD_INT, "0", RECU_DYNAMIC, RR_NULL, RECC_STR, "[0-1]", RECA_NULL}
   ,
+  {RECT_CONFIG, "proxy.config.quic.max_recv_udp_payload_size_in", RECD_INT, "65527", RECU_DYNAMIC, RR_NULL, RECC_NULL, "^[0-9]+$", RECA_NULL}
+  ,
+  {RECT_CONFIG, "proxy.config.quic.max_recv_udp_payload_size_out", RECD_INT, "65527", RECU_DYNAMIC, RR_NULL, RECC_NULL, "^[0-9]+$", RECA_NULL}
+  ,
+  {RECT_CONFIG, "proxy.config.quic.max_send_udp_payload_size_in", RECD_INT, "65527", RECU_DYNAMIC, RR_NULL, RECC_NULL, "^[0-9]+$", RECA_NULL}
+  ,
+  {RECT_CONFIG, "proxy.config.quic.max_send_udp_payload_size_out", RECD_INT, "65527", RECU_DYNAMIC, RR_NULL, RECC_NULL, "^[0-9]+$", RECA_NULL}
+  ,
+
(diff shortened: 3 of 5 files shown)


brbzull0:
[approve ci autest]

maskit:
I just found that `max_udp_payload_size` is actually a Transport Parameter of QUIC. It may be more appropriate to have the setting under `ts.config.quic`. And if so, we should have ones with `_in` and `_out` in the names to support outgoing QUIC connection in the future.

> max_udp_payload_size (0x03):
The maximum UDP payload size parameter is an integer value that limits the size of UDP payloads that the endpoint is willing to receive. UDP datagrams with payloads larger than this limit are not likely to be processed by the receiver.[¶](https://datatracker.ietf.org/doc/html/rfc9000#section-18.2-4.8.1)

> The default for this parameter is the maximum permitted UDP payload of 65527. Values below 1200 are invalid.[¶](https://datatracker.ietf.org/doc/html/rfc9000#section-18.2-4.10.1)

brbzull0:
> I just found that `max_udp_payload_size` is actually a Transport Parameter of QUIC. It may be more appropriate to have the setting under `ts.config.quic`. And if so, we should have ones with `_in` and `_out` in the names to support outgoing QUIC connection in the future.
> 
> > max_udp_payload_size (0x03):
> > The maximum UDP payload size parameter is an integer value that limits the size of UDP payloads that the endpoint is willing to receive. UDP datagrams with payloads larger than this limit are not likely to be processed by the receiver.[¶](https://datatracker.ietf.org/doc/html/rfc9000#section-18.2-4.8.1)
> 
> > The default for this parameter is the maximum permitted UDP payload of 65527. Values below 1200 are invalid.[¶](https://datatracker.ietf.org/doc/html/rfc9000#section-18.2-4.10.1)

Ok, I saw that on the spcecs,

So,

`max_udp_payload_size_in|out` should be part of the `ts.quic` block, but `max_udp_send_payload_size` is not a transport param but rather a generic UDP, so the later should be part of the `ts.udp` , now as we set this on quiche, do we still need the `in|out` for the `send_payload_size`?

maskit:
Ah, Quiche API uses similar names for send and recv, but only the one for recv is a transport parameter.

We should probably have the both recv and send under `ts.quic` because I think we need `in|out` for the both. `in|out` are for inbound|outbound *connections*, so having it under ts.udp doesn't make sense.

We could also have `ts.udp.send_payload_size` without `in|out`, but that'd be complicated. Either `ts.udp.send_payload_size` or `ts.quic.max_udp_send_payload_size_in|out` would have to override the other.

brbzull0:
grand

brbzull0:
> Ah, Quiche API uses similar names for send and recv, but only the one for recv is a transport parameter.
> 
> We should probably have the both recv and send under `ts.quic` because I think we need `in|out` for the both. `in|out` are for inbound|outbound _connections_, so having it under ts.udp doesn't make sense.
> 
> We could also have `ts.udp.send_payload_size` without `in|out`, but that'd be complicated. Either `ts.udp.send_payload_size` or `ts.quic.max_udp_send_payload_size_in|out` would have to override the other.

grand, will update this PR then.

brbzull0:
[approve ci autest]

doc/admin-guide/files/records.yaml.en.rst
```diff
@@ -4667,6 +4667,23 @@ removed in the future without prior notice.
    that |TS| is willing to store. The value MUST be at least 2.
    Transport Parameter.
 
+.. ts:cv:: CONFIG proxy.config.quic.max_recv_udp_payload_size_in INT 65527
+
+   This value will be advertised as ``max_udp_payload_size`` Transport Parameter.
+
+.. ts:cv:: CONFIG proxy.config.quic.max_recv_udp_payload_size_out INT 65527
+
+   This value will be advertised as ``max_udp_payload_size`` Transport Parameter.
```
ywkaras, line 4676: Two configs with exactly the same description.

doc/admin-guide/files/records.yaml.en.rst
```diff
@@ -4667,6 +4667,23 @@ removed in the future without prior notice.
...
+
+.. ts:cv:: CONFIG proxy.config.quic.max_send_udp_payload_size_in INT 65527
+
+   Specified the maximum outgoing UDP payload size.
+
+.. ts:cv:: CONFIG proxy.config.quic.max_send_udp_payload_size_out INT 65527
+
+   Specified the maximum outgoing UDP payload size.
```
ywkaras, line 4684: Ditto.

iocore/net/QUICNetProcessor_quiche.cc
```diff
@@ -90,8 +90,8 @@ QUICNetProcessor::start(int, size_t stacksize)
   }
 
   quiche_config_set_max_idle_timeout(this->_quiche_config, params->no_activity_timeout_in());
-  quiche_config_set_max_recv_udp_payload_size(this->_quiche_config, 16384);
-  quiche_config_set_max_send_udp_payload_size(this->_quiche_config, 16384);
+  quiche_config_set_max_recv_udp_payload_size(this->_quiche_config, params->get_max_recv_udp_payload_size_in());
+  quiche_config_set_max_send_udp_payload_size(this->_quiche_config, params->get_max_send_udp_payload_size_in());
```
ywkaras, line 94: in or out?

ywkaras: Seems a bit odd that the default value for these config variables is (about) double the previous hard-coded value.

doc/admin-guide/files/records.yaml.en.rst
brbzull0: Yes, they reference to the same transport parameter.  I think this eventually will contain more details, or even a link to the current rfc definition.

iocore/net/QUICNetProcessor_quiche.cc
brbzull0: It seems to be the default for [this](https://www.rfc-editor.org/rfc/rfc9000.html#section-18.2-4.10.1) transport parameter.
> The default for this parameter is the maximum permitted UDP payload of 65527. Values below 1200 are invalid.well, being hardcoded before means nothing different

maskit: The hard-coded value means nothing. It could be 12000 or 32768 if I made the commit on a different day.

maskit: And `_in` is correct for the both here. It's for inbound connections. The `_out` will be used when we support H*3* to  origin.

src/records/RecordsConfig.cc
```diff
@@ -258,6 +258,10 @@ static const RecordElement RecordsConfig[] =
   ,
   {RECT_CONFIG, "proxy.config.udp.enable_gso", RECD_INT, "1", RECU_NULL, RR_NULL, RECC_NULL, nullptr, RECA_NULL}
   ,
+  {RECT_CONFIG, "proxy.config.udp.max_recv_payload_size", RECD_INT, "65527", RECU_DYNAMIC, RR_NULL, RECC_NULL, "^[0-9]+$", RECA_NULL}
```
maskit, line 261: We are going to have `proxy.config.quic.max_recv_udp_payload_size_in|out`, right?

doc/admin-guide/files/records.yaml.en.rst
ywkaras: I suggest adding "for incoming/outgoing" packets.

iocore/net/QUICNetProcessor_quiche.cc
ywkaras: From the names of the the Quiche API functions, it doesn't look like you can set different max packet sizes when the near end is the server versus the client.

src/records/RecordsConfig.cc
brbzull0: yes, my bad, made a mess with names and all that stuff. Will fix.

brbzull0:
Made a mess with my branches, this is missing the actual records. Will fix. Sorry again guys.

iocore/net/QUICNetProcessor_quiche.cc
ywkaras: Yes, that's what I thought.  But I don't see how Quiche is allowing us to configure those values differently depending on whether the near end is the client or the server.  If we can't, we only need two new config variables, not four.

maskit: Oh you mean different packet sizes for inbound connection and outbound connection? We'll need another quiche_config object for outbound connections. That is why we use `_in` settings here. We'll have similar code that use `_out` settings somewhere else in the future.

maskit (approved):
Thanks!

PR #9491, opened 2023-03-03
Adds a new percentage option, cleans up pressure calculations

zwoop:
The new option is --iprep_percentage, and allows us to start rate limiting at any percentage of the max allowed. Before, it would rate limit (with the IP reputation) at 90% always.

Also, this fixes a problem where the IP reputation is always on, I think that's not a good idea. You have to explicitly set the bucket sizes at a minimum.

This needs to go into 9.2.x as well, since the old code doesn't work as-is.

### DIFF
--- a/doc/admin-guide/plugins/rate_limit.en.rst
+++ b/doc/admin-guide/plugins/rate_limit.en.rst
@@ -60,21 +60,21 @@ are available:
    queued transactions we will allow. When this threshold is reached, all
    additional transactions are immediately served with an error message.
 
-   The queue is effectively disabled if this is set to `0`, which implies
+   The queue is effectively disabled if this is set to ``0``, which implies
    that when the transaction limit is reached, we immediately start serving
    error responses.
 
-   The default queue size is `UINT_MAX`, which is essentially unlimited.
+   The default queue size is ``UINT_MAX``, which is essentially unlimited.
 
 .. option:: --error
 
    An optional HTTP status error code, to be used together with the
-   :option:`--queue` option above. The default is `429`.
+   :option:`--queue` option above. The default is ``429``.
 
 .. option:: --retry
 
-   An optional retry-after value, which if set will cause rejected (e.g. `429`)
-   responses to also include a header `Retry-After`.
+   An optional retry-after value, which if set will cause rejected (e.g. ``429``)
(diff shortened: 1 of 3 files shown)


doc/admin-guide/plugins/rate_limit.en.rst
```diff
@@ -161,10 +161,10 @@ The following options are available:
 .. option:: --iprep_buckets
 
    The number of LRU buckets to use for the IP reputation. A good number here
-   is 10, but can be configured. The reason for the different buckets is to
-   account for a pseudo-sorted list of IPs on the frequency seen. Too few buckets
-   will not be enough to keep such a sorting, rendering the algorithm useless. To
-   function in our setup, the number of buckets must be less than ``100``.
+   is `10`, which is the default, but can be configured. The reason for the different
```
bneradt, line 164: Need two ticks around 10. Otherwise this is italicized instead of a code space string.

There's a few examples of this in this patch, but just marking this one.

bneradt (changes requested):
Looks good. Just a doc formatting issue I noticed.

zwoop:
@bneradt I've addressed the docs issues.

doc/admin-guide/plugins/rate_limit.en.rst
```diff
@@ -161,21 +161,28 @@ The following options are available:
...
+   few buckets will not be enough to keep such sorting, rendering the algorithm useless.
+   To function in our setup, the number of buckets must be less than ``100``.
 
 .. option:: --iprep_bucketsize
 
-   This is the size of the largest LRU bucket (the `entry bucket`), `15` is a good
-   value. This is a power of 2, so `15` means the largest LRU can hold `32768` entries.
-   Note that this option must be bigger then the `--iprep_buckets` setting, for the
+   This is the size of the largest LRU bucket (the ``entry bucket``), ``15`` is a good
+   value. This is a power of 2, so ``15``` means the largest LRU can hold ``32768`` entries.
```
bneradt, line 172: Three ticks after the 15.

bneradt (changes requested):
Looks good.

bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9500, opened 2023-03-06
HttpSM: fix no_dns_forward_to_parent for cache stale

traeak:
Fixes the case where a stale asset will result in a DNS lookup on the remap.config destination with this option enabled.


cmcfarlen (approved):

cmcfarlen:
[approve ci cmake]

bneradt:
[approve ci cmake]

zwoop:
Cherry-picked to v9.2.x

PR #9502, opened 2023-03-07
Cleanup: More const-crrectness of cache read

masaori335:
I noticed `Vol::open_read` and similar reading functions can be const.


bneradt (approved):

bneradt:
[approve ci autest]

PR #9504, opened 2023-03-07
records.yaml - Fix record name comparison logic.

brbzull0:
There was a bug when you were trying to get a record by name from the registered records. Without this fix you could request  for instance a shorter record name and the comparison will still be valid.

So you could have configured:
```yaml
ts:
  diags:
    debug:
      enable: 1 # note the missing "d" at the end. 
```
and still be valid, even though the registered record is: `proxy.config.diags.debug.enabled`


maskit (approved):

shukitchan:
I think you mean "d" in the above use case. right ?

shukitchan (approved):

brbzull0:
> I think you mean "d" in the above use case. right ?

yes, that's what I meant, thanks.

PR #9508, opened 2023-03-07
9.2.x: http2_flow_control.test.py: Correctly enable debug logs

bneradt:
The http2_flow_control.test.py AuTest was ported from 10.x which supports a value of 3 for `proxy.config.diags.debug.enabled`. 9.2.x, however, does not support 3 and therefore debug logging hasn't been working for this test in 9.2.x. This fixes that. This, incidentally, seems to make the test run more reliably in the opensource CI environment.


bneradt:
Note that this change is a 9.2.x only change. It is not applicable to 10.x since 10.x supports a value of 3 for `proxy.config.diags.debug.enabled`.

bneradt:
[approve ci autest]

PR #9511, opened 2023-03-08
Fix records events deps

cmcfarlen:
This PR moves the code depending on iocore/eventsystem into eventsystem to avoid circular dependencies.  I'm open to better naming for files and functions or a better method for encapsulating the event system code, but this does remove the circular deps.

### DIFF
--- a/src/traffic_server/traffic_server.cc
+++ b/src/traffic_server/traffic_server.cc
@@ -77,7 +77,7 @@ extern "C" int plock(int);
 #include "tscore/I_Layout.h"
 #include "I_Machine.h"
 #include "records/I_RecordsConfig.h"
-#include "records/I_RecProcess.h"
+#include "RecProcess.h"
 #include "Transform.h"
 #include "ConfigProcessor.h"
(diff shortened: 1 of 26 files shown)


brbzull0 (dismissed):
just a minor typo.

brbzull0 (approved):
Thanks.

src/traffic_server/traffic_server.cc
```diff
@@ -77,7 +77,7 @@ extern "C" int plock(int);
 #include "tscore/I_Layout.h"
 #include "I_Machine.h"
 #include "records/I_RecordsConfig.h"
-#include "records/I_RecProcess.h"
+#include "RecProcess.h"
```
bryancall, line 80: Since ` RecProcess.h` is an header file used by other subsystems, shouldn't it be in the` include/` directory?

cmcfarlen: Maybe, but there isn't an include/iocore directory and everything that depends on this already has iocore/eventsystem in their include path.

PR #9518, opened 2023-03-14
Fix, add metrics testing to tls_tunnel Au test.

ywkaras:


ywkaras:
Keeping this a draft because I'm not sure the values of the metrics are correct.

ywkaras:
[approve ci docs]

proxy/http/HttpTransact.cc
```diff
@@ -5723,9 +5723,13 @@ HttpTransact::initialize_state_variables_from_request(State *s, HTTPHdr *obsolet
     s->cache_info.action = CACHE_DO_NO_ACTION;
   }
 
+  bool do_increment_stat = (0 == s->method);
+
   s->method = incoming_request->method_get_wksidx();
 
-  if (s->method == HTTP_WKSIDX_GET) {
+  if (!do_increment_stat) {
```
maskit, line 5730: Why adding test requires this change? Does this actually fix something?

ywkaras: Yes , it fixes the associated issue.

maskit: Name the PR accordingly. Who expects behavior change from the current title.

proxy/http/HttpTransact.cc
```diff
@@ -5723,9 +5723,13 @@ HttpTransact::initialize_state_variables_from_request(State *s, HTTPHdr *obsolet
     s->cache_info.action = CACHE_DO_NO_ACTION;
   }
 
+  bool do_increment_stat = (0 == s->method);
```
shinrich, line 5726: So without this check unset methods count towards tunnel.   Won't this remove some legitimate tunnel counts like tr_pass tunnels?  In what cases are we seeing unset methods?

shinrich:
Seems reasonable enough.  But a description of what it means for the initialize method to be called with method set to 0 would be useful.  Otherwise, I'm unsure about what is being fixed here.

proxy/http/HttpTransact.cc
ywkaras: OK done.

ywkaras:
> Seems reasonable enough. But a description of what it means for the initialize method to be called with method set to 0 would be useful. Otherwise, I'm unsure about what is being fixed here.

I've added a new commit to address this.

ywkaras:
@shinrich ^^^

ezelkow1:
[approve ci centos]

SolidWallOfCode (approved):

PR #9520, opened 2023-03-14
prefetch: add --two-tier option with header to deprecate --front

traeak:
This adds yet another header to pass along which allows the plugin to detect if its in front or back mode.  This is important when something like edge peering is in play.


bneradt:
[approve ci cmake]

traeak:
ended up not being necessary.

PR #9528, opened 2023-03-16
Remove premature context updates in TSSslSecretSet().

ywkaras:
This PR partially reverts PR #8368.  SSL contexts (SSL_CTX) should only be updated by TSSslSecretUpdate() (which is only called for the main, not the related secret).  The update in TSSslSecretSet() is not merely redundant, it causes errors.  When an update is done, OpenSSL will check that the cert public key matches the private key.  Since TSSslSecretSet() can only set one at time, if it also updates, they will not match.

Contexts for a loading confguration never have to be updated, becasue no SSL_CTX's have been created using it yet.


ywkaras:
Waiting on a confirmation for a Yahoo prod engineer that this fixes the problem they're having with one of our plugins.

ywkaras:
Leaving this as a draft, until the TSSslSecretXxx functions are fixed ( see https://github.com/apache/trafficserver/issues/9562 ).

bneradt:
[approve ci cmake]

ezelkow1:
[approve ci centos]

PR #9547, opened 2023-03-21
Add TS_HAS_QUICHE feature variable.

brbzull0:
Some of the quic features may only work with the quiche implementation. Having this will make easy in particular for unit tests to opt out depending of the variable value.

Exmaple:
```python
Test.SkipUnless(
   Condition.HasATSFeature('TS_HAS_QUICHE')
)
```
This also makes the variable available for all binaries.

```bash
/opt/ats/bin/traffic_layout info --features | grep QUICHE
#define TS_HAS_QUICHE 1
```


brbzull0:
[approve ci autest]

maskit:
`HAVE_QUICHE_H` is used at some places. We might want to consistently use `TS_HAS_QUICHE` if we add it.

brbzull0:
> `HAVE_QUICHE_H` is used at some places. We might want to consistently use `TS_HAS_QUICHE` if we add it.

you mean change all the `HAVE_QUICHE_H` usage and use `TS_HAS_QUICHE` instead?

maskit:
> you mean change all the HAVE_QUICHE_H usage and use TS_HAS_QUICHE instead?

Yes, I'm fine with either one but using the both may cause confusion.

brbzull0:
[approve ci rocky]

brbzull0:
convert to draft, I'm not happy with this, will re-open once I'm satisfied.

maskit (approved):
Thanks!

PR #9551, opened 2023-03-22
records to yaml convert script update.

brbzull0:
The script will generate a backup(`records.yaml.{timestamp}`) file if the output file already exists; this option lets you avoid the creation of the backup file.

By default it will behave as before(generate a backup file) unless the new argument is set (`-x|--no-backup`).


cmcfarlen (approved):
Thanks!

bneradt (commented):
Looks good.

PR #9552, opened 2023-03-22
Add tr-allow-plain

shinrich:
This is the implementation of what I suggested on the ats-dev list "[api] Add tr-try-plain port descriptor".  With this code change if a port has tr-allow-plain and ssl, a connection that does not look like TLS will get converted to a UnixNetVConnection and try to process as non-TLS HTTP.

I have an implementation running over 9.1.3 using the tr-pass port descriptor. The conversion of the SSLNetVConnection to UnixNetVConnection works.

I need to write an autest for this. I was going to first write a transparent mode test, but I guess this feature strictly speaking does not need transparent mode although that is likely the case where this would be most likely used.

### DIFF
--- a/iocore/net/SSLNetVConnection.cc
+++ b/iocore/net/SSLNetVConnection.cc
@@ -1333,12 +1333,11 @@ SSLNetVConnection::sslServerHandShakeEvent(int &err)
     err = errno;
     SSLVCDebug(this, "SSL handshake error: %s (%d), errno=%d", SSLErrorName(ssl_error), ssl_error, err);
 
-    // start a blind tunnel if tr-pass is set and data does not look like ClientHello
+    // If this doesn't look like a ClientHello, convert this connection to a UnixNetVC and send the
+    // packet for Http Processing
     char *buf = handShakeBuffer ? handShakeBuffer->buf() : nullptr;
-    if (getTransparentPassThrough() && buf && *buf != SSL_OP_HANDSHAKE) {
-      SSLVCDebug(this, "Data does not look like SSL handshake, starting blind tunnel");
-      this->attributes   = HttpProxyPort::TRANSPORT_BLIND_TUNNEL;
-      sslHandshakeStatus = SSL_HANDSHAKE_ONGOING;
+    if (getTransparentAllowPlain() && buf && *buf != SSL_OP_HANDSHAKE) {
+      this->_migrateFromSSL();
       return EVENT_CONT;
     }
   }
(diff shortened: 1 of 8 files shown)


iocore/net/SSLNetVConnection.cc
```diff
@@ -2457,3 +2456,68 @@ SSLNetVConnection::_ssl_read_buffer(void *buf, int64_t nbytes, int64_t &nread)
...
+  // vio.read buffer gets processed
+  target->readSignalDone(VC_EVENT_READ_COMPLETE, get_NetHandler(t));
+}
+
+/*
+ * Replaces the current SSLNetVConnection with a UnixNetVConnection
+ * Propagates any data in the SSL handShakeBuffer to be processed
+ * by the UnixNetVConnection logic
+ */
+UnixNetVConnection *
```
ywkaras, line 2485: This does not seem to need a return value.

proxy/ProtocolProbeSessionAccept.cc
```diff
@@ -170,8 +170,14 @@ ProtocolProbeSessionAccept::mainEvent(int event, void *data)
...
-    NetVConnection *netvc          = static_cast<NetVConnection *>(data);
-    ProtocolProbeTrampoline *probe = new ProtocolProbeTrampoline(this, netvc->mutex, nullptr, nullptr);
+    NetVConnection *netvc = static_cast<NetVConnection *>(data);
+    ProtocolProbeTrampoline *probe;
+    UnixNetVConnection *unix_netvc = dynamic_cast<UnixNetVConnection *>(netvc);
+    if (unix_netvc != nullptr && unix_netvc->read.vio.get_writer() != nullptr) {
+      probe = new ProtocolProbeTrampoline(this, netvc->mutex, unix_netvc->read.vio.get_writer(), unix_netvc->read.vio.get_reader());
+    } else {
+      probe = new ProtocolProbeTrampoline(this, netvc->mutex, nullptr, nullptr);
+    }
```
ywkaras, line 180: `probe` looks like it was leaked, even before your change.

proxy/http/HttpProxyServerMain.cc
```diff
@@ -227,7 +227,7 @@ MakeHttpProxyAcceptor(HttpProxyAcceptor &acceptor, HttpProxyPort &port, unsigned
   ProtocolSessionCreateMap.insert({TS_ALPN_PROTOCOL_INDEX_HTTP_1_1, create_h1_server_session});
 
   if (port.isSSL()) {
-    SSLNextProtocolAccept *ssl = new SSLNextProtocolAccept(probe, port.m_transparent_passthrough);
+    SSLNextProtocolAccept *ssl = new SSLNextProtocolAccept(probe, port.m_transparent_passthrough, port.m_transparent_allow_plain);
```
ywkaras, line 230: So the new object takes ownership of `probe`?

shinrich: @SolidWallOfCode should be able to talk to the memory ownership of the probe logic.

shinrich:
Will reopen a new PR.  I ended up building my local version which this attribute disconnected from the transparent feature.

PR #9554, opened 2023-03-23
Make magick plugin buildable with BoringSSL

maskit:
BoringSSL's `EVP_PKEY_assign_RSA` doesn't accept `char *`.

I don't see any reason to have the template.


bneradt (approved):
Thanks. With this I can set up CI boringssl builds.

zwoop:
/
Cherry-picked to v9.2.x

PR #9558, opened 2023-03-29
Require use of BoringSSL if Quiche is used

maskit:
#9347 will require this. We can remove this commit if Quiche supports OpenSSL in the future.


bneradt:
[approve ci cmake]

bneradt (approved):

PR #9561, opened 2023-03-29
Drop support for old quiche

maskit:
This just removes `HAVE_QUICHE_CONFIG_SET_ACTIVE_CONNECTION_ID_LIMIT` which is to support old versions of Quiche.


brbzull0 (commented):
Shouldn't we remove this one too?https://github.com/apache/trafficserver/blob/de8e2439fdd4dc491bb7af512b8f8603a81a7be1/build/quiche.m4#L73

maskit:
Good catch. I updated quiche.m4. Thanks.

bneradt (approved):

brbzull0 (approved):
looks good. Thanks.

PR #9565, opened 2023-03-30
Cleanup: Remove lib/perl code.

brbzull0:
This makes no more sense in the current status as we have moved from records.config to records.yaml.
If configuration needs to be set from within a script there are other options right now:
1 - JSONRPC API, any scripting language would do it.
2 - traffic_ctl config set --cold should write the file only.
3 - Use a YAML library and work directly with the file.


brbzull0:
[approve ci cmake]

cmcfarlen (approved):

PR #9566, opened 2023-03-30
Cleanup: Use swoc::meta instead of ts::meta.

brbzull0:


src/traffic_ctl/CtrlPrinters.cc
```diff
@@ -24,7 +24,7 @@
 #include <string_view>
 
 #include "jsonrpc/ctrl_yaml_codecs.h"
-#include "tscpp/util/ts_meta.h"
+#include "swoc/swoc_meta.h"
```
randall, line 27: should this be `<swoc/swoc_meta.h>` like the other includes in this file?

randall (approved):

PR #9567, opened 2023-03-30
Add timestamp to SOH

ezelkow1:
Add `current_time_epoch_ms` stat to be appended before the server version. This allows computation of stats externally based on the cache time frame. This can help alleviate issues with sliding windows between various stats programs that generate discrepencies


bneradt (approved):

anonymous-bugs:
Can the docs be updated that this stats variable exists. It seems to be missing about this addition

PR #9571, opened 2023-03-31
fixup cmake build for master and add conditional for io_uring support

cmcfarlen:


bneradt (approved):
Builds for me in my local rockylinux:8 container.

I haven't used cmake before, but it seems great. It configures super fast and the resultant make scripts build faster than autotools make scripts.

Thanks for working on this.

PR #9572, opened 2023-03-31
Add logic to make the server.policy and server.properties settings reloadable

### LINKED ISSUE #9563: proxy.config.ssl.client.verify.server.policy and proxy.config.ssl.client.verify.server.properties are not reloadable
Although the docs and traffic_ctl say they should be.  I have seen this in 9.1.3, and fixed it.  A quick glance through the branches seems that is it not reloadable in newer versions either.  Hope to get an afternoon to add an autest and fix this for master.

shinrich:
Registered callback and added autest.

This closes #9563


brbzull0:
[approve ci cmake]

iocore/net/SSLConfig.cc
```diff
@@ -189,6 +189,65 @@ set_paths_helper(const char *path, const char *filename, char **final_path, char
...
+  if (strcmp(verify_server, "SIGNATURE") == 0) {
+    verifyServerProperties = YamlSNIConfig::Property::SIGNATURE_MASK;
+  } else if (strcmp(verify_server, "NAME") == 0) {
+    verifyServerProperties = YamlSNIConfig::Property::NAME_MASK;
+  } else if (strcmp(verify_server, "ALL") == 0) {
+    verifyServerProperties = YamlSNIConfig::Property::ALL_MASK;
+  } else if (strcmp(verify_server, "NONE") == 0) {
+    verifyServerProperties = YamlSNIConfig::Property::NONE;
+  } else {
+    Warning("%s is invalid for proxy.config.ssl.client.verify.server.properties.  Should be one of SIGNATURE, NAME, or ALL",
```
brbzull0, line 229: I know this was already in the code but(not big issue) I would probably mention that `NONE` will be set in this case as per the docs  `ALL` seems the default, it could be misleading and someone may think `ALL` will be used in this case.

iocore/net/SSLConfig.cc
```diff
@@ -189,6 +189,65 @@ set_paths_helper(const char *path, const char *filename, char **final_path, char
...
+SSLConfigParams::SetServerPolicy(const char *verify_server)
+{
+  if (strcmp(verify_server, "DISABLED") == 0) {
+    verifyServerPolicy = YamlSNIConfig::Policy::DISABLED;
+  } else if (strcmp(verify_server, "PERMISSIVE") == 0) {
+    verifyServerPolicy = YamlSNIConfig::Policy::PERMISSIVE;
+  } else if (strcmp(verify_server, "ENFORCED") == 0) {
+    verifyServerPolicy = YamlSNIConfig::Policy::ENFORCED;
+  } else {
+    Warning("%s is invalid for proxy.config.ssl.client.verify.server.policy.  Should be one of DISABLED, PERMISSIVE, or ENFORCED",
```
brbzull0, line 245: Same here, I would mention that this is now `DISABLED`

brbzull0 (approved):
Looks good, just a minitor comment that can be fixed in another PR probably, no biggie.

shinrich:
@brbzull0 good points. I pushed up another commit to address.

brbzull0 (approved):
Thanks!

zwoop:
@shinrich @bneradt Do we want this back ported to 9.2.x ?

shinrich:
Seems like a bug fix that would be good to backport.  I'm not relying on this fix however.  So getting it in 10.x would be good enough for me.  Seems low risk in any case.

zwoop:
Cherry-picked to v9.2.x

PR #9588, opened 2023-04-07
swoc/TextView.h: add cstdint include

bneradt:
This fixes clang-16 builds by adding cstdint for intmax_t and related
definitions. This has a corresponding libswoc PR:

https://github.com/SolidWallOfCode/libswoc/pull/84

### DIFF
--- a/lib/swoc/include/swoc/TextView.h
+++ b/lib/swoc/include/swoc/TextView.h
@@ -13,6 +13,7 @@
 
 #pragma once
 #include <bitset>
+#include <cstdint>
 #include <iosfwd>
 #include <memory.h>
 #include <string>


maskit:
I thought we don't touch external library snapshots like this. Do we?

bneradt:
> I thought we don't touch external library snapshots like this. Do we?

I have a libswoc PR with this patch here: https://github.com/SolidWallOfCode/libswoc/pull/84 which Alan has merged in.

But I'd like to get a quick cherry-pick with that in locally so that I can enable fedora:38 builds in CI.

bneradt:
This patch got included in https://github.com/apache/trafficserver/pull/9592

PR #9589, opened 2023-04-07
Converting typedef aliases to using aliases

bneradt:
using aliases are pretty universally preferred over typedefs. Modern
editors complain about them which is distracting. This is a pass through
all typedefs, changing the non-c uses of them to using aliases.

### DIFF
--- a/include/tscore/Extendible.h
+++ b/include/tscore/Extendible.h
@@ -98,9 +98,9 @@ public:
 // C API
 
 // context (internally FieldDesc*)
-typedef void const *ExtFieldContext;
-typedef void *DerivedPtr;
-typedef void *FieldPtr;
+using ExtFieldContext = const void *;
+using DerivedPtr      = void *;
+using FieldPtr        = void *;
 
 FieldPtr ExtFieldPtr(DerivedPtr derived, ExtFieldContext field_context, int *size = nullptr);
 
--- a/plugins/esi/fetcher/HttpDataFetcherImpl.h
+++ b/plugins/esi/fetcher/HttpDataFetcherImpl.h
@@ -105,7 +105,7 @@ private:
   TSCont _contp;
   char _debug_tag[64];
 
-  typedef std::list<FetchedDataProcessor *> CallbackObjectList;
+  using CallbackObjectList = std::list<FetchedDataProcessor *>;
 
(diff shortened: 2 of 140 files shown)


bneradt:
[approve ci fedora]

include/tscore/Extendible.h
```diff
@@ -98,9 +98,9 @@ template <typename T> class has_super_type
 // C API
 
 // context (internally FieldDesc*)
-typedef void const *ExtFieldContext;
-typedef void *DerivedPtr;
-typedef void *FieldPtr;
+using ExtFieldContext = const void *;
```
masaori335, line 101: Note: `void const` is equivalent to `const void`

plugins/esi/fetcher/HttpDataFetcherImpl.h
```diff
@@ -122,10 +122,10 @@ class HttpDataFetcherImpl : public HttpDataFetcher
     RequestData() {}
   };
 
-  typedef __gnu_cxx::hash_map<std::string, RequestData, EsiLib::StringHasher> UrlToContentMap;
+  using UrlToContentMap = __gnu_cxx::hash_map<std::string, RequestData, EsiLib::StringHasher>;
```
masaori335, line 125: Note: https://github.com/apache/trafficserver/issues/9595

include/tscore/Extendible.h
bneradt: True. And I generally prefer `something const` over `const somethin`, but this is what VSCode auto fixed to. Most of this PR is an application of autofix from VS Code. If you want I'll change it, but the the more the autofix did over me manually doing it, the safer I feel. :)

masaori335 (approved):
Great job! 

I'm surprised we still have so many `typedef`, although we have the `modernize-use-using` rule of the clang-tidy since 2019.

https://github.com/apache/trafficserver/commit/191d794253637559644d53fbe192cd30dfb5c3d1

include/tscore/Extendible.h
masaori335: I'm +1 with staying auto-fixed code.

I was saying to myself that this change is good because I am confused sometimes with the combination of const and pointer :)

PR #9590, opened 2023-04-07
Adds a new hash_url directive to strategies

zwoop:
In addition, this deprecates the old hash_key: cache_key directive. Not only is it replaces with this feature, the existing implementation is confusing because it's not the cache-key that's hashed on. It's the URL as set with TSHttpTxnParentSelectionUrlSet().

### DIFF
--- a/doc/admin-guide/files/strategies.yaml.en.rst
+++ b/doc/admin-guide/files/strategies.yaml.en.rst
@@ -182,8 +182,17 @@ Each **strategy** in the list may using the following parameters:
    #. **path**: (**default**) Creates a hash over the path portion of the request URL.
    #. **path+query**: Same as **path** but adds the **query string** in the request URL.
    #. **path+fragment**: Same as **path** but adds the fragment portion of the URL.
-   #. **cache_key**: Uses the hash key from the **cachekey** plugin.  defaults to **path** if the **cachekey** plugin is not configured on the **remap**.
+   #. **cache_key**: Deprecated, use the URL designation instead.
    #. **url**: Creates a hash from the entire request url.
+
+- **hash_url**: The specific URL to use for the **hash_key**. If the specified URL type does
+  not have a value (e.g. not set by an API), we default back to the reqeust URL. Use one of:
+
+   #. **request**: (**default**) Use the client request URL.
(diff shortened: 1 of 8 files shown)


zwoop:
[approve ci-clang-analyzer]

jrushford (commented):
@zwoop, I don't see any unit tests for hash_url.  See proxy/http/remap/unit-tests/  the yaml file there are setup with differing strategies such as one using cache_key and the tests check it.  Otherwise this looks good.

doc/admin-guide/files/strategies.yaml.en.rst
```diff
@@ -182,9 +182,18 @@ Each **strategy** in the list may using the following parameters:
    #. **path**: (**default**) Creates a hash over the path portion of the request URL.
    #. **path+query**: Same as **path** but adds the **query string** in the request URL.
    #. **path+fragment**: Same as **path** but adds the fragment portion of the URL.
-   #. **cache_key**: Uses the hash key from the **cachekey** plugin.  defaults to **path** if the **cachekey** plugin is not configured on the **remap**.
+   #. **cache_key**: Deprecated, use the URL designation instead.
```
ezelkow1, line 185: Since we are deprecating/changing how some of this works it would be good if there was a deprecation note on what settings to use so that this can act how it used to pre-changes. Just helps with transition

zwoop:
> @zwoop, I don't see any unit tests for hash_url. See proxy/http/remap/unit-tests/ the yaml file there are setup with differing strategies such as one using cache_key and the tests check it. Otherwise this looks good.

I modified a few of the autests to use the new directive. Hopefully it still passes :).

doc/admin-guide/files/strategies.yaml.en.rst
zwoop: Added.

jrushford:
> > @zwoop, I don't see any unit tests for hash_url. See proxy/http/remap/unit-tests/ the yaml file there are setup with differing strategies such as one using cache_key and the tests check it. Otherwise this looks good.
> 
> 
> 
> I modified a few of the autests to use the new directive. Hopefully it still passes :).

Ok

jrushford (approved):
Looks good

PR #9591, opened 2023-04-07
Use FetchSM for OCSP HTTP requests

maskit:
This is a step to support OCSP, without boringocsp library that has never been open sourced.

ATS has been using `OCSP_REQ_CTX` and related functions, which make HTTP requests for OCSP, but many of them were deprecated on OpenSSL 3.0 and they are unavailable on BoringSSL. On this PR, I replaced those functions with ATS's FetchSM and functions that are available on BoringSSL.

It still doesn't work with BoringSSL because there are still other functions only available on OpenSSL. I'm going to replace them on next PR.

Changes on FetchSM are needed because OCSP stuff basically runs on ET_OCSP thread and the requests can be made before FetchSM gets ready. It should not affect existing plugins because those won't run until everything is initialized.

### DIFF
--- a/iocore/net/OCSPStapling.cc
+++ b/iocore/net/OCSPStapling.cc
@@ -186,7 +186,7 @@ public:
   get_response_body(int *len)
   {
     SCOPED_MUTEX_LOCK(lock, mutex, this_ethread());
-    char *buf = static_cast<char *>(malloc(MAX_RESP_LEN));
+    char *buf = static_cast<char *>(ats_malloc(MAX_RESP_LEN));
     *len      = this->_fsm->ext_read_data(buf, MAX_RESP_LEN);
     return reinterpret_cast<unsigned char *>(buf);
   }
@@ -562,12 +562,12 @@ query_responder(const char *uri, const char *user_agent, OCSP_REQUEST *req, int 
   if (httpreq.is_success()) {
     // Parse the response
     int len;
-    const unsigned char *res = httpreq.get_response_body(&len);
-    const unsigned char *p   = res;
-    resp                     = reinterpret_cast<OCSP_RESPONSE *>(ASN1_item_d2i(nullptr, &p, len, ASN1_ITEM_rptr(OCSP_RESPONSE)));
+    unsigned char *res     = httpreq.get_response_body(&len);
+    const unsigned char *p = res;
+    resp                   = reinterpret_cast<OCSP_RESPONSE *>(ASN1_item_d2i(nullptr, &p, len, ASN1_ITEM_rptr(OCSP_RESPONSE)));
 
     if (resp) {
-      free(res);
+      ats_free(res);
       return resp;
     }
 
@@ -577,7 +577,7 @@ query_responder(const char *uri, const char *user_agent, OCSP_REQUEST *req, int 
       Error("failed to parse a response from OCSP server; uri=%s len=%d data=%02x%02x%02x%02x%02x...", uri, len, res[0], res[1],
             res[2], res[3], res[4]);
     }
-    free(res);
+    ats_free(res);
     return nullptr;
   }
 
--- a/iocore/net/OCSPStapling.cc
+++ b/iocore/net/OCSPStapling.cc
@@ -540,7 +540,10 @@ query_responder(const char *uri, const char *user_agent, OCSP_REQUEST *req, int 
 
   // Content-Type, Content-Length, Request Body
   if (use_post) {
-    httpreq.set_body("application/ocsp-request", ASN1_ITEM_rptr(OCSP_REQUEST), (const ASN1_VALUE *)req);
+    if (httpreq.set_body("application/ocsp-request", ASN1_ITEM_rptr(OCSP_REQUEST), (const ASN1_VALUE *)req) != 1) {
+      Error("failed to make a request for OCSP server; uri=%s", uri);
+      return nullptr;
+    }
   }
 
   // Send request
--- a/iocore/net/OCSPStapling.cc
+++ b/iocore/net/OCSPStapling.cc
@@ -562,12 +562,21 @@ query_responder(const char *uri, const char *user_agent, OCSP_REQUEST *req, int 
   if (httpreq.is_success()) {
     // Parse the response
     int len;
-    const unsigned char *p = httpreq.get_response_body(&len);
-    resp                   = reinterpret_cast<OCSP_RESPONSE *>(ASN1_item_d2i(nullptr, &p, len, ASN1_ITEM_rptr(OCSP_RESPONSE)));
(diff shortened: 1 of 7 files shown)


bneradt:
[approve ci fedora]

iocore/net/OCSPStapling.cc
```diff
@@ -373,84 +511,72 @@ stapling_check_response(certinfo *cinf, OCSP_RESPONSE *rsp)
...
-    return resp;
+  // User-Agent header
+  if (user_agent != nullptr) {
+    httpreq.add_header("User-Agent", user_agent);
   }
 
-  Error("failed to connect to OCSP server; host=%s path=%s", host, path);
+  // Content-Type, Content-Length, Request Body
+  if (use_post) {
+    httpreq.set_body("application/ocsp-request", ASN1_ITEM_rptr(OCSP_REQUEST), (const ASN1_VALUE *)req);
```
ywkaras, line 543: You are ignoring the return value of this function.  Should set_body() return void, and call set_error() for errors?

maskit: Good catch. I added error handling.

iocore/net/OCSPStapling.cc
```diff
@@ -57,6 +59,142 @@ struct certinfo {
   time_t expire_time;
 };
 
+extern ClassAllocator<FetchSM> FetchSMAllocator;
+
+class HTTPRequest : public Continuation
```
ywkaras, line 64: This looks like it should be in an anonymous namespace.

iocore/net/OCSPStapling.cc
```diff
@@ -57,6 +59,142 @@ struct certinfo {
...
+  }
+
+  void
+  set_request_line(bool use_post, const char *uri)
+  {
+    struct sockaddr_in sin;
+    memset(&sin, 0, sizeof(sin));
+    sin.sin_family      = AF_INET;
+    sin.sin_addr.s_addr = inet_addr("127.0.0.1");
+    sin.sin_port        = 65535;
```
ywkaras, line 106: What process is listening on this port?

ywkaras:
Would it be an option to use libcurl instead?

iocore/net/OCSPStapling.cc
```diff
@@ -373,84 +511,75 @@ stapling_check_response(certinfo *cinf, OCSP_RESPONSE *rsp)
...
-  if (BIO_do_connect(cbio) <= 0 && !BIO_should_retry(cbio)) {
-    Debug("ssl_ocsp", "process_responder: failed to connect to OCSP server; host=%s port=%s path=%s", host, port, path);
-    goto end;
-  }
-  resp = query_responder(cbio, host, path, user_agent, req, req_timeout);
+  if (httpreq.is_success()) {
+    // Parse the response
+    int len;
+    const unsigned char *p = httpreq.get_response_body(&len);
+    resp                   = reinterpret_cast<OCSP_RESPONSE *>(ASN1_item_d2i(nullptr, &p, len, ASN1_ITEM_rptr(OCSP_RESPONSE)));
```
ywkaras, line 561: Seems like you need to free `p` here.

iocore/net/OCSPStapling.cc
maskit: Nothing. I don't think the address and port are actually used for something.

iocore/net/OCSPStapling.cc
```diff
@@ -373,84 +511,75 @@ stapling_check_response(certinfo *cinf, OCSP_RESPONSE *rsp)
...
 
-end:
-  if (cbio) {
-    BIO_free_all(cbio);
+    if (resp) {
+      return resp;
+    }
   }
-  return resp;
```
ywkaras, line 433: Probably also be good to put the value of `p` in the trace output for bad responses.

iocore/net/OCSPStapling.cc
maskit: added anonymous namespace.

iocore/net/OCSPStapling.cc
maskit: Now error log has first 5 bytes of a response if ATS fails to parse it.

iocore/net/OCSPStapling.cc
```diff
@@ -373,84 +516,84 @@ stapling_check_response(certinfo *cinf, OCSP_RESPONSE *rsp)
...
+    const unsigned char *res = httpreq.get_response_body(&len);
+    const unsigned char *p   = res;
+    resp                     = reinterpret_cast<OCSP_RESPONSE *>(ASN1_item_d2i(nullptr, &p, len, ASN1_ITEM_rptr(OCSP_RESPONSE)));
 
-end:
-  if (cbio) {
-    BIO_free_all(cbio);
+    if (resp) {
+      return resp;
+    }
```
ywkaras, line 571: It seems to me that the array of char that is returned by `httpreq.get_response_body() is leaked.  I think you need to change this code to:
```
    // Parse the response
    int len;
    const unsigned char *res = httpreq.get_response_body(&len);
    const unsigned char *p   = res;
    resp                     = reinterpret_cast<OCSP_RESPONSE *>(ASN1_item_d2i(nullptr, &p, len, ASN1_ITEM_rptr(OCSP_RESPONSE)));
    free(res);

    if (resp) {
      if ((res + len) == p) {
        return resp;
      }
      Error("garbage at the end of response from OCSP server; uri=%s len=%d", uri, len);
      return nullptr;
    }
```

maskit: Fixed the memory leak.

Why do we need to check `(res + len) == p` ?

ywkaras: To check for garbage at the end of the response.

ywkaras:
How did you test this?  Can you add a regression test?

iocore/net/OCSPStapling.cc
maskit: OpenSSL's implementation (OSSL_HTTP_REQ_CTX_nbio_d2i) doesn't seem to check it. The check could be too strict and might find unexpected incompatibility. I don't want to have it on this PR.

maskit:
The test I did is completely manual. Setup a test web server that serves ocsp_response.der used in tls_ocsp autest, setup ATS with certs for the autest, then launch ATS. I checked ssl_ocsp debug log.

It would be great if we could have an autest for this, but I don't think we can have it because the URL for OCSP responder (http://localhost:8888) is written in the cert. I guess autest can open a port with a fixed number but I don't think that's something we want to do.

Also, I'm testing the code on several of our prod servers. Although the memory leak exist, everything looks fine so far.

maskit:
[approve ci]

ywkaras (approved):

PR #9593, opened 2023-04-09
OpenSuSE fix - became unable to find intmax_t, uintmax_t

AD5GB:
Not sure what changed since none of my other OSs had the problem.  Suddenly, OpenSuSE (Tumbleweed) started griping about being unable to find intmax_t and uintmax_t with G++-13. This solved that problem and doesn't interfere with Fedora, CentOs or Raspberry Pi OS.  Up to y'all if you want to accept and commit this patch.  It's simple enough I can add it by hand if necessary when I'm rebuilding ATS.

### DIFF
--- a/lib/swoc/include/swoc/TextView.h
+++ b/lib/swoc/include/swoc/TextView.h
@@ -19,6 +19,8 @@
 #include <string>
 #include <string_view>
 #include <limits>
+#include <cstdint>
+#include <cinttypes>
 
 #include "swoc/swoc_version.h"
 #include "swoc/string_view_util.h"


bneradt (changes requested):
Thank you Randy. Can you please retest on latest master? I think Alan's latest libswoc merge from Saturday should have already fixed this issue you're seeing.

AD5GB:
I've  gone ahead and closed the PR.

bneradt:
Sounds good @AD5GB. Thank you for helping keep the compilers happy.

PR #9597, opened 2023-04-11
Cmake plugins and install things

cmcfarlen:


CMakeLists.txt
```diff
@@ -183,3 +190,6 @@ add_subdirectory(mgmt/utils)
 add_subdirectory(mgmt/config)
 add_subdirectory(mgmt/rpc)
 add_subdirectory(src/traffic_server)
+add_subdirectory(plugins)
+
+add_subdirectory(configs)
```
bneradt, line 195: https://stackoverflow.com/a/44704969/629530

bneradt (changes requested):

plugins/cache_promote/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+#  the License.
+#
+#######################
+
+add_atsplugin(cache_promote
+        cache_promote.cc
+        configs.cc
+        policy.cc
+        lru_policy.cc
+        policy_manager.cc
```
bneradt, line 23: Out of curiosity: how do header files relate to this? Will CMake detect changes in header files and rebuild the required files?

bneradt:
This is great progress. Thanks @cmcfarlen for working on this.

CMakeLists.txt
cmcfarlen: I had this for vim, but been using clion lately.  I'll look up the equivalent.

ywkaras:
How does CMake handle installing headers for externally-built plugings?  As done here in Makefile.am:  https://github.com/apache/trafficserver/blob/master/include/tscpp/api/Makefile.am .

plugins/cache_promote/CMakeLists.txt
cmcfarlen: Yes! I did an experiment where I just changed a random header file.  It recompiled .cc files that included that, recreated libraries and then rebuild other targets that depended on that target (even transitive ones)

bneradt (approved):

PR #9598, opened 2023-04-11
TS API - Add support for checking the YAML-CPP version used in the core.

brbzull0:
This will be handy for several reasons:

1 - Avoid binary compatibility issues: Make sure that any plugin using a TS API with a `TSYaml` as part of their signature runs the same YAML-CPP version as in core.
2 - if any JSONRPC Plugin handler wants to make sure they have the same yaml version as in core(before registering by TSRPCRegister, or even after using the constants) so they can avoid issues.


ezelkow1:
[approve ci centos]

PR #9600, opened 2023-04-11
Replace __gnu_cxx::hash_map with std::unordered_map

### LINKED ISSUE #9595: Replace __gnu_cxx::hash_map in the ESI plugin
The ESI plugin is using `__gnu_cxx::hash_map`. Probably we can use `std::unordered_map`.
https://github.com/apache/trafficserver/blob/7651796f643ae17e70c23115d46063e145d97289/plugins/esi/fetcher/HttpDataFetcherImpl.h#L125

https://github.com/apache/trafficserver/blob/7651796f643ae17e70c23115d46063e145d97289/plugins/esi/lib/StringHash.h#L39

lzx404243:
This resolves #9595.

### DIFF
--- a/plugins/esi/Makefile.inc
+++ b/plugins/esi/Makefile.inc
@@ -21,10 +21,7 @@ ESI_CPPFLAGS = \
 	-I$(srcdir)/esi/test \
 	$(AM_CPPFLAGS)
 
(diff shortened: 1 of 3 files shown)


masaori335 (approved):
+1, thanks for addressing the issue.

PR #9603, opened 2023-04-12
records.yaml - Remove duplicate code.

brbzull0:
Just a clean up around the YAML tag definitions used in several places, better have them in a single place and also available for plugins.


masaori335 (approved):
Looks reasonable.

PR #9604, opened 2023-04-12
Make a couple of the threads configs correct

zwoop:


bneradt (approved):
I double checked each of these. `proxy.config.udp.threads` and `proxy.config.cache.threads_per_disk` are clearly only read on process startup. `proxy.config.log.preproc_threads` is more complicated because it is read every `Log::periodic_tasks`, but I think it's main use is utilized only during startup as well.

PR #9610, opened 2023-04-14
traffic_ctl: plugin msg command, print out the response from server.

brbzull0:


brbzull0:
> I read through the code. It seems the only time the response isn't empty is if there was a parsing error. I'd be cool if there was a way to get data out of a plugin!

The server can respond with an error like:
```
$ traffic_ctl plugin msg plugin_xyz dump -f rpc
{"jsonrpc": "2.0", "error": {"code": 10, "message": "Unauthorized action", "data": [{"code": 2, "message": "Denied privileged API access for uid=105106 gid=100"}]}, "id": "5a30ae5f-03f4-4d44-ba33-2f51e8a77f82"}
```

```
Server Error found:
[10] Unauthorized action
- [2] Denied privileged API access for uid=1001 gid=1001
```

> I'd be cool if there was a way to get data out of a plugin!

You can do this already. A JSON-RPC handler can be located inside a plugin and respond with data to the outside.

PR #9615, opened 2023-04-15
Ensure a reason phrase when sending an HTTP/1 response

bneradt:
With an HTTP/1 client and an HTTP/2 origin, responses will come from the
origin without reason phrases because reason phrases are explicitly
removed from the HTTP/2 RFC. When expanding test coverage of reason
phrases it was noticed that when converting responses from HTTP/2 to
HTTP/1 to send toward the client, the reason phrases were empty. This
updates the conversion logic to ensure that HTTP/1 clients receive a
reason phrase.

### DIFF
--- a/proxy/http/HttpTransact.cc
+++ b/proxy/http/HttpTransact.cc
@@ -7834,7 +7834,19 @@ HttpTransact::build_response(State *s, HTTPHdr *base_response, HTTPHdr *outgoing
                              HTTPStatus status_code, const char *reason_phrase)
 {
   if (reason_phrase == nullptr) {
-    reason_phrase = http_hdr_reason_lookup(status_code);
+    if (status_code != HTTP_STATUS_NONE) {
+      reason_phrase = http_hdr_reason_lookup(status_code);
+      Debug("http_transact", "Using reason phrase from status %d: %s", status_code, reason_phrase);
+    } else if (base_response != nullptr && base_response->status_get() != HTTP_STATUS_NONE) {
+      HTTPStatus const base_response_status = base_response->status_get();
+      reason_phrase                         = http_hdr_reason_lookup(base_response_status);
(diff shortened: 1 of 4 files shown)


proxy/http/HttpTransact.cc
```diff
@@ -7834,7 +7834,13 @@ HttpTransact::build_response(State *s, HTTPHdr *base_response, HTTPHdr *outgoing
                              HTTPStatus status_code, const char *reason_phrase)
 {
   if (reason_phrase == nullptr) {
-    reason_phrase = http_hdr_reason_lookup(status_code);
+    if (status_code != HTTP_STATUS_NONE) {
+      Debug("http_transact", "Using reason phrase from status %d: %s", status_code, reason_phrase);
```
maskit, line 7838: reason_phrase is always nullptr here.

proxy/http/HttpTransact.cc
```diff
@@ -7834,7 +7834,13 @@ HttpTransact::build_response(State *s, HTTPHdr *base_response, HTTPHdr *outgoing
                              HTTPStatus status_code, const char *reason_phrase)
 {
   if (reason_phrase == nullptr) {
-    reason_phrase = http_hdr_reason_lookup(status_code);
+    if (status_code != HTTP_STATUS_NONE) {
+      reason_phrase = http_hdr_reason_lookup(status_code);
+      Debug("http_transact", "Using reason phrase from status %d: %s", status_code, reason_phrase);
+    } else if (base_response->status_get() != HTTP_STATUS_NONE) {
```
maskit, line 7840: We may want to check the availability of `base_response`, although it doesn't make sense if it has `nullptr` on this line.

bneradt: I agree. We're best off making sure it isn't nullptr to be safe.

maskit (approved):

PR #9620, opened 2023-04-18
Return current parent max retry count for parent proxy

### LINKED ISSUE #9576: Regression tests fail with -R 3
Our CI jobs currently run the regression tests with `-R 1` and the tests are pretty stable with that option. However, we have noticed that they fail due to a , at least in CI, if we pass `-R 3`. We should get to the bottom of why it's failing with `-R 3` and configure at least some of the regression tests with `-R 3` since that's a superset of `-R 1`.

## Reproduction Steps

```
docker run -it -u 1200:1200 --name regression_failure_repro ci.trafficserver.apache.org/ats/rockylinux:8 /bin/bash

# Now, from within that container.

cd /var/tmp
git clone https://github.com/apache/trafficserver.git
cd trafficserver/
source /opt/rh/gcc-toolset-11/enable
autoreconf -fiv
./configure --with-openssl=/opt/openssl-quic --prefix=/tmp/ats/ --enable-werror --enable-debug
make -j `nproc`
make install
/tmp/ats/bin/traffic_server -K -k -R 3
```

### Expected Results
The tests should run and pass

### What Happens

It fails with a crash:

```
REGRESSION TEST SDK_API_HttpParentProxySet_Fail started
Regression test(SDK_API_HttpParentProxySet_Fail) still in progress
RPRINT SDK_API_HttpParentProxySet_Fail: setting synserver parent proxy to 127.0.0.1:3300
Fatal: HttpTransact.h:611: failed assertion `_v <= configured_connect_attempts_max_retries`
[Mar 31 21:39:22.254] traffic_crashlo NOTE: crashlog started, target=34868, debug=false syslog=true, uid=1200 euid=1200
[Mar 31 21:39:22.254] traffic_crashlo NOTE: logging to 0x13170c0
[Mar 31 21:39:22.318] traffic_crashlo ERROR: wrote crash log to /tmp/ats/var/log/trafficserver/crash-2023-03-31-213922.log
traffic_server - STACK TRACE: 
/tmp/ats/bin/traffic_server(_Z19crash_logger_invokeiP9siginfo_tPv+0xc3)[0x95bc55]
/lib64/libpthread.so.0(+0x12ce0)[0x7f8ad7acace0]
/lib64/libc.so.6(gsignal+0x10f)[0x7f8ad6c0ea4f]
/lib64/libc.so.6(abort+0x127)[0x7f8ad6be1db5]
/tmp/ats//lib/libtscore.so.10(_Z11ink_warningPKcz+0x0)[0x7f8ad9150423]
/tmp/ats//lib/libtscore.so.10(_Z17ats_base64_encodePKhmPcmPm+0x0)[0x7f8ad914b01e]
/tmp/ats/bin/traffic_server(_ZN12HttpTransact12_CurrentInfo8Attempts9incrementEl+0x4a)[0xab1154]
/tmp/ats/bin/traffic_server(_ZN12HttpTransact27handle_response_from_parentEPNS_5StateE+0xc86)[0xa8ea52]
/tmp/ats/bin/traffic_server(_ZN12HttpTransact14HandleResponseEPNS_5StateE+0xb5b)[0xa8d551]
/tmp/ats/bin/traffic_server(_ZN6HttpSM32call_transact_and_set_next_stateEPFvPN12HttpTransact5StateEE+0x78)[0xa4fa86]
/tmp/ats/bin/traffic_server(_ZN6HttpSM17handle_api_returnEv+0x177)[0xa2ab87]
/tmp/ats/bin/traffic_server(_ZN6HttpSM17state_api_calloutEiPv+0xe70)[0xa2a82a]
/tmp/ats/bin/traffic_server(_ZN6HttpSM18state_api_callbackEiPv+0x289)[0xa298fd]
/tmp/ats/bin/traffic_server(TSHttpTxnReenable+0x109)[0x992957]
/tmp/ats/bin/traffic_server[0x9e22d1]
/tmp/ats/bin/traffic_server(_ZN15INKContInternal12handle_eventEiPv+0x236)[0x984b22]
/tmp/ats/bin/traffic_server(_ZN12Continuation11handleEventEiPv+0xdd)[0x962053]
/tmp/ats/bin/traffic_server(_ZNK7APIHook6invokeEiPv+0xeb)[0x98579d]
/tmp/ats/bin/traffic_server(_ZN6HttpSM17state_api_calloutEiPv+0xa0b)[0xa2a3c5]
/tmp/ats/bin/traffic_server(_ZN6HttpSM23do_api_callout_internalEv+0x21d)[0xa4515d]
/tmp/ats/bin/traffic_server(_ZN6HttpSM14do_api_calloutEv+0x27)[0xa59d23]
/tmp/ats/bin/traffic_server(_ZN6HttpSM33state_read_server_response_headerEiPv+0xd2a)[0xa2dca0]
/tmp/ats/bin/traffic_server(_ZN6HttpSM12main_handlerEiPv+0x36b)[0xa32c09]
/tmp/ats/bin/traffic_server(_ZN12Continuation11handleEventEiPv+0xdd)[0x962053]
/tmp/ats/bin/traffic_server[0xd84370]
/tmp/ats/bin/traffic_server[0xd84d4e]
/tmp/ats/bin/traffic_server[0xd85407]
/tmp/ats/bin/traffic_server(_ZN18UnixNetVConnection11net_read_ioEP10NetHandlerP7EThread+0x2b)[0xd87b0f]
/tmp/ats/bin/traffic_server(_ZN10NetHandler18process_ready_listEv+0x9b)[0xd781c7]
/tmp/ats/bin/traffic_server(_ZN10NetHandler15waitForActivityEl+0x4e0)[0xd788da]
/tmp/ats/bin/traffic_server(_ZN7EThread15execute_regularEv+0x462)[0xedd578]
/tmp/ats/bin/traffic_server(_ZN7EThread7executeEv+0x107)[0xedd71f]
/tmp/ats/bin/traffic_server[0xedc002]
/lib64/libpthread.so.0(+0x81cf)[0x7f8ad7ac01cf]
/lib64/libc.so.6(clone+0x43)[0x7f8ad6bf9d83]
Aborted
```

lzx404243:
As titled. This prevents the `SDK_API_HttpParentProxySet_Fail` regression test from crashing. 

This resolves #9576. (`SDK_API_HttpParentProxySet_Fail` is the only test that's crashing/failing)

### DIFF
--- a/proxy/http/HttpTransact.cc
+++ b/proxy/http/HttpTransact.cc
@@ -3641,12 +3641,12 @@ HttpTransact::handle_response_from_parent(State *s)
       return;
     }
 
-    if (s->current.retry_attempts.get() < s->txn_conf->parent_connect_attempts) {
+    if (s->current.retry_attempts.get() < (s->txn_conf->parent_connect_attempts - 1)) {
       HTTP_INCREMENT_DYN_STAT(http_total_parent_retries_stat);
-      s->current.retry_attempts.increment(s->configured_connect_attempts_max_retries());
+      s->current.retry_attempts.increment();
 
       // Are we done with this particular parent?
-      if ((s->current.retry_attempts.get() - 1) % s->txn_conf->per_parent_connect_attempts != 0) {
+      if (s->current.retry_attempts.get() % s->txn_conf->per_parent_connect_attempts != 0) {
         // No we are not done with this parent so retry
         HTTP_INCREMENT_DYN_STAT(http_total_parent_switches_stat);
         s->next_action = how_to_open_connection(s);
@@ -3864,7 +3864,7 @@ HttpTransact::retry_server_connection_not_open(State *s, ServerState_t conn_stat
   // disable keep-alive for request and retry //
   //////////////////////////////////////////////
   s->current.server->keep_alive = HTTP_NO_KEEPALIVE;
-  s->current.retry_attempts.increment(s->configured_connect_attempts_max_retries());
+  s->current.retry_attempts.increment();
 
   TxnDebug("http_trans", "retry attempts now: %d, max: %d", s->current.retry_attempts.get(), max_retries);
 
--- a/proxy/http/HttpTransact.h
+++ b/proxy/http/HttpTransact.h
@@ -605,10 +605,9 @@ public:
       }
 
       void
-      increment(MgmtInt configured_connect_attempts_max_retries)
+      increment()
       {
         ++_v;
-        ink_assert(_v <= configured_connect_attempts_max_retries);
       }
 
       unsigned
@@ -944,7 +943,17 @@ public:
     MgmtInt
     configured_connect_attempts_max_retries() const
     {
-      return txn_conf->connect_attempts_max_retries;
+      if (dns_info.looking_up != ResolveInfo::PARENT_PROXY) {
+        return txn_conf->connect_attempts_max_retries;
+      }
+      // For parent proxy, return the maximum retry count for the current parent
+      // intead of the max retries for the whole parent group.  The max retry
+      // count for the current parent is derived from rounding the current
+      // attempt up to next multiple of ppca.
+      auto ppca                    = txn_conf->per_parent_connect_attempts;
+      auto cur_tries               = current.retry_attempts.get() + 1;
+      auto cur_parent_max_attempts = ((cur_tries + ppca - 1) / ppca) * ppca;
+      return std::min(cur_parent_max_attempts, txn_conf->parent_connect_attempts) - 1;
     }
 
   private:


serrislew:
Why is `proxy.config.http.parent_proxy.total_connect_attempts` used instead of `proxy.config.http.parent_proxy.fail_threshold`? It seems like for the os case, `proxy.config.http.connect_attempts_max_retries` is the max connection retries before marking an origin as dead which is what the latter config `fail_threshold` uses before considering the parent unavailable

ywkaras:
It looks like this logic is not used when next hop is determined using strategies.yaml.  If the consensus is parent.config is headed for deprecation, we should consider that is deciding how much effort to put into this.

serrislew:
Maybe the docs need to be updated then since it seems like these configs are used for attempts before failing a request, not making the parent down. Or is that implied? 

It seems like for the parents, there are 2 kinds of attempt limits: 1 for the connection attempts for a specific transaction before failing it and 1 for the connection attempts on a parent before making it unavailable. And from my previous comment, it seems like `proxy.config.http.connect_attempts_max_retries` for OS is used to determine how many attempts are needed before marking this origin dead, which is the latter attempt limit. If it's implied that after `proxy.config.http.parent_proxy.per_parent_connect_attempts` is reached and so we are done with this particular parent that we will mark it down, why do we have this separate config `proxy.config.http.parent_proxy.fail_threshold`?

proxy/http/HttpTransact.h
```diff
@@ -944,7 +944,22 @@ class HttpTransact
...
+      // The max attempt count for the current parent is calculated by rounding
+      // the current attempt up to next multiple of ppca.
+      auto ppca                    = txn_conf->per_parent_connect_attempts;
+      auto cur_attempts            = current.attempts.get();
+      auto cur_parent_max_attempts = ((cur_attempts + ppca - 1) / ppca) * ppca;
+      if (cur_attempts == cur_parent_max_attempts) {
+        // If the current attempt is already a multiple of ppca, get the next
+        // multiple.
+        cur_parent_max_attempts += ppca;
+      }
```
ywkaras, line 961: I don't think you want this `if` statement.  If `ppca` is 10 and `cur_attempts` is 20, that means you're on the last try to connect to the second origin in the next hop protection group.  If there's a non-retryable error,  you don' t want to raise the number of attempts to 30.  That would mean you'd never try the last origin in the next hop protection group.

lzx404243: Without this, when `cur_attempts` is already a multiple of `ppca` when calling this function, the calculation would always return the same `cur_attempts`, which leads to an assertion error in `increment()` down the line.

I do see your point though and we don't want to skip parents. Question: when `cur_attempts` is 20, is it the last try for the second origin, or is it the first try for the last origin?

ywkaras: This call: https://github.com/apache/trafficserver/blob/6e8933f35ebf5a50211e3b06aa52749ac0ec3c3b/proxy/http/HttpTransact.cc#L3646 can't cause an assert.  `configured_connect_attempts_max_retries()` can't be more than `txn_conf->parent_connect_attempts`.  The test in the `if` statement at line 3644 ensures that an assert is impossible if execution reaches line 3644.

The only other call to the increment function is https://github.com/apache/trafficserver/blob/6e8933f35ebf5a50211e3b06aa52749ac0ec3c3b/proxy/http/HttpTransact.cc#L3865.  Not clear under what circumstances this call is reached.  The very large HttpTransact and HttpSM classes have little internal scoping or even encapsulation.  It could takes days or weeks to analyze possible code paths.  As I previously said, it's very hard to work on a source bases like this without taking leaps of faith and making reasonable assumptions.  So I suggest making the reasonable assumption that a retry will be blocked if the number of retries is already at the maximum.  To back that up, you could write long Au test(s) exercising parent.config (as I did for strategies.yaml).  But if parent.config is soon to be deprecated, it's questionable if the effort involved is worth it.

lzx404243:
@serrislew Thanks for the question. I share the same question as to what the `fail_threshold` does and the relationship with the other parent configs and I believe the doc can be improved to clarify that. 

This `configured_connect_attempts_max_retries()` function is currently used when setting the attempt counts. We choose to return values based on the `parent_proxy.total_connect_attempt` and `parent_proxy.per_parent_connect_attempts` because we see the attempt count is only compared against those values, such as in https://github.com/apache/trafficserver/blob/master/proxy/http/HttpTransact.cc#L3644 .

As for `parent_proxy.fail_threshold`, guessing from the source code I think it might be a next-tier accounting that takes into account a timing component.  From https://github.com/apache/trafficserver/blob/master/proxy/ParentSelectionStrategy.cc#L90-L93 , it seems a parent can be marked "down" several times and is only set to unavailable when it passed the `fail_threshold`.

Based on the above, my impression is that `parent_proxy.total_connect_attempt` and `parent_proxy.per_parent_connect_attempts` configs work with the attempt count to determine whether to use parent proxy and when to switch to the next parent. `parent_proxy.fail_threshold` seems to be separate from this.

proxy/http/HttpTransact.h
lzx404243: Right. But the same logic applies when calling `maximize()` with the `configured_connect_attempts_max_retries()`:
```
current.attempts : 0
per_parent_connect_attempts: 2
parent_proxy.total_connect_attempts: 100
```
the `configured_connect_attempts_max_retries()` returns 0, which is passed into `maximize()`. As a result, `current.attempts` is unchanged after the call, while what we want is to "maximize" the attempt count so that the current parent is skipped. There seems to be an issue here too. Do you have any suggestions on this?

ywkaras: From looking at the code, I thought the attempt numbers are 1 to N.  That is the number is 1 during the first attempt, not zero.  So the attempt number would never be zero when `maximize()` is called.

ywkaras: OK, I see now, "retries" doesn't include the first try, so max tries = max retries + 1.

ywkaras: How about?
```
    MgmtInt
    configured_connect_attempts_max_retries() const
    {
      if (parent_params != nullptr) {
        auto ppca = txn_conf->per_parent_connect_attempts;
        auto tries = current.attempts.get() + 1;
        // Round g up to next multiple of ppca.
        auto ru = ((tries + ppca - 1) / ppca) * ppca;
        return std::min(ru, txn_conf->parent_connect_attempts) - 1;
      }
      return txn_conf->connect_attempts_max_retries;
    }
```
But this would also require removing the parameter and assert in `CurrentInfo::Attempts::increment()`, 
and changing line 3644 in HttpTransact.cc to:
```
    if (s->current.attempts.get() < (s->txn_conf->parent_connect_attempts - 1)) {
```
and changing line 3649 in HttpTransact.cc to:
```
      if (s->current.attempts.get() % s->txn_conf->per_parent_connect_attempts != 0) {
```

lzx404243: Yep after some more digging and thoughts, I believe this is correct. I reach out to Susan and she confirms that the current `HttpTransact::_CurrentInfo.attempt` is really the retry attempt. Based on this understanding, what you suggested here makes a lot of sense and I think we are good to go.

proxy/http/HttpTransact.h
```diff
@@ -605,10 +605,9 @@ class HttpTransact
       }
 
       void
-      increment(MgmtInt configured_connect_attempts_max_retries)
+      increment()
       {
         ++_v;
-        ink_assert(_v <= configured_connect_attempts_max_retries);
```
serrislew, line 611: why wouldn't this assert apply now with the new changes? seems like increment doesn't have to change now that you check attempts - 1 before you increment

lzx404243: This is a great question. The check before the increment ensures that `attempts - 1 < the global parent_max_retries`. The problem with this assert is that the argument passed in `configured_connect_attempts_max_retries` may not be the global parent max retry, but the max retry count for the current parent. Currently, the argument is set to the the output of `s->configured_connect_attempts_max_retries()`, which is returning:

- if connecting to the OS, returns the `txn_conf->connect_attempts_max_retries` 
- if connecting to a parent proxy, returns the max retries for the current parent(bounded by the global max retry for all parents)
In the second case, the assert can fail as after the increment, it went over the max for the current parent(which should be allowed as we switch to the next parent).

serrislew: but doesn't this line `auto cur_parent_max_attempts = ((cur_tries + ppca - 1) / ppca) * ppca;` take into account the max retries for the parent but also in relation to the total? even when we hit the max retries for the current parent, don't your new changes in `configured_connect_attempts_max_retries()` account for the next parent. each increment should be protected. im trying to find the edge case when this assert fails, could you provide an example

proxy/http/HttpTransact.cc
```diff
@@ -3641,12 +3641,12 @@ HttpTransact::handle_response_from_parent(State *s)
       return;
     }
 
-    if (s->current.attempts.get() < s->txn_conf->parent_connect_attempts) {
```
serrislew, line 3644: isn't this an off by one error then? if `s->txn_conf->parent_connect_attempts` is 3 so we can try 3 times and we are on the 3rd attempt but since we don't increment yet, `s->current.attempts.get() = 2` so that conditional would fail so we don't try the 3rd time?

proxy/http/HttpTransact.h
ywkaras: The assert can be triggered if parent_connect_attempts > (connect_attempts_max_retries + 1).  If you can find checking logic in the config code that enforces this, it would be safe to restore the assert.

This was the original version of increment():
```
      void
      increment()
      {
        ++_v;
        ink_assert(_v <= configured_connect_attempts_max_retries());
      }
```
This was CurrentInfo::configured_connect_attempts_max_retries():
```
inline MgmtInt
HttpTransact::CurrentInfo::Attempts::configured_connect_attempts_max_retries() const
{
  return reinterpret_cast<const HttpTransact::State *>(reinterpret_cast<const char *>(this) -
                                                       offsetof(HttpTransact::State, current.attempts))
    ->configured_connect_attempts_max_retries();
}
```
Even though this code was verified in Yahoo prod, and this sort of pointer arithmetic is how inheritance is implemented in C++, this code was rejected in review.  I never would have put in the assert if I thought in would require adding a parameter to increment().

proxy/http/HttpTransact.cc
lzx404243: Good point. I think the confusion arises from the naming. `s->current.attempts.get()` is actually the retry count, so when `s->current.attempts.get() = 2`, it has retried for 2 times already(3 attempts in total).

The naming confusion is addressed in another PR: https://github.com/apache/trafficserver/pull/9655 .

proxy/http/HttpTransact.h
lzx404243: > even when we hit the max retries for the current parent, don't your new changes in `configured_connect_attempts_max_retries()` account for the next parent?

`configured_connect_attempts_max_retries()` normally would be increment to the next multiple of `ppca`(accounts for the next parent), except for the case where the attempt/(retry + 1) is already a multiple of ppca(current parent hits max retries), in which case the calculation basically returns the same attempt/retry count as before. 

> im trying to find the edge case when this assert fails, could you provide an example

The above logic make sense when we call `maximize()` with the `configured_connect_attempts_max_retries()` to disable further retries, since we already hit the max retries for the current parent, we are not setting it to anything larger.

However, it will fail the assert in `increment()` with the `configured_connect_attempts_max_retries()`, since `n = configured_connect_attempts_max_retries(n).`

serrislew (approved):

bneradt:
[approve ci autest]

PR #9622, opened 2023-04-19
quic: Fix session cleanup assert.

brbzull0:


maskit (approved):
Thanks, that's what I wanted to check here.

PR #9631, opened 2023-04-21
Change remap filter behavior to match ip_allow.yaml

zwoop:
This could possibly be considered a bug, but certainly it's not a great behavior as it was. Before this patch, filters (as defined and used with .definefilter and .activatefilter), are always all evaluated, with a priority towards deny.

ip_allow.yaml however works by evaluating only filters up until one matches, and whatever action has triggered, is used.

This patch changes these remap filters to follow the behavior of ip_allow.yaml. First filter that matches is applied, and evaluation of filters stops.

I would really have liked to see this in 9.2.x, but I'm not so sure that it's a good idea since it does also break compatibility even if we consider this a bug.


zwoop:
Seems likely that this would also address some of the issues report in #1971.

PR #9655, opened 2023-04-28
rename attempts to retry_attempts

lzx404243:
From the name, the `HttpTransact::_CurrentInfo.attempts` seems to represent the total connection attempt, but it's currently really the retry attempt:
- it's initialized to `0`
- it's incremented only in the OS or parent retry code path

This PR renames this to `HttpTransact::_CurrentInfo.retry_attempts` to avoid confusion. Note that there is no functionality changes. Changes are to just the naming and some related log messages(plus one assertion).


bneradt (commented):
Looks good to me. The previous naming has confused people in the past, so thank you for updating this.

I'll let a non-Yahoo person give an official approval.

cmcfarlen (approved):
Thanks!

PR #9656, opened 2023-04-28
Prefer the __linux__ macro over linux.

jpeach:
Depending on the standards mode, GCC may not predefine the "linux" macro. However, "__linux__" is always predefined, so we can use that consistently and reliably.


masaori335 (approved):
Looks good to me.

PR #9657, opened 2023-04-29
Remove deprecated debug output functions from 10 source files.

ywkaras:


iocore/cache/RamCacheLRU.cc
```diff
@@ -58,6 +58,12 @@ struct RamCacheLRU : public RamCache {
   RamCacheLRUEntry *remove(RamCacheLRUEntry *e);
 };
 
+#ifdef DEBUG
+
+static DbgCtl dbg_ctl_ram_cache{"ram_cache"};
```
cmcfarlen, line 63: anon namespace here too?

iocore/cache/P_CacheInternal.h
```diff
@@ -49,9 +49,9 @@ struct EvacuationBlock;
 #endif
 
 #ifdef DEBUG
-#define DDebug(tag, fmt, ...) Debug(tag, fmt, ##__VA_ARGS__)
+#define DDbg(dbg_ctl, fmt, ...) Dbg(dbg_ctl, fmt, ##__VA_ARGS__)
```
cmcfarlen, line 52: These seem pretty useful generally. Should we move the up next to Dbg?

include/tscore/Trie.h
```diff
@@ -30,9 +30,16 @@
 #include "tscore/List.h"
 #include "tscore/Diags.h"
 
+class TrieImpl
+{
+protected:
+  inline static DbgCtl dbg_ctl_insert{"Trie::Insert"};
+  inline static DbgCtl dbg_ctl_search{"Trie::Search"};
+};
```
cmcfarlen, line 38: Is this so these only get constructed if a Trie does?

cmcfarlen (approved):
Might want to stick to the anon namespaces

iocore/cache/P_CacheInternal.h
ywkaras: Perhaps, but these PRs are big enough without mission creep.

cmcfarlen (approved):

PR #9667, opened 2023-05-02
Add new settings to specify TLS versions to use

maskit:
I found an incompatibility between OpenSSL and BoringSSL.

If `valid_tls_versions_in` has `[TLSv1, TLSv1_2, TLSv1_3]` (note that `TLSv1_1` isn't in the list), BoringSSL only allows TLSv1 where OpenSSL allows all the three versions. It sounds like a bug but it's actually not.

https://boringssl.googlesource.com/boringssl/+/86ada1ea2f51ff41baa2919337e5d721bd27f764/ssl/ssl_versions.cc#225
> // If there is a disabled version after the first enabled one, all versions
// after it are implicitly disabled.

And I also found that the way (API) we currently use to specify available TLS versions is deprecated.

The old way allow us to specify versions individually (like above), but the new way only allows us to specify a range of versions.

Old:
https://www.openssl.org/docs/man3.1/man3/SSL_CTX_set_options.html
https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_OP_NO_TLSv1

New:
https://www.openssl.org/docs/man3.1/man3/SSL_CTX_set_min_proto_version.html
https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_set_min_proto_version

Since the old way is no longer supported by SSL libraries, we should deprecate the current way and only allow ATS users to specify a range.

----

This adds these new settings:
```
records.yaml:
  proxy.config.ssl.server.version.min
  proxy.config.ssl.server.version.max
  proxy.config.ssl.client.version.min
  proxy.config.ssl.client.version.max
sni.yaml:
  valid_tls_version_min_in
  valid_tls_version_max_in
```

and deprecates these settings:
```
records.yaml:
  proxy.config.ssl.TLSv1
  proxy.config.ssl.TLSv1_1
  proxy.config.ssl.TLSv1_2
  proxy.config.ssl.TLSv1_3
  proxy.config.ssl.client.TLSv1
  proxy.config.ssl.client.TLSv1_1
  proxy.config.ssl.client.TLSv1_2
  proxy.config.ssl.client.TLSv1_3
sni.yaml:
  valid_tls_versions_in
```

### DIFF
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -3583,11 +3583,45 @@ SSL Termination
 
    This configuration works with OpenSSL v1.0.2 and above.
 
+.. ts:cv:: CONFIG proxy.config.ssl.server.version.min INT -1
+
+   Specifies the minimum TLS version that will be offered to clients during the TLS negotiation.
+
+   ======== ===================================================================
+   Value    Version
+   ======== ===================================================================
+   ``0``    TLS 1.0
+   ``1``    TLS 1.1
+   ``2``    TLS 1.2
+   ``3``    TLS 1.3
+   ======== ===================================================================
+
+.. ts:cv:: CONFIG proxy.config.ssl.server.version.max INT -1
+
+   Specifies the maximum TLS version that will be offered to clients during the TLS negotiation.
+
+   ======== ===================================================================
+   Value    Version
+   ======== ===================================================================
+   ``0``    TLS 1.0
+   ``1``    TLS 1.1
+   ``2``    TLS 1.2
+   ``3``    TLS 1.3
+   ======== ===================================================================
+
 .. ts:cv:: CONFIG proxy.config.ssl.TLSv1 INT 0
+   :deprecated:
+
+   This setting is depreated in favor of :ts:cv:`proxy.config.ssl.server.version.min` and
+   :ts:cv:`proxy.config.ssl.server.version.min`, and will be ignored if those new settings are used.
 
    Enables (``1``) or disables (``0``) TLSv1.0. If not specified, disabled by default.
 
 .. ts:cv:: CONFIG proxy.config.ssl.TLSv1_1 INT 0
+   :deprecated:
+
+   This setting is depreated in favor of :ts:cv:`proxy.config.ssl.server.version.min` and
+   :ts:cv:`proxy.config.ssl.server.version.min`, and will be ignored if those new settings are used.
 
    Enables (``1``) or disables (``0``) TLS v1.1.  If not specified, disabled by default.  [Requires OpenSSL v1.0.1 and higher]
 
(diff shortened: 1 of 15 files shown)


doc/admin-guide/files/records.yaml.en.rst
```diff
@@ -3583,11 +3583,45 @@ SSL Termination
...
+   ``0``    TLS 1.0
+   ``1``    TLS 1.1
+   ``2``    TLS 1.2
+   ``3``    TLS 1.3
+   ======== ===================================================================
+
 .. ts:cv:: CONFIG proxy.config.ssl.TLSv1 INT 0
+   :deprecated:
+
+   This setting is depreated in favor of :ts:cv:`proxy.config.ssl.server.version.min` and
```
brbzull0, line 3615: type: deprecated
there are more probably from the copy/paste

iocore/net/SSLConfig.cc
```diff
@@ -280,47 +280,85 @@ SSLConfigParams::initialize()
...
-    ssl_ctx_options |= SSL_OP_NO_TLSv1;
-  }
+  REC_ReadConfigInteger(client_tls_ver_min, "proxy.config.ssl.client.version.min");
+  REC_ReadConfigInteger(client_tls_ver_max, "proxy.config.ssl.client.version.max");
+  if (client_tls_ver_min < 0 || client_tls_ver_max < 0) {
+    REC_ReadConfigInteger(option, "proxy.config.ssl.client.TLSv1");
+    if (!option) {
+      ssl_client_ctx_options |= SSL_OP_NO_TLSv1;
+    } else {
+      // This is disabled by default. It it's used if it's enabled.
```
brbzull0, line 290: typo: `It it's`

maskit:
[approve ci]

iocore/net/P_SNIActionPerformer.h
```diff
@@ -341,15 +343,25 @@ class TLSValidProtocols : public ActionItem
...
-      auto ssl_vc            = dynamic_cast<SSLNetVConnection *>(snis);
+    if (this->min_ver >= 0 || this->max_ver >= 0) {
       const char *servername = snis->get_sni_server_name();
-      Debug("ssl_sni", "TLSValidProtocol param 0%x, fqdn [%s]", static_cast<unsigned int>(this->protocol_mask), servername);
-      ssl_vc->set_valid_tls_protocols(protocol_mask, TLSValidProtocols::max_mask);
+      Debug("ssl_sni", "TLSValidProtocol min=%d, max=%d, fqdn [%s]", this->min_ver, this->max_ver, servername);
+      auto ssl_vc = dynamic_cast<SSLNetVConnection *>(snis);
+      ssl_vc->set_valid_tls_version_min(this->min_ver);
+      ssl_vc->set_valid_tls_version_max(this->max_ver);
+    } else {
```
ywkaras, line 357: Why not `} else if (!unset) {` ?

ywkaras (approved):

iocore/net/P_SNIActionPerformer.h
maskit: I guess I wanted to make it clear that `if (!unset) {` is the first line for the case min/max isn't used. It enables us to add some code before the `if` easily.

PR #9668, opened 2023-05-02
QUIC: Add a unit tests to validate that the qlog file is generated

brbzull0:
While investigating another issue I found out that I was having some crashes inside quiche because (possible) the qlog string. So I am adding these tests to make sure we detect any crash related to the qlog generation.

This also uses buffer writter to build up the qlog name(instead of snprintf).


bneradt:
[approve ci rocky autest]

maskit:
Can we have a gold-file to examine the generated qlog file? I don't think sleep-and-ls is a common way for autest.

brbzull0:
> Can we have a gold-file to examine the generated qlog file? I don't think sleep-and-ls is a common way for autest.

the issue is that we do not have the trace id to know which file is the one we are supposed to read to then compare with a gold file.

I guess I can make a quick script to rename whatever TS is droping as qlog files and the read the renamed file's content(and then apply the gold file). 
How does this looks to you?

maskit:
Oh, I see, filename. Renaming is an option.

@bneradt Do you have any suggestions?

bneradt:
> > Can we have a gold-file to examine the generated qlog file? I don't think sleep-and-ls is a common way for autest.
> 
> the issue is that we do not have the trace id to know which file is the one we are supposed to read to then compare with a gold file. An example of a file would be `test_qlog-9b9eca8989ab274e14a3acd9e95e1a23247f.sqlog`
> 
> I guess I can make a quick script to rename whatever TS is droping as qlog files and the read the renamed file's content(and then apply the gold file). How does this looks to you?

I don't think there's a way to handle run-time naming. , this seems like the way to go:

1. `cp` the output to a well known file.
2. Use a gold file, or ContainsExpression, to verify the contents.

PR #9624, opened 2023-04-19
Remove dependency on OpenSSL's OCSP API

maskit:
Changes:
- Remove all boringocsp stuff
- Copy OCSP_ functions from OpenSSL 3 (Apache License 2) and add TS_ prefix.
- Copy ASN1 definitions for OCSP from OpenSSL 3
- Modify code not to use functions only available on OpenSSL
- Modify code to make it C++ (original code is C and has issues compiling as C++ code)
- Change ERR_raise to ATS's Debug

PR #9672, opened 2023-05-02
Fixes crashes around OCSP with FetchSM

maskit:
We saw two kind of crashes while testing #9591 and #9624, and the both were triggered by OCSP request timeout. In a nutshell, the crashes are due to accessing unavailable continuations. Continuations become unavailable when OCSP requests timeout, but the status were not properly updated/checked.

### DIFF
--- a/iocore/net/OCSPStapling.cc
+++ b/iocore/net/OCSPStapling.cc
@@ -288,12 +288,10 @@ public:
     if (event == TS_EVENT_IMMEDIATE) {
       this->fetch();
     } else {
-      auto fsm = reinterpret_cast<FetchSM *>(e);
-      auto req = reinterpret_cast<HTTPRequest *>(fsm->ext_get_user_data());
       if (event == TS_FETCH_EVENT_EXT_BODY_DONE) {
-        req->set_done();
+        this->set_done();
       } else if (event == TS_EVENT_ERROR) {
-        req->set_error();
+        this->set_error();
       }
     }
 
@@ -310,7 +308,6 @@ public:
     sin.sin_port        = 65535;
 
     this->_fsm = FetchSMAllocator.alloc();
-    this->_fsm->ext_set_user_data(this);
     if (use_post) {
       this->_fsm->ext_init(this, "POST", uri, "HTTP/1.1", reinterpret_cast<sockaddr *>(&sin), 0);
     } else {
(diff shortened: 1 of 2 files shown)


iocore/net/OCSPStapling.cc
```diff
@@ -355,31 +355,43 @@ class HTTPRequest : public Continuation
   fetch()
   {
     SCOPED_MUTEX_LOCK(lock, mutex, this_ethread());
+    this->_result = 0;
```
ywkaras, line 358: Looking at this:
```
wkaras ~/REPOS/TS
O$ grep -e '\Wfetch(' -e '\Wset_done(' -e '\Wset_error(' $(findsrc)
./include/tscore/ArgParser.h:  void set_error(std::string e);
./iocore/net/OCSPStapling.cc:      this->fetch();
./iocore/net/OCSPStapling.cc:        req->set_done();
./iocore/net/OCSPStapling.cc:        req->set_error();
./iocore/net/OCSPStapling.cc:  fetch()
./iocore/net/OCSPStapling.cc:  set_done()
./iocore/net/OCSPStapling.cc:  set_error()
./plugins/experimental/inliner/cache.h:  fetch(const std::string &k, A &&...a)
./plugins/experimental/slice/prefetch.cc:  if (bg->fetch(data)) {
./plugins/experimental/slice/prefetch.cc:BgBlockFetch::fetch(Data *const data)
./plugins/experimental/slice/prefetch.h:  bool fetch(Data *const data);
./plugins/lua/ts_lua_fetch.c:  /* ts.fetch() */
./src/tscore/ArgParser.cc:ArgParser::set_error(std::string e)
./src/tscore/unit_tests/test_ArgParser.cc:  parser2.set_error("error");
wkaras ~/REPOS/TS
O$
```
it's clear that `fetch()`, `set_done()` and `set_error()` are only called by `event_handler()`.  So, the continuation mutex will already be locked when they are called.  It also means these member functions should be private.  I suggest you make them private as a part of this PR.

ywkaras: In `event_handler()`, won't `req` always be the same as `this`?  What's the point of it then?

iocore/net/OCSPStapling.cc
```diff
@@ -1016,13 +1028,21 @@ query_responder(const char *uri, const char *user_agent, TS_OCSP_REQUEST *req, i
...
     ink_hrtime_sleep(HRTIME_MSECONDS(1));
   } while (!httpreq.is_done() && (Thread::get_hrtime() < end));
 
+  if (!httpreq.is_done()) {
+    Error("OCSP request was timed out; uri=%s", uri);
+    if (!httpreq.is_initiated()) {
+      Debug("ssl_ocsp", "Request is not initiated yet. Cancelling the event.");
+      e->cancel(&httpreq);
+    }
+  }
```
ywkaras, line 1044: Add a `}' to close the block and release the mutex.

iocore/net/OCSPStapling.cc
maskit: I moved the functions private and removed unnecessary calls for ext_set/get_user_data.

> So, the continuation mutex will already be locked when they are called. 

This may be true, but I kept `SCOPED_MUTEX_LOCK`, which should be almost no-op if the lock is acquired. Because `e->mutex` and `e->continuation->mutex` can be different under some conditions and `e->continuation->mutex` is not locked in that case. It may not be the case for this, but I'd rather explicitly acquire the lock.

PR #9674, opened 2023-05-02
Add error log for invalid OCSP response

maskit:


maskit:
@jpeach Thanks for pointing out the unchecked return value.

iocore/net/OCSPStapling.cc
```diff
@@ -957,8 +957,10 @@ stapling_check_response(certinfo *cinf, TS_OCSP_RESPONSE *rsp)
   if (!TS_OCSP_resp_find_status(bs, cinf->cid, &status, &reason, &rev, &thisupd, &nextupd)) {
     // If ID not present just pass it back to client
     Error("stapling_check_response: certificate ID not present in response for %s", cinf->certname);
-  } else {
-    TS_OCSP_check_validity(thisupd, nextupd, 300, -1);
+  }
+  if (!TS_OCSP_check_validity(thisupd, nextupd, 300, -1)) {
```
jpeach, line 961: If you failed to get the status, is it still OK to check the validity?

iocore/net/OCSPStapling.cc
```diff
@@ -957,8 +957,10 @@ stapling_check_response(certinfo *cinf, TS_OCSP_RESPONSE *rsp)
...
+    // The check is just for logging and pass the response back to client anyway
+    Error("stapling_check_response: status in response for %s is not valid already/yet", cinf->certname);
```
jpeach, line 963: Is there any way (or need to) to propagate the actual error out?

iocore/net/OCSPStapling.cc
```diff
@@ -957,8 +957,10 @@ stapling_check_response(certinfo *cinf, TS_OCSP_RESPONSE *rsp)
   if (!TS_OCSP_resp_find_status(bs, cinf->cid, &status, &reason, &rev, &thisupd, &nextupd)) {
     // If ID not present just pass it back to client
     Error("stapling_check_response: certificate ID not present in response for %s", cinf->certname);
-  } else {
-    TS_OCSP_check_validity(thisupd, nextupd, 300, -1);
+  }
```
jpeach, line 960: IIUC, even if the response has a not-good status or we find it is invalid, we still cache it and send it to the client? Is that the right thing to do?

maskit: I'm not sure if we should cache it, but I think we should just tell the truth to the client to give the information that the response is invalid.

ywkaras (approved):

PR #9684, opened 2023-05-05
Replace curl with proxy verifier in proxy protocol tests

### LINKED ISSUE #9652: AuTest: Switch from curl for proxy_protocol.test.py
The current proxy_protocol.test.py uses curl to verify Proxy Protocol usage. The gold file for the test expects curl to output the Proxy Protocol header it sends. Recent versions of curl do not output this string anymore. Maybe that's not a big deal - we can probably assume it does send the Proxy Protocol header. But it would probably be preferable to use Proxy Verifier for this test.

lzx404243:
Per #9652 , it would be good to replace curl with proxy verifier in `proxy_protocol.test.py`. 

This PR updates the tests to use proxy verifier. As a result, some files are cleaned up:
* all the gold files are cleaned up(except `access.gold`) as verifications are done in the replay files
* `proxy_protocol_server.py` is no longer needed as proxy verifier supports PROXY header.

Also added a few test cases:

- test pp-in for v2 on tcp port
- test pp-in for v2 on ssl port

This resolves #9652 .


tests/gold_tests/proxy_protocol/proxy_protocol.test.py
```diff
@@ -17,36 +17,37 @@
...
 
+    replay_file = "replay/proxy_protocol_in.replay.yaml"
+
     def __init__(self):
         self.setupOriginServer()
+        self.setupPVOriginServer()
         self.setupTS()
 
     def setupOriginServer(self):
         self.httpbin = Test.MakeHttpBinServer("httpbin")
```
masaori335, line 36: Do we still need the `httpbin` backend?

lzx404243: Nope. That's a good catch. I cleaned up all usages of `httpbin`.

masaori335 (approved):
Looks good. Thanks for the conversion.

PR #9697, opened 2023-05-10
Send a 502 to the client when there is an invalid Transfer-Encoding header value from the origin

bryancall:
Did a git mv on chunked.replay.yaml to chunked_diabled.replay.yaml

Added proxy verifier tests to test:
Transfer-Encoding header not equal to chunked from origin
Transfer-Encoding and Content-Length headers missing from origin
valid Transfer-Encoding: chunked response from the origin


masaori335:
[approve ci]

bneradt:
[approve ci]

bneradt (changes requested):
Looks great.

Rather than the _2 test, I'd just add your test class to the current chunked_encoding.test.py.

ezelkow1:
[approve ci centos]

serrislew:
If chunking is disabled but origin server responds with invalid Transfer-Encoding header, will it send 502 (from bad response) or 200 (since it is not even considered)?

bryancall:
If any invalid Tranfer-Encoding header comes from the origin traffic server will respond with a 502, regardless if chunking is on or off.

serrislew (approved):

bryancall:
[approve ci]

PR #9698, opened 2023-05-10
Fix PROXY protocol out with tls

lzx404243:
While updating the PROXY protocol tests, I observe the https connections to the origin would hang when PROXY protocol out is enabled in ATS. This PR addresses this.

### root cause

After the #9366 changes, the logic of kicking off state machine is in `ConnectingEntry` and is triggered by callbacks upon one of `VC_EVENT_READ_COMPLETE`, `VC_EVENT_WRITE_READY` or `VC_EVENT_WRITE_COMPLETE` events.  The root cause of the hang has to do with none of these events is triggered:
* Without PROXY protocol enabled, the `VC_EVENT_WRITE_READY` is triggered and the connection flows. However when PROXY protocol out is enabled, as PROXY protocol data is sent before ssl handshake,  the accounting of `vio.nbytes` and `vio.done`changed , leading to `VC_EVENT_WRITE_READY` not being triggered. 
* `VC_EVENT_READ_COMPLETE` is not triggered because `vio` read is not enabled. Prior to #9366, `attach_server_session()` enables vio read and is called prior to the handshake. After #9366, this is called after the handshake so the `vio` read is disabled during the handshake. As read is disabled, the `VC_EVENT_READ_COMPLETE` can't be signaled to the handler.

This PR adds a 0-byte read so the `vio` read is enabled and `ConnectingEntry::state_http_server_open()` handler can be signaled and called back with the `VC_EVENT_READ_COMPLETE` event.

### DIFF
--- a/tests/gold_tests/post_slow_server/gold/post_slow_server_max_requests_in_0_stderr.gold
+++ b/tests/gold_tests/post_slow_server/gold/post_slow_server_max_requests_in_0_stderr.gold
@@ -1,5 +1,5 @@
 ``
 > POST / HTTP/1.1
 ``
-< HTTP/1.1 502 Connection refused
+< HTTP/1.1 502 Broken pipe
 ``
--- a/tests/gold_tests/proxy_protocol/proxy_protocol.test.py
+++ b/tests/gold_tests/proxy_protocol/proxy_protocol.test.py
@@ -242,9 +242,9 @@ class ProxyProtocolOutTest:
 ProxyProtocolInTest().run()
 
 # non-tunnling HTTP to origin
-ProxyProtocolOutTest(pp_version=-1, is_tunnel=False).run()
-ProxyProtocolOutTest(pp_version=1, is_tunnel=False).run()
-ProxyProtocolOutTest(pp_version=2, is_tunnel=False).run()
+ProxyProtocolOutTest(pp_version=-1, is_tunnel=False, is_tls_to_origin=False).run()
+ProxyProtocolOutTest(pp_version=1, is_tunnel=False, is_tls_to_origin=False).run()
+ProxyProtocolOutTest(pp_version=2, is_tunnel=False, is_tls_to_origin=False).run()
 # non-tunnling HTTPS to origin
 ProxyProtocolOutTest(pp_version=-1, is_tunnel=False, is_tls_to_origin=True).run()
 ProxyProtocolOutTest(pp_version=1, is_tunnel=False, is_tls_to_origin=True).run()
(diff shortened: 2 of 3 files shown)


tests/gold_tests/proxy_protocol/proxy_protocol.test.py
```diff
@@ -227,7 +225,7 @@ def setupTS(self, tr: 'TestRun') -> None:
         )
 
         self._ts.Disk.remap_config.AddLine(
-            f"map / http://backend.pp.origin.com:{self._server.Variables.external_port}/")
+            f"map / https://backend.pp.origin.com:{self._server.Variables.external_port}/")
```
maskit, line 228: Why don't you add a new test? Is current test invalid?

lzx404243: Good point. Added some https pp_out tests in addition to the existing http ones.

tests/gold_tests/post_slow_server/gold/post_slow_server_max_requests_in_0_stderr.gold
```diff
@@ -1,5 +1,5 @@
 ``
 > POST / HTTP/1.1
 ``
-< HTTP/1.1 502 Connection refused
+< HTTP/1.1 502 Broken pipe
```
lzx404243, line 4: Note to reviewers: 

The `post_slow_server_max_requests_in` was failing with my changes, as it expects a `502 Connection refused`, but gets a `502 Broken pipe`. 

The expectation changed from `502 Broken pipe`->`502 Connection refused` as part of the #9366. Discussed with @bneradt and we believe we should update/revert the expectation to the original. After this PR, certain things seem to be more aligned to how they work pre H2ToOrigin.

bneradt (changes requested):
Looks great. Thanks for fixing and updating the tests.

tests/gold_tests/proxy_protocol/proxy_protocol.test.py
```diff
@@ -235,8 +241,14 @@ def run(self) -> None:
 
 ProxyProtocolInTest().run()
 
+# non-tunnling HTTP to origin
 ProxyProtocolOutTest(pp_version=-1, is_tunnel=False).run()
 ProxyProtocolOutTest(pp_version=1, is_tunnel=False).run()
 ProxyProtocolOutTest(pp_version=2, is_tunnel=False).run()
```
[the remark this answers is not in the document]
lzx404243: Make sense, thx. Updated the tests.

bneradt:
[approve ci]

ezelkow1:
[approve ci centos]

cmcfarlen (approved):

bneradt (approved):

PR #9702, opened 2023-05-11
Don't throw exceptions if io_uring cannot be initialized. 

cmcfarlen:
If the kernel doesn't support io_uring, but it's compiled in, we would throw an exception and crash ATS.  This PR just marks the ring bad and provides a function to check if io_uring is valid.  This way ATS can respond at runtime and fallback to non io_uring code.

### DIFF
--- a/iocore/aio/AIO.cc
+++ b/iocore/aio/AIO.cc
@@ -561,9 +561,9 @@ prep_read(io_uring_sqe *sqe, AIOCallbackInternal *op)
 void
 prep_readv(io_uring_sqe *sqe, AIOCallbackInternal *op)
 {
-  op->iovec.iov_len  = op->aiocb.aio_nbytes;
-  op->iovec.iov_base = op->aiocb.aio_buf;
-  io_uring_prep_readv(sqe, op->aiocb.aio_fildes, &op->iovec, 1, op->aiocb.aio_offset);
+  op->iov.iov_len  = op->aiocb.aio_nbytes;
+  op->iov.iov_base = op->aiocb.aio_buf;
+  io_uring_prep_readv(sqe, op->aiocb.aio_fildes, &op->iov, 1, op->aiocb.aio_offset);
 }
 
 void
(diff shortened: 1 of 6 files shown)


bneradt:
[approve ci]

cmcfarlen:
[approve ci autest]

iocore/aio/AIO.cc
```diff
@@ -542,22 +549,111 @@ AIOThreadInfo::aio_thread_main(AIOThreadInfo *thr_info)
...
+    prep_ops[LIO_READ]  = prep_readv;
+    prep_ops[LIO_WRITE] = prep_writev;
+  }
+}
+
+void
+io_uring_prep_ops_internal(AIOCallbackInternal *op_in, int op_type)
+{
+  IOUringContext *ur      = IOUringContext::local_context();
+  AIOCallbackInternal *op = op_in;
```
bryancall, line 604: Would be nice if we could iterate over AIOCallbackInternal.

iocore/aio/AIO.cc
```diff
@@ -573,28 +669,15 @@ AIOCallbackInternal::handle_complete(io_uring_cqe *cqe)
...
-      op->aiocb.aio_lio_opcode = LIO_READ;
-      if (op->then) {
-        sqe->flags |= IOSQE_IO_LINK;
-      } else {
-        op->aio_op = static_cast<AIOCallbackInternal *>(op_in);
-      }
-
-      op = static_cast<AIOCallbackInternal *>(op->then);
-    }
+    io_uring_prep_ops_internal(static_cast<AIOCallbackInternal *>(op_in), LIO_READ);
```
bryancall, line 680: I would push the static_cast into io_uring_prep_ops_internal.

iocore/aio/AIO.cc
```diff
@@ -220,7 +226,8 @@ ink_aio_init(ts::ModuleVersion v, AIOBackend backend)
   case AIOBackend::AIO_BACKEND_AUTO: {
     // detect if io_uring is available and can support the required features
     auto *ctx = IOUringContext::local_context();
-    if (ctx && ctx->supports_op(IORING_OP_WRITE) && ctx->supports_op(IORING_OP_READ)) {
+    if (ctx && ctx->valid()) {
+      setup_prep_ops(ctx);
```
bryancall, line 230: Now use_io_uring can be true even if the setup failed to initialize prep_ops?

bryancall (changes requested):
I more concerned about use_io_uring being set to true when prep_ops didn't get initialized.  Also before you were checking for IOURING_OP_WRITE and IOURING_OP_READ support and now it only checks for IOURING_OP_READ.  Was there a reason for the change?
Other stuff is just stylistic comments.

iocore/aio/AIO.cc
cmcfarlen: It should default to the always available readv/writev but I notice I use readv/writev for the READ supported case as well.  I'll fix that and make a comment to make it more clear.

iocore/aio/AIO.cc
cmcfarlen: Sometimes this is called with a pointer type that is already AIOCallbackInternal.  I'm not a fan of these casts at all, but they are here to coexist with the thread AIO mode.

bryancall (approved):

PR #9707, opened 2023-05-14
Add a new configuration to ignore query params in cache keys

zwoop:
This adds a new configuration option:

    CONFIG proxy.config.http.cache.ignore_query INT 0

This is off by default, but it is overridable. The typical use case for this would be to use e.g. conf_remap.so plugin, to set this for services / URLs which shouldn't use the query parameters in the HTTP cache key calculations.

This would be significantly less expensive (less CPU) than using e.g. the cachekey.so plugin to simply strip the query parameters from the key.


zwoop:
Moderately tested, but first without setting this configuration:

```
leif@Hel ~/ » curl -D -s -o /dev/null "http://hel.ogre.com/bench/100.bmp" | grep HTTP/1.1
HTTP/1.1 200 OK
leif@Hel ~/ » curl -D -s -o /dev/null -H "Cache-Control: only-if-cached" "http://hel.ogre.com/bench/100.bmp" | grep HTTP/1.1
HTTP/1.1 200 OK
leif@Hel ~/ » curl -D -s -o /dev/null -H "Cache-Control: only-if-cached" "http://hel.ogre.com/bench/100.bmp?foo=bar" | grep HTTP/1.1
```

Then setting it via `@plugin=conf_remap.so @pparam=proxy.config.http.cache.ignore_query=1` :

```
leif@Hel ~/ » curl -D -s -o /dev/null "http://hel.ogre.com/bench/100.bmp" | grep HTTP/1.1
HTTP/1.1 200 OK
leif@Hel ~/ » curl -D -s -o /dev/null -H "Cache-Control: only-if-cached" "http://hel.ogre.com/bench/100.bmp" | grep HTTP/1.1
HTTP/1.1 200 OK
leif@Hel ~/ » curl -D -s -o /dev/null -H "Cache-Control: only-if-cached" "http://hel.ogre.com/bench/100.bmp?foo=bar" | grep HTTP/1.1
HTTP/1.1 200 OK
```

So, clearly getting the cache hit on the request with the query params. I tested with various other query params, and they of course also hits the same cache object.

zwoop:
Benchmarking this against the cachekey plugin, the difference is surprisingly small. This option is about 10-15% faster than the plugin. But it's an simple addition, so worth it I think.

proxy/http/HttpConfig.cc
```diff
@@ -1312,6 +1312,7 @@ HttpConfig::startup()
   HttpEstablishStaticConfigLongLong(c.oride.cache_guaranteed_max_lifetime, "proxy.config.http.cache.guaranteed_max_lifetime");
 
   HttpEstablishStaticConfigLongLong(c.oride.cache_max_stale_age, "proxy.config.http.cache.max_stale_age");
+
```
ywkaras, line 1315: I've often looked at line 1314 and thought it really needed a blank line after it.

proxy/http/HttpConfig.h
```diff
@@ -618,19 +619,19 @@ struct OverridableHttpConfigParams {
   /////////////////////////////////////////////////
   MgmtByte response_suppression_mode = 0; // proxy.config.body_factory.response_suppression_mode
 
+  //////////////////
+  // Redirection  //
+  //////////////////
+  MgmtByte redirect_use_orig_cache_key = 0;
+  MgmtInt number_of_redirections       = 0;
+
```
ywkaras, line 627: It would be zwoopy of me to say you need a separate PR to move these line.

ywkaras (approved):

ezelkow1:
[approve ci centos]

proxy/http/HttpConfig.h
zwoop: @ywkaras Well played. I can move these out if you prefer, having these out of "byte order" so to speak just rubs me the wrong way since they add so much uneccesary padding to the structs.

proxy/http/HttpConfig.cc
ywkaras: This should be a 1-line PR but I'll let it slide.

PR #9708, opened 2023-05-14
Increase BUFSIZE to modern sizing

zwoop:
This BUFSIZE, in URL.cc, is used exclusively for allocating buffers on the stack. The problem with keeping it small is that URLs larger than the current 512 bytes can never use the fast path. Granted, odds are that you would be using query parameters for this, but it seems safe and generally better to increase this to 4K (with no risk).


bryancall (approved):

ezelkow1:
[approve ci centos]

PR #9346, opened 2023-01-31
Make Http3App class responsible for H3 transaction creation/deletion

maskit:

PR #9714, opened 2023-05-16
Fix H3 transaction leak

maskit:
#9346 made memory leak, assertion failures, and crashes.

### DIFF
--- a/proxy/http3/Http3Transaction.cc
+++ b/proxy/http3/Http3Transaction.cc
@@ -570,6 +570,7 @@ Http09Transaction::state_stream_open(int event, void *edata)
     if (this->_read_event == edata) {
       this->_read_event = nullptr;
     }
+    [[fallthrough]];
   case VC_EVENT_READ_COMPLETE: {
     int64_t len = this->_process_read_vio();
     // if no progress, don't need to signal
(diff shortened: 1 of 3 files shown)


proxy/http3/Http3Transaction.cc
```diff
@@ -595,11 +623,17 @@ Http09Transaction::state_stream_closed(int event, void *data)
 
   switch (event) {
   case VC_EVENT_READ_READY:
+    if (this->_read_event == data) {
+      this->_read_event = nullptr;
+    }
```
brbzull0, line 628: funny the way complier complains about `[[fallthrough]]` as here seems not to care (probably because it just `break`s)

proxy/http3/Http3Transaction.cc
```diff
@@ -208,6 +206,7 @@ HQTransaction::transaction_done()
 {
   // TODO: start closing transaction
   super::transaction_done();
+  delete this;
```
brbzull0, line 209: I am always scare about [this](https://isocpp.org/wiki/faq/freestore-mgmt#delete-this)

brbzull0 (approved):

proxy/http3/Http3Transaction.cc
maskit: I'm not a big fun of this pattern too, but this is probably the right place to delete/free/release transaction instance if we follow the design. H2 tries to free transactions here and there and it's a mess, but If you look at H1ClientTransaction, it's released here in `transaction_done()`.

PR #9715, opened 2023-05-16
Fix a potential crash due to use-after-free on QUIC connections

maskit:
Hope this fixes the crash we see on the docs server.


ezelkow1:
[approve ci centos]

maskit:
Yes, the crash seemed to be related to quiche timeout. If a connection closes in an unexpected way, an ATS event to check connection state remains active in the event queue even after QUICNetVC is freed. I don't remember how the backtrace looked like but it might have been the line which checks continuation's mutex you mentioned above.

brbzull0 (approved):
looks good to me! Thanks

PR #9716, opened 2023-05-16
Fix protocol version in request Via header

maskit:


bneradt (changes requested):
Looks reasonable.

We have a via.test.py. Can a test for this be added there?

maskit:
I added a test case for H3, and I realized HTTP version is correct but it says TCP is used. I'll try to fix that part.

maskit:
> I added a test case for H3, and I realized HTTP version is correct but it says TCP is used. I'll try to fix that part.

I didn't need to fix it. I thought there is a bug because autest passed where the gold file says QUIC uses TCP, but it was just the test is skipped. I updated the gold file.

brbzull0 (approved):
Looks good. Thanks.

ezelkow1:
[approve ci centos]

bneradt (approved):

PR #9719, opened 2023-05-18
Fix cqpv log field value on H3 connections

maskit:
`cqpv` field is a human readable protocol name such as "http/1.1" and "http/2", but "h3",  which is ALPN ID, was logged on H3 connections.


ezelkow1:
[approve ci centos]

bneradt (approved):

PR #9722, opened 2023-05-19
Fix ttmsh log field

maskit:
Log field `ttmsh` has never worked. The documentation says `ttmsh` but the implementation is `ttmh`. It's hex version of `ttms`, so `ttmsh` is correct.

https://docs.trafficserver.apache.org/en/latest/admin-guide/logging/formatting.en.html#timestamps-and-durations


bryancall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9724, opened 2023-05-21
Fix TS_HTTP_REQUEST_TRANSFORM_HOOK Tunnel Processing

bneradt:
This fixes a bug in client request processing for transform tunnels. Without this fix, the transform request tunnel was not created for requests with a body which made it so that request transform plugins were not dispatched to for the various events concerning requests with a body.

### DIFF
--- a/proxy/http/HttpSM.cc
+++ b/proxy/http/HttpSM.cc
@@ -6054,8 +6054,8 @@ HttpSM::handle_http_server_open()
 
     int method = t_state.hdr_info.server_request.method_get_wksidx();
     if (method != HTTP_WKSIDX_TRACE &&
-        server_txn->has_request_body(t_state.hdr_info.response_content_length,
-                                     t_state.server_info.transfer_encoding == HttpTransact::CHUNKED_ENCODING) &&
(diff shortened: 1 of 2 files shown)


proxy/http/HttpSM.cc
```diff
@@ -6054,8 +6054,8 @@ HttpSM::handle_http_server_open()
 
     int method = t_state.hdr_info.server_request.method_get_wksidx();
     if (method != HTTP_WKSIDX_TRACE &&
-        server_txn->has_request_body(t_state.hdr_info.response_content_length,
-                                     t_state.server_info.transfer_encoding == HttpTransact::CHUNKED_ENCODING) &&
+        server_txn->has_request_body(t_state.hdr_info.request_content_length,
+                                     t_state.client_info.transfer_encoding == HttpTransact::CHUNKED_ENCODING) &&
```
shukitchan, line 6058: Interesting. That's a typo kind of mistake here.

shukitchan (approved):

SolidWallOfCode (approved):

PR #9728, opened 2023-05-23
Make config.proxy.http.no_dns_just_forward_to_parent overridable

traeak:
This makes the setting for config.proxy.http.no_dns_just_forward_to_parent configurable per transaction.

Tested by temporarily adding debug to HttpSM::do_hostdb_lookup that dumps the no_dns_forward_to_parent value.


masaori335 (approved):
Looks good.

PR #9729, opened 2023-05-23
Fix a crash when H3 transaction ends early

maskit:
Similar to #9715 but this one is for H3 layer.

### DIFF
--- a/proxy/http3/Http3Transaction.cc
+++ b/proxy/http3/Http3Transaction.cc
@@ -324,7 +324,7 @@ HQTransaction::_schedule_write_complete_event()
 void
 HQTransaction::_unschedule_write_complete_event()
 {
-  if (this->_read_complete_event) {
+  if (this->_write_complete_event) {
     this->_write_complete_event->cancel();
     this->_write_complete_event = nullptr;
   }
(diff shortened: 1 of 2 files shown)


proxy/http3/Http3Transaction.cc
```diff
@@ -342,30 +429,16 @@ Http3Transaction::~Http3Transaction()
 }
 
 int
-Http3Transaction::state_stream_open(int event, void *edata)
+Http3Transaction::state_stream_open(int event, Event *edata)
 {
   // TODO: should check recursive call?
-  if (this->_thread != this_ethread()) {
```
maskit, line 348: Note for reviewers: This block was copied from H2 but thread mismatch should not happen. If it happens, we should fix the code that schedules events on a wrong thread.

proxy/http3/Http3Transaction.cc
```diff
@@ -234,24 +229,112 @@ HQTransaction::direction() const
...
+    return;
+  }
+
+  this->_write_complete_event = this->_thread->schedule_imm(this, VC_EVENT_WRITE_COMPLETE, &this->_write_vio);
+}
+
+void
+HQTransaction::_unschedule_write_complete_event()
+{
+  if (this->_read_complete_event) {
```
brbzull0, line 327: shouldn't we be checking `_write_complete_event`?

maskit: Yes. Thanks.

brbzull0 (approved):
looks good to me. Thanks.

Issue #5230, opened 2019-04-05
Bypass parsing HTTP request on HttpSM when request is HTTP/2


masaori335:
When Http2Stream passes HTTP request to HttpSM, it dump HTTP/1.1 request header in text from HTTPHdr. HttpSM needs to parse the text and build HTTPHdr again. I'm not sure how much this has performance overhead. But, obviously, this is not ideal. Currently, HTTP/3 impl is following same logic.

https://github.com/apache/trafficserver/blob/089316d6e2926a2bd9c9c4dfd828fe0292bfd7c0/proxy/http2/Http2Stream.cc#L166

maskit:
I think it's a blocker of H2 (and H3) to origin. Because headers sent by a client with Literal Header Field Never Indexed representation have to be transferred with the same representation. We can't retain the information on HTTP/1.1 header representation (plain text).

masaori335:
A similar overhead exists on sending responses and it's significant. ( HttpSM dump response header in a buffer and HTTP/2 parse it again ).
30% of `Http2Stream::update_write_request` is `HTTPHdr::parse_resp`.

<img width="1203" alt="Screen Shot 2020-03-17 at 16 18 42" src="https://user-images.githubusercontent.com/741391/76832014-d1457180-686b-11ea-8ed3-acec759ffa44.png">

shinrich:
@maskit I am not following your comment about the "Literal Field Never Indexed".  Can you say a few more words on this?

maskit:
It's a header field representation defined on HPACK.

https://tools.ietf.org/html/rfc7541#section-6.2.3
>    A literal header field never-indexed representation results in
   appending a header field to the decoded header list without altering
   the dynamic table.  Intermediaries MUST use the same representation
   for encoding this header field.

https://tools.ietf.org/html/rfc7541#section-7.1.3
>    An intermediary MUST NOT re-encode a value that uses the never-
   indexed literal representation with another representation that would
   index it.  If HPACK is used for re-encoding, the never-indexed
   literal representation MUST be used.

shinrich:
[...]

PR #9730, opened 2023-05-23
Add support for direct header passing

maskit:
This is for #5230. HttpSM only reads/writes HTTP request/response from/to VConnection. Although it's simple, it requires that the input/output format is text based HTTP/1.1 form. It works fine for H1, but not H2 and H3. The interface disables propagating information that cannot be transformed into HTTP/1.1 form (e.g. header sensitivity). Also, it's wasteful to marshal HTTPHdr and parse the text data again (An H2 or H3 transaction has a HTTPHdr as a result of HPACK/QPACK decoding).

This PR adds a new interface to pass/receive HTTPHdr instance to bypass header marshaling. A subclass of ProxyTransaction needs to overwrite `supports_direct_header_passing()` to use the new interface.

### DIFF
--- a/proxy/ProxyTransaction.h
+++ b/proxy/ProxyTransaction.h
@@ -56,6 +56,10 @@ public:
   virtual bool expect_receive_trailer() const;
   virtual void set_expect_receive_trailer();
 
+  virtual bool supports_direct_header_passing() const;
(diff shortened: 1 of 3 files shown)


maskit:
I forgot about transform plugins

maskit:
Found that it cannot handle 100 continue. Closing for now.

PR #9731, opened 2023-05-23
Pass HTTPHdr directly on H3 transaction

maskit:
The first commit is #9730, and the second commit is to use the new interface for H3.

This also enables POST request on H3.


maskit:
100-continue is not supported

PR #9732, opened 2023-05-23
Use Dbg() for debug output in both core and plugins.

ywkaras:
Eliminate TSDbg(), TSIsDbgCtlSet(), TSDbgCtlCreate(), TSDbgCtlDestroy().

Eliminate TSDbgCtlUniqPtr type.


ywkaras:
Doesn't yet have all the doc changes that would be necessary.

include/tscore/DbgCtl.h
```diff
@@ -25,38 +25,4 @@
...
-
-  static void _rm_reference();
-
-  class _RegistryAccessor;
-
-  friend TSDbgCtl const *TSDbgCtlCreate(char const *tag);
-
-  friend void TSDbgCtlDestroy(TSDbgCtl const *dbg_ctl);
-};
+using DbgCtl = TSDbgCtl;
```
ywkaras, line 28: Maybe just absorb this header into DiagsTypes.h

ywkaras:
After thinking about it some, I realize it would cause problems for Yahoo (and perhaps others) to change TSDbgCtl to a class.  I'd prefer to add the namespace `tsapi` for C++ stuff in the TS API.  So we'd add the class DbgCtl and the function Dbg() to this namespace.  We could also have a member function operator for DbgCtl to optionally shorted debug output calls a bit, I think.

ywkaras:
[approve ci]

ywkaras:
This sets the stage for removing and replacing slow TSDebug() calls.

ywkaras:
I will make the doc changes for this in a follow on PR that also removes TSDebug().

include/ts/DbgCtl.h
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  bool
+  on() const
+  {
+    auto m{_config_mode.load(std::memory_order_relaxed)};
+    if (!m) {
+      return false;
+    }
+    if (!_ptr->second) {
```
ywkaras, line 62: Technically `second` should be std::atomic (or at least volatile), since it can be written from another thread.  But doing that risks causing the generation of less performant code.  And, it seems unlikely that any register buffering of the value would cause actual problems.

ywkaras:
[approve ci]

zwoop (approved):
Love it. We still have work to do in many plugins though, right?

ywkaras:
> Love it. We still have work to do in many plugins though, right?

Yes

PR #9755, opened 2023-05-28
Fix OCSP detection during build

midchildan:
Backport of #9754 targeting 9.1. Looking at downstream packages listed in Repology, it's mostly packages for 9.1 and 9.2 releases that seem to be affected by the problem.

https://repology.org/projects/?search=trafficserver


bryancall:
We normally don't backport changes to branches that we no longer maintain or do releases on.  Is it necessary to have this backported to this branch?

PR #9756, opened 2023-05-28
Disable proxies when running autest

midchildan:
When HTTP proxies are enabled on the system, autest would fail to run properly because requests should never leave localhost. Now, proxies can be disabled by unsetting the relevant environment variables in the CLI. However, it's not ideal in this case because the autest script uses pipenv, which fetches Python packages online, to run the actual tests.

This change fixes that by unsetting the environment variables after pipenv is finished.


bryancall:
[approve ci]

bryancall (approved):

randall:
@midchildan Can you rebase this PR?

bneradt (approved):

PR #9757, opened 2023-05-30
Update tools script to pass shellcheck.

jpeach:
shellcheck is a linter to shell scripts. Update the scripts in the tools directory with changes suggested by shellcheck. Most of these changes are quoting variable expansions.

### DIFF
--- a/ci/coverage
+++ b/ci/coverage
@@ -16,13 +16,16 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+set -o errexit
+set -o pipefail
+
 LCOV=${LCOV:-lcov}
 GENHTML=${GENHTML:-genhtml}
 
 TMPDIR=${TMPDIR:-/tmp}
 BUILDID="org.apache.trafficserver.$$"
 
-SRCROOT=${SRCROOT:-$(cd $(dirname $0)/.. && pwd)} # where the source lives
+SRCROOT=${SRCROOT:-$(cd "$(dirname "$0")/.." && pwd)} # where the source lives
 OBJROOT=${OBJROOT:-"$TMPDIR/$BUILDID/obj"} # where we are building
 DSTROOT=${DSTROOT:-"$TMPDIR/$BUILDID/dst"} # where we are installing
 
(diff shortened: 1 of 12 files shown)


ci/coverage
```diff
@@ -16,36 +16,43 @@
...
 
 autogen() {
   (
     cd "$SRCROOT"
-    [ configure -nt configure.ac -a Makefile.in -nt Makefile.am ] || autoreconf -fi
+    if [ configure -nt configure.ac ] && [ Makefile.in -nt Makefile.am ] ; then
+        :
+    else
+        autoreconf -fi
+    fi
```
bneradt, line 47: I prefer the original.

If not the original, then at least let's reverse the boolean and not have the empty then.

tools/clang-format.sh
```diff
@@ -46,13 +46,13 @@ function main() {
...
   # Default to sha256sum, but honor the env variable just in case
-  if [ $(which sha256sum) ] ; then
+  if [ -n "$(which sha256sum)" ] ; then
     SHASUM=${SHASUM:-sha256sum}
   else
     SHASUM=${SHASUM:-shasum -a 256}
   fi
 
-  ARCHIVE=$ROOT/$(basename ${URL})
+  ARCHIVE="$ROOT/$(basename "${URL}")"
```
bneradt, line 55: Do these two nested quotes work?

tools/git/pre-commit
```diff
@@ -44,19 +49,19 @@ if [ ! -x "$FORMAT" ]; then
...
     exit 1
 fi
-source ${AUTOPEP8_VENV}/bin/activate
+source "${AUTOPEP8_VENV}/bin/activate"
 
 # Where to store the patch
 clang_patch_file=$(mktemp -t clang-format.XXXXXXXXXX)
 autopep8_patch_file=$(mktemp -t autopep8.XXXXXXXXXX)
-trap "rm -f $clang_patch_file $autopep8_patch_file" 0 1 2 3 5 15
+trap 'rm -f $clang_patch_file $autopep8_patch_file' 0 1 2 3 15
```
bneradt, line 61: This change to teh single quote will no longer expand $clang_patch_file and $autopep8_patch_file. Is that an intentional change?

bneradt (changes requested):
Nice.

Maybe we can think about including something like this in the pre-commit hooks.

ci/coverage
jpeach: This change is based on https://www.shellcheck.net/wiki/SC2166.The work happens in the else because I didn't want to invert the conditional.

I do agree that it looks a bit weird :)

tools/clang-format.sh
jpeach: Yep, because the inner quotes are in a subshell.

tools/git/pre-commit
jpeach: The shellcheck issue is:
```
In tools/git/pre-commit line 61:
trap "rm -f $clang_patch_file $autopep8_patch_file" 0 1 2 3 15
            ^---------------^ SC2064 (warning): Use single quotes, otherwise this expands now rather than when signalled.
                              ^------------------^ SC2064 (warning): Use single quotes, otherwise this expands now rather than when signalled.

For more information:
  https://www.shellcheck.net/wiki/SC2064 -- Use single quotes, otherwise this...
```

In this case, I think that expanding these variable when the trap is called is OK,since their values arenever changed.

bryancall:
@jpeach Can you please resolve the conflicts.

Issue #9750, opened 2023-05-26
trafficserver 9.1.1 traffic_server ERROR: plugin 'regex_remap.so' failed to load:


volgaleoking:
After migrating the configuration from the 8.0 server, the error module error appears at startup regex_remap.so , with the following lines from the log file:

[May 26 15:59:28.658] traffic_server ERROR: plugin 'regex_remap.so' failed to load: /run/trafficserver/d176adad-6a1c-4ec6-85cc-86b9696f7129/usr/lib/trafficserver/modules/regex_remap.so: failed to map segment from shared object
[May 26 15:59:28.658] traffic_server ERROR: [ReverseProxy] failed to add remap rule at /etc/trafficserver/remap.config line 202: /run/trafficserver/d176adad-6a1c-4ec6-85cc-86b9696f7129/usr/lib/trafficserver/modules/regex_remap.so: failed to map segment from shared objectfailed to remove runtime copy: Success
[May 26 15:59:28.658] traffic_server WARNING: something failed during BuildTable() -- check your remap plugins!
[May 26 15:59:28.658] traffic_server FATAL: remap.config failed to load

remap.config:
map http://static.noname.com/v/vh/vod_hds/definst/amlst http://192.168.25.3:1935/ibms3/definst/amlst @plugin=regex_remap.so @pparam=maps.reg

maps.reg:
^/v/vh/vod_hds/definst/amlst(.*)\.(.*)$ http://192.168.25.3:1935/ibms3/definst/amlst$1.$2

how to solve this problem, despite the fact that there were no such problems in either version 6 or 8.

randall:
Try adding 
```
CONFIG proxy.config.plugin.dynamic_reload_mode INT 0
```

to `records.config`

volgaleoking:
Thank you so much for your help, I spent several hours not understanding what the problem is. The only thing that is not clear is why in the new configuration this parameter is not 0 by default, since the plugins cannot be dynamically reloaded?
After all, it is logical to set the default settings to working states, or I don’t understand something!

bryancall:
@randall Do you know why this is happening?

bryancall:
This was resolved by disabling `proxy.config.plugin.dynamic_reload_mode` (set to `0`). The dynamic reload feature introduced in ATS 9.x copies plugins to a temporary directory before loading, which caused "failed to map segment from shared object" errors in some environments. The workaround of disabling dynamic reload resolved the issue.

[...]

PR #9760, opened 2023-05-31
Improve PluginDso error formatting.

jpeach:
Add a trivial helper to make sure that PluginDso always separates appended error messages with a colon and space.


jpeach:
You can observe a malformed error message in #9750

PR #9762, opened 2023-05-31
Fix for strategies abort where ParentResult rec isn't being set

traeak:
This abort showed up during testing with a somewhat complex configuration.  The issue shows up due to the dispatch code in HttpTransact's parent functions and compatibility layer with parent.config handling classes.


traeak:
lost how to recreate this.

PR #9765, opened 2023-06-01
UDP/QUIC: Make sure quiche gets the entire rx buffer from the udp packet.

brbzull0:
It seems that it was only getting the first slice of the entire chain. Currently each of the iov buffer size is hardcoded to 2k, if happens that we need to fit more than 2k inside the buffer the current code will only hand over the first part of the chain which will make QUICHE to complain as the data is incomplete.

This change makes sure that quiche gets the entire buffer by building the entire buffer together.

Issue can be reproduced by including the ` --max-udp-payload-size 3k` in the `h2load` tool. Actually, any size above 2k should hit the issue.
example:
```bash
/opt/bin/h2load -n 1 -c 1 --npn-list=h3 -H="my-hn1: test01" https://<server-ip>:4443/cache/1000 --max-udp-payload-size 3k
```

FYI: For now this seems fine but as soon as we introduce GRO this will more likely to change.

### DIFF
--- a/iocore/net/UnixUDPNet.cc
+++ b/iocore/net/UnixUDPNet.cc
@@ -134,9 +134,40 @@ UDPPacket::free()
   if (p.conn)
     p.conn->Release();
   p.conn = nullptr;
+
+  if (this->_payload) {
+    _payload.release();
+  }
   udpPacketAllocator.free(this);
 }
 
+uint8_t *
+UDPPacket::get_entire_chain_buffer(size_t *buf_len)
+{
+  if (this->_payload == nullptr) {
+    IOBufferBlock *block = this->getIOBlockChain();
+
+    // No need to allocate, we will use the first slice. With the current slice size
+    // 2048 more likely we will using this block most of the time.
+    if (block && block->next.get() == nullptr) {
+      *buf_len = this->getPktLength();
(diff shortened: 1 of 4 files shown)


iocore/net/UnixUDPNet.cc
```diff
@@ -134,9 +134,40 @@ UDPPacket::free()
...
+    IOBufferBlock *block = this->getIOBlockChain();
+
+    // No need to allocate, we will use the first slice. With the current slice size
+    // 2048 more likely we will using this block most of the time.
+    if (block && block->next.get() == nullptr) {
+      *buf_len = this->getPktLength();
+      return reinterpret_cast<uint8_t *>(block->buf());
+    }
+
+    // Store it. Should we try to avoid allocating here?
```
maskit, line 157: > Should we try to avoid allocating here?

Yes, it's very unfortunate to alloc another buffer and copy data. But as we talked offline, let's fix the bug for now and improve it with a GRO friendly design.

maskit (approved):

brbzull0:
[approve ci Ubuntu]

brbzull0:
[approve ci ubuntu]

brbzull0:
[approve ci ubuntu]

PR #9770, opened 2023-06-03
Break the circular dependency between iocore and tscore.

### LINKED ISSUE #9768: Circular dependency between iocore and tscore
If you try to run tests without doing a full build first, the build fails because of ordering issues:

```
$ make -j4 check
...
make[3]: Entering directory '/home/jpeach/src/trafficserver/src/tscore'
...
make[3]: *** No rule to make target '../../iocore/eventsystem/libinkevent.a', needed by 'test_tscore'.  Stop.
```

However, it's not possible to re-order the build so that this works, because iocore and tscore depend on each other.

In tscore, the `MTHashTable` class uses `ProxyMutex` from iocore for locking. `ProxyMutex` pulls in `EThread` and all sorts of other iocore symbols. iocore depends on tscore all over the place, which seems reasonable and intended.

AFAICT, `MTHashTable` is only used in one place (QUIC support in iocore, see https://github.com/apache/trafficserver/issues/2478).

jpeach:
The re-introduction of the MTHashTable class created a dependency from tscore to iocore, since it uses proxy mutexes, which are tightly coupled to the event system. We can break this circular dependency by moving the class into the QUIC library as an implementation detail, since that already depends on the event system

This fixes #9768.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -108,7 +108,7 @@ lib/perl/lib/Apache/TS.pm
 iocore/net/test_certlookup
 iocore/net/test_UDPNet
 iocore/net/test_libinknet
-iocore/net/quic/test_QUIC*
+iocore/net/quic/test_*
 iocore/aio/test_AIO
 iocore/eventsystem/test_IOBuffer
 iocore/eventsystem/test_EventSystem
--- a/iocore/net/quic/MTHashTable.h
+++ b/iocore/net/quic/MTHashTable.h
@@ -23,7 +23,7 @@
 
 /****************************************************************************
 
-  MT_hashtable.h
+  MTHashTable.h
 
   Multithread Safe Hash table implementation
 
--- a/iocore/net/quic/Makefile.am
+++ b/iocore/net/quic/Makefile.am
@@ -54,6 +54,20 @@ libquic_a_SOURCES = \
 # Check Programs
 #
 
+check_PROGRAMS := test_MTHashTable
+
+test_MTHashTable_CPPFLAGS = \
+	$(AM_CPPFLAGS) \
+	-I$(abs_top_srcdir)/tests/include
+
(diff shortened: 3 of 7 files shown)


jpeach:
```
     Setting up : Copying '/home/jenkins/workspace/Github_Builds/autest/src/tests/gold_tests/pluginTest/polite_hook_wait/.libs/polite_hook_wait.so' to '/tmp/sandbox/polite_hook_wait/ts/plugin' - Passed
        Reason: Returned Value -9 != 0
        Reason: Returned value: 10.503296613693237 < 600.0
```

Seems like a CI problem?

jpeach:
Wait, the copy isn't the error, I guess it's a return code check from this somewhere?

```
tr = Test.AddTestRun()
tr.Processes.Default.StartBefore(server)
tr.Processes.Default.StartBefore(ts)
tr.Processes.Default.Command = (
    'curl --verbose --ipv4 --header "Host:myhost.test" http://localhost:{}/ 2>curl.txt'.format(ts.Variables.port)
)

tr = Test.AddTestRun()
tr.Processes.Default.Command = (
    'curl --verbose --ipv4 --header "Host:myhost.test" http://localhost:{}/ 2>curl.txt'.format(ts.Variables.port)
)

tr = Test.AddTestRun()
tr.Processes.Default.Command = "grep -F HTTP/ curl.txt"
tr.Processes.Default.Streams.stdout = "curl.gold"
```

But it's not clear which one fails or why. Exit code 9 from curl is " FTP access denied", so maybe it's the grep that fails? If there's an expected output file, I wonder why the test runs grep manually?

jpeach:
[approve ci autest]

cmcfarlen (approved):
fantastic!

maskit (approved):

randall (approved):

PR #9771, opened 2023-06-05
tools: use a private directory to build H3 dependencies

jpeach:
Build H3 dependencies using a private work directory. This prevents dependencies being splatted into the current directory if the build is interrupted or fails.


bneradt (approved):
Looks good. I use this script regularly and think this change would be helpful.

PR #9616, opened 2023-04-16
Add support for CONNECT method on HTTP/2 connection

maskit:
This blog post motivated me to implement CONNECT method support.
https://daniel.haxx.se/blog/2023/04/14/curl-speaks-http-2-with-proxy/

```
$ ~/opt/curl/bin/curl -vk --proxy-insecure --proxy-http2 -x https://localhost:8443/ https://www.apache.org/ > /dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying 127.0.0.1:8443...
* Connected to localhost (127.0.0.1) port 8443 (#0)
* ALPN: offers h2,http/1.1
} [5 bytes data]
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
} [512 bytes data]
* TLSv1.2 (IN), TLS handshake, Server hello (2):
{ [122 bytes data]
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
} [1 bytes data]
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
{ [15 bytes data]
* TLSv1.3 (IN), TLS handshake, Certificate (11):
{ [1444 bytes data]
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
{ [520 bytes data]
* TLSv1.3 (IN), TLS handshake, Finished (20):
{ [36 bytes data]
* TLSv1.3 (OUT), TLS handshake, Finished (20):
} [36 bytes data]
* SSL connection using TLSv1.3 / TLS_AES_128_GCM_SHA256
* ALPN: server accepted h2
* Proxy certificate:
*  subject: C=AU; ST=Some-State; O=Internet Widgits Pty Ltd; CN=localhost
*  start date: Apr 16 01:33:15 2023 GMT
*  expire date: Apr 15 01:33:15 2024 GMT
*  issuer: C=AU; ST=Some-State; O=Internet Widgits Pty Ltd; CN=localhost
*  SSL certificate verify result: self signed certificate (18), continuing anyway.
* CONNECT tunnel: HTTP/2 negotiated
* Establish HTTP/2 proxy tunnel to www.apache.org:443
} [5 bytes data]
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
{ [214 bytes data]
[...]

PR #9781, opened 2023-06-06
Added Autest for H2 CONNECT and fix a crash triggered by the test

lzx404243:
In #9616, @maskit wrote an H2 CONNECT Autest but couldn't include that in the final PR because of a Proxy Verifier issue. 
Now that the Proxy Verifier issue is resolved,  the Autest is added in this PR(with a few tweaks).

ATS crashes with the new test executing HTTP/2 tunneling traffic. This PR also includes a fix to resolve that.

### DIFF
--- a/proxy/http/HttpSM.cc
+++ b/proxy/http/HttpSM.cc
@@ -4119,6 +4119,11 @@ HttpSM::tunnel_handler_ssl_producer(int event, HttpTunnelProducer *p)
   STATE_ENTER(&HttpSM::tunnel_handler_ssl_producer, event);
 
   switch (event) {
+  case VC_EVENT_READ_READY:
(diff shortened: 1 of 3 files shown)


lzx404243:
Note: The updated `connect` Autest is dependent on a newer release of Proxy Verifier. It would fail before the Proxy Verifier version is updated in ATS testing.

bneradt:
> Note: The updated `connect` Autest is dependent on a newer release of Proxy Verifier. It would fail before the Proxy Verifier version is updated in ATS testing.

Just to record it, Proxy Verifier was updated to the required v2.8.1 via: https://github.com/apache/trafficserver/pull/9834

moonchen (approved):

PR #9792, opened 2023-06-07
Avoid crash on traffic_server shutdown

maskit:


maskit:
[approve ci]

bneradt (approved):
I'm glad you found this.

The crash is occasioned by some updated system package. The old CI image worked fine, but if you do a `sudo yum update` on the old image, the certifier test starts crashing again.

I suggest we:

1. Pull in this patch so the crash stops for now.
2. Update the CI image with the newer packages.
3. File an issue for the crash and push to get it assigned and worked on.

PR #9795, opened 2023-06-07
Add a shutdown handler to the certifier plugin

### LINKED ISSUE #9794: certifier plugin crash on master
I'm a part of the team that manages the ASF ATS CI images. I did a routine update of the ci.trafficserver.apache.org/ats/rockylinux:8 image to make it slimmer by cleaning up build artifacts. We noticed, however, that with that update, the certifier.test.py test started failing due to what looked like a shutdown crash from ATS. Through some investigation, I observed:

1. Start with the previous ci.trafficserver.apache.org/ats/rockylinux:8 docker image and run the certifier autest, it passes fine. The original image is available on docker1 in the CI system, which I have access to.
3. Now run `sudo yum update` and update the system packages.
4. Then rebuild ATS after a full git clean.
5. Then re-run the certifier autest. The test now fails due to the ATS crash.

@maskit and @bryancall both reproduced this crash on later Fedora versions, so this isn't a peculiarity of the docker image itself. It's just occasioned by some package update.

@maskit found that the crash could be avoided by commenting out the debug log from a destructor in the certifier.cc file, see: https://github.com/apache/trafficserver/pull/9792

We agreed to merge that in temporarily to keep CI green so that PRs can be merged. However, we need to prioritize figuring out why that log results in a crash with that certifier test.

bneradt:
This adds a shutdown handler to the certifier plugin so that its SSL data can be cleaned up in a callback rather than on process shutdown.

This also reverts commenting out the debug call from SslData which was commented out as a workaround to a crash which was happening on shutdown.

Fixes #9794

### DIFF
--- a/plugins/certifier/certifier.cc
+++ b/plugins/certifier/certifier.cc
@@ -102,9 +102,7 @@ private:
     SslData *next = nullptr;
 
     SslData() = default;
-    ~SslData()
-    { /* TSDebug(PLUGIN_NAME, "Deleting ssl data for [%s]", commonName.c_str()); */
-    }
+    ~SslData() { TSDebug(PLUGIN_NAME, "Deleting ssl data for [%s]", commonName.c_str()); }
   };
 
   using scoped_SslData = std::unique_ptr<SslLRUList::SslData>;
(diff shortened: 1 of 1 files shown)


bneradt:
Rather than adding a shutdown handler, we'll address this by using the TSDbg interface which is safer in shutdown scenarios: #9809.

PR #9809, opened 2023-06-08
certifier.so: Convert to TSDbg

### LINKED ISSUE #9794: certifier plugin crash on master

bneradt:
The TSDbg works better in shutdown conditions. The certifier plugin with the old logging interface was experiencing a shutdown crash. Converting to TSDbg addresses this crash.

Fixes #9794

### DIFF
--- a/plugins/certifier/certifier.cc
+++ b/plugins/certifier/certifier.cc
@@ -46,6 +46,7 @@
 #include "ts/ts.h"
 
 const char *PLUGIN_NAME = "certifier";
+auto dbg_ctl            = TSDbgCtlCreate(PLUGIN_NAME);
 
 /// Override default delete for unique ptrs to openSSL objects
 namespace std
@@ -102,9 +103,7 @@ private:
     SslData *next = nullptr;
 
     SslData() = default;
-    ~SslData()
-    { /* TSDebug(PLUGIN_NAME, "Deleting ssl data for [%s]", commonName.c_str()); */
-    }
+    ~SslData() { TSDbg(dbg_ctl, "Deleting ssl data for [%s]", commonName.c_str()); }
   };
 
   using scoped_SslData = std::unique_ptr<SslLRUList::SslData>;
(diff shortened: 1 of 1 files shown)


bryancall:
[approve ci autest]

ywkaras (changes requested):
We should not merge this until we decide what to do with:  https://github.com/apache/trafficserver/pull/9732

bneradt:
[approve ci rat]

jpeach:
Might as well take all the changes from #9975 as well.

jpeach (approved):

bneradt:
The certifier plugin has been converted to DbgCtl via #10390:
https://github.com/apache/trafficserver/pull/10390/files#diff-14664431da891cb710c866874630968fad88531ac96f9e6e7b2558ee13988f7e

Closing this PR.

PR #9811, opened 2023-06-08
Add support for hwloc library to CMake build

JosiahWI:
The library will be used automatically if it is found on the system. If the library was found, a status message should be printed with the path to the library.


cmcfarlen (approved):
🏆 I like the style of the find package creating the interface lib target.  Thanks so much!

PR #9813, opened 2023-06-08
Untangle some classes in iocore

moonchen:
- Use dynamic dispatch for EventIO
- Isolate the following classes into their own files 
  - NetEvent
  - NetHandler 
  - NetVCOptions 
  - PollCont 
  - AcceptOptions
- Fix some includes other files to compile cleanly


masaori335:
"Use dynamic dispatch for EventIO" sounds cool, but would you separate cleanup into other PR? We easily miss something with this size of change.

PR #9817, opened 2023-06-09
Resolve circular dep in esi components

JosiahWI:
Some files in esi/fetcher depended on files in esi/lib and vice-versa. I was able to move the shared dependencies into esi/common, and now the dependency hierarchy is a little cleaner. This made it easy to add the esi plugin and its tests to the CMake build, so this refactoring is a precursor to bringing in those changes as well.


shukitchan (approved):
Looks ok. 
So now I think it is 
lib -> common & fetcher
fetcher -> common

JosiahWI:
Some of the header paths in the Makefile weren't updated properly.

ezelkow1 (approved):

shukitchan (approved):

PR #9818, opened 2023-06-09
Deprecated APIs removal

zwoop:


traeak (approved):
These look fine to me.

shukitchan (changes requested):
Please update the documents for lua plugin here as well.

https://docs.trafficserver.apache.org/en/latest/admin-guide/plugins/lua.en.html#ts-http-client-packet-tos-set
https://docs.trafficserver.apache.org/en/latest/admin-guide/plugins/lua.en.html#ts-http-server-packet-tos-set

shukitchan (approved):

PR #9822, opened 2023-06-09
Add esi plugin and tests to CMake build

JosiahWI:
This compiles the esi plugin and registers all its tests with CTest. The functionality of the installed plugin cannot be easily verified because autest does not work with the CMake install yet.


JosiahWI:
The segfault was caused by the asserts being omitted from the build in Release mode. Only this test was heavily impacted, but all the esi tests will need to be updated, in a separate PR.

cmcfarlen (approved):

PR #9825, opened 2023-06-12
Clean up UnixNetProcessor entanglements.

jpeach:
The NetProcessor class sometimes downcasts itself to a UnixNetProcessor because it implicitly knows that it is a singleton and that UnixNetProcessor is the actual implementation. We can remove this oddity by simply making the relevant NetProcessor operations abstract and moving the implementations to UnixNetProcessor.

While we are doing this, we can remove some unused member variables, make createNetAccept protected (only subclasses of UnixNetProcessor should call it), remove unnecessary casting, and add a missing lock to a naVec traversal.

### DIFF
--- a/iocore/net/UnixNetProcessor.cc
+++ b/iocore/net/UnixNetProcessor.cc
@@ -30,7 +30,23 @@
 // For Stat Pages
 #include "StatPages.h"
 
-int net_accept_number = 0;
+// naVecMutext protects access to naVec.
+Ptr<ProxyMutex> naVecMutex;
+
+std::vector<NetAccept *> naVec;
+
+unsigned int
+net_next_connection_number()
+{
+  static int net_connection_number = 1;
+
+  unsigned int res = 0;
+  do {
+    res = ink_atomic_increment(&net_connection_number, 1);
+  } while (!res);
+  return res;
+}
+
 NetProcessor::AcceptOptions const NetProcessor::DEFAULT_ACCEPT_OPTIONS;
 
 NetProcessor::AcceptOptions &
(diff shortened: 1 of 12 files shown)


jpeach:
[ci autest]

jpeach:
[approve ci autest]

ywkaras:
This may incur the wrath of @zwoop .

jpeach:
> This may incur the wrath of @zwoop .

Added him to reviewers.

iocore/net/UnixNetProcessor.cc
```diff
@@ -30,7 +30,19 @@
 // For Stat Pages
 #include "StatPages.h"
 
-int net_accept_number = 0;
+static int net_accept_number     = 0;
+static int net_connection_number = 1;
```
ywkaras, line 34: Why not an anonymous namespace?

ywkaras (dismissed):

iocore/net/UnixNetProcessor.cc
```diff
@@ -259,19 +260,11 @@ UnixNetProcessor::connect_re_internal(Continuation *cont, sockaddr const *target
...
-}
-
-struct PollCont;
-
 // This needs to be called before the ET_NET threads are started.
 void
 UnixNetProcessor::init()
 {
-  EventType etype = ET_NET;
+  naVecMutex = new_ProxyMutex();
```
moonchen, line 267: It's not intuitive to me that the NetProcessor should be responsible for initializing this mutex, which is defined in UnixNetAccept.cc.  I recommending leaving the initialization in `main` as before.

Alternatively, if you move `naVec` and `naVecMutex` inside of `UnixNetProcessor`, then this would be a good place to initialize `naVecMutex`.

iocore/net/UnixNetProcessor.cc
```diff
@@ -56,34 +68,21 @@ NetProcessor::AcceptOptions::reset()
...
-
 Action *
-NetProcessor::accept(Continuation *cont, AcceptOptions const &opt)
+UnixNetProcessor::accept(Continuation *cont, AcceptOptions const &opt)
 {
   Debug("iocore_net_processor", "NetProcessor::accept - port %d,recv_bufsize %d, send_bufsize %d, sockopt 0x%0x", opt.local_port,
         opt.recv_bufsize, opt.send_bufsize, opt.sockopt_flags);
 
-  return ((UnixNetProcessor *)this)->accept_internal(cont, NO_FD, opt);
+  return this->accept_internal(cont, NO_FD, opt);
```
moonchen, line 77: `this->` is unnecessary.

iocore/net/UnixNetProcessor.cc
```diff
@@ -56,34 +68,21 @@ NetProcessor::AcceptOptions::reset()
...
 
 Action *
-NetProcessor::main_accept(Continuation *cont, SOCKET fd, AcceptOptions const &opt)
+UnixNetProcessor::main_accept(Continuation *cont, SOCKET fd, AcceptOptions const &opt)
 {
-  UnixNetProcessor *this_unp = static_cast<UnixNetProcessor *>(this);
   Debug("iocore_net_processor", "NetProcessor::main_accept - port %d,recv_bufsize %d, send_bufsize %d, sockopt 0x%0x",
         opt.local_port, opt.recv_bufsize, opt.send_bufsize, opt.sockopt_flags);
-  return this_unp->accept_internal(cont, fd, opt);
+  return this->accept_internal(cont, fd, opt);
```
moonchen, line 85: ditto

iocore/net/UnixNetProcessor.cc
```diff
@@ -30,7 +30,19 @@
 // For Stat Pages
 #include "StatPages.h"
 
-int net_accept_number = 0;
+static int net_accept_number     = 0;
```
moonchen, line 33: It's not a big deal, but I think a good place for these variables is as static variables inside the functions that they're used.  The reason is that they're only used in those functions, and not meant to be touched anywhere else.

moonchen (changes requested):

iocore/net/UnixNetProcessor.cc
jpeach: > 



> It's not intuitive to me that the NetProcessor should be responsible for initializing this mutex, which is defined in UnixNetAccept.cc. I recommending leaving the initialization in `main` as before.

I agree that it's not great, but it is better than doing it in traffic_server.cc, whi8ch has no business at all knowing about how internal iocore variables need to be initialized. UnixNetProcessor has some reason to know this since it is the one that knows how to create UnixNetAccept. I'll think about how to clean this up in a separate PR.

> Alternatively, if you move `naVec` and `naVecMutex` inside of `UnixNetProcessor`, then this would be a good place to initialize `naVecMutex`.

OK, I don't mind doing that, since UnixNetProcessor actually populated the vector.

iocore/net/UnixNetProcessor.cc
jpeach: It makes me sad to do that, since `this->` improves readability so much, but I agree with you that it is consistent to remove it.

iocore/net/UnixNetProcessor.cc
jpeach: Changes to local static.

jpeach:
[approve ci autest]

jpeach:
[approve ci]

moonchen (approved):

PR #9826, opened 2023-06-12
Remove TS_HAS_LIBZ

JosiahWI:
Zlib is required to compile ATS. We don't need to keep this option around.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,6 +99,7 @@ endif()
 find_package(PCRE)
 include(FindOpenSSL)
 find_package(OpenSSL)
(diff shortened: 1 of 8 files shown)


cmcfarlen (approved):

cmcfarlen:
are we sure this doesn't break builds without zlib that would have worked with the CPP conditionals?

JosiahWI:
I just tried building without it, and autotools gave me an error that it could not find zlib.

JosiahWI:
[approve ci autest]

cmcfarlen (approved):

PR #9831, opened 2023-06-12
Allow slice plugin to purge requests

serrislew:
Allow purge requests when slice plugin is enabled
- respects `ref-relative` config, purging ref block (slice 0) if processed
    - autests check both behaviors
- will stop purging remaining blocks when 404 (not found in cache) is received


traeak (approved):
Changes look fine and they pass some cursory hand tests.

bryancall:
Cherry picked to the 9.2.x branch

PR #9832, opened 2023-06-12
Add support for TCMalloc in CMake build

JosiahWI:
This abstracts some of the logic for finding a malloc replacement so that either jemalloc or TCMalloc can be used, and mimalloc support added easily at a later point. It also fixes some incorrect generator expression usage. If jemalloc and TCMalloc are both found, an error will be issued. It looks like this is what the autotools build does.


bneradt:
[approve ci autest]

cmcfarlen (approved):

PR #9836, opened 2023-06-13
Rename h3 settings parameter name

### LINKED ISSUE #9426: Enforce H3 header size constraints
Need to support `SETTINGS_MAX_FIELD_SECTION_SIZE`.
https://datatracker.ietf.org/doc/html/rfc9114#name-header-size-constraints

There are similar thing in H2 as well and we have `proxy.config.http2.max_header_list_size` for it.
https://datatracker.ietf.org/doc/html/rfc9113#name-limits-on-field-block-size

Note that the size is not the total of the lengths of field names and values.
> The size of a field list is calculated based on the uncompressed size of fields, including the length of the name and value in bytes plus an overhead of 32 bytes for each field.

duke8253:
this should resolve #9426

### DIFF
--- a/proxy/http3/Http3.cc
+++ b/proxy/http3/Http3.cc
@@ -24,10 +24,10 @@
 #include "Http3.h"
 
 // Default values of settings defined by specs (draft-17)
-const uint32_t HTTP3_DEFAULT_HEADER_TABLE_SIZE     = 0;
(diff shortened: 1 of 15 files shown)


brbzull0:
[approve ci autest]

maskit (approved):

PR #9838, opened 2023-06-13
Untangle some classes in iocore

moonchen:
- Isolate the following classes into their own files
  - NetEvent
  - NetHandler
  - NetVCOptions
  - PollCont
  - AcceptOptions
- Fix some includes other files to compile cleanly


iocore/net/EventIO.cc
```diff
@@ -0,0 +1,228 @@
...
+int
+EventIO::close()
+{
+  if (!this->syscall) {
+    return 0;
+  }
+
+  stop();
+  switch (type) {
+  default:
```
maskit, line 214: Curious why `default` is on the top.

iocore/net/Makefile.am
```diff
@@ -122,12 +122,16 @@ test_libinknet_LDADD = \
 	@HWLOC_LIBS@ @OPENSSL_LIBS@ @LIBPCRE@ @YAMLCPP_LIBS@ @SWOC_LIBS@
 
 libinknet_a_SOURCES = \
+  AcceptOptions.cc \
```
maskit, line 125: Wrong indent width

iocore/net/EventIO.cc
moonchen: Not sure why, but the code was moved from here: https://github.com/apache/trafficserver/blob/9d3bda4cb241f5ee574753506e970d383ee4d495/iocore/net/P_UnixNet.h#L622

iocore/net/Makefile.am
moonchen: Thanks!  Fixed.  Turns out that file had mixed spaces and tabs.

iocore/net/EventIO.cc
maskit: Ok, I was just curious. I don't think we should change it on this PR then, because we could accidentally break something.

moonchen:
Rebased

ywkaras (approved):

PR #9839, opened 2023-06-13
Add mimalloc support to CMake build

JosiahWI:
We don't have a feature flag for mimalloc, but we do support it in the autotools build. This updates the CMake build to reflect that.


cmcfarlen (approved):
I wonder if failing the build when multiple mallocs are found will be annoying and we should instead just have a default preference with an option to pick?

JosiahWI:
Options are a good idea, although I doubt anyone will run into this problem other than explicitly testing malloc support. It was hard enough to get tcmalloc and mimalloc installed that I don't think it will happen accidentally.

cmcfarlen (approved):

PR #9841, opened 2023-06-14
Search for and enable backtrace in CMake build

JosiahWI:
This makes sure that backtrace is enabled if it is available from the OS.


JosiahWI:
[approve ci debian]

cmcfarlen (approved):

jpeach:
@cmcfarlen Is this change correct? Looking at the automake, it's checking to see whether `execinfo.h` is present to set the `has_backtrace` symbol. On FreeBSD, it also checks for `libexecinfo`.

I googled the make docs, and I think that this should be using [FindBacktrace](https://cmake.org/cmake/help/latest/module/FindBacktrace.html)
```
FindBacktrace()
if (Backtrace_FOUND)
  FindBacktrace
  -- do something with the output variables here
endif()
```

PR #9842, opened 2023-06-14
Search for and link profiler in CMake build

JosiahWI:
This looks for the gperftools profiler and links it if found. It also puts both profiler and TCMalloc targets under the gperftools namespace.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,7 @@ execute_process(COMMAND id -ng OUTPUT_VARIABLE BUILD_GROUP OUTPUT_STRIP_TRAILING
 execute_process(COMMAND uname -n OUTPUT_VARIABLE BUILD_MACHINE OUTPUT_STRIP_TRAILING_WHITESPACE)
 
 # Options
+option(ENABLE_PROFILER "Use gperftools profiler (default OFF)")
 set(DEFAULT_STACK_SIZE 1048576 CACHE STRING "Default stack size (default 1048576)")
 set(TS_MAX_HOST_NAME_LEN 256 CACHE STRING "Max host name length (default 256)")
 set(TS_USE_SET_RBIO 1 CACHE STRING "Use openssl set_rbio (default 1)")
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -88,25 +88,36 @@ CHECK_INCLUDE_FILE(arpa/nameser_compat.h HAVE_ARPA_NAMESER_COMPAT_H)
...
+
 find_package(PCRE)
+
 include(FindOpenSSL)
 find_package(OpenSSL)
+
+find_package(profiler)
+if(profiler_FOUND)
+    set(TS_HAS_PROFILER TRUE)
+endif()
```
cmcfarlen, line 114: We should add an option for looking for the profiler as we don't always want to profile.  The auto tools build has `--with-profiler` for this.

cmcfarlen: Maybe just `PROFILE` as an option and default to off.

cmcfarlen (dismissed):
Please add an option for enabling profiling

CMakeLists.txt
JosiahWI: Maybe `ENABLE_PROFILER` would be better? The autotools build uses `--enable-xxx` for enabling/disabling features, and I think this shows the intent well. Note that the autotools build has no way to explicitly disable the profiler that I could find.

CMakeLists.txt
```diff
@@ -108,8 +109,8 @@ find_package(PCRE)
 include(FindOpenSSL)
 find_package(OpenSSL)
 
-find_package(profiler)
-if(profiler_FOUND)
+if(ENABLE_PROFILER)
+    find_package(profiler REQUIRED)
     set(TS_HAS_PROFILER TRUE)
```
cmcfarlen, line 114: `set(TS_HAS_PROFILER ${profiler_FOUND})`

cmcfarlen (changes requested):

cmcfarlen (approved):

PR #9844, opened 2023-06-14
Fix misuse of unique_ptr.release.

brbzull0:
The idea was to release the used memory not to release ownership which would create a leak.


bneradt (commented):
I won't approve since we're both with Yahoo, but looks good to me.

I reviewed all other uses of `release` in our code and they seem all related to configProcessor, which is unrelated.

randall (approved):

maskit (approved):

PR #9849, opened 2023-06-15
Remove unneeded DEBUG conditionals.

jpeach:
ink_assert is already compiled out in non-debug builds, so we don't need to protect assertions with "#if DEBUG". Since we are touching this file, make the type of "cancelled" bool, since it is already treated as if it was.


brbzull0 (approved):
Looks good. Thanks.

Toppole6 (commented):
[]()

PR #9851, opened 2023-06-15
Check for 128bit CAS in CMake build

JosiahWI:
This adds a script that sets the following variables: TS_HAS_128BIT_CAS, TS_NEEDS_MCX16_FOR_CAS. The source code for the compile check was taken from the autotools build with some minor modifications (removed void arg to main, removed extraneous semicolon, brought main body back to 1 level of indentation).

Testing Note: I checked the behavior of both paths through the module by simulating the first check failing, but I cannot verify that the first check will actually fail.


cmcfarlen (approved):
I'll approve this but I wonder if this is necessary at all.  Are there any compilers or architectures that don't support this or is there even a reason to make a distinction anymore?

PR #9854, opened 2023-06-15
Check for SO_MARK in CMake build

JosiahWI:
This sets the CMake variable TS_HAS_SO_MARK appropriately.


JosiahWI:
[approve ci cmake fedora]

JosiahWI:
[approve ci cmake]

cmcfarlen (commented):
This seem unnecessary.  Maybe we should just use `#if SO_MARK` instead of making a new definition for this?  afaict this is used in one place in the code.

JosiahWI:
I think it's a good idea to look into which options we don't need anymore, but the autotools build supports this option and it's documented as being available. Maybe we should do that separately?

JosiahWI:
Rebased on master.

cmcfarlen (approved):

PR #9862, opened 2023-06-16
Add support for libunwind in CMake build

JosiahWI:
This adds an option that defaults to ON: ENABLE_UNWIND. On Linux, CMake will look for the library and link it globally if it is found. On other OSs, CMake will skip the check, because libunwind is only available on Linux.


JosiahWI:
[approve ci autest]

cmcfarlen (changes requested):
Should this new import target be linked anywhere?

JosiahWI:
It's being linked globally. The header isn't included anywhere specifically in ATS.

cmcfarlen (approved):

PR #9863, opened 2023-06-16
Add option to build regression tests

JosiahWI:
This adds a new option that defaults to ON: BUILD_REGRESSION_TESTING. Because the regression tests are part of the traffic_server binary, this is independent from the BUILD_TESTING option added by CTest, which builds separate test binaries.


cmcfarlen (approved):

bneradt (approved):
Fantastic. I'll make a note to run the regression tests now in the jenkins cmake CI jobs.

PR #9793, opened 2023-06-07
Update HTTP/3 error codes

duke8253:
This should resolve #9427

PR #9864, opened 2023-06-16
Enforce SETTINGS frame rules

### LINKED ISSUE #9428: Prohibit a second SETTINGS frame
We have `max_settings_per_frame`, `max_settings_per_minute`, and `max_settings_frames_per_minute` for H2, but sending multiple SETTINGS frame on H3 is not allowed.

> SETTINGS frames always apply to an entire HTTP/3 connection, never a single stream. A SETTINGS frame MUST be sent as the first frame of each [control stream](https://datatracker.ietf.org/doc/html/rfc9114#control-streams) (see [Section 6.2.1](https://datatracker.ietf.org/doc/html/rfc9114#control-streams)) by each peer, and it MUST NOT be sent subsequently. If an endpoint receives a second SETTINGS frame on the [control stream](https://datatracker.ietf.org/doc/html/rfc9114#control-streams), the endpoint MUST respond with a [connection error](https://datatracker.ietf.org/doc/html/rfc9114#errors) of type [H3_FRAME_UNEXPECTED](https://datatracker.ietf.org/doc/html/rfc9114#H3_FRAME_UNEXPECTED).[¶](https://datatracker.ietf.org/doc/html/rfc9114#section-7.2.4-2)

https://datatracker.ietf.org/doc/html/rfc9114#name-settings

It's testable through `H3FrameDispatcher` like the unit test does (Need to add Http3SettingsHandler instead of the mock handler).
https://github.com/apache/trafficserver/blob/2402444b5864411f2415ae142ba81d7637656531/proxy/http3/test/test_Http3FrameDispatcher.cc#L31-L47

It's also testable by calling `Http3SettingsHandler::handleFrame` directly.
https://github.com/apache/trafficserver/blob/2402444b5864411f2415ae142ba81d7637656531/proxy/http3/Http3App.h#L99

duke8253:
this resolves #9428

### DIFF
--- a/proxy/http3/Http3FrameDispatcher.cc
+++ b/proxy/http3/Http3FrameDispatcher.cc
@@ -41,11 +41,12 @@ Http3FrameDispatcher::add_handler(Http3FrameHandler *handler)
 }
 
 Http3ErrorUPtr
-Http3FrameDispatcher::on_read_ready(QUICStreamId stream_id, IOBufferReader &reader, uint64_t &nread)
+Http3FrameDispatcher::on_read_ready(QUICStreamId stream_id, Http3StreamType stream_type, IOBufferReader &reader, uint64_t &nread)
 {
   std::shared_ptr<const Http3Frame> frame(nullptr);
   Http3ErrorUPtr error = Http3ErrorUPtr(new Http3NoError());
   nread                = 0;
+  uint32_t frame_count = 0;
 
   while (true) {
     // Read a length of Type field and hopefully a length of Length field too
@@ -95,6 +96,7 @@ Http3FrameDispatcher::on_read_ready(QUICStreamId stream_id, IOBufferReader &read
       if (frame == nullptr) {
         break;
       }
+      ++frame_count;
 
       // Consume buffer if frame is created
       nread += frame_len;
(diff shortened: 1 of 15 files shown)


proxy/http3/Http3Types.h
duke8253, line 1: Default error code is not `0` anymore due to #9793. I can make a separate PR for this.

proxy/http3/Http3FrameDispatcher.cc
```diff
@@ -95,13 +96,23 @@ Http3FrameDispatcher::on_read_ready(QUICStreamId stream_id, IOBufferReader &read
...
       }
+      ++frame_count;
 
       // Consume buffer if frame is created
       nread += frame_len;
       reader.consume(frame_len);
 
       // Dispatch
       Http3FrameType type = frame->type();
+      if (stream_type == Http3StreamType::CONTROL) {
```
maskit, line 107: Can you put this into a private function? This block will be very long eventually. For example, we also need to check this:

> If a DATA frame is received on a control stream, the recipient MUST respond with a connection error of type H3_FRAME_UNEXPECTED.

I was going to make `ProtocolEnforcer` that inherits `FrameHandler` and handle all frame types (as you can see we can have multiple frame handlers for a frame type). With the enforcer, a dispatcher can focus on dispatching. The Single Responsibility Principle.

maskit (approved):
Looks good. Thanks for updating the code.

maskit:
[approve ci autest]

PR #9869, opened 2023-06-16
Extract HttpUserAgent class from HttpSM

JosiahWI:
This extracts a new class with 3 fields (entry, raw_buffer_reader, and txn). It doesn't have much responsibility yet. But it does provide a starting point for moving some behavior out of the state machine.

Edit: We extracted a lot more fields, and moved HttpVCTable to its own file.


proxy/http/HttpUserAgent.h
```diff
@@ -0,0 +1,110 @@
...
+  MIOBuffer *write_buffer;
+  VIO *read_vio;
+  VIO *write_vio;
+  HttpSMHandler vc_read_handler;
+  HttpSMHandler vc_write_handler;
+  HttpVC_t vc_type;
+  HttpSM *sm;
+  bool eos;
+  bool in_tunnel;
+};
```
maskit, line 56: These are obviously not just for "UserAgent". Probably need another header file if they can't stay in HttpSM.h.

proxy/http/HttpSM.h
```diff
@@ -101,20 +92,6 @@ enum BackgroundFill_t {
...
-  VIO *write_vio;
-  HttpSMHandler vc_read_handler;
-  HttpSMHandler vc_write_handler;
-  HttpVC_t vc_type;
-  HttpSM *sm;
-  bool eos;
-  bool in_tunnel;
-};
-
 struct HttpVCTable {
```
maskit, line 95: Having `HttpVCTableEntry` and `HttpVCTable` at different places doesn't look like a right thing.

proxy/http/HttpSM.h
```diff
@@ -569,9 +545,7 @@ class HttpSM : public Continuation, public PluginUserArgs<TS_USER_ARGS_TXN>
 
   HttpVCTable vc_table;
 
-  IOBufferReader *ua_raw_buffer_reader = nullptr;
-
-  HttpVCTableEntry *ua_entry     = nullptr;
+  HttpUserAgent _ua{};
```
maskit, line 548: It looks like we could do a similar thing for `server_entry` and `server_txn`. What would it be called? `HttpServer`? It would be nice if the two had a same prefix or suffix in their names to indicate they are basically the same thing but one for client side and one for server side.

bneradt:
[approve ci]

maskit (changes requested):
I'm fine with making changes step by step, but I'd like to see the big picture before making changes.

`HttpSMHandler`, `HttpVC_t` and `HttpVCTableEntry` probably need a better place.

proxy/http/HttpUserAgent.h
JosiahWI: Do you have a name in mind for this header file?

proxy/http/HttpSM.h
JosiahWI: Good catch. They should probably go in the same header.

proxy/http/HttpSM.h
JosiahWI: `HttpServer` seems like a good name. It might be a good idea to try this in a separate PR, unless you are ok with having a pretty large diff? This one is already a little bigger than I thought would be nice to review.

JosiahWI:
My objective is to move some of the logic for response headers into this new class. For example, `HttpSM::setup_push_read_response_header()` can be moved entirely or in part into this new class.

proxy/http/HttpUserAgent.h
JosiahWI: I've selected HttpVCTable.h.

maskit:
Hmm, the goal is still unclear to me. If it's going to be like we can get anything about a user agent from `HttpUserAgent`, and anything about a server from `HttpServer` that kinda make sense. But we already have `client_info` and `server_info` in `HttpTransact::State`. What would be the difference? And why don't you move other variables that are related to client (e.g. `client_request_hdr_bytes`, `client_tcp_reused`, client_protocol`, etc) into the class?

proxy/http/HttpSM.h
JosiahWI: I hadn't thought as far ahead as extracting `HttpServer` and making a base class, but I think that's a good direction to take this.

JosiahWI:
I wanted to split up the class a bit based on the single responsibility principle to make it more manageable. I did not have a long-term vision of how the refactoring would continue, but I'd like to try moving the other client fields over (in this PR) like you suggested, and splitting out `HttpServer` in a separate one.

I'm pretty new to the codebase and it's not very clear to me yet which fields belong together. I thought client and user agent were somehow different concepts in the codebase, because they are named differently.

JosiahWI:
The goal is to encapsulate these fields such that the state machine doesn't need to access this data at all. If it has less data it has to manage directly it will be a lot easier to follow the logic. I want to move operations on client and server data out of the state machine, so that the state machine focuses on changing between states correctly and delegates the details of changing the state to those new classes.

maskit:
Thanks for the clarification.

> I wanted to split up the class a bit based on the single responsibility principle to make it more manageable.

Sounds reasonable. You've already answered my question on the other comment but I was not sure what the responsibility of the new class where you have only one for user agent side.

> I'm pretty new to the codebase and it's not very clear to me yet which fields belong together. I thought client and user agent were somehow different concepts in the codebase, because they are named differently.

I'd say nobody fully understand the concepts 🙂 The two terms might have been used for a client on HTTP layer and a client on TCP layer, but this is just my guess.

> The goal is to encapsulate these fields such that the state machine doesn't need to access this data at all. If it has less data it has to manage directly it will be a lot easier to follow the logic.

Sounds great. Let's make it happen.

JosiahWI:
Rebased on master.

proxy/http/HttpSM.h
```diff
@@ -679,10 +622,58 @@ HttpTransact::State::state_machine_id() const
...
+{
+  return _ua.get_client_curve();
+}
+
+inline int
+HttpSM::get_client_alpn_id() const
+{
+  return _ua.get_client_alpn_id();
+}
+
```
maskit, line 672: So, we have to add a function to `HttpSM` when we add a function to `HttpUserAgent`? Is it a bad idea to have `HttpSM::get_client` or `HttpSM::get_useragent` so we don't have to have these?

JosiahWI: I was asking myself this question as well. Since the number of getters here is slightly getting out of hand, I think we should go with your suggestion. The goal is to be able to get rid of many of these getters, but that change will also be easier to handle if we don't have to remove getters in 2 places every time.

JosiahWI:
Rebased on master.

bryancall:
[approve ci autest]

JosiahWI:
The autest failure on this one indicates a real problem. I've been working on it but debugging the state machine without unit tests is a relatively large time sink.

JosiahWI:
I found the bug. I missed a single !, and I think I spent as much time finding that single missing ! as I spent making this entire PR. I'm going to add a unit test that covers that case, but I'm encountering a lot of segfaults in the test setup.

JosiahWI:
Rebased on master. Some unfinished changes also got pushed by mistake.

JosiahWI:
Rebased on master to resolve merge conflict.

JosiahWI:
[approve ci autest]

proxy/http/HttpSM.h
```diff
@@ -679,10 +615,16 @@ HttpTransact::State::state_machine_id() const
...
 }
 
+inline HttpUserAgent const &
+HttpSM::get_user_agent() const
+{
+  return _ua;
+}
+
 inline ProxyTransaction *
 HttpSM::get_ua_txn()
```
maskit, line 625: I wondered if we should keep this because now we can do `sm->get_user_agent()->get_txn()`. Maybe we should keep it, so we can get a user-agent transaction and a server transaction in a consistent way.

maskit (approved):
Looks good. Assuming functions are just moved and there's no logic change.

JosiahWI:
There isn't supposed to be any logic change. The AuTests passing gives me a little confidence that the behavior is preserved, and I am planning to add some more unit tests in a followup PR.

PR #9881, opened 2023-06-19
Fix autests for fedora 38

bneradt:
I ran autest on fedora:38 and noticed that all but three passed. This commit updates our autests to work with the newer openssl s_client and Python3.11.

The most significant set of changes come from hyper being abandoned an no longer working on Python 3.11. This converts the ad-hoc clients we had written in hyper to the h2 framework.

I kept the work for the three tests as three separate commits.


bneradt:
[approve ci rocky]

dragon512 (approved):

serrislew (approved):

PR #9883, opened 2023-06-20
Add max thread count options to CMake build

JosiahWI:
This adds an option that defaults to 4096: TS_MAX_EVENT_THREADS, and an option that defaults to 3072: TS_MAX_THREADS_PER_TYPE.


JosiahWI:
There is a discrepancy between the default MAX_THREADS_PER_TYPE value in the CMake build and in I_EventProcessor.h. The former has a default value of 3072, and the latter has a default value of 3071.

cmcfarlen (approved):

PR #9891, opened 2023-06-21
tests: minor autest script fixes

jpeach:
* switch to the tests directory before trying to run scripts from it
* don't check for the unnecessary Python devel tools
* exit if Python isn't available
* use the preferred bash, not the (obsolete on macOS) one in /bin

### DIFF
--- a/tests/test-env-check.sh
+++ b/tests/test-env-check.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#! /usr/bin/env bash
 
 #  Licensed to the Apache Software Foundation (ASF) under one
 #  or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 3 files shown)


JosiahWI:
Are the dev tools not required to compile some test dependencies?

dragon512 (approved):
I don't know why the pushd and popd had been added. It should have been the -D option. I also don't know why the python-config logic existed.

jpeach:
> Are the dev tools not required to compile some test dependencies?

Not AFAICT.

tests/test-env-check.sh
```diff
@@ -26,20 +26,11 @@ _END_
...
 
-# check for python development header -- for autest
-python3-config &> /dev/null
-if [ $? = 1 ]; then
-    echo "python3-dev/devel detected!"
-else
-    echo "python3-dev/devel is not installed. "
-    return
-fi
-
```
bneradt, line 42: Huh... on my fedora:38 system, `pipenv install` fails without python3-devel. I guess we should leave this check in.

PR #9833, opened 2023-06-13
cmake builds with quiche/quic support

bryancall:

PR #9892, opened 2023-06-21
Add support for quic to CMake build

JosiahWI:
This adds an option that defaults to OFF: ENABLE_QUICHE. When quiche is enabled, the quic static library and the http3 static library will be built and linked appropriately. The traffic_quic executable will also be built.

It also does some cleanup of targets in the CMake build. Some targets are now namespaced in ts:: to make it explicit which targets are part of trafficserver, and to make usage style consistent with that of third-party libraries. Some dependencies have also been corrected, and comments added where dependencies are misrepresented by necessity.

I opened this PR after consulting with @bryancall. It is an alternative to #9833, and part of it is taken from that PR (albeit heavily modified).


JosiahWI:
[approve ci]

CMakeLists.txt
```diff
@@ -313,6 +327,9 @@ add_subdirectory(mgmt/config)
 add_subdirectory(mgmt/rpc)
 add_subdirectory(src/traffic_server)
 add_subdirectory(src/traffic_ctl)
+if(TS_USE_QUIC)
+    add_subdirectory(src/traffic_quic)
```
maskit, line 331: Thanks for taking care of this, but actually we no longer maintain it. https://github.com/apache/trafficserver/pull/9726

JosiahWI: Thanks, I'll follow suite and obliterate it from the CMake build as well.

JosiahWI:
Test "TSVConnFD" failed with code -9.

JosiahWI:
[approve ci autest]

cmcfarlen (approved):

PR #9893, opened 2023-06-21
Install libswoc headers by default, can be disabled instead

zwoop:


JosiahWI:
Does this require a similar change to the CMake build?

zwoop:
> Does this require a similar change to the CMake build?

Good question. It may ? But probably that would be in the upstream version of the CMakeLists.txt file? @SolidWallOfCode thoughts?

bneradt:
[approve ci autest]

SolidWallOfCode (approved):

PR #7931, opened 2021-06-09
Adds new X-Effective-URL header to the xdebug plugin

jhiapple:
By adding "X-Effective-URL" to the request header "X-Debug",
the header "X-Effective-URL" header will be inserted to the
request headers with the effective URL value.

Exposes the TSHttpTxnEffectiveUrlStringGet() result, which is
the URL after the remap.

Note: the inserted header is intentionally X-prefixed for consistency
with the other headers inserted by the xdebug plugin.

PR #9899, opened 2023-06-22
For xdebug plugin, add doublequotes around the URL of X-Effective-URL.

jhiapple:
URLs can have commas which makes the URLs fragile when multiple URLs are returned, and the values of multiple X-Effective-URL headers are merged.

Follow-up on 170e12b4 (Adds new X-Effective-URL header to the xdebug plugin (#7931))

### DIFF
--- a/doc/admin-guide/plugins/xdebug.en.rst
+++ b/doc/admin-guide/plugins/xdebug.en.rst
@@ -127,6 +127,10 @@ X-Remap
     If the URL was remapped for a request, this header gives the *to* and *from* field from the line in remap.config that caused
     the URL to be remapped.
 
+X-Effective-URL
+    If the URL was remapped for a request, this header gives the URL resulting from the remapping. Note that if there are
+    multiple remaps, this header aggregates the URLs, space-comma-separated. The URLs are inside doublequotes.
+
 X-ParentSelection-Key
     The ``X-ParentSelection-Key`` header contains the URL that is used to
     determine parent selection for an object in the Traffic Server. This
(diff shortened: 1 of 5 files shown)


zwoop:
[approve ci]

bneradt:
[approve ci]

bneradt:
[approve ci]

bneradt:
[approve ci cmake]

ywkaras:
I suggest asking on the users mailing list if this would cause problems for anyone.

jhiapple:
> I suggest asking on the users mailing list if this would cause problems for anyone.

I'd be really surprised if anyone else than me is using the X-Effective-URL debug feature, which I added two years ago for a very particular use case I have.

Google search for

trafficserver "x-effective-url"

returns exactly two (2) hits, one in the 9.x "what's new" and one in the 9.2.0 changeling.

bryancall:
RFC on URL Encoding: https://datatracker.ietf.org/doc/html/rfc1738#section-2.2

maskit:
Note: the reason I suggested to back port this change to 9.2.x is that I think this is a bug (not just an enhancement). URLs that contain commas are problematic if those are in header values. See also the example on the RFC.
https://www.rfc-editor.org/rfc/rfc9110.html#name-field-values

ywkaras:
Can you add documentation for this option in doc/admin-quide/plugins/xdebug.en.rst ?

bneradt:
[approve ci clang-format]

zwoop:
Not 100% sure this should go into 9.2.x, considering it does change behavior.

ywkaras:
This is the spec for a quoted string in an HTTP header field value:  https://www.rfc-editor.org/rfc/rfc9110.html#name-quoted-strings

So, the only potential non-compliance would be if a double quote was in the URL (without being URL encoded as %22).  Can we be sure this won't happen?

maskit:
Double-quote is one of unsafe characters, and "All unsafe characters must always be encoded within a URL" according to RFC1738.
https://datatracker.ietf.org/doc/html/rfc1738#section-2.2

jhiapple:
Added documentation for the feature.

ywkaras (approved):

PR #9905, opened 2023-06-23
Add support for recvmmsg and UDP GRO

brbzull0:
### Descrition

This PR adds support for UDP GRO when reading the socket using `recvm(m)sg`,  also includes support for `recvmmsg` when it is available by the OS.

There is now a new configuration variable:

```yaml
# records.yaml
ts:
  udp:
    enable_gro: 1 # Enabled by default.
```
----

### Notes

All the tests were done using GET only and setting the h2load parameter to `--max-udp-payload-size=2k`.

|Feature       |Total # Req | # Clients | # recvmmsg | # recvmsg |
|--------------|------------|-----------|------------|-----------|
| recvmmsg+gro | 100k       | 100       |      22834 | 0         |
| recvmmsg     | 100k       | 100       |      23501 | 0         |
| recvmsg+gro  | 100k       | 100       |       | 247970         |
| recvmmsg     | 100k       | 100       |       | 249097         |

As I didn't see much of a difference between having GRO enabled or disabled I did run a test with debug
enabled and I was able to count how many packets came with the GRO set in the ancillary data. 
Then I realized why I wasn't actually seeing much difference, not too many packets were spliced by the kernel.

Packets with (No GRO/GRO) = Number of times that ATS detected a packet with spliced with GRO.

|Feature       |Total # Req | # Clients | # recvmmsg | # recvmsg | # Packets with (No GRO/GRO)  | GRO % |
|--------------|------------|-----------|------------|-----------|---------------|-------|
| recvmmsg+gro | 100k       | 100       |     13929  | 0         |    223080 / **180** | 0.8% |
| recvmsg+gro  | 100k       | 100       |      0| 278718         |    231760 / **34484** | 14% |

**Using recvmmsg reduced the number of syscall significantly**

This is what I used for this tests:

**bpftrace**

```bash
bpftrace -e 'tracepoint:syscalls:sys_enter_recvm* { @[comm, probe] = count();  }' 
```
**h2load**
```bash
/opt/bin/h2load -n 100000 -c 100 --npn-list=h3 URL --max-udp-payload-size=2k
```

I was using only **GET** requests with around 15 header with about 80 bytes each.
I have not tried POST yet which I believe will give us better insights.

I tested this on `Ubuntu 20.04` and `FreeBSD 13.1`. All figures here are from a tests on Ubuntu. 
This was tested on `FreeBSD 13.1` which does not support GSO/GRO to make sure the functionallity works the same.

### DIFF
--- a/iocore/net/UnixUDPNet.cc
+++ b/iocore/net/UnixUDPNet.cc
@@ -1717,8 +1717,9 @@ UDPNetHandler::is_gro_enabled() const
 {
 #ifndef SOL_UDP
   return false;
-#endif
+#else
   return this->_cfg.enable_gro;
+#endif
 }
 
 int
--- a/iocore/net/UnixUDPNet.cc
+++ b/iocore/net/UnixUDPNet.cc
@@ -37,6 +37,7 @@
 #include "P_DNSConnection.h"
 #include "P_Net.h"
 #include "P_UDPNet.h"
+#include "tscore/ink_inet.h"
 
 #include "tscore/ink_sock.h"
 #include <math.h>
@@ -54,6 +55,13 @@ using UDPNetContHandler = int (UDPNetHandler::*)(int, void *);
 
 ClassAllocator<UDPPacket> udpPacketAllocator("udpPacketAllocator");
 EventType ET_UDP;
+
+namespace
+{
+#ifdef HAVE_RECVMMSG
+const uint32_t MAX_RECEIVE_MSG_PER_CALL{16}; //< VLEN parameter for the recvmmsg call.
+#endif
+}; // namespace
 
 UDPPacket *
 UDPPacket::new_UDPPacket()
@@ -312,10 +320,40 @@ get_ip_address_from_cmsg(struct cmsghdr *cmsg, sockaddr_in6 *toaddr)
 #endif
   return false;
 }
+
+unsigned int
+build_iovec_block_chain(unsigned max_niov, int64_t size_index, Ptr<IOBufferBlock> &chain, struct iovec *out_tiovec)
+{
+  unsigned int niov;
+  IOBufferBlock *b, *last;
+
+  // build struct iov
+  // reuse the block in chain if available
+  b    = chain.get();
+  last = nullptr;
+  for (niov = 0; niov < max_niov; niov++) {
+    if (b == nullptr) {
+      b = new_IOBufferBlock();
+      b->alloc(size_index);
+      if (last == nullptr) {
+        chain = b;
+      } else {
+        last->next = b;
+      }
+    }
+
+    out_tiovec[niov].iov_base = b->buf();
+    out_tiovec[niov].iov_len  = b->block_size();
+
+    last = b;
+    b    = b->next.get();
+  }
+  return niov;
+}
 } // namespace
 
 void
-UDPNetProcessorInternal::udp_read_from_net(UDPNetHandler *nh, UDPConnection *xuc)
+UDPNetProcessorInternal::read_single_message_from_net(UDPNetHandler *nh, UDPConnection *xuc)
 {
   UnixUDPConnection *uc = (UnixUDPConnection *)xuc;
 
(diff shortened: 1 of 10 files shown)


JosiahWI:
The CMake build will need to be updated as well.

brbzull0:
> The CMake build will need to be updated as well.

Thanks for the comment. 
You mean this: https://github.com/apache/trafficserver/pull/9905/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R1393 ?

Cheers.

JosiahWI:
I took a closer look; it looks like we don't check this yet in CMake, but we probably should. We might want to make our own function:
```cmake
include(CheckIncludeFile)
function(REQUIRE_SYSTEM_HEADER HEADER)
    check_include_file("${HEADER}" HAVE_HEADER)
    if(NOT HAVE_HEADER)
        message(FATAL_ERROR "Missing system header ${HEADER}")
    endif()
endfunction()

# sendmsg, recvmsg, bunch of other stuff
require_system_header("sys/socket.h")
```

That's a little bigger change than I anticipated; if you don't want to do it in this PR I could do it afterwards.

JosiahWI:
Sorry, I didn't read carefully, that only works for sendmsg and recvmsg. I think for recvmmsg and sendmmsg we need to check for the function, which I have not figured out how to do (autotools has its pros...). I would ignore this for now.

iocore/net/UnixUDPNet.cc
```diff
@@ -326,61 +415,54 @@ UDPNetProcessorInternal::udp_read_from_net(UDPNetHandler *nh, UDPConnection *xuc
...
 #endif
-#if defined(IPV6_PKTINFO) || defined(IPV6_RECVPKTINFO)
-      case IPV6_PKTINFO: // IPV6_RECVPKTINFO uses IPV6_PKTINFO too
-        if (cmsg->cmsg_level == IPPROTO_IPV6) {
-          struct in6_pktinfo *pktinfo = reinterpret_cast<struct in6_pktinfo *>(CMSG_DATA(cmsg));
-          memcpy(toaddr.sin6_addr.s6_addr, &pktinfo->ipi6_addr, 16);
+    }
+
+    // If we got the gso size, then we need to find out in how many parts this was spliced.
+    int const parts = gso_size ? static_cast<int>(ceil(static_cast<double>(r) / static_cast<double>(gso_size))) : 1;
```
brbzull0, line 434: Not sure how bad would be using `ceil` just right here.

iocore/net/UnixUDPNet.cc
```diff
@@ -1452,13 +1709,22 @@ net_signal_hook_callback(EThread *thread)
...
   SET_HANDLER(&UDPNetHandler::startNetEvent);
 }
 
+bool
+UDPNetHandler::is_gro_enabled() const
+{
+#ifndef SOL_UDP
+  return false;
+#endif
+  return this->_cfg.enable_gro;
```
maskit, line 1725: It may be ok, but some analyzer may say this line is unreachable. I'd have this line in an else clause.

brbzull0: This is ugly, I meant to have an `else` here. Thanks

iocore/net/UnixUDPNet.cc
brbzull0: We can have a remainder bigger than 0.

brbzull0: > Oh, you mean `r` can be 2500 (1000 + 1000 + 500) where `gso_size` is 1000?
> 

Exactly.


> I don't have any better idea to calculate parts faster.
> https://quick-bench.com/q/IC9D3TNEcLCwWgEfoh_Ll3igs2k

does not seems to  change much I think.


> Another approach would be get rid of parts. Maybe we can subtract gso_size from r on every iteration instead (gso_size > would need to be set to r if gso_size is 0)?

I could probably do that I think.

brbzull0:
[approve ci autest]

brbzull0:
[approve ci autest]

brbzull0:
removed the `ceil` function and the parts handling, it now relies on the received size and work out with the remining size based on each part/packet size (if gro).

cc @maskit

iocore/net/UnixUDPNet.cc
```diff
@@ -37,20 +37,32 @@
 #include "P_DNSConnection.h"
 #include "P_Net.h"
 #include "P_UDPNet.h"
+#include "tscore/ink_inet.h"
 
 #include "tscore/ink_sock.h"
-
-#include "netinet/udp.h"
+#include <math.h>
```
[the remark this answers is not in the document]
brbzull0: defintelly not. Thanks.

maskit (approved):
Looks good. Thanks.

PR #9921, opened 2023-06-27
Fixes defines for the version strings

cmcfarlen:
I noticed the version was incorrect in headers so I fixed this in the CMakeLists.txt


JosiahWI (approved):
We might want to set the version in the `project` command and set all these defines from `PROJECT_VERSION`, `PROJECT_VERSION_MAJOR`, `PROJECT_VERSION_MINOR`, and so on. But this doesn't improve the clarity enough to be worthwhile now, I think.

JosiahWI (approved):
Very clean. +1 on removing the LIBTOOL_VERSION stuff, it looks like it was only applicable for the autotools build.

bneradt:
[approve ci ubuntu]

bneradt (approved):

PR #9923, opened 2023-06-27
Remove proxy.config.stat_api.max_stats_allowed, it's dead Jim

zwoop:


ywkaras (approved):
You can keep the red shirt.

PR #9928, opened 2023-06-28
Detect SSL_CTX_set_keylog_callback in CMake build

JosiahWI:
This looks for the symbol SSL_CTX_set_keylog_callback declared in ssl.h and defined in libssl.so. It defines the TS_HAS_TLS_KEYLOGGING appropriately so that we can run autests on TLS features, since the autests assume that TLS keylogging is enabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -213,8 +213,6 @@ check_symbol_exists(IN6_IS_ADDR_UNSPECIFIED "netinet/in.h" TS_HAS_IN6_IS_ADDR_UN
 check_symbol_exists(IP_TOS "netinet/ip.h" TS_HAS_IP_TOS)
 check_symbol_exists(SO_MARK "sys/socket.h" TS_HAS_SO_MARK)
 check_symbol_exists(SO_PEERCRED "sys/socket.h" TS_HAS_SO_PEERCRED)
-check_symbol_exists(TLS1_3_VERSION "${OPENSSL_INCLUDE_DIR}/openssl/ssl.h" TS_USE_TLS13)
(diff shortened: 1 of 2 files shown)


bneradt (commented):
Looks good to me.

CMakeLists.txt
```diff
@@ -254,6 +257,7 @@ check_symbol_exists(BIO_meth_get_ctrl "openssl/bio.h" HAVE_BIO_METH_GET_CTRL)
 check_symbol_exists(BIO_meth_get_create "openssl/bio.h" HAVE_BIO_METH_GET_CREATE)
 check_symbol_exists(BIO_meth_get_destroy "openssl/bio.h" HAVE_BIO_METH_GET_DESTROY)
 check_symbol_exists(DH_get_2048_256 "openssl/dh.h" TS_USE_GET_DH_2048_256)
+check_symbol_exists(SSL_CTX_set_keylog_callback openssl/ssl.h TS_HAS_TLS_KEYLOGGING)
```
randall, line 260: should this be a quoted include file (like all the previous lines) or unquoted (like the next line)?

JosiahWI: Great question. In this case it's a matter of style. Since the CMake documention has them quoted, and because @cmcfarlen quoted most of them, I suggest we quote them all. Also, if there happened to be a header with a space in the name, then the quotes would be necessary. I hope that never happens, but theoretically that would be a reason to prefer quotes.

bneradt:
[approve ci fedora]

bneradt:
[approve ci]

randall (approved):

PR #6108, opened 2019-10-31
Add new log field to output server name sent by client in TLS handshake.

ywkaras:

PR #9931, opened 2023-06-29
Use string copy for the SNI server name.

jpeach:
Since the server name is a string, we don't need to manually copy, we can just copy it into a std::string.

### DIFF
--- a/proxy/private/SSLProxySession.cc
+++ b/proxy/private/SSLProxySession.cc
@@ -21,8 +21,6 @@
   limitations under the License.
  */
 
-#include <cstring>
-
 #include "SSLProxySession.h"
 #include "I_EventSystem.h"
 #include "I_NetVConnection.h"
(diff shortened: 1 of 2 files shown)


bneradt:
[approve ci]

maskit:
I thought it is a workaround to ensure the string is available when ATS makes a log entry.
@ywkaras thoughts?

cmcfarlen (approved):

cmcfarlen:
> I thought it is a workaround to ensure the string is available when ATS makes a log entry.
> @ywkaras thoughts?

The lifetime should be the same, right?

maskit:
Ah, I read `unique_ptr` as `shared_ptr`... Now I don't get what it is for.

bneradt (approved):

maskit:
I found the related discussion about this variable. Seems like there **was** a lifetime issue and the type **was** shared_ptr on the first proposal. Maybe we don't have a reason not to use std:string now.
https://github.com/apache/trafficserver/pull/6108#pullrequestreview-310186898

PR #9933, opened 2023-06-29
Define DEBUG and _DEBUG in CMake Debug mode

JosiahWI:
Our debug asserts weren't working because these debug symbols weren't defined when using a Debug configuration.


JosiahWI:
@bneradt Consider adding -DCMAKE_BUILD_TYPE=Debug to one of the CMake builds on CI if it isn't already present.

cmcfarlen (approved):

bneradt:
> @bneradt Consider adding -DCMAKE_BUILD_TYPE=Debug to one of the CMake builds on CI if it isn't already present.

Done

PR #9934, opened 2023-06-29
Add more debug to the regex_revalidate main handler.

traeak:
This also updates and adds to the autests and treats the time_t like an int64_t (cleans up the logic).  Also adds extra warning to documentation about using this with MISS(refetch) which requires the parent be an origin and not another cdn.


plugins/regex_revalidate/regex_revalidate.c
```diff
@@ -462,20 +468,20 @@ load_config(plugin_state_t *pstate, invalidate_t **ilist)
             }
             if (i) {
               iptr->next = i;
-              TSDebug(PLUGIN_NAME, "Loaded %s %d %d %s", i->regex_text, (int)i->epoch, (int)i->expiry, strForResult(i->new_result));
+              TSDebug(PLUGIN_NAME, "Loaded %s %jd %jd %s", i->regex_text, i->epoch, i->expiry, strForResult(i->new_result));
```
ywkaras, line 471: What guarantees that the times will be of type intmax_t?  You could change to `intmax_t(i->epoch)` to be sure.

plugins/regex_revalidate/regex_revalidate.c
```diff
@@ -462,20 +468,20 @@ load_config(plugin_state_t *pstate, invalidate_t **ilist)
...
+        TSDebug(PLUGIN_NAME, "Skipping line %d, too few fields", ln);
       }
     }
     pcre_free(config_re);
     fclose(fs);
     pstate->last_load = s.st_mtime;
     return true;
   } else {
-    TSDebug(PLUGIN_NAME, "File mod time is not newer: %d >= %d", (int)pstate->last_load, (int)s.st_mtime);
+    TSDebug(PLUGIN_NAME, "File mod time is not newer: ftime: %jd <= last_load: %jd", s.st_mtime, pstate->last_load);
```
ywkaras, line 484: ditto

plugins/regex_revalidate/regex_revalidate.c
```diff
@@ -502,13 +508,13 @@ list_config(plugin_state_t *pstate, invalidate_t *i)
     iptr = i;
     while (iptr) {
       char const *const typestr = strForResult(iptr->new_result);
-      TSDebug(PLUGIN_NAME, "%s epoch: %d expiry: %d result: %s", iptr->regex_text, (int)iptr->epoch, (int)iptr->expiry, typestr);
+      TSDebug(PLUGIN_NAME, "%s epoch: %jd expiry: %jd result: %s", iptr->regex_text, iptr->epoch, iptr->expiry, typestr);
```
ywkaras, line 511: ditto all of these too.

traeak: ditto

plugins/regex_revalidate/regex_revalidate.c
traeak: ditto

plugins/regex_revalidate/regex_revalidate.c
```diff
@@ -251,9 +256,9 @@ prune_config(invalidate_t **i)
     iptr  = *i;
     ilast = NULL;
     while (iptr) {
-      if (difftime(iptr->expiry, now) < 0) {
-        TSDebug(PLUGIN_NAME, "Removing %s expiry: %d type: %s now: %d", iptr->regex_text, (int)iptr->expiry,
-                strForResult(iptr->new_result), (int)now);
+      if (iptr->expiry <= now) {
+        TSDebug(PLUGIN_NAME, "Removing %s expiry: %jd type: %s now: %jd", iptr->regex_text, (intmax_t)iptr->expiry,
+                strForResult(iptr->new_result), (intmax_t)now);
```
ywkaras, line 261: Why don' t you wanna do these C++ style not C style?  `intmax_t(iptr->expiry)` not `(intmax_t)iptr->expiry`.

traeak: because the plugin is still 'c' and it won't compile.

ywkaras (approved):

PR #9935, opened 2023-06-30
Remove socket option globals.

jpeach:
Remove the SOCKOPT_ON and SOCKOPT_OFF globals. These weren't used consistently and cand easily be replaced by a helper function. Apply the new helper function everywhere that toggles socket options on.


brbzull0 (approved):
Much cleaner. Thanks.

PR #9946, opened 2023-07-04
build_h3_tools: update for lib64

bneradt:
On some systems, such as amd64 Linux, OpenSSL 3.0 installs into a lib64 directory instead of lib. This updates the build_h3_tools for this.


bneradt:
[approve ci autest]

randall (approved):

PR #9947, opened 2023-07-05
OpenSSL 3.0 tls autest updates

bneradt:
This updates the curl invocations for the tls_client_versions and tls_client_versions_minmax autests so that it will run with the updated OpenSSL 3.0 which is more strict. This avoids the issue by toning down curl's security level to 0. A future PR should assess our proxy.config.ssl.server.cipher_suite configurations and update accordingly.

This also fixes a crash that tls.test.py found for OpenSSL 3.0 in which
SSLNetVConnection::load_buffer_and_write should return a non-zero in the
event of SSL_ERROR_SYSCALL.

### DIFF
--- a/tests/gold_tests/tls/tls_client_versions.test.py
+++ b/tests/gold_tests/tls/tls_client_versions.test.py
@@ -71,28 +71,32 @@ ts.Disk.sni_yaml.AddLines([
 tr = Test.AddTestRun("foo.com TLSv1_2")
 tr.Processes.Default.StartBefore(server)
 tr.Processes.Default.StartBefore(Test.Processes.ts)
-tr.Processes.Default.Command = "curl -v --tls-max 1.2 --tlsv1.2 --resolve 'foo.com:{0}:127.0.0.1' -k  https://foo.com:{0}".format(
+# Newer versions of OpenSSL further restrict the ciphers they accept. Setting
+# the security level to 0 "retains compatibility with previous versions of
+# OpenSSL." See:
+# https://www.openssl.org/docs/manmaster/man3/SSL_CTX_set_security_level.html
+tr.Processes.Default.Command = "curl -v --ciphers DEFAULT@SECLEVEL=0 --tls-max 1.2 --tlsv1.2 --resolve 'foo.com:{0}:127.0.0.1' -k  https://foo.com:{0}".format(
     ts.Variables.ssl_port)
 tr.ReturnCode = 35
 tr.StillRunningAfter = ts
 
 # Target foo.com for TLSv1.  Should succeed
 tr = Test.AddTestRun("foo.com TLSv1")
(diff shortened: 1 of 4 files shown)


bneradt:
[approve ci autest]

tests/gold_tests/tls/tls_client_versions.test.py
```diff
@@ -71,28 +71,28 @@
 tr = Test.AddTestRun("foo.com TLSv1_2")
 tr.Processes.Default.StartBefore(server)
 tr.Processes.Default.StartBefore(Test.Processes.ts)
-tr.Processes.Default.Command = "curl -v --tls-max 1.2 --tlsv1.2 --resolve 'foo.com:{0}:127.0.0.1' -k  https://foo.com:{0}".format(
+tr.Processes.Default.Command = "curl -v --ciphers DEFAULT@SECLEVEL=0 --tls-max 1.2 --tlsv1.2 --resolve 'foo.com:{0}:127.0.0.1' -k  https://foo.com:{0}".format(
```
jpeach, line 74: Please add a comment to explain why you need to set the ciphers flag, and to refer the reader to the docs that explain the syntax

https://www.openssl.org/docs/manmaster/man1/openssl-ciphers.html

tests/gold_tests/tls/tls_client_versions_minmax.test.py
```diff
@@ -75,35 +75,35 @@
 tr = Test.AddTestRun("foo.com TLSv1_2")
 tr.Processes.Default.StartBefore(server)
 tr.Processes.Default.StartBefore(Test.Processes.ts)
-tr.Processes.Default.Command = "curl -v --tls-max 1.2 --tlsv1.2 --resolve 'foo.com:{0}:127.0.0.1' -k  https://foo.com:{0}".format(
+tr.Processes.Default.Command = "curl -v --ciphers DEFAULT@SECLEVEL=0 --tls-max 1.2 --tlsv1.2 --resolve 'foo.com:{0}:127.0.0.1' -k  https://foo.com:{0}".format(
```
jpeach, line 78: Please add a comment to explain why you need to set the ciphers flag, and to refer the reader to the docs that explain the syntax

https://www.openssl.org/docs/manmaster/man1/openssl-ciphers.html

jpeach (dismissed):
Looks good. Please go ahead and merge after adding the extra comments.

The two tests look extremely similar - could they be combined into one?

tests/gold_tests/tls/tls_client_versions.test.py
bneradt: Sounds good. Comment added. I chose to link an explanation of why security level is being set rather than the syntax, because I figure the former is more relevant to the reader here and the syntax would be easier to find.

maskit (approved):

PR #9960, opened 2023-07-06
Adding host URL to NextHop Strategies debug message during host mark down.

### LINKED ISSUE #8050: Add PS debug to nexthop during mark up/down
In relation to #8049 , we should add similar debug to the new nexthop strategies during parent mark down and up

cybertunnel:
This adds a more verbose debug line for when ATS is determining the next hop while using strategies and marks a host down.
This feature was requested by #8050.

Example log line:
`[Jul  3 09:49:57.824] [ET_NET 9] NOTE: [0] NextHop initially marked as down p1.example.com for request example.com`

### DIFF
--- a/proxy/http/remap/NextHopHealthStatus.cc
+++ b/proxy/http/remap/NextHopHealthStatus.cc
@@ -76,6 +76,7 @@ NextHopHealthStatus::markNextHop(TSHttpTxn txn, const char *hostname, const int 
   int64_t fail_threshold  = sm->t_state.txn_conf->parent_fail_threshold;
   int64_t retry_time      = sm->t_state.txn_conf->parent_retry_time;
   uint32_t new_fail_count = 0;
+  const char *host         = sm->t_state.request_data.get_host();
 
   // make sure we're called back with a result structure for a parent
   // that is being retried.
@@ -123,7 +124,9 @@ NextHopHealthStatus::markNextHop(TSHttpTxn txn, const char *hostname, const int 
           new_fail_count = h->failCount += 1;
         }
       } // end lock guard
-      NH_Note("[%" PRId64 "] NextHop %s marked as down %s", sm_id, (result.retry) ? "retry" : "initially", h->hostname.c_str());
+
+      NH_Note("[%" PRId64 "] NextHop %s marked as down %s for request %s", sm_id, (result.retry) ? "retry" : "initially",
+              h->hostname.c_str(), host);
     } else {
       // if the last failure was outside the retry window, set the failcount to 1 and failedAt to now.
       { // lock guard
--- a/proxy/http/remap/NextHopHealthStatus.cc
+++ b/proxy/http/remap/NextHopHealthStatus.cc
@@ -124,12 +124,9 @@ NextHopHealthStatus::markNextHop(TSHttpTxn txn, const char *hostname, const int 
           new_fail_count = h->failCount += 1;
         }
       } // end lock guard
-      if (host == nullptr) {
-        NH_Note("[%" PRId64 "] NextHop %s marked as down %s", sm_id, (result.retry) ? "retry" : "initially", h->hostname.c_str());
-      } else {
-        NH_Note("[%" PRId64 "] NextHop %s marked as down %s for request %s", sm_id, (result.retry) ? "retry" : "initially",
-                h->hostname.c_str(), host);
-      }
+
+      NH_Note("[%" PRId64 "] NextHop %s marked as down %s", sm_id, (result.retry) ? "retry" : "initially", host ? host: "(null)");
+      
     } else {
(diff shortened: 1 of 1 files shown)


shukitchan:
[approve ci]

ezelkow1:
[approve ci]

cybertunnel:
@randall, @ezelkow1 can't review since we are under the same company 😢

proxy/http/remap/NextHopHealthStatus.cc
```diff
@@ -76,6 +76,7 @@ NextHopHealthStatus::markNextHop(TSHttpTxn txn, const char *hostname, const int
   int64_t fail_threshold  = sm->t_state.txn_conf->parent_fail_threshold;
   int64_t retry_time      = sm->t_state.txn_conf->parent_retry_time;
   uint32_t new_fail_count = 0;
+  const char *host        = sm->t_state.request_data.get_host();
```
ywkaras, line 79: Should we add `ink_assert(host != nullptr);` here?

cybertunnel: If I understand asserts in c++ properly, this would essentially kill ATS if the host is null, right? Do we really want to kill the entire ATS program if there is a single malformed URL host being sent?

cybertunnel: We could check for nullptr and write a more verbose message if host is not null. If it is, we can write a generic response.

cybertunnel: See [9ce7bdd](https://github.com/apache/trafficserver/pull/9960/commits/9ce7bdda709a5107b2c000f7c63451b531c5b3ae) for what I am talking about.

ywkaras: I guess it's possible/likely that the printf-like debug output call checks for and handles nullptr.

Yes, I think it's best that, when we make wrong assumptions about code invariants, that the application exit in a way that makes it clear what our wrong assumption is.  Rather than seeing a crash, and not being sure what the proximate cause was.

ywkaras:
[approve ci autest]

randall:
> @randall, @ezelkow1 can't review since we are under the same company 😢

I wasn't aware of that. He can still give his verbal here. I made him a reviewer since he created the original issue.

proxy/http/remap/NextHopHealthStatus.cc
ywkaras: Remember that ink_assert() checks are only done in debug builds.  It would not affect release builds.

ywkaras:
Why are you sure `sm->t_state.request_data.get_host()` would never return a null pointer here?  Seems likely intuitively, but, unless there's more reason to be sure, the check in debug builds seems like a good idea.

proxy/http/remap/NextHopHealthStatus.cc
```diff
@@ -123,7 +124,12 @@ NextHopHealthStatus::markNextHop(TSHttpTxn txn, const char *hostname, const int
           new_fail_count = h->failCount += 1;
         }
       } // end lock guard
-      NH_Note("[%" PRId64 "] NextHop %s marked as down %s", sm_id, (result.retry) ? "retry" : "initially", h->hostname.c_str());
+      if (host == nullptr) {
+        NH_Note("[%" PRId64 "] NextHop %s marked as down %s", sm_id, (result.retry) ? "retry" : "initially", h->hostname.c_str());
+      } else {
+        NH_Note("[%" PRId64 "] NextHop %s marked as down %s for request %s", sm_id, (result.retry) ? "retry" : "initially",
+                h->hostname.c_str(), host);
```
ywkaras, line 131: Why not
```
        NH_Note("[%" PRId64 "] NextHop %s marked as down %s for request %s", sm_id, (result.retry) ? "retry" : "initially",
                h->hostname.c_str(), host ? host : "(null)");
```

cybertunnel: Because I am fairly new to C++. So have a lot to still learn. Unsure how the `?` and `:` work in that statement.

ywkaras: It's a less verbose equivalent to using a lambda expression with an if statement:

`([](char const *h) -> char const * { if (h) return h; else return "(null)"; })(host)`

cybertunnel: Thank you for explaining, sounds like it is a `??` from Swift, where an optional can have a "default" value set. I made your suggested changes, should be updated now.

cybertunnel:
[approve ci]

ywkaras (approved):

ezelkow1 (approved):

PR #9965, opened 2023-07-06
Build binaries for bigobj AuTest

JosiahWI:
This builds the 2 binaries needed for the bigojb AuTest. It places them in the source tree (ugh) where the AuTest expects to find them.


cmcfarlen (approved):
ew

PR #9970, opened 2023-07-07
RPC client: Increase message buffer size.

brbzull0:
We need this to be bigger to eventually allocate a bigger number of records/metrics.

We are running some tests where the metrics are quite a large number and this current buffer isn't enough. 

This have only affect in traffic_ctl and traffic_top.

### DIFF
--- a/include/shared/rpc/IPCSocketClient.h
+++ b/include/shared/rpc/IPCSocketClient.h
@@ -27,7 +27,7 @@
 #include <sys/socket.h>
 #include <sys/un.h>
 
-#include <tscore/BufferWriter.h>
+#include <swoc/BufferWriter.h>
 
 namespace shared::rpc
 {
@@ -57,7 +57,7 @@ struct IPCSocketClient {
   self_reference send(std::string_view data);
 
   /// Read all the content from the socket till the passed buffer is full.
-  ReadStatus read_all(ts::FixedBufferWriter &bw);
+  ReadStatus read_all(swoc::FixedBufferWriter &bw);
 
   /// Closes the socket.
   void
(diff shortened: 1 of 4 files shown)


zwoop (approved):

zwoop:
[approve ci autest]

brbzull0:
This actually broke the RPCClient as the `LocalBuffer` seems not to be happy with the configured size. Will investigate.,

brbzull0:
[approve ci autest]

bryancall:
Can we make it so we stream buffers instead of creating one huge buffer?

brbzull0:
> Can we make it so we stream buffers instead of creating one huge buffer?

@bryancall 

Yes, I don't like that either, this code is tight with `FixedBufferWriter` which seems it does not support linked buffers unfortunally, I know Alan may be doing something about this at some point.
Refactoring this client may take another effort(which I am willing to take), the thing is in the meantime if you have a large amount of metrics(as Leif have) `traffic_ctl` will error out.
This allocation is only done in `traffic_ctl` and `traffic_top` and not in `traffic_server`

zwoop:
@brbzull0 The lifetime of this buffer is short?  It's not like we're going to keep adding a ton of these buffers, significantly increasing memory foot print are we ?

I don't like this either, but it's better than before (which was seriously limited)

zwoop (approved):

brbzull0:
> @brbzull0 The lifetime of this buffer is short? It's not like we're going to keep adding a ton of these buffers, significantly increasing memory foot print are we ?
> 
> I don't like this either, but it's better than before (which was seriously limited)


Exactly, `traffic_ctl` will create the buffer and automatically exit once done. 

In any case I will be changing this to better approach.
https://github.com/apache/trafficserver/issues/10052

PR #9971, opened 2023-07-07
Update to autest version 1.10.3

bneradt:
1.10.3 updates IPv6 port selection logic so it works on systems where /etc/hosts doesn't have localhost associated with v6 addresses. This happens not infrequently on recent Ubuntu releases.

### DIFF
--- a/tests/Pipfile
+++ b/tests/Pipfile
@@ -26,7 +26,7 @@ pyflakes = "*"
 [packages]
 
 # Keep init.cli.ext updated with this required autest version.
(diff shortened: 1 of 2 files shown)


bneradt:
Marking this for 9.2.x because without this version update, any v6 autests will fail on certain systems. It's a test-only change, so it should be safe for the 9.2.x branch.

maskit (approved):

zwoop:
Cherry-picked to v9.2.x

PR #9972, opened 2023-07-07
Remove the useless std::vector calls in plugin_agents_cleanup.

bneradt:
@maskit observed that there is an unneccesary use of vectors in HttpSM::plugin_agents_cleanup. I had in mind using the vector to reduce the duplicated logic across the request/response hooks, but that never got implemented and using that doesn't add much value, if any. I'll simply remove the use of the vectors to simplify the code.

---
# For Reviewer

Selecting "Hide whitespace" diff configuration makes this easier to read.


maskit (approved):
Thanks.

PR #9973, opened 2023-07-07
Don't use std::min when we don't care which value is smaller

maskit:
I found that `std::min` is unnecessarily used in `Http2ConnectionState::restart_stream()`. This is a tiny optimization but we don't care which value is smaller here, and it doesn't look like the use of `std::min`  contributes to readability either.

`h2load -n 20000 -m 1 -c 1  "https://localhost:8443/256k"`

image


bneradt (approved):
Makes sense.

PR #9974, opened 2023-07-07
Switch libtscore to static and deal with the fallout

cmcfarlen:


zwoop:
I'm working on the corresponding changes to automake. I drew the short straw.

cmcfarlen:
[approve ci ubuntu]

cmcfarlen:
[approve ci centos]

JosiahWI:
It looks like some targets are linking to pcap that don't use it. Could we mark those with a # transitive comment? If we don't keep track of which usage requirements are transitive, then figuring it out later will be a pain.

zwoop:
I've tested this with my Makefile.am branch, and it works fine. I will make a PR for my branch as soon as this lands, since there are dependencies.

JosiahWI:
> > It looks like some targets are linking to pcap after these changes that don't use it. Could we mark those with a # transitive comment? If we don't keep track of which usage requirements are transitive, then figuring it out later will be a pain.
> 
> Yeh, we can go back and fix those once the larger refactoring is done. The Makefile patch I have is large, 2000 lines of changes.

Sounds good. Note that that my comment only applied to the CMake build, though. Automake doesn't manage usage requirements the same way.

bryancall:
[approve ci rocky]

JosiahWI:
> It looks like some targets are linking to pcap after these changes that don't use it. Could we mark those with a # transitive comment? If we don't keep track of which usage requirements are transitive, then figuring it out later will be a pain.

src/traffic_server/CMakeLists.txt
```diff
@@ -80,6 +80,9 @@ endif (TS_USE_LINUX_IO_URING)
 if (TS_USE_LINUX_NATIVE_AIO)
     target_link_libraries(traffic_server PRIVATE aio)
 endif (TS_USE_LINUX_NATIVE_AIO)
+if(TS_USE_POSIX_CAP)
+    target_link_libraries(traffic_server PUBLIC cap::cap)
+endif()
```
JosiahWI, line 85: I don't think traffic_server has a direct dependency on libpcap, so it should not be linked here. Let CMake take care of resolving the transitive dependencies.

```cmake
add_library(dependency dependency.cc)

target_link_libraries(dependency PUBLIC transitive::transitive)

add_executable(main main.cc)

target_link_libraries(main PRIVATE dependency)
```

CMake will link main to transitive::transitive if dependency is STATIC; if dependency is SHARED, then include paths for transitive::transitive will be added to main, but the library will not be linked.

This is the ideal. We can't do it everywhere in ATS yet because our dependency graph is a nightmare; if we really have to link directly to the transitive dependency for now, let's mark it so we know it should be fixed later.

This applies to a number of targets changed in this PR.

JosiahWI (changes requested):
Please address the above concern before this is merged.

JosiahWI (commented):
The CMake part of it looks pretty good now for a start. I'm planning to go through and try to improve the rest of the usage requirements separately at some point.

src/traffic_server/CMakeLists.txt
cmcfarlen: I fixed this one.  I'm sure there are more link commands that are over specified, but we can address those in a future pass as we further consolidate targets.

cmcfarlen:
[approve ci autest]

bneradt (approved):

PR #9975, opened 2023-07-09
Clean up certifier plugin debug messages.

jpeach:
The certifier plugin manually write out the function name in debug messages, and these were wrong in some cases. Clean this up to use __func__ and a more consistent message format. Remove the "t" mode from fopen call, which is a Windows-ism.

### DIFF
--- a/plugins/certifier/certifier.cc
+++ b/plugins/certifier/certifier.cc
@@ -211,7 +211,7 @@ public:
 
         // Remove oldest node if size exceeds limit
         if (++size > limit) {
-          TSDebug(PLUGIN_NAME, "Removing %s", tail->commonName.c_str());
+          TSDebug(PLUGIN_NAME, "%s: removing '%s'", __func__, tail->commonName.c_str());
           auto iter = cnDataMap.find(tail->commonName);
           if (iter != cnDataMap.end()) {
             local = std::move(iter->second); // copy ownership
@@ -224,7 +224,7 @@ public:
         }
       }
     }
-    TSDebug(PLUGIN_NAME, "%s Prepend to LRU list...List Size:%d Map Size: %d", data->commonName.c_str(), size,
+    TSDebug(PLUGIN_NAME, "%s: prepending '%s' to LRU ... list size: %d, map size: %d", __func__, data->commonName.c_str(), size,
             static_cast<int>(cnDataMap.size()));
 
     TSMutexUnlock(list_mutex);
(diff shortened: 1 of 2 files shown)


plugins/certifier/certifier.cc
```diff
@@ -211,7 +211,7 @@ class SslLRUList
 
         // Remove oldest node if size exceeds limit
         if (++size > limit) {
-          TSDebug(PLUGIN_NAME, "Removing %s", tail->commonName.c_str());
+          TSDebug(PLUGIN_NAME, "%s: removing '%s'", __func__, tail->commonName.c_str());
```
bneradt, line 214: Let's definitely remove explicit function names, but I don't think we need to add them to the log via `__func__` either since we support automatically adding them:
https://docs.trafficserver.apache.org/en/latest/admin-guide/files/records.yaml.en.html#proxy-config-diags-show-location

Or does that not apply to plugin messages?

bneradt (changes requested):
Thanks for cleaning this up.

@lzx404243 : can you review this too?

bneradt:
Looks like the autest might need an expected log expression update:
https://ci.trafficserver.apache.org/job/Github_Builds/job/autest/9715/console

```
     file /tmp/sandbox/certifier/ts2/log/traffic.out : Should reuse the existing certificate - Failed
        Reason: Contents of /tmp/sandbox/certifier/ts2/log/traffic.out did not contains expression: "Reuse existing cert and context for www.tls.com"
```

plugins/certifier/certifier.cc
jpeach: Yeh `TSDebug` doesn't know the source location. It wouldn't be too hard to fix that though.
```C
tsapi void TSDebug(const char *tag, const char *format_str, ...) TS_PRINTFLIKE(2, 3);
```

bneradt: Ah, OK. I'll leave it up to you then whether you want to keep the `__func__` here or add TSDebug support for it in a separate PR.

Thanks for looking into that.

jpeach: It would definitely be a separate PR.

jpeach:
[approve ci]

plugins/certifier/certifier.cc
```diff
@@ -419,25 +421,25 @@ shadow_cert_generator(TSCont contp, TSEvent event, void *edata)
...
+  TSDebug(PLUGIN_NAME, "%s: cert file is expected at %s", __func__, cert_filename.c_str());
   /// If cert file exists and is readable
   if (fp != nullptr) {
     cert.reset(PEM_read_X509(fp, nullptr, nullptr, nullptr));
     fclose(fp);
 
     if (cert == nullptr) {
       /// Problem with cert file / openssl read
-      TSError("[%s] [shadow_cert_generator] Problem with loading certs", PLUGIN_NAME);
+      TSError("[%s] %s: failed to loading certificated from '%s'", PLUGIN_NAME, __func__, cert_filename.c_str());
```
lzx404243, line 432: `failed to loading certificated` -> `failed to load certificate`

plugins/certifier/certifier.cc
```diff
@@ -419,25 +421,25 @@ shadow_cert_generator(TSCont contp, TSEvent event, void *edata)
...
       std::remove(cert_filename.c_str());
     } else {
-      TSDebug(PLUGIN_NAME, "shadow_cert_generator(): Loaded cert from file");
+      TSDebug(PLUGIN_NAME, "%s: loaded certificates from '%s'", __func__, cert_filename.c_str());
```
lzx404243, line 435: I think here we are loading a single certificate so `certificates` -> `certificate`

tests/gold_tests/pluginTest/certifier/certifier.test.py
lzx404243, line 1: Looks like the following expectation needs to be updated too:
```
       self.ts.Disk.traffic_out.Content += Testers.ContainsExpression(
            "Creating shadow certs",
            "Verify the certifier plugin generates the certificate dynamically.")
```

lzx404243 (changes requested):
Thanks for the clean up. Looks good overall. Some minor comments.

plugins/certifier/certifier.cc
```diff
@@ -529,28 +531,33 @@ cert_retriever(TSCont contp, TSEvent event, void *edata)
...
     SSL_set_SSL_CTX(ssl, ref_ctx);
     TSVConnReenable(ssl_vc);
+    return TS_SUCCESS;
   }
 
-  /// For scheduled connections, the scheduled continuation will handle the reenabling
+  // If no existing context, schedule TASK thread to generate.
+  TSDebug(PLUGIN_NAME, "%s: scheduling thread to generate/retrieve cert for %s", __func__, servername);
+  TSCont schedule_cont = TSContCreate(shadow_cert_generator, TSMutexCreate());
+  TSContDataSet(schedule_cont, (void *)servername);
```
ywkaras, line 558: I don' t think you need to cast to `void *`.  In any case, C-style casts are deprecated in new code.

plugins/certifier/certifier.cc
```diff
@@ -529,28 +531,33 @@ cert_retriever(TSCont contp, TSEvent event, void *edata)
...
+    TSDebug(PLUGIN_NAME, "%s: reusing existing cert and context for %s", __func__, servername);
     SSL_set_SSL_CTX(ssl, ref_ctx);
     TSVConnReenable(ssl_vc);
+    return TS_SUCCESS;
   }
 
-  /// For scheduled connections, the scheduled continuation will handle the reenabling
+  // If no existing context, schedule TASK thread to generate.
+  TSDebug(PLUGIN_NAME, "%s: scheduling thread to generate/retrieve cert for %s", __func__, servername);
+  TSCont schedule_cont = TSContCreate(shadow_cert_generator, TSMutexCreate());
```
ywkaras, line 557: Why not include `<tscpp/api/Cleanup.h>` and change this to:
```
  atscppapi::TSContUniqPtr schedule_cont{TSContCreate(shadow_cert_generator, TSMutexCreate());
```
So it will be cleaned up on program exit.

jpeach: This didn't change any code. Why clean up continuations on exit? Just for leak detection?

plugins/certifier/certifier.cc
jpeach: This was in the original code. Fixed.

jpeach:
[approve ci]

bryancall:
[approve ci]

ywkaras:
[approve ci autest]

ywkaras:
@bneradt is this good now?

bneradt (approved):

PR #9983, opened 2023-07-10
Add config to disable http/0.9 over QUIC by default

### LINKED ISSUE #9749: Add a setting to disable HTTP/0.9 over QUIC
Have a configuration option to disable HTTP/0.9 support.  Disabled by default.

duke8253:
this resolves #9749

### DIFF
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -4854,6 +4854,10 @@ removed in the future without prior notice.
 
    Specified the maximum outgoing UDP payload size.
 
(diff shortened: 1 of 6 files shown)


maskit:
Is it too tricky to exclude "hq" from the protocol list for ALPN?

duke8253:
Hmm, I guess can move the config check to `negotiated_application_name()` .

maskit:
The list is set by `quiche_config_set_application_protos`. I'd rather remove "hq" from the list so it won't be negotiated in the first place.

duke8253:
Ahh gotcha.

maskit (approved):
Thanks.

PR #9984, opened 2023-07-10
Strategies vhost support

ezelkow1:
Adds a host_override option to a strategy definition. This will modify the requests host header to be that of the selected parent instead of the original request host header (some origins depend on this sort of functionality)


ezelkow1:
[approve ci autest]

ywkaras (approved):

PR #9985, opened 2023-07-10
prefetch: widen evaluations from 32 to 64 bits

traeak:
useful when someone wants to use grossly huge numbers for segment indices.


moonchen:
I think that making it 64-bit is good, but we should update the documentation:

https://github.com/apache/trafficserver/blob/7a3aaa4b350a6fe3ee6bb17727176757151772a6/doc/admin-guide/plugins/prefetch.en.rst?plain=1#L163

I think this change also re-introduces overflow behavior on addition.

moonchen (changes requested):
I think that making it 64-bit is good, but we should update the documentation:

https://github.com/apache/trafficserver/blob/7a3aaa4b350a6fe3ee6bb17727176757151772a6/doc/admin-guide/plugins/prefetch.en.rst?plain=1#L163

I think this change also re-introduces overflow behavior on addition.

moonchen (approved):
Looks great!

PR #9986, opened 2023-07-10
Changes libtscore to an archive lib for the standard Makefile.am build

zwoop:
This requires #9974 to succeed the CI, but I've tested them both combined.


cmcfarlen:
[approve ci]

PR #9990, opened 2023-07-11
Build tslua plugin in CMake build

JosiahWI:
This also adds a Find script for LuaJIT, which uses pkg-config to find the LuaJIT library.


shukitchan (dismissed):
It looks good.

JosiahWI:
@shukitchan Somewhat related to this PR, I'm getting an interesting Lua panic in all the lua plugin AuTests with this PR.

traffic.out:
```
[Jul 11 13:19:19.135] traffic_crashlo ERROR: missing password database entry for ''
... (more content ommitted)
PANIC: unprotected error in call to Lua API (bad light userdata pointer)
```

I've included an error from traffic_crashlog here as well, in case the panic is a symptom of that.

shukitchan:
What is the version of luajit? 
It looks like a problem with luajit version. 

I think it won't work with anything below 2.0.5

JosiahWI:
Version 2.1 I believe. @bneradt can run the AuTests just fine on the fedora38 image; so it's not an issue with the build configuration.

shukitchan:
I see that the PR build autests all passed now?

JosiahWI:
Those were built with automake. There's nothing in CI that tests that the CMake build is correct, only that it compiles.

bneradt:
> What is the version of luajit? It looks like a problem with luajit version.
> 
> I think it won't work with anything below 2.0.5

Huh, yeah. This could likely be the difference. I bet fedora:38 has a more recent luajit.

PR #10002, opened 2023-07-12
Use cmake-format for consistent CMakeLists.txt formatting

zwoop:
This doesn't enforce this (yet), but easy to run manually with something like

```
cmake-format -i $(find . -name CMakeLists.txt)
```

For now, don't run this on the yaml-cpp and libswoc CMakeLists.txt files. To install cmake-format, I ran

```
pip3 install cmake-format
```


JosiahWI:
Definitely very different from the personal style I adhere to, but it's important to have a consistent style.

JosiahWI:
We currently have a very high throughput of CMake changes. We want to make sure we enforce this in CI quickly.

zwoop:
@JosiahWI If you have suggestions for changes to make to the .clang-format.yaml, please post here.

JosiahWI:
When a command is split across multiple lines, I prefer to put the variable or target of interest on the same line as the command:
```cmake
target_compile_options(my_target
  PRIVATE
    -Wall
    -Wextra
)
 
 This distinguishes the subject of the command, or in a variable in a `set()` command, from the rest of the arguments.
 I put usage requirement scopes on their own line:
 ```cmake
 PRIVATE
     -Wall
```
instead of
```cmake
PRIVATE -Wall
```
I think this is easier to type and looks nicer. It's probably easier for tools to auto-indent correctly as well.

I also use 2-space indents. I used 4-space indents in this repo because most of it was already 4-space indents, but Kitware seems to use 2-space indents and I picked up the habit from looking at their repo. I've personally found it nice, because CMake lines tend to get pretty long and in some cases cannot be split across multiple lines easily. This also supports the previous point.

Everything else I noticed looks good to me.

JosiahWI:
```cmake
target_include_directories(
    yaml-cpp PUBLIC $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include> $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
    PRIVATE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>
)
```

Just noticed this in the changes. This is not good. This is obscuring behavior by destroying the visual separation between the two generators.

JosiahWI:
I'm interesting in what @cmcfarlen prefers. Most of the CMake is pretty consistent and I think we're generally in agreement about how to format it, but he may have some good reasons for doing things differently that I haven't considered.

zwoop:
I don't know that cmake-format has options for what you are asking for, but if you can figure it out, please post.

zwoop:
And I should say, I rather live with some ugliness, as long as it's consistently ugly, rather than pretty inconsistency.

zwoop:
I made one config change, which I think reduces some of the ugliness mentioned. I can't find a way to keep the first word on the line with the leading '('.

cmcfarlen:
I also can't figure out how to get the first token up with the `(`. Although this does seem to happen for `set` and some others.  
I agree with @JosiahWI about the change to the directory grouping.  That's a problem.

setting `max_subgroups_hwrap: 4` makes some things better and some worse 😢

.cmake-format.yaml
```diff
@@ -0,0 +1,245 @@
...
+format:
+  _help_disable:
+    - Disable formatting entirely, making cmake-format a no-op
+  disable: false
+  _help_line_width:
+    - How wide to allow formatted cmake files
+  line_width: 120
+  _help_tab_size:
+    - How many spaces to tab for indent
+  tab_size: 4
```
bneradt, line 33: Let's use the default value of 2 here since that seems to be the expected default.

Everything else looks fine. As usual it's not perfect, but the consistency is worth it.

bneradt (changes requested):
Looks good. Thanks for working on this.

iocore/dns/CMakeLists.txt
```diff
@@ -15,24 +15,19 @@
...
-        ${CMAKE_SOURCE_DIR}/iocore/aio
-        ${CMAKE_SOURCE_DIR}/iocore/net
-        ${CMAKE_SOURCE_DIR}/iocore/cache
-        ${CMAKE_SOURCE_DIR}/iocore/hostdb
-        ${CMAKE_SOURCE_DIR}/proxy
-        ${CMAKE_SOURCE_DIR}/proxy/http
-        ${CMAKE_SOURCE_DIR}/proxy/hdrs
-        ${CMAKE_SOURCE_DIR}/mgmt
-        ${CMAKE_SOURCE_DIR}/mgmt/utils
-)
```
masaori335, line 38: Does `cmake-format` have a option like `InsertNewlineAtEOF` of `clang-format`?

zwoop: Not that I could see :/

JosiahWI:
I wonder how hard it would be to roll our own formatter.

JosiahWI:
I asked on the CMake Matrix room about formatters and the maintainer of https://github.com/Decodetalkers/neocmakelsp/ answered me. It's a full language server, so it does more than we need, and I could not get it to pick up the .editconfig file as described in their README, but I am pretty happy with the default format. Maybe @cmcfarlen would take a look at it as well.

bryancall (approved):

PR #10004, opened 2023-07-12
Add support for request/response transform for Wasm plugin

shukitchan:


shukitchan:
[approve ci autest]

plugins/experimental/wasm/wasm_main.cc
```diff
@@ -46,6 +46,250 @@ struct WasmInstanceConfig {
...
+
+      c->clearTransformResult();
+    }
+
+    if (status == proxy_wasm::FilterDataStatus::StopIterationNoBuffer) {
+      c->clearTransformResult();
+    }
+
+    if (eos && !towrite) { // EOS
+      eos = 1;
```
serrislew, line 218: I don't think this does anything

plugins/experimental/wasm/wasm_main.cc
```diff
@@ -46,6 +46,250 @@ struct WasmInstanceConfig {
...
+          status == proxy_wasm::FilterDataStatus::StopIterationAndWatermark) &&
+         eos && !towrite)) {
+      res = c->getTransformResult(&res_len);
+
+      if (res && res_len > 0) {
+        if (!ti->output_vio) {
+          if (eos && !towrite) {
+            ti->output_vio = TSVConnWrite(output_conn, contp, ti->output_reader, res_len); // HttpSM go on
+          } else {
+            ti->output_vio = TSVConnWrite(output_conn, contp, ti->output_reader, ti->downstream_bytes); // HttpSM go on
```
serrislew, line 201: why would this be different? only `transform_result_` gets written right so shouldn't the number of bytes written always be `res_len`?

plugins/experimental/wasm/wasm_main.cc
```diff
@@ -46,6 +46,250 @@ struct WasmInstanceConfig {
...
+    TSContCall(TSVIOContGet(input_vio), TS_EVENT_ERROR, input_vio);
+    break;
+
+  case TS_EVENT_VCONN_WRITE_COMPLETE:
+    TSDebug(WASM_DEBUG_TAG, "[%s] event vconn write complete", __FUNCTION__);
+    TSVConnShutdown(TSTransformOutputVConnGet(contp), 0, 1);
+    break;
+
+  case TS_EVENT_VCONN_WRITE_READY:
+  default:
```
serrislew, line 284: does `TS_EVENT_VCONN_EOS` not have a different course of action or is it assumed that it will eventually go to `TS_EVENT_VCONN_WRITE_COMPLETE`

plugins/experimental/wasm/wasm_main.cc
```diff
@@ -46,6 +46,250 @@ struct WasmInstanceConfig {
...
+  } else {
+    input_avail   = 0;
+    upstream_done = 0;
+    toread        = 0;
+    eos           = 1;
+  }
+
+  if (input_avail > 0) {
+    // move to the reserved.buffer
+    TSIOBufferCopy(ti->reserved_buffer, input_reader, input_avail, 0);
```
serrislew, line 127: why is there a reserved buffer and reserved reader? it seems like it's acting as an intermediate buffer

plugins/experimental/wasm/wasm_main.cc
```diff
@@ -46,6 +46,250 @@ struct WasmInstanceConfig {
...
+  do {
+    TSDebug(WASM_DEBUG_TAG, "[%s] inside transform handler loop", __FUNCTION__);
+    proxy_wasm::FilterDataStatus status = proxy_wasm::FilterDataStatus::Continue;
+
+    if (towrite == 0 && empty_input == 0) {
+      break;
+    }
+
+    TSDebug(WASM_DEBUG_TAG, "[%s] retrieving text and calling the wasm handler function", __FUNCTION__);
+    if (empty_input == 0) {
```
serrislew, line 150: also this might be more nit-picky but it seems like this conditional is checked a lot and could be refactored to handle this case separately instead of having it checked every block of code

plugins/experimental/wasm/wasm_main.cc
shukitchan: that's right

plugins/experimental/wasm/wasm_main.cc
shukitchan: I think it will be TS_EVENT_VCONN_WRITE_COMPLETE eventually. 
I am following the null transform example plugin on the event handling

plugins/experimental/wasm/wasm_main.cc
shukitchan: "reserved" is for temporarily storing the current block. That block will be passed to the wasm module for the provided function to process. And "output" is for actually taking the processed result out to pass it back to ATS

plugins/experimental/wasm/wasm_main.cc
shukitchan: That is probably right. Will try this out.

shukitchan: Actually there is a check on line 197 for output_vio. 
so for a long response that comes in multiple block, we will issue TSVConnWrite with INT64_MAX once. For a short response that comes in one single block, we will issue TSVConnWrite with the exact size. 
That's what the code is trying to do.

shukitchan:
@serrislew would you pls review again? Thanks.

plugins/experimental/wasm/wasm_main.cc
serrislew: oh for a large response with multiple blocks, it will iterate and keep writing so you are only calling TSVConnWrite to initialize for all of the bytes. got it, thanks for clarifying

plugins/experimental/wasm/wasm_main.cc
```diff
@@ -46,6 +46,249 @@ struct WasmInstanceConfig {
...
+  if (eos && !ti->output_vio) {
+    ti->output_vio = TSVConnWrite(output_conn, contp, ti->output_reader, 0);
+  }
+
+  if (write_down || eos) {
+    TSVIOReenable(ti->output_vio);
+  }
+
+  if (toread > input_avail) { // upstream not finished.
+    if (eos) {
```
serrislew, line 232: is this even possible? if `toread > input_avail`, `eos` will always be false since there is more to read? also from how `eos` is set above?

plugins/experimental/wasm/wasm_main.cc
serrislew: actually from `transform_handler` logic (in my previous comment), I don't think `TS_EVENT_VCONN_EOS` would even be called

plugins/experimental/wasm/wasm_main.cc
```diff
@@ -46,6 +46,249 @@ struct WasmInstanceConfig {
 
 static std::unique_ptr<WasmInstanceConfig> wasm_config = nullptr;
 
+// handler for transform event
+static int
+transform_handler(TSCont contp, ats_wasm::TransformInfo *ti, TSEvent event)
```
serrislew, line 51: TSEvent event not needed in transform_handler

plugins/experimental/wasm/wasm_main.cc
```diff
@@ -46,6 +46,249 @@ struct WasmInstanceConfig {
...
+      eos = false;
+    }
+  } else {
+    input_avail   = 0;
+    upstream_done = 0;
+    toread        = 0;
+    eos           = true;
+  }
+
+  if (input_avail > 0) {
```
serrislew, line 125: if `toread` < `input_avail`, would we only want to copy `toread` bytes? or does it not matter if we copy more than needed?

shukitchan: The input_avail is from the buffer. So we do the buffer copy with the number of bytes available in the buffer, not with the data left to be read. 

I guess in reality you probably will not get toread < input_avail.

plugins/experimental/wasm/wasm_main.cc
shukitchan: let me remove this

plugins/experimental/wasm/wasm_main.cc
shukitchan: it can be true by the function in the wasm module. I forgot to retrieve that value and set eos . 

Good find. So that's why you can still have the EOS event below.

shukitchan: I just double checked the proxy-wasm spec. It currently does not allow changing the eos. But I expect it be possible in the future. So I want to keep the code as if. But let me add a bunch of comments to explain the idea

shukitchan:
@serrislew ready for another review

serrislew:
LGTM

serrislew (approved):

PR #10050, opened 2023-07-15
cmake: -Wall -Wextra flags

bneradt:
This turns on -Wall -Wextra. Our practice with autoconf before and with
cmake here is to enable -Werror via the commandline,
-DCMAKE_COMPILE_WARNING_AS_ERROR=ON in this case.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -335,6 +335,13 @@ if (OPENSSL_FOUND)
 endif(OPENSSL_FOUND)
 
 add_subdirectory(lib)
+
+# Keep this after lib because lib is made up of third party libraries, so if
+# there are warnings, we can't do anything about it.
+# Note: -DCMAKE_COMPILE_WARNING_AS_ERROR=ON will turn warnings into errors.
+add_compile_options(-pipe -Wall -Wextra -Wno-unused-parameter -Wno-format-truncation -Wno-cast-function-type -Wno-stringop-overflow)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -335,6 +335,13 @@ if (OPENSSL_FOUND)
 endif(OPENSSL_FOUND)
 
 add_subdirectory(lib)
+
+# Keep this after lib because lib is made up of third party libraries, so if
+# there are warnings, we can't do anything about it.
+add_compile_options(-pipe -Wall -Wextra -Werror -Wno-unused-parameter -Wno-format-truncation -Wno-cast-function-type -Wno-stringop-overflow)
+add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wno-noexcept-type>)
+add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wsuggest-override>)
```
bneradt, line 343: I could not figure out how to add two options to a generator like this without calling add_compile_options twice. I tried a variety of quoting permutations, but nothing worked. If someone else can figure it out, I'll pull it in.

JosiahWI:
```cmake
add_compile_options("$<$<COMPILE_LANGUAGE:CXX>:-Wno-except-type;-Wsuggest-override>"
```

JosiahWI (approved):

SolidWallOfCode (approved):

PR #10048, opened 2023-07-15
prefetch: make char explicitly signed

bneradt:
The plugins/prefetch/evaluate.cc sub() function assumed that char would be signed. char is signed for most x86 systems but not for all systems, such as arm64. This makes the variable explicitly signed.

Fixes #10047

PR #10062, opened 2023-07-18
prefetch: fix evaluate for unsigned char platforms

moonchen:
Use int8_t instead of char to ensure that we don't underflow on platforms where char is unsigned.


bneradt:
Looks like a duplicate of: https://github.com/apache/trafficserver/pull/10048

This should be merged into master already.

PR #10053, opened 2023-07-17
Bump minimum CMake version from 3.20 to 3.26

JosiahWI:
Kitware's FindImageMagick module provides imported targets since version 3.26. One of the experimental plugins depends on ImageMagick, so it would be nice to have those imported targets available.

PR #10065, opened 2023-07-19
Revert "Bump minimum CMake version from 3.20 to 3.26"

bneradt:
Reverts apache/trafficserver#10053

We'll change this to a range from 3.20-3.26 since we decided to keep support for at least 3.20 to maintain support for older OS's.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@
 #
 #######################
 
-cmake_minimum_required(VERSION 3.26)
+cmake_minimum_required(VERSION 3.20)
 project(ats VERSION 10.0.0)
 
 set(TS_VERSION_MAJOR ${PROJECT_VERSION_MAJOR})


bneradt:
> We discussed this in the PR review meeting and I was told it was easy to upgrade to CMake 3.26. What changed?

It is easy to upgrade. But we decided it was best to still keep things at 3.20 which is what rockylinux:8 has, so that should be both reasonable for cmake dev and convenient for users.

randall (approved):

JosiahWI:
> What is the intention of allowing a range instead of requiring just a minimum version? (sounds more permissive, but I'm not sure what is actually allowed by that)

We want to test with our minimum version in CI, but we also don't want things to break when we increase the minimum version in the future. Setting a version range should set policies to `NEW` in newer CMake versions, so we'll know if a policy change causes an issue.

One note on this: I just ran into a case where a version range apparently had no effect on policies. I don't know whether there's something I was missing, or whether it's a CMake bug of some sort.

maskit:
> Setting a version range should set policies to NEW in newer CMake versions, so we'll know if a policy change causes an issue

I'm not familiar enough with cmake to understand this part, but my understanding is that the motivation is basically the same as running CI jobs with C++17 and also C++20 where we only use C++17 features at the movement but ensure nothing won't be broken when we move to C++20.. Sounds reasonable to me.

maskit (approved):

PR #10085, opened 2023-07-25
Fix proxy dependency tree in CMake build

JosiahWI:
viz


cmcfarlen (approved):

ywkaras:
What do the dashed vs. solid arrows mean in the diagrams?  I assume that double borders around module nodes man they are in dynamically-linked libraries?

ywkaras:
Can diagrams like these be automatically generated with some tool?

cmcfarlen:
https://cmake.org/cmake/help/latest/module/CMakeGraphVizOptions.html

`The different dependency types PUBLIC, INTERFACE and PRIVATE are represented as solid, dashed and dotted edges.`

Issue #9800, opened 2023-06-08
Manage QUICStream instances without relying on quiche


maskit:
`quiche_conn_stream_init_application_data` and `quiche_conn_stream_application_data` will be unavailable on next version of quiche. Actually I found there is a case that we cannot rely on them and I was going to stop using it.

maskit:
We cannot close this issue. We need to modify our code so we do not use the removed functions.

Issue #10061, opened 2023-07-18
Can't build plugins on Mac


maskit:
I see this build errors on my Mac. Clean build didn't work.
```
Making all in plugins
  CXXLD    esi/libesicore.la
  CXXLD    esi/libtest.la

*** Warning: Linking the shared library esi/libtest.la against the loadable module
*** libesicore.so is not portable!
*** Warning: lib libesicore.so is a module, not a shared library

*** And there doesn't seem to be a static archive available
*** The link will probably fail, sorry
ld: can't link with bundle (MH_BUNDLE) only dylibs (MH_DYLIB) file 'esi/.libs/libesicore.so' for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [esi/libtest.la] Error 1
```

```
*** Warning: Linking the shared library experimental/sslheaders/sslheaders.la against the loadable module
*** libsslhdr.so is not portable!
*** Warning: lib libsslhdr.so is a module, not a shared library

*** And there doesn't seem to be a static archive available
*** The link will probably fail, sorry
```

bryancall:
I was able to get ATS to build with autotools on the mac, but it wants to load the wrong libcrypto.dylib on startup:

```
16:15:20 chaos:(master)~/dev/apache/trafficserver/build-Darwin_gcc_h3$ /opt/ats/bin/traffic_server -T quic
dyld[2790]: Symbol not found: _ASN1_STRING_get0_data
  Referenced from: <A471AA6C-2E17-311B-8A63-BE4F5ECA52A8> /opt/ats/bin/traffic_server
  Expected in:     <B70E508F-94CD-3330-ACD3-23F309B60F8B> /usr/lib/libcrypto.dylib
Abort trap: 6
```

jpeach:
Building with this:

```
$ cat c.sh
#! /usr/bin/env bash

set -o errexit

make distclean || true

autoreconf -fi
./configure \
    --enable-debug \
    --prefix=/opt/ats \
    --with-openssl=$(brew --prefix openssl) \
    --enable-experimental-plugins \
    --enable-example-plugins \
    --enable-test-tools

make clean
make -j$(getconf NPROCESSORS_ONLN) CC="mj clang" CXX="mj clang++"
make -j$(getconf NPROCESSORS_ONLN) check CC="mj clang" CXX="mj clang++"
mj --collate

find . -name *.db.json | xargs rm
```

Fails in the tests:
```
Making check in remap
[...]

PR #10087, opened 2023-07-25
Several fixes around the H3 and QUICHE build.

brbzull0:
This PR tackles a few issues:

1 - ATS wasn't built with quiche. (TS_HAS_QUICHE=0)

Fixing the `rpath`  in the  `quiche.m4` solves this issue as the autoconf now successfully detects quiche, so it does define the macro we need.

2 - Build with the latest quiche.
This was a problem because quiche introduced [this](https://github.com/cloudflare/quiche/pull/1537) change which was incompatible with our code.

**This is a short term solution, the proper fix is to update our code to the latest API** This is tracked [here](https://github.com/apache/trafficserver/issues/9800)

```bash
Environment config finished. Running AuTest...
Running Test qlog_quiche:..... Passed
Running Test quic_no_activity_timeout:... Passed
Running Test quic_poll_timeout:.. Passed

Total of 3 test
  Unknown: 0
  Exception: 0
  Failed: 0
  Warning: 0
  Skipped: 0
  Passed: 3

```

3 - There were few issues when building h3 related to the new libtscore archive.

### DIFF
--- a/m4/quiche.m4
+++ b/m4/quiche.m4
@@ -58,7 +58,7 @@ if test "$has_quiche" != "0"; then
   quiche_have_libs=0
   if test "$quiche_base_dir" != "/usr"; then
     TS_ADDTO(CPPFLAGS, [-I${quiche_include}])
-    TS_ADDTO(LDFLAGS, [-L${quiche_ldflags} -rpath ${quiche_ldflags}])
+    TS_ADDTO(LDFLAGS, [-L${quiche_ldflags} -Wl,-rpath,${quiche_ldflags}])
     TS_ADDTO(LIBS, [-lquiche])
     TS_ADDTO_RPATH(${quiche_ldflags})
   fi
--- a/iocore/net/quic/Makefile.am
+++ b/iocore/net/quic/Makefile.am
@@ -62,6 +62,8 @@ test_MTHashTable_LDADD = \
 	$(top_builddir)/iocore/eventsystem/libinkevent.a \
 	$(top_builddir)/src/tscore/libtscore.a \
 	$(top_builddir)/src/tscpp/util/libtscpputil.la \
+	$(LIBPCRE) \
+	$(LIBCAP) \
 	@SWOC_LIBS@ @OPENSSL_LIBS@
 
 test_MTHashTable_SOURCES = test_MTHashTable.cc
--- a/proxy/http3/Makefile.am
+++ b/proxy/http3/Makefile.am
@@ -72,10 +72,11 @@ test_LDADD = \
   $(top_builddir)/iocore/net/TLSKeyLogger.o \
   $(top_builddir)/iocore/eventsystem/libinkevent.a \
   $(top_builddir)/src/records/librecords_p.a \
-  $(top_builddir)/src/tscore/libtscore.a \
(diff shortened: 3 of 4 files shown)


m4/quiche.m4
```diff
@@ -58,7 +58,7 @@ if test "$has_quiche" != "0"; then
   quiche_have_libs=0
   if test "$quiche_base_dir" != "/usr"; then
     TS_ADDTO(CPPFLAGS, [-I${quiche_include}])
-    TS_ADDTO(LDFLAGS, [-L${quiche_ldflags} -rpath ${quiche_ldflags}])
+    TS_ADDTO(LDFLAGS, [-L${quiche_ldflags} -Wl,-rpath,${quiche_ldflags}])
```
maskit, line 61: Although this way of passing rpath looks more familiar, why do we need to do this where we have `TS_ADDTO_RPATH` below?

maskit: After a bit of googling, it seems like `-R`, which `TS_ADDTO_RPATH` adds, is not a common way. I feel like we should modify `TS_ADDTO_RPATH` to add `-Wl,-rpath` instead of `-R` (or we can add the both if needed).

In any case, this actually solves the weird link issue I faced (https://github.com/apache/trafficserver/issues/10061).

bryancall:
```
diff --git a/iocore/net/quic/Makefile.am b/iocore/net/quic/Makefile.am
index aeb98a391..fce167482 100644
--- a/iocore/net/quic/Makefile.am
+++ b/iocore/net/quic/Makefile.am
@@ -62,6 +62,8 @@ test_MTHashTable_LDADD = \
        $(top_builddir)/iocore/eventsystem/libinkevent.a \
        $(top_builddir)/src/tscore/libtscore.a \
        $(top_builddir)/src/tscpp/util/libtscpputil.la \
+  $(LIBPCRE) \
+  $(LIBCAP) \
        @SWOC_LIBS@ @OPENSSL_LIBS@

 test_MTHashTable_SOURCES = test_MTHashTable.cc
```

m4/quiche.m4
brbzull0: yes, makes sense, I think we can land this and then I will update the `common.m4` to do it right. https://github.com/apache/trafficserver/pull/10096

brbzull0:
> With this added change the make -j check passes.
> 
> ```
> diff --git a/iocore/net/quic/Makefile.am b/iocore/net/quic/Makefile.am
> index aeb98a391..fce167482 100644
> --- a/iocore/net/quic/Makefile.am
> +++ b/iocore/net/quic/Makefile.am
> @@ -62,6 +62,8 @@ test_MTHashTable_LDADD = \
>         $(top_builddir)/iocore/eventsystem/libinkevent.a \
>         $(top_builddir)/src/tscore/libtscore.a \
>         $(top_builddir)/src/tscpp/util/libtscpputil.la \
> +  $(LIBPCRE) \
> +  $(LIBCAP) \
>         @SWOC_LIBS@ @OPENSSL_LIBS@
> 
>  test_MTHashTable_SOURCES = test_MTHashTable.cc
> ```

Thanks Bryan, just added that + other stuffs.

maskit (approved):

PR #10102, opened 2023-07-26
Add quiche openssl lib to LDFLAGS

ezelkow1:
Building in my ubuntu 20.04 VM I ran into an issue where due to the order of the -L directives the build kept trying to pull openssl from the system library locations first, rather than the specified path before the library link. Adding the quiche openssl location to the LDFLAGS ensured it showed up first in the library locations and allowed the build to complete

### DIFF
--- a/tools/build_h3_tools.sh
+++ b/tools/build_h3_tools.sh
@@ -247,7 +247,7 @@ fi
   PKG_CONFIG_PATH=${BASE}/lib/pkgconfig:${OPENSSL_LIB}/pkgconfig \
   CFLAGS="${CFLAGS}" \
   CXXFLAGS="${CXXFLAGS}" \
-  LDFLAGS="${LDFLAGS}" \
+  LDFLAGS="${LDFLAGS} -L${OPENSSL_LIB}" \
   --enable-http3 \
   ${ENABLE_APP}
 ${MAKE} -j ${num_threads}


ezelkow1:
Before the change: 
`g++ -std=c++14 -O3 -g -Wl,-rpath -Wl,/opt/openssl-quic-openssl-3.0.9+quic/lib64 -o .libs/nghttp util.o http2.o timegm.o app_helper.o nghttp2_gzip.o nghttp.o HtmlParser.o tls.o -pthread  ../lib/.libs/libnghttp2.so ../third-party/.libs/liburl-parser.a ../third-party/.libs/libllhttp.a -ljemalloc -L/usr/lib/x86_64-linux-gnu -lxml2 -lev -L/opt/lib /opt/lib/libnghttp3.so /opt/lib/libngtcp2_crypto_openssl.so /opt/lib/libngtcp2.so -L/opt/openssl-quic-openssl-3.0.9+quic/lib64 -lssl -lcrypto -lcares -lsystemd -ljansson -lz -ldl -pthread -Wl,-rpath -Wl,/opt/lib`

quic openssl is added to the rpath via the LDFLAGS but then the ordering of locations is -Lsystem -L/opt/lib -L/opt/quic-openssl. So my build was always pulling the ssl lib from the system location. This adds a -L/opt/quic-openssl to LDFLAGS to ensure it's first

bneradt (approved):

PR #10103, opened 2023-07-26
autest get_port: update to use psutil

bneradt:
This updates the get_port logic to pre-filter possible ports via psutil. This should work quite effectively on Linux systems which can gather all listening ports on the system via that module. On Mac, psutil is limited to the user's processes, but it should still help there. Note that the previous logic is kept, it's just that whatever ports are listening via psutil are not considered available before using that logic.

I'm hopeful this will make some of the autests in CI more reliable.

### DIFF
--- a/tests/gold_tests/autest-site/ports.py
+++ b/tests/gold_tests/autest-site/ports.py
@@ -20,6 +20,7 @@ import socket
 import subprocess
 import os
 import platform
+import psutil
 
 import hosts.output as host
 
@@ -37,12 +38,18 @@ class PortQueueSelectionError(Exception):
     pass
 
 
-def PortOpen(port, address=None):
+def PortOpen(port: int, address: str = None, listening_ports: set[int] = None) -> bool:
     """
     Detect whether the port is open, that is a socket is currently using that port.
 
     Open ports are currently in use by an open socket and are therefore not available
     for a server to listen on them.
+
+    Args:
+        port: The port to check.
+        address: The address to check. Defaults to localhost.
+        listening_ports: A set of ports that are currently listening. If a port
+            is in this set, it is considered open.
 
     Returns:
         True if there is a connection currently listening on the port, False if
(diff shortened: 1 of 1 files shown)


tests/gold_tests/autest-site/ports.py
```diff
@@ -115,6 +127,30 @@ def _get_available_port(queue):
...
+    except psutil.AccessDenied:
+        # Mac OS X doesn't allow net_connections() to be called without root.
+        for proc in psutil.process_iter(['pid', 'name']):
+            try:
+                connections = proc.connections(kind='all')
+            except (psutil.AccessDenied, psutil.NoSuchProcess):
+                continue
+            for conn in connections:
+                if conn.status == psutil.CONN_LISTEN:
+                    ports.add(conn.laddr.port)
```
cmcfarlen, line 150: not familiar with python and scoping, but to me this for loop looks odd outside the try that assigns to connections

bneradt: Thank you for checking. Yes, Python is different that way with scoping. Variables are scoped by function, not by inner control blocks within a function. Thus `connections` in that try are accessible (in scope) outside of the try in the loop.

https://stackoverflow.com/a/2829642/629530

bneradt: Note that this is a handy feature, btw. You don't have to "declare" a variable in python outside of code blocks to access them later. Thus this is fine:

```python3
╰─➤  cat e.py       
#!/usr/bin/env python3

i = 3
if i > 0:
    connections = "one two three"
else:
    connections = 4

print(connections)
```

You don't have to explicitly declare `connections` earlier with an otherwise useless assignment:

```python3
╰─➤  cat e.py       
#!/usr/bin/env python3

i = 3
connections = None # <---- Not needed.
if i > 0:
    connections = "one two three"
else:
    connections = 4

print(connections)
```

zwoop:
Cherry-picked to v9.2.x

PR #10107, opened 2023-07-27
Fix ports.py type hint for sets on older Python

bneradt:
Older versions of Python do not support set[int] as a type hint. Using typing.Set to accomodate them. The offending type hint was introduced by PR #10103


cmcfarlen (approved):

zwoop:
Cherry-picked to v9.2.x

PR #10110, opened 2023-07-28
Check for SSL_CTX_set_client_hello_cb in CMake

JosiahWI:
This checks for the symbol in ssl.h and sets TS_USE_HELLO_CB if found.


bneradt (commented):
Looks good.

cmcfarlen (approved):

PR #10114, opened 2023-07-29
Pull INKAPIInternal out of InkAPI.cc as inkcont

JosiahWI:
This is the first of three parts of a refactoring to break out FeatureAPIHooks, making it possible to unit test hook callbacks.

This moves the implementation of INKAPIInternal from InkAPI.cc to InkAPIInternal.cc, still in the traffic_server directory. I wrapped it in a new library, inkcont, as a convenient way to document its dependencies and link it into unit tests.

This was by no means an easy refactor, and the result seems far from perfect. I would really appreciate feedback. The biggest roadblock was that parts of this proxy that this depends on also depend on `HostStatus`. This particular arrangement happens to work for unit testing so far, I think because the API hooks don't use any of those functions and thus the linker doesn't look for their definitions. I tried to move the source file to `tsapi` instead, but could not get it to compile without `HostStatus`.


JosiahWI:
[approve ci autest]

JosiahWI:
I misstyped INKContInternal as InkAPIntertal. Fixed.

src/traffic_server/CMakeLists.txt
```diff
@@ -15,6 +15,16 @@
 #
 #######################
 
+add_library(inkcont InkContInternal.cc)
```
cmcfarlen, line 18: I think we should be explicit about STATIC here to avoid surprises if anyone sets BUILD_SHARED_LIBS

cmcfarlen (changes requested):
Mark inkcont library STATIC

JosiahWI:
Force pushed the requested change to make inkcont explicitly `STATIC`. My local branch is ahead of the remote, and I wanted to move the commit back in the history, but I accidentally squashed it instead out of habit.

JosiahWI:
[approve ci fedora]

cmcfarlen (approved):

Issue #7096, opened 2020-08-10
Crash due to bug in Thread Context migration


sudheerv:
Ran into the below crash in ATS9 after a couple of weeks of running stable. This crash happens about every other minute on some random box out of the 100s of boxes we've in production. 

```
(gdb) bt
#0  0x00002b74100ed31e in asn1_string_embed_free () from /lib/libcrypto.so.1.1
#1  0x00002b74100f8eaf in asn1_primitive_free () from /lib/libcrypto.so.1.1
#2  0x00002b74100f8e89 in asn1_primitive_free () from /lib/libcrypto.so.1.1
#3  0x00002b74100f9240 in asn1_template_free () from /lib/libcrypto.so.1.1
#4  0x00002b74100f8fde in asn1_item_embed_free () from /lib/libcrypto.so.1.1
#5  0x00002b74100f9240 in asn1_template_free () from /lib/libcrypto.so.1.1
#6  0x00002b74100f8fde in asn1_item_embed_free () from /lib/libcrypto.so.1.1
#7  0x00002b74100f9240 in asn1_template_free () from /lib/libcrypto.so.1.1
#8  0x00002b74100f8fde in asn1_item_embed_free () from /lib/libcrypto.so.1.1
#9  0x00002b74100f9240 in asn1_template_free () from /lib/libcrypto.so.1.1
#10 0x00002b74100f8fde in asn1_item_embed_free () from /lib/libcrypto.so.1.1
#11 0x00002b74100f9185 in ASN1_item_free () from /lib/libcrypto.so.1.1
#12 0x00002b7410246f11 in OPENSSL_sk_pop_free () from /lib/libcrypto.so.1.1
#13 0x00002b740fe03eca in SSL_SESSION_free () from /lib/libssl.so.1.1
#14 0x00002b740fdfdbba in SSL_free () from /lib/libssl.so.1.1
#15 0x0000000000736c78 in SSLNetVConnection::clear (this=0x2b7592c85aa0) at SSLNetVConnection.cc:928
#16 0x0000000000737374 in SSLNetVConnection::free (this=0x2b7592c85aa0, t=0x2b7413c06000) at SSLNetVConnection.cc:972
#17 0x000000000075d2af in NetHandler::free_netevent (this=0x2b7413c0a0a0, ne=ne@entry=0x2b7592c85c50) at UnixNet.cc:364
[...]

PR #10117, opened 2023-07-31
HttpSessionManager: add global_locked pool option.

traeak:
Use only when transaction loads starve the session pool to the point that the pool is mostly bypassed.
I'm not sure this is that useful unless someone has severe congestion issues.  Performance is not negatively impacted if the global locked isn't used.

NOTE: the critical section of the session pool was expanded due to the PR associated with this issue:  https://github.com/apache/trafficserver/issues/7096


doc/admin-guide/files/records.yaml.en.rst
```diff
@@ -1067,24 +1067,36 @@ mptcp
...
+
+   For a hybrid pool, the operation starts as the global pool, but sessons
+   are returned to the local thread pool if the global pool lock is not
+   acquired rather than just closing the origin connection as is the
+   case in standard global mode.
+
+   For a ``global_locked`` pool connections are managed by a blocking
+   mutex instead of the normal try mutex.  Under extreme transaction
+   loads the connection pool starvation may result in most transactions
+   bypassing the connection pool resulting in runaway upstream
```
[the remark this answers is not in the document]
traeak: retrying the lock would result in even further pool starvation.

traeak:
[approve ci rocky]

doc/admin-guide/files/records.yaml.en.rst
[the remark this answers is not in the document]
traeak: retry on that lock would just cause congestion to go critical that much faster.

proxy/http/HttpSessionManager.cc
```diff
@@ -146,52 +146,75 @@ ServerSessionPool::acquireSession(sockaddr const *addr, CryptoHash const &hostna
...
+    // This is broken out because only in this case do we check
+    // the host hash first. The range must be checked to verify
+    // an upstream that matches port and SNI name is selected.
+    // Walk backwards to select oldest.
+    // Prefer a session that's already on this thread.
+    in_port_t const port = ats_ip_port_cast(addr);
+    auto iter            = m_fqdn_pool.find(hostname_hash);
+    while (iter != m_fqdn_pool.end() && iter->hostname_hash == hostname_hash) {
+      Debug("http_ss", "Compare port 0x%x against 0x%x", port, ats_ip_port_cast(iter->get_remote_addr()));
+      auto *const netvc = dynamic_cast<UnixNetVConnection *>(iter->get_netvc());
```
masaori335, line 164: Do we need to compare `NetHandler`? Comparing thread like below (avoiding dynamic_cast) is enough? 
```
NetVConnection *const netvc = iter->get_netvc();
...
  if (netvc->thread == ethread) {
```

traeak: got it thanks.

proxy/http/HttpSessionManager.cc
```diff
@@ -146,52 +146,75 @@ ServerSessionPool::acquireSession(sockaddr const *addr, CryptoHash const &hostna
...
-          zret = HSM_DONE;
-          break;
+      while (iter != m_ip_pool.end() && ats_ip_addr_port_eq(iter->get_remote_addr(), addr)) {
+        auto *const netvc = iter->get_netvc();
+        if ((!(match_style & TS_SERVER_SESSION_SHARING_MATCH_MASK_HOSTONLY) || iter->hostname_hash == hostname_hash) &&
+            (!(match_style & TS_SERVER_SESSION_SHARING_MATCH_MASK_SNI) || validate_sni(sm, netvc)) &&
+            (!(match_style & TS_SERVER_SESSION_SHARING_MATCH_MASK_HOSTSNISYNC) || validate_host_sni(sm, netvc)) &&
+            (!(match_style & TS_SERVER_SESSION_SHARING_MATCH_MASK_CERT) || validate_cert(sm, netvc))) {
+          // Check if on same thread
+          if (netvc->thread == ethread) {
```
shinrich, line 203: How is this only triggered in the global locked case?

shinrich:
I like the idea of this.  One question in the code.

How did this compare to hybrid?  Fewer connections to origin opened?

Did you see performance degradation due to walking longer session lists in the global_locked vs global case?

proxy/http/HttpSessionManager.cc
traeak: it's not this one is actually a change to the pools in general.  if necessary i could split this into 2 prs.

traeak:
I'm going to run performance tests on one of our heavily loaded linear only
caches in the near future to see how it performs.  I still need to check
the performance difference since removing the dynamic cast for each list
item.

proxy/http/HttpSessionManager.cc
traeak: done and tested.

traeak:
I ran some performance tests and the session pool walking has a negative affect if there's a large number of matching pool entries.  I will cut that number back to only checking the first few (whatever that number is, 4 or 8) for same thread session, otherwise fall back to taking the first (most idle) session.

traeak:
> I like the idea of this. One question in the code.
> 
> How did this compare to hybrid? Fewer connections to origin opened?
> 
> Did you see performance degradation due to walking longer session lists in the global_locked vs global case?

Walking the list can hurt performance.  I will test using a walk threshold and generate a new PR if it's worth it.

masaori335 (approved):
Having an option of  blocking mutex instead of try lock looks reasonable because it's faster in some cases. Let's land this and add `prefer session on same thread` feature later.

PR #10122, opened 2023-08-01
Remove unused cache Store code.

jpeach:
Quite a few of the Store methods are not used by anything. Perhaps there was once a need for these APIs, but I expect that they are unlikely to be needed any more.


masaori335 (approved):
+1 for removing dead code.

PR #10135, opened 2023-08-02
TXN var config string set. Add nullptr check for non-specific variable string handling.

brbzull0:
Without this we will try to store a nullptr which will cause a crash.


brbzull0:
[approve ci autest]

masaori335 (approved):
Looks reasonable

brbzull0:
[approve ci autest]

PR #10136, opened 2023-08-02
cmake: improve installing config files

cmcfarlen:
This (hopefully) follows the auto tools scheme for installing but not overwriting config files.


cmake/install_configs.cmake
```diff
@@ -0,0 +1,33 @@
...
+
+# This is a script intended to be passed to install(SCRIPT ...)
+# This copies default config files to the destination  without overwriting existing files.
+# If the config file already exists, the source config is copied with the '.default' extension
+
+file(GLOB CONFIG_FILES ${CONFIG_SOURCE_GLOBS})
+file(MAKE_DIRECTORY ${CONFIG_DEST_PATH})
+foreach(CONFIG_FILE ${CONFIG_FILES})
+  cmake_path(GET CONFIG_FILE STEM LAST_ONLY CONFIG_FILE_NAME)
+  if (EXISTS "${CONFIG_DEST_PATH}/${CONFIG_FILE_NAME}")
```
JosiahWI, line 26: `"${CONFIG_DEST_PATH}/${CONFIG_FILE_NAME}"` appears 3 times. Would it look better with a temporary variable to remove duplication?

cmake/install_configs.cmake
```diff
@@ -0,0 +1,33 @@
...
+#######################
+
+# This is a script intended to be passed to install(SCRIPT ...)
+# This copies default config files to the destination  without overwriting existing files.
+# If the config file already exists, the source config is copied with the '.default' extension
+
+file(GLOB CONFIG_FILES ${CONFIG_SOURCE_GLOBS})
+file(MAKE_DIRECTORY ${CONFIG_DEST_PATH})
+foreach(CONFIG_FILE ${CONFIG_FILES})
+  cmake_path(GET CONFIG_FILE STEM LAST_ONLY CONFIG_FILE_NAME)
```
JosiahWI, line 25: Why do we throw out the file extension? I seem to recall config files that AuTest copies into its sandbox keeping their extension (e.g. "sni.yaml", not "sni")

JosiahWI, line 25: Should we do a sanity check to make sure `CONFIG_FILE_NAME` isn't an empty string?

> If a requested component is not present in the path, an empty string will be stored in <out-var>.

I'm pretty sure the current implementation will error out later if this happens, but nobody wants the config directory to be accidentally wiped in an unexpected edge case.

cmcfarlen: They all have the 'default' extension in the source tree.

cmcfarlen: I'm not sure they can be empty based on the glob of "*.default"

cmake/install_configs.cmake
cmcfarlen: yeah, that will get rid of the else too!

cmake/install_configs.cmake
JosiahWI: Ah, that's good then. Maybe it would help to add a comment since that context is missing otherwise.

JosiahWI: Probably safe then. And I did think through it more carefully and I am very confident that even if it is empty it will not do anything crazy.

JosiahWI (changes requested):
Please add a comment pointing out that the `cmake_path` directive is intended to remove the `.default` extension present on all the configs in the source tree.

JosiahWI (approved):
Looks good. Thank you for this fix.

bneradt (approved):

Issue #10137, opened 2023-08-02
ATS forwards malformed empty `Content-Length` headers, in violation of RFC 9110


kenballus:
## Description
RFC 9110 defines the following ABNF rule for the acceptable values in a `Content-Length` header:
```
Content-Length = 1*DIGIT
```
(i.e. 1 or more ASCII digits, so empty values are invalid)

A few paragraphs later, it says this:
> Likewise, a sender MUST NOT forward a message with a Content-Length header field value that does not match the ABNF above, with one exception: a recipient of a Content-Length header field value consisting of the same decimal value repeated as a comma-separated list (e.g, "Content-Length: 42, 42") MAY either reject the message as invalid or replace that invalid field value with a single instance of the decimal value, since this likely indicates that a duplicate was generated or combined by an upstream message processor.

Thus, ATS is obligated to reject (or at least, not forward) requests that do not match that ABNF rule.

However, ATS currently forwards empty `Content-Length` headers as-is, which is a violation of the RFC.

## To reproduce
1. Build and start ATS master on port 80, along with a logging webserver of your choice on port 8000.
2. Send the following request to port 80:
```
GET / HTTP/1.1\r\n
Host: whatever\r\n
Content-Length: \r\n
\r\n
```
3. Observe that it forwards the invalid CL header as-is to your backend server, without correction or rejection.

### /usr/local/etc/trafficserver/records.yaml
```yaml
ts:
  http:
    server_ports: 80 80:ipv6
```

### /usr/local/etc/trafficserver/remap.config
```
map / http://127.0.0.1:8000
```

## Versions
ATS: master branch, commit e6182d9ac9c3f611cb33b3ef6dc98327df41c3d6
OS: Debian Bookworm container running on top of Arch Linux with a 6.4.3 kernel

masaori335:
Thanks for the report. I agree that the empty value Content-Length header is a malformed header.

PR #10144, opened 2023-08-03
Strictly follow Content-Length header ABNF rule

masaori335:
Address #10137. 

[RFC 9110 8.6. Content-Length](https://datatracker.ietf.org/doc/html/rfc9110#section-8.6)

```
Content-Length = 1*DIGIT
```

If the value filed is empty, it should be handled as a malformed header.

### DIFF
--- a/proxy/hdrs/HTTP.cc
+++ b/proxy/hdrs/HTTP.cc
@@ -27,6 +27,7 @@
 #include <cassert>
 #include <cstdio>
 #include <cstring>
+#include <string_view>
 #include "HTTP.h"
 #include "HdrToken.h"
 #include "tscore/Diags.h"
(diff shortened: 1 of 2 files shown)


proxy/hdrs/HTTP.cc
```diff
@@ -1264,9 +1264,14 @@ validate_hdr_content_length(HdrHeap *heap, HTTPHdrImpl *hh)
     int content_length_len         = 0;
     const char *content_length_val = content_length_field->value_get(&content_length_len);
 
-    // RFC 7230 section 3.3.2
+    // RFC 9110 section 8.6.
     // Content-Length = 1*DIGIT
     //
+    if (content_length_len == 0) {
```
jpeach, line 1270: `content_length_len` is int, so while `value_get` doesn't actually ever return negative, people reading the call sires still have to keep that i mind.

Could we remove this version of `value_get` so that the unambiguous `string_view` version is the only version?

masaori335: Changed to `std::string_view`. PTAL.

jpeach (approved):
Looks good. Would you mind making a follow-up PR to remove the non-sotringview version of value_get?

masaori335:
I'll file an issue to deprecate the `value_get` at least. It'll be a big change because the function is widely used.

PR #10146, opened 2023-08-03
Unrequire remap rules for OCSP

maskit:
The change to use FetchSM for OCSP requests (#9591) unintentionally made ATS rely on remap rules, and that effectively broke OCSP if a user sets `remap_required` to 1. To recover the original behavior which does not rely on remap rules, this introduces a new flag for FetchSM, `TS_FETCH_FLAGS_SKIP_REMAP`, and the flag enables skipping remap on a transaction initiated by `FetchSM` even if `remap_required` is set to 1.

Since the flag is part of TS API, this also enables plugins to make HTTP requests for other servers without remap rules (if FetchSM is used).


zwoop:
I think I understand now (Thanks to @maskit ). The issue is that when setting up the FetchSM, there's no way to signal the HttpSM that it shouldn't require remap. The existing configurations and APIs works on the transactions, which you don't have yet when setting up the FetchSM (obviously).

bneradt (approved):

PR #10149, opened 2023-08-03
Fix the libloader plugin makefile.

jpeach:
The dangling header name doesn't break the build now, but since automake moves statements around, perturbing things by adding a new plugin can lead to invalid build rules.


jpeach:
@keesspoelstra

randall (approved):

zwoop:
Cherry-picked to v9.2.x

PR #10150, opened 2023-08-04
Eliminate function-scope thread_local in Regex.cc.

ywkaras:
It appears to be the case that, in the C/C++ runtime for Linux, there is a mutex that is locked both:
1)  When initializing non-local variables in a shared library, while it is being loaded by calling dlopen(), and
2)  When execution encounters the definition of a thread_local variable within a function.  It seems that the mutex is locked even if the variable has already been initialized, presumably to check whether it was initialized.

This has been causing problems for the DbgCtl class.  When an instance of DbgCtl is initialized, it locks a mutex (for the registry of instances).  The instance initialization then (indirectly) calls a function (in Regex.cc) that defines a thread_local varaiable.  The result is scenarios that sometimes lead to deadlock.  For example:
1)  dlopen() is called to load a plugin.  The runtime's mutex is taken.
2)  Another thread starts initializing a DbgCtl instance, and takes the instance registry mutex.
3)  The non-local initialization of the shared library starts the intialization of another DbgCtl instance.  This thread now blocks trying to lock the registry mutex.
4)  The other thread now enters the function in Regex.cc that defines a thread_local object.  It blocks on the runtime's mutex, which it needs to lock in order to do/check the thread_local instances.  Deadlock.

This PR fixes the problem, by eliminating the thread_local variable that's defined in a function.

### DIFF
--- a/src/tscore/Regex.cc
+++ b/src/tscore/Regex.cc
@@ -29,22 +29,31 @@
 #include "tscore/Regex.h"
 
 #ifdef PCRE_CONFIG_JIT
-static pcre_jit_stack *
-get_jit_stack(void *data ATS_UNUSED)
+namespace
 {
-  thread_local struct JitStack {
-    JitStack()
-    {
-      jit_stack = pcre_jit_stack_alloc(ats_pagesize(), 1024 * 1024); // 1 page min and 1MB max
+thread_local pcre_jit_stack *jit_stack;
+
+struct JitStackCleanup {
+  ~JitStackCleanup()
+  {
+    if (jit_stack) {
+      pcre_jit_stack_free(jit_stack);
     }
-    ~JitStack() { pcre_jit_stack_free(jit_stack); }
+  }
+};
+thread_local JitStackCleanup jsc;
 
-    pcre_jit_stack *jit_stack = nullptr;
-  } stack;
-
-  return stack.jit_stack;
+pcre_jit_stack *
+get_jit_stack(void *)
+{
+  if (!jit_stack) {
+    jit_stack = pcre_jit_stack_alloc(ats_pagesize(), 1024 * 1024); // 1 page min and 1MB max
+  }
+  return jit_stack;
 }
-#endif
+
+} // end anonymous namespace
+#endif // def PCRE_CONFIG_JIT
 
 Regex::Regex(Regex &&that) noexcept : regex(that.regex), regex_extra(that.regex_extra)
 {


ywkaras:
This PR is an alternative to:  https://github.com/apache/trafficserver/pull/10143

cmcfarlen:
While not quite the same code, we did try to move the thread local outside of the function and it did still deadlock when @bneradt tested it.  We just moved the thread_local struct outside like so:

```
#ifdef PCRE_CONFIG_JIT
thread_local struct JitStack {
  JitStack()
  {
    jit_stack = pcre_jit_stack_alloc(ats_pagesize(), 1024 * 1024); // 1 page min and 1MB max
  }
  ~JitStack() { pcre_jit_stack_free(jit_stack); }

  pcre_jit_stack *jit_stack = nullptr;
} stack;

static pcre_jit_stack *
get_jit_stack(void *data ATS_UNUSED)
{
  return stack.jit_stack;
}
#endif
```

Maybe having a thread_local raw pointer behaves differently.

cmcfarlen:
Using two thread locals avoids the deadlock because without the thread local object access, `get_jit_stack` doesn't call the TLS init function which ends up calling __cxx_thread_atexit(which locks the dl_whatever mutex).  Since the raw pointer doesn't have a destructor to call, it doesn't need to call this.  Interestingly, `get_jit_stack` was calling the TLS init function to setup the destructor call at thread exit whether or not the class was declared in the function body.

So really it was the destructor that was causing this.

I like this solution better than the other one. 👍   Thanks!

bneradt:
Looks great to me, @ywkaras. I've verified in CI that with this patch we no longer experience the deadlock when starting ATS with TSVConnFd.

@cmcfarlen : since I'm also with Yahoo, I'll let you approve if you think this is OK.

cmcfarlen (approved):
Thanks @ywkaras and @bneradt for creating and confirming this.

jpeach:
> 



> Using two thread locals avoids the deadlock because without the thread local object access, `get_jit_stack` doesn't call the TLS init function which ends up calling __cxx_thread_atexit(which locks the dl_whatever mutex). Since the raw pointer doesn't have a destructor to call, it doesn't need to call this. Interestingly, `get_jit_stack` was calling the TLS init function to setup the destructor call at thread exit whether or not the class was declared in the function body.
> 
> So really it was the destructor that was causing this.
> 
> I like this solution better than the other one. 👍 Thanks!

Can you please add this explanation in a code comment to make it easier for future maintainers to understand this change?

PR #10151, opened 2023-08-05
[Fuzzing] add fuzzing test

0x34d:
Moving the fuzzing harness from [oss-fuzz](https://github.com/google/oss-fuzz/tree/master/projects/trafficserver/fuzzer) to the main project for better management.

Note:-

`fuzz_hpack` and `fuzz_http` are not working under UndefinedBehaviorSanitizer.

Error:
```
/home/0x34d/project/trafficserver/src/tscore/ink_queue.cc:165:3: runtime error: member call on null pointer of type 'Diags'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/legion/0x34d/trafficserver/src/tscore/ink_queue.cc:165:3 in 
```

So i removed both targets from `UBSan` https://github.com/google/oss-fuzz/pull/10800/files#diff-45181139e12414262e5dd1b734e02937a00ea123a7b2e66ff146b3a9786b7ce3R38-R42


shukitchan:
[approve ci]

CMakeLists.txt
```diff
@@ -323,6 +324,13 @@ set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 add_compile_definitions(${HOST_OS} PACKAGE_NAME="ats" PACKAGE_VERSION="${TS_VERSION_STRING}")
 add_compile_options($<$<COMPILE_LANGUAGE:CXX>:-Wno-invalid-offsetof>)
 
+# Enable fuzzing
+if(ENABLE_FUZZING)
+    if(NOT CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
JosiahWI, line 329: Why `MATCHES` instead of `STREQUAL`?

0x34d: > Why MATCHES instead of STREQUAL?

Ah My bad, I can change this to `STREQUAL`.

JosiahWI: `MATCHES` is used for regular expression comparisons. If what you really want is a string comparison, then `MATCHES` is doing more than you intended and is sloppy from a performance and readability perspective. It could also introduce subtle unexpected behavior: any compiler that happens to have "Clang" in the name will match. It's also possible that you want that behavior - maybe you want Apple Clang to count. Whichever way you intend to do it, you want to understand why you did it that way and which alternatives you decided against and why.

0x34d: >  It could also introduce subtle unexpected behavior: any compiler that happens to have "Clang" in the name will match. 

I understand, But this is  what I want.

When [AFL](https://github.com/AFLplusplus/AFLplusplus) is used for fuzzing, the compilers are:

```bash
CC=/src/aflplusplus/afl-clang-fast
CXX=/src/aflplusplus/afl-clang-fast++
```

Doing it the other way around will break the AFL fuzzer build.


#### My comment:

> if(NOT CMAKE_CXX_COMPILER_ID MATCHES "Clang")

I don't wan't `AFL` and `honggfuzz`  build to break.

NOTE:  AFL and Honggfuzz dose support GCC, but GCC instrumentation is kind of crap.

JosiahWI: Thanks for thinking through it with me. All the CMake changes look good to me.

ezelkow1:
[approve ci]

shukitchan:
Is this supposed to work together with https://github.com/google/oss-fuzz/pull/10800 ?

0x34d:
> Is this supposed to work together with [google/oss-fuzz#10800](https://github.com/google/oss-fuzz/pull/10800) ?

that's correct.

shukitchan (changes requested):
1) Is fuzz_esi.cc supposed to replace FuzzEsi.cc and fuzz_http.cc to replace FuzzHttp.cc? If yes, perhaps we should copy those tests over. Right now it looks like there are some differences.

2) I tried to rebuild the fuzzer with oss-fuzz and it looks good but the check_build is stuck on fuzz_yamlcpp somehow and can't finish. Is that working for you?

shukitchan:
Also ubsan is running fine with FuzzHTTP. So let's see why it can't be run with fuzz_http.

0x34d:
> 1. Is fuzz_esi.cc supposed to replace FuzzEsi.cc and fuzz_http.cc to replace FuzzHttp.cc? If yes, perhaps we should copy those tests over. Right now it looks like there are some differences.

Alright Done.

> 2. I tried to rebuild the fuzzer with oss-fuzz and it looks good but the check_build is stuck on fuzz_yamlcpp somehow and can't finish. Is that working for you?

Did you try this PR : https://github.com/google/oss-fuzz/pull/10800 ?

> Also ubsan is running fine with FuzzHTTP. So let's see why it can't be run with fuzz_http.

I don't know; the similar error I was facing in `fuzz_proxy_protocol.c` when I added 
```cpp
DiagsPtr::set(new Diags("fuzzing", "", "", nullptr));
delete diags();
```
then it's gone, but in fuzz_hpack and fuzz_http, it's not.

Maybe you can ask your dev team about it?
For me, anything like this ever happens, I blame CPP for its clusterf##k.

shukitchan:
[approve ci centos]

0x34d:
@shukitchan ping.

shukitchan:
I tested it again and it worked fine this time without the problem with fuzz_yamlcpp. 
Can you update and resolve the conflicts? Then I think it is good to go.

0x34d:
[CMakeLists.txt](https://github.com/apache/trafficserver/pull/10151/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR444-R449)
```cmake
add_subdirectory(src/tests)
if(ENABLE_AUTEST)
    add_subdirectory(tests)
endif()
if(ENABLE_FUZZING)
    add_subdirectory(tests/fuzzing)
endif()
add_subdirectory(plugins)
````

Can you check?

shukitchan:
> [CMakeLists.txt](https://github.com/apache/trafficserver/pull/10151/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR444-R449)
> 
> ```cmake
> add_subdirectory(src/tests)
> if(ENABLE_AUTEST)
>     add_subdirectory(tests)
> endif()
> if(ENABLE_FUZZING)
>     add_subdirectory(tests/fuzzing)
> endif()
> add_subdirectory(plugins)
> ```
> 
> Can you check?

@JosiahWI can you confirm again that this change is ok?

JosiahWI (commented):
@shukitchan Yes, all the CMake changes look good to me. Thanks!

Note: I did test the RegEx just to convince myself that it works. There's something that my brain finds odd about "Clang" as a Regex, I guess because it has no RegEx symbols in it.

shukitchan (approved):

PR #10152, opened 2023-08-07
CID 1516688: Fix uninitialized member of AcceptOptions

### LINKED ISSUE #10013: CID 1516688:  Uninitialized members  (UNINIT_CTOR)
```
*** CID 1516688:  Uninitialized members  (UNINIT_CTOR)
/iocore/net/AcceptOptions.h: 95 in AcceptOptions::AcceptOptions()()
89     
90       /// Proxy Protocol enabled
91       bool f_proxy_protocol;
92     
93       /// Default constructor.
94       /// Instance is constructed with default values.
   CID 1516688:  Uninitialized members  (UNINIT_CTOR)
   Non-static class member "defer_accept" is not initialized in this constructor nor in any functions that it calls.
95       AcceptOptions() { this->reset(); }
96       /// Reset all values to defaults.
97       self &reset();
```

masaori335:
Fix #10013

----

Update:

Cleanup AcceptOptions
1. Remove the `reset()` function, because it's only called by the constructor.
2. Remove the AcceptOptions.cc file because it has the `reset()` function only.
3. Set in-class initializers to AcceptOptions members.
`IpAddr local_ip` is an exception. Calling `invalidate()` is the same as the initialization of `IpAddr`. 

https://github.com/apache/trafficserver/blob/f9aa1e0adb7d23cb9d9dd4fb421b80a24654ae85/include/tscore/ink_inet.h#L1302-L1308
https://github.com/apache/trafficserver/blob/f9aa1e0adb7d23cb9d9dd4fb421b80a24654ae85/include/tscore/ink_inet.h#L1317


iocore/net/AcceptOptions.cc
```diff
@@ -37,6 +37,7 @@ AcceptOptions::reset()
   frequent_accept       = true;
   recv_bufsize          = 0;
   send_bufsize          = 0;
+  defer_accept          = 0;
```
jpeach, line 40: Why is there a `reset()` member? It's never called, so could we just use normal constructor initialization?

masaori335: Oh, it's never called? Well, setting default value is better.

masaori335:
[approve ci freebsd]

iocore/net/AcceptOptions.h
```diff
@@ -25,52 +25,54 @@
...
   /// @note Ignored if an explicit incoming address is set in the
   /// the configuration (@c local_ip). If neither is set IPv4 is used.
-  int ip_family;
+  int ip_family = AF_INET;
   /// Should we use accept threads? If so, how many?
-  int accept_threads;
+  int accept_threads = -1;
   /// Event type to generate on accept.
-  EventType etype;
+  EventType etype = ET_NET;
```
jpeach, line 48: I think you could remove this (and not have to include "I_Net.h") because it is never set to anything other than ET_NET.

jpeach (approved):

iocore/net/AcceptOptions.h
```diff
@@ -25,52 +25,54 @@
 #pragma once
 
 #include "tscore/ink_inet.h"
+
 #include "I_Event.h"
+#include "I_Net.h"
```
maskit, line 30: Doesn't this bring a lot of dependency? I think that's the reason we have AcceptOptions.cc separately.

@moonchen What do you think?

maskit (dismissed):
We need to think about I_Net.h. @bryancall and some others are going to look into it.

iocore/net/AcceptOptions.h
bryancall: Yes, I don't see it being set to anything else.  You could get rid of the member variable and just use ET_NET in the code.

moonchen:
I'm for hard-coding the etype to ET_NET.  If we want to keep it as a variable, I suggest we pass in etype from the constructor of AcceptOptions so that ET_NET is not a dependency of AcceptOptions.

masaori335:
Removed `AcceptOptions::etype` and rebased on the latest master.

bryancall (approved):

PR #10157, opened 2023-08-07
Fix KeyEqual param for ConnectingIpPool declaration

wangshichen:
the 4th parameter, `KeyEqual`, is missing in ConnectingIpPool declaration, which may lead to the comparison of two `IpEndpoint` object won't work as we expected. That is, the default one is unlikely to work with `sockaddr` type.
In another way, IpHelper contains the compare operator for IpEndpoint type, so IpHelper should be used as the KeyEqual parameter of ConnectingIpPool declaration.

### DIFF
--- a/proxy/http/ConnectingEntry.h
+++ b/proxy/http/ConnectingEntry.h
@@ -70,7 +70,7 @@ struct IpHelper {
   }
 };
 
-using ConnectingIpPool = std::unordered_multimap<IpEndpoint, ConnectingEntry *, IpHelper>;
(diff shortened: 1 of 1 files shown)


bneradt:
[approve ci]

bneradt (commented):
Looks correct. Default IpEndpoint operator== is just address comparison, not including the port. So this makes sense. Especially since `IpHelper` includes the key_eq check, this seems like an oversight to not include it in the template instantiation.

I'll let a non-Yahoo approve. Maybe @shinrich is appropriate.

shinrich (approved):
Seems like a good clarification.

PR #10158, opened 2023-08-07
Add AutoOptionHelpers.cmake

JosiahWI:
This module assists with making the usage of optional libraries consistent and in agreement with automake. It provides functions to register new options, and to print out the state of all associated features.

Each option based on a package will follow the format ENABLE_XXX, and will have an associated feature variable (e.g. TS_HAS_JEMALLOC). Its value can be ON, 1, TRUE, OFF, 0, FALSE, or AUTO.

When the states of the features are printed out, features that are ON or OFF will generally be color coded green or red respectively to help configuration mistakes stand out. Options that were explicitly turned off are given special treatment: if the feature is off, it will not be color coded, because this is expected; if the feature is found to be enabled despite the user's request, it will be colored red and a CMake warning will be issued. All of this should help keep things consistent and avoid any potentially annoying bugs caused by misconfiguration.


JosiahWI:
[approve ci freebsd]

JosiahWI:
[approve ci freebsd]

cmcfarlen (approved):
Very cool! Thanks!

cmake/AutoOptionHelpers.cmake
```diff
@@ -0,0 +1,104 @@
...
+  set_property(CACHE ${_NAME} PROPERTY STRINGS AUTO ON OFF)
+
+  set(LOCAL_AUTO_OPTION_VARS ${GLOBAL_AUTO_OPTION_VARS})
+  list(APPEND LOCAL_AUTO_OPTION_VARS "${_NAME}")
+  set(GLOBAL_AUTO_OPTION_VARS ${LOCAL_AUTO_OPTION_VARS} PARENT_SCOPE)
+endfunction()
+
+# Add a new auto feature that uses find_package.
+# Creates an option ENABLE_<PACKAGE_NAME>.
+macro(AUTO_FEATURE_PACKAGE _PACKAGE_NAME _FEATURE_VAR _DESCRIPTION)
```
JosiahWI, line 37: This name is not great. I've come up with a much more usable interface that's also more flexible and more idiomatic. Argument parsing can be used to specify types of package dependencies using arguments such as `PACKAGES`, `SYMBOLS`, etc.

cmcfarlen (approved):
Looks good!  This will be real nice!

PR #10163, opened 2023-08-08
Make sure that the thread local time is updated timely

zwoop:
This tries to fix the problem Yahoo is seeing with times on short lived requests are skewed now when the thread local times aren't updated as often as they need to be. This does

* Revert the thread_local time cache
* Always call clock_gettime() (it's cheap / fast, with native TSC handled by the kernel / VDSO
* It also switches to CLOCK_REALTIME_COARSE, for now at least, we can reevaluate if that accuracy is enough.
* Finally it removes some old code, and cleans up the interfaces.

co-author: Chris McFarlen, Craig Taylor


ywkaras:
How about this instead?  https://github.com/apache/trafficserver/pull/10164

bneradt:
[approve ci autest]

bneradt (dismissed):
I like that this does away with the `get_hrtime_updated` vs `get_hrtime` distinction. It sounds like your benchmarks shows it's performant too.

I approve. I would appreciate hearing @masaori335's thoughts as well.

zwoop:
> How about this instead? #10164

I think 10164 essentially reverts everything back to as before the time "cache" became thread_local. Since these are high res values, odds are that you will always write the updated time. It will fix yahoo's problem, but essentially is the same as reverting 72e660233155b8c825bc215324c6fbd9e2197c97 in almost all cases.

zwoop:
I'm not happy with the arbitrary 0.5ms at 2Ghz numbers here. 0.5ms first of all is perhaps too long, and I'd really like to get a better estimate of the CPU frequency (but having no luck with hwloc so far :/).

iocore/eventsystem/I_Thread.h
```diff
@@ -157,16 +157,35 @@ class Thread
...
+    return tsc;
+#else
+    unsigned int lo, hi;
+    asm volatile("rdtsc" : "=a"(lo), "=d"(hi));
 
-      @note This also updates the cached time.
-  */
-  static ink_hrtime get_hrtime_updated();
+    return ((uint64_t)hi << 32) | lo;
+#endif
```
bneradt, line 174: Can we use `#include <intrin.h>` with `__rdtsc()` instead?

bneradt: This comment is not applicable anymore with the latest commits in this PR.

bneradt (approved):

bneradt:
[approve ci freebsd]

c-taylor:
To record here also as 'coarse' clocks are mentioned. Underlying kernel config affects the resolution of that clock source.
CONFIG_HZ does vary by distribution, but defaults (currently) to 1000Hz. Because of the variance I would continue to default (as the PR does) to CLOCK_REALTIME.

https://www.kernel.org/doc/html/latest/core-api/timekeeping.html

bneradt (approved):
We tested with this patch internally and confirmed it resolves the latency metrics issue we saw with thread_local.

Thanks for working on this!

doc/admin-guide/files/records.yaml.en.rst
```diff
@@ -5204,6 +5204,13 @@ Sockets
    ``2`` Tracks IO Buffer Memory and OpenSSL Memory allocations and releases
    ===== ======================================================================
 
+.. ts:cv:: CONFIG proxy.config.system_clock INT 0
+
+   *For advanced users only*. This allows to specify the underlying system clock
+   used by ATS. The default is ``CLOCK_REALTIME`` (``0``), but a higher performance
+   option could be ``CLOCK_REALTIME_COARSE`` (``5``). See ``clock_gettime(2)``_ for
+   more details.
+
```
bneradt, line 5213: Should we says something about how to ascertain the values of these? My man page for clock_gettime lists the names but not values (0, 5, etc). Or, since it is low level, do we assume they can figure out how to get the values from the appropriate .h file?

zwoop: I can add something about where to find the numbers. But if they can’t find that they probably shouldn’t be messing with this :)

masaori335:
I'm +1 of getting rid of the `cur_time` cache from ATS. It looks like the modern Linux Kernel can handle it well from our benchmark.

- vanilla (includes thread_local cur_time patch): 541,994 req/sec
- Revert PR#9184: 463,053 req/sec (85.4%)
- PR#10163 with `CLOCK_REALTIME`: 535,093 req/s (98.7%)
- PR#10163 with `CLOCK_REALTIME_COARSE`: 536,563 req/s (99.0%)

doc/admin-guide/files/records.yaml.en.rst
masaori335: It's defined in the `time.h` on Linux.

Linux (https://github.com/torvalds/linux/blob/v6.4/include/uapi/linux/time.h)
```
#define CLOCK_REALTIME			0
#define CLOCK_REALTIME_COARSE		5
```

Now, another CAVEAT is the value depends on the platform.

FreeBSD (https://github.com/freebsd/freebsd-src/blob/release/13.2.0/sys/sys/_clock_id.h)
```
#define CLOCK_REALTIME		0
#define CLOCK_REALTIME_FAST	10
#define	CLOCK_REALTIME_COARSE	CLOCK_REALTIME_FAST
```

macOS doesn't support `CLOCK_REALTIME_COARSE`
```
_CLOCK_REALTIME __CLOCK_AVAILABILITY = 0,
#define CLOCK_REALTIME _CLOCK_REALTIME
```

zwoop:
[approve ci autest]

zwoop:
[approve ci]

zwoop:
[approve ci centos]

zwoop:
[approve ci fedora]

ezelkow1:
[approve ci]

zwoop:
[approve ci]

zwoop:
[approve ci autest]

bneradt:
[approve ci autest]

doc/admin-guide/files/records.yaml.en.rst
```diff
@@ -5204,6 +5204,13 @@ Sockets
    ``2`` Tracks IO Buffer Memory and OpenSSL Memory allocations and releases
    ===== ======================================================================
 
+.. ts:cv:: CONFIG proxy.config.system_clock INT 0
+
+   *For advanced users only*. This allows to specify the underlying system clock
+   used by ATS. The default is ``CLOCK_REALTIME`` (``0``), but a higher performance
+   option could be ``CLOCK_REALTIME_COARSE`` (``5``). See ``clock_gettime(2)``_ for
+   more details. On Linux, these definitions can be found in ``<linux/time.h>``.
+
```
bneradt, line 5213: I think this isn't rendering how we would like it to. Do we want that underscore after the `clock_gettime(2)` reference? Here's the CI built version of the docs:

https://ci.trafficserver.apache.org/job/Github_Builds/job/docs/2600/artifact/output/10163/docbuild/html/admin-guide/files/records.yaml.en.html#proxy-config-system-clock

bneradt (changes requested):

doc/admin-guide/files/records.yaml.en.rst
zwoop: Fixed I hope

ywkaras: Looks alright to me:  https://ci.trafficserver.apache.org/job/Github_Builds/job/docs/2604/Sphinx_20Docs/admin-guide/files/records.yaml.en.html#proxy-config-system-clock

bneradt (approved):

PR #10164, opened 2023-08-09
Avoid unnecessary (multi-core) writes of Thread::cur_time.

ywkaras:
This variable is used frequently in most threads.  By avoiding unnecessary writes of it, locks of its cache line in (all) per-core caches are reduced.  This should reduce CPU core stalls.  The variable thus does not need to be thread_local.

### DIFF
--- a/iocore/eventsystem/I_Thread.h
+++ b/iocore/eventsystem/I_Thread.h
@@ -175,7 +175,7 @@ public:
 protected:
   Thread();
 
-  static thread_local ink_hrtime cur_time;
+  static ink_hrtime cur_time;
 };
 
 extern Thread *this_thread();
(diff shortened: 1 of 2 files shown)


ywkaras:
[approve ci freebsd]

zwoop:
This doesn't make a whole lot of sense. You may as well get rid of the cache completely, and just call ink_get_hrtime_internal() directly. I made a version of my patch that does exactly that, and it's reasonably fast, but not as fast as current master.

zwoop:
I'd imagine in almost all cases, now > cur_time will be true, which means this code is the same as before making the data thread local. In that case, I made a version of my patch that just calls clock_gettime() directly every time. It's fast, but not quite as fast as the current thread local version with the update.

ywkaras:
So, in a nutshell, the issue is, how, and in what weird ways, are we going to fuzz the high-res clock, in order to improve performance.  Yahoo prod says that the current fuzzing (that resulted from thread_local cur_time) changes ATS stats, making it look like transactions take longer.

Hard to know the relative performance of my change without testing.  Depends how often get_hrtime_updated() is called.  (Typically several hundred instructions execute per nanosecond I think.)  The performance would also be better if ink_get_hrtime_internal() returns a time that is already fuzzed.  That is, the granularity is not really one nanosecond.

We could improve performance of my change by adding fuzzing, by only updating cur_time if it's less than now by some delta.

zwoop:
@ywkaras I got rid of all the TSC stuff, so that's all gone. This version is as fast as current master, solves all of Yahoo's problems, simplifies the usage, and guarantees 1ms or better accuracy.

ywkaras:
Seems like switching to a coarse clock may make the clock more fuzzy.  So, maybe Yahoo prod would still see squirrelly metrics.

PR #10172, opened 2023-08-11
Makes cmake build again, on macOS

zwoop:


JosiahWI (approved):
Thanks for the fix!

bneradt (approved):

ywkaras:
Where do we order our MCBA baseball caps?

PR #10173, opened 2023-08-11
Remove autoconf headers during CMake configuration

JosiahWI:
The headers cause duplicate symbol warnings if they are in the source tree during a configuration. This should finally put that annoyance to rest.


JosiahWI:
[approve ci fedora]

zwoop (approved):

PR #10184, opened 2023-08-14
Check SNI in h3

### LINKED ISSUE #10089: Check SNI extension is used on H3
>    QUIC version 1 uses TLS version 1.3 or greater as its handshake
   protocol.  HTTP/3 clients MUST support a mechanism to indicate the
   target host to the server during the TLS handshake.  If the server is
   identified by a domain name ([DNS-TERMS]), clients MUST send the
   Server Name Indication (SNI; [RFC6066]) TLS extension unless an
   alternative mechanism to indicate the target host is used.

https://www.rfc-editor.org/rfc/rfc9114.html#section-3.2-2

What to do on the server side doesn't seem to be defined, but we can probably use `H3_GENERAL_PROTOCOL_ERROR` and close the connection if a client does not use SNI extension.

Note that this is from H3 spec and QUIC does not require use of SNI extension (i.e. we should do this check in `proxy/http3`).

lzx404243:
This closes #10089.

### DIFF
--- a/proxy/http3/Http3SessionAccept.cc
+++ b/proxy/http3/Http3SessionAccept.cc
@@ -48,6 +48,12 @@ Http3SessionAccept::accept(NetVConnection *netvc, MIOBuffer *iobuf, IOBufferRead
     Warning("QUIC client '%s' prohibited by ip-allow policy", ats_ip_ntop(client_ip, ipb, sizeof(ipb)));
     return false;
   }
+  // RFC9114, section 3.2-2: Client must send the SNI extension.
+  if (auto sni = netvc->get_service<TLSSNISupport>(); !sni || sni->get_sni_server_name()[0] == '\0') {
+    ip_port_text_buffer ipb;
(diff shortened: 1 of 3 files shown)


lzx404243:
@maskit What do you think about placing the SNI check in `Http3SessionAccept::accept()`?  On one hand, it makes sense as this is right before the start of a H3 session and once we know that client didn't send a SNI we can reject it and don't need to start the session at all. The downside for this is we can't send a H3 response from here. We just rejects the session.

lzx404243:
> what happens after ATS rejects a connection at `accept()`?

`netvc->do_io_close()` will be called which supposedly closes the quic connection. 

However, I noticed in the Autest that the curl/proxy verifier client times out. And doing a packet capture on the rejection case, I don't see ATS closes the quic connection(no `CONNECTION_CLOSE ` frame etc). It seems there is some silent drop going on here.

maskit:
Ok, I think this PR is fine. I don't think I implemented a way to initiate connection close on H3 layer.

maskit (approved):

PR #10188, opened 2023-08-15
Fixes a problem which can decrement milestone metrics unintentionally

zwoop:
I believe what can happen here is that if a milestone is 0 (because it never triggered), we will decrement that metric with 1. This often has little impact, because of other transaction adding hundreds if not thousands of msec to those milestone metrics. An example when this happens is when you have cache misses, and the server connection milestones are all 0.

### DIFF
--- a/proxy/Milestones.h
+++ b/proxy/Milestones.h
@@ -64,10 +64,10 @@ public:
    * @return The difference time in milliseconds
    */
   int64_t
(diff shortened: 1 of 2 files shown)


zwoop:
[approve ci autest]

bryancall (approved):

zwoop:
Cherry-picked to v9.2.x

Issue #10196, opened 2023-08-16
CID 1518060: Uninitialized scalar variable


maskit:
```
1095  if (local_addr.network_order_port() || !is_any_address) {
    	19. Condition -1 == SocketManager::ink_bind(fd, &local_addr.sa, ats_ip_size(&local_addr.sa), 0), taking false branch.
1096    if (-1 == SocketManager::ink_bind(fd, &local_addr.sa, ats_ip_size(&local_addr.sa))) {
1097      char buff[INET6_ADDRPORTSTRLEN];
1098      Debug("udpnet", "ink bind failed on %s", ats_ip_nptop(local_addr, buff, sizeof(buff)));
1099      goto SoftError;
1100    }
1101
    	
CID 1518060 (#1 of 1): Uninitialized scalar variable (UNINIT)
20. uninit_use_in_call: Using uninitialized value local_addr_len when calling safe_getsockname. [[show details](https://scan6.scan.coverity.com/eventId=34735411-25&modelId=34735411-0&fileInstanceId=162995658&filePath=%2Fsrc%2Ftscore%2Fink_sock.cc&fileStart=223&fileEnd=230)]
1102    if (safe_getsockname(fd, &local_addr.sa, &local_addr_len) < 0) {
1103      Debug("udpnet", "CreateUdpsocket: getsockname didn't work");
1104      goto HardError;
1105    }
1106  }
```

PR #10212, opened 2023-08-16
Coverity 1518060: Uninitialized scalar variable

maskit:
This closes https://github.com/apache/trafficserver/issues/10196.

### DIFF
--- a/iocore/net/UnixUDPNet.cc
+++ b/iocore/net/UnixUDPNet.cc
@@ -1014,8 +1014,8 @@ bool
 UDPNetProcessor::CreateUDPSocket(int *resfd, sockaddr const *remote_addr, Action **status, NetVCOptions const &opt)
 {
   int res = 0, fd = -1;
-  int local_addr_len;
   IpEndpoint local_addr;
+  int local_addr_len = sizeof(local_addr.sa);
 
   // Need to do address calculations first, so we can determine the
   // address family for socket creation.


iocore/net/UnixUDPNet.cc
```diff
@@ -1014,8 +1014,8 @@ bool
 UDPNetProcessor::CreateUDPSocket(int *resfd, sockaddr const *remote_addr, Action **status, NetVCOptions const &opt)
 {
   int res = 0, fd = -1;
-  int local_addr_len;
   IpEndpoint local_addr;
+  int local_addr_len = sizeof(local_addr.sa);
```
bneradt, line 1018: Am I reading this right that this is only used as a throw-away, non-read output variable below? If so, I suggest just declaring and initializing it to zero just before the point of use.

bneradt (changes requested):

iocore/net/UnixUDPNet.cc
maskit: I thought so too, but actually initializing it with 0 is wrong.

> getsockname() returns the current address to which the socket
       sockfd is bound, in the buffer pointed to by addr.  The addrlen
       argument should be initialized to indicate the amount of space
       (in bytes) pointed to by addr.  On return it contains the actual
       size of the socket address.

https://man7.org/linux/man-pages/man2/getsockname.2.html

bneradt: Ah, good to know. Thanks.

bneradt (approved):

PR #10231, opened 2023-08-16
Move Cleanup.h from include/tscpp/api to include/ts.

ywkaras:
Change the namespace of its declarations from atscppapi to tsapi::c_support. This is in support of removing the main, high-overhead parts of the ATS C++ API.

I used `git mv` to move the file, frustrating that github shows it as a deletion and a new file.


maskit:
Does this mean everything in Cleanup.h is going to be part of TS API and require API proposal and deprecation period when we change it? Or is it still just an utility that does not guarantee any compatibility between ATS versions?

ywkaras:
> Does this mean everything in Cleanup.h is going to be part of TS API and require API proposal and deprecation period when we change it? Or is it still just an utility that does not guarantee any compatibility between ATS versions?

Yes I think so.

cmcfarlen:
Should this go in the new `include/api` directory?

ywkaras:
> Should this go in the new `include/api` directory?

I haven't heard about that directory, maybe I wasn't in the right meeting (or I slipped into a meeting coma).

maskit:
> > Does this mean everything in Cleanup.h is going to be part of TS API and require API proposal and deprecation period when we change it? Or is it still just a utility that does not guarantee any compatibility between ATS versions?

> Yes I think so.

Which one?

maskit:
> It will be part of the TS API.

Then it's an addition in terms of TS API, and a proposal need to be sent on the dev list.

ywkaras:
I have brought it up on the mailing list.

maskit:
You mean the one you sent last year?

include/ts/Cleanup.h
```diff
@@ -0,0 +1,201 @@
...
+#pragma once
+
+#include <type_traits>
+#include <memory>
+
+#include <ts/ts.h>
+
+namespace tsapi
+{
+// This sub-namespace is for declarations specifically designed to work with declarations in tsapi::c.
```
maskit, line 33: Where's `tsapi::c`?

ywkaras: In ts.h.

ywkaras:
> You mean the one you sent last year?

https://lists.apache.org/list?dev@trafficserver.apache.org:2023-8

include/ts/Cleanup.h
maskit: Ok, so it was added recently. I didn't know such namespace was added. I don't think I saw the proposal on the list.

ywkaras:
> You proposed removing C++ API and saving Cleanup.h, but promoting Cleanup.h to TS API is not part of the proposal. You might have meant it, but I didn't think that is included. And the thread is about "Removing example/plugins/cpp-api" and it's not even started by you. I don't think it's a valid proposal.

https://lists.apache.org/list?dev@trafficserver.apache.org:2023-8

PR #10246, opened 2023-08-17
Fix ip_allow optional methods specification

lzx404243:
According to doc, the `methods` field is optional so the following should be a valid ip_allow config which allows all HTTP methods:
```yaml
ip_allow:
  - apply: in
    ip_addrs: 0/0
    action: allow
```
However, ATS reports an error when parsing it. 

The root cause is that yaml-cpp returns an invalid node for non-existent keys, which is fine, but it throws an exception when assigning the invalid node to another node. This PR prevents this and adds a test to exercise this case.


SolidWallOfCode (approved):
This is, IMHO, a bug in YAML-CPP, that this is fixed by using a new variable.

PR #10247, opened 2023-08-17
Python 3.12: microserver.test.ext wrap_socket update

bneradt:
ssl.wrap_socket was deprecated in Python 3.6 and is removed in Python 3.12. This patch replaces it with the recommended SSLContext.wrap_socket method instead. This fixes the extension for Python 3.12.


maskit (approved):
Looks good, though I don't know Python very much.

Should we back port this to 9.2.x? The new way has been there for years so it should be possible.

zwoop:
Cherry-picked to v9.2.x

PR #10250, opened 2023-08-18
This dependency is not needed, cmake did it right

zwoop:


maskit:
I guess I had to add the dependency because some function was un-inlined on my cleanup (#8995) and some recent cleanup made it unnecessary.

bneradt (approved):

PR #10256, opened 2023-08-21
Python 3.12: add h2 dependencies to autest Pipfile

bneradt:
When testing with Python 3.12, the autests that used the h2 package failed because the Pipenv didn't pull in the hyperframe and hpack dependencies. This explicitly pulls those dependencies in via the autest Pipfile.


maskit (approved):

zwoop:
Cherry-picked to v9.2.x

PR #10257, opened 2023-08-21
Python 3.12: Make autest regex strings raw strings

bneradt:
Python 3.12 makes the use of unexpected escape sequences a SyntaxError. This is simply avoided by making these regexes, which are intentional and passed down to other modules or shell commands, raw character sequences so that Python doesn't incorrectly try to interpret them.

### DIFF
--- a/tests/gold_tests/headers/general-connection-failure-502.test.py
+++ b/tests/gold_tests/headers/general-connection-failure-502.test.py
@@ -42,6 +42,6 @@ tr.Processes.Default.StartBefore(Test.Processes.ts)
 # Do not start the origin server: We wish to simulate connection refused while hopefully no one else uses this port.
 tr.Processes.Default.Command = \
     (f"{sys.executable} tcp_client.py 127.0.0.1 {ts.Variables.port} www.connectfail502.test-get.txt | "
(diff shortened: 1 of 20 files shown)


bneradt:
[approve ci]

bneradt:
[approve ci]

jpeach (approved):

bneradt:
[approve ci]

bneradt:
[approve ci]

bneradt:
[approve ci]

bneradt:
[approve ci]

moonchen (approved):

bneradt:
I marked this for 9.2.x since we'll need this for autest of that branch for anything running Python 3.12. Otherwise, CI will fail.

zwoop:
This fixes something that's new via #9572, but looking at that one, kinda feels we also want #9572 in 9.2.x?

bneradt:
> This fixes something that's new via #9572, but looking at that one, kinda feels we also want #9572 in 9.2.x?

I wouldn't pull in #9572 just because of this because that includes a production change. Rather, let's take this PR out of 9.2.x and I'll create a new PR without the patch for that new test.

zwoop:
Cherry-picked to v9.2.x

PR #10283, opened 2023-08-25
cmake: add tests for proxy/hdrs

cmcfarlen:
This also simplifies linking hwloc

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -302,12 +302,6 @@ if(ENABLE_DOCS)
   find_program(PipEnv pipenv REQUIRED)
 endif()
 
-# Catch2 for tests
-set(CATCH_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/lib/catch2)
-
-add_library(catch::catch INTERFACE IMPORTED)
-target_include_directories(catch::catch INTERFACE ${CMAKE_SOURCE_DIR}/include ${CATCH_INCLUDE_DIR})
-
 include(CTest)
 set(TS_HAS_TESTS ${BUILD_REGRESSION_TESTING})
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -305,6 +305,9 @@ endif()
 # Catch2 for tests
 set(CATCH_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/lib/catch2)
 
+add_library(catch::catch INTERFACE IMPORTED)
+target_include_directories(catch::catch INTERFACE ${CMAKE_SOURCE_DIR}/include ${CATCH_INCLUDE_DIR})
+
 include(CTest)
 set(TS_HAS_TESTS ${BUILD_REGRESSION_TESTING})
 
--- a/iocore/eventsystem/CMakeLists.txt
+++ b/iocore/eventsystem/CMakeLists.txt
@@ -49,3 +49,7 @@ target_link_libraries(inkevent
         tscpputil # transitive
         yaml-cpp::yaml-cpp # transitive
 )
+
+if(TS_USE_HWLOC)
+        target_link_libraries(inkevent PRIVATE hwloc::hwloc)
+endif()
--- a/iocore/eventsystem/CMakeLists.txt
+++ b/iocore/eventsystem/CMakeLists.txt
@@ -49,7 +49,3 @@ target_link_libraries(inkevent
         tscpputil # transitive
         yaml-cpp::yaml-cpp # transitive
 )
-
-if(TS_USE_HWLOC)
-        target_link_libraries(inkevent PRIVATE hwloc::hwloc)
-endif()
--- a/proxy/hdrs/CMakeLists.txt
+++ b/proxy/hdrs/CMakeLists.txt
@@ -53,6 +53,7 @@ add_executable(test_proxy_hdrs
     unit_tests/test_HdrUtils.cc
     unit_tests/test_HdrHeap.cc
     unit_tests/test_HeaderValidator.cc
+    unit_tests/test_Huffmancode.cc
     unit_tests/test_mime.cc
     unit_tests/test_URL.cc
     unit_tests/unit_test_main.cc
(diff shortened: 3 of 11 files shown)


proxy/hdrs/unit_tests/test_Huffmancode.cc
```diff
@@ -88,13 +89,20 @@ random_test()
   free(dst_start);
 }
 
+TEST_CASE("Huffmancode Random", "[proxy][huffman]")
+{
+  // This doesn't check anything ...
```
JosiahWI, line 94: Then why does it exist?

CMakeLists.txt
```diff
@@ -307,6 +307,9 @@ endif()
 # Catch2 for tests
 set(CATCH_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/lib/catch2)
 
+add_library(catch::catch INTERFACE IMPORTED)
+target_include_directories(catch::catch INTERFACE ${CMAKE_SOURCE_DIR}/include ${CATCH_INCLUDE_DIR})
```
JosiahWI, line 311: I thought there was already a target? `catch2::catch2` maybe? Their own official target last I checked is `Catch2::Catch2WithMain`, which we don't use (but maybe should for consistency).

proxy/hdrs/CMakeLists.txt
```diff
@@ -46,3 +46,26 @@ target_link_libraries(hdrs
...
+    unit_tests/test_HdrUtils.cc
+    unit_tests/test_HdrHeap.cc
+    unit_tests/test_HeaderValidator.cc
+    unit_tests/test_Huffmancode.cc
+    unit_tests/test_mime.cc
+    unit_tests/test_URL.cc
+    unit_tests/unit_test_main.cc
+)
+target_link_libraries(test_proxy_hdrs PRIVATE ts::hdrs ts::tscore ts::inkevent catch::catch)
+add_test(NAME test_proxy_hdrs COMMAND $<TARGET_FILE:test_proxy_hdrs>)
```
JosiahWI, line 62: Is the generator expression necessary? My understanding of the `add_test` docs was that if the command is the name of a target, it automatically uses that targets binary output.

CMakeLists.txt
cmcfarlen: We vendor the catch header file so we don't use the find package.  I just wanted the prettier way of getting the include :smile:

proxy/hdrs/CMakeLists.txt
cmcfarlen: I don't think it is necessary but I was just following suit from other invocations of `add_test`

CMakeLists.txt
JosiahWI: lib/CMakeLists.txt#L35-36

cmcfarlen: oh, that's not in the convenient spot!

src/traffic_layout/CMakeLists.txt
```diff
@@ -34,10 +34,6 @@ target_link_libraries(traffic_layout
         ${OPENSSL_LIBRARY}
 )
 
-if(TS_USE_HWLOC)
-    target_link_libraries(traffic_layout PRIVATE hwloc::hwloc)
-endif()
-
```
JosiahWI, line 40: info.cc needs to get the API version from hwloc if it's enabled, so hwloc is a usage requirement.

iocore/eventsystem/CMakeLists.txt
```diff
@@ -49,7 +49,3 @@ target_link_libraries(inkevent
         tscpputil # transitive
         yaml-cpp::yaml-cpp # transitive
 )
-
-if(TS_USE_HWLOC)
-        target_link_libraries(inkevent PRIVATE hwloc::hwloc)
-endif()
```
JosiahWI, line 55: UnixEventProcessor.cc uses symbols from hwloc, so it is a usage requirement.

proxy/hdrs/CMakeLists.txt
```diff
@@ -46,3 +46,26 @@ target_link_libraries(hdrs
...
+target_link_libraries(test_proxy_hdrs PRIVATE ts::hdrs ts::tscore ts::inkevent catch2::catch2)
+add_test(NAME test_proxy_hdrs COMMAND test_proxy_hdrs)
+
+add_executable(test_proxy_hdrs_xpack
+    XPACK.cc
+    HuffmanCodec.cc
+    unit_tests/test_XPACK.cc
+)
+target_include_directories(test_proxy_hdrs_xpack PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+target_link_libraries(test_proxy_hdrs_xpack PRIVATE ts::tscore libswoc catch2::catch2)
```
JosiahWI, line 70: Also depends on tscpputil. Doesn't depend on libswoc.

JosiahWI (changes requested):
I wonder whether it's possible to put all the hwloc stuff in one place instead of sprinkling conditional code around in multiple places.

proxy/hdrs/CMakeLists.txt
cmcfarlen: tscore needs libswoc but doesn't have it PUBLIC

iocore/eventsystem/CMakeLists.txt
cmcfarlen: I removed this because its pulled in transitively

proxy/hdrs/CMakeLists.txt
JosiahWI: It almost certainly should be PUBLIC.

iocore/eventsystem/CMakeLists.txt
JosiahWI: If the hwloc related code is that tightly coupled, it probably shouldn't be in 2 different components. If that's not the case, then imagine a scenario where we no longer need the hwloc code in tscore. Eventsystem will fail to compile, and the author will have to patch the inkevent CMakeLists as well, requiring a change in 2 places instead of 1.

src/traffic_layout/CMakeLists.txt
```diff
@@ -35,7 +35,7 @@ target_link_libraries(traffic_layout
 )
 
 if(TS_USE_HWLOC)
-    target_link_libraries(traffic_layout PRIVATE hwloc::hwloc)
+        target_link_libraries(inkevent PRIVATE hwloc::hwloc)
```
JosiahWI, line 38: Should be traffic_layout, not inkevent.

JosiahWI (changes requested):

JosiahWI (approved):
Looks good, thanks!

bneradt (approved):

PR #10285, opened 2023-08-25
Fix slice head request memory issue

### LINKED ISSUE #10280: Slice plugin: HEAD requests causing memory issues
Recent changes to enable PURGE requests for slice plugin resulted in HEAD requests to constantly loop (on the same request) on the same header heap
- client side is unaffected but server will continue to loop on HEAD requests
- HdrHeap nodes were constantly being allocated, causing high memory usage 
- https://github.com/apache/trafficserver/pull/9831

Currently working on a fix

serrislew:
Fixes #10280 
- regression bug: change in #9831 caused loop in HEAD requests


zwoop:
Cherry-picked to v9.2.x

PR #10287, opened 2023-08-25
Fix a crash due to bad disks

moonchen:
Hold the volume lock when handling a bad io result so that dir_delete doesn't crash.


masaori335:
[approve ci fedora]

SolidWallOfCode (approved):

zwoop:
Cherry-picked to v9.2.x

PR #10298, opened 2023-08-29
Adjust CMakeLists with git worktree

masaori335:
In the git worktree directory, the `.git` is just an ASCII text file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -479,7 +479,7 @@ add_custom_target(format
         COMMENT "formatting all files"
         )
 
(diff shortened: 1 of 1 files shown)


cmcfarlen (dismissed):

jpeach (approved):
This change looks fine, but IIUC the cmake correctly, changes to the pre-commit script won't ever take effect once it is installed the first time.

masaori335:
It looks like `NOT EXISTS ${CMAKE_SOURCE_DIR}/.git/hooks/pre-commit` is always true and the `configure_file` fails in the worktree directory.
```
-- Installing github hook
CMake Error at CMakeLists.txt:484 (configure_file):
  Not a directory
```

PR #10306, opened 2023-08-29
Add support for OCSP requests by GET method

maskit:
This adds support of OCSP requests by GET method. Requests by GET method enables caching their responses.

If `proxy.config.ssl.ocsp.request_mode` is set to 1, ATS uses GET method for OCSP requests if possible. If the encoded URL is going to be too long, it fallbacks to POST method. The setting is 0 by default (no behavior change).

### DIFF
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -4254,6 +4254,14 @@ OCSP Stapling Configuration
 
    Number of seconds before an OCSP response expires in the stapling cache.
 
+.. ts:cv:: CONFIG proxy.config.ssl.ocsp.request_mode INT 0
+
+   Set the request method to prefer when querying OCSP responders. The default
+   is zero, or POST, and a value of 1 will cause ATS to attempt a GET request.
+   Because the length of the encoded request must be less than 255 bytes per RFC
+   6960, Appendix A, ATS will fall back to the POST request method when the
+   encoded size exceeds this limit.
+
 .. ts:cv:: CONFIG proxy.config.ssl.ocsp.request_timeout INT 10
    :units: seconds
 
--- a/iocore/net/OCSPStapling.cc
+++ b/iocore/net/OCSPStapling.cc
@@ -25,11 +25,11 @@
 #include <openssl/x509v3.h>
 #include <openssl/asn1.h>
 #include <openssl/asn1t.h>
-#include <cstring>
 
 #include "tscore/ink_memory.h"
 #include "tscore/Encoding.h"
 #include "tscore/ink_base64.h"
+#include "tscore/ink_string.h"
 #include "P_Net.h"
 #include "P_SSLConfig.h"
 #include "P_SSLUtils.h"
@@ -1163,7 +1163,7 @@ make_url_for_get(TS_OCSP_REQUEST *req, const char *base_url)
   url = new_IOBufferBlock();
   url->alloc(buffer_idx);
 
-  int written = strlcpy(url->end(), base_url, url->write_avail());
+  int written = ink_strlcpy(url->end(), base_url, url->write_avail());
   url->fill(written);
 
   // Append '/' if base_url does not end with it
@@ -1172,7 +1172,7 @@ make_url_for_get(TS_OCSP_REQUEST *req, const char *base_url)
     url->fill(1);
   }
 
-  written = strlcat(url->end(), ocsp_escaped, url->write_avail());
+  written = ink_strlcat(url->end(), ocsp_escaped, url->write_avail());
   url->fill(written);
   Debug("ssl_ocsp", "appended encoded data to path: %s", url->buf());
 
--- a/iocore/net/OCSPStapling.cc
+++ b/iocore/net/OCSPStapling.cc
@@ -27,6 +27,8 @@
 #include <openssl/asn1t.h>
 
 #include "tscore/ink_memory.h"
+#include "tscore/Encoding.h"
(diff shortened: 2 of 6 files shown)


maskit:
[approve ci autest]
[approve ci rocky]

doc/admin-guide/files/records.yaml.en.rst
```diff
@@ -4254,6 +4254,14 @@ OCSP Stapling Configuration
...
    Number of seconds before an OCSP response expires in the stapling cache.
 
+.. ts:cv:: CONFIG proxy.config.ssl.ocsp.request_mode INT 0
+
+   Set the request method to prefer when querying OCSP responders. The default
+   is zero, or POST, and a value of 1 will cause ATS to attempt a GET request.
+   Because the length of the encoded request must be less than 255 bytes per RFC
+   6960, Appendix A, ATS will fall back to the POST request method when the
+   encoded size exceeds this limit.
+
```
bneradt, line 4264: As it is, it seems like it would be clearer to name this "prefer_get_method" or some such, and 1 enables it. Are you anticipating other possible configurations that might be set for this?

iocore/net/OCSPStapling.cc
```diff
@@ -308,10 +315,10 @@ class HTTPRequest : public Continuation
...
 
     this->_fsm = FetchSMAllocator.alloc();
-    if (use_post) {
-      this->_fsm->ext_init(this, "POST", uri, "HTTP/1.1", reinterpret_cast<sockaddr *>(&sin), TS_FETCH_FLAGS_SKIP_REMAP);
-    } else {
+    if (use_get) {
       this->_fsm->ext_init(this, "GET", uri, "HTTP/1.1", reinterpret_cast<sockaddr *>(&sin), TS_FETCH_FLAGS_SKIP_REMAP);
+    } else {
+      this->_fsm->ext_init(this, "POST", uri, "HTTP/1.1", reinterpret_cast<sockaddr *>(&sin), TS_FETCH_FLAGS_SKIP_REMAP);
     }
```
bneradt, line 322: I suggest setting the char* in a conditional:

```cpp
    char const *const method = use_get ? "GET": "POST";
    this->_fsm->ext_init(this, method, uri, "HTTP/1.1", reinterpret_cast<sockaddr *>(&sin), TS_FETCH_FLAGS_SKIP_REMAP);

```

bneradt (changes requested):

bneradt:
Looks good. Thanks for working on the OCSP fetch.

iocore/net/OCSPStapling.cc
```diff
@@ -1097,7 +1204,21 @@ stapling_refresh_response(certinfo *cinf, TS_OCSP_RESPONSE **prsp)
     goto err;
   }
 
-  *prsp = query_responder(cinf->uri, cinf->user_agent, req, SSLConfigParams::ssl_ocsp_request_timeout);
+  if (SSLConfigParams::ssl_ocsp_request_mode) { // True: GET, False: POST
```
bneradt, line 1207: This also would be clearer as `ssl_ocsp_prefer_get` or some such. Otherwise maybe create an enum and switch on that value?

```cpp
  bool use_get = false;
  switch(SSLConfigParams::ssl_ocsp_request_mode) {
    case USE_GET:
      use_get = true;
  }

doc/admin-guide/files/records.yaml.en.rst
maskit: I don't have any plans to add another mode at the moment. A possible mode would be "Always try GET regardless of the request size". The documentation uses "must" for the size limitation, but strictly speaking it's not a "MUST" on the RFC.

>   HTTP-based OCSP requests can use either the GET or the POST method to
   submit their requests.  To enable HTTP caching, small requests (that
   after encoding are less than 255 bytes) MAY be submitted using GET.
   If HTTP caching is not important or if the request is greater than
   255 bytes, the request SHOULD be submitted using POST.

I'm not sure if there are OCSP responders that accept big GET requests, but if some of them do, it can be an option.

@elsloo What do you think?

iocore/net/OCSPStapling.cc
maskit: Embedded the ternary operator into the ext_init line.

doc/admin-guide/files/records.yaml.en.rst
elsloo: I asked myself the same question when I worked on this initially, and had originally named the configuration parameter `request_method` and later renamed to `request_mode` to be more generic. I agree that the RFC is mostly clear on the behavior here but it isn't absolutely clear (SHOULD vs MUST), and as a result, we may in the future need another option to support a mixed mode of some kind bound by a size check to switch between the two request methods dynamically. I don't think that we should let that concern override the improvements here, and if someone does have GET requests that exceed the RFC-defined length, they can a) develop the support for variable request methods based on size which would require a new mode in configuration, or b) fall back to using POST for all requests.

I would personally prefer to keep this variable generic due to the RFC, which allows us to retain the variable if we need to extend it in the future. This is also somewhat of an advanced configuration topic, and most operators will likely read the documentation if they have any reason to move from POST to GET.

bneradt: Sounds good.

bneradt (approved):

iocore/net/OCSPStapling.cc
bneradt: We can stick with things as is for now and tweak this if we add new modes.

PR #10308, opened 2023-08-29
Add TS_HTTP_TUNNEL_START_HOOK and TSHttpTxnTypeGet API

### LINKED ISSUE #10307: Allow a plugin to decide whether a tunnel should continue
With tr-pass, traffic that does not parse a HTTP or TLS will be tunneled.  For some cases, it would be desirable to have a plugin that can look at other information, e.g. source or destination address or external policy, to determine whether the tunneled traffic should be allowed.

shinrich:
This PR adds the TS_HTTP_TUNNEL_START_HOOK and the TSHttpTxnTypeGet API. 

It also adds documentation and autest. 

This closes #10307


shinrich:
[ approve ci cmake ]

shinrich:
[ approve ci CMake ]

shinrich:
[ approve ci ]

shinrich:
[approve ci CMake]

shinrich:
[approve ci]

shinrich:
[approve ci Rocky]

shinrich:
[approve ci rocky]

tests/tools/plugins/Makefile.inc
```diff
@@ -67,6 +67,9 @@ tools_plugins_ssl_verify_test_la_SOURCES = tools/plugins/ssl_verify_test.cc
 noinst_LTLIBRARIES += tools/plugins/ssntxnorder_verify.la
 tools_plugins_ssntxnorder_verify_la_SOURCES = tools/plugins/ssntxnorder_verify.cc
 
+noinst_LTLIBRARIES += tools/plugins/hook_tunnel_plugin.la
```
moonchen, line 70: Please also add this to the CMake build.  Suggested change in tests/tools/plugins/CMakeLists.txt: `add_autest_plugin(hook_tunnel_plugin hook_tunnel_plugin.cc)`

moonchen (changes requested):

tests/tools/plugins/Makefile.inc
shinrich: Added

doc/developer-guide/api/functions/TSHttpTxnTypeGet.en.rst
```diff
@@ -0,0 +1,37 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+   implied.  See the License for the specific language governing
+   permissions and limitations under the License.
+
+.. include:: ../../../common.defs
+
+.. default-domain:: c
+
+TSHttpTxnTypeGet
```
maskit, line 21: I pointed this out on the dev list but I'm commenting here as well for visibility.

I don't think transactions on a connection/session don't have different types. Shouldn't it a type of a connection, a session, or a port?

shinrich: Hmm.  Guess I called it Txn, since I wanted to call this from a hook where I have a txn object.  I want to distinguish between tunnel type (which is a connection characteristic) and a HTTP transaction (which I guess is a characteristic of the connection as well). However from an implementation perspective this is information we pull from the HttpSM.  Let me think a bit on how invasive it would be to add this information to the vconn.

shinrich: Added a test case for the CONNECT method.  This gets processed as a HTTP Request type even though it does eventually set up a tunnel.  The tr-pass case and sni config case will get reported as various tunnel transactions since they are processed as tunnels directly not a transaction sent to the proxy first.

maskit: Just to confirm, does that mean `TS_TXN_TYPE_HTTP` will be returned on a transaction that receives CONNECT method?

I'm wondering if that is the expected and desired behavior. In other words, would you mind if somebody added `TS_TXN_TYPE_CONNECT_TUNNEL`?

In either case, it would be nice to have the behavior documented.

shinrich: Correct, as it stands a txn created from a CONNECT request will be reported as TS_TXN_TYPE_HTTP.  If one wanted to look at the method, they could differentiate that as TS_TXN_TYPE_CONNECT_TUNNEL.  I would be ok with that addition.  In the meantime, I'll review my proposed documentation to make sure this point is clear.

maskit (approved):
Looks good to me.

moonchen (approved):

PR #10315, opened 2023-08-30
Move InkAPI.cc into src/api

lzx404243:
This PR moves InkAPI.cc into src/api, so that it's built into the libtsapi.so. A new `libtsapicore.a` is created to separate out the stuff also used in the core(e.g. Metrics) from the high-level plugin APIs.  

Also moved some files around:
* Moved `Hoststatus.cc` from `src/traffic_server/` to `proxy/` where the header is.
* Moved `FetchSM.h/cc` from `src/traffic_server/` to `include/api` and `src/api` respectively.
* Moved `InkContInternal.cc` and `InkIOCoreAPI.cc` from `src/traffic_server/` to `src/api`.
* Moved `InkAPIInternal.h` from `proxy/` to `include/api`.


cmcfarlen (approved):
Great progress on cleanup efforts!  Thanks!

lzx404243:
@cmcfarlen Thanks for review and all the suggestions!

PR #10321, opened 2023-08-31
Disable copying/moving for DbgCtl.

ywkaras:


include/ts/DbgCtl.h
```diff
@@ -90,6 +90,10 @@ class DbgCtl
   static void print(char const *tag, char const *file, char const *function, int line, char const *fmt_str, ...)
     TS_PRINTFLIKE(5, 6);
 
+  // No copying/moving.
+  DbgCtl(DbgCtl const &)            = delete;
```
jpeach, line 94: Do we also need to delete `DbgCtl(DbgCtl const &&)` these days?

jpeach (approved):

include/ts/DbgCtl.h
ywkaras: Moving defaults to whatever copying is:  https://godbolt.org/z/cETc71dxr .

PR #10322, opened 2023-08-31
Update Proxy Verifier to v2.10.1

bneradt:
This includes the ability to specify separate DATA frames for a stream as well as a way to keep a session open after the final transaction.


bneradt:
Marking for 9.2 since the new PV release is backwards compatible and some new tests which might go into 9.2 may depend upon the updated release.

randall (approved):

zwoop:
@bneradt This gets merge conflicts, I think there's quite a few PRs that would have to go in for this, or you have to make a 9.2.x branch PR.

PR #10324, opened 2023-08-31
Add TSVConnFdGet api

### LINKED ISSUE #10323: Create helper TS API to return the file descriptor associated with a SSL connection
I needed this to simplify writing a plugin that needed some access to the SSL object information.  For a variety of reasons, I didn't want to pull in a SSL library to fetch this information directly from the SSL object I could have retrieved from TSVConnSslConnectionGet.

I also had created a wrapper API to return the SNI, but I see that someone has already contributed that via the TSVConnSslSniGet function.

shinrich:
Adding a convenience API for working with SSL connections.

This closes #10323

### DIFF
--- a/doc/developer-guide/api/functions/TSVConnSslFdGet.en.rst
+++ b/doc/developer-guide/api/functions/TSVConnSslFdGet.en.rst
@@ -1,0 +1,34 @@
+.. Licensed to the Apache Software Foundation (ASF) under one or more
+   contributor license agreements.  See the NOTICE file distributed
+   with this work for additional information regarding copyright
+   ownership.  The ASF licenses this file to you under the Apache
+   License, Version 2.0 (the "License"); you may not use this file
+   except in compliance with the License.  You may obtain a copy of
+   the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+   implied.  See the License for the specific language governing
+   permissions and limitations under the License.
+
+
+TSVConnSslFdGet
+===============
+
+Synopsis
+--------
+
+.. code-block:: cpp
+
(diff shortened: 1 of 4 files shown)


doc/developer-guide/api/functions/TSVConnSslFdGet.en.rst
```diff
@@ -0,0 +1,34 @@
...
+.. code-block:: cpp
+
+    #include <ts/ts.h>
+
+.. c:function:: int TSVConnSslFdGet(TSVConn sslp)
+
+
+Description
+-----------
+Returns the file descriptor associated with the SSL connection :arg:`sslp`.
```
maskit, line 33: Why do we need to limit this just for SSL connection? Are we going to add `TSVConnFdGet` and `TSVConnQuicFdGet` if we needed? Plugins would need to call isSSL() and isQUIC() then call the corresponding TSVConnXFdGet()`. It doesn't sounds handy.

shinrich: Yeah, that crossed my mind as I pushed the branch up.  Let me take a pass at generalizing.

include/ts/ts.h
```diff
@@ -1315,6 +1315,8 @@ namespace c
   TSReturnCode TSVConnTunnel(TSVConn sslp);
   /*  Return the SSL object associated with the connection */
   TSSslConnection TSVConnSslConnectionGet(TSVConn sslp);
+  /* Return the file descriptoer associated with the ssl connection */
```
maskit, line 1318: Please remove "ssl".

maskit:
The change looks almost fine, but I don't see API proposal for this on the list.

include/ts/ts.h
zwoop: From the comment :)

zwoop (changes requested):

shinrich:
Sent the API email to the list a couple days ago.  Will make a new PR with the remaining changes next week.

bneradt (approved):

zwoop (approved):

PR #10319, opened 2023-08-31
Fix lua plugin build

masaori335:
Fix #10318. We start using c++ compiler and the symbol `log` conflicts with `math.h`.

```
lua/ts_lua_log.cc:21:24: error: redefinition of 'log' as different kind of symbol
static TSTextLogObject log;
                       ^
/opt/brew/opt/llvm@16/bin/../include/c++/v1/math.h:827:1: note: previous definition is here
log(_A1 __x) _NOEXCEPT {return __builtin_log((double)__x);}
^
```

PR #10326, opened 2023-08-31
Fix example plugins build

masaori335:
More fix of the `log` symbol issue. https://github.com/apache/trafficserver/pull/10319

### DIFF
--- a/doc/developer-guide/plugins/plugin-management/logging-api.en.rst
+++ b/doc/developer-guide/plugins/plugin-management/logging-api.en.rst
@@ -65,14 +65,14 @@ The steps below show how the logging API is used in the
 
    .. code-block:: c
 
-         static TSTextLogObject log;
+         static TSTextLogObject ts_log;
 
 #. In ``TSPluginInit``, a new log object is allocated:
 
    .. code-block:: c
 
            TSReturnCode error = TSTextLogObjectCreate("denylist",
-                                TS_LOG_MODE_ADD_TIMESTAMP, &log);
+                                TS_LOG_MODE_ADD_TIMESTAMP, &ts_log);
 
    The new log is named ``denylist.log``. Each entry written to the log
    will have a timestamp. The ``nullptr`` argument specifies that the new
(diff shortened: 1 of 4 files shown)


zwoop (approved):

jpeach:
I went down a bit of a rabbit hole on this one ....

IIUC this is a lllvm libc++ problem. `<math.h>` comes in via `<atomic>` which comes in via `<DbgCtl.h>`
```
.. ../../../include/ts/DbgCtl.h
... /usr/local/opt/llvm/bin/../include/c++/v1/atomic
.... /usr/local/opt/llvm/bin/../include/c++/v1/cstring
.... /usr/local/opt/llvm/bin/../include/c++/v1/cmath
..... /usr/local/opt/llvm/bin/../include/c++/v1/math.h
```

Relevant libc++ header says:
```C
  #if !defined(_LIBCPP_REMOVE_TRANSITIVE_INCLUDES) && _LIBCPP_STD_VER <= 20
  #  include <cmath>
  #  include <compare>
  #  include <type_traits>
  #endif
```

So in theory, setting that define should help. However, if we do that, we end up with a different path:
```
... /usr/local/opt/llvm/bin/../include/c++/v1/utility
.... /usr/local/opt/llvm/bin/../include/c++/v1/compare
..... /usr/local/opt/llvm/bin/../include/c++/v1/__compare/compare_partial_order_fallback.h
...... /usr/local/opt/llvm/bin/../include/c++/v1/__compare/partial_order.h
....... /usr/local/opt/llvm/bin/../include/c++/v1/__compare/compare_three_way.h
....... /usr/local/opt/llvm/bin/../include/c++/v1/__compare/weak_order.h
........ /usr/local/opt/llvm/bin/../include/c++/v1/__compare/strong_order.h
......... /usr/local/opt/llvm/bin/../include/c++/v1/cmath
.......... /usr/local/opt/llvm/bin/../include/c++/v1/math.h
```

And in `<cmath>` we find code like this, which unconditionally pulls `<math.h>` into the global namespace:
```C
...
#include <math.h>
...
_LIBCPP_BEGIN_NAMESPACE_STD
...
_LIBCPP_END_NAMESPACE_STD
...
```

So yeh, c++ seems like it's a bug mess with regard to `<math.h>` symbols.

jpeach (approved):

PR #10327, opened 2023-09-01
Cmake autest

cmcfarlen:
This allows running autest with a cmake target.  It also provides a helper script to run it again for iterative development.  I imagine CI will use the target while a developer would use the target to set it up for them and then use the script with filters to iterate on creating tests.

I tried to install trafficserver with DESTDIR and run autest that way but traffic_layout returns the wrong paths in this case so that will have to get sorted if we want that to work.  Just throwing ats in /tmp seems just fine, and CI builds don't really need this feature.

This does build the test plugins out of source now, so no more .libs directories in the source tree.

The helper script ends up at ${CMAKE_BINARY_DIRECTORY}/tests/autest.sh.  It has all of the options for resolving ats bin and everything but you can add additional options to filter output etc.

```
$ cmake -B test-autest-build -GNinja -DENABLE_AUTEST=ON -DCMAKE_INSTALL_PREFIX=/tmp/ats-autest .
... build generation, proxy-verifier install, python/pipenv detection
$ cmake --build test-autest-build --target autest
... build project, install, pipenv install, autest
$ cd cmake-autest-build/tests
$ ./autest.sh --filter TSVConnFd
Running Test TSVConnFd:.. Passed

Total of 1 test
  Unknown: 0
  Exception: 0
  Failed: 0
  Warning: 0
  Skipped: 0
  Passed: 1
```

The status of autest builds with cmake is still quite poor, but at least this will get us all going to get those fixed.


CMakeLists.txt
```diff
@@ -314,11 +315,19 @@ check_symbol_exists(
...
 if(ENABLE_DOCS)
-  find_package(Python3 REQUIRED)
-  find_package(Java COMPONENTS Runtime REQUIRED)
-  find_program(PipEnv pipenv REQUIRED)
+    find_package(Java COMPONENTS Runtime REQUIRED)
+endif()
+
+if(ENABLE_AUTEST)
+  set(PROXY_VERIFIER_VERSION "v2.10.1")
+  set(PROXY_VERIFIER_HASH "SHA1=0f189a37596d7488b5b81b5547df6fc1eadf56a1")
```
bneradt, line 329: Kind of handy to have these both specified here.

bneradt (commented):
Looks promising.

bneradt (approved):

PR #10359, opened 2023-09-01
(DNM) update WASM plugin to support http calls and return value of TSHttpTxnIsInternal() calls

thebadpete:


shukitchan:
[approve ci]

plugins/experimental/wasm/ats_context.cc
```diff
@@ -565,11 +565,17 @@ Context::httpCall(std::string_view target, const Pairs &request_headers, std::st
       path = value;
     } else if (key == ":authority") {
       authority = value;
+    } else if (key == ":protocol") {
```
shukitchan, line 568: It should be ":scheme" instead according to 
https://portswigger.net/burp/documentation/desktop/http2/http2-basics-for-burp-users
Please check out the section in pseudo-headers.

plugins/experimental/wasm/ats_context.cc
```diff
@@ -1325,6 +1331,10 @@ Context::getProperty(std::string_view path, std::string *result)
     int64_t total   = h_bytes + b_bytes;
     result->assign(reinterpret_cast<const char *>(&total), sizeof(int64_t));
     return WasmResult::Ok;
+  } else if (path.substr(0, p_txn_is_internal.size()) == p_txn_is_internal) {
+    int internal = TSHttpTxnIsInternal(txnp_);
+    result->assign(reinterpret_cast<const char *>(&internal), sizeof(int));
+    return WasmResult::Ok;
```
shukitchan, line 1337: I am actually thinking of using foreign function instead for this. e.g. in envoy it has "compress and "uncompress" foreign function exposed to wasm modules.

This is how it is defined - 
https://github.com/envoyproxy/envoy/blob/d12d47b05a832ce04d9130a869fa72f7ddbcef28/source/extensions/common/wasm/foreign.cc#L43

This is how it is used -
https://github.com/envoyproxy/envoy/blob/d12d47b05a832ce04d9130a869fa72f7ddbcef28/test/extensions/common/wasm/test_data/test_cpp.cc#L158

Please take a look and see if we can implement this using foreign function

shukitchan (changes requested):
Please take a look at my comments. Thanks.

shukitchan:
And let's move this PR to draft for now till it is ready.

thebadpete:
closing this pr for now, will refile after addressing the feeedback

Issue #10363, opened 2023-09-01
CID 1518569: String not null terminated


maskit:
money_trace.cc
```
232        bwriter.write(uuidstr, strlen(uuidstr));
233        bwriter.write(sep);
234        bwriter.write(parentid);
235        bwriter.write(new_parent);
236        bwriter.write(sep);
237        bwriter.write(spanid);
238        bwriter.print("{}", TSHttpTxnIdGet(txnp));
239
   	
CID 1518569 (#1 of 1): String not null terminated (STRING_NULL)
6. string_null: Passing unterminated string bwriter._buffer to view, which expects a null-terminated string.
240        header = std::string{bwriter.view()};
241      } else {
242        LOG_ERROR("Error getting uuid string");
243      }
```

PR #10373, opened 2023-09-05
money_trace cid 1518569: string not null terminated

traeak:
Fixes coverity #10363

Probably coverity issue, assigning std::string from swoc::BufferWriter.


maskit:
Yeah, looks like a false positive, but I'm fine with either way.

maskit (approved):

PR #10383, opened 2023-09-07
Check return values, log errors, and more tests

### LINKED ISSUE #10138: H3 error handling issue
It seems the h3 error handling logic needs to be added/updated. Currently, the `Http3Error` returned by the frame handler and dispatcher is not checked in `Http3Transaction`.

duke8253:
This resolves #10138

### DIFF
--- a/proxy/http3/Http3Transaction.cc
+++ b/proxy/http3/Http3Transaction.cc
@@ -563,8 +563,12 @@ Http3Transaction::_process_read_vio()
   SCOPED_MUTEX_LOCK(lock, this->_info.read_vio->mutex, this_ethread());
 
   uint64_t nread = 0;
-  this->_frame_dispatcher.on_read_ready(this->_info.adapter.stream().id(), Http3StreamType::UNKNOWN,
-                                        *this->_info.read_vio->get_reader(), nread);
+  auto error     = this->_frame_dispatcher.on_read_ready(this->_info.adapter.stream().id(), Http3StreamType::UNKNOWN,
+                                                         *this->_info.read_vio->get_reader(), nread);
+  if (error && error->cls != Http3ErrorClass::UNDEFINED) {
+    Http3TransDebug("Error occured while processing read vio: %hu, %s", error->get_code(), error->msg);
+    return 0;
+  }
   this->_info.read_vio->ndone += nread;
   return nread;
 }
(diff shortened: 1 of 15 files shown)


duke8253:
[approve ci cmake]

proxy/http3/Http3Types.h
```diff
@@ -95,29 +95,37 @@ enum class Http3ErrorCode : uint16_t {
...
+    : cls(Http3ErrorClass::APPLICATION), code(error_code), msg(error_msg), type(error_type){};
 };
 
 class Http3ConnectionError : public Http3Error
 {
 public:
   Http3ConnectionError() : Http3Error() {}
-  Http3ConnectionError(const Http3ErrorCode error_code, const char *error_msg = nullptr) : Http3Error(error_code, error_msg){};
+  Http3ConnectionError(const Http3ErrorCode error_code, const char *error_msg = nullptr)
+    : Http3Error(error_code, error_msg, Http3ErrorType::CONNECTION_ERROR){};
```
maskit, line 128: If we are going to have error type, we no longer need this sub class (and the one for stream as well).

proxy/http3/Http3Transaction.cc
```diff
@@ -581,8 +585,12 @@ Http3Transaction::_process_write_vio()
 
   size_t nwritten = 0;
   bool all_done   = false;
-  this->_frame_collector.on_write_ready(this->_info.adapter.stream().id(), *this->_info.write_vio->get_writer(), nwritten,
-                                        all_done);
+  auto error      = this->_frame_collector.on_write_ready(this->_info.adapter.stream().id(), *this->_info.write_vio->get_writer(),
+                                                          nwritten, all_done);
+  if (error && error->cls != Http3ErrorClass::NONE) {
```
maskit, line 590: What case does `error->cls` become `Http3ErrorClass::NONE` in?

maskit: It was probably needed when we had NoError class. I think we can remove it now.

duke8253: Yeah, good point, I'll make the changes.

duke8253: I think it is because the `QUICErrorClass` uses a similar structure:
```
enum class QUICErrorClass {
  UNDEFINED,
  TRANSPORT,
  APPLICATION,
};
```

But in the docs I didn't see there are any other types for `Http3Errors`, so I think we can just remove the error class complete for `Http3Errors`

maskit: Let's do that.

maskit:
I'll approve this when CI passes.

maskit (approved):

PR #10384, opened 2023-09-07
Coverity 1518612: Remove dead code

serrislew:
Fixes: #10332 

`out_end` cannot be nullptr, remove conditional 
- `*out_len` > 0 is always true

### DIFF
--- a/plugins/experimental/cookie_remap/strip.cc
+++ b/plugins/experimental/cookie_remap/strip.cc
@@ -189,7 +189,7 @@ get_stripped(const char *in, ssize_t in_len, char *out, int *out_len, unsigned i
   char *w, *out_end;      /* where we write to, write limit */
 
   /* validate params */
-  if (in == nullptr || in_len < 0 || out_len == nullptr || *out_len < 0 || (out == nullptr && *out_len > 0) ||
+  if (in == nullptr || in_len < 0 || out_len == nullptr || *out_len <= 0 || (out == nullptr && *out_len > 0) ||
       (flags & (~allowed_flags))) {
     if (out != nullptr && out_len != nullptr && *out_len > 0) {
       *out     = '\0';
--- a/plugins/experimental/cookie_remap/strip.cc
+++ b/plugins/experimental/cookie_remap/strip.cc
@@ -228,7 +228,7 @@ get_stripped(const char *in, ssize_t in_len, char *out, int *out_len, unsigned i
   retval = stripped_core(r, in_end, &w, out_end, flags);
 
   /* null terminate */
-  out_end += out_end ? 1 : 0;            /* undo decrement at start */
+  out_end++;                             /* undo decrement at start */
   write_char_if_room(&w, out_end, '\0'); /* try to term at end of output */
 
   /* report the required/used length */


plugins/experimental/cookie_remap/strip.cc
```diff
@@ -228,7 +228,7 @@ get_stripped(const char *in, ssize_t in_len, char *out, int *out_len, unsigned i
   retval = stripped_core(r, in_end, &w, out_end, flags);
 
   /* null terminate */
-  out_end += out_end ? 1 : 0;            /* undo decrement at start */
+  out_end++;                             /* undo decrement at start */
```
bneradt, line 231: A simple enough change. But for my benefit, can you explain why out_end is always non-zero? It wasn't obvious to me reading the function that it would be. At 211 it's set to:

```cpp
out_end = out + *out_len;
```

I think char is signed with some compilers, so one or both of those could be negative, which complicates things.

serrislew: maybe always was misleading but `get_stripped` only gets called here [`CookieJar::verify_value`] (https://github.pie.apple.com/ats/trafficserver/blob/master/plugins/experimental/cookie_remap/cookiejar.cc#L178) with the above code setting `out_len` > 0

bneradt: OK, so this isn't an issue with your patch, but I'm concerned this code has a one byte buffer overrun potential:

1. The caller passes in a pointer to a buffer that will be filled called `out` and a length for the allocated space called `out_len`.
2. Then we set `out_end` accordingly: `out_end = out + *out_len;`
3. OK, so out_end points to the end of the allocated buffer.
4. Then we increment one past the end of that buffer: `out_end++;`
5. And then write into the `out` buffer pointed to by `w`.

The fact is, we are probably always well under `out_len`, so this doesn't cause problems. But isn’t that a one byte buffer overrun waiting to happen? 

In other words, the caller said "Don't write more than `out_len` bytes to `out`." Then we say, "OK, I'll write at most `out_len + 1` bytes to `out`.

Let me know whether I'm missing something.

bneradt: @serrislew pointed out that earlier, in line 202, `out_len` was decremented, so this increment for `\0` should be safe per the incoming buffer.

bneradt (approved):

bneradt (commented):
So is this OK if `out` is non-null, and `*out_len` is 0? In that case, we won't decement `*out_len` for a null terminator.

plugins/experimental/cookie_remap/strip.cc
bneradt, line 231: Should we add a parameter check at line 192 that if `out` is non-nullptr, then `*out_len > 0` ?

bneradt (changes requested):

plugins/experimental/cookie_remap/strip.cc
bneradt: I think, in that case, there will be no space for the `out_end++`, right?

bneradt (approved):

PR #10385, opened 2023-09-07
Allow origins to do TLS renegotiation

bryancall:
Ran into an issue where IIS will do a TLS renegotiation after the first request is made.  Since the handshake has already been completed it is failing this check.

CI has a test for TS_SSL_VERIFY_SERVER_HOOK and I will see if this change breaks that functionality.

Discussion about IIS: https://security.stackexchange.com/questions/24554/should-i-use-ssl-tls-renegotiation

### DIFF
--- a/iocore/net/P_SSLNetVConnection.h
+++ b/iocore/net/P_SSLNetVConnection.h
@@ -89,7 +89,7 @@ typedef enum {
   SSL_HOOK_OP_LAST = SSL_HOOK_OP_TERMINATE ///< End marker value.
 } SslVConnOp;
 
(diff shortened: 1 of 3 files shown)


bryancall:
This commit has the logic change: https://github.com/apache/trafficserver/pull/10385/commits/ede1f34e6260551726bc9e466c886278d305980c the rest of the PR is just cleanup.

shinrich:
Logic seems reasonable.  Do we want to make this configurable?  I dimly recall that renegotiation is frowned up these days.  If it is configurable, I assume that is something we can set during the TLS negotiation to indicate whether the client will handle renegotiations.

Oh I see the 10 year old article you reference does discuss some of the renegotiation concerns.

shinrich (approved):
Server initiated renegotiation is less concerning.

PR #10304, opened 2023-08-29
Preserve unmapped url regardless of need for remapping

maskit:
https://github.com/apache/trafficserver/pull/10146 discovered that marking a parent host down causes a crash if a request is set to skip remapping. Not only OCSP requests but also any requests that skip remapping can cause the crash.

```
(lldb) bt
* thread #3, name = '[ET_NET 0]', stop reason = signal SIGABRT
  * frame #0: 0x000000018c190764 libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x000000018c1c7c28 libsystem_pthread.dylib`pthread_kill + 288
    frame #2: 0x000000018c0d5ae8 libsystem_c.dylib`abort + 180
    frame #3: 0x0000000101060878 traffic_server`ink_abort(message_format="%s:%d: failed assertion `%s`") at ink_error.cc:99:3
    frame #4: 0x00000001010555bc traffic_server`_ink_assert(expression="valid()", file="/Users/mkitajo/src/github.com/trafficserver/proxy/hdrs/URL.h", line=595) at ink_assert.cc:37:3
    frame #5: 0x00000001000b17c0 traffic_server`URL::host_get(this=0x0000000113017718, length=0x000000010af12760) at URL.h:595:3
    frame #6: 0x000000010038cbcc traffic_server`HttpSM::mark_host_failure(this=0x0000000113016ae0, info=0x0000000113016cb8, time_down=ts_time @ 0x000000010af126a0) at HttpSM.cc:5750:62
    frame #7: 0x000000010038b824 traffic_server`HttpSM::do_hostdb_update_if_necessary(this=0x0000000113016ae0) at HttpSM.cc:4532:11
    frame #8: 0x00000001004771e4 traffic_server`HttpTransact::handle_server_connection_not_open(s=0x0000000113016be8) at HttpTransact.cc:3899:21
    frame #9: 0x000000010048ca78 traffic_server`HttpTransact::handle_response_from_server(s=0x0000000113016be8) at HttpTransact.cc:3799:7
    frame #10: 0x0000000100452034 traffic_server`HttpTransact::HandleResponse(s=0x0000000113016be8) at HttpTransact.cc:3450:5
```
The backtrace above shows ink_abort but a real crash happens on production (non-debug build).

PR #10386, opened 2023-09-07
Don't set port number as part of hostname

maskit:
One of autest failed on https://github.com/apache/trafficserver/pull/10304 , and it discovered that there is a place mishandling Host header value.

The original code used the entire Host header value as a hostname, and it confused `url_print()` because "hostname" contains a colon. `url_print()` treated the hostname as IPv6 address and it added brackets to the hostname like `[foo.example:8888]`. If it's used in an URL (i.e. `http://[foo.example:8888]/`), it means/will be read as an (invalid) IPv6 address `foo.exmample:8080` and port `80`.

This PR fix the issue. The behavior is described in the comments.

### DIFF
--- a/proxy/http/HttpSM.cc
+++ b/proxy/http/HttpSM.cc
@@ -4376,7 +4376,40 @@ HttpSM::do_remap_request(bool run_inline)
       int host_len;
       const char *host_name = host_field->value_get(&host_len);
       if (host_name && host_len) {
+        int port = -1;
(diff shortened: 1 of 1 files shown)


maskit:
@treak Basically what we discussed yesterday. I added IPv6 support.

traeak (approved):
This should take care of the ipv6 host/port setting issue.

zwoop:
Cherry-picked to v9.2.x

PR #9267, opened 2022-12-20
fix ProxyVerifier version check

duke8253:

PR #10389, opened 2023-09-08
9.2.x: Proxy Verfier Upgrade to v2.10.1

bneradt:
This is a set of Proxy Verifier cherry-picks from master into 9.2.x to get 9.2.x up to date with the latest v2.10.1 Proxy Verifier release.

---

Fix the autest Proxy Verifier version check (#9365)

The Proxy Verifier autest version check logic assumed that verifier-client was in the PATH. This is most often not the case, and even if it is it can be the wrong one since the test will by default use the unpacked Proxy Verifier from bintray. This fixes the check so that it explicitly refers to the verifier-client that the test will use.

(cherry picked from commit edbd10a8ad0be7df2cf6fa9474833b9006bcc994)

Upgrade to Proxy Verifier v2.7.0 (#9611)

Proxy Verifier v2.7.0 adds:

* Proxy Protocol support.
* HTTP response reason string verification

Some of the h2origin AuTest tests were over-specified with respect to reason strings. HTTP/2 explicitly removes reason strings, so they will not be in the responses nor will the verifier client find them. This removes those incorrect specifications.

(cherry picked from commit b449992b1d14cb1897202e845ec04c9eaebaf8c5)

Fix a crash caused by a HTTP/2 GET request with a body (#9738)

* fix a crash caused by a HTTP/2 GET request with a body

* Update to Proxy Verifier v2.8.0

This updates our test suite to use Proxy Verifier v2.8.0. This release comes with HTTP/2 frame replay updates.

In the cherry-pick from master to 9.2.x, I removed the following test updates because they were HTTP/2 to origin specific:

* h2get_with_body
* background_fill

(cherry picked from commit 0f70092d22f439681858cbe7aa31fe6dfb383819)

---------

Co-authored-by: bneradt <bneradt@yahooinc.com>
(cherry picked from commit 50b6aa9217d48b0922cc06bd0f11b765dfbca207)

Update to Proxy Verifier v2.8.1 (#9834)

Proxy Verifier v2.8.1 has fixes for the way Proxy Verifier relates to HTTP/2 CONNECT method request pseudo header fields. This will be helpful for testing HTTP/2 CONNECT requests.

(cherry picked from commit d5c47a7b743bd798b9827c175b35747f59f40eca)

Proxy Verifier: update to v2.9.0 (#10083)

Proxy Verifier v2.9.0 adds HTTP/2 trailer support as well as containing updates for Expect: 100-continue and quic timeout.

(cherry picked from commit 183aa3efe325dc439797d7884a15d4517574b8ea)

Update Proxy Verifier to v2.10.1 (#10322)

This includes the ability to specify separate DATA frames for a stream as well as a way to keep a session open after the final transaction.

(cherry picked from commit c5d7b19d60646de902fc847f80f57030779a4bfc)

fix ProxyVerifier version check (#9267)

(cherry picked from commit a2ab0b7d7544b2486ab133f9400907ce9218ec33)


bneradt:
Huh. Looks like `tls_verify4` fails, but it fails for the 9.2.x base branch too. We should investigate that and address that with a separate PR.

PR #10390, opened 2023-09-08
Remove TSDebug() and TSIsDebugTagSet() TS API functions.

ywkaras:
Replace with Dbg() and DbgCtl::on().  This required extensive changes to the ESI plugin unit tests.

Doc changes were also necessary.

The redirect_1 example plugin is changed to illustrate how to implement debug trace output in a plugin that can compile and run with either these ATS10 changes or for older versions of ATS.


ywkaras:
@cmcfarlen this is passing CI now.

ywkaras:
@cmcfarlen It's successfully rebased.

doc/developer-guide/api/functions/TSTrafficServerVersionGet.en.rst
```diff
@@ -55,6 +55,8 @@ Example
 
     #define PLUGIN_NAME "hello_world"
 
+    DbgCtl dbg_ctl{PLUGIN_NAME};
```
cmcfarlen, line 58: Do you want to establish the convention of putting these in anonymous namespaces here?

example/plugins/c-api/protocol/Protocol.h
```diff
@@ -36,6 +36,12 @@
 
 #define PLUGIN_NAME "protocol"
 
+namespace protocol_ns
+{
+extern DbgCtl dbg_ctl;
+}
+using namespace protocol_ns;
```
cmcfarlen, line 43: I know this is just an example, but putting this `using namespace` in a header might lead to user astonishment.

plugins/cache_promote/policy.h
```diff
@@ -28,6 +28,10 @@ extern int TXN_ARG_IDX;
 #define MAX_STAT_LENGTH (1 << 8)
 extern const char *PLUGIN_NAME;
 
+extern DbgCtl cache_promote_dbg_ctl;
+
+#define DBG(...) Dbg(cache_promote_dbg_ctl, __VA_ARGS__)
```
cmcfarlen, line 33: Why is this plugin converted with a macro rather than `Dbg`?

plugins/esi/esi.cc
```diff
@@ -139,6 +154,7 @@ struct ContData {
       input_type(DATA_TYPE_RAW_ESI),
       packed_node_list(""),
       gzipped_data(""),
+      dbg_ctl{createDebugTag(DEBUG_TAG, contptr)},
```
cmcfarlen, line 157: Won't this insert a lot of tags into the registry? How often is `ContData` instantiated?

cmcfarlen (commented):
This is mostly mechanical.  I am concerned that the esi plugin will generate a lot of tags for the tag registry.  Can you explain how that works?

doc/developer-guide/api/functions/TSTrafficServerVersionGet.en.rst
ywkaras: I think it's not that simple.  If I add an anonymous namespace, `check_ts_version()` probably belongs in it also.  But, if I make such changes consistently, in all the files I'm touching, it will likely double the size of this PR.  I get a lot of surprising push back in reviews on this project.  For example, push back that seems to dispute the value of RAII.  Which makes me feel like adding any sort of cleanup risks adding further delay to merging PRs.

example/plugins/c-api/protocol/Protocol.h
ywkaras: I see your point, but what do you see as the better alternative?

TS developers must steel themselves to endure astonishment, and not in a good way.

plugins/cache_promote/policy.h
ywkaras: I think I did this before I thought of the other approach with `using namespace`.

plugins/esi/esi.cc
ywkaras: If you create multiple DbgCtl instances for the same tag, they will all point to the same entry in the registry map.  In any case, I think each instance of this continuation class will have a unique debug tag.  So there's a net increase, per class instance, of 8 bytes for the pointer in the DbgCtl, plus the overhead of an additional std::map entry.

doc/developer-guide/api/functions/TSTrafficServerVersionGet.en.rst
cmcfarlen: I just mean in this example, since folks tend to copy pasta examples.  If we want to prefer anonymous namespaces, having that in the example will help with proliferation.

plugins/esi/esi.cc
cmcfarlen: It looks like a new ContData instance is created per transaction that this plugin handles.  Since the tag is created using the address of the `contptr` instance, I feel like there is a good chance that there will be a lot of additional tags due to this.

ywkaras: But my changes don't increase the number of instances of ContData.  I see the issue, but what can I do about it?

cmcfarlen: It seems like these tags are generated to add correlation context to the log statements and not as a way to filter them with the tagging mechanism.  One idea would be to move this context from the tag name into the log message and have a fixed number of dbgctl tags.

ywkaras: But does that really belong in *this* PR?  The change you suggest would reduce memory use in the existing code.  It can't clearly be claimed it's addressing an issue that I created.

ywkaras: OK, valid point.

ywkaras: Fixed

cmcfarlen (approved):
Nicely done.  Thank you!

PR #10394, opened 2023-09-11
CID 1518583: Unchecked return value

jasmine-nahrain:


plugins/experimental/cache_fill/cache_fill.cc
```diff
@@ -72,7 +72,10 @@ cont_check_cacheable(TSHttpTxn txnp)
     return false;
   }
   int lookupStatus;
-  TSHttpTxnCacheLookupStatusGet(txnp, &lookupStatus);
+  if (TSHttpTxnCacheLookupStatusGet(txnp, &lookupStatus) == TS_ERROR) {
+    TSError("[esi][%s] Couldn't get cache status of object", __FUNCTION__);
```
masaori335, line 76: `esi` is the name of the other plugin. Please replace it with `cache_fill`.

masaori335:
[approve ci]

plugins/experimental/cache_fill/cache_fill.cc
```diff
@@ -72,7 +72,10 @@ cont_check_cacheable(TSHttpTxn txnp)
     return false;
   }
   int lookupStatus;
-  TSHttpTxnCacheLookupStatusGet(txnp, &lookupStatus);
+  if (TSHttpTxnCacheLookupStatusGet(txnp, &lookupStatus) == TS_ERROR) {
+    TSError("[cache_fill][%s] Couldn't get cache status of object", __FUNCTION__);
```
bneradt, line 76: Better to refer to PLUGIN_NAME rather than the literal "cache_fill". See:
https://github.com/apache/trafficserver/pull/10394/files#diff-896f422b7d09fb7725036af2885bbbca071b822167a1f8a99fbcb3ccd9d46f40R124

Also, it's probably better to not include the `__FUNCTION__` in an error message. You already worded the message to be unique enough to identify where the Error is coming from, so we should be in good shape as far as that goes.

bneradt (changes requested):
Looks good. I just have a quick comment about the TSError code.

bneradt (approved):

PR #10397, opened 2023-09-11
CID-1508974

sharkleash:
Check return value of TSMutexTryLock before calling continuation, retry if return value is TS_Error


duke8253 (approved):
Seems fine by me.

bryancall (approved):

PR #10409, opened 2023-09-13
Fixed HTTP method metrics not updating

lzx404243:
Currently, the HTTP request method metrics such as `proxy.process.http.get_requests` are not updating properly for non-tunneled traffic. This PR fixes this and added some test coverage.

In addition, while I am here, I cleaned up the redundant call to `build_error_response()` in `HttpTransact::SelfLoop`, as this is already called right before in `HttpTransact::will_this_request_self_loop()`, with the same arguments but more informative reason phrase.

### DIFF
--- a/proxy/http/HttpTransact.cc
+++ b/proxy/http/HttpTransact.cc
@@ -891,7 +891,6 @@ HttpTransact::SelfLoop(State *s)
 {
   TxnDebug("http_trans", "Request will selfloop.");
   bootstrap_state_variables_from_request(s, &s->hdr_info.client_request);
-  build_error_response(s, HTTP_STATUS_BAD_REQUEST, "Direct self loop detected", "request#cycle_detected");
   TRANSACT_RETURN(SM_ACTION_SEND_ERROR_CACHE_NOOP, nullptr);
 }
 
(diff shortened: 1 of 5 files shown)


bneradt (commented):
Looks good to me. I'll let a non-Yahoo person approve.

Thank you @lzx404243 for fixing these stats.

zwoop:
Can you update this with the new metrics? :)

proxy/http/HttpTransact.h
```diff
@@ -956,6 +956,8 @@ class HttpTransact
       return std::min(cur_parent_max_attempts, txn_conf->parent_connect_attempts) - 1;
     }
 
+    bool is_method_stats_incremented = false;
```
zwoop, line 959: This feels like an odd place to add this member. Can you move it up where all the other bool members are?

proxy/http/HttpTransact.cc
```diff
@@ -5807,6 +5783,34 @@ HttpTransact::initialize_state_variables_from_request(State *s, HTTPHdr *obsolet
   s->dns_info.lookup_name = s->server_info.name;
 }
 
+void
```
zwoop, line 5786: While we're at it, would this not be better / nicer with a switch statement ?

proxy/http/HttpTransact.cc
```diff
@@ -891,7 +891,6 @@ HttpTransact::SelfLoop(State *s)
 {
   TxnDebug("http_trans", "Request will selfloop.");
   bootstrap_state_variables_from_request(s, &s->hdr_info.client_request);
-  build_error_response(s, HTTP_STATUS_BAD_REQUEST, "Direct self loop detected", "request#cycle_detected");
```
zwoop, line 894: Is this line intentionally removed ?

proxy/http/HttpTransact.h
lzx404243: Fixed.

proxy/http/HttpTransact.cc
lzx404243: I tried, but values like `HTTP_WKSIDX_GET` is not compile-time constants so I left it the way it was.

proxy/http/HttpTransact.cc
lzx404243: Yes this is not related to the fix. This is a cleanup while I am here and I noted that in the description.

lzx404243:
> Can you update this with the new metrics? :)

@zwoop  Updated to use the new metrics. I like it 😄 .

zwoop (approved):

PR #10410, opened 2023-09-13
Add a pristine option for strategies peering

ezelkow1:
Adds a `use_pristine` option to the strategies configuration. Currently when using peering we use the post-remap URL which does not work for peering as then all peers have to have remaps of each other in order to handle the new fqdn. This option replaces it with the original pristine URL requested by the client so all peers can maintain the same set of remaps


ywkaras:
Would be good to have some Au test coverage of this.

jrushford (approved):
Looks good.

PR #10442, opened 2023-09-14
Fix Coverity issue in inliner plugin.

ywkaras:
CID 1508893

### DIFF
--- a/plugins/experimental/inliner/cache-handler.h
+++ b/plugins/experimental/inliner/cache-handler.h
@@ -326,12 +326,12 @@ namespace inliner
       request             += std::string(b1, i);
       request             += "\r\n\r\n";
 
-      std::unique_ptr<ats::io::IO> io{new io::IO()};
+      auto io{std::make_unique<io::IO>()};
 
       TSDebug(PLUGIN_TAG, "request:\n%s", request.c_str());
 
-      auto r{io->copy(request)};
-      ats::get(std::move(io), r, AnotherClass(src_));
+      auto size{io->copy(request)};
+      ats::get(std::move(io), size, AnotherClass(src_));
     }
   };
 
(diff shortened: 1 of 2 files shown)


zwoop:
I think this looks good, but I'd like @cmcfarlen  to review.

plugins/experimental/inliner/cache-handler.h
```diff
@@ -326,11 +326,12 @@ namespace inliner
       request             += std::string(b1, i);
       request             += "\r\n\r\n";
 
-      ats::io::IO *const io = new io::IO();
+      std::unique_ptr<ats::io::IO> io{new io::IO()};
 
       TSDebug(PLUGIN_TAG, "request:\n%s", request.c_str());
```
ywkaras, line 331: It's not clear what exactly Coverity's gripe is.  My best guess is that it's seeing that `*io` would get leaked if TSDebug() threw an exception.  But there must be a bazillion cases of that in our code, why would it single this one out?  Are there instructions anywhere about how to run Coverity, to be sure we've appeased it?

zwoop: No unfortunately we don’t have enough karma to run PRs against coverity. So it’ll have to land on master for the next schedule run, rinse repeats…

cmcfarlen: It looks like from reading the defect report, coverity doesn't trace into the second get call to see that it is indeed assigned eventually.  The std::move will hopefully realize that ownership is handed off.

cmcfarlen (approved):
This looks good to me.

cmcfarlen (approved):

ywkaras:
[autest ci autest]

ywkaras:
[approve ci autest]

PR #10443, opened 2023-09-15
CID1508860: double lock confusion

### LINKED ISSUE #10430: CID 1508860: Double lock
AIO.cc
```
529      if (op->thread == AIO_CALLBACK_THREAD_AIO) {
530        SCOPED_MUTEX_LOCK(lock, op->mutex, thr_info->mutex->thread_holding);
531        op->handleEvent(EVENT_NONE, nullptr);
   	12. Falling through to end of if statement.
   	24. Falling through to end of if statement.
   	37. Condition op->thread == NULL, taking true branch.
532      } else if (op->thread == AIO_CALLBACK_THREAD_ANY) {
533        eventProcessor.schedule_imm(op);
   	38. Falling through to end of if statement.
534      } else {
535        op->thread->schedule_imm(op);
536      }
   	25. lock: ink_mutex_acquire locks my_aio_req->aio_mutex. [[show details](https://scan6.scan.coverity.com/eventId=34764104-39&modelId=34764104-0&fileInstanceId=164784741&filePath=%2Finclude%2Ftscore%2Fink_mutex.h&fileStart=46&fileEnd=52)]
   	
CID 1508860 (#1-2 of 2): Double lock (LOCK)
39. double_lock: ink_mutex_acquire locks my_aio_req->aio_mutex while it is locked.
537      ink_mutex_acquire(&my_aio_req->aio_mutex);
```

cmcfarlen:
I wonder if coverity is confused by following the `current_req` pointer for the unlock call vs `my_aio_req` for the lock.  I removed the `current_req` variable as it isn't necessary.

I also moved the release call before incrementing metrics since the lock doesn't need to be held for that.

Fixes #10430

### DIFF
--- a/iocore/aio/AIO.cc
+++ b/iocore/aio/AIO.cc
@@ -423,9 +423,8 @@ ink_aio_thread_num_set(int thread_num)
 void *
 AIOThreadInfo::aio_thread_main(AIOThreadInfo *thr_info)
 {
-  AIO_Reqs *my_aio_req  = thr_info->req;
-  AIO_Reqs *current_req = nullptr;
-  AIOCallback *op       = nullptr;
+  AIO_Reqs *my_aio_req = thr_info->req;
+  AIOCallback *op      = nullptr;
   ink_mutex_acquire(&my_aio_req->aio_mutex);
   for (;;) {
     do {
@@ -433,7 +432,6 @@ AIOThreadInfo::aio_thread_main(AIOThreadInfo *thr_info)
         ink_mutex_release(&my_aio_req->aio_mutex);
         return nullptr;
       }
-      current_req = my_aio_req;
       /* check if any pending requests on the atomic list */
       aio_move(my_aio_req);
       if (!(op = my_aio_req->aio_todo.pop())) {
@@ -442,8 +440,10 @@ AIOThreadInfo::aio_thread_main(AIOThreadInfo *thr_info)
 #ifdef AIO_STATS
       num_requests--;
       current_req->queued--;
-      ink_atomic_increment((int *)&current_req->pending, 1);
+      ink_atomic_increment((int *)&my_aio_req->pending, 1);
 #endif
+      ink_mutex_release(&my_aio_req->aio_mutex);
+
       // update the stats;
       if (op->aiocb.aio_lio_opcode == LIO_WRITE) {
         Metrics::increment(aio_rsb.write_count);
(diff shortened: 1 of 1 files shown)


iocore/aio/AIO.cc
```diff
@@ -442,8 +440,10 @@ AIOThreadInfo::aio_thread_main(AIOThreadInfo *thr_info)
 #ifdef AIO_STATS
```
ywkaras, line 440: How the outer `for` loop not an infinite loop?  The `break` on line 438 breaks out of the `do` loop only, no?

iocore/aio/AIO.cc
```diff
@@ -442,8 +440,10 @@ AIOThreadInfo::aio_thread_main(AIOThreadInfo *thr_info)
...
       num_requests--;
       current_req->queued--;
-      ink_atomic_increment((int *)&current_req->pending, 1);
+      ink_atomic_increment((int *)&my_aio_req->pending, 1);
```
ywkaras, line 443: You could get rid of the C-style cast.

iocore/aio/AIO.cc
```diff
@@ -423,17 +423,15 @@ ink_aio_thread_num_set(int thread_num)
 void *
 AIOThreadInfo::aio_thread_main(AIOThreadInfo *thr_info)
```
ywkaras, line 424: Not directly relevant to this PR, but `thr_info` is redundant, it's  the same as the `this` pointer in this function.

ywkaras:
The AIO code seems scary bad in general.

iocore/aio/AIO.cc
```diff
@@ -452,9 +452,8 @@ AIOThreadInfo::aio_thread_main(AIOThreadInfo *thr_info)
         Metrics::increment(aio_rsb.read_count);
         Metrics::increment(aio_rsb.kb_read, op->aiocb.aio_nbytes >> 10);
       }
-      ink_mutex_release(&current_req->aio_mutex);
       cache_op((AIOCallbackInternal *)op);
-      ink_atomic_increment(&current_req->requests_queued, -1);
+      ink_atomic_increment(&my_aio_req->requests_queued, -1);
 #ifdef AIO_STATS
       ink_atomic_increment((int *)&current_req->pending, -1);
```
ywkaras, line 458: It lives.

iocore/aio/AIO.cc
```diff
@@ -442,8 +440,10 @@ AIOThreadInfo::aio_thread_main(AIOThreadInfo *thr_info)
 #ifdef AIO_STATS
       num_requests--;
       current_req->queued--;
-      ink_atomic_increment((int *)&current_req->pending, 1);
+      ink_atomic_increment((int *)&my_aio_req->pending, 1);
 #endif
+      ink_mutex_release(&my_aio_req->aio_mutex);
```
ywkaras, line 445: Releasing early looks OK.

iocore/aio/AIO.cc
cmcfarlen: Yep, I missed some inside the conditional code.  Thanks!

ywkaras (approved):

PR #10447, opened 2023-09-15
Stop using functions that are unavailable on the latest quiche

### LINKED ISSUE #9800: Manage QUICStream instances without relying on quiche

maskit:
This fixes #9800.


zwoop (approved):

brbzull0 (approved):
Looks good to me!
Thanks.

PR #10457, opened 2023-09-18
Cmake presets

cmcfarlen:
CMakePresets.json will let you configure the project with a known set of options.  This can be useful for reproducing builds and should give us more outside control over CI pipeline builds.

To configure with a preset just use the `--preset` option to cmake and provide the name.  For example, to build `autests`:

```
cmake . --preset autest
cmake --build --preset autest
```

You can provide a `CMakeUserPresets.json` file to add additional presets.  This file is not to be checked in, and is included in `.gitignore`.  Here is an example:

```json
{
  "version": 2,
  "cmakeMinimumRequired": {
    "major": 3,
    "minor": 20,
    "patch": 0
  },
  "configurePresets": [
    {
      "name": "my-dev",
      "displayName": "my development",
      "description": "My Development Presets",
      "inherits": ["dev"],
      "cacheVariables": {
        "CMAKE_INSTALL_PREFIX": "/opt/ats-cmake",
        "jemalloc_ROOT": "/opt/homebrew"
      }
    }
  ]
}
```

This is like the `dev` preset, but changes the install prefix and the jemalloc root.

This is just to get us started.  As we get closer to using cmake, I'm sure we will expand this.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,112 @@
+{
+  "version": 2,
+  "cmakeMinimumRequired": {
+    "major": 3,
+    "minor": 20,
+    "patch": 0
+  },
+  "configurePresets": [
+    {
+      "name": "default",
+      "displayName": "Default build with ninja",
+      "description": "Default build using Ninja generator",
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/build-default",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug",
+        "CMAKE_COMPILE_WARNING_AS_ERROR": "ON"
+      }
+    },
+    {
+      "name": "autest",
+      "inherits": ["default"],
+      "binaryDir": "${sourceDir}/build-autest",
+      "cacheVariables": {
+        "ENABLE_AUTEST": "ON",
+        "CMAKE_INSTALL_PREFIX": "/tmp/ts-autest",
+        "BUILD_EXPERIMENTAL_PLUGINS": "ON",
+        "ENABLE_EXAMPLE": "ON"
+      }
+    },
+    {
+      "name": "dev",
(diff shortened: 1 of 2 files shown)


cmcfarlen:
[approve ci autest]

bneradt (changes requested):
Look great.

I suggest adding `CMAKE_COMPILE_WARNING_AS_ERROR` to more configurations. Devs find it annoying if they use one of these locally and find in CI that they had some warnings they missed because they weren't treated as errors.

cmcfarlen:
> Look great.
> 
> I suggest adding `CMAKE_COMPILE_WARNING_AS_ERROR` to more configurations. Devs find it annoying if they use one of these locally and find in CI that they had some warnings they missed because they weren't treated as errors.

With the inheritance, every preset should have this set to `ON`.  The top level ones are `default` and `ci`

bneradt:
> > Look great.
> > I suggest adding `CMAKE_COMPILE_WARNING_AS_ERROR` to more configurations. Devs find it annoying if they use one of these locally and find in CI that they had some warnings they missed because they weren't treated as errors.
> 
> With the inheritance, every preset should have this set to `ON`. The top level ones are `default` and `ci`

Ah, great. Thanks.

CMakePresets.json
```diff
@@ -0,0 +1,112 @@
...
+    },
+    {
+      "name": "dev",
+      "displayName": "development",
+      "description": "Development Presets",
+      "inherits": ["default"],
+      "binaryDir": "${sourceDir}/cmake-build-${presetName}",
+      "generator": "Ninja",
+      "cacheVariables": {
+        "CMAKE_EXPORT_COMPILE_COMMANDS": "ON",
```
bneradt, line 39: Nice.

bneradt (approved):

Issue #10463, opened 2023-09-19
Adjust PreWarm stats with new `ts::Metrics`


masaori335:
PreWarm is using DynamicStats, a wrapper of old librecord stats, but we can avoid it with the new `ts::Metrics` introduced by #9988.

PR #10464, opened 2023-09-19
Metrics: Add read/write function by id

masaori335:
While working on adjusting PreWarming stats to new `ts::Metircs` (#10463), I found it's helpful having sugar to read & write by id.


masaori335:
It turns out we don't need this. https://github.com/apache/trafficserver/pull/10465#discussion_r1332408210

PR #10465, opened 2023-09-19
Use ts::Metrics for PreWarm stats

masaori335:
Address #10463. Main change is getting rid of `DynamicStats stats` from `PreWarmManager` and directly accessing metrics via `ts::Metrics`.

~~Requires #10464.~~

### DIFF
--- a/iocore/net/PreWarmManager.h
+++ b/iocore/net/PreWarmManager.h
@@ -136,7 +136,7 @@ enum class Stat {
   LAST_ENTRY,
 };
 
-using StatsIds          = std::array<ts::Metrics::IdType, static_cast<size_t>(PreWarm::Stat::LAST_ENTRY)>;
+using StatsIds          = std::array<ts::Metrics::IntType *, static_cast<size_t>(PreWarm::Stat::LAST_ENTRY)>;
 using SPtrConstStatsIds = std::shared_ptr<const StatsIds>;
 using StatsIdMap        = std::unordered_map<SPtrConstDst, SPtrConstStatsIds, DstHash, DstKeyEqual>;
 
--- a/iocore/net/PreWarmManager.h
+++ b/iocore/net/PreWarmManager.h
@@ -32,8 +32,8 @@
 #include "NetTimeout.h"
 #include "Milestones.h"
 
+#include "api/Metrics.h"
 #include "records/I_RecHttp.h"
-#include "records/DynamicStats.h"
 
 #include <map>
 #include <unordered_map>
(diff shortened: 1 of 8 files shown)


iocore/net/PreWarmManager.h
```diff
@@ -136,7 +136,7 @@ enum class Stat {
   LAST_ENTRY,
 };
 
-using StatsIds          = std::array<int, static_cast<size_t>(PreWarm::Stat::LAST_ENTRY)>;
+using StatsIds          = std::array<ts::Metrics::IdType, static_cast<size_t>(PreWarm::Stat::LAST_ENTRY)>;
```
zwoop, line 139: Definitely not a big deal either way, but if you make this an array of `Metrics::IntType*`, you can use the static Metric::increment without needing the Metrix singleton instance.

masaori335: Yup, I thought it a bit, but I left it because it'll be big code change. 

Getting instance doesn't have noticeable over head, right?

zwoop: Yeh getting the instance should have no cost, the compiler should be smart about that. It’s a little bit of cost doing the indexing into the arrays

zwoop: But why would the code change be any bigger? You still just have to call the increment it write functions ?

zwoop: That pointer is the pointer directly to the atomic btw. It’s safe to use because the memory never moves or reallocates

masaori335: Each PreWarm Stat has 8 entries and getting metrics by the offset from the first entry. So changing it pointer makes... hmm...  let me try using `Metrics::IntType *`. My guess might be wrong.

https://github.com/apache/trafficserver/blob/84502fa3696c246bc8324e8ec05e556a26194ba6/iocore/net/PreWarmManager.h#L125-L137

masaori335: All right, I was just wrong. Using `Metrics::IntType *`, `ts::Metrics::increment` and ` `ts::Metrics::store` is much better. We don't need any sugar #10464. Thanks for checking.

src/records/RecordsConfig.cc
```diff
@@ -947,8 +947,6 @@ static const RecordElement RecordsConfig[] =
   ,
   {RECT_CONFIG, "proxy.config.tunnel.prewarm.enabled", RECD_INT, "0", RECU_RESTART_TS, RR_NULL, RECC_INT, "[0-1]", RECA_NULL}
   ,
-  {RECT_CONFIG, "proxy.config.tunnel.prewarm.max_stats_size", RECD_INT, "100", RECU_RESTART_TS, RR_NULL, RECC_INT, "[5-65536]", RECA_NULL}
```
zwoop, line 950: Love that we can get rid of this settings! :)

zwoop (approved):
Looks good, but probably should get a review from a Yahoo or Comcaster too :)

ywkaras (approved):

PR #10468, opened 2023-09-19
Set the proper variable when find_package fails to find the package

cmcfarlen:
I found another bug testing a setup without hwloc.


JosiahWI (approved):
Thanks for the fix. The addition of `break()` is unrelated to the fix and very likely has no performance improvement for us so I'm calling you out on it, but I think maybe it's what people would expect to see so I'm approving.

bneradt (approved):

PR #10471, opened 2023-09-19
stub.cc cleanup

lzx404243:
* Pruned down `cache/test/stub.cc` and `net/libinknet_stub.cc`. 
* Re-arranged sourced files within `src/api`


cmcfarlen (approved):
I think this is ok for now.  I feel like some of this will need to move, but we are going to have a meeting about layout that should straighten it out.

PR #10472, opened 2023-09-19
Refactor and rename restart metrics

zwoop:
Much of this is remnants from when we had traffic_manager. So this PR will remove a few metrics that either makes no sense, or have simply stopped being updated already. A few gets a new name, because we don't track "restarts" any more.


brbzull0 (approved):
Looks good!  Thanks

PR #10474, opened 2023-09-20
add conveinience function to lookup name->IntType*

cmcfarlen:


zwoop (approved):

include/api/Metrics.h
```diff
@@ -88,6 +89,16 @@ class Metrics
   IdType lookup(const std::string_view name) const;
   IntType *lookup(IdType id, std::string_view *name = nullptr) const;
 
+  std::optional<IntType *>
+  lookupPtr(const std::string_view name) const
```
bneradt, line 93: A const string_view strikes me as odd, but I see it is done elsewhere in the file.

bneradt (approved):

Issue #10428, opened 2023-09-14
CID 1508880: Structurally dead code


maskit:
rpc/RPCClient.h
```
 52  std::string
 53  invoke(std::string_view req)
 54  {
 55    std::string text; // for error messages.
 56    std::unique_ptr<char[]> buf(new char[BUFFER_SIZE]);
 57    swoc::FixedBufferWriter bw{buf.get(), BUFFER_SIZE};
 58    try {
 59      _client.connect();
 60      if (!_client.is_closed()) {
 61        _client.send(req);
 62        switch (_client.read_all(bw)) {
 63        case IPCSocketClient::ReadStatus::NO_ERROR: {
 64          _client.disconnect();
 65          return {bw.data(), bw.size()};
 66        }
 67        case IPCSocketClient::ReadStatus::BUFFER_FULL: {
 68          throw std::runtime_error(
 69            swoc::bwprint(text, "Buffer full, not enough space to read the response. Buffer size: {}", BUFFER_SIZE));
 70        } break;
 71        default:
 72          throw std::runtime_error("Something happened, we can't read the response");
 73          break;
 74        }
 75      } else {
 76        throw std::runtime_error(swoc::bwprint(text, "Node seems not available: {}", std ::strerror(errno)));
 77      }
 78    } catch (std::exception const &ex) {
 79      _client.disconnect();
 80      throw std::runtime_error(swoc::bwprint(text, "RPC Node Error: {}", ex.what()));
 81    }
 82
   	
CID 1508880 (#1 of 1): Structurally dead code (UNREACHABLE)
unreachable: This code cannot be reached: return std::__cxx11::string();.
 83    return {};
 84  }
```

PR #10475, opened 2023-09-20
CID-1508880: Re-work function to avoid dead-code.

brbzull0:
closes https://github.com/apache/trafficserver/issues/10428

### DIFF
--- a/include/shared/rpc/RPCClient.h
+++ b/include/shared/rpc/RPCClient.h
@@ -52,7 +52,7 @@ public:
   std::string
   invoke(std::string_view req)
   {
-    std::string text; // for error messages.
+    std::string err_text; // for error messages.
     std::unique_ptr<char[]> buf(new char[BUFFER_SIZE]);
     swoc::FixedBufferWriter bw{buf.get(), BUFFER_SIZE};
     try {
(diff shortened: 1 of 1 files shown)


include/shared/rpc/RPCClient.h
```diff
@@ -64,23 +64,22 @@ class RPCClient
...
     } catch (std::exception const &ex) {
-      _client.disconnect();
-      throw std::runtime_error(swoc::bwprint(text, "RPC Node Error: {}", ex.what()));
+      swoc::bwprint(err_text, "RPC Node Error: {}", ex.what());
     }
-
-    return {};
+    _client.disconnect();
+    // If we've got this far, then there must be an error to report back.
+    throw std::runtime_error(err_text);
```
brbzull0, line 82: not sure if coverity will complain about this as there is no return :P

moonchen:
The change looks correct.  I'm curious: do any compilers complain if all you do is remove that last return statement?

brbzull0:
> The change looks correct. I'm curious: do any compilers complain if all you do is remove that last return statem
more likely they will complain, I've seen it. But not sure if modern compilers do it.
In any case I think this looks better than before.

moonchen (approved):

PR #10484, opened 2023-09-21
Remove REC_ConfigReadCounter

### LINKED ISSUE #10414: CID 1508919: Unchecked return value
RecCore.cc
```
1141RecCounter
1142REC_ConfigReadCounter(const char *name)
1143{
1144  RecCounter t = 0;
    	
CID 1508919 (#1 of 1): Unchecked return value (CHECKED_RETURN)
1. check_return: Calling RecGetRecordCounter without checking return value (as is done elsewhere 4 out of 5 times).
1145  RecGetRecordCounter(name, &t);
1146  return t;
1147}
```

maskit:
This closes #10414.

According to a comments, the function is left for backward compatibility and it should be removed, and it's not used.


zwoop (approved):
This probably became unused after the Metrics refactoring / changes ? There is likely to be more of these going forward :).

PR #10488, opened 2023-09-21
Coverity 1521541: Uninitialized pointer field

bryancall:


zwoop (approved):
This really shouldn’t be needed but it’s also fine and harmless.

PR #10491, opened 2023-09-21
Coverity 1518592: check cachekey valid

### LINKED ISSUE #10345: CID 1518592: Uninitialized pointer field
cachekey.cc
```
257      if (TS_SUCCESS != TSHttpTxnPristineUrlGet(_txn, &_buf, &_url)) {
258        /* Failing here is unlikely. No action seems the only reasonable thing to do from within this plug-in */
259        CacheKeyError("failed to get pristine URI handle");
   	
CID 1518592 (#2 of 2): Uninitialized pointer field (UNINIT_CTOR)
7. uninit_member: Non-static class member _hdrs is not initialized in this constructor nor in any functions that it calls.
260        return;
261      }
262      CacheKeyDebug("using pristine uri '%s'", getUri(_buf, _url).c_str());
263    } else {
264      _buf = rri->requestBufp;
265      _url = rri->requestUrl;
266      CacheKeyDebug("using remap uri '%s'", getUri(_buf, _url).c_str());
267    }
```

serrislew:
Fixes #10345 by checking cachekey is valid

### DIFF
--- a/plugins/cachekey/cachekey.h
+++ b/plugins/cachekey/cachekey.h
@@ -68,6 +68,12 @@ public:
   bool appendUaClass(Classifier &classifier);
   bool finalize() const;
 
+  bool
+  isValid()
+  {
+    return _valid;
+  }
+
   // noncopyable
   CacheKey(const CacheKey &)            = delete; // disallow
   CacheKey &operator=(const CacheKey &) = delete; // disallow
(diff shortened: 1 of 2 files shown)


zwoop:
I'm generally fine with this, except, makes you wonder how we are even starting up with a configuration that could create broken (invalid) cache keys? Seems we ought to error out if so, rather than just marking it invalid ?

serrislew:
@zwoop that's true but this is called on `TSRemapDoRemap` when the transaction is called so this is after starting up, so would we want to error out (stop ATS) for this case? is that better than just skipping the config? 

I guess you mean by checking the configs before each reaching this point but it also seems like it can be invalid despite the configs [TSHttpTxnPristineUrlGet()](https://github.com/apache/trafficserver/blob/master/plugins/cachekey/cachekey.cc#L257-L260)

bneradt (approved):

PR #10505, opened 2023-09-22
Coverity 1497353: Use of 32-bit time_t

bryancall:
Found a couple bugs when fixing the 32bit time_t issue:
1. ink_atoui was returning a max value of INT_MAX instead of UINT_MAX.
2. mime_field_value_set_uint64 had too short of a buffer.


proxy/hdrs/MIME.h
```diff
@@ -1856,6 +1857,8 @@ MIMEHdr::set_age(time_t value)
     if (sizeof(time_t) > 4) {
       value_set_int64(MIME_FIELD_AGE, MIME_LEN_AGE, value);
     } else {
```
bryancall, line 1859: Maybe just get rid of the 32 check.  I don't think we support any OSes that have a 32bit time_t.

zwoop (approved):

PR #10515, opened 2023-09-25
Fix problem with debug and add support for Wasmtime for wasm plugin

shukitchan:


plugins/experimental/wasm/wasm_main.cc
```diff
@@ -62,23 +66,23 @@ transform_handler(TSCont contp, ats_wasm::TransformInfo *ti)
 
   ats_wasm::Context *c;
 
-  Dbg(dbg_ctl, "[%s] transform handler begins", __FUNCTION__);
+  Dbg(ats_wasm::dbg_ctl, "[%s] transform handler begins", __FUNCTION__);
```
serrislew, line 69: would it be more consistent (with the other plugins/definitions) to add `using namespace ats_wasm` instead of explicitly defining the namespace for each debug statement?

shukitchan: I probably can. But I was just trying to make it consistent inside this file (wasm_main.cc) and not using namespace.

plugins/experimental/wasm/wasm_main.cc
```diff
@@ -28,6 +28,10 @@
 #include "include/proxy-wasm/wasmedge.h"
 #endif
 
+#ifdef WASMTIME
+#include "include/proxy-wasm/wasmtime.h"
```
serrislew, line 32: will this have any conflict with WARM above? Or will WASMTIME not be defined if WAMR is?

shukitchan: Yup. That's controlled by Makefile.inc

serrislew (approved):

PR #10518, opened 2023-09-25
Move sandbox path into runtime directory

cmcfarlen:
In some autest setups the basic_plugin_handler fails due to not being able to write out its state file.  This PR updates the path to be in the traffic server `runtime` directory which seems to always be writable (for me, so far).

What say you @brbzull0?


brbzull0 (approved):
Makes sense yes.  We do create quite a few stuffs in the runtime dir, so should be fine I think.

Issue #10020, opened 2023-07-13
CID 1513224:  Resource leaks  (RESOURCE_LEAK)


zwoop:
```
*** CID 1513224:  Resource leaks  (RESOURCE_LEAK)
/src/traffic_server/[traffic_server.cc](http://traffic_server.cc/): 1006 in load_plugin(plugin_type_t, const swoc::_1_5_1::file::path &, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> &)()
1000     static bool
1001     load_plugin(plugin_type_t plugin_type, const fs::path &plugin_path, std::string &error)
1002     {
1003       switch (plugin_type) {
1004       case plugin_type_t::GLOBAL: {
1005         void *handle, *initptr;
   CID 1513224:  Resource leaks  (RESOURCE_LEAK)
   Variable "handle" going out of scope leaks the storage it points to.
1006         return plugin_dso_load(plugin_path.c_str(), handle, initptr, error);
1007       }
1008       case plugin_type_t::REMAP: {
1009         auto temporary_directory  = fs::temp_directory_path();
1010         temporary_directory      /= fs::path(std::string("verify_plugin_") + std::to_string(getpid()));
1011         std::error_code ec;
```

PR #10520, opened 2023-09-25
CID 1513224: Cleanup dl handle while testing a plugin

bneradt:
traffic_server has a mechanism to test whether a shared object file satisfies some basic symbol requirements of a plugin. This mechanism loaded that shared object file but never closed the resources. This addresses that by calling dlclose if the load is successful.

Fixes: #10020

### DIFF
--- a/proxy/Plugin.cc
+++ b/proxy/Plugin.cc
@@ -114,6 +114,8 @@ plugin_dso_load(const char *path, void *&handle, void *&init, std::string &error
   if (!init) {
     error.assign("unable to find TSPluginInit function in '").append(path).append("': ").append(dlerror());
     Error("%s", error.c_str());
+    dlclose(handle);
+    handle = nullptr;
     return false;
   }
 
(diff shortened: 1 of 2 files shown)


bryancall:
Here is another change you might want to include in the PR.   It closes the handle when we can't find the `TSPluginInit` symbol.

```
diff --git a/proxy/Plugin.cc b/proxy/Plugin.cc
index 99a7688cf..2f71dc44d 100644
--- a/proxy/Plugin.cc
+++ b/proxy/Plugin.cc
@@ -114,6 +114,8 @@ plugin_dso_load(const char *path, void *&handle, void *&init, std::string &error
   if (!init) {
     error.assign("unable to find TSPluginInit function in '").append(path).append("': ").append(dlerror());
     Error("%s", error.c_str());
+    dlcose(handle);
+    handle = nullptr;
     return false;
   }
```

masaori335 (dismissed):
Looks good.

bneradt:
> proxy/Plugin.cc

Done.

masaori335 (approved):

masaori335:
[approve ci ubuntu]

bneradt:
[approve ci ubuntu]

zwoop:
Cherry-picked to v9.2.x

PR #10521, opened 2023-09-25
CID 1518613: Traffic Dump: fix a use after free of a mutex

bneradt:
Coverity found that Traffic Dump had code that locked a mutex via std::lock_guard, freed the owner of the mutex (and thus the mutex), and then, when the std::lock_guard went out of scope, unlocked the free'd mutex. This use after free is addressed in this patch by extending the lifetime of the SessionData object owning the mutex until after the lock is released.

Fixes: #10331

---

## For Reviewers
This patch is easier to read if configured to ignore white space.


moonchen (approved):

zwoop:
@bneradt This gets merge conflicts because this plugin has moved on master. If you want this on 9.2.x, please make another branch PR.

Issue #10539, opened 2023-09-28
LSan: Unit Test Leaks with `--enable-malloc-allocator`


masaori335:
Combination of LSan & MallocAllocator uncovered existing memory leaks of unit tests.

```
==22727==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 14336 byte(s) in 7 object(s) allocated from:
    #0 0x55e02bcdac9d in aligned_alloc (/workspace/proxy/hdrs/.libs/test_proxy_hdrs+0xcbc9d) (BuildId: 7e145db0b261d0ac150935c6337987155a34438e)
    #1 0x55e02bd89cbc in MallocAllocator::alloc_void() /workspace/iocore/eventsystem/../../include/tscore/Allocator.h:141:13
    #2 0x55e02bd89cbc in thread_alloc(MallocAllocator&, ProxyAllocator&) /workspace/iocore/eventsystem/ProxyAllocator.cc:38:12
    #3 0x55e02bd710e5 in new_HdrHeap(int) /workspace/proxy/hdrs/HdrHeap.cc:114:35
    #4 0x55e02bd5546b in HTTPHdr::create(HTTPType, HTTPVersion, HdrHeap*) /workspace/proxy/hdrs/./HTTP.h:700:14
    #5 0x55e02bd5546b in (anonymous namespace)::test_http_hdr_null_char(int, char const*, char const*) /workspace/proxy/hdrs/unit_tests/test_Hdrs.cc:290:7
    #6 0x55e02bd5546b in C_A_T_C_H_T_E_S_T_4() /workspace/proxy/hdrs/unit_tests/test_Hdrs.cc:1191:16
    #7 0x55e02bcf4cfe in Catch::TestCase::invoke() const /workspace/lib/catch2/catch.hpp:14167:15
    #8 0x55e02bcf4cfe in Catch::RunContext::invokeActiveTestCase() /workspace/lib/catch2/catch.hpp:13027:27
    #9 0x55e02bcf4cfe in Catch::RunContext::runCurrentTest(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>&) /
workspace/lib/catch2/catch.hpp:13000:17
    #10 0x55e02bcf420e in Catch::RunContext::runTest(Catch::TestCase const&) /workspace/lib/catch2/catch.hpp:12761:13
[...]

Issue #10541, opened 2023-09-28
LSan: test_proxy_hdrs crash


masaori335:
As part of #10539, trying fix of memory leak in the `test_http_hdr_print_and_copy_aux`, I faced a crash.

The leak trying to fix

```
==26521==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 14336 byte(s) in 7 object(s) allocated from:
    #0 0x5591d1c59c9d in aligned_alloc (/workspace/proxy/hdrs/.libs/test_proxy_hdrs+0xcbc9d) (BuildId: 1e34a10e595cf9e4b7f649457b87d110220a4d1b)
    #1 0x5591d1d08ddc in MallocAllocator::alloc_void() /workspace/iocore/eventsystem/../../include/tscore/Allocator.h:141:13
    #2 0x5591d1d08ddc in thread_alloc(MallocAllocator&, ProxyAllocator&) /workspace/iocore/eventsystem/ProxyAllocator.cc:38:12
    #3 0x5591d1cf0205 in new_HdrHeap(int) /workspace/proxy/hdrs/HdrHeap.cc:114:35
    #4 0x5591d1cd3ffc in HTTPHdr::create(HTTPType, HTTPVersion, HdrHeap*) /workspace/proxy/hdrs/./HTTP.h:700:14
    #5 0x5591d1cd3ffc in (anonymous namespace)::test_http_hdr_print_and_copy_aux(int, char const*, char const*, char const*, char const*) /workspace/proxy/hdrs/unit_tests/test_Hdrs.cc:413:15
    #6 0x5591d1cd3ffc in C_A_T_C_H_T_E_S_T_4() /workspace/proxy/hdrs/unit_tests/test_Hdrs.cc:1191:20
    #7 0x5591d1c73cfe in Catch::TestCase::invoke() const /workspace/lib/catch2/catch.hpp:14167:15
    #8 0x5591d1c73cfe in Catch::RunContext::invokeActiveTestCase() /workspace/lib/catch2/catch.hpp:13027:27
    #9 0x5591d1c73cfe in Catch::RunContext::runCurrentTest(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>&) /workspace/lib/catch2/catch.hpp:13000:17
[...]

PR #10543, opened 2023-09-28
LSan: Fix memory leak of test_libhttp2

masaori335:
Part of #10539.

### DIFF
--- a/proxy/http2/unit_tests/test_HpackIndexingTable.cc
+++ b/proxy/http2/unit_tests/test_HpackIndexingTable.cc
@@ -71,7 +71,10 @@ TEST_CASE("HPACK low level APIs", "[hpack]")
       HpackIndexingTable indexing_table(4096);
 
       for (const auto &i : indexed_test_case) {
-        std::unique_ptr<HTTPHdr> headers(new HTTPHdr);
+        std::unique_ptr<HTTPHdr, void (*)(HTTPHdr *)> headers(new HTTPHdr, [](HTTPHdr *hdr) {
+          hdr->destroy();
+          delete hdr;
+        });
         headers->create(HTTP_TYPE_REQUEST);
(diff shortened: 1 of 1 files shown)


proxy/http2/unit_tests/test_HpackIndexingTable.cc
```diff
@@ -71,7 +71,10 @@ TEST_CASE("HPACK low level APIs", "[hpack]")
       HpackIndexingTable indexing_table(4096);
 
       for (const auto &i : indexed_test_case) {
-        std::unique_ptr<HTTPHdr> headers(new HTTPHdr);
+        std::unique_ptr<HTTPHdr, void (*)(HTTPHdr *)> headers(new HTTPHdr, [](HTTPHdr *hdr) {
+          hdr->destroy();
+          delete hdr;
```
masaori335, line 76: It looks like the destructor of `HTTPHdr` should call  `destroy()`, but it triggers a crash that the same with #10541.

masaori335:
[approve ci autest]

proxy/http2/unit_tests/test_HpackIndexingTable.cc
bneradt, line 76: This gets reused a few times in here. Can we:
1. Create a separate function for this.
2. Add a comment there explaining why this is destroy call is needed. (Basically `HTTPHdr` needs to be destroyed as well as deleted to be cleaned up properly).

bneradt (changes requested):
Makes sense.

proxy/http2/unit_tests/test_HpackIndexingTable.cc
masaori335: Done. Please take another look.

masaori335:
@bneradt The issue is addressed. Please take another look.

bneradt (approved):

zwoop:
Cherry-picked to v9.2.x

PR #10548, opened 2023-09-30
Build unit tests only with BUILD_TESTING

JosiahWI:
This makes all the tests conditional on BUILD_TESTING being set. Respecting BUILD_TESTING is a good idea because it is officially documented as part of CTest and people will expect it to work. Also, until now we have no way to not build the tests if we should want to.


cmcfarlen (changes requested):
I noticed looking at this we call `enable_testing` twice in CMakeLists.txt.  I think we should remove both calls.  The docs for `enable_testing` state:

```
This command is automatically invoked when the CTest module is included, except if the BUILD_TESTING option is turned off.
```

The difference seems to be if these calls are included with BUILD_TESTING off, a `test` target is created, but there are no tests.  Without the calls to `enable_testing`, there isn't even a test target.

JosiahWI:
Thanks, that's an excellent observation. This is definitely a good place to fix that.

cmcfarlen (approved):
Thanks!

PR #10554, opened 2023-10-02
cmake: Support external yaml-cpp

cmcfarlen:
This adds support for using external yaml-cpp library.

You can use this like so:
```
cmake . -B build -DEXTERNAL_YAML_CPP -Dyaml-cpp_ROOT=/opt/yaml
```

The caveat is we require 0.8.0 since that is what ATS expects and prior versions of yaml-cpp don't expose the same cmake targets.  Further, using newer versions is risky since yaml-cpp declares `AnyNewerVersion` for compatibility which could be just a giant lie.  We currently use the latest version, so this is a problem for the future.

As a side-effect, this PR also cleans up some link_libraries calls due to removing the global includes for libswoc and yaml-cpp.


lib/swoc/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ cmake_policy(SET CMP0087 NEW)
 # override "lib64" to be "lib" unless the user explicitly sets it.
 set(CMAKE_INSTALL_LIBDIR "lib" CACHE STRING "directory for libraries" FORCE)
 include(GNUInstallDirs)
+include(CMakeDependentOption)
```
cmcfarlen, line 10: This is already fixed in upstream libswoc so should be fine after updating

bneradt:
https://ci.trafficserver.apache.org/job/Github_Builds/job/cmake/1719/console

```
Total Test time (real) =  34.38 sec
+ /tmp/ats/bin/traffic_server -K -k -R 1
/tmp/ats/bin/traffic_server: error while loading shared libraries: libyaml-cppd.so.0.8: cannot open shared object file: No such file or directory
```

bneradt (approved):

PR #10558, opened 2023-10-04
Move mgmt/ and iocore/ files to include/ and src/

lzx404243:
As titled.


dragon512 (approved):
Big move. Everything look correct from what I can see for the move of source

zwoop:
I agree with everything Chris said. :-).

zwoop:
Also, with this landed, everyone better be ready and prepared to make branch PRs, because cherry-picking from master to 9.x is not going to be trivial anymore. :)

dragon512:
I don't have any disagreement with the change with P_ or I_ values. However this PR was about moving the source as that is a big move, then making change to the source files as PR after the move. I am concerned about trying to move and change at the same time with a lot of "smaller" PRs on in that the small PRs will be bigger than we want because of the "ball of Wax" issues that exists with the current headers/source files. @zwoop is concerned as I understand about making cherry picks, and the point of having all the source file move in one PR vs many was to make the possibility of moving source in an updated 9 easier than harder. At least that is the intent here.

I feel we need to get the main sources files moved then clean up the smaller items such as the which headers go where or renames of files. I don't want "the perfect to be the enemy of the good" here

cmcfarlen:
Maybe Just have a couple of PRs then?

1. Move the directories
2. Move the I_* headers but drop the I_ part

PR #10559, opened 2023-10-04
Refactors rate_limit plugin with YAML configurations

zwoop:
Besides majorly refactoring a lot of the global plugin code, this PR makes the following changes:

* YAML configuration
* Reloadable configurations (even as a global)
* SNI aliases (one rate limiter can apply to many SNIs)
* The IP reputation objects are now shareable for many SNIs. This means an attacker can circumvent the reputation by spreading the DDoS traffic across many SNIs.

### DIFF
--- a/doc/admin-guide/plugins/rate_limit.en.rst
+++ b/doc/admin-guide/plugins/rate_limit.en.rst
@@ -123,30 +123,33 @@ configuration file. The basic use is as::
 
 
 The YAML configuration can have the following format, where the varies sections
-and nodes are documented below::
-   selector:
-   - sni: test1.example.com
-     limit: 1000
-     queue:
-       size: 1000
-      max-age: 30
-     metrics:
-       tag: example.com
-       prefix: ddos
-     ip-rep: main
-   - sni: test2.example.com
-    limit: 100
+and nodes are documented below.
 
-   ip-rep:
-     - name: main
-       buckets: 10
-       size: 15
-       percentage: 90
-       max-age: 300
-       perma-block:
-         limit: 100
-         threshold: 1
-         max-age: 1800
+   .. code-block:: yaml
+
+      selector:
+         - sni: test1.example.com
+            limit: 1000
+            queue:
+               size: 1000
+               max-age: 30
+            metrics:
+               tag: example.com
+               prefix: ddos
+            ip-rep: main
+         - sni: test2.example.com
+            aliases: [test3.example.com, test4.example.com]
+            limit: 100
+      ip-rep:
+         - name: main
+            buckets: 10
+            size: 15
+            percentage: 90
+            max-age: 300
+            perma-block:
+               limit: 100
+               threshold: 1
+               max-age: 1800
 
 For the top level `selector` node, the following options are available:
 
@@ -157,6 +160,10 @@ For the top level `selector` node, the following options are available:
 .. option:: limit
 
    The maximum number of active client transactions.
+
+.. option:: aliases
+
+      A list of aliases for the SNI, which will also be matched by this rate limiter.
 
 .. option:: ip-rep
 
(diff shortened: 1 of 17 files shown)


shukitchan:
I can review this when it is ready

zwoop:
> I can review this when it is ready

Great! Almost there :). The refactoring focuses entirely on the global plugin only, leaving the remap plugin as-is. One hope with this refactoring and change to a reloadable YAML configuration is that we can add more features and rate limiting rules more easily.

shukitchan:
@zwoop let me know when it is ready to review. It looks like it is but I am not sure if you are having more changes in your mind.

bneradt:
[approve ci autest]

plugins/experimental/rate_limit/sni_selector.cc
```diff
@@ -16,117 +16,240 @@
...
-sni_queue_cont(TSCont cont, TSEvent event, void *edata)
+bool
+SniSelector::yamlParser(const std::string &yaml_file)
 {
-  SniSelector *selector = static_cast<SniSelector *>(TSContDataGet(cont));
+  YAML::Node config;
+
+  try {
+    config = YAML::LoadFile(yaml_file);
+  } catch (YAML::BadFile &e) {
```
brbzull0, line 35: You are not going to change the exception right? if not then why not `const`.
the one below too.

zwoop: I just copied this pattern from other places in the code :). I can change it.

plugins/experimental/rate_limit/sni_selector.cc
```diff
@@ -139,6 +262,26 @@ SniSelector::setupQueueCont()
...
+  TSReleaseAssert(sni_cont);
+
+  _instance.store(new SniSelector());
+  TSHttpHookAdd(TS_SSL_CLIENT_HELLO_HOOK, sni_cont);
+  TSHttpHookAdd(TS_VCONN_CLOSE_HOOK, sni_cont);
+
+  auto config_cont = TSContCreate(sni_config_cont, TSMutexCreate());
+
+  TSReleaseAssert(config_cont);
+  TSContScheduleEveryOnPool(config_cont, std::chrono::milliseconds{10000}.count(), TS_THREAD_POOL_TASK);
```
shukitchan, line 286: Is there a reason why we want to change the config file every 10 seconds vs using TSMgmtUpdateRegister() and wait for an explicit config reload 

just to make it more convenient?

Can a race condition happen when the config yaml is only partial updated when the config_cont is scheduled to run and therefore resulting in some inconsistent behavior?

plugins/experimental/rate_limit/sni_selector.cc
brbzull0: hope it wasn't my code! that would be a shame :rofl:

plugins/experimental/rate_limit/sni_selector.cc
zwoop: Yeh, for convenience, but we can change to be triggered with traffic_ctl.

shukitchan: I think it is better to be triggered by traffic_ctl

shukitchan (dismissed):
I think it looks ok.

zwoop:
[approve ci]

doc/admin-guide/plugins/rate_limit.en.rst
```diff
@@ -122,51 +121,117 @@ The basic use is as::
...
 
-.. option:: --tag
+   The queue option can include a `size` and a `max-age` option. The size is
+   default to ``UINT_MAX``, which is essentially unlimited. The max-age is
+   default to ``0``, which means no age limit.
 
-   An optional metric tag to use instead of the default. When a tag is not specified
-   the plugin will use the FQDN of the SNI associated with each rate limiter instance
-   created during plugin initialization.
+   No queue is enable without this configuration directive, but it can also be
```
shukitchan, line 196: enabled?

plugins/experimental/rate_limit/CMakeLists.txt
```diff
@@ -15,6 +15,8 @@
 #
 #######################
 
+project(rate_limit)
+
 add_atsplugin(
   rate_limit
   ip_reputation.cc
```
shukitchan, line 22: lists.cc is not needed herer?

shukitchan (changes requested):

plugins/experimental/rate_limit/CMakeLists.txt
zwoop: Not sure why it didn't fail without lists.cc ... Added.

zwoop:
I updated this to use the updated version of the registration API.

doc/admin-guide/plugins/rate_limit.en.rst
zwoop: fixed

PR #10577, opened 2023-10-09
Move DbgCtl.cc to libtsapi so plugins can link against it.

ywkaras:


bneradt:
[approve ci autest]

ywkaras:
[approve ci rocky]

ywkaras:
[approve ci cmake]

cmcfarlen:
This change is not necessary and adds a ton of complexity to testing and maintenance.  We are trying to reduce the amount of stub code required for tests, not add more.  I agree that we need to move some of the lowest level functionality to a different library to increase testability and for plugins to use outside of tscore, but we need to finish making a plan for this.

ywkaras:
It's been a settled matter for some time that definitions in the TS API would be available in libtsapi.so.  It's a big issue for Yahoo if we reverse this, even if only temporarily.  Can we merge this, and replace it with the golden solution later, without that being a blocker for other work?

dragon512:
I don't disagree with you @cmcfarlen . However I think @ywkaras is only trying to address the regression in our commit to correctly link our code correctly.

The issues here is simple to fix.. we need to have a .so for plugin only code and for code that is to be shared in core and plugins. However people seem to want to say "well we can sort of not do it and it works... and we been doing it this way for a long time and I have not seen an issue" as an argument.

We cannot make code better and have the build chain work for us is we are hung up on making the basic stuff hard.. @ywkaras is only doing what is needed to make this work, given the lets make this hard argument.

cmcfarlen:
DbgCtl needs to be available to tests as well without linking every other library in the project or stubbing half of it.  tsapicore is a good place for code like DbgCtl (Metrics is there, for example, another bit that plugins can use), but currently there is code in there (FetchSM) that needs to moved to the higher level tsapi to avoid extra dependencies.  We need to clean that up so tests and plugins can use tsapicore directly for those exposed features.

Further, if tsapi doesn't declare its dependencies property then it places a burden on any target that uses it.  This is still a conversation we need to have.  I'd like to better understand the situation you are up against, so we can all take that into account.

At the last ATS weekly meeting we agreed to have a separate meeting to discuss tsapi.  I think we should do that soon.

ywkaras:
As I understand it, libtsapicore is for definitions that are needed by both libtsapi and core, but are not part of the TS API.  DbgCtl.cc has definitions which are (now) part of the TS API.

ywkaras:
If stubs are clearly a bad thing in unit tests, why are tools like gMock widely used?

cmcfarlen:
> If stubs are clearly a bad thing in unit tests, why are tools like gMock widely used?

Mocking is about changing behavior of a code dependency in order to facilitate testing other aspects of a system.  This PR removes functionality from tests that are very useful for creating and troubleshooting tests.  This is nowhere near the same thing.  IMO stubs(and mocking) are bad if they are incurred incidental complexity to writing a test.  Not to mention the code duplication in this PR.

cmcfarlen:
> As I understand it, libtsapicore is for definitions that are needed by both libtsapi and core, but are not part of the TS API. DbgCtl.cc has definitions which are (now) part of the TS API.

This is definitely part of the conversation we need to have.  But even as a stop-gap, it is preferable to this approach.

ywkaras:
https://stackoverflow.com/questions/2665812/what-is-mocking

Empty stubs for console output functions make sense when the console output and function return values are not pertinent to the test.  The stub duplication could be eliminated.

dragon512:
can we just fix the build to have a way to build the ts core infra for plugins to be built as an .so and an .a ( so people like me can build this with two .so files for plugins to link with). Apple can do what ever they want to have a full .a build. The basic design of an software is that if you have extensions.. the code shared with the extension needs to be a .so . This is simple .. everything else is complex. 

I agree with @cmcfarlen about the testing concerns, however if we want this flat .a build we have to make stuff complex or make the build weak to assume it is correct, which devalues the tools we are using and allows for human mistakes to creep in when we could have automated that out 100%. 100% correct is better than "might be correct"

cmcfarlen:
#10585

Here is an alternative moving DbgCtl to tsapicore.  Most tests already link to this already so it's a lighter move and still gets it out of core.  I haven't updated auto tools files for this yet, so only the CI jobs that use cmake will pass for now.

ywkaras:
> #10585
> 
> Here is an alternative moving DbgCtl to tsapicore. Most tests already link to this already so it's a lighter move and still gets it out of core. I haven't updated auto tools files for this yet, so only the CI jobs that use cmake will pass for now.

Jason would have to change his scons/parts files to link against tsapicore as well as tsapi.  That will expose definitions in tsapicore that are not part of the TS API to plugins.  If that's OK with him, it's fine with me.

dragon512:
I am fine with having two .so to link with for plugins.. I sort of expected this would be needed. This will make life a lot easier in the end.

ywkaras:
> I am fine with having two .so to link with for plugins.. I sort of expected this would be needed. This will make life a lot easier in the end.

How about the option of just putting everything in tscore.so and linking plugins with that?

PR #10578, opened 2023-10-10
Cleanup for optimised builds

cmcfarlen:
LTO builds were having linking problems with clang 16 due to lack of PIC.  This marks static libs PIC and cleans up some link library calls.


JosiahWI (approved):
Thanks, this looks great! The way the esi plugin is laid out is kind of messy and it would be good to improve it in a separate PR.

PR #10581, opened 2023-10-10
lua plugin: update CMakeLists.txt for ts_lua_vconn.cc

bneradt:
Updating plugins/lua/CMakeLists.txt to build ts_lua_vconn.cc. Otherwise the plugin fails to find the symbols defined in there when the plugin is run.


randall (approved):

shukitchan:
👍

PR #10585, opened 2023-10-11
Move DbgCtl.cc from tscore to tsapicore

cmcfarlen:


ywkaras:
I think, for this to work, we would also have to turn tsapicore into a dynamically linked lib.

dragon512 (approved):

dragon512 (commented):
For the cmake part of this

cmcfarlen:
> I think, for this to work, we would also have to turn tsapicore into a dynamically linked lib.

I tried this but there is a dependency cycle, so we will have to move some items around.  We need to do this anyway.

iocore/net/Makefile.am
```diff
@@ -53,6 +53,7 @@ test_certlookup_SOURCES = \
 test_certlookup_LDADD = \
 	$(top_builddir)/iocore/eventsystem/libinkevent.a \
 	$(top_builddir)/src/tscore/libtscore.a \
+	$(top_builddir)/src/api/libtsapicore.a \
```
ywkaras, line 56: I'm pretty sure tsapicore.a need to become dynamically loaded and installed.  @dragon512 can confirm.

bneradt (approved):
Looks great. Thanks for making this happen.

PR #10586, opened 2023-10-11
Enforce h2 MAX_* settings before SETTINGS ACK

bneradt:
Per the RFC, MAX_HEADER_SIZE and MAX_CONCURRENT_STREAMS are infinite by default. ATS usually doesn't enforce configured h2 SETTINGS until the client ACKs those changes. This patch updates ATS to enforce the configured MAX_* settings even before the client ACKS the settings update so that a poorly behaving client can't run ATS out of resources by opening up an unreasonable amount of streams.


bneradt:
Note that this is a 10.x only issue since the `acknowledged_local_settings` implementation is new to 10.x.

maskit (approved):
Makes sense.

PR #10587, opened 2023-10-12
Break cycle between iocore/inknet and proxy/http

JosiahWI:
Summary of changes:

- Remove unused headers.
- Break the PreWarm namespace out of PreWarmManager and leave it in iocore/net.
- Move PreWarmManager back to proxy/http (I had moved it to iocore earlier).
- Use a callback to reconfigure the prewarm manager when sni.yaml is reloaded.

Removing the unused headers is simply cleanup. They have been in the source code for a long time and aren't currently used for anything. The other two changes were more significant. I think there's a tradeoff between putting the PreWarm namespace in iocore/net or proxy/http. The former makes it easy to avoid the dependency cycle, but I think the code is more closely related to the rest of the PreWarmManager than to iocore/net. The last change was to address the issue of reloading the pre-warm configuration when the SNI was reloaded, because SNI rules could change pre-warm config. The codebase currently handled this by calling `reconfigure()` on the global `prewarmManager`, so I proposed to @masaori335 that the prewarmManager could instead register its `reconfigure()` method as a callback. Because that method is static, I think this ended up being kind of messy.

### DIFF
--- a/iocore/net/PreWarm.h
+++ b/iocore/net/PreWarm.h
@@ -1,0 +1,132 @@
+/** @file
+
+  Pre-Warming NetVConnection
+
+  @section license License
+
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+
+#pragma once
+
+// inknet
+#include "SSLTypes.h"
+#include "YamlSNIConfig.h"
+
+// api
+#include "api/Metrics.h"
+
+// records
+#include "records/I_RecHttp.h"
+
+// tscore
+#include "tscore/CryptoHash.h"
+#include "tscore/ink_hrtime.h"
+
+#include <netinet/in.h>
+
+#include <array>
+#include <cstddef>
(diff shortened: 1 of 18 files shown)


JosiahWI:
There's no AuTest verifying that the reconfiguration takes place, and I don't see a way to verify it. So unless someone can propose a good way to test, I'm going to leave it untested.

proxy/http/PreWarmManager.h
```diff
@@ -312,6 +216,16 @@ class PreWarmQueue : public Continuation
...
 {
 public:
+  PreWarmManager()
+  {
+    // We use the callback because it would introduce a circular dependency
+    // for SNIConfig to explicitly call prewarmManager.reconfigure().
+    // Because prewarmManager is global there's not a good place to set
+    // this, but as long as there's only one instance of this class
+    // there should not be an issue.
+    SNIConfig::set_on_reconfigure_callback([this]() { this->reconfigure(); });
```
cmcfarlen, line 226: 🥇

iocore/net/PreWarm.h
```diff
@@ -0,0 +1,132 @@
...
+struct Dst {
+  Dst(std::string_view h, in_port_t p, SNIRoutingType t, int a) : host(h), port(p), type(t), alpn_index(a) {}
+
+  std::string host;
+  in_port_t port      = 0;
+  SNIRoutingType type = SNIRoutingType::NONE;
+  int alpn_index      = SessionProtocolNameRegistry::INVALID;
+};
+
+using SPtrConstDst = std::shared_ptr<const Dst>;
```
masaori335, line 61: TLSTunnelSupport needs only `PreWarm::Dst` and `PreWarm::SPtrConstDst`. We can leave rest of PreWarm namespace code in the `proxy/http/PreWarmManager.h`. That'll clarify which structure is required by `iocore/net` and `proxy/http`. What do you think?

JosiahWI: Great idea!

proxy/http/PreWarmManager.h
masaori335: Nice idea to get rid of the cyclic dependency 👍

masaori335 (approved):
Looks good! Thanks for breaking the cyclic dependency.

PR #10588, opened 2023-10-12
Honor BUILD_SHARED_LIBS for tscore.

cmcfarlen:
This change will build `tscore` as a shared library if `BUILD_SHARED_LIBS` is set to `ON`.  By default this will be built statically.  We can add more libraries to honor this settings, but libraries cannot be shared if they have circular deps(pretty much every other library) so we'll have to work through that if we want to expand this.

### DIFF
--- a/plugins/esi/common/CMakeLists.txt
+++ b/plugins/esi/common/CMakeLists.txt
@@ -15,7 +15,7 @@
 #
 #######################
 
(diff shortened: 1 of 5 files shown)


JosiahWI (approved):
It looks like this PR also makes the esi libs explicitly static, which is consistent with the rest of our libs.

dragon512 (approved):

PR #10589, opened 2023-10-12
Use returned length instead of calculating again with strlen

duke8253:
Remove redundant string length calculations


maskit:
Looks like `LogAccess::strlen` doesn't simply count the length. It adds 1 byte for null termination and also consider about alignment. Is it actually unnecessary?

bryancall (approved):

duke8253:
> Looks like `LogAccess::strlen` doesn't simply count the length. It adds 1 byte for null termination and also consider about alignment. Is it actually unnecessary?

Good point. Though we're seeing some very weird error values in the logs, so I think we can leave this PR for now and wait for a few days to see whether the we still see those values in the logs.

duke8253:
Closing this for now incase it gets merged by accident.

PR #10591, opened 2023-10-12
Fix typo in block_errors documentation

maskit:
https://github.com/apache/trafficserver/blob/6c03b168ae244db87c2bb32090671301d378ea08/plugins/experimental/block_errors/block_errors.cc#L48


bneradt (approved):
Looks fine to me. Unless we want to switch things the other way (i.e., we should use `enable` instead of `enabled`). But I think that `enabled` is more typical. So I say go forward with your patch as is.

bryancall (approved):

bryancall:
@maskit We should have a PR for 8.1.x separately so the change can be part of the changelog and tied to 8.1.10.  I cherry-picked this change into 9.2.x for the next 9.2.4 release.

maskit:
https://github.com/apache/trafficserver/pull/10618

PR #10595, opened 2023-10-12
Break cycle between iocore/hostdb and proxy/http libraries

cmcfarlen:
This moves some http related config from hosted to http.


cmcfarlen:
[approve ci cmake]

bneradt (approved):

PR #10596, opened 2023-10-12
Add QUICSupport as a NetVC service

maskit:
Trying to catch the trend. This removes `#include <P_QUICNetVConnection.h>` from `proxy/`, and use more abstract `QUICSupport.h` instead.


brbzull0 (approved):
Nice!!
Thanks!

PR #10597, opened 2023-10-12
Break inkdns->inkcache and inkhostdb->inkcache cycles

JosiahWI:
This removes unused inkcache includes from the inkdns and inkhostdb components, breaking the dependency on inkcache for both.


proxy/http/HttpTransactCache.cc
```diff
@@ -29,6 +29,10 @@
 #include <ctime>
 #include "HTTP.h"
 #include "HttpCompat.h"
+
+// inkcache
+#include "P_Cache.h"
```
cmcfarlen, line 34: Can we get away with including `I_Cache.h` here?

JosiahWI: `I_Cache.h` doesn't bring in the symbols it needs.

JosiahWI: It was previously getting the symbols by a transient include of `P_Cache.h` in a header in `include/ts/InkAPIPrivateIOCore.h`.

JosiahWI: Otherwise I could include the specific headers it needs. `P_Cache.h` doesn't contain any declarations - it just brings in all the other Cache headers.

cmcfarlen: We should stick to including the `I_` headers for includes outside of the library if we can.

cmcfarlen (approved):

bneradt (approved):

PR #10598, opened 2023-10-12
Add library dirs to luajit target

cmcfarlen:


zwoop (approved):

cmake/FindLuaJIT.cmake
```diff
@@ -44,5 +44,6 @@ endif()
 if(LuaJIT_FOUND AND NOT TARGET LuaJIT::LuaJIT)
     add_library(LuaJIT::LuaJIT INTERFACE IMPORTED)
     target_include_directories(LuaJIT::LuaJIT INTERFACE ${LuaJIT_INCLUDE_DIRS})
+    target_link_directories(LuaJIT::LuaJIT INTERFACE ${LuaJIT_LIBRARY_DIRS})
     target_link_libraries(LuaJIT::LuaJIT INTERFACE ${LuaJIT_LIBRARIES})
```
JosiahWI, line 48: Is this line still needed?

cmcfarlen: `LuaJIT_LIBRARIES` just has the `luajit-5.1`, but if that isn't in a known path then it fails to find it.  So I think it needs both.

cmcfarlen: From CMakeCache.txt

```
LuaJIT_LIBRARY_DIRS:INTERNAL=/opt/homebrew/Cellar/luajit/2.1.0-beta3-20230430.1/lib
LuaJIT_LIBRARIES:INTERNAL=luajit-5.1
```

JosiahWI: Interesting. Thanks for the fix.

JosiahWI (approved):

bneradt (approved):

PR #10599, opened 2023-10-12
Split CacheVC into its own source file

JosiahWI:
This is a low-priority cleanup, but it should be easy to review because I only moved code between files and didn't make any source changes, besides taking the time to carefully list out all the includes for the new source file.


masaori335 (approved):
Nice cleanup! Somehow (probably due to dependency issue), many implementation of functions are spreaded across files.

masaori335:
We can move these later, but please note that some CacheVC function implementations are still in the CacheRead.cc and CacheWrite.cc.

https://github.com/apache/trafficserver/blob/2eb6377e8e59ae862c4e0f29aabc139b6f8b1acc/iocore/cache/CacheRead.cc#L186

https://github.com/apache/trafficserver/blob/2eb6377e8e59ae862c4e0f29aabc139b6f8b1acc/iocore/cache/CacheWrite.cc#L76

JosiahWI:
Good catch. I'll add that to my TODO.

masaori335:
Also in the CacheVol.cc. Thanks for taking care those.
https://github.com/apache/trafficserver/blob/2eb6377e8e59ae862c4e0f29aabc139b6f8b1acc/iocore/cache/CacheVol.cc#L62

PR #10601, opened 2023-10-13
Move CacheVC struct and methods from Vol

JosiahWI:
This is part of the follow-up requested by @masaori335 in #10599. The remaining work is to deal with the methods defined in CacheRead.cc and CacheWrite.cc.


iocore/cache/CacheVC.cc
```diff
@@ -21,12 +21,17 @@
   limitations under the License.
  */
 
+#include "I_Cache.h"
 #include "I_CacheDefs.h"
 #include "P_CacheDisk.h"
 #include "P_CacheHttp.h"
+// #include "P_CacheHosting.h"
```
cmcfarlen, line 28: delete this line.

iocore/cache/CacheVC.cc
```diff
@@ -21,12 +21,17 @@
...
 #include "P_CacheInternal.h"
 #include "P_CacheVol.h"
 
+// must be included after the others
```
cmcfarlen, line 32: Should CacheVC.h include the required files itself?

cmcfarlen (changes requested):

iocore/cache/CacheVC.cc
JosiahWI: I couldn't get that to work. I spent significantly more time trying to get that to work than I did making the change.

cmcfarlen (approved):

masaori335:
Thanks for addressing the issue.

For CacheRead.cc and CacheWrite.cc, we need to figure out what we want. Current code layout makes it clear which functions are related to reading or writing cache.

JosiahWI:
Agreed. I think we would be losing something by doing away with that structure, but maybe we can come up with a better way to organize it.

PR #10602, opened 2023-10-13
sslhdr library needs public ssl

cmcfarlen:
This fixes some cases where the test for the sslheaders plugin can't find openssl headers.

### DIFF
--- a/plugins/experimental/sslheaders/CMakeLists.txt
+++ b/plugins/experimental/sslheaders/CMakeLists.txt
@@ -16,7 +16,7 @@
 #######################
 
 add_library(sslhdr STATIC expand.cc util.cc)
-target_link_libraries(sslhdr PRIVATE OpenSSL::SSL)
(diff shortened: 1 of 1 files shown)


zwoop (approved):
Love it!

plugins/experimental/sslheaders/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@
 #######################
 
 add_library(sslhdr STATIC expand.cc util.cc)
-target_link_libraries(sslhdr PRIVATE OpenSSL::SSL)
+target_link_libraries(sslhdr PUBLIC OpenSSL::SSL)
```
JosiahWI, line 19: I think we should keep this `PRIVATE` and link `test_sslhdr` to `OpenSSL::SSL` instead. The missing headers are included directly by the test; it's not a transitive dependency.

JosiahWI (changes requested):

bneradt (approved):

PR #10603, opened 2023-10-13
Extract CacheEvacuateDocVC from CacheVC

JosiahWI:
This extracts a subclass from `CacheVC` called `CacheEvacuateDocVC` and moves the logic for evacuating a doc to the new subclass. By having less code in it, `CacheVC` should be a little simpler. While doing the extraction, I also read through the logic and tried to clean up the logic flow a little bit. This last bit was risky, and I'm not confident we have enough test coverage to be sure it was done correctly. I did my best to double check my reasoning on every change, but please review carefully.

EDIT: The following changes are no longer part of this PR per @masaori335's request.

~~The two dangerous refactorings I did are split into two commits.~~ They were supposed to be split into two commits, but it looks like that didn't happen. The first one should be the simplest and safest. It just extracts a method which returns a status boolean, and remembers to check that boolean to abort on failure. The second refactoring was more involved. I will copy the commit message, which explains the refactoring, for convenience.

> This extracts some methods and attempts to make the logic flow more
self-evident. Most importantly, it makes it obvious that the method
locates a particular evacuation block to work with, and terminates after
it finishes. The previous logic flow obscured this by putting all the
operations on the block within the search loop body.

### DIFF
--- a/iocore/cache/CacheEvacuateDocVC.h
+++ b/iocore/cache/CacheEvacuateDocVC.h
@@ -1,0 +1,72 @@
+/** @file
+
+  A brief file description
+
+  @section license License
+
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+
+#pragma once
+
+/*
+#include "P_CacheDir.h"
+*/
+
+#include "CacheVC.h"
+
+// eventsystem
+#include "I_Continuation.h"
+#include "I_EThread.h"
+#include "I_Event.h"
+#include "I_ProxyAllocator.h"
+
+// tscore
+#include "tscore/ink_assert.h"
+
+// ts
+#include "ts/DbgCtl.h"
+
+class CacheEvacuateDocVC : public CacheVC
+{
+public:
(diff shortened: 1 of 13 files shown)


masaori335:
> While doing the extraction, I also read through the logic and tried to clean up the logic flow a little bit.

I'd prefer decouple moving code and logic change even if small simple changes. From our experience, it easily goes wrong way. Also when we got trouble, it makes debugging hard.

JosiahWI:
That's a good point. Would it be good for me to split the commits into 2 PRs?

iocore/cache/CacheEvacuateDocVC.h
```diff
@@ -0,0 +1,72 @@
...
+#include "I_Event.h"
+#include "I_ProxyAllocator.h"
+
+// tscore
+#include "tscore/ink_assert.h"
+
+// ts
+#include "ts/DbgCtl.h"
+
+class CacheEvacuateDocVC : public CacheVC
```
masaori335, line 44: Any member variables can be move into this class from `CacheVC`?

masaori335:
It's better to change `int Vol::evacuateWrite(CacheVC *evacuator, int event, Event *e)` to `int Vol::evacuateWrite(CacheEvacuateDocVC *evacuator, int event, Event *e)` to make sure what Continuation is given.

JosiahWI:
Excellent observation, thanks!

iocore/cache/CacheEvacuateDocVC.h
JosiahWI: None stood out, but I should do a more thorough check.

JosiahWI: @masaori335 There's a flag that indicates whether the CacheVC is an evacuator or not. In `agg_copy` evacuated docs have to have different handling, so the flag is used to determine which algorithm to use. That is the only place `overwrite_dir` is set on a `CacheVC` instead of `CacheEvacuateDocVC`. If we move `agg_copy` into `CacheVC`, we can use dynamic dispatch to get the correct copy algorithm, move `overwrite_dir` to the `CacheEvacuateDocVC`, and completely do away with the `f.evacuator` flag. But that's quite a lot of changes so maybe we should do it in a follow-up PR.

JosiahWI:
Rebased because of CMakeLists.txt conflict.

iocore/cache/CacheEvacuateDocVC.h
masaori335: Thanks for checking, let's leave it as is for now.

masaori335 (approved):
Looks good.

PR #10610, opened 2023-10-13
Move ControlMatcher.cc into .inl file

JosiahWI:
The ControlMatcher template had to be instantiated for all its dependent components because the definitions were in a source file. This created a dependency on inkdns. After this change, there is only one dependency on inkdns remaining in proxy, so this is a step towards simplifying the dependency tree. The change was made in such a way as to retain git history in ControlMatcher.inl.


JosiahWI:
[approve ci ubuntu fedora debian centos]

bneradt:
[approve ci debian]

JosiahWI:
@SolidWallOfCode and I took a look at this and decided a better approach is to change the SplitDNS configuration to YAML.

PR #10612, opened 2023-10-14
Change CMAKE_CXX_STANDARD to cache variable

### LINKED ISSUE #10609: cmake: support setting CMAKE_CXX_STANDARD on the commandline
The way `CMAKE_CXX_STANDARD` is currently used in CMakeLists.txt, a user cannot override it on the command line invocation of cmake. It would be nice to update things so that users can easily test `-std=c++20` builds if they like.

Currently CI has a couple builds that run with `-std=c++20` just to ensure we don't break things for C++20. We don't currently support using C++20 features, but we want to make sure we don't use deprecated features that are removed in 20 and thus break the 20 build.

JosiahWI:
There are multiple ways to approach this. ~~Since the ability to build with C++20 is intended for CI, I opted to add an option that can be preset. An alternative that I tested was to simply respect CMAKE_CXX_STANDARD if it was set, but I don't like this, because it doesn't work on most projects and consistency is good. I also think the option makes the purpose explicit, whereas allowing CMAKE_CXX_STANDARD to be set from the command line is more flexible, but more mysterious.~~

@cmcfarlen  pointed out that it is extra work to maintain a new option. I considered this and have to admit respecting `CMAKE_CXX_STANDARD` is much simpler and more flexible, so I have converted `CMAKE_CXX_STANDARD` to a cache variable.

Fixes #10609


bneradt:
[approve ci]

bneradt:
[approve ci]

bneradt:
[approve ci]

JosiahWI:
[approve ci rocky]

bneradt:
[approve ci debian freebsd osx rocky]

bneradt:
[approve ci rocky]

bneradt (approved):
Looks good to me.

@cmcfarlen : I'll let you review this before merging. If we do approve and merge this, let me know and I'll update our CI to use this instead of the ugly sed stuff I did there temporarily.

bryancall (approved):

cmcfarlen:
IMO just setting CMAKE_CXX_STANDARD directly is less mysterious than a custom setting for the same.  Also, we will have to maintain this one when we want to build with C++23 or whatever other wacky standard version code name compiler vendors come up with (C++2d perhaps?).  Its also more inline with cmake's idiom of not setting compiler flags directly but letting that be done from outside at configuration time.

JosiahWI:
This is a good point. I can change it back.

bneradt:
[approve ci freebsd]

bneradt:
[approve ci freebsd]

bneradt:
[approve ci freebsd]

cmcfarlen (approved):

PR #10614, opened 2023-10-16
Add rat target

cmcfarlen:
This adds the `rat` target that behaves the same as the autotools version.  CI will still need to post-process the output to determine if there are violations.


bryancall (approved):
Tested the commit locally.

bneradt:
Thanks @cmcfarlen .

PR #10633, opened 2023-10-17
fix case on cache var

cmcfarlen:
This was causing a cmake warning


JosiahWI (approved):
Java habit bit me when I added this. Thanks for fixing the case.

randall (approved):

PR #10634, opened 2023-10-18
Fetch dependencies using cmake instead of internal copies

cmcfarlen:
This will download and build yaml-cpp and libswoc from github instead of using the internal copies.  For  yaml-cpp, it will attempt to find version 0.8.0 or newer using find_package before fetching it.  For libswoc it always fetches.

This is a draft PR so we can look at this, try it out and see if we like this approach for ATS.

I didn't do a lot of testing with this, just verified it downloads and builds on my mac and that it installed yaml-cpp and libswoc correctly.


bryancall:
[approve ci]

cmcfarlen:
[approve ci]

PR #10642, opened 2023-10-18
Improve the TSMgmtUpdateRegister with an optional file name

zwoop:
Providing this filename makes the update registry keep track of file changes automatically. When used, the continuation provided will only be called if changes to the configuration file is detected.

Also, this eliminates the superflous code for calling a named plugin reload. That code is never used, and I don't like having code there "just because", when nothing uses it.

### DIFF
--- a/doc/developer-guide/api/functions/TSMgmtUpdateRegister.en.rst
+++ b/doc/developer-guide/api/functions/TSMgmtUpdateRegister.en.rst
@@ -28,7 +28,7 @@ Synopsis
 
     #include <ts/ts.h>
 
-.. function:: void TSMgmtUpdateRegister(TSCont contp, const char * plugin_name)
+.. function:: void TSMgmtUpdateRegister(TSCont contp, const char *plugin_name, const char *plugin_file_name=nullptr)
 
 Description
 ===========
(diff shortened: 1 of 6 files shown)


zwoop:
[approve ci].

shukitchan (changes requested):
Let's update documentation as well - https://docs.trafficserver.apache.org/en/latest/developer-guide/api/functions/TSMgmtUpdateRegister.en.html

zwoop:
[approve ci]

zwoop:
[approve ci autest]

shukitchan:
[approve ci autest]

shukitchan:
1) Since it is an API change, should we send an email to the mailing list of the change?

2) The "name" is really useless. So it does not have to be a map. A list will do. Right?. I am ok with the current change unless you want to improve it further.

zwoop:
> 1. Since it is an API change, should we send an email to the mailing list of the change?

Yeh, I can email, but it doesn't break or modify any existing plugins (because we can now provide defaults in API changes).

> 2. The "name" is really useless. So it does not have to be a map. A list will do. Right?. I am ok with the current change unless you want to improve it further.

With # 2, yes, it's useless, but removing it would break compatibility unnecessarily. I'd rather keep it, if not only because it could at least allow for easier debugging later.

shukitchan (approved):

brbzull0 (approved):
looks good!
Thanks.

PR #10645, opened 2023-10-19
Rename cache/test/ to cache/unit_tests/

masaori335:
For consistency with other subsystems.


masaori335:
I added a fix for CacheDir test to pass CI.

bneradt (approved):
@cmcfarlen : I wonder whether the lock change here would address that lock issue we saw with the cache unit tests in Ubuntu earlier.

PR #10650, opened 2023-10-19
cmake: Install include files

cmcfarlen:
This installs the include files based on what is installed from autotools.

I moved the install target for libswoc into libswoc's CMakeLists.txt.  That file has deviated from libswoc upstream substantially, so I also ran cmake-format on it.

There doesn't seem to be consensus on the best way to setup header files for installation, so I just picked one that wasn't overly complicated.  The bare minimum is to set a PUBLIC_HEADER property on the target and then specify a destination for the same on the install target.  There is a target_include_directories component, but it has more to do with exporting cmake scripts for other projects to consume.  That task will be a different PR so that aspect will likely need adjustment.

Edit: I added `tscpp/api` as well

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -38,4 +38,4 @@ set(SWOC_INCLUDE_DIR
 add_library(catch2::catch2 INTERFACE IMPORTED GLOBAL)
 target_include_directories(catch2::catch2 INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/catch2")
 
-install(TARGETS libswoc fastlz)
+install(TARGETS fastlz)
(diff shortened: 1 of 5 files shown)


cmcfarlen:
[approve ci ubuntu]

bneradt:
[approve ci fedora autest]

bneradt:
[approve ci autest]

bneradt:
[approve ci autest]

JosiahWI (changes requested):
Could we use `PROJECT_SOURCE_DIR` instead of `CMAKE_SOURCE_DIR`? I don't think it's all that likely that ATS won't be the top-level project, but I'm kind of paranoid about it.

bneradt (approved):

PR #10652, opened 2023-10-20
Convert CacheDisk and RamCache regression tests into unit tests

masaori335:


src/iocore/cache/unit_tests/test_CacheDisk.cc
```diff
@@ -0,0 +1,98 @@
...
+        ++from;
+      } else {
+        ++to;
+      }
+    }
+  }
+  Dbg(dbg_ctl_cache_test,
+      "Cache stability difference - "
+      "delta = %d of %d : %d to, %d from, originally %d slots, now %d slots (net gain = %d/%d)",
+      to + from, VOL_HASH_TABLE_SIZE, to, from, then, now, now - then, to - from);
```
masaori335, line 94: @SolidWallOfCode Originally, this test is introduced by TS-2505. f8cb33617947d6ca27208a57cff7758fc7529391

How the stability was checked with these values?

PR #10655, opened 2023-10-21
List all includes explicitly in dns headers

JosiahWI:
This follows the 'include what you use' principle for all the headers in dns. The source files remain unchanged - I did the headers first because they get included by other components, and I ran into missing declarations very quickly when I started to write unit tests. ~~All the dns headers now use include paths relative to include/ as @zwoop has championed, with the exception of P_ headers.~~ (done across all of ATS by @lzx404243 🎉)

Note: inkdns has a public dependency on a private header in inknet, but we can't link that directly because the cyclic dependency causes linking problems with RegressionSM.

### DIFF
--- a/src/iocore/dns/CMakeLists.txt
+++ b/src/iocore/dns/CMakeLists.txt
@@ -21,7 +21,7 @@ add_library(ts::inkdns ALIAS inkdns)
 target_include_directories(
   inkdns
   PUBLIC "${PROJECT_SOURCE_DIR}/include"
-  PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}" "${PROJECT_SOURCE_DIR}/src/iocore/net"
+  PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}"
 )
 
 target_link_libraries(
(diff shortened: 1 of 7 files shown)


JosiahWI:
[approve ci centos]

src/iocore/dns/CMakeLists.txt
```diff
@@ -18,4 +18,15 @@
 add_library(inkdns STATIC DNS.cc DNSConnection.cc DNSEventIO.cc Inline.cc SplitDNS.cc)
 add_library(ts::inkdns ALIAS inkdns)
 
-target_link_libraries(inkdns PUBLIC ts::inkevent ts::inkhostdb ts::proxy ts::tscore)
+target_include_directories(
+  inkdns
+  PUBLIC "${PROJECT_SOURCE_DIR}/include"
+  PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}" "${PROJECT_SOURCE_DIR}/src/iocore/net"
```
cmcfarlen, line 24: Is this include into src/iocore/net necessary?

JosiahWI: It's truthful. `dns` includes some of inknet's `P_` headers.

JosiahWI: Now that the `P_` includes are relative this should not be necessary anymore.

cmcfarlen (changes requested):
See if this works without the extra src include for net.

src/iocore/dns/CMakeLists.txt
JosiahWI: They don't look like they're relative to the project root. This might be a little shaky.

cmcfarlen (approved):

PR #10656, opened 2023-10-21
Remove unused getbyxxx methods from DNSProcessor

JosiahWI:
These old methods that used const char* are superceded by the ones that use std::string_view.

### DIFF
--- a/include/iocore/dns/DNSProcessor.h
+++ b/include/iocore/dns/DNSProcessor.h
@@ -121,8 +121,11 @@ struct DNSProcessor : public Processor {
   // NOTE: the HostEnt *block is freed when the function returns
   //
 
+  // To avoid lifetime issues, the name will be copied out of the view's
+  // buffer before the query is scheduled.
   Action *gethostbyname(Continuation *cont, std::string_view name, Options const &opt);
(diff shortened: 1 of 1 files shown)


JosiahWI:
[approve ci ubuntu]

JosiahWI:
I'm seeing a segfault in the CacheDir test on CI, but I can't reproduce it locally. It's plausible that the underlying string could go out of scope, causing the `std::string_view` to become invalid and generating a segfault; however, the `const char *` was getting wrapped in a `std::string_view` anyway by `DNSQueryData`, so I don't yet understand why this change would make a difference.

JosiahWI:
[approve ci centos ubuntu]

JosiahWI:
The CacheDir test has nothing to do with DNSProcessor. It looks like my changes are not responsible for the segfaults.

masaori335 (dismissed):
Looks good.

JosiahWI:
Don't merge yet. The semantics are correct, but it's non-obvious what's going on and extremely risky. I would like to add some comments in the appropriate places.

bryancall:
Please make this a non-Draft PR and we can review it.

masaori335 (approved):

PR #10667, opened 2023-10-24
generate svg files from uml for docs

cmcfarlen:
@bneradt noticed the cmake docs build was missing some images.  This PR generates the `svg` files from `uml` so they can be included in the docs.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -29,6 +29,49 @@ configure_file(ext/traffic-server.cmake.in.py ext/traffic-server.py)
 configure_file(conf.cmake.in.py conf.py)
 configure_file(manpages.cmake.in.py manpages.py)
 
+# copy static directory to build
+add_custom_target(
+  generate_docs_setup
+  COMMENT "Copy static files to build directory"
+  COMMAND ${CMAKE_COMMAND} -E copy_directory
+          ${CMAKE_CURRENT_SOURCE_DIR}/static
+          ${CMAKE_CURRENT_BINARY_DIR}/static
+)
+
+# generate svg files from uml files
+set(UML_FILES
+  uml/JsonRPCManager.uml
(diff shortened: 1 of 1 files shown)


doc/CMakeLists.txt
```diff
@@ -29,6 +29,49 @@ configure_file(ext/traffic-server.cmake.in.py ext/traffic-server.py)
...
+  uml/hdr-heap-str-alloc.plantuml
+  uml/host-resolve.plantuml
+  uml/l4-basic-sequence.uml
+  uml/l4-example-cdn-layout.uml
+  uml/l4-pre-warming-overview.uml
+  uml/l4-sni-routing-seq.uml
+  uml/l4-tcp-routing.uml
+  uml/traffic_ctl-class-diagram.uml
+  uml/url_rewrite.plantuml
+)
```
bneradt, line 56: Can we glob these? That should make things a bit easier for people adding these.

bneradt (changes requested):
This definitely works as can be seen here from the docs generated by this change:
https://ci.trafficserver.apache.org/job/Github_Builds/job/docs/3437/artifact/output/10667/docbuild/html/admin-guide/layer-4-routing.en.html

bneradt:
[approve ci ubuntu]

doc/CMakeLists.txt
cmcfarlen: We can glob them, but then a user will have to know to manually re-run cmake if they add a new file to the directory.  OTOH, adding a file to this list will auto generate cmake and then continue the build.  Its a trade off.

cmake doesn't support pattern matching that makefiles can do because other builds systems (ninja, Xcode, visual studios) don't have a similar feature.

bneradt: I see. Let's leave it as explicit then as you did for this patch. Thanks.

bneradt (approved):

PR #10671, opened 2023-10-24
Unified include paths

lzx404243:
This PR does the following
* updates the common include path used by all modules.
* The #include lines in code are updated to the relative path to the updated include paths or to where the including files are.
* Most `include_directories` are cleaned up in cmake files.

The common include path are the following:
```
# Common includes for everyone
include_directories(${CMAKE_SOURCE_DIR}/include ${CMAKE_BINARY_DIR}/include)
```

Also, since we are here,
* moved the private header (`P_xxx`)  and test headers from `include/records` back to the `src/records`.


lzx404243:
[approve ci autest]

cmcfarlen (approved):

PR #10672, opened 2023-10-25
Don't make boringssl mandatory when QUIC & quiche is enabled/used.

brbzull0:
If quiche is enabled, then we should allow for openssl/quictls to be used, specially if we have quiche/openssl support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -275,6 +275,7 @@ endif()
 find_package(PCRE REQUIRED)
 
 include(CheckOpenSSLIsBoringSSL)
+include(CheckOpenSSLIsQuictls)
 find_package(OpenSSL REQUIRED)
 check_openssl_is_boringssl(OPENSSL_IS_BORINGSSL BORINGSSL_VERSION "${OPENSSL_INCLUDE_DIR}")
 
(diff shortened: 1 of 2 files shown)


maskit:
The check itself is fine, but the name of CheckOpenSSLSupportsQUIC may be a bit of confusing because the official OpenSSL 3.2 supports QUIC too and we don't have plan to use their API at the moment (and we'd need different code for it). What we really want to check is whether the library is quictls. I'd name it CheckOpenSSLIsQuictls.

brbzull0:
> The check itself is fine, but the name of CheckOpenSSLSupportsQUIC may be a bit of confusing because the official OpenSSL 3.2 supports QUIC too and we don't have plan to use their API at the moment (and we'd need different code for it). What we really want to check is whether the library is quictls. I'd name it CheckOpenSSLIsQuictls.

You are right. I will change the name.
Thanks.

bneradt:
[approve ci clang-format]

brbzull0:
[approve ci]

brbzull0:
[approve ci]

brbzull0:
[approve ci format]

maskit:
@brbzull0 Do you want to merge this before the pr for quiche gets merged? It looks like the latest quiche which was released a few hours ago does not support quictls yet.

brbzull0:
> @brbzull0 Do you want to merge this before the pr for quiche gets merged? It looks like the latest quiche which was released a few hours ago does not support quictls yet.

We should probably merge this anyway, it helps with development while using quictls.  I don't know exactly when quiche/quictls wil be part of master.

maskit:
It would be nice if we could check whether special quiche which supports QuicTLS is used. I'm also fine with printing a message that says special quiche is required to use QuitTLS so users should not assume they can use the official quiche.

brbzull0:
> It would be nice if we could check whether special quiche which supports QuicTLS is used. I'm also fine with printing a message that says special quiche is required to use QuitTLS so users do not assume they can use the official quiche.

Make sense, warning added. For now it will just warn you. Once we get this into quiche/master we can use the version to validate.

maskit (approved):

PR #10673, opened 2023-10-25
Do not overwrite the error code of GOAWAY frame

duke8253:
We've been seeing some strange error codes being logged with H2 traffic, such as this one `S20666169` (where `S` stands for session).

Turns out when ATS processes GOAWAY frames, it overwrote error code with `Additional Debug Data` (https://datatracker.ietf.org/doc/html/rfc7540#section-6.8). This patch fixes the issue.

### DIFF
--- a/src/proxy/http2/Http2ConnectionState.cc
+++ b/src/proxy/http2/Http2ConnectionState.cc
@@ -837,7 +837,7 @@ Http2ConnectionState::rcv_goaway_frame(const Http2Frame &frame)
 {
   Http2Goaway goaway;
   char buf[HTTP2_GOAWAY_LEN];
-  unsigned nbytes               = 0;
+  char *end;
   const Http2StreamId stream_id = frame.header().streamid;
 
   Http2StreamDebug(this->session, stream_id, "Received GOAWAY frame");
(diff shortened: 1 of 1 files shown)


src/proxy/http2/Http2ConnectionState.cc
```diff
@@ -852,9 +852,14 @@ Http2ConnectionState::rcv_goaway_frame(const Http2Frame &frame)
   while (nbytes < frame.header().length) {
     unsigned read_bytes = read_rcv_buffer(buf, sizeof(buf), nbytes, frame);
 
-    if (!http2_parse_goaway(make_iovec(buf, read_bytes), goaway)) {
-      return Http2Error(Http2ErrorClass::HTTP2_ERROR_CLASS_CONNECTION, Http2ErrorCode::HTTP2_ERROR_PROTOCOL_ERROR,
-                        "goaway failed parse");
+    if (nbytes == HTTP2_GOAWAY_LEN) {
```
maskit, line 855: Can't we simply remove the while loop instead? I think the while loop is unnecessary for this frame (it makes sense for SETTINGS frame).

duke8253: Yeah, I think we can, but I'm not sure whether we plan on handling the `Additional Debug Data` part in the future. I'm down for removing the loop all together.

maskit (approved):

zwoop:
Cherry-picked to v9.2.x

PR #10680, opened 2023-10-26
Fix cmake test for 128 bit CAS.

ywkaras:
The cmake check_c_source_compiles() function seems to want to successfully link as well as compile the sample code.

https://godbolt.org/z/oGGcGY7Gn

https://stackoverflow.com/questions/62391538/how-can-i-link-sync-bool-compare-and-swap-16

### DIFF
--- a/cmake/Check128BitCas.cmake
+++ b/cmake/Check128BitCas.cmake
@@ -25,13 +25,10 @@
 
 set(CHECK_PROGRAM
     "
-    int
-    main()
+    int main(void)
     {
         __int128_t x = 0;
-        __sync_bool_compare_and_swap(&x,0,10);
-
-        return 0;
+        return __sync_bool_compare_and_swap(&x,0,10);
     }
     "
 )
@@ -41,7 +38,7 @@ check_c_source_compiles("${CHECK_PROGRAM}" TS_HAS_128BIT_CAS)
 
 if(NOT TS_HAS_128BIT_CAS)
     unset(TS_HAS_128BIT_CAS CACHE)
-    set(CMAKE_REQUIRED_FLAGS "-Werror" "-mcx16")
+    set(CMAKE_REQUIRED_FLAGS "-Werror -mcx16")
     check_c_source_compiles("${CHECK_PROGRAM}" TS_HAS_128BIT_CAS)
     set(NEED_MCX16 ${TS_HAS_128BIT_CAS})
     unset(CMAKE_REQUIRED_FLAGS)


cmake/Check128BitCas.cmake
```diff
@@ -25,18 +25,16 @@
...
     "
-    int
-    main()
+    int main(void)
     {
         __int128_t x = 0;
-        __sync_bool_compare_and_swap(&x,0,10);
-
-        return 0;
+        return __sync_bool_compare_and_swap(&x,0,10);
```
JosiahWI, line 31:
```suggestion
        return __sync_bool_compare_and_swap(&x,0,10) ? 0 : 1;
```
As the name suggests, `__sync_bool_compare_and_swap` returns a `bool`. https://gcc.gnu.org/onlinedocs/gcc-4.1.2/gcc/Atomic-Builtins.html

cmake/Check128BitCas.cmake
```diff
@@ -25,18 +25,16 @@
...
     }
     "
 )
 
 include(CheckCSourceCompiles)
+set(CMAKE_REQUIRED_FLAGS "-march=native")
 check_c_source_compiles("${CHECK_PROGRAM}" TS_HAS_128BIT_CAS)
```
JosiahWI, line 39: It might be wise to `unset(CMAKE_REQUIRED_FLAGS)`, because otherwise the variable persists and could influence the next call to `check_c_source_compiles` unintentionally.

JosiahWI (changes requested):
Thank you for the fix! I noticed a couple small things I'd like to address.

JosiahWI:
It looks like it needs to check for `__sync_val_compare_and_swap` as well?

cmake/Check128BitCas.cmake
ywkaras: Hmm except that `bool` isn't a type in all dialects of C:  https://godbolt.org/z/vrdE116WG

cmcfarlen (changes requested):
I think the bug here is line 45:
```
set(CMAKE_REQUIRED_FLAGS "-Werror" "-mcx16")
```
needs to be:
```
set(CMAKE_REQUIRED_FLAGS "-Werror -mcx16")
```

cmake/Check128BitCas.cmake
ywkaras: Not sure the sample code really needs to change at all.  Pedantic C doesn't allow nothing between the braces when declaring a function.  Some compilers might warn about ignoring the return value of __sync_bool_compare_and_swap().

ywkaras:
> It looks like it needs to check for `__sync_val_compare_and_swap` as well?

Autotools configure.ac doesn't check for that.  I guess it's assumed that whole group of functions will be available if one of them is.

JosiahWI:
> I think the bug here is line 45:
> 
> ```
> set(CMAKE_REQUIRED_FLAGS "-Werror" "-mcx16")
> ```
> 
> needs to be:
> 
> ```
> set(CMAKE_REQUIRED_FLAGS "-Werror -mcx16")
> ```

I would expect this to introduce a bug. The flags should be a list, so in CMake it should be semicolon-seperated, not space-separated, or am I missing something obvious here?

cmake/Check128BitCas.cmake
JosiahWI: I think the main issue isn't whether it returns a `bool` but whether the return value is correct. `__sync_bool_compare_and_swap` will return a truthy value if it succeeds, which will indicate an exit failure if main returns it. That shouldn't interfere with CMake's check, but it seems risky.

ywkaras:
> > I think the bug here is line 45:
> > ```
> > set(CMAKE_REQUIRED_FLAGS "-Werror" "-mcx16")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > needs to be:
> > ```
> > set(CMAKE_REQUIRED_FLAGS "-Werror -mcx16")
> > ```
> 
> @cmcfarlen I would expect this to introduce a bug. The flags should be a list, so in CMake it should be semicolon-seperated, not space-separated, or am I missing something obvious here?

Maybe you could use either `"-Werror -mcx16"` or `-Werror ; -mcx16`.

Perhaps someday, somebody will write some low-voodoo CM tools.

JosiahWI:
> String of additional flags to pass to the compiler. The string must be space-delimited--a [;-list](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#cmake-language-lists) will not work. The contents of [CMAKE_<LANG>_FLAGS](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS.html#variable:CMAKE_%3CLANG%3E_FLAGS) and its associated configuration-specific variable are automatically added to the compiler command before the contents of CMAKE_REQUIRED_FLAGS.

`CMAKE_REQUIRED_FLAGS` doesn't take a list - every other variable in the module does. Yay for inconsistent APIs. Thanks for the fix @cmcfarlen!

cmake/Check128BitCas.cmake
cmcfarlen: I don't think `check_c_source_compiles` actually runs the program

cmcfarlen (approved):

cmcfarlen:
> > String of additional flags to pass to the compiler. The string must be space-delimited--a [;-list](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#cmake-language-lists) will not work. The contents of [CMAKE__FLAGS](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS.html#variable:CMAKE_%3CLANG%3E_FLAGS) and its associated configuration-specific variable are automatically added to the compiler command before the contents of CMAKE_REQUIRED_FLAGS.
> 
> `CMAKE_REQUIRED_FLAGS` doesn't take a list - every other variable in the module does. Yay for inconsistent APIs. Thanks for the fix @cmcfarlen!

yeah, I was surprised by this when I looked at the cmake log.  It just silently ignores anything past the first argument...  I did test this way and it worked with the space.

PR #10685, opened 2023-10-27
make doc lang configurable with cmake

traeak:
Should allow docs to be built with custom language using something like:

cmake -B docbuild -DDOC_LANG:STRING='ja' -DENABLE_DOCS=ON


bneradt (changes requested):
We had some discussion earlier about whether we would continue to support the Japanese language in the docs. @masaori335 @maskit : can you remind us our decision? I seem to recall that since they were not being maintained currently that we intended to drop them.

maskit:
The decision was to remove it if we have issues with building localized docs and nobody wants to fix it. The feedback from Japanese community was basically "it's nice to have and we cannot blame anyone even if it's removed because we are not maintaining it".

Since @traeak have already made this fix, I don't think we have a reason to remove it now.

CMakeLists.txt
```diff
@@ -611,6 +611,7 @@ if(ENABLE_EXAMPLE)
 endif()
 
 if(ENABLE_DOCS)
+  set(DOC_LANG en CACHE STRING "Document language option")
```
cmcfarlen, line 614: Either move this set outside of ENALBE_DOCS here or add a conditional below for `cmake_print_variables` or it will be printing empty except for a docs build.

cmcfarlen (changes requested):
Or maybe just remove the print?

CMakeLists.txt
traeak: decided to check DOC_LANG before printing.

cmcfarlen (approved):

PR #10687, opened 2023-10-27
Update ACCEPT thread configurations

bneradt:
Before this change, ACCEPT NetHandler instances would not have their configurations updated, leaving them at their initialized defaults (usually 0). This should fix things for get_additional_accepts calls.

### DIFF
--- a/src/iocore/net/UnixNetAccept.cc
+++ b/src/iocore/net/UnixNetAccept.cc
@@ -54,8 +54,7 @@ net_accept(NetAccept *na, void *ep, bool blockable)
   UnixNetVConnection *vc = nullptr;
   Connection con;
 
-  EThread *t             = e->ethread;
-  int additional_accepts = get_NetHandler(t)->get_additional_accepts();
+  int additional_accepts = NetHandler::get_additional_accepts();
 
   if (!blockable) {
     if (!MUTEX_TAKE_TRY_LOCK(na->action_->mutex, e->ethread)) {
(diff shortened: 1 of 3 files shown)


src/iocore/net/UnixNetAccept.cc
```diff
@@ -571,8 +571,10 @@ NetAccept::acceptLoopEvent(int event, Event *e)
 {
   (void)event;
   (void)e;
-  EThread *t = this_ethread();
-
+  EThread *t    = this_ethread();
+  NetHandler *h = get_NetHandler(t);
```
moonchen, line 575: Since you're on a dedicated accept thread, and NetHandler is only initialized on ET_NET threads, the pointer h now points to an uninitialized NetHandler object.  (The memory belongs to PER_THREAD_DATA, and is actually zeroed in EThread::EThread)

My suggestion would be to store this config in a thread-local block of memory, or to have the dedicated accept threads read the config values from the global config, if the config is not expected to change often.

bneradt: Interesting. Is this line a problem already then?
https://github.com/apache/trafficserver/blob/fc3f12e5eb360a32845ddbf3cba64294c8cb1ddf/src/iocore/net/UnixNetAccept.cc#L58

There's no verification at that point that the thread is a NET type. There is logic later in the function to verify NET threads, but only conditionally:
https://github.com/apache/trafficserver/blob/fc3f12e5eb360a32845ddbf3cba64294c8cb1ddf/src/iocore/net/UnixNetAccept.cc#L114

sharkleash (approved):
👍 thanks for doing this

moonchen (approved):
lgtm

bneradt:
[approve ci centos]

PR #10693, opened 2023-10-30
cmake: resolve install prefix later to pickup prefix override

cmcfarlen:
If you `cmake --install <build> --prefix <prefix>` then the default config files were installed to the wrong place.


cmake/install_configs.cmake
```diff
@@ -22,11 +22,12 @@
 
 
 file(GLOB CONFIG_FILES ${CONFIG_SOURCE_GLOBS})
-file(MAKE_DIRECTORY $ENV{DESTDIR}${CONFIG_DEST_PATH})
+file(MAKE_DIRECTORY $ENV{DESTDIR}${CMAKE_INSTALL_PREFIX}${CONFIG_DEST_PATH})
```
JosiahWI, line 25: Does the argument here need to be quoted now?

JosiahWI (approved):
Nice fix, thanks!

bneradt (approved):

PR #10694, opened 2023-10-30
Refactor new metrics with Counter vs Gauge types

zwoop:
This adds some more tooling around `Counter` vs `Gauge` type of metrics, assuring that Counters for example can only increment. There's no change in functionality here, other than the naming  and creation details. But it will assure that the underlying APIs are used correctly.

Performance wise this should make no difference, at least in a release build. The compiler will optimize this down to just the `fetch_add()` just as before.


masaori335:
[approve ci clang-analyzer]

ywkaras:
Maybe you can fix the Clang Analyzer defect by inserting `ink_assert(d != nullptr);` after line 1269.

zwoop:
> Maybe you can fix the Clang Analyzer defect by inserting `ink_assert(d != nullptr);` after line 1269.

I could, but the thing is, this is not my fault :). I asked @masaori335 to look at this, I'm suspecting that something in my PR makes clang-analyzer either confused, or brilliant, in detecting this old issue.

ywkaras:
Clang Analyzer punishes for unconfessed sins and those from previous lifetimes.

zwoop:
> Clang Analyzer punishes for unconfessed sins and those from previous lifetimes.

True dat.

masaori335:
[approve ci clang-analyzer]

include/api/Metrics.h
```diff
@@ -289,10 +280,126 @@ class Metrics
...
+  // These are sort of factory classes, using the Metrics singleton for all storage etc.
+  class Gauge
+  {
+  public:
+    using self_type = Gauge;
+    using SpanType  = Metrics::SpanType;
+
+    class AtomicType : public Metrics::AtomicType
+    {
+    };
```
ywkaras, line 297: Why not just `using AtomicType = Metrics::AtomicType;` ?

include/api/Metrics.h
```diff
@@ -45,23 +45,47 @@ class Metrics
 private:
   using self_type = Metrics;
 
+  class AtomicType
+  {
+    friend class Metrics;
+
+  public:
+    AtomicType()          = default;
+    virtual ~AtomicType() = default;
```
ywkaras, line 54: Making this virtual doubles the size of instances of this type.  Why is it necessary?

include/api/Metrics.h
```diff
@@ -289,10 +280,126 @@ class Metrics
...
+
+      return instance._create(name);
+    }
+
+    static AtomicType *
+    createPtr(const std::string_view name)
+    {
+      auto &instance = Metrics::instance();
+
+      return reinterpret_cast<AtomicType *>(instance.lookup(instance._create(name)));
```
ywkaras, line 312: This looks like the overload of `lookup()` that returns `IdType/uint32_t`.  How can that work, to cast it to a64-bit pointer?

include/api/Metrics.h
```diff
@@ -289,10 +280,126 @@ class Metrics
...
+
+    static void
+    increment(AtomicType *metric, uint64_t val = 1)
+    {
+      ink_assert(metric);
+      metric->_value.fetch_add(val, MEMORY_ORDER);
+    }
+
+    static int64_t
+    load(AtomicType *metric)
```
ywkaras, line 395: This can't be const?

include/api/Metrics.h
```diff
@@ -103,16 +125,9 @@ class Metrics
     return std::nullopt;
   }
```
ywkaras, line 126: This function is not currently used.

include/proxy/http2/Http2ServerSession.h
```diff
@@ -72,13 +72,13 @@ class Http2ServerSession : public PoolableSession, public Http2CommonSession
   void
   increment_current_active_connections_stat() override
   {
-    Metrics::increment(http2_rsb.current_active_server_connection_count);
+    Metrics::Gauge::increment(http2_rsb.current_active_server_connection_count);
```
ywkaras, line 75: Since you had to change all of these anyway, why didn't you make increment and decrement non-static member functions:
`http2_rsb.current_active_server_connection_count->increment()`.  It's shorter, and you would get a compiler error if you tried to decrement a counter rather than a gauge.

include/api/Metrics.h
zwoop: Because then the compiler won't give errors or warnings about wrong types.

include/api/Metrics.h
zwoop: Same as above :/

include/api/Metrics.h
zwoop: That's a different lookup no? I'd have to look again, but remember there are two lookups, one to that returns the IdType, and one that returns a pointer directly to the atomic types.

include/api/Metrics.h
zwoop: Maybe? I can try.

include/api/Metrics.h
```diff
@@ -289,10 +280,126 @@ class Metrics
...
+    }
+
+    static int64_t
+    load(AtomicType *metric)
+    {
+      ink_assert(metric);
+      return metric->_value.load();
+    }
+
+  }; // class Counter
```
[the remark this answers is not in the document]
zwoop: clear? Why clear? All counters starts at zero, and can always go up. You should not set it back to 0 IMO.

include/api/Metrics.h
zwoop: It's used in the tests :). But it could be used, maybe even should be used. Certainly in plugins.

include/proxy/http2/Http2ServerSession.h
zwoop: I thought of that, and even asked @cmcfarlen , we both preferred the static. I can change it if we think it's better. It just reads better with the static IMO.

ywkaras: If you did Guage::increment() on a counter, or Counter::increment() on a guage, you would not get a compiler error.  Is this part of @cmcfarlen 's philosphy that C++ code should look like Python?

include/api/Metrics.h
ywkaras: above what?

include/api/Metrics.h
ywkaras: OK I see but it shouldn't need the case then.

include/api/Metrics.h
```diff
@@ -289,10 +280,126 @@ class Metrics
...
+
+      return instance._create(name);
+    }
+
+    static AtomicType *
+    createPtr(const std::string_view name)
+    {
+      auto &instance = Metrics::instance();
+
+      return reinterpret_cast<AtomicType *>(instance.lookup(instance._create(name)));
```
ywkaras, line 376: Another superfluous cast.

include/api/Metrics.h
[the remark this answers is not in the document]
ywkaras: Yes there are pros and cons to having clear().  I was thinking we had set() to do a clear, so just have clear().

src/iocore/aio/test_AIO.cc
```diff
@@ -223,11 +223,11 @@ dump_summary()
 
   auto &m = Metrics::getInstance();
 
-  Metrics::IntType *completed = m.lookup(m.lookup("proxy.process.io_uring.completed"));
-  Metrics::IntType *submitted = m.lookup(m.lookup("proxy.process.io_uring.submitted"));
+  Metrics::Counter::AtomicType *completed = m.lookup(m.lookup("proxy.process.io_uring.completed"));
+  Metrics::Counter::AtomicType *submitted = m.lookup(m.lookup("proxy.process.io_uring.submitted"));
```
ywkaras, line 227: You could have had two more useless casts here.

src/iocore/aio/test_AIO.cc
```diff
@@ -223,11 +223,11 @@ dump_summary()
...
 
-  printf("submissions: %lu\n", Metrics::read(submitted));
-  printf("completions: %lu\n", Metrics::read(completed));
+  printf("submissions: %lu\n", Metrics::Gauge::load(submitted));
```
ywkaras, line 229: Really %lu quit slacking https://en.cppreference.com/w/cpp/types/integer

include/api/Metrics.h
```diff
@@ -289,10 +280,126 @@ class Metrics
...
+    static Metrics::IdType
+    create(const std::string_view name)
+    {
+      auto &instance = Metrics::instance();
+
+      return instance._create(name);
+    }
+
+    static AtomicType *
+    createPtr(const std::string_view name)
```
ywkaras, line 308: const std::string_view is overkill, all the member functions are const I think.

src/iocore/hostdb/P_RefCountCache.h
```diff
@@ -428,21 +428,19 @@ RefCountCache<C>::RefCountCache(unsigned int num_partitions, int size, int items
...
-  this->rsb.refcountcache_current_items        = intm.newMetricPtr((metrics_prefix + "current_items").c_str());
-  this->rsb.refcountcache_current_size         = intm.newMetricPtr((metrics_prefix + "current_size").c_str());
-  this->rsb.refcountcache_total_inserts        = intm.newMetricPtr((metrics_prefix + "total_inserts").c_str());
-  this->rsb.refcountcache_total_failed_inserts = intm.newMetricPtr((metrics_prefix + "total_failed_inserts").c_str());
-  this->rsb.refcountcache_total_lookups        = intm.newMetricPtr((metrics_prefix + "total_lookups").c_str());
-  this->rsb.refcountcache_total_hits           = intm.newMetricPtr((metrics_prefix + "total_hits").c_str());
-  this->rsb.refcountcache_last_sync_time       = intm.newMetricPtr((metrics_prefix + "last_sync.time").c_str());
-  this->rsb.refcountcache_last_total_items     = intm.newMetricPtr((metrics_prefix + "last_sync.total_items").c_str());
-  this->rsb.refcountcache_last_total_size      = intm.newMetricPtr((metrics_prefix + "last_sync.total_size").c_str());
+  this->rsb.refcountcache_current_items        = Metrics::Gauge::createPtr((metrics_prefix + "current_items").c_str());
```
ywkaras, line 435: Why are you doing string -> c string -> string_view instead of just string -> string_view directly?

zwoop: Probably a remnant from the code as it was prior to changing to new metrics.

src/iocore/aio/test_AIO.cc
zwoop: Yeh, not my code.

include/api/Metrics.h
zwoop: No, I don't think it's superfluous, it won't compile otherwise I'm fairly certain.

include/api/Metrics.h
zwoop: does it hurt? Other than your eyes ?

include/api/Metrics.h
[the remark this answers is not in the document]
zwoop: Well, I'll probably argue that if you need to do clear(), you are using this wrong, and it should be a gauge. Counters are supposed to always increase, randomly setting it to 0 (or any value) is wrong, and why I even bothered adding this difference of Counters vs Gauges.

include/api/Metrics.h
zwoop:
```
/Users/leif/apache/trafficserver/include/api/Metrics.h:312:14: error: cannot initialize return object of type 'AtomicType *' (aka 'ts::Metrics::Gauge::AtomicType *') with an rvalue of type 'AtomicType *' (aka 'ts::Metrics::AtomicType *')
      return instance.lookup(instance._create(name));
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

src/iocore/aio/test_AIO.cc
zwoop: You implying we should have another lookup function, that avoids both lookups here/ This is an unusual use case honestly, the the IdType lookups are kept mostly because of the old APIs and use of "ids" instead of pointers.

We can probably eliminate some of this in v11 once we clean out more old stuff completely. Or, if we could convince Strostrup to allow overloading to take the return type into the resolution.

include/api/Metrics.h
zwoop: Ah yeh, I remember now, it can't. I blame the Danes.

```
/Users/leif/apache/trafficserver/include/api/Metrics.h:395:30: error: static member function cannot have 'const' qualifier
    load(AtomicType *metric) const
```

zwoop: Ok, fixed, and uploaded with a new commit for your review suggestions (will add more as if we identify more things that needs fixing).

include/api/Metrics.h
zwoop: It had to be virtual, otherwise I couldn't subclass it later as I did. And I had to sub class it, rather than aliases, to make the compiler detect bad usage.

zwoop: Maybe the right solution here is to have 

```
Metrics::Counter::lookup(stats_id);
Metrics::Gauge::lookup(stats_id);
```

?

include/api/Metrics.h
ywkaras: Why don't you give it the constructor `explicit AtomicType(Metrics::AtomicType a) : Metrics::AtomicType(a) {}` so you won't need the reinterpret_cast.

include/api/Metrics.h
ywkaras: Alan will sentence you to hard labor on his pumpkin farm for this.

ywkaras (approved):

include/api/Metrics.h
zwoop: @ywkaras I added those static's, which allows us to eliminate the virtual. Let me know what you think, it's in the last commit that I lovingly made just for you.

include/api/Metrics.h
zwoop: Ok, let me look tomorrow morning.

ywkaras:
How about:  https://godbolt.org/z/7Wv444EfE

include/api/Metrics.h
zwoop: Removed.

bneradt:
[approve ci]

bneradt:
[approve ci]

ywkaras (dismissed):

src/proxy/http/PreWarmManager.cc
```diff
@@ -1122,48 +1133,64 @@ PreWarmManager::_parse_sni_conf(PreWarm::ParsedSNIConf &parsed_conf, const SNICo
...
+
+    // First the Counters
+    for (int j = 0; j < static_cast<int>(PreWarm::CounterStat::LAST_ENTRY); ++j) {
       char name[STAT_NAME_BUF_LEN];
 
-      if (dst->alpn_index != SessionProtocolNameRegistry::INVALID) {
-        std::string_view alpn_name = alpn_name_for_stat(dst->alpn_index);
+      _makeName(dst, COUNTER_STAT_ENTRIES[j], name, sizeof(name));
+
+      auto metric = Metrics::Counter::createPtr(name); // This will do a lookup if it already exists
```
masaori335, line 1168: This looks much convenient 👍

masaori335 (approved):
This makes metrics type explicit. Looks good to me.

masaori335 (approved):

ywkaras (approved):

PR #10699, opened 2023-10-31
Add len to urlsig debug printing

ezelkow1:
We have seen some logged instances of junk in urlsig error logging. I havent been able to reproduce it but this seems to be one obvious path where its possible some of these are not null terminated on the internal API call. So just sending in the url len to the debug prints


traeak (approved):
These all look fine.  Using std::string_view could be more "fun" but not necessary.

bneradt (approved):

PR #10701, opened 2023-10-31
Support upstream MPTCP

serrislew:
MPTCP is now supported in upstream linux kernels. Allow ATS to use `IPPROTO_MPTCP` if configured
- for ATS logging, 5.16+ kernel needed to detect mptcp_info struct

### DIFF
--- a/mgmt/rpc/overridable_txn_vars.cc
+++ b/mgmt/rpc/overridable_txn_vars.cc
@@ -1,0 +1,1 @@
+#include "../../src/shared/overridable_txn_vars.cc"
--- a/mgmt/rpc/overridable_txn_vars.cc
+++ b/mgmt/rpc/overridable_txn_vars.cc
@@ -1,1 +1,0 @@
-#include "../../src/shared/overridable_txn_vars.cc"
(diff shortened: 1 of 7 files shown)


bneradt (changes requested):
I assume we'll need a change to CMakeLists.txt for cmake builds now that that's the supported build system for master. Or am I missing something? Be aware that all of PR CI now runs cmake for master, so it will not cover anything done in configure.ac.

mgmt/rpc/overridable_txn_vars.cc
```diff
@@ -0,0 +1 @@
+#include "../../src/shared/overridable_txn_vars.cc"
```
cmcfarlen, line 1: The mgmt code moved to src/mgmt, but this file shouldn't be necessary regardless.  Please remove.

serrislew: Oh I don't even remember including this.. will remove

bneradt:
[approve ci clang-format]

bneradt (approved):

bneradt:
[approve ci clang-format]

PR #10706, opened 2023-11-01
Extract Stripe::aggregate_pending_writes

JosiahWI:
This extracts aggregate_pending_writes from aggWrite and adds a javadoc comment intended to document the operation of the method for future developers. This is an internal API so it intentionally goes into implementation details in an attempt to make this part of the write sequence less mysterious.

### DIFF
--- a/src/iocore/cache/P_CacheVol.h
+++ b/src/iocore/cache/P_CacheVol.h
@@ -218,6 +218,27 @@ struct Stripe : public Continuation {
 
   int aggWriteDone(int event, Event *e);
   int aggWrite(int event, void *e);
+
+  /**
+   * Copies virtual connection buffers into the aggregate write buffer.
+   *
+   * Pending write data will only be copied while space remains in the aggregate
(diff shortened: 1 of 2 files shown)


src/iocore/cache/P_CacheVol.h
```diff
@@ -218,6 +218,27 @@ struct Stripe : public Continuation {
...
+   *
+   *     - sync (only if CacheVC::f.use_first_key): inserted into sync queue
+   *     - evacuator: handler invoked - probably evacuateDocDone
+   *     - otherwise: inserted into tocall for handler to be scheduled later
+   *
+   * @param tocall Out parameter; a queue of virtual connections with handlers that need to
+   *     invoked at the end of aggWrite.
+   * @see aggWrite
+   */
+  void aggregate_pending_writes(Queue<CacheVC, Continuation::Link_link> &tocall);
```
masaori335, line 241: We can make it later, but how's make this private function?

masaori335 (approved):
No logical change. Thanks for adding description of the function.

PR #10708, opened 2023-11-01
Removes the cache_key_genid plugin, as agreed

zwoop:


shukitchan (approved):
I don't see any document on this plugin. So i think we are fine.

PR #10709, opened 2023-11-01
Create library target for overridable_txn_vars

JosiahWI:
This creates a new STATIC library target to avoid dependency issues, and moves some includes from overridable_txn_vars.h to
overridable_txn_vars.cc to keep the dependency on ts::inknet PRIVATE.


cmcfarlen (approved):
Wonderful! Thank you!

PR #10717, opened 2023-11-02
cmake: adding a release preset

bryancall:
Used for benchmarking and users can use it

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -37,6 +37,19 @@
       }
     },
     {
+      "name": "release",
+      "displayName": "Release build",
+      "description": "Release build with Ninja generator",
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/build-release",
+      "cacheVariables": {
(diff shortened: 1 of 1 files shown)


JosiahWI (approved):

cmcfarlen (approved):

CMakePresets.json
```diff
@@ -36,6 +36,19 @@
...
       }
     },
+    {
+      "name": "release",
+      "displayName": "Release build",
+      "description": "Release build with Ninja generator",
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/build-release",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Release",
```
ywkaras, line 46: versus "Debug"?

Do -D command line options override the presets?

JosiahWI: You can set variables in the cache once it exists:
```
cmake --preset release
cmake -B build-release -DCMAKE_BUILD_TYPE=Debug
```

But I do not think CMake supports doing this in one step.

PR #10710, opened 2023-11-01
Removes all the stats and cache inspector pages

zwoop:

PR #10718, opened 2023-11-02
Extract class AggregateWriteBuffer from Stripe

JosiahWI:
This encapsulates the `agg_` fields of `Stripe` in a new class. This is in preparation to move all the aggregation behavior into this new class, including the behavior to enqueue virtual connections and update the number of bytes waiting to be written to the buffer, which is currently being done by the virtual connections themselves.

Note: `Stripe` does not use its own getters to access aggregate buffer, but calls the getters on the `AggregateWriteBuffer` itself. This is because I want to get rid of as many of the getters on `Stripe` as possible.

### DIFF
--- a/src/iocore/cache/P_CacheVol.h
+++ b/src/iocore/cache/P_CacheVol.h
@@ -26,6 +26,7 @@
 #include "P_CacheDir.h"
 #include "P_CacheStats.h"
 #include "P_RamCache.h"
+#include "iocore/cache/AggregateWriteBuffer.h"
 
 #include "tscore/CryptoHash.h"
 
@@ -75,7 +76,7 @@
 #define DOC_NO_CHECKSUM ((uint32_t)0xA0B0C0D0)
 
 struct Cache;
-struct Stripe;
+class Stripe;
 struct CacheDisk;
 struct StripeInitInfo;
 struct DiskStripe;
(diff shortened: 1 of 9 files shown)


JosiahWI:
HostDB regression test failed.

JosiahWI:
[approve ci autest]

JosiahWI:
[approve ci ubuntu]

JosiahWI:
[approve ci autest centos]

JosiahWI:
[approve ci autest]

JosiahWI:
I will rebase and mark ready for review after #10710 lands.

src/iocore/cache/P_CacheVol.h
```diff
@@ -274,12 +271,18 @@ struct Stripe : public Continuation {
...
-  ~Stripe() override { ats_free(agg_buffer); }
+  Queue<CacheVC, Continuation::Link_link> &get_pending_writers();
+  char *get_agg_buffer();
+  int get_agg_buf_pos() const;
+  void reset_agg_buf_pos();
+  int get_agg_todo_size() const;
+  void add_agg_todo(int size);
+
+private:
+  AggregateWriteBuffer write_buffer;
```
masaori335, line 285: Please add `_` as prefix to follow the naming convention.

masaori335 (approved):

PR #10720, opened 2023-11-02
Update CMakeLists.txt

dragon512:
Allow plugin to build if required library exists


bneradt:
[approve ci rocky]

JosiahWI (changes requested):
Thanks for the fix and the status message. Please run cmake-format - the body of the if statement should be indented.

dragon512:
@JosiahWI Please make sure when you make such change requests are you more exact in what you request. What is cmake-format?? I don't see this as a command on my system.

JosiahWI:
We need to set up something similar to the clang-format hook for cmake-format. To document a workflow here:

```
python3 -m pip install cmake-format
cmake-format -i <path-to-CMakeLists.txt>
```

JosiahWI:
[ci approve clang-format]

JosiahWI (approved):

JosiahWI:
[approve ci clang-format]

bneradt:
[approve ci]

plugins/experimental/maxmind_acl/CMakeLists.txt
```diff
@@ -14,11 +14,16 @@
 #  the License.
 #
 #######################
+find_package(maxminddb QUIET)
```
JosiahWI, line 17: @traeak observed that `QUIET` silences the messages indicating the package was found. We probably don't want that. 
```cmake
find_package(maxminddb)
```
will keep the informational messages.

We want to change this to an `auto_option` but we can change this and `stek_share` together in a separate PR.

JosiahWI (changes requested):

bneradt:
[approve ci autest]

dragon512:
> We need to set up something similar to the clang-format hook for cmake-format. To document a workflow here:
> 
> ```
> python3 -m pip install cmake-format
> cmake-format -i <path-to-CMakeLists.txt>
> ```

I agree.. we have a pipenv file in test for testing.. we might want to consider having this file in the root as part of a better setup process.

dragon512:
[approve ci autest]

dragon512:
Closing PR.. will remake a new one with other tweaks as the better pattern is now known for us to use

PR #10721, opened 2023-11-02
Remove the direct dependency for HttpCacheSM from iocore/cache

maskit:
This PR changes the way of telling `HttpCacheSM` that cache will be RWW. The new way use an event instead of calling `HttpCacheSM::set_readwhilewrite_inprogress()` directly, and that removes the necessity of knowing `HttpCacheSM`. Although an event is used, the event is dispatched by calling `handleEvent` so it's still processed instantly.

test_RWW needed a bit of adjustment to handle the new event, so the new event won't be caught by `default` case which make tests fail. In other words, the new event is dispatched to the intended continuation / event handler.


bneradt:
[approve ci osx]

bneradt:
[approve ci osx]

bneradt:
[approve ci format]

SolidWallOfCode (approved):

PR #10723, opened 2023-11-03
uri_signing cmake: clean up unit test cmake and change suggested github repo

traeak:
Some cmake cleanup and change the suggested source for the cjose repo from the unmaintained cisco to one maintained by openidc.

### DIFF
--- a/plugins/experimental/uri_signing/unit_tests/CMakeLists.txt
+++ b/plugins/experimental/uri_signing/unit_tests/CMakeLists.txt
@@ -26,8 +26,6 @@ add_executable(
   ${PROJECT_SOURCE_DIR}/match.cc
   ${PROJECT_SOURCE_DIR}/normalize.cc
   ${PROJECT_SOURCE_DIR}/parse.cc
-  ${PROJECT_SOURCE_DIR}/timing.cc
-  ${PROJECT_SOURCE_DIR}/uri_signing.cc
 )
 target_compile_definitions(uri_signing_test PRIVATE UNITTEST)
 target_include_directories(uri_signing_test PRIVATE ${PROJECT_SOURCE_DIR})
--- a/plugins/experimental/uri_signing/unit_tests/CMakeLists.txt
+++ b/plugins/experimental/uri_signing/unit_tests/CMakeLists.txt
@@ -26,6 +26,7 @@ add_executable(
   ${PROJECT_SOURCE_DIR}/match.cc
   ${PROJECT_SOURCE_DIR}/normalize.cc
   ${PROJECT_SOURCE_DIR}/parse.cc
+  ${PROJECT_SOURCE_DIR}/timing.cc
 )
 target_compile_definitions(uri_signing_test PRIVATE UNITTEST)
 target_include_directories(uri_signing_test PRIVATE ${PROJECT_SOURCE_DIR})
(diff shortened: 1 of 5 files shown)


JosiahWI (approved):
Improves cohesion.

JosiahWI (changes requested):
Please keep the `auto_option`s.

traeak:
[approve ci clang-format]

JosiahWI (approved):
Summary of changes:

## Changes to the uri_signing plugin
- The auto options for cjose and jansson were removed in favor of calling find_package conditional on the experimental plugins being enabled.
- The documentation was updated to refer admins to a more maintained cjose repo with security patches.
- Unneeded source files were removed from the sources list for the unit tests

## Changes to informational messages
- The maxmind_acl, stek_share, and uri_signing plugins now print informational messages indicating whether or not they will be built.
- The maxmind_acl and uri_signing plugins no longer silence informational messages about whether required packages were found or not. This change was not made for the stek_share plugin because it looks for its dependency in `CONFIG` mode, and the warning when it is not found is unpleasantly verbose.

traeak:
Removed the maxmind_acl and stek_share changes.

JosiahWI (approved):

JosiahWI (approved):

plugins/experimental/uri_signing/unit_tests/CMakeLists.txt
```diff
@@ -26,8 +26,6 @@ add_executable(
   ${PROJECT_SOURCE_DIR}/match.cc
   ${PROJECT_SOURCE_DIR}/normalize.cc
   ${PROJECT_SOURCE_DIR}/parse.cc
-  ${PROJECT_SOURCE_DIR}/timing.cc
```
ywkaras, line 29: What is the point of timing.cc?

ywkaras: This file doesn' t set URI_SIGNING_UNIT_TEST that I can see.  Did you build/run the unit test after you make this change?

traeak: it builds just fine (using the autest preset).
In a previous PR I only removed the timing calls from the uri_signing.cc.  They were only returning ns level values which aren't particularly useful and junked up the uri_signing debug.  I wouldn't mind removing the rest of the references but perhaps in a later PR.

plugins/experimental/uri_signing/unit_tests/CMakeLists.txt
```diff
@@ -26,8 +26,6 @@ add_executable(
...
-  ${PROJECT_SOURCE_DIR}/uri_signing.cc
 )
 target_compile_definitions(uri_signing_test PRIVATE UNITTEST)
```
ywkaras, line 30: Looks like the unit test stuff still needs some cleanup.  The -D for the unit test is URI_SIGNING_UNIT_TEST ( https://github.com/apache/trafficserver/blob/6e00344df420db067e2b3cddbe8a7a48b7559700/plugins/experimental/uri_signing/common.cc#L21 ), not UNITTEST.  But, since the test works, it looks like the code triggered by URI_SIGNING_UNIT_TEST can just be deleted.  But we could do all that in a separate PR.

ywkaras (dismissed):

traeak:
[approve ci clang-format]

ywkaras (approved):
I'm still confused (but perhaps just my natural state).  If the unit test worked with the wrong -D, doesn't that mean we don't need the -D, or the code it enables?  But anyway, approved, go with God.

PR #10725, opened 2023-11-03
Cleanup build with tsapibackend

cmcfarlen:
This introduces a new target: tsapicommon

The build has been very fragile with dependencies around the tsapi library, especially with unit tests

This PR hopes to ameliorate much of that by introducing a 3rd layer of "api" dependency to detangle some linker issues.  The intent being that libraries can depend on tsapicommon without also pulling in tsapi which drags in the entire project.

There is still the issue of FetchSM which uses tspai, but iocore/net OCSP stapling also uses.  For this, I have added stubs so that unit tests may compile without tsapi.

Now the grouping of symbols for the 3 api layers are:

tsapicore - Stuff that even tscore can use (metrics, diags, dbctl, regex etc) that the plugins can also use
tsapicommon - symbols necessary to interact with the proxy side of tsapi internals (hooks, callbacks, etc)
tsapi - the actual plugin API that ends up pulling in the entire ts project

Ideally, these three libraries can be built as either shared or static where most likely they will be build both ways in the end.

With this change introduces, I was finally able to move the tests in src/tests back to their respective homes and greatly simplify their link stages.

EDIT: Also, tsapicore and tsapicommon follow BUILD_SHARED_LIBS to determine static or shared libraries, but that won't work until we sort out some more cycles (more PRs like #10721).


cmcfarlen:
[approve ci]

cmcfarlen:
[approve ci]

src/api/CMakeLists.txt
```diff
@@ -32,23 +37,32 @@ set(TSAPI_PUBLIC_HEADERS
...
+  InkVConnInternal.cc # proxy, http
+  APIHook.cc # iocore
+  APIHooks.cc # proxy
+  LifecycleAPIHooks.cc # proxy, http
+  HttpHookState.cc # proxy, http
+)
+add_library(ts::tsapicommon ALIAS tsapicommon)
+target_link_libraries(tsapicommon
+  PUBLIC ts::inknet ts::http
+  PRIVATE ts::proxy ts::hdrs ts::tscore)
```
bneradt, line 53: Just to record it here: we decided in the PR/issue meeting to rename this to something other than `tsapicommon`. Maybe `tsapibackend`?

bneradt (changes requested):

bneradt:
[approve ci]

cmcfarlen:
[approve ci]

cmcfarlen:
[approve ci]

bneradt (dismissed):
Looks fine to me. Let me know when you rebase to resolve conflicts and I'll re-approve.

bneradt (approved):

PR #10726, opened 2023-11-03
Harden the tunnel_transform Autest

lzx404243:
The `tunnel_transform` test failed internally in certain environment as the actual TLS traffic byte counts fell out of the hard-coded range of expected value. 

The TLS byte count can varied across different openssl/curl versions, so this PR adds a simple forwarding proxy to keep track of the exact byte count between the client and server, as suggested in the test comment. The exact byte count is used in the validation.

Note: originally I was trying a `nc` + pipe solution. It worked in my environment but I am concerned about the portability of different versions of `nc` and different characteristics of named pipe on different platforms(unidirectional VS bidirectional). So sticking with the hand-rolled proxy for portability.

### DIFF
--- a/tests/gold_tests/tunnel/dumb_proxy.py
+++ b/tests/gold_tests/tunnel/dumb_proxy.py
@@ -41,7 +41,7 @@ def parse_args() -> argparse.Namespace:
     parser.add_argument(
         '--forwarding_port',
         type=int,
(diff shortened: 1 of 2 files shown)


lzx404243:
[approve ci]

shinrich (approved):
Looks good.  "Some bright soul" came along.

PR #10728, opened 2023-11-04
Cleanup of the rate_limit simulator

zwoop:
Note that this doesn't quite fix the simulator, we have to figure out how to feed it a YAML initialization.


shukitchan (approved):
This looks reasonable after the previous refactor.

Eventually it may be helpful to document the simulator as well with instruments to compile and run it.

PR #10734, opened 2023-11-06
Add logic to find nuraft

dragon512:
Use standard pattern to find libraries and naming for constancy


JosiahWI:
This will break the ability to find `NuRaftConfig.cmake` on case sensitive systems. I can submit a PR to NuRaft to change the name to `nuraft-config.cmake`.

We could just rely on the Find module, but I think in general it is a good idea not to write Find modules if we do not have to. It's code that we then have to maintain.

dragon512:
> This will break the ability to find `NuRaftConfig.cmake` on case sensitive systems. I can submit a PR to NuRaft to change the name to `nuraft-config.cmake`.
> 
> We could just rely on the Find module, but I think in general it is a good idea not to write Find modules if we do not have to. It's code that we then have to maintain.

I think you just found that everything is still working with no change on your PR ... is the correct @JosiahWI

dragon512:
That is fine with me. We need this now. we can change it later if we want to after a new nuraft is made. Till then I believe it better to follow convention we have in CMake to not mix cases

cmcfarlen (dismissed):

cmcfarlen (approved):

Issue #10628, opened 2023-10-16
Rename Vol to Stripe 


masaori335:
Cache subsystem has a struct called `Vol`, but actually it's `Stripe` in the docs and comments.

https://github.com/apache/trafficserver/blob/fffcfb2cda7e918175171ce86939c6361722e1b1/iocore/cache/P_CacheVol.h#L125

PR #10739, opened 2023-11-07
Cleanup: Move stripe init functions

masaori335:
Part of #10628.

### DIFF
--- a/src/iocore/cache/Stripe.cc
+++ b/src/iocore/cache/Stripe.cc
@@ -37,30 +37,6 @@ DbgCtl dbg_ctl_cache_init{"cache_init"};
 short int const CACHE_DB_MAJOR_VERSION_COMPATIBLE = 21;
 
 void
-vol_init_data_internal(Stripe *vol)
-{
-  // step1: calculate the number of entries.
-  off_t total_entries = (vol->len - (vol->start - vol->skip)) / cache_config_min_average_object_size;
-  // step2: calculate the number of buckets
-  off_t total_buckets = total_entries / DIR_DEPTH;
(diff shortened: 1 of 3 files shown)


src/iocore/cache/Stripe.cc
```diff
@@ -36,30 +36,6 @@ DbgCtl dbg_ctl_cache_init{"cache_init"};
...
-  vol->start = vol->skip + 2 * vol->dirlen();
-}
-
-void
-vol_init_data(Stripe *vol)
-{
-  // iteratively calculate start + buckets
-  vol_init_data_internal(vol);
-  vol_init_data_internal(vol);
-  vol_init_data_internal(vol);
```
masaori335, line 60: We don't need to call this function 3 times...right?

masaori335: Good point. I'll make it call 3 times as is.

src/iocore/cache/unit_tests/test_CacheDir.cc
```diff
@@ -88,7 +88,7 @@ class CacheDirTest : public CacheInit
       return EVENT_DONE;
     }
 
-    vol_dir_clear(vol);
+    vol->clear_dir();
```
bneradt, line 91: Do we plan to rename this to `stripe` instead of `vol` later?

bneradt (approved):

src/iocore/cache/unit_tests/test_CacheDir.cc
masaori335: Yes. Renaming local variable will be in another PR. I sliced changes into multiple PRs, because it becomes over 1K lines changes.

PR #10741, opened 2023-11-07
Add auto options for maxmind_acl, stek_share, and uri_signing plugins

JosiahWI:
This helps clean up the dependency management and gives fine grained control over whether these plugins are built. I am planning to add all the experimental plugins, but I want to get these changes in quickly because we are still in the middle of fixing the plugins, and other PRs can build on these changes.

`BUILD_EXPERIMENTAL_PLUGINS` determines whether these plugins are `OFF` or `AUTO` on the first configuration. Changing the value of `BUILD_EXPERIMENTAL_PLUGINS` subsequently will not change whether those plugins are built, because their enabled/disabled state is cached separately. I have not come up with a good way to fix this. Suggestions welcome.


cmcfarlen (approved):
Looks good.  I like how the default follows BUILD_EXPERIMENTAL_PLUGINS

PR #10743, opened 2023-11-07
[9.2.x] Support upstream mptcp

serrislew:
Same as #10701 but for 9.2.x branch

### DIFF
--- a/iocore/net/Connection.cc
+++ b/iocore/net/Connection.cc
@@ -42,6 +42,7 @@
 // #define SEND_BUF_SIZE            (1024*64)
 #define FIRST_RANDOM_PORT 16000
 #define LAST_RANDOM_PORT 32000
+#define MPTCP_V1 2
 
 int
 get_listen_backlog()
@@ -229,7 +230,7 @@ Server::setup_fd_for_listen(bool non_blocking, const NetProcessor::AcceptOptions
 #endif
   }
 
-  if ((opt.sockopt_flags & NetVCOptions::SOCK_OPT_NO_DELAY) && !opt.f_mptcp &&
+  if ((opt.sockopt_flags & NetVCOptions::SOCK_OPT_NO_DELAY) && (opt.f_mptcp != MPTCP_V1) &&
       safe_setsockopt(fd, IPPROTO_TCP, TCP_NODELAY, SOCKOPT_ON, sizeof(int)) < 0) {
     goto Lerror;
   }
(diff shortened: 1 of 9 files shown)


serrislew:
[approve ci debian]

serrislew:
[approve ci clang-format]

serrislew:
[approve ci]

bneradt:
[approve ci]

iocore/net/Connection.cc
```diff
@@ -229,7 +229,7 @@ Server::setup_fd_for_listen(bool non_blocking, const NetProcessor::AcceptOptions
 #endif
   }
 
-  if ((opt.sockopt_flags & NetVCOptions::SOCK_OPT_NO_DELAY) &&
+  if ((opt.sockopt_flags & NetVCOptions::SOCK_OPT_NO_DELAY) && !opt.f_mptcp &&
       safe_setsockopt(fd, IPPROTO_TCP, TCP_NODELAY, SOCKOPT_ON, sizeof(int)) < 0) {
     goto Lerror;
   }
```
bneradt, line 235: Can you help me understand this change? I assume TCP_NODELAY is not applicable to MPTCP?

iocore/net/Connection.cc
```diff
@@ -390,6 +385,7 @@ Server::listen(bool non_blocking, const NetProcessor::AcceptOptions &opt)
     fd = NO_FD;
   }
 
-  Error("Could not bind or listen to port %d (error: %d)", ats_ip_port_host_order(&addr), res);
+  Fatal("Could not bind or listen to port %d, mptcp enabled: %d (error: %d) %s %d", ats_ip_port_host_order(&addr),
+        prot == IPPROTO_MPTCP, errno, strerror(errno), res);
```
bneradt, line 389: This makes sense. If you configured to listen on a port and can't, probably best to Fatal.

bneradt (approved):
Looks fine. I have a question, but the change is intentional enough that I trust you're doing the right thing with the socket options.

iocore/net/Connection.cc
serrislew: It seems like now that we are using MPTCP as a protocol and not a socket option, the socket options for IPPROTO_TCP doesn't apply for sockets that are not defined for IPPROTO_TCP

iocore/net/Connection.cc
serrislew: Oh this is Fatal on master but maybe it should be consistent with 9.2.x. Thoughts?

bneradt: Ah, gotcha. Yeah, probably fine to pull it in as fatal for 9.2.x too while we're at it.

zwoop:
@serrislew Does this remove the support for the old MPTCP stack? That's fine (and expected) on master, but we wouldn't want that on 9.2.x branch.

PR #10746, opened 2023-11-07
cmake-format target and pre-commit hook

bneradt:
This adds the cmake-format cmake target as well as updates the
pre-commit hook to check cmake-format on changed files.

This also has, as a separate commit, the cmake-format changes applied to our repo.


bneradt:
[approve ci]

cmcfarlen (approved):
Thank you!

PR #10749, opened 2023-11-08
Fix sni.yaml's ip_allow

bneradt:
The sni.yaml's version of ip_allow had an inverted conditional check. This adds a test for the feature and fixes the condition.

### DIFF
--- a/src/iocore/net/SNIActionPerformer.cc
+++ b/src/iocore/net/SNIActionPerformer.cc
@@ -346,7 +346,7 @@ TLSValidProtocols::SNIAction(SSL &ssl, const Context & /* ctx */) const
   return SSL_TLSEXT_ERR_OK;
 }
 
-SNI_IpAllow::SNI_IpAllow(std::string &ip_allow_list, std::string const &servername)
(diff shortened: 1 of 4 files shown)


SolidWallOfCode (approved):

src/iocore/net/SNIActionPerformer.cc
```diff
@@ -373,12 +373,12 @@ SNI_IpAllow::load(swoc::TextView content, swoc::TextView server_name)
...
     if (swoc::IPRange r; r.load(token)) {
-      Dbg(dbg_ctl_ssl_sni, "%.*s is not a valid format", static_cast<int>(token.size()), token.data());
-      break;
-    } else {
       Dbg(dbg_ctl_ssl_sni, "%.*s added to the ip_allow token %.*s", static_cast<int>(token.size()), token.data(),
           int(server_name.size()), server_name.data());
       ip_addrs.fill(r);
+    } else {
+      Dbg(dbg_ctl_ssl_sni, "%.*s is not a valid format", static_cast<int>(token.size()), token.data());
+      break;
```
bneradt, line 381: This is the functional change. It inverts the conditional because r.load returns true if the token is a valid IP format.

PR #10751, opened 2023-11-08
Cleanup: Remove unimplemented event handlers from CacheVC

masaori335:
These event handlers are declared but don't have implementations.


JosiahWI (approved):
Thanks for updating the comment as well!

ywkaras (approved):

bneradt (approved):

PR #10752, opened 2023-11-08
Add check for config existence before cont creation

### LINKED ISSUE #10658: stats_over_http always registers a config reload continuation, even without a config file
There's no reason, as far as I can tell, for this code if the plugin has not been given a YAML configuration (to reload).

```
  /* Create continuation for management updates to re-read config file */
  config_cont = TSContCreate(config_handler, TSMutexCreate());
  TSContDataSet(config_cont, (void *)config_holder);
  TSMgmtUpdateRegister(config_cont, PLUGIN_NAME);
```

The easy fix here is to just not call this if there's no config file. However, with #10642, we may also want to clean up the configuration reload, such that we let the core deal with the file timestamps etc. If the file name is given to TSMgmtUpdateRegister(), that continuation will only be called if the file has actually changed.

ezelkow1:
Checks if a config path is set to monitor for a config file, if not set dont create the mgmt cont

closes #10658

### DIFF
--- a/plugins/stats_over_http/stats_over_http.cc
+++ b/plugins/stats_over_http/stats_over_http.cc
@@ -737,9 +737,12 @@ init:
   TSHttpHookAdd(TS_HTTP_READ_REQUEST_HDR_HOOK, main_cont);
 
   /* Create continuation for management updates to re-read config file */
-  config_cont = TSContCreate(config_handler, TSMutexCreate());
-  TSContDataSet(config_cont, (void *)config_holder);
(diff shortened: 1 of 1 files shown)


JosiahWI:
HostDBProcessor regression test and regexcookie AuTest failed.

JosiahWI:
AuTests psubstitute and matchuri failed.

JosiahWI:
[approve ci autest]

ezelkow1:
[approve ci autest]

ywkaras (approved):

PR #10758, opened 2023-11-09
Move find_package() to root build file

dragon512:
This is basically clean up to make sure we can easily see what packages are needed as a whole in trafficserver
fmt as a library was also discussed as a general library to be used going forward

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -267,6 +267,12 @@ check_include_file(siginfo.h HAVE_SIGINFO_H)
 check_include_file(endian.h HAVE_ENDIAN_H)
 
 # Find libraries
+if(ENABLE_CRIPTS)
+  # needed for cripts
+  # however this might become a general requirement
+  find_package(fmt 8.1 REQUIRED)
+endif()
+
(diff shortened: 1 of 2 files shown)


JosiahWI (approved):

CMakeLists.txt
```diff
@@ -267,6 +267,13 @@ check_include_file(siginfo.h HAVE_SIGINFO_H)
 check_include_file(endian.h HAVE_ENDIAN_H)
 
 # Find libraries
+find_package(PkgConfig REQUIRED) # needed to find various libraries
```
JosiahWI, line 270: Could you help me understand why this was added? There is no corresponding removal of `find_package(PkgConfig REQUIRED)` from another location.

dragon512: hmm  This was just part of me moving find_package() to root.

I should pull that out to a different PR

JosiahWI (changes requested):
Please pull out the `PkgConfig` change to another PR.

JosiahWI (approved):

cmcfarlen (approved):

PR #10759, opened 2023-11-09
Fix capitalization of LuaJIT package and update

dragon512:
This addresses issues with the configure working as expected. It also addresses messages to tell us what is being used for lua

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,7 +102,7 @@ auto_option(
   HEADER_DEPENDS
   "sys/prctl.h"
 )
-auto_option(LUAJIT PACKAGE_DEPENDS LuaJIT)
+auto_option(LUAJIT PACKAGE_DEPENDS luajit)
 auto_option(UNWIND FEATURE_VAR TS_USE_REMOTE_UNWINDING PACKAGE_DEPENDS unwind)
 
 option(ENABLE_ASAN "Use address sanitizer (default OFF)")
--- a/cmake/Findluajit.cmake
+++ b/cmake/Findluajit.cmake
@@ -15,33 +15,34 @@
 #
 #######################
 
-# FindLuaJIT.cmake
+# Findluajit.cmake
 #
 # This will define the following variables
 #
-#     LuaJIT_FOUND
+#     luajit_FOUND
 #
 # and the following imported targets
 #
-#     LuaJIT::LuaJIT
+#     luajit::luajit
 #
 
-# LuaJIT exports their own config since LuaJIT-1.5.0, but it isn't
-# present in the OpenSUSE libLuaJIT-devel-1.7.1 package and maybe others.
-
 find_package(PkgConfig REQUIRED)
-pkg_check_modules(LuaJIT luajit)
+pkg_check_modules(luajit luajit)
 
 include(FindPackageHandleStandardArgs)
-find_package_handle_standard_args(LuaJIT REQUIRED_VARS LuaJIT_LIBRARIES LuaJIT_INCLUDE_DIRS)
+find_package_handle_standard_args(
+  luajit
+  REQUIRED_VARS luajit_INCLUDE_DIRS luajit_LINK_LIBRARIES luajit_LIBRARIES luajit_LIBRARY_DIRS
+  HANDLE_COMPONENTS
+)
 
-if(LuaJIT_FOUND)
-  set(LuaJIT_INCLUDE_DIRS ${LuaJIT_INCLUDE_DIR})
+if(luajit_FOUND)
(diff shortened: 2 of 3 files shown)


cmake/FindLuaJIT.cmake
```diff
@@ -15,35 +15,36 @@
...
+#     luajit_FOUND
 #
 # and the following imported targets
 #
-#     LuaJIT::LuaJIT
+#     luajit::luajit
 #
 
-# LuaJIT exports their own config since LuaJIT-1.5.0, but it isn't
-# present in the OpenSUSE libLuaJIT-devel-1.7.1 package and maybe others.
```
JosiahWI, line 30: I do not know where I got this information - it appears to be completely wrong. Might as well remove it entirely.

cmake/FindLuaJIT.cmake
```diff
@@ -15,35 +15,36 @@
...
 find_package(PkgConfig REQUIRED)
-pkg_check_modules(LuaJIT luajit)
+pkg_check_modules(luajit luajit)
 
 include(FindPackageHandleStandardArgs)
-find_package_handle_standard_args(LuaJIT
-    REQUIRED_VARS LuaJIT_LIBRARIES LuaJIT_INCLUDE_DIRS
+find_package_handle_standard_args(luajit
+    REQUIRED_VARS luajit_INCLUDE_DIRS luajit_LINK_LIBRARIES luajit_LIBRARIES luajit_LIBRARY_DIRS
+    HANDLE_COMPONENTS
```
JosiahWI, line 38: Does the LuaJIT from PkgConfig use any components?

cmake/FindLuaJIT.cmake
```diff
@@ -15,35 +15,36 @@
...
-#     LuaJIT::LuaJIT
+#     luajit::luajit
 #
 
-# LuaJIT exports their own config since LuaJIT-1.5.0, but it isn't
-# present in the OpenSUSE libLuaJIT-devel-1.7.1 package and maybe others.
+# luajit exports their own config since luajit-1.5.0, but it isn't
+# present in the OpenSUSE libluajit-devel-1.7.1 package and maybe others.
 
 find_package(PkgConfig REQUIRED)
```
JosiahWI, line 32: This can go in the root CMakeLists.txt.

JosiahWI (changes requested):

cmake/FindLuaJIT.cmake
dragon512: yes

JosiahWI: Probably good to have this then, thanks for adding it. 👍

dragon512:
> Please fix the case on the FindLuaJIT.cmake file (It should be Findluajit.cmake). Also, this adds a requirement of pkg-config being available on the system. We have not required pkg-config so far with cmake, so users that don't have that will have broken builds. I don't know if it was required for the autotools build. I think we could simplify finding reps if we also try png-config for other deps as well, so I'm fine adding it, but I think it should be included in the top cmakelists.txt file.
> 
> If we make that optional, then we can optionally use pkg-config to try to find deps and reduce the need to provide _ROOT variables for non-standard dep locations.

I fixed the case issues. I thought I had done it, but I must have messed up the git command. The original code had the pkg-config call. I only fixed up the code so that the pkg-config works as expected and to have it print out the message about what is found or not found for this library.

now we have a 
```-- Checking for module 'luajit'
--   Package 'luajit', required by 'virtual:world', not found
-- Could NOT find LuaJIT (missing: LuaJIT_LIBRARIES LuaJIT_INCLUDE_DIRS) 
```
when not found and message of which library that is being used ( which was not reported before)
```
-- Checking for module 'luajit'
--   Found luajit, version 2.1.0-beta3
-- Found luajit: /usr/include/luajit-2.1   
```

JosiahWI:
This already used PkgConfig prior to these changes.

JosiahWI:
Autotools used PkgConfig in some places, but maybe we should convert this Find module to follow the same format as the others. It is the only Find module we have that uses PkgConfig.

dragon512:
> Autotools used PkgConfig in some places, but maybe we should convert this Find module to follow the same format as the others. It is the only Find module we have that uses PkgConfig.

I think the issue with this library is how it names and layouts its outputs. It does some ugly stuff with the include path we don't want to have to manually support. The .pc file defines the extra include values so it just works.

cmcfarlen (approved):

dragon512:
[approve ci autest]

shukitchan (approved):

shukitchan:
I checked that out in the open source PR as well.

PR #10761, opened 2023-11-10
Fix sni.yaml ip_allow tunnel_route

bneradt:
When combining a rejection of an incoming connection via sni.yaml ip_allow and tunnel_route, ATS still initiated an unneccessary connection to the origin. This constituted an unused TLS handshake with the origin because it would immediately be closed after the client side handshake ended. This patch fixes ATS so it doesn't do anything on the origin side if the client side TLS handshake failed already at the CLIENT_HELLO.

---
# Note to Reviewer
Review this with the "ignore whitespace" option. The production patch is functionally a one line change.

### DIFF
--- a/src/iocore/net/SSLNetVConnection.cc
+++ b/src/iocore/net/SSLNetVConnection.cc
@@ -21,6 +21,7 @@
   limitations under the License.
  */
 
+#include "iocore/net/NetVConnection.h"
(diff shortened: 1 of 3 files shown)


bneradt:
[approve ci autest]

randall (approved):

PR #10762, opened 2023-11-10
Display auto_option hint in all cases

JosiahWI:
This takes the HINT added by @dragon512 a step further. The cases are as follows:
    - feature enabled, option enabled: hints how to disable
    - feature enabled, option disabled: no hint, something broke
    - feature disabled, option enabled (AUTO): hints how to require
    - feature disabled, option disabled: hints how to enable

I made a distinction between enabling and requiring. If the option is set to AUTO, it is already enabled, in some sense. If the user sets an option to ON, the user is requiring that the corresponding feature must be turned on, and should expect an error if a dependency is missing.

### DIFF
--- a/cmake/AutoOptionHelpers.cmake
+++ b/cmake/AutoOptionHelpers.cmake
@@ -132,7 +132,6 @@ endmacro()
 # Prints a colorized summary of one auto option.
 function(PRINT_AUTO_OPTION _NAME)
   string(ASCII 27 ESC)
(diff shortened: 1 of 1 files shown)


bneradt:
[approve ci clang-format]

dragon512 (approved):

SolidWallOfCode (approved):

bneradt:
[approve ci format]

Issue #9847, opened 2023-06-14
OpenSSL 3.0 MD5 deprecation warnings in WCCP


JosiahWI:
```
[12/426] Building CXX object src/wccp/CMakeFiles/wccp.dir/WccpMsg.cc.o
/trafficserver/src/wccp/WccpMsg.cc: In member function 'wccp::SecurityComp& wccp::SecurityComp::secure(const wccp::MsgBuffer&)':
/trafficserver/src/wccp/WccpMsg.cc:483:13: warning: 'int MD5_Init(MD5_CTX*)' is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  483 |     MD5_Init(&ctx);
      |     ~~~~~~~~^~~~~~
In file included from /trafficserver/src/wccp/WccpMsg.cc:25:
/usr/include/openssl/md5.h:49:27: note: declared here
   49 | OSSL_DEPRECATEDIN_3_0 int MD5_Init(MD5_CTX *c);
      |                           ^~~~~~~~
/trafficserver/src/wccp/WccpMsg.cc:484:15: warning: 'int MD5_Update(MD5_CTX*, const void*, size_t)' is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  484 |     MD5_Update(&ctx, key, KEY_SIZE);
      |     ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
/usr/include/openssl/md5.h:50:27: note: declared here
   50 | OSSL_DEPRECATEDIN_3_0 int MD5_Update(MD5_CTX *c, const void *data, size_t len);
      |                           ^~~~~~~~~~
/trafficserver/src/wccp/WccpMsg.cc:485:15: warning: 'int MD5_Update(MD5_CTX*, const void*, size_t)' is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  485 |     MD5_Update(&ctx, msg.getBase(), msg.getCount());
      |     ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/openssl/md5.h:50:27: note: declared here
   50 | OSSL_DEPRECATEDIN_3_0 int MD5_Update(MD5_CTX *c, const void *data, size_t len);
      |                           ^~~~~~~~~~
[...]

PR #10763, opened 2023-11-10
Removes all WCCP code and utilities, updates docs

zwoop:


JosiahWI:
Resolves #9847.

JosiahWI (changes requested):
There seem to be some fixes to the autest.sh script that are not directly related to the WCCP removal. Can we do these separately?

JosiahWI (commented):
What was the change to `jobs.yaml`? The whole file was reformatted and has a large diff.

zwoop:
> There seem to be some fixes to the autest.sh script that are not directly related to the WCCP removal. Can we do these separately?

It "fixes" the b0rken indentation :).

zwoop:
> What was the change to `jobs.yaml`? The whole file was reformatted and has a large diff.

Because we don't have a standard YAML format :). I fixed it.

randall (approved):

JosiahWI (approved):
Looks good. I could not find anything that was missed. I checked whether there were any other configuration definitions that we could remove, but I did not find any.

PR #10765, opened 2023-11-11
cmake: Fix CMAKE_INSTALL_CACHEDIR STRING value

hnakamur:
I suppose it is a leftover after copy-and-paste from `CMAKE_INSTALL_LOGDIR` above.


JosiahWI (approved):
Thank you for the fix!

masaori335 (approved):

PR #10769, opened 2023-11-13
remove odd include paths

dragon512:
This addresses some odd build paths I was finding in the build.

for example `-Iplugins/experimental/stek_share/nuraft::nuraft`

CMake seems to be adding the path correctly without adding the `target_include_directories()` call

### DIFF
--- a/plugins/experimental/stek_share/CMakeLists.txt
+++ b/plugins/experimental/stek_share/CMakeLists.txt
@@ -18,7 +18,6 @@
 if(nuraft_FOUND)
   add_atsplugin(stek_share common.cc log_store.cc stek_share.cc stek_utils.cc)
 
-  target_include_directories(stek_share PRIVATE nuraft::nuraft OpenSSL::SSL yaml-cpp::yaml-cpp)
   target_link_libraries(stek_share PRIVATE nuraft::nuraft OpenSSL::SSL yaml-cpp::yaml-cpp)
 else()
   message(STATUS "skipping stek_share plugin (missing nuraft)")
--- a/src/proxy/http/CMakeLists.txt
+++ b/src/proxy/http/CMakeLists.txt
@@ -45,8 +45,6 @@ if(BUILD_REGRESSION_TESTING)
   target_sources(http PRIVATE RegressionHttpTransact.cc)
 endif()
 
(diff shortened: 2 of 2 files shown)


plugins/experimental/stek_share/CMakeLists.txt
```diff
@@ -18,7 +18,7 @@
 if(nuraft_FOUND)
   add_atsplugin(stek_share common.cc log_store.cc stek_share.cc stek_utils.cc)
 
-  target_include_directories(stek_share PRIVATE nuraft::nuraft OpenSSL::SSL yaml-cpp::yaml-cpp)
+  target_include_directories(stek_share PRIVATE yaml-cpp::yaml-cpp)
```
JosiahWI, line 21: This is also incorrect. Targets should not be used as arguments to `target_include_directories`. It might need to be `${YAML_INCLUDE_DIR}`.

JosiahWI (changes requested):

plugins/experimental/stek_share/CMakeLists.txt
dragon512: libyaml is broken independent of this fix.

I will have a different PR for libyaml

JosiahWI (approved):

src/proxy/http/CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ if(BUILD_REGRESSION_TESTING)
   target_sources(http PRIVATE RegressionHttpTransact.cc)
 endif()
 
-target_include_directories(http PRIVATE ${YAMLCPP_INCLUDE_DIR} OpenSSL::SSL)
+target_include_directories(http PRIVATE ${YAMLCPP_INCLUDE_DIR})
```
cmcfarlen, line 48: This line can be removed as well.  http doesn't include yaml-cpp and if it did then it should be listed as a dependency below.

cmcfarlen (changes requested):
remove the other target_include_directories line as well

src/proxy/http/CMakeLists.txt
dragon512: done

cmcfarlen:
[approve ci osx]

cmcfarlen (approved):

PR #10771, opened 2023-11-13
Remove unused EventName.[h|cc]

lzx404243:
As titled. Don't see any usage of these.


bneradt (commented):
At first I thought maybe it should be used, but then, looking at the implementation, this wouldn't help much. :) It would take some work to make it complete.

I approved, but will let a non-Yahoo person give an official approval.

randall (approved):

bneradt:
[approve ci format]

PR #10774, opened 2023-11-14
tscore depends on tscpputil

cmcfarlen:
Since `tscore/DiagsType.h` includes `tscpp/util/Regex.h`, `tscore` needs a PUBLIC dependency on `tscpputil`.


JosiahWI (approved):
Straightforward fix. Thanks for the description.

SolidWallOfCode (approved):

PR #10776, opened 2023-11-14
cmake: Update pre-commit even if it exists

cmcfarlen:
This will generate the pre-commit file even if it exists.  This will allow cmake to always update the pre-commit hook if it changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -697,9 +697,13 @@ add_custom_target(
   COMMENT "formatting all files"
 )
 
-if(IS_DIRECTORY ${CMAKE_SOURCE_DIR}/.git AND NOT EXISTS ${CMAKE_SOURCE_DIR}/.git/hooks/pre-commit)
-  message(STATUS "Installing github hook")
+if(IS_DIRECTORY ${CMAKE_SOURCE_DIR}/.git)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -697,9 +697,13 @@ add_custom_target(
   COMMENT "formatting all files"
 )
 
-if(IS_DIRECTORY ${CMAKE_SOURCE_DIR}/.git AND NOT EXISTS ${CMAKE_SOURCE_DIR}/.git/hooks/pre-commit)
-  message(STATUS "Installing github hook")
+if(IS_DIRECTORY ${CMAKE_SOURCE_DIR}/.git)
+  file(TIMESTAMP ${CMAKE_SOURCE_DIR}/.git/hooks/pre-commit PRE_COMMIT_BEFORE)
   configure_file(${CMAKE_SOURCE_DIR}/tools/git/pre-commit ${CMAKE_SOURCE_DIR}/.git/hooks/pre-commit COPYONLY)
```
JosiahWI, line 702: Does `configure_file(... COPYONLY)` not always update the timestamp?

cmcfarlen: Only if the source is different from the destination

JosiahWI (approved):
Thanks for the fix.

bneradt (approved):

bneradt:
[approve ci format]

PR #10782, opened 2023-11-14
Remove HAVE_ALLOCA_H build var

phongn:

This PR does so for `alloca.h`

### DIFF
--- a/include/tscore/ink_platform.h
+++ b/include/tscore/ink_platform.h
@@ -122,7 +122,7 @@ struct ifafilt;
 #ifdef HAVE_VALUES_H
 #include <values.h>
 #endif
-#ifdef HAVE_ALLOCA_H
+#if __has_include(<alloca.h>)
 #include <alloca.h>
 #endif
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -215,7 +215,6 @@ include(CheckTypeSize)
 include(CheckSourceCompiles)
 include(CheckStructHasMember)
 
-check_include_file(alloca.h HAVE_ALLOCA_H)
 check_include_file(cpio.h HAVE_CPIO_H)
 check_include_file(dlfcn.h HAVE_DLFCN_H)
 check_include_file(float.h HAVE_FLOAT_H)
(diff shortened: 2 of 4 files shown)


JosiahWI (approved):

bneradt:
[approve ci]

bneradt:
[approve ci]

include/tscore/ink_platform.h
```diff
@@ -122,7 +122,7 @@ struct ifafilt;
 #ifdef HAVE_VALUES_H
 #include <values.h>
 #endif
-#ifdef HAVE_ALLOCA_H
+#if __has_include(<alloca.h>)
```
dragon512, line 125: So the question is ... does this code compile if there is no alloca.h? If not there is no point in testing for the header at all.

dragon512 (changes requested):

bneradt:
[approve ci osx]

include/tscore/ink_platform.h
maskit: Looks like FreeBSD has `alloca()` in "stdlib.h", which is included above, and it doesn't have "alloca.h". Linux has `alloca()` in "alloca.h". So I suppose the intention is "If you build this on a system that has alloca.h, you need it".

phongn: I've been reading up on man pages on recent Linux, FreeBSD and macOS: 

Linux: 


> Normally, `gcc(1)` translates calls to `alloca()` with inlined code.  This is not done when either the -`ansi`, `-std=c89`, `-std=c99`, or the `-std=c11` option is given and the header `<alloca.h>` is not included.  Otherwise, (without an `-ansi` or `-std=c*` option) the glibc version  of  `<stdlib.h>` includes `<alloca.h>` and that contains the lines:
> 
> `#ifdef  __GNUC__`
> `#define alloca(size)   __builtin_alloca (size)`
> `#endif`
>
> with messy consequences if one has a private version of this function.
>
> The fact that the code is inlined means that it is impossible to take the address of this function, or to change its behavior by linking with a different library.
>
> The inlined code often consists of a single instruction adjusting the stack pointer, and does not check for stack overflow. Thus, there is no NULL error return.

FreeBSD: 

> `#include	<stdlib.h>`

macOS: 

> `#include <alloca.h>` or
> `#include <stdlib.h>`

phongn: I think in this case I would prefer to explicitly include `alloca.h` if available (which it is on recent macOS and Linux)

maskit: Yeah, that sounds reasonable to me.

dragon512 (approved):

maskit (approved):

PR #10783, opened 2023-11-14
Remove HAVE_CPIO_H build var

phongn:
It appears that `HAVE_CPIO_H` was used to undefine `cpio.h`'s `MAGIC` global variable. While examining our code, it appears that we also declare `MAGIC` but namespace it properly, meaning that it is unlikely that we will see a conflict. Compiling this code shows no shadowing warnings. 

No code in TS directly includes `cpio.h`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -216,7 +216,6 @@ include(CheckTypeSize)
 include(CheckSourceCompiles)
 include(CheckStructHasMember)
 
-check_include_file(cpio.h HAVE_CPIO_H)
 check_include_file(dlfcn.h HAVE_DLFCN_H)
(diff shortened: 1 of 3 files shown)


JosiahWI (approved):

bneradt:
[approve ci]

bneradt:
[approve ci]

maskit:
The second commit makes sense. CPIO does not seem to be used anywhere. We should be able to remove it.

phongn:
> The second commit makes sense. CPIO does not seem to be used anywhere. We should be able to remove it.

Yes, I further investigated the original use of this and it appears unnecessary. See updated PR commit notes.

maskit (approved):

PR #10785, opened 2023-11-15
Remove HAVE_STDLIB_H build var

phongn:

This PR does so for `stdlib.h` (which should always exist as a standard header, and thus all guards are removed).


bneradt:
[approve ci]

JosiahWI (approved):

bneradt:
[approve ci]

include/tscore/ink_platform.h
```diff
@@ -60,7 +60,7 @@
 struct ifafilt;
 #include <net/if.h>
 
-#ifdef HAVE_STDLIB_H
+#ifdef __has_include(<stdlib.h>)
```
JosiahWI, line 63:
```suggestion
#if __has_include(<stdlib.h>)
```

JosiahWI (changes requested):

JosiahWI (approved):

include/tscore/ink_platform.h
```diff
@@ -60,7 +60,7 @@
 struct ifafilt;
 #include <net/if.h>
 
-#ifdef HAVE_STDLIB_H
+#if __has_include(<stdlib.h>)
```
dragon512, line 63: When would we ever not have stdlib.h ... it is a standard C header
We should remove the check for this

dragon512 (changes requested):

include/tscore/ink_platform.h
phongn: I agree.

maskit: Yeah, I agree with @dragon512 on this one . I can't think of any system that doesn't need stdlib.h. But at the same time, this is to remove an unnecessary check in CMakeLists.txt. If there is any doubt I'm ok with this change.

PR #10786, opened 2023-11-15
Obliterate Stripe::reset_agg_buf_pos

JosiahWI:
There was only one use of `Stripe::reset_agg_buf_pos`, and this gets rid of it by moving the code to flush the aggregate buffer from CacheDir.cc into Stripe.cc and AggregateWriteBuffer.cc.

I also removed two `#define`s from P_CacheVol.h that were supposed to be moved to AggregateWriteBuffer.h, but got left in due to a rebase mistake and were duplicated in both places.

### DIFF
--- a/src/iocore/cache/Stripe.cc
+++ b/src/iocore/cache/Stripe.cc
@@ -32,6 +32,7 @@ namespace
 {
 
 DbgCtl dbg_ctl_cache_init{"cache_init"};
+DbgCtl dbg_ctl_cache_dir_sync{"dir_sync"};
 
 // This is the oldest version number that is still usable.
 short int const CACHE_DB_MAJOR_VERSION_COMPATIBLE = 21;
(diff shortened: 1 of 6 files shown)


src/iocore/cache/Stripe.cc
```diff
@@ -917,3 +918,23 @@ Stripe::_init_data()
   this->_init_data_internal();
   this->_init_data_internal();
 }
+
+bool
+Stripe::flush_aggregate_write_buffer()
+{
+  Dbg(dbg_ctl_cache_dir_sync, "Dir %s: flushing agg buffer first", this->hash_text.get());
```
JosiahWI, line 925: I wonder whether this debug message should be updated now that the context has changed.

masaori335: I agree with adjusting message.

bneradt:
[approve ci autest]

bneradt:
[approve ci autest]

src/iocore/cache/Stripe.cc
JosiahWI: I looked at the code again and opted to move the message back to `sync_cache_dir_on_shutdown`, because the message goes along with the other debug messages in that function, and I think it would be good to keep it.

masaori335: That makes sense 👍

masaori335 (approved):
Looks good.

PR #10792, opened 2023-11-15
Remove code for old openssl

maskit:
This removes code for old OpenSSL (< 1.1.1) which relies on `OPENSSL_VERSION_NUMBER`.

Code that relies on function availability checks remains (e.g. `TS_HAS_TLS_EARLY_DATA`).  It's not easy because we need to check the availability on BoringSSL, in addition to the availability on OpenSSL >=1.1.1. And we don't have the minimum supported "version" for BoringSSL.


bneradt (approved):
It's nice to see this code simplified.

PR #10793, opened 2023-11-15
stek_share: resize instead of reserve

bneradt:
stek_share indexes into an offset on a string that it reserved space for, but indexing past size() is undefined behavior. This would result in crashing on some systems. reszing adjusts the length which satisfies the indexing requirement.


duke8253 (approved):
+1 

Still need someone outside of Yahoo to approve though

randall (approved):

PR #10795, opened 2023-11-16
Cleanup Cript includes, and metrics

zwoop:
This changes Cripts to use the new Metrics, keeping it's convenience wrapper and storage class (for instances etc.).

We could technically use the lower level Metrics.h directly, but this is more inline with the goal of simplicity with Cripts.


zwoop:
[approve ci autest]

bneradt (approved):

zwoop:
@bneradt Need another +1 please, after the merge conflict fixes.

randall (approved):

PR #10796, opened 2023-11-16
Add check in ts.h that language is C++, version 17 or later.

ywkaras:


JosiahWI:
This is also be enforced by CMake, but I guess it doesn't hurt to have a check.

ywkaras:
> This is also be enforced by CMake, but I guess it doesn't hurt to have a check in the source code.

It's possible someone might not want to use cmake to build their plugin.

JosiahWI:
> > This is also be enforced by CMake, but I guess it doesn't hurt to have a check in the source code.
> 
> It's possible someone might not want to use cmake to build their plugin.

Aha, this change makes sense now!

JosiahWI (approved):

maskit:
Would it cause errors if a plugin is built with C++ 11 where ATS itself is built with C++ 17?

We probably want to keep our right to use C++ 17 in ts.h, but I don't think we need to force plugin developers to use C++ 17 until we add something that requires C++ 17.

ywkaras:
Do C++11 and C++17 have the same runtime library?  Even if we are OK with the memory usage of having both loaded, will they have duplicate identifiers?  Are there ABI issues?

@zwoop, do you want to change your mind about this?

https://www.open-std.org/JTC1/SC22/WG21/docs/papers/2018/p1319r0.html#removed

https://www.open-std.org/JTC1/SC22/WG21/docs/papers/2018/p0636r3.html#removed

dragon512 (approved):
GIven we we need c++ 17 or better. I think this is a wise option to be ABI working

SolidWallOfCode (approved):

PR #10799, opened 2023-11-16
cmake: Add -fPIC when checking for symbols

bneradt:
I noticed that on ubuntu:23:04 with ci-ubuntu, the symbol checks for getresuid and the like were failing because the compilation of the simple programs to check the symbols failed with complaints saying that the code had to be compiled with PIC enabled. This caused cmake to incorrectly think that these symbols didn't exist when in fact they did. Adding -fPIC for the symbol checks fixed this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -373,6 +373,7 @@ include(ConfigureTransparentProxy)
 include(find_ccache)
 
 # Check for IO faculties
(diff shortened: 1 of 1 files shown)


bneradt:
[approve ci ubuntu]

cmcfarlen (approved):
This should be ok.  Its kinda odd that the test program require PIC, but other executables do not.

bneradt:
[approve ci centos]

PR #10801, opened 2023-11-16
Add missing Openssl depends

dragon512:
The only item here I am concerned about if if you pull in your own libfmt, by default it does not build a .so, only a static .a. This will lot link as it is not built with -fPic. I am not sure if there is a way to tell this to use a shared library for the fmt library, Other than that this adds the missing depends on OpenSSL.


dragon512:
we have https://github.com/apache/trafficserver/pull/10800 that includes this fix and more.. might want to close this one if we that the other PR from @cmcfarlen

dragon512:
closing this PR as the other PR was merged

PR #10803, opened 2023-11-16
Remove HAVE_FLOAT_H build var

phongn:

This PR does so for `float.h` (which should always exist as a standard header, and thus all guards are removed).


cmcfarlen (changes requested):
Can you try to just remove including float.h?  I don't think its necessary

phongn:
> Can you try to just remove including float.h? I don't think its necessary

`src/iocore/cache/HttpTransactCache.cc` and `src/records/RecUtils.cc` rely on `float.h` via `include/tscore/ink_platform.h`

cmcfarlen (approved):

PR #10804, opened 2023-11-16
Fix compiler warning on m1 macs printing time_t

bryancall:


masaori335 (approved):
Looks good.

src/iocore/hostdb/P_RefCountCache.h
```diff
@@ -236,7 +236,7 @@ RefCountCachePartition<C>::put(uint64_t key, C *item, int size, time_t expire_ti
 
   // add expiry_entry to expiry queue, if the expire time is positive (otherwise it means don't expire)
   if (expire_time >= 0) {
-    Dbg(dbg_ctl, "partition %d adding entry with expire_time=%" PRId64, this->part_num, expire_time);
+    Dbg(dbg_ctl, "partition %d adding entry with expire_time=%" PRIdMAX, this->part_num, expire_time);
```
ywkaras, line 239: The more portable file for this would be:
```
Dbg(dbg_ctl, "partition %d adding entry with expire_time=%" PRIdMAX, this->part_num, intmax_t(expire_time));
```
It time_t and the bus/word width is N bits, and intmax_t is 2*N bits, then using expire_time without the cast would mean only one word would get pushed on the stack (or put in a register if that were the ABI).  But the format would cause the printf-like function to expect two words.

PR #10807, opened 2023-11-16
Make dubiously correct changes to Japanese docs for TSDebug -> Dbg.

ywkaras:


masaori335 (approved):
This change is fine.

I'd point out there're ton of outdated msgids like this. We need to update po files periodically.

PR #10810, opened 2023-11-17
For vscode users, automatically use ms-vscode.cmake-tools

phongn:
As we move to CMake by default, perhaps it would be useful for VSCode users to automatically set the configuration provider to `ms-vscode.cmake-tools`? 

@dragon512


dragon512:
I don't see an issue with this. @zwoop do you have any concerns?

zwoop (approved):
Looks good.

zwoop:
Actually, there's a missing carriage return at the end of the file, can you add that please?

PR #10815, opened 2023-11-17
Remove HAVE_NETINET_*_H and HAVE_NETDB_H build vars

phongn:

This PR does so for `netinet/*.h` and `netdb.h` (which should always exist as UNIX standard internet headers, and thus all guards are removed). 

In addition, it reorganizes and renames the CMake build environment variables for detecting various features and flavors of `tcp_info`.


dragon512:
[approve ci osx]

cmcfarlen (changes requested):
There is a check for `HAVE_STRUCT_TCP_INFO` on line 1296 in `plugins/header_rewrite/conditions.cc` that I think needs to be `HAVE_STRUCT_LINUX_TCP_INFO`

phongn:
> There is a check for `HAVE_STRUCT_TCP_INFO` on line 1296 in `plugins/header_rewrite/conditions.cc` that I think needs to be `HAVE_STRUCT_LINUX_TCP_INFO`

Right now that block of code currently does a (¬FreeBSD ∨ glibc) check for the Linux-style `tcp_info`, and then devolves to the FreeBSD-style `tcp_info` by default. I've made this more explicit in the latest commit to using the CMake `HAVE_STRUCT_LINUX_TCP_INFO` define for the first case and adding a new `HAVE_STRUCT_FREEBSD_TCP_INFO` in the second case. 

In doing some research, there's also a macOS `tcp_connection_info` struct that behaves similarly but that is a future project to implement (and at any rate I am not aware of anyone using macOS in production, just for development).

phongn:
There are a few checks that look like this: 

```C
#if defined(TCP_INFO) && defined(HAVE_STRUCT_TCP_INFO)
```

Is this redundant? The existence of one should imply the other, unless someone is doing crazy things (and then we should have a redefinition preprocessor warning). If we are willing to accept preprocessor protection, we could remove the `HAVE_STRUCT_TCP_INFO` CMake build env. 

cc: @dragon512  @cmcfarlen

PR #10821, opened 2023-11-17
Add missing stdint.h to ink_file.h

phongn:
While testing some code on Linux and with GCC 13, I discovered that `include/tscore/ink_file.h` declared a `uint64_t` field but did not include `stdint.h`, which is now against strict conformance. 

This PR adds it in.

### DIFF
--- a/include/tscore/ink_file.h
+++ b/include/tscore/ink_file.h
@@ -36,6 +36,7 @@
 #include <cstdio>
 #include <sys/types.h>
 #include <dirent.h>
+#include <stdint.h>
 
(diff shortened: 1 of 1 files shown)


traeak (approved):

phongn:
> Isn't cstdint better to use than stdint.h ? Okay cstdint isn't guaranteed to put the types in the global namespace so that's fine.

It would be better to do so and the C++ specification requires it. However, all of the present code assumes use of the global namespace, and changing to `cstdint` and the accompanying `std::` namespace is a broader conversation for this project.

PR #10825, opened 2023-11-18
ConnectionTracker: Remove queue implementation

bneradt:
Connection count queueing functionality was removed via #7302. This removes some dead code that was still associated with that now-removed feature.

### DIFF
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -5884,9 +5884,6 @@ HttpSM::handle_post_failure()
 void
 HttpSM::handle_http_server_open()
 {
-  // The request is now not queued. This is important because server retries reuse the t_state.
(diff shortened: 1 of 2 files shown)


src/proxy/http/HttpSM.cc
```diff
@@ -5884,9 +5884,6 @@ HttpSM::handle_post_failure()
 void
 HttpSM::handle_http_server_open()
 {
-  // The request is now not queued. This is important because server retries reuse the t_state.
-  t_state.outbound_conn_track_state.dequeue();
-
```
bneradt, line 5889: This might seem like a concerning removal of a use of dequeue, but note that enqueue was never called. Thus this call was always a no-op.

SolidWallOfCode (approved):

PR #10827, opened 2023-11-18
Remove Stripe::add_agg_todo

JosiahWI:
I could not find a way to do this without moving any logic around. I changed the logic as minimally as possible and split it into 3 commits as follows:
- The first commit reorders some of the logic to split it apart and prepare part of it to be extracted to a new method.
- The second commit does the extraction.
- The third commit cleans up to remove duplication and get rid of `Stripe::add_agg_todo`.

The logic change was quite limited and straightforward, so I don't think this one is very risky, but please review carefully anyway. One oddity here is `CACHE_AGG_FAIL_RATE`. Is this define still used by anyone for testing? Should we be testing with it?

### DIFF
--- a/src/iocore/cache/CacheWrite.cc
+++ b/src/iocore/cache/CacheWrite.cc
@@ -262,26 +262,9 @@ CacheVC::handleWrite(int event, Event * /* e ATS_UNUSED */)
     return handleEvent(AIO_EVENT_DONE, nullptr);
   }
 
+  // Make sure the size is correct for checking error conditions before calling add_writer(this).
   agg_len = stripe->round_to_approx_size(write_len + header_len + frag_len + sizeof(Doc));
-  stripe->add_agg_todo(agg_len);
-  bool agg_error = (agg_len > AGG_SIZE || header_len + sizeof(Doc) > MAX_FRAG_SIZE ||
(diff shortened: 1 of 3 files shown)


masaori335:
> One oddity here is CACHE_AGG_FAIL_RATE. Is this define still used by anyone for testing? Should we be testing with it?

At least, our CI is not using this. However, this is a test for error handling, so it'd be great if you test with enabling it. 
Eventually, we need to make it a build option or change it with a more sophisticated approach.

JosiahWI:
> > One oddity here is CACHE_AGG_FAIL_RATE. Is this define still used by anyone for testing? Should we be testing with it?
> 
> At least, our CI is not using this. However, this is a test for error handling, so it'd be great if you test with enabling it. Eventually, we need to make it a build option or change it with a more sophisticated approach.

The unit tests and the level 1 and 3 regression tests pass with it enabled. I don't think I have a powerful enough machine to run all the AuTests in a reasonable timeframe.

JosiahWI:
I observed a failure in the `cache-request-method` AuTest, which went away on subsequent runs, so it seems to be due to `CACHE_AGG_FAIL_RATE` being turned on. I don't think this is unexpected, and it would confirm that it did simulate a failure.

masaori335 (approved):
This change looks a bit complicated in total, but if I follow the three commits one-by-one, each commit is pretty simple cleanups.

PR #10831, opened 2023-11-21
Pull out down origin check from HttpSM::apply_ip_allow_filter

bneradt:
In a refactor on master for the ip_allow implementation in HttpSM (see #9845) put into HttpSM::apply_ip_allow_filter. However, the down origin feature is orthogonal to ip_allow and it is confusing to implement its feature in a function that communicates that it implements ip_allow filtering. This separates the down origin check from that function like it was before the refactor.


JosiahWI (approved):
Straightforward change. It shouldn't be less efficient, because the calling method returns early if the ip allow check fails.

I like the addition of the debug message.

bneradt:
[approve ci]

PR #10832, opened 2023-11-22
TS_USE_MMAP: mmap+memcopy instead pread+pwrite

monkuta:
Allows for only-in-memory caching for usecases where such behavior is needed (i.e. Live Video caching). It maps storage directly into user space memory, so reads and writes are being done by a memcpy (without calls to io).

apt install libaio-dev
autoreconf -if
./configure --enable-mmap --enable-experimental-linux-native-aio
./configure --enable-mmap

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1704,6 +1704,15 @@ AS_IF([test "x$enable_linux_native_aio" = "xyes"], [
 AC_MSG_RESULT([$enable_linux_native_aio])
 TS_ARG_ENABLE_VAR([use], [linux_native_aio])
 
+AC_MSG_CHECKING([enable mmap instead read/write])
+AC_ARG_ENABLE([mmap],
+  [AS_HELP_STRING([--enable-mmap], [WARNING this is experimental mmap instead read and write support @<:@default=no@:>@])],
+  [enable_mmap="${enableval}"],
+  [enable_mmap=no]
+)
+AC_MSG_RESULT([$enable_mmap])
+TS_ARG_ENABLE_VAR([use], [mmap])
+
 # Check for hwloc library.
 # If we don't find it, disable checking for header.
 use_hwloc=0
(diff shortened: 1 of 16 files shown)


bryancall:
Do you have a benchmark comparing the performance of mmap vs pread?

New features would go into the master branch and we most likely wouldn't backport this change into 9.x at this point in time because we are focusing on releasing ATS 10.

moleksy:
@bryancall 
In a Sapphire Rapids system with 100Gbps NIC we observed 10x lower mean latency (2.7ms vs 28ms), also maximum obtainable transfer rate was higher by 22% (9.4GBps bs 7.7GBps).
Test was done using vegeta, live video 1080p simulation (1MB objects), 30% cold cache, 70% cached, total volume of 14GB of cached objects.

moleksy:
@SolidWallOfCode 
Hello Alan, would you like us to provide additional information so that the PR can be approved?
Best Regards
Maciej

moleksy:
@bryancall Would you be able to review this PR?

zwoop (changes requested):
Wait, why is this against the 9.2.x branch? Is this a back port of something from Master?

masaori335:
@moleksy Our development flow is a). merge in to the master branch first, and b). backport commits to the latest branch if needed

Please open a PR against the master branch. (or change the target branch of this PR) You might need to adjust your autotools config into CMake because we're in the middle of building tool migration.

> In a Sapphire Rapids system with 100Gbps NIC we observed 10x lower mean latency (2.7ms vs 28ms), also maximum obtainable transfer rate was higher by 22% (9.4GBps bs 7.7GBps).

We also need more benchmark with this, but the number you talked is pretty interesting.

JosiahWI:
@monkuta has made the PR against the master branch as requested.

PR #10840, opened 2023-11-22
Core implementation of ip_category for ip_allow.yaml

bneradt:
Adds the ip_category feature for ip_allow.yaml so that users can specify arbitrary IP descriptions like the following:

```yaml
  ip_allow:
    - apply: in
      ip_category: ACME_INTERNAL
      action: allow
      methods:
        - GET
        - HEAD
        - POST
        - PUSH
        - DELETE
```

Whether a given IP belongs to ACME_INTERNAL is then indicated by a custom plugin using the TS_HTTP_IP_ALLOW_CATEGORY_HOOK callback via the TSHttpIpAllowInfoContainsSet API.

---

This is a new feature. I'll leave this as a draft now until I receive feedback from an email I am sending to the dev mailing alias. See email with subject **New feature suggestion: ip_category for ip_allow.yaml**

TODO:

- [ ] Receive feedback from the dev mailing list.
- [ ] Add test coverage.
- [ ] Verify the docs render appropriately.
- [ ] Verify the hook state diagram. Maybe ip_allow functions on remap instead of where I put it in accept?


example/plugins/c-api/ip_category/ip_category.cc
```diff
@@ -0,0 +1,122 @@
...
+
+  return TS_SUCCESS;
+}
+
+} // anonymous namespace
+
+void
+TSPluginInit(int argc, const char *argv[])
+{
+  auto cont = TSContCreate(ip_category_callback, nullptr);
```
ywkaras, line 120: Does the callback for this hook need to be thread-safe?  Is the IPAllow match function called in each accept thread without mutual exclusion?

bneradt: There's nothing from the core-side requiring mutual exclusion - the core simply calls the handler with a thread-local variable, so there's no critical region to serialize execution around.

bneradt:
After feedback, I'm working on a different implementation of IP categories.

PR #10841, opened 2023-11-22
Coverity 1518588: Out-of-bounds read in healthchecks plugin

bryancall:


plugins/healthchecks/healthchecks.cc
```diff
@@ -213,10 +213,10 @@ hc_thread(void *data ATS_UNUSED)
     }
 
     if (len >= 0) {
-      int i = 0;
+      uint32_t i = 0;
 
       /* coverity[ -tainted_data] */
-      while (i < len) {
+      while (i <= (len - sizeof(struct inotify_event))) {
```
maskit, line 219: What should we do for the remaining data? It might be garbage and we can discard, but it might be incomplete next event.

bryancall: Maybe we just log if `len % sizeof(struct inotify_event) != 0` and ignore the events?

Currently we are reading over the buffer and have an incomplete next event.

maskit: `len % sizeof(struct inotify_event) != 0`  wouldn't work because it the `len` includes the length of event. `i` is updated like `i += sizeof(struct inotify_event) + event->len;`.

We should probably keep the remaining data for next loop and concatenate it with the data from next read.

PR #10852, opened 2023-11-28
Move FetchSM out of tsapi to break cycle

lzx404243:
Currently FetchSM is in tsapi which is dependent on inknet. However,  it's also used by OCSPStapling(inknet), thus creating a dependency cycle(which IIUC is a blocker for the tsapi.so work). 

This PR cleans this up:
* Eliminated the usage of public API in FetchSM
* Moved FetchSM to src/proxy


maskit (approved):
This is probably a good first step to cleanup FetchSM and the dependency around it.

We have to deal with the new dependency that is from `iocore/net/OCSPStapling` to `proxy/FetchSM` later. I'm not too sure what's the best structure, but I feel like we should decouple the stapling part and the fetching part, and move the fetching part into `proxy/` or `traffic_server/`.

PR #10856, opened 2023-11-28
Fix netcat usage for accept_timeout test

shinrich:
Adjust to use the reduced argument list for netcat that is available in the ubuntu 22 image.

### DIFF
--- a/tests/gold_tests/timeout/create_request.sh
+++ b/tests/gold_tests/timeout/create_request.sh
@@ -16,5 +16,20 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-(printf "GET / HTTP/1.1" |  nc -w 11 127.0.0.1 $1)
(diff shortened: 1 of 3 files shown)


tests/gold_tests/timeout/create_request.sh
```diff
@@ -16,5 +16,42 @@
...
+# so falling back to shell
+
+# Send part of the request and then stop
+request() {
+	printf "GET / HTTP/1.1"
+}
+
+nc_name="nc"
+ncat --help
+if [ $? -le 1 ]
```
bneradt, line 29: I'm surprised by `-le 1` rather than `-eq 0`. Do some `ncat` commands return `1` with `--help`? I assume checking for `1` is intentional, but just verifying.

Related to this, we might want to consider using the built-in `command -v ncat` instead of `ncat --help`.

ywkaras:
[approve ci]

PR #10859, opened 2023-11-28
add CUSTOM_LINKER to CMakeLists.txt

traeak:
Adds CUSTOM_LINKER option so that bfd/lld/gold/mold may be selected.


cmcfarlen:
This is the kind of config that IMO we should NOT have in the cmakelists.txt.  What is provided will very often not be enough to get this to work (you will need to add paths and other linker flags as well).  CMake has some guidelines for compiler config that falls outside of the common things that cmake has cross-compiler support for (i.e. build type, warnings as errors) which is basically let the user set flags and variable externally.  There is also a full feature (toolchains) that is more appropriate for this sort of setup.

PR #10860, opened 2023-11-28
yapf: use yapf instead of autopep8

bneradt:
autopep8 does not support Python 3.12, which is a requirement for moving
to fedora:39. The following ticket has been open without comment from
the development community for 6 weeks:
    
https://github.com/hhatto/autopep8/issues/712
    
Transitioning to yapf is easy, and the community behind it seems
stronger. I think moving forward it is a better Python formatter for us.

---

# For Review

Note that this review is split up into two commits:

* The first updates the formatting scripts to use yapf.
* The second has the yapf updated format for the Python files.


bneradt:
[approve ci]

bneradt:
[approve ci]

.style.yapf
```diff
@@ -0,0 +1,421 @@
...
+# would reformat to:
+#
+#    call_func_that_takes_a_dict({
+#        'key1': 'value1',
+#        'key2': 'value2',
+#    })
+coalesce_brackets=False
+
+# The column limit.
+column_limit=99
```
masaori335, line 104: Assuming this is limit of line length, our autopep8 setting was 132. I hope using `132` will reduce style change made by this migration.
https://github.com/apache/trafficserver/blob/40961fbf69d33d9a19e8e0be706101efef4a3a7b/tools/autopep8.sh#L91

bneradt: I initially tried 132 locally, but I think the diff is actually larger if we set the line length to that. In reality, if you look at our .py tests before this patch, the community has kind of implicitly respected about a 80-100 character limit while autpep8 didn't "expand" things to 132 characters. yapf will actively expand function parameters and arguments and the like to 132 if we set that size. I'm not opposed to setting 132, I'm just stating it didn't seem like using 132 reduced the patch size but rather increased it. I can switch things to 132 if we like.

The reason the patch is so large is because of the three big Python formatters (yapf, autopep8, black), autopep8 is the least "opinionated". yapf enforces much more style over autopep8. It enforces things at the level more akin to clang-format, for instance. I'm not sure that's bad, but it is how it is.

I guess the big question is whether we think the changes are a net benefit or not. I feel like we kind of need to move off of autopep8 because I've lost confidence in the project - 6 weeks after an issue was filed about Python 3.12 and there's not even so much as a comment on the issue from any maintainers. Not even when I explicitly (and politely) asked for status with an @ mention of the repo author have they replied after 8 days.

I intend, if I have time, to create a separate patch seeing how things look with the black formatter so we can compare. yapf has a pretty strong community which is a big plus, so I'll first investigate who and how much black is supported before trying them. Like all formatters, the goal is to find something "acceptable". "Beyond awesome" is probably not attainable. 

Anyway, thanks for looking at the patch. Feel free to provide other thoughts. Or even to try things out locally, if you like. It's pretty easy to experiment by using my first to commits in this PR.

masaori335: Fair enough, let's go with `99`, if we want, we can do it later. I agree with we should move away from autopep8 now. Also, `yapf` might not be the best formatter, but as far as it works like clang-format, it's acceptable.

Thanks for working on this 👍 Please resolve conflicts.

bneradt: I changed things to a 132 line limit and it looks better than I remembered. I think things look OK.

zwoop (approved):

PR #10823, opened 2023-11-17
Remove HAVE_SYS_*_H build envs

phongn:
Replace automake-style `HAVE_INCLUDE_H` compile flags with `#if __has_include` preprocessor directives.

This PR does so for `sys/*.h` and does minor cleanup for various system specific build flags.

PR #10863, opened 2023-11-29
Cleanup ink_thread.h of wrong BSD checks assuming missing pthread functions

phongn:
Commit https://github.com/apache/trafficserver/commit/26c408071d6a559bfb30da8a67854a4e888cfb71 attempted to replace some FreeBSD pthread function checks with function checking. This was reverted in https://github.com/apache/trafficserver/commit/012df52981084d3cd82d1e8aa87074683dd22594 as feature detection should be done via autoconf (or now, CMake). 

However, every remotely recent version of FreeBSD has the pthread functions in question. There seems to be no reason to bother to check for `pthread_cancel` or `pthread_getschedparam`. 

In doing this, this allows us to remove `HAVE_SYS_CAPABILITY_H` and `HAVE_SYS_PRCTL_H`, replacing them with `__has_include` directives. We also introduce a direct `HAVE_PRCTL` feature check to replace the implied existence of it in `HAVE_SYS_PRCTL_H`. 

In addition, the merge of #10823 into #11029 dropped a fix in which `HAVE_PTHREAD_MUTEXATTR_SETTYPE` was removed but not what called it. `pthread_mutexattr_settype` is available on every modern operating system supporting pthread; we should no longer need to check for its existence.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -234,12 +234,6 @@ if(brotli_FOUND)
   set(HAVE_BROTLI_ENCODE_H TRUE)
 endif()
 
-if(TS_USE_POSIX_CAP)
-  # These headers must have been found for capabilites to be enabled.
-  set(HAVE_SYS_CAPABILITY_H TRUE)
-  set(HAVE_SYS_PRCTL_H TRUE)
-endif()
-
 find_package(LibLZMA)
 if(LibLZMA_FOUND)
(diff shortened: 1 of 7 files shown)


cmcfarlen (dismissed):

ywkaras:
[approve ci]

phongn:
[approve ci]

cmcfarlen (approved):

PR #10864, opened 2023-11-29
Replace nonstandard darwin platform detection

phongn:
We do not a few feature checks for operating systems using deprecated flags. This replaces the nonstandard `darwin` with `__APPLE__` ⋀ `__MACH__`.

A future PR will do so for FreeBSD and OpenBSD.

### DIFF
--- a/src/iocore/net/test_certlookup.cc
+++ b/src/iocore/net/test_certlookup.cc
@@ -234,7 +234,7 @@ main(int argc, const char **argv)
   ink_freelists_dump(stdout);
 
 // On Darwin, fail the tests if we have any memory leaks.
-#if defined(darwin)
+#if (defined(__APPLE__) && defined(__MACH__))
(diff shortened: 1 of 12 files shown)


src/iocore/net/test_certlookup.cc
```diff
@@ -234,7 +234,7 @@ main(int argc, const char **argv)
   ink_freelists_dump(stdout);
 
 // On Darwin, fail the tests if we have any memory leaks.
```
randall, line 236: Maybe change the deferences to Darwin in comments to macOS

phongn: I thought about it, but it's not actually wrong either (and `uname` returns `Darwin` on macOS). It's really just the preprocessor directives that should be changed. Also, a number of places look for the `Darwin` string as well for various purposes.

cmcfarlen:
We do define `darwin`, is that not working? Why do we need this?

PR #10865, opened 2023-11-29
Update links in records.yaml to be valid

jasmine-nahrain:
Currently most of the links in records.yaml do not work and lead nowhere. Updating them to the new url structure.


masaori335:
[approve ci]

ywkaras (approved):

PR #10866, opened 2023-11-29
Remove netcat from bigobj test

shinrich:
Can use standard Http clients to exercise the PUSH method.  Removed netcat usage.


tests/gold_tests/bigobj/bigobj.test.py
```diff
@@ -52,65 +47,88 @@
...
+    f.write("x" * obj_bytes)
+    f.close()
+    return True
+
 
 tr = Test.AddTestRun("PUSH an object to the cache")
 # Delay on readiness of TS IPv4 ssl port
-tr.Processes.Default.StartBefore(ts)
-#
+tr.Processes.Default.StartBefore(ts, ready=lambda: create_pushfile())
```
ywkaras, line 75: Does this augment or replace the default ready test?

ywkaras:
One of the goals of this test was, you could test multi-MB when you wanted to:  https://github.com/apache/trafficserver/pull/10866/files#diff-70d5d4c78bcc7e78c3f88898d7a8ab453f657979a39bc5720d630b6c7970a718L23 .  There was a bug that was triggered by this test, when the object was large enough, as I recall.

tests/gold_tests/bigobj/bigobj.test.py
```diff
@@ -52,65 +47,88 @@
...
 #
 obj_kilobytes = 10 * 1024
+obj_bytes = obj_kilobytes * 10
+header = "HTTP/1.1 200 OK\r\nContent-length: {}\r\n\r\n".format(obj_bytes)
+
+
+def create_pushfile():
+    f = open(Test.RunDirectory + "/objfile", "w")
+    f.write(header)
+    f.write("x" * obj_bytes)
```
ywkaras, line 68: I think there' s some value in having a ramping pattern, rather than a constant value for all bytes in the file.  For example, if somehow ATS swapped the order of two IOBuffers.

tests/gold_tests/bigobj/bigobj.test.py
```diff
@@ -52,65 +47,88 @@
...
+
 
 tr = Test.AddTestRun("PUSH an object to the cache")
 # Delay on readiness of TS IPv4 ssl port
-tr.Processes.Default.StartBefore(ts)
-#
+tr.Processes.Default.StartBefore(ts, ready=lambda: create_pushfile())
 # Put object with URL http://localhost/bigobj in cache using PUSH request.
-tr.Processes.Default.Command = (
-    f'./push_request {obj_kilobytes} | nc localhost {ts.Variables.port}'
```
ywkaras, line 75: This runs netcat with no arguments.  But there's still some incompatibility with Ubuntu retro netcat?

ywkaras:
Whenever I run any Au test on my Ubuntu laptop, there is is a very (minutes) long pause at start up.  Are you seeing that?

tests/gold_tests/bigobj/bigobj.test.py
shinrich: This augments the default port ready test as far as I can tell @dragon512 can you verify that?  It looks like there is an obj ready test which I assume includes the port ready for ts as well as the explicitly specified ready function.

tests/gold_tests/bigobj/bigobj.test.py
shinrich: Seems reasonable

ywkaras:
It looks like, in general `Test.Setup.Copy()` in Au tests does not work on Ubuntu, for example:  `Test.Setup.Copy(os.path.join(Test.Variables.AtsBuildGoldTestsDir, 'tls', 'ssl-post'))`.

tests/gold_tests/bigobj/bigobj.test.py
shinrich: Since curl does support PUSH, that seems more reliable and maintainable than relying on netcat. 

The specific failure I was seeing on u22 was with the last testcase
> file /home/ubuntu/ats10/build-mydev/tests/_sandbox/bigobj/_output/6-tr-Default/stream.stdout.txt : The PUSH request should have received a 403 response. - ^[[0m^[[31mFailed^[[0m^[[1m
          Reason: Contents of /home/ubuntu/ats10/build-mydev/tests/_sandbox/bigobj/_output/6-tr-Default/stream.stdout.txt did not contains expression: "403 Access Denied"

Specifically the stream.stdout.txt file was empty.  I assume there is a connection close difference or something between the distros.  Rather than debugging that, since nc isn't really needed, I felt it would be easier to deal with by using curl instead.

ywkaras:
> It looks like, in general `Test.Setup.Copy()` in Au tests does not work on Ubuntu, for example: `Test.Setup.Copy(os.path.join(Test.Variables.AtsBuildGoldTestsDir, 'tls', 'ssl-post'))`.

Nevermind this, I forgot the `--build-root` option on autest.sh.

tests/gold_tests/bigobj/bigobj.test.py
ywkaras: I only get this far:
```
   Run: GET bigobj: TLS, HTTP/2, IPv6: Failed
     Starting TestRun 4-tr : No Issues found - Passed
        Reason: Started!
     Process: Default: Failed
          Reason: Returned Value 1 != 0
```
The failure is here:  https://github.com/apache/trafficserver/blob/870e242eee6d382750ad383abde03e416314bba0/tests/gold_tests/bigobj/check_ramp.c#L41

I'll approve this, but I feel like we're whistling past the graveyard a bit.  Ideally, our white box regression testing should be robust on all supported platforms.

ywkaras (approved):

PR #10867, opened 2023-11-29
Remove POSIX_THREAD constants and fix unmigrated HAVE_PTHREAD_MUTEXATTR_SETTYPE

phongn:
We define two constants, `POSIX_THREAD` and `POSIX_THREAD_10031c` that appear to be old legacy code and predates the project move to git. They are essentially hardcoded at this point. This PR removes them. 

Per conversation with @cmcfarlen, we remove the old code that sets `PTHREAD_PROCESS_SHARED` on mutexes as it has been dead code since at least ATS v2.0.1.

This also fixes an autoconf-to-CMake bug wherein `HAVE_PTHREAD_MUTEXATTR_SETPSHARED` was not migrated over. This bug only occurs on debugging to set `PTHREAD_MUTEX_ERRORCHECK ` on a mutex and likely has been unnnoticed. 

**Please carefully review**, especially `src/tscore/ink_mutex.cc`.


src/tscore/ink_mutex.cc
```diff
@@ -35,7 +35,7 @@ class x_pthread_mutexattr_t
   x_pthread_mutexattr_t()
   {
     pthread_mutexattr_init(&attr);
-#ifndef POSIX_THREAD_10031c
+#if HAVE_PTHREAD_MUTEXATTR_SETPSHARED
     pthread_mutexattr_setpshared(&attr, PTHREAD_PROCESS_SHARED);
```
phongn, line 39: I am not sure if this is the correct change here. The old code essentially disabled this call and it dates to the original SVN import. 

Should it continue to be ignored, and if so, should it be deleted?

phongn: Continued investigation of the old SVN import indicates that this dates to at least v2.0.1(!) and SVN r831143 (the initial open-source import) and that `POSIX_THREAD_10031c` was defined on a Linux, FreeBSD or Darwin system.

cmcfarlen: I do not think we should be setting mutexes pshared at all.  I don't think it matters anyways unless the mutexes are in shared memory regions.  I think we could just delete this code.

cmcfarlen (approved):

ywkaras:
[approve ci]

PR #10868, opened 2023-11-29
Remove unnecessary nc to fix u22 test

shinrich:
Replaced use of nc with curl.  Without that change, the following test case would not emit any data when run on ubuntu22.


masaori335 (approved):
Looks good.

PR #10869, opened 2023-11-29
Fix connect test for ubuntu22

shinrich:
Slight output difference between the curl on redhat and the curl on u22.  u22 prints (nil) if there is no hostname.  It appears that the redhat curl prints the ip address gain.  Adding wildcard to avoid the difference in the gold file.


masaori335 (approved):
Looks reasonable.

Issue #10052, opened 2023-07-17
RPCClient: Avoid having such big memory allocation on every command.


brbzull0:
I had to increase the buffer size of the RPC Client read socket data. 
https://github.com/apache/trafficserver/pull/9970

This issues is to avoid having such a big buffer and make it more dynamic.

PR #10870, opened 2023-11-29
traffic_ctl: Remove the hardcoded big buffer and use a stream buffer instead.

brbzull0:
This is only for `traffic_ctl`

The approach is to use a simple stream buffer instead of the big allocation. In the past we had to create a super buffer as a workaround to allocate big responses from the JSONRPC server,  this pr is made tackle that issue so on small responses this will just use the fixed buffer(100x smaller than before), if response contains more than the fixed buffer then it will save the data into a stringstream and reuse the already allocated buffer.

Fixes https://github.com/apache/trafficserver/issues/10052


brbzull0:
[approve ci autest]

zwoop (approved):

PR #10872, opened 2023-11-29
tools: Update build script(h3) to have the option to build quiche with either openssl/quictls or boringssl vendor.

brbzull0:
this should do the trick.
```
$ ./build_h3_tools.sh quictls
```
By default it will build quiche+boringssl.

Also, removing the `recursive` flag from quiche clone as it's not needed.

https://github.com/apache/trafficserver/issues/10873


brbzull0:
https://github.com/apache/trafficserver/pull/11228

PR #10874, opened 2023-11-29
Fix double to watch for done like the other continuation tests

shinrich:
The other three tests in the continuations area do the wait for done this way.  The original shell script would not work on u22 since /bin/sh is linked to dash instead of bash.

### DIFF
--- a/tests/gold_tests/continuations/double.test.py
+++ b/tests/gold_tests/continuations/double.test.py
@@ -104,7 +104,7 @@ watcher = Test.Processes.Process("watcher", "sleep 20")
 
 tr = Test.AddTestRun("Wait for stats")
 tr.Processes.Default.StartBefore(watcher, ready=make_done_stat_ready(ts.Env))
-tr.Processes.Default.Command = "traffic_ctl metric get continuations_verify.test.done",
+tr.Processes.Default.Command = "traffic_ctl metric get continuations_verify.test.done"
 tr.Processes.Default.ReturnCode = 0
 tr.Processes.Default.Env = ts.Env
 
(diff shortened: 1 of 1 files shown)


tests/gold_tests/continuations/double.test.py
```diff
@@ -81,26 +81,30 @@
...
+        return retval.returncode == 0 and b'1' in retval.stdout
+
+    return done_stat_ready
+
+
+watcher = Test.Processes.Process("watcher", "sleep 20")
+
 # number of sessions/transactions opened and closed are equal
 tr = Test.AddTestRun("Check Ssn")
+tr.Processes.Default.StartBefore(watcher, ready=make_done_stat_ready(ts.Env))
```
ywkaras, line 107: If this ever failed, it would probably be less confusing in a separate test run with a dummy echo command.

ywkaras:
You could also consider using `tests/tools/stdout_wait` .

bneradt (dismissed):

bneradt:
> You could also consider using `tests/tools/stdout_wait` .

Either way. @shinrich: if you want to use stdout_await, let me know when you push and I can re-review.

tests/gold_tests/continuations/double.test.py
```diff
@@ -81,24 +81,33 @@
...
+        return retval.returncode == 0 and b'1' in retval.stdout
+
+    return done_stat_ready
+
+
+watcher = Test.Processes.Process("watcher", "sleep 20")
+
+tr = Test.AddTestRun("Wait for stats")
+tr.Processes.Default.StartBefore(watcher, ready=make_done_stat_ready(ts.Env))
+tr.Processes.Default.Command = "traffic_ctl metric get continuations_verify.test.done"
```
ywkaras, line 107: Seems like this could just be `echo waits for stats done`.

shinrich: It could.  Doesn't really matter what the command is.  Does it?

PR #10875, opened 2023-11-29
Adjust gold file to deal curl case differences

shinrich:
Evidently the versions of curl differ between ubuntu22 and redhat.


ywkaras (approved):

ywkaras:
I think, long term, tests like these should be rewritten to use Proxy Verifier.

shinrich:
> I think, long term, tests like these should be rewritten to use Proxy Verifier.

Agreed.

PR #10876, opened 2023-11-29
Coverity 1528709: Pointer to local outside scope in PreWarmSM::state_dns_lookup

bryancall:
The scope of `srv_hostname` and `hostname` needs to be the same since `hostname` can refer to `srv_hostname`.


masaori335 (approved):
Looks reasonable. Thanks.

PR #10878, opened 2023-11-29
Adjust nc usage in fix-post-early-return test

shinrich:
Fixing post-early-return test to use the base options of netcat so it will work on ubuntu22 as well.


shinrich:
> LGTM (but I'm in the kill the use of nc as much of possible camp).

Agreed, but when you are playing with timings, nc is often the easiest choice. The alternative is writing your own client or server.

PR #10882, opened 2023-11-29
Coverity 1497456: Time of check time of use in LogFile::trim_rolled

bryancall:
Instead of calling `access` and `opendir` separately using the filename, I `open` the directory and then use the file descriptor to call `faccessat` and `fdopendir`.

### DIFF
--- a/src/proxy/logging/LogFile.cc
+++ b/src/proxy/logging/LogFile.cc
@@ -281,7 +281,7 @@ LogFile::trim_rolled(size_t rolling_max_count)
   ats_free(name);
 
   // Open the directory
-  int dirfd = ::open(logfile_dir.c_str(), O_RDONLY);
+  int dirfd = open(logfile_dir.c_str(), O_RDONLY);
   if (dirfd < 0) {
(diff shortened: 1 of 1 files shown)


src/proxy/logging/LogFile.cc
```diff
@@ -275,31 +275,37 @@ struct RolledFile {
...
   if (err < 0) {
-    Error("Error accessing logging directory %s: %s.", logfile_dir.c_str(), strerror(errno));
+    Error("Error accessing logging directory %s: %s", logfile_dir.c_str(), strerror(errno));
     return false;
   }
 
-  /* Open logging directory */
-  DIR *ld = ::opendir(logfile_dir.c_str());
+  // Open the logging directory
+  DIR *ld = ::fdopendir(dirfd);
```
cmcfarlen, line 302: If faccessat or fdopendir fails and we return from this function, I think we need to close `dirfd`.  If fdopendir succeeds then I believe the fd will close when closedir is called.

cmcfarlen (changes requested):
I think we need to close dirfd in some cases

PR #10885, opened 2023-11-29
Fix remap_inc test for systems with dash

shinrich:
Removed the non-basic shell code test. Will not run with /bin/sh == dash on ubuntu 22.  Replaced the check with content checks on the output.

### DIFF
--- a/tests/gold_tests/traffic_ctl/remap_inc/remap_inc.test.py
+++ b/tests/gold_tests/traffic_ctl/remap_inc/remap_inc.test.py
@@ -74,8 +74,8 @@ tr.Processes.Default.ReturnCode = 0
 tr.StillRunningAfter = ts
 
 tr = Test.AddTestRun("Get response from generator")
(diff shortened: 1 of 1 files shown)


tests/gold_tests/traffic_ctl/remap_inc/remap_inc.test.py
```diff
@@ -74,8 +74,8 @@
 tr.StillRunningAfter = ts
 
 tr = Test.AddTestRun("Get response from generator")
-tr.Processes.Default.Command = (
-    f'test $$(curl --proxy 127.0.0.1:{ts.Variables.port} http://example.four/nocache/5 | wc -c) == 5'
```
ywkaras, line 78: This also works:
```
    f"bash -c 'test $$(curl --proxy 127.0.0.1:{ts.Variables.port} http://example.four/nocache/5 | wc -c) == 5'"
```

ywkaras (approved):

PR #10888, opened 2023-11-30
Add new config option to support CUBIC

### LINKED ISSUE #9743: Support CUBIC and compare with current impl
- Add a new setting for QUIC congestion control algorithm
- Set the setting value by calling `quiche_config_set_cc_algorithm`

Performance comparison is optional.

duke8253:
this resolves #9743

### DIFF
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -4785,6 +4785,17 @@ removed in the future without prior notice.
 
    Disables HTTP/0.9 over QUIC by default.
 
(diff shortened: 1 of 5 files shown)


doc/admin-guide/files/records.yaml.en.rst
```diff
@@ -4785,6 +4785,17 @@ removed in the future without prior notice.
 
    Disables HTTP/0.9 over QUIC by default.
 
+.. ts:cv:: CONFIG proxy.config.quic.quiche_cc_algorithm INT 0
```
maskit, line 4788: I don't think we should have "quiche" in the setting name. We wouldn't want "quic.openssl_cc_althorithm" in addition to this.

PR #10889, opened 2023-11-30
Remove dependency of apicore on tscore

cmcfarlen:
This moves code related to diagnostics, errata and buffer writing to src/api.  I tried to move as little as possible.


src/api/CMakeLists.txt
```diff
@@ -52,12 +51,28 @@ target_link_libraries(
...
+add_library(tsapicore SHARED Assert.cc Metrics.cc DbgCtl.cc SourceLocation.cc ts_diags.cc)
 add_library(ts::tsapicore ALIAS tsapicore)
-target_link_libraries(tsapicore PRIVATE ts::tscore)
-set_target_properties(tsapicore PROPERTIES POSITION_INDEPENDENT_CODE TRUE)
+set_target_properties(tsapicore PROPERTIES POSITION_INDEPENDENT_CODE TRUE PUBLIC_HEADER "${TSAPICORE_PUBLIC_HEADERS}")
+target_link_libraries(tsapicore PUBLIC libswoc)
 
-install(TARGETS tsapi tsapicore tsapibackend PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/ts)
+install(TARGETS tsapi tsapibackend PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/ts)
+install(TARGETS tsapicore PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/api)
```
bneradt, line 75: Just to make sure I understand this, this is the big change which is the goal of this patch, right? `tsapicore` no longer lists `ts::tscore` as a dependency, so the circular dependency is removed.

bneradt (approved):

src/api/CMakeLists.txt
cmcfarlen: Yes, that is goal!

Skn0tt:
I believe `tsapicore` also needs to be added in here:

https://github.com/apache/trafficserver/blob/78ccc61c9c803c39b3a874588fe2b3b2d52e9a0d/tests/fuzzing/CMakeLists.txt#L31-L37

I'll try to validate that locally, and open a PR with a fix.

Skn0tt:
Turned out that shared libs need to be copied into a special `/lib` directory for fuzz tests. Fix is in https://github.com/apache/trafficserver/pull/10930.

PR #10891, opened 2023-11-30
Coverity 1497333: Resource leak in object in LogConfig

bryancall:
hostname wasn't freed and minor cleanup


masaori335 (approved):
Looks good.

PR #10899, opened 2023-12-06
remove ts_bwf.cc and update locations where its used

cmcfarlen:
This remove tscpp/util/ts_bwf.cc and moves the used functionality to Diags.cc

The way that ts_bwf.cc registers globals does not play well with static libraries and was causing problems removing a circular dependency.  Also, this functionality (afaict) is only used in tscore/Diags.cc, so rather than carry this complexity for small benefit, I have factored this file out.

### DIFF
--- a/src/tscore/Diags.cc
+++ b/src/tscore/Diags.cc
@@ -214,6 +214,9 @@ Diags::~Diags()
   deactivate_all(DiagsTagType_Action);
 }
 
+namespace
+{
+
 struct DiagTimestamp {
   std::chrono::time_point<std::chrono::system_clock> ts = std::chrono::system_clock::now();
 };
@@ -244,6 +247,8 @@ bwformat(swoc::BufferWriter &w, swoc::bwf::Spec const &spec, DiagThreadname cons
   bwformat(w, spec, std::string_view{n.name});
   return w;
 }
+
+} // namespace
 
 //////////////////////////////////////////////////////////////////////////////
 //


src/tscore/Diags.cc
```diff
@@ -213,6 +214,37 @@ Diags::~Diags()
...
+  DiagThreadname() { ink_get_thread_name(name, sizeof(name)); }
+};
+
+swoc::BufferWriter &
+bwformat(swoc::BufferWriter &w, swoc::bwf::Spec const &spec, DiagThreadname const &n)
+{
+  bwformat(w, spec, std::string_view{n.name});
+  return w;
+}
+
```
ywkaras, line 247: Should all these added definitions be in an anonymous namespace?

cmcfarlen: adding to an anonymous namespace hides them from consideration as an overload.

ywkaras: Are you using the Swift compiler again?
```
wkaras ~/REPOS/TS
O$ git status
On branch cleanup-ts-bwf
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   src/tscore/Diags.cc

no changes added to commit (use "git add" and/or "git commit -a")
wkaras ~/REPOS/TS
O$ git diff
diff --git a/src/tscore/Diags.cc b/src/tscore/Diags.cc
index 0092f9d9d..d1f870326 100644
--- a/src/tscore/Diags.cc
+++ b/src/tscore/Diags.cc
@@ -214,6 +214,9 @@ Diags::~Diags()
   deactivate_all(DiagsTagType_Action);
 }
 
+namespace
+{
+
 struct DiagTimestamp {
   std::chrono::time_point<std::chrono::system_clock> ts = std::chrono::system_clock::now();
 };
@@ -245,6 +248,8 @@ bwformat(swoc::BufferWriter &w, swoc::bwf::Spec const &spec, DiagThreadname cons
   return w;
 }
 
+} // end anonymous namespace
+
 //////////////////////////////////////////////////////////////////////////////
 //
 //      void Diags::print_va(...)
wkaras ~/REPOS/TS
O$ cmake --build build
[0/2] Re-checking globbed directories...
[54/54] Linking CXX executable plugins/experimental/uri_signing/unit_tests/uri_signing_test
wkaras ~/REPOS/TS
O$ 
```

ywkaras (approved):

PR #10902, opened 2023-12-07
Coverity 1523647: No virtual destructor

maskit:
`HttpBodySet` inherits `HttpBodySetRawData` and we have `HttpBodySet` has a destructor but `HttpBodySetRawData` doesn't have a virtual destructor.


maskit:
[approve ci freebsd]

bryancall (approved):

PR #10914, opened 2023-12-08
Coverity 1497395: Out-of-bounds access

maskit:


include/tscore/ink_inet.h
```diff
@@ -742,6 +742,8 @@ ats_ip_copy(sockaddr *dst,      ///< Destination object.
   }
   if (n) {
     if (src != dst) {
+      // It can access beyond sizeof(sockaddr) but the real size is checked above
+      // coverity[access_dbuff_const]
       memcpy(dst, src, n);
```
bryancall, line 747: The problem is with this  function is you don't know how large the caller allocated for sockaddr.  Here are the sizes of the different sockaddrs..

sizeof(sockaddr) = 16
sizeof(sockaddr_in) = 16
sizeof(sockaddr_in6) = 28

maskit: And we check the actual size above, right? Should we not trust `src->sa_family` ?

PR #10923, opened 2023-12-11
Fix cache.volume_N.stripes metric issue by cleaning up CacheProcessor::cacheInitialized

### LINKED ISSUE #10921: `proxy.process.cache.volume_N.stripes` are not incremented if `ram_cache.size` is auto
In the cache initialization sequence, `vol_rsb.stripes` is incremented, but this happens only if `cache_config_ram_cache_size != AUTO_SIZE_RAM_CACHE`.

https://github.com/apache/trafficserver/blob/c1832e43bc0c57aa6b8a53590c48fc2709763012/src/iocore/cache/Cache.cc#L717

- e.g.

```
» traffic_ctl metric match stripe
proxy.process.cache.stripes 2
proxy.process.cache.volume_1.stripes 0
proxy.process.cache.volume_2.stripes 0
```

masaori335:
Alternative approach of #10922 to fix #10921.

The root cause is `CacheProcessor::cacheInitialized` has ton of duplicated code for `AUTO_SIZE_RAM_CACHE` case and configured case. This cleanup remove these duplicated code and increment metrics for both cases.

### DIFF
--- a/src/iocore/cache/Cache.cc
+++ b/src/iocore/cache/Cache.cc
@@ -576,9 +576,13 @@ CacheProcessor::diskInitialized()
 void
 CacheProcessor::cacheInitialized()
 {
-  int i;
+  if (theCache == nullptr) {
+    Dbg(dbg_ctl_cache_init, "theCache is nullptr");
+    return;
+  }
 
-  if (theCache && (theCache->ready == CACHE_INITIALIZING)) {
+  if (theCache->ready == CACHE_INITIALIZING) {
+    Dbg(dbg_ctl_cache_init, "theCache is initializing");
     return;
   }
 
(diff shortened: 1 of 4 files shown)


src/iocore/cache/Cache.cc
```diff
@@ -645,45 +634,22 @@ CacheProcessor::cacheInitialized()
...
+
       // let us calculate the Size
       if (cache_config_ram_cache_size == AUTO_SIZE_RAM_CACHE) {
         Dbg(dbg_ctl_cache_init, "CacheProcessor::cacheInitialized - cache_config_ram_cache_size == AUTO_SIZE_RAM_CACHE");
-        for (i = 0; i < gnvol; i++) {
-          stripe = gvol[i];
-
-          if (gvol[i]->cache_vol->ramcache_enabled) {
-            gvol[i]->ram_cache->init(stripe->dirlen() * DEFAULT_RAM_CACHE_MULTIPLIER, stripe);
-            ram_cache_bytes += gvol[i]->dirlen();
```
masaori335, line 656: This is another bug, this should be multiplied by `DEFAULT_RAM_CACHE_MULTIPLIER`.

masaori335:
Rebased on the latest master and add AuTest.

bneradt (approved):

PR #10925, opened 2023-12-12
Use slice req method type in Data obj

serrislew:
Set `m_method_type` in Data instead of Config since the former is allocated per transaction and the latter is allocated for a single remap instance so the data is shared across transactions.`m_method_type` is specific to a transaction not config definition


traeak (dismissed):
Good fix, my apologies for not seeing this in a previous PR.

serrislew:
[approve ci autest]

traeak (approved):

bryancall:
Cherry picked into 9.2.x

PR #10927, opened 2023-12-12
Change volatile to atomic in EThread.h.

ywkaras:


include/iocore/eventsystem/EThread.h
```diff
@@ -479,7 +481,7 @@ class EThread : public Thread
     /// The slices.
     std::array<Slice, N_SLICES> _slice;
 
-    Slice *volatile current_slice = nullptr; ///< The current slice.
```
masaori335, line 482: Just curious about what race `current_slice` has. It looks like it's just pointing to one of the `_slice` array and only access from this EThread.

ywkaras:
Ooops pushed to Apache repo.

include/iocore/eventsystem/EThread.h
ywkaras: It is also accessed from Metrics::record_loop_time() and from Metrics::summarize().  I had to make a new PR because I accidentally pushed my branch to the main repo, not my fork.

masaori335: Ah, `EThread::Metrics::summarize` is called from `EventMetricStatSync`. Got it, thanks.

PR #10930, opened 2023-12-13
Copy tsapicore library into /lib

Skn0tt:
Follow-up to https://github.com/apache/trafficserver/pull/10889. Running fuzz tests throws `error while loading shared libraries: libtsapicore.so: cannot open shared object file: No such file or director`. This is because `tsapicore` is a shared library now, and it needs to be copied into the `/lib` directory of fuzz tests.

This should fix the fuzzing build (broken since yesterday, take a look at [today](https://storage.googleapis.com/oss-fuzz-coverage/trafficserver/reports/20231213/linux/src/trafficserver/tests/fuzzing/report.html) vs [yesterday](https://storage.googleapis.com/oss-fuzz-coverage/trafficserver/reports/20231212/linux/src/trafficserver/tests/fuzzing/report.html))

### DIFF
--- a/tests/fuzzing/CMakeLists.txt
+++ b/tests/fuzzing/CMakeLists.txt
@@ -49,5 +49,6 @@ add_custom_command(
   POST_BUILD
   COMMAND ${CMAKE_COMMAND} -E copy_if_different $<TARGET_FILE:libswoc> ${CMAKE_CURRENT_BINARY_DIR}/lib/
   COMMAND ${CMAKE_COMMAND} -E copy_if_different $<TARGET_FILE:ts::tsapi> ${CMAKE_CURRENT_BINARY_DIR}/lib/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different $<TARGET_FILE:ts::tsapicore> ${CMAKE_CURRENT_BINARY_DIR}/lib/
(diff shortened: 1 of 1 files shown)


Skn0tt:
Are fuzz tests being built as part of the CI? It might make sense to compile + try running them, so these kinds of regressions can be caught before merging.

cmcfarlen:
tsapicore will be merged with tscpputil to form tsutil in this pr: #10928 

The tsutil library will follow the BUILD_SHARED_LIBS variable to determine if it is build static or shared (defaulting to static).  So you might have to check that variable to determine if you need to do this strange copy.  It looks like I already fixed this in the PR.  I'm not sure how the fuzz testing it run or why you need to copy these libraries.  Could you point me to some info on that?

Skn0tt:
Thanks for pointing that out! I'm also new to the fuzzing tests here. This talks about how to run them locally, but it's a bit involved: https://google.github.io/oss-fuzz/getting-started/new-project-guide/#testing-locally

I can also try running the fuzz tests on https://github.com/apache/trafficserver/pull/10928 in the coming days, and can double-check if that fixes it.

shukitchan:
Yeah. Perhaps we should wait till after #10928 is done and update this pull request accordingly.

Issue #9488, opened 2023-03-03
Set DF bit to UDP datagrams for QUIC


maskit:
> UDP datagrams MUST NOT be fragmented at the IP layer. In IPv4 [[IPv4](https://datatracker.ietf.org/doc/html/rfc791)], the Don't Fragment (DF) bit MUST be set if possible, to prevent fragmentation on the path.[¶](https://datatracker.ietf.org/doc/html/rfc9000#section-14-7)

ywkaras:
Can you give a reference explaining why this is necessary?  When IP fragments a UDP packet, does it reassemble it, or pass the fragments through to the application?

PR #10931, opened 2023-12-13
UDP: Set  the "Don't Fragment" (DF) bit  at IP layer for QUIC.

brbzull0:
Did some tests and looks it's being set. 
```
Flags: 0x4000, Don't fragment
    0... .... .... .... = Reserved bit: Not set
    .1.. .... .... .... = Don't fragment: Set
    ..0. .... .... .... = More fragments: Not set

```
Likely the linux kernel will set it too, but in this case it's being forced by us.

Note:
As per man page:
```
IP_PMTUDISC_WANT           Use per-route settings.
IP_PMTUDISC_DONT           Never do Path MTU Discovery.
IP_PMTUDISC_DO             Always do Path MTU Discovery.
IP_PMTUDISC_PROBE          Set DF but ignore Path MTU.
```
wondering if `IP_PMTUDISC_DO` is what we  should use instead and not ignore PATH MTU Discovery, as per docs:
```
...
for  non-SOCK_STREAM sockets, IP_PMTUDISC_DO forces the don't-fragment flag to be set on all outgoing packets
...
```

closes: https://github.com/apache/trafficserver/issues/9488

### DIFF
--- a/src/iocore/net/UnixUDPNet.cc
+++ b/src/iocore/net/UnixUDPNet.cc
@@ -1173,6 +1173,12 @@ UDPNetProcessor::UDPBind(Continuation *cont, sockaddr const *addr, int fd, int s
       Dbg(dbg_ctl_udpnet, "setsockeopt for pktinfo failed");
       goto Lerror;
     }
+#ifdef IP_MTU_DISCOVER
+    int probe = IP_PMTUDISC_PROBE; // Set DF but ignore Path MTU
+    if (safe_setsockopt(fd, IPPROTO_IP, IP_MTU_DISCOVER, &probe, sizeof(probe)) == -1) {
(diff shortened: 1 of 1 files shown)


maskit:
https://datatracker.ietf.org/doc/html/rfc9000#name-handling-of-icmp-messages-b
Looks like DPLPMTUD is the preferred way.

maskit:
DF bit will be set for all UDP sockets ATS open (not only ones for QUIC), but I don't think that's going to be a problem.

brbzull0:
> DF bit will be set for all UDP sockets ATS open (not only ones for QUIC), but I don't think that's going to be a problem.

yes, I was thinking to just do it for quic, but as you said, wont be a problem. The kernel may do this for us some us. Anyway, we'll see, we can change it eventually. 
Thanks.

PR #10935, opened 2023-12-14
cmake: Add tsutil/Assert.h to public headers

hnakamur:
I suppose `tsutil/Assert.h` belongs to public headers since it is included from `tsutil/TsSharedMutex.h` which also belongs to public headers.


cmcfarlen (approved):
Thanks! Good catch!

PR #10938, opened 2023-12-14
Update cmake for cleaner install with static vs shared

cmcfarlen:
Also fixes issue with libswoc forcing CMAKE_INSTALL_LIBDIR


cmcfarlen:
[approve ci autest]

cmcfarlen:
The last commit was necessary for the OSX build linking traffic_logstats.  The references to tsapi from proxy were causing linker problems

bneradt (approved):

PR #10945, opened 2023-12-18
cmake: small fixes to install

cmcfarlen:


src/api/CMakeLists.txt
```diff
@@ -51,7 +51,7 @@ target_link_libraries(
 )
 
 install(
-  TARGETS tsapi tsapibackend
```
cmcfarlen, line 54: I should have removed this in the prior PR

cmake/post_install.cmake
```diff
@@ -17,8 +17,7 @@
 
 if("$ENV{USER}" STREQUAL root)
   foreach(DIR ${CHOWN_DIRS})
-    message(STATUS "Changing ${DIR} ownership to ${OWNER_USER}:${OWNER_GROUP}")
+    message(STATUS "Changing $ENV{DESTDIR}${DIR} ownership to ${OWNER_USER}:${OWNER_GROUP}")
```
JosiahWI, line 20: It shouldn't have `${CMAKE_INSTALL_PREFIX}` as well like the line below it?

cmcfarlen: Actually, the other should look like this one! I missed this picking from internal code.  Thanks!

JosiahWI (approved):
This also makes the CMakeUserPresets in .gitignore relative to the .gitignore location.

Looks good.

SolidWallOfCode (approved):

PR #10952, opened 2023-12-20
Cleanup: Span and Store

masaori335:
No logic change.

1. Remove unused functions and members from `struct Span`
2. Rename `disk` to `span`
3. Include only used headers

### DIFF
--- a/include/iocore/cache/Store.h
+++ b/include/iocore/cache/Store.h
@@ -79,29 +79,14 @@ struct Span {
   unsigned hw_sector_size = DEFAULT_HW_SECTOR_SIZE;
   unsigned alignment      = 0;
   span_diskid_t disk_id;
-  int forced_volume_num = -1; ///< Force span in to specific volume.
-private:
-  bool is_mmapable_internal = false;
-
-public:
-  bool file_pathname = false; // the pathname is a file
+  int forced_volume_num = -1;    ///< Force span in to specific volume.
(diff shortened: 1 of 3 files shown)


src/iocore/cache/Store.cc
```diff
@@ -486,7 +484,6 @@ Span::init(const char *path, int64_t size)
   }
 
   // A directory span means we will end up with a file, otherwise, we get what we asked for.
-  this->set_mmapable(ink_file_is_mmappable(S_ISDIR(sbuf.st_mode) ? static_cast<mode_t>(S_IFREG) : sbuf.st_mode));
```
masaori335, line 489: This is setting `Span::is_mmapable_internal`, but it's used nowhere.

oknet:
If we don't need to recall the old `MultiCache`, then these cleanups are OK.

include/iocore/cache/Store.h
```diff
@@ -189,25 +141,23 @@ struct Store {
...
       }
-      n_disks = i;
+      n_spans = i;
     }
   }
 
   void delete_all();
 
-  Store();
+  Store(){};
```
bneradt, line 155: Suggest: `Store() = default;`

But not a big deal.

bneradt (approved):
Looks good.

include/iocore/cache/Store.h
masaori335: I'll cleanup later.

PR #10956, opened 2024-01-02
boringssl: Add a type declaration

maskit:
Some declarations around X509 were moved to an internal header file on recent BoringSSL because they thought those were not used externally.
https://boringssl.googlesource.com/boringssl/+/1a7233d191952cc163133a5ab235a0ab0f31c812

And they realized that some of those were actually used externally and moved those back to a public header.
https://boringssl.googlesource.com/boringssl/+/faac623b09d6b14fbb06fe5150016de78b359eea
https://boringssl.googlesource.com/boringssl/+/538b2a6cf0497cf8bb61ae726a484a3d7a34e54e

However, these commits do not completely revert the removal commit, so there are still missing declarations which we use. This PR adds declarations for the missing ones, and make ATS buildable with the latest BoringSSL (b6e0eba6e62333652290514e51b75b966b27b27c).

### DIFF
--- a/src/iocore/net/OCSPStapling.cc
+++ b/src/iocore/net/OCSPStapling.cc
@@ -53,6 +53,9 @@ extern ClassAllocator<FetchSM> FetchSMAllocator;
 // clang-format off
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wunused-function"
+#ifndef GENERAL_NAME_it
+DECLARE_ASN1_ITEM(GENERAL_NAME)
+#endif
 // RFC 6960
 using TS_OCSP_CERTID = struct ocsp_cert_id {
   X509_ALGOR *hashAlgorithm;


src/iocore/net/OCSPStapling.cc
```diff
@@ -53,6 +53,9 @@ extern ClassAllocator<FetchSM> FetchSMAllocator;
 // clang-format off
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wunused-function"
+#ifndef GENERAL_NAME_it
+DECLARE_ASN1_ITEM(GENERAL_NAME)
+#endif
```
ywkaras, line 58: `GENERAL_NAME_it` in openssl is declared in the expansion of this macro:
https://github.com/openssl/openssl/blob/806bbafe2df5b699feac6ef26e50c14e701950cf/include/openssl/x509v3.h.in#L530
For some reason, boringssl has removed most of the contents of x509v3.h, moving some (but not all of it) to x509.h .

See:  https://github.com/openssl/openssl/blob/806bbafe2df5b699feac6ef26e50c14e701950cf/include/openssl/asn1.h.in#L244

maskit: Right, they (BoringSSL) assumed it's only used internally and move it into a private header.

ywkaras: OK hopefully this will be temporary.  Should we raise an issue in BoringSSL? (Is that even possible?)

PR #10957, opened 2024-01-02
Support elevated access option when loading client certs

lzx404243:
With this change, `proxy.config.ssl.cert.load_elevated` also takes affect when loading client certificates.


maskit (approved):
Looks reasonable.

PR #10958, opened 2024-01-03
Added config to support exit on client cert load failure

lzx404243:
This PR adds the `proxy.config.ssl.client.cert.exit_on_load_fail` configuration, which is analogous to the existing `proxy.config.ssl.server.multicert.exit_on_load_fail` config on the (SSL) server side. 

Unlike `proxy.config.ssl.server.multicert.exit_on_load_fail` the new config is default to `0`(don't exit when fails to load client cert), same as the current behavior without the config.


lzx404243:
[approve ci autest]

maskit (approved):

src/iocore/net/SSLConfig.cc
```diff
@@ -542,7 +544,13 @@ SSLConfigParams::initialize()
   // can cause HTTP layer to connect using SSL. But only if SSL
   // initialization hasn't failed already.
   client_ctx = this->getCTX(this->clientCertPath, this->clientKeyPath, this->clientCACertFilename, this->clientCACertPath);
-  if (!client_ctx) {
+  if (client_ctx) {
+    return;
+  }
+  // Can't get SSL client context.
+  if (this->clientCertExitOnLoadError) {
+    Fatal("Can't initialize the SSL client, HTTPS in remap rules will not function");
```
ywkaras, line 552: I think this should be Emergency(), not Fatal().  Emergency prevents immediate restart, which won't work if the error is a misconfiguration error:  https://docs.trafficserver.apache.org/developer-guide/api/functions/TSDebug.en.html#c.TSEmergency

maskit: Maybe? I wonder how it's controlled where we no longer have traffic_manager.
https://github.com/apache/trafficserver/blob/7ccdbd30ac8f03667d734fa40f04f1c250eed2c8/include/tscore/Diags.h#L86-L92

lzx404243: Thank you @ywkaras for pointing this out. I put up #11108 to address this. @maskit  Do you want to take a look at that?

Also wonder how the magic works :)

PR #10959, opened 2024-01-03
Fix a bug in OCSP update with dual cert configuration

maskit:
ATS uses two cert storages (one for RSA, the other for EC) because BoringSSL does not natively support dual certs, unlike OpenSSL. It works fine in terms of selecting a cert for a connection, but I found OCSP status update only uses the storage for RSA.

This PR surrounds the existing code by a loop that iterates over certification types in an array to cover all certs in the both storages. Hiding whitespace changes makes the diff easy to read.

Also, the change for SSLUtils is for a logging issue. It always printed a filename for either RSA or EC cert if dual certs is configured.

### DIFF
--- a/src/iocore/net/OCSPStapling.cc
+++ b/src/iocore/net/OCSPStapling.cc
@@ -1256,33 +1256,40 @@ ocsp_update()
   time_t current_time;
 
   SSLCertificateConfig::scoped_config certLookup;
-  const unsigned ctxCount = certLookup ? certLookup->count() : 0;
 
   Debug("ssl_ocsp", "updating OCSP data");
-  for (unsigned i = 0; i < ctxCount; i++) {
-    SSLCertContext *cc = certLookup->get(i);
-    if (cc) {
-      ctx = cc->getCtx();
-      if (ctx) {
-        certinfo *cinf    = nullptr;
-        certinfo_map *map = stapling_get_cert_info(ctx.get());
-        if (map) {
-          // Walk over all certs associated with this CTX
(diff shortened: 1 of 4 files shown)


randall (dismissed):

src/iocore/net/OCSPStapling.cc
```diff
@@ -1256,33 +1256,40 @@ ocsp_update()
...
-              ink_mutex_release(&cinf->stapling_mutex);
-              if (stapling_refresh_response(cinf, &resp)) {
-                Debug("ssl_ocsp", "Successfully refreshed OCSP for %s certificate. url=%s", cinf->certname, cinf->uri);
-                Metrics::Counter::increment(ssl_rsb.ocsp_refreshed_cert);
+#ifndef OPENSSL_IS_BORINGSSL
+  const SSLCertContextType ctxTypes[] = {SSLCertContextType::GENERIC};
+#else
+  const SSLCertContextType ctxTypes[] = {SSLCertContextType::RSA, SSLCertContextType::EC};
+#endif
+  for (auto &ctxType : ctxTypes) {
```
bneradt, line 1266: I think `auto const &ctxType` would work here. (i.e., making it a const reference).

bneradt (changes requested):

bneradt (dismissed):

src/iocore/net/OCSPStapling.cc
maskit: Added `const`.

bneradt (approved):
Looks good.

PR #10960, opened 2024-01-04
Persist bad disks

moonchen:
Add a config option to have traffic server remember which disks are bad.  Previously, cache disks that were determined to be bad at runtime were forgotten upon restart.  This could cause corrupt cache metadata to be loaded from disk, leading to crashes, or slowdowns when accessing the known bad disk.

If the option is turned on, bad disks are remembered until the `known_bad_disks` file is edited or deleted.  If the system administrator replaces the disks, they need to perform this manual step.  Because of this requirement, this option is off by default.

### DIFF
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -2488,6 +2488,29 @@ Cache Control
    used from the asynchronous IO threads when IO finishes and the ``CacheVC`` lock or stripe lock is
    required.
 
+.. ts::cv:: CONFIG proxy.config.cache.max_disk_errors INT 5
+
+   Cache disks sometimes fail due to hardware problems.  |TS| keeps count of
+   the number of times it encounters I/O errors when accessing each cache disk.
+   If the number of errors on a disk reaches this setting, |TS| removes that
+   disk from the cache.
+
+   Note that the count of errors is kept in memory, and is reset to zero when
+   |TS| is restarted.  By default, |TS| will not remember which cache disk has
+   been removed in this way when it restarts.  If you wish to change this behavior
+   and prevent known bad disks from re-joining the cache upon restart, change
+   the setting :ts:cv:`proxy.config.cache.persist_bad_disks`.
+
+.. ts::cv:: CONFIG proxy.config.cache.persist_bad_disks INT 0
+
+   When enabled (``1``), |TS| will remember which cache disks have been
+   marked as failed through :ts:cv:`proxy.config.cache.max_disk_errors`,
+   even after a restart.  A list of known bad cache disks is written to
+   ``localstatedir/known_bad_disks``.  If you replace the known bad disks,
+   delete this file so that |TS| will use it in the cache again.
+
+   When disabled (``0``), |TS| will ignore the ``known_bad_disks`` file.
+
 RAM Cache
 =========
 
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -2477,7 +2477,7 @@ Cache Control
    write vector. For further details on cache write vectors, refer to the
    developer documentation for :cpp:class:`CacheVC`.
 
-.. ts::cv:: CONFIG proxy.config.cache.mutex_retry_delay INT 2
+.. ts:cv:: CONFIG proxy.config.cache.mutex_retry_delay INT 2
    :reloadable:
    :units: milliseconds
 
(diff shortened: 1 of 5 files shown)


bneradt:
[approve ci ubuntu]

doc/admin-guide/files/records.yaml.en.rst
```diff
@@ -2488,6 +2488,29 @@ Cache Control
    used from the asynchronous IO threads when IO finishes and the ``CacheVC`` lock or stripe lock is
    required.
 
+.. ts::cv:: CONFIG proxy.config.cache.max_disk_errors INT 5
```
bneradt, line 2491: Interestingly, these don't render:
https://ci.trafficserver.apache.org/job/Github_Builds/job/docs/4133/artifact/output/10960/docbuild/html/admin-guide/files/records.yaml.en.html#ram-cache

I think the issue is `ts:cv::` is correct, but `ts::cv::` (note the two colons after `ts` instead of one) is incorrect. It looks like you're the victim of copying and pasting from a previous mistake. And I'm glad you are so we can fix it.

Can you please search for all `ts::cv::` in this file and replace it with the correct `ts:cv::`? I think I count four such mistakes (including the two in this patch).

bneradt (changes requested):

doc/admin-guide/files/records.yaml.en.rst
```diff
@@ -2488,6 +2488,29 @@ Cache Control
...
+.. ts::cv:: CONFIG proxy.config.cache.persist_bad_disks INT 0
+
+   When enabled (``1``), |TS| will remember which cache disks have been
+   marked as failed through :ts:cv:`proxy.config.cache.max_disk_errors`,
+   even after a restart.  A list of known bad cache disks is written to
+   ``localstatedir/known_bad_disks``.  If you replace the known bad disks,
+   delete this file so that |TS| will use it in the cache again.
+
+   When disabled (``0``), |TS| will ignore the ``known_bad_disks`` file.
+
```
bneradt, line 2513: It's typical to say something explicit about the default for something like this, like: 

```
This feature is disabled by default.
```

src/iocore/cache/Cache.cc
```diff
@@ -336,6 +344,15 @@ CacheProcessor::start_internal(int flags)
       opts |= O_CREAT;
     }
 
+    // Check if the disk is known to be bad
```
bneradt, line 347: nit: end with a period.

src/iocore/cache/Cache.cc
```diff
@@ -336,6 +344,15 @@ CacheProcessor::start_internal(int flags)
...
+    if (cache_config_persist_bad_disks && !known_bad_disks.empty()) {
+      if (known_bad_disks.find(paths[gndisks]) != known_bad_disks.end()) {
+        Note("%s is a known bad disk.  Skipping.", paths[gndisks]);
```
bneradt, line 350: Probably should be a warning.

include/tscore/Filenames.h
```diff
@@ -47,6 +47,7 @@ namespace filename
   // Various other file names
   constexpr const char *RECORDS_STATS = "records.snap";
   constexpr const char *HOST_RECORDS  = "host_records.yaml";
+  constexpr const char *BAD_DISKS     = "bad_disks";
```
bneradt, line 50: Probably good to add a `.txt` suffix. That can be helpful for some editors.

bneradt (changes requested):
I have some small comments.

I see BAD_DISKS being written to but not read from. I might simply be missing it, but where is the file used on startup to persist the information?

doc/admin-guide/files/records.yaml.en.rst
moonchen: Thanks for the catch!  I've fixed the offending `ts::cv::` lines.

moonchen:
> I have some small comments.
> 
> I see BAD_DISKS being written to but not read from. I might simply be missing it, but where is the file used on startup to persist the information?

The code that loads BAD_DISKS was accidentally missed when I rebased to master branch.  I've added it back.

moonchen:
[approve ci]

bneradt (approved):

randall:
One drive by comment: should this really be a plain text file? I'm guessing this is for ease of administration, but it does seem weird to not have this in YAML since that's where we'd like to be (all yaml config files)

moonchen:
> One drive by comment: should this really be a plain text file? I'm guessing this is for ease of administration, but it does seem weird to not have this in YAML since that's where we'd like to be (all yaml config files)

In the general case, this file isn't meant to be a config file.  It's there to keep track of which disks were marked bad automatically during previous runs of ATS.  I think an administrator would probably not need to edit this file manually, and a text file would help keep the feature simple.

PR #10961, opened 2024-01-04
Cleanup of HdrHeap::HeapGuard and HdrStrHeap classes.

ywkaras:


src/proxy/hdrs/HdrHeap.cc
```diff
@@ -1168,12 +1178,9 @@ HdrStrHeap::free()
...
 HdrStrHeap::allocate(int nbytes)
 {
-  char *new_space;
-
-  if (m_free_size >= static_cast<unsigned>(nbytes)) {
-    new_space     = m_free_start;
-    m_free_start += nbytes;
-    m_free_size  -= nbytes;
+  if (_avail_size >= static_cast<unsigned>(nbytes)) {
+    char *new_space  = reinterpret_cast<char *>(this) + _total_size - _avail_size;
```
masaori335, line 1182: I'd like to add a function that calculates this pointer with checking values.

src/proxy/hdrs/HdrHeap.cc
```diff
@@ -241,39 +240,50 @@ HdrHeap::allocate_str(int nbytes)
...
-  }
-  // Try to allocate of our read/write string heap
-  new_space = m_read_write_heap->allocate(nbytes);
+  for (;;) {
+    if (coalesce) {
+      switch (coalesce) {
+      case 2:
+        Warning("HdrHeap=%p coalescing twice", this);
+        break;
+      case 3:
```
masaori335, line 251: Have you ever seen `coalesce` become more than `3`?

masaori335 (approved):
It looks like just cleanups.

src/proxy/hdrs/HdrHeap.cc
ywkaras: No.  The original code allowed coalescing to be repeated an unlimited number of times.  I think it should only happen once, but I'm not 100% sure, so I didn' t want to add assert to check it.  If we see coalescing happening repeatedly, we should try to figure out if it make sense for that to happen.

src/proxy/hdrs/HdrHeap.cc
ywkaras: The checking is already present, on line 1181.  It results in a nullptr return if the checking fails.  The only place this function is called is in HdrHeap::allocate_str(), which does handle the null return value.

PR #10964, opened 2024-01-04
multiplexer: debug logging and variable name updates

bneradt:
This is a multiplexer cleanup PR that adds debug logging to post.cc and updates variable names for the transform in order to make their purpose explicit (for the origin, for a copy, etc.). In post.cc, for instance, both the incoming and outgoing VIO were named `vio`, with the difference being whether it was called off of the state `s` variable or as a local variable. This updates the names to clarify the purpose of each.

Other than logging changes, this patch makes no functional change to multiplexer.

### DIFF
--- a/plugins/multiplexer/dispatch.cc
+++ b/plugins/multiplexer/dispatch.cc
@@ -31,6 +31,8 @@
 #error Please define a PLUGIN_TAG before including this file.
 #endif
 
+using multiplexer_ns::dbg_ctl;
+
 extern Statistics statistics;
 
 size_t timeout;
@@ -229,6 +231,7 @@ generateRequests(const Origins &o, const TSMBuffer buffer, const TSMLoc location
     const std::string &host = *iterator;
     assert(!host.empty());
     request.hostHeader(host);
+    Dbg(dbg_ctl, "Preparing request for \"%s\"", host.c_str());
     r.push_back(Request(host, buffer, location));
   }
 }
--- a/plugins/multiplexer/post.cc
+++ b/plugins/multiplexer/post.cc
@@ -24,20 +24,24 @@
 #include <limits>
 
 #include "post.h"
+#include "ts/ts.h"
+#include "tsutil/DbgCtl.h"
 
 #ifndef PLUGIN_TAG
 #error Please define a PLUGIN_TAG before including this file.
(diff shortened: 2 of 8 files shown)


bneradt:
[approve ci]

plugins/multiplexer/post.cc
```diff
@@ -48,55 +52,65 @@ postTransform(const TSCont c, PostState &s)
...
-    assert(s.vio != nullptr);
+    s.output_vio = TSVConnWrite(output_vconn, c, origin_reader, std::numeric_limits<int64_t>::max());
+    assert(s.output_vio != nullptr);
   }
 
-  if (!TSVIOBufferGet(vio)) {
-    TSVIONBytesSet(s.vio, TSVIONDoneGet(vio));
-    TSVIOReenable(s.vio);
+  if (!TSVIOBufferGet(input_vio)) {
+    if (s.output_vio) {
```
bneradt, line 80: This is the only "interesting" change in this PR in which I add a nullptr check just in case.

shukitchan (approved):
This looks straightforward. Thanks.

plugins/multiplexer/dispatch.cc
```diff
@@ -229,6 +231,7 @@ generateRequests(const Origins &o, const TSMBuffer buffer, const TSMLoc location
     const std::string &host = *iterator;
     assert(!host.empty());
     request.hostHeader(host);
+    Dbg(dbg_ctl, "Prepaing request for \"%s\"", host.c_str());
```
randall, line 234: Spelling (Preparing)

randall (dismissed):

plugins/multiplexer/dispatch.cc
bneradt: Good catch. Thanks! Updated.

randall (approved):

bneradt:
[approve ci autest]

plugins/multiplexer/post.cc
oknet: The value of `s.output_vio` is always `non-null` because it returned by `INKVConnInternal::do_io_write()`.

It's good job to prefix the variable names with `input` and `output`.

PR #10965, opened 2024-01-04
multiplexer: Correctly Handle Large Content-Length POST/PUT Request Bodies

bneradt:
The multiplexer plugin assumed that for POST/PUT requests it could pass INT64_MAX to the `TSVConnWrite` api for transforming the request bodies. After all, INT64_MAX is often used for transforming response bodies. INT64_MAX accidentally worked for small request bodies that got sent in a single READ_READY call, but for larger bodies, HttpSM complained about the unsupported INT64_MAX call for requests. It is unsupported for requests because INT64_MAX correlates with chunked requests which our transform mechanism doesn't support.

Thus, multiplexer was not correctly abiding by the expected API requirement that for POST/PUT requests, `TSVConnWrite` requires an explicit length passed in. This fixes that bug. We should consider for a future PR whether we want to support transforming chunked requests.

### DIFF
--- a/tests/gold_tests/pluginTest/multiplexer/multiplexer.test.py
+++ b/tests/gold_tests/pluginTest/multiplexer/multiplexer.test.py
@@ -31,6 +31,7 @@ class MultiplexerTestBase:
     """
 
     client_counter = 0
+    dns_counter = 0
     server_counter = 0
     ts_counter = 0
 
@@ -39,7 +40,13 @@ class MultiplexerTestBase:
         self.multiplexed_host_replay_file = multiplexed_host_replay_file
 
         self.setupServers()
+        self.setupDns()
         self.setupTS(skip_post)
+
+    def setupDns(self):
+        counter = MultiplexerTestBase.dns_counter
+        MultiplexerTestBase.dns_counter += 1
+        self.dns = Test.MakeDNServer(f"dns_{counter}", default='127.0.0.1')
 
     def setupServers(self):
         counter = MultiplexerTestBase.server_counter
@@ -57,6 +64,7 @@ class MultiplexerTestBase:
             'X-Multiplexer: original', 'Verify the HTTP multiplexed host does not receive an "original".')
         self.server_https.Streams.All += Testers.ExcludesExpression(
             'X-Multiplexer: original', 'Verify the HTTPS multiplexed host does not receive an "original".')
+        self.server_https.Streams.All += Testers.ExcludesExpression(r'\[ERROR\]', 'Verify there were no errors in the replay.')
 
         # In addition, the original server should always receive the POST and
         # PUT requests.
@@ -64,6 +72,7 @@ class MultiplexerTestBase:
             'uuid: POST', "Verify the client's original target received the POST transaction.")
         self.server_origin.Streams.All += Testers.ContainsExpression(
             'uuid: PUT', "Verify the client's original target received the PUT transaction.")
+        self.server_origin.Streams.All += Testers.ExcludesExpression(r'\[ERROR\]', 'Verify there were no errors in the replay.')
 
         # Under all configurations, the GET request should be multiplexed.
         self.server_origin.Streams.All += Testers.ContainsExpression(
@@ -74,6 +83,7 @@ class MultiplexerTestBase:
         self.server_http.Streams.All += Testers.ContainsExpression(
             'X-Multiplexer: copy', 'Verify the HTTP server received a "copy" of the request.')
         self.server_http.Streams.All += Testers.ContainsExpression('uuid: GET', "Verify the HTTP server received the GET request.")
+        self.server_http.Streams.All += Testers.ExcludesExpression(r'\[ERROR\]', 'Verify there were no errors in the replay.')
 
         self.server_https.Streams.All += Testers.ContainsExpression(
             'X-Multiplexer: copy', 'Verify the HTTPS server received a "copy" of the request.')
@@ -96,6 +106,8 @@ class MultiplexerTestBase:
                 "proxy.config.ssl.client.verify.server.policy": 'PERMISSIVE',
                 'proxy.config.diags.debug.enabled': 1,
                 'proxy.config.diags.debug.tags': 'http|multiplexer',
+                'proxy.config.dns.nameservers': f'127.0.0.1:{self.dns.Variables.Port}',
+                'proxy.config.dns.resolv_conf': 'NULL',
             })
         self.ts.Disk.ssl_multicert_config.AddLine('dest_ip=* ssl_cert_name=server.pem ssl_key_name=server.key')
         skip_remap_param = ''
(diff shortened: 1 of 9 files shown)


tests/gold_tests/pluginTest/multiplexer/multiplexer.test.py
```diff
@@ -96,33 +106,37 @@ def setupTS(self, skip_post):
...
+                'proxy.config.dns.resolv_conf': 'NULL',
             })
         self.ts.Disk.ssl_multicert_config.AddLine('dest_ip=* ssl_cert_name=server.pem ssl_key_name=server.key')
         skip_remap_param = ''
         if skip_post:
             skip_remap_param = ' @pparam=proxy.config.multiplexer.skip_post_put=1'
         self.ts.Disk.remap_config.AddLines(
             [
-                f'map https://origin.server.com https://127.0.0.1:{self.server_origin.Variables.https_port} '
+                f'map https://origin.server.com https://backend.origin.server.com:{self.server_origin.Variables.https_port} '
```
ywkaras, line 118: Does MicroDNS return the loopback address for any hostname it wasn't configured with?  If so, does that mean we can't use MicroDNS to test DNS lookup failures?

bneradt: Typically, we configure the default to resolve everything to 127.0.0.1. In this PR, you can see that here:

https://github.com/apache/trafficserver/pull/10965/files#diff-ab6abc922014fa32a8c9d9a16e3b9be6e89f16a45a671728146d270ba2d848cfR49

I _think_ MicroDNS can be used to test failures, but I'm not sure offhand. It might require some modification of it.

oknet:
There are 2 types of transforms: `Post Transform` and `Response Transform`.

The `Transform` read data from upstream and write data to downstream.

As the writer of plugin, you can do it by InkAPI `TSVConnRead and `TSVConnWrite`.

There is a parameter "nbytes" indicating the bytes to be read or written.

The downstream of the last `Transform Plugin` is handled by the `TransformTerminus`.

When the last `Transform Plugin` calling the InkAPI `TSVConnWrite` on the downsteam VC, 

- the `TransformTerminus` will call back `TRANSFORM_READ_READY` with `nbytes` as `data` to the HttpSM, 
- and the event handled by `HttpSM::state_request_wait_for_transform_read` or `HttpSM::state_response_wait_for_transform_read` per the type of transforms.

For the `Response Transform`, if the `HttpSM::state_response_wait_for_transform_read` received `TRANSFORM_READ_READY` event with `INT64_MAX`, it tries to re-chunk the data from `TransformVC`.

Please refer these 2 options `proxy.config.http.chunking_enabled` and `proxy.config.http.chunking.size` to enable the re-chunk feature.

But the `HttpSM::state_request_wait_for_transform_read` does not accept `INT64_MAX`, because it directly fill the `nbytes` to the `Content-Length` header. 

The `HttpTunnel` decode the chunked data If the POST payload is chunked, and then write de-chunked data into `TransformVC`. But the `HttpSM` does not re-chunk the data received from `TransformVC`, even `HttpTunnel` can do it. 

This pull request is intended to serve `multiplexer plugin`, as `multiplexer plugin` does not change the length of the payload.

In order to support widely scene, it is better to enable / implement the re-chunk feature for `Post Transform`.

bneradt:
> This pull request is intended to serve multiplexer plugin, as multiplexer plugin does not change the length of the payload.
> 
> In order to support widely scene, it is better to enable / implement the re-chunk feature for Post Transform.

I see. Thank you @oknet, this is a very helpful feedback and context. Reading through your explanation, I see two sets of things that can happen:

1. The multiplexer plugin should simply read the Content-Length header and use that for the `TSVConnWrite` api. This won't handled chunked POST requests of course, but it will at least fix things for large Content-Length bodies. And it correctly uses the `TSVConnWrite` api since it doesn't support INT64_MAX for POST requests.
2. If there are chunked requests, we'll have to handle that by updating the rechunk feature of Post Transform, as you state.

For this pull request, I'll stick with simply addressing the first of those issues. This should cover the majority of contexts - in my experience, chunked POST requests are pretty rare in comparison to Content-Length.

oknet:
By review your current code, let me confirm the implement and details with you:

1. dispatch HTTP request directly except ( POST and PUT )
2. For the POST and PUT request with valid `Content-Length` header,
    a. Collect data into `s.origin_buffer` by transform plugin,
    b. Get all data by `s.clone_reader` while the transformVC closed,
    c. Then dispatch the request.
3. For the POST and PUT request without `Content-Length` header, does not dispatch the request.

I'll approve your commit if the above list is all you have done at this stage.

For step 2.a, the data is kept in memory until all data is received, this is not a good design, I hope you can improve the design in the future.

Correction for my last comment:

If the POST payload is chunked, the `HttpTunnel` decodes the chunked data  **and verifies the integrity of the chunked data,  but it still writes the unchanged data** to `TransformVC`. But the `HttpSM` does not re-chunk the data received from `TransformVC`, even `HttpTunnel` can do it.

In addition to chunking, there is also `Content-Type: multipart/mixed` type which not supported by `HttpSM` and `HttpTunnel`. Maybe that's why the `HttpTunnel` only validates the request's payload and passes it unchanged to `transformVC`. 

For those requests without a "Content-Length" header, you can defer calling `TSVConnWrite(output_vconn, ...)` till upstream's `vio->ntodo() == 0`, and you have to copy all data from `upsteam’s vio` to a temporary `IOBuffer` then consume all data from  `upstream's vio` in order to make NetHandler read more data.

https://github.com/apache/trafficserver/blob/8f5d558461cc0ad2f91f579f40754678c2e6803a/plugins/multiplexer/post.cc#L75

The upstream's `vio->ntodo() == 0` means all data was sent. (refer to doc/developer-guide/plugins/io/transformations.en.rst )

https://github.com/apache/trafficserver/blob/8f5d558461cc0ad2f91f579f40754678c2e6803a/doc/developer-guide/plugins/io/transformations.en.rst?plain=1#L156-L163

bneradt:
> By review your current code, let me confirm the implement and details with you:
> 
> 1. dispatch HTTP request directly except ( POST and PUT )
> 2. For the POST and PUT request with valid `Content-Length` header,
>    a. Collect data into `s.origin_buffer` by transform plugin,
>    b. Get all data by `s.clone_reader` while the transformVC closed,
>    c. Then dispatch the request.
> 3. For the POST and PUT request without `Content-Length` header, does not dispatch the request.
> 
> I'll approve your commit if the above list is all you have done at this stage.

Yes, this is correct. All that my current patch does now is set the length explicitly in the `TSVConnWrite` because that's what the API currently requires for POST bodies. Basically, this current patch fixes multiplexer to behave like the current API requires. Which is at least a step in a good direction, because it make multiplexer work at least for Content-Length bodies.

oknet (approved):

PR #10970, opened 2024-01-05
tools/install_cmake.sh

bneradt:
This adds tools/install_cmake.sh which installs a recent version of CMake.

### DIFF
--- a/tools/install_cmake.sh
+++ b/tools/install_cmake.sh
@@ -1,0 +1,92 @@
+#! /usr/bin/env bash
+#
+#  Grab and install a recent version of CMake.
+#
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


maskit:
I don't mind having this script as long as I'm not required to use it, but I wonder why the script uses Homebrew for MacOS where the repository has binaries for MacOS as well. The script wouldn't have the install path limitation if it consistently grab an official release from the GitHub repository.

bneradt:
> I don't mind having this script as long as I'm not required to use it, 

You're definitely not required to use it. I was asked to write this to make it easier for people to get the latest version of cmake.

> I wonder why the script uses Homebrew for MacOS where the repository has binaries for MacOS as well. The script wouldn't have the install path limitation if it consistently grab an official release from the GitHub repository.

Right, MacOS is a bit more challenging than the Linux case. I'm open to suggestions, but I don't think that using the `.dmg`/`.app` version avoids the install path problem. I don't think that a `.app` file can just be copied anywhere. At least generally speaking they seem to need to go into the `/Applications` directory, or maybe the `~/Applications` directory. And that's initially how I was writing this, but I didn't feel comfortable with this script installing stuff in there. That seemed a bit invasive for an ATS tools script to install stuff into those directories.

`brew` on the other hand is a much easier solution. It's just a single command rather than a wget, dmg mount, and copy. `brew` exists after all to make stuff like this easy. Also, `brew` is pretty good about keeping up with the latest version. They already install 3.28.1, which is latest, for instance.

Anyway, that's why I chose the `brew` route.

That being the case, maybe this script should just not handle MacOS? It can just echo a statement saying "This script does not handle MacOS. Consider the Kitware .dmg file <here> or using brew to install cmake".

maskit:
Oh, I assumed the tar.gz has a command line executable and we can put it into anywhere. I just found it actually has CMake.app instead. Now it looks like homebrew is a reasonable way to install the command (and this is how I install the command).

We should probably ask real users what they want. I don't mind dropping the support for MacOS. I can't imagine a situation that this script helps people trying to build ATS. If a person has to rely on this script to install cmake, the person probably doesn't know how to install dependent libraries either.

maskit (approved):

PR #10971, opened 2024-01-08
Fix problem with the negation in TextView/libswoc

shukitchan:
ABS_MIN is 9223372036854775808
intmax_t(ABS_MIN) is -9223372036854775808
negation of that is overflow for intmax_t data type with max value of 9223372036854775807
And I think that's not the intention as well. 

So I need to make this change.

### DIFF
--- a/lib/swoc/src/TextView.cc
+++ b/lib/swoc/src/TextView.cc
@@ -63,7 +63,12 @@ svtoi(TextView src, TextView *out, int base) {
         out->assign(start, parsed.data_end());
       }
       if (neg) {
-        zret = -intmax_t(std::min<uintmax_t>(n, ABS_MIN));
+        uintmax_t temp = std::min<uintmax_t>(n, ABS_MIN);
+        if (temp == ABS_MIN) {
+          zret = std::numeric_limits<intmax_t>::min();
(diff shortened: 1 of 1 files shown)


shukitchan:
This is needed for the problem found in oss fuzz - https://oss-fuzz.com/testcase-detail/5196561539530752

shukitchan:
/src/trafficserver/lib/swoc/src/TextView.cc:66:16: runtime error: negation of -9223372036854775808 cannot be represented in type 'intmax_t' (aka 'long'); cast to an unsigned type to negate this value to itself

bneradt:
[approve ci]

bneradt (commented):
Ideally we would add a unit test for this, but I guess our copied libswoc doesn't have that. 

Can you make a libswoc PR for this change and add a unit test here:
https://github.com/SolidWallOfCode/libswoc/blob/master/unit_tests/test_TextView.cc#L453

moonchen (approved):
Agreed with @bneradt about the unit test.  Otherwise, the code LGTM.

shukitchan:
Created https://github.com/SolidWallOfCode/libswoc/pull/92

PR #10976, opened 2024-01-09
Cleanup of Thread and EThread classes.

ywkaras:


maskit (approved):
Looks good.

This is unrelated to the change, but I don't get the point of `THREAD_MUTEX_THREAD_HOLDING` at all.

```
#define THREAD_MUTEX_THREAD_HOLDING (-1024 * 1024)
```

ywkaras:
`ProxyMutex` is a recursive mutex.  Mostly, `nthread_holding` counts the number of current recursive locks.
```
O$ printf "%x\n" $(( -1024 << 10 ))
fffffffffff00000
```
My best guess is, the point of doing this is to be able to recognize a Thread mutex when `nthread_holding` is examined in the debugger.

Issue #10955, opened 2023-12-27
ATS 9.2.x and origin 301/302 chase redirections


mkrug1981:
We are running Trafficserver 9.2.3 and observed an issue where trafficserver seems not to work as expected by following 302 redirects.

Following the redirect from Origin, ATS now serves the content on the first attempt. Unfortunately, re-querying the same object always results in a timeout.
After some debugging, it appears that the second and subsequent requests never get the content and always hit a timeout.
The debug logs show 'cache open read failure' in a retry loop.

```
[Dec 22 09:13:55.428] [ET_NET 19] DEBUG: <HttpTransact.cc:1499 (HandleRequest)> (http_seq) [453] request valid.
[Dec 22 09:13:55.429] [ET_NET 19] DEBUG: <HttpTransact.cc:2148 (DecideCacheLookup)> (http_trans) [453] Will do cache lookup.
[Dec 22 09:13:55.429] [ET_NET 19] DEBUG: <HttpTransact.cc:2149 (DecideCacheLookup)> (http_seq) [453] Will do cache lookup
[Dec 22 09:13:55.429] [ET_NET 19] DEBUG: <HttpSM.cc:7636 (call_transact_and_set_next_state)> (http) [453] State Transition: SM_ACTION_API_POST_REMAP -> SM_ACTION_CACHE_LOOKUP
[Dec 22 09:13:55.429] [ET_NET 19] DEBUG: <HttpSM.cc:4860 (do_cache_lookup_and_read)> (http_seq) [453] Issuing cache lookup for URL http://ida-mkrug-cdn.check.com/aaa/test100MB.nff
[Dec 22 09:13:56.440] [ET_NET 19] DEBUG: <HttpCacheSM.cc:100 (state_cache_open_read)> (http_cache) [453] [&HttpCacheSM::state_cache_open_read, CACHE_EVENT_OPEN_READ_FAILED/TS_EVENT_CACHE_OPEN_READ_FAILED]
[Dec 22 09:13:56.451] [ET_NET 19] DEBUG: <HttpCacheSM.cc:100 (state_cache_open_read)> (http_cache) [453] [&HttpCacheSM::state_cache_open_read, EVENT_INTERVAL]
[Dec 22 09:13:56.451] [ET_NET 19] DEBUG: <HttpCacheSM.cc:142 (state_cache_open_read)> (http_cache) [453] [state_cache_open_read] cache open read failure 1. retrying cache open read...
[...]

PR #10981, opened 2024-01-09
Improve redirect handling

elsloo:
* Do not permit a 302 to be cached if a redirect is in progress.
* Could be more permissive if we remove the check for `HTTP_STATUS_MOVED_TEMPORARILY` in `HttpTransact::is_response_cacheable()`
* Refactor HttpSM::is_redirect_required() to make it easier to call when state is incomplete
* May address issue #10955

### DIFF
--- a/include/proxy/http/HttpSM.h
+++ b/include/proxy/http/HttpSM.h
@@ -258,6 +258,7 @@ public:
 
   bool is_private() const;
   bool is_redirect_required();
+  bool is_redirect_required_for_status(HTTPStatus *status);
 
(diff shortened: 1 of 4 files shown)


elsloo:
Changing this PR to a draft until I complete testing and add the AuTests to this PR with a description of what it solves. Without this PR we have corner cases that can lead to a 302 being cached unintentionally, which then leads to secondary problems due to the 302 itself being possibly stuck in cache and served to other caches.

bryancall:
[approve ci]

PR #10982, opened 2024-01-10
Update build_h3_tools.sh

bneradt:
This updates build_h3_tools.sh to the latest library dependencies. The curl library dependencies are updated per this:
https://curl.se/docs/http3.html

@maskit guided me on the boringssl version.


maskit (approved):

bneradt:
I ran autests with the changes locally and they all passed.

PR #10983, opened 2024-01-10
CMAKE_INSTALL_FULL_LOGDIR and CMAKE_INSTALL_FULL_CACHEDIR

bneradt:
We use the FULL versions of CMAKE_INSTALL_LOGDIR and CMAKE_INSTALL_CACHEDIR, but GNUInstallDirs doens't know about these since they are custom for us and therefore doesn't automatically create them. This explicitly sets the values of these.

### DIFF
--- a/cmake/layout.cmake
+++ b/cmake/layout.cmake
@@ -62,8 +62,20 @@ set(CMAKE_INSTALL_LOGDIR
     "${CMAKE_INSTALL_LOCALSTATEDIR}/log/trafficserver"
     CACHE STRING "logdir"
 )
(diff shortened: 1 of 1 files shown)


cmcfarlen (approved):
I think this is fine, but it might need to consider DESTDIR in the places that use it.

We might also consider not setting permissions on install paths and leave it to post-install steps or packager to ensure ownership matches their config.

Even if we use the configured user/group a user could still change that in the config so would also have to adjust the permissions.

This is also only an issue when installed from source.  Package maintainers will have to ensure the permissions are correct regardless.  And package users will have no knowledge of this at all.

PR #10985, opened 2024-01-11
CMake: Make sure pthread is available when checking for some of the library functions.

brbzull0:
At least in my ubuntu 20.04(probably in others as well) `check_symbol_exists` seems not to detect the passed funcions to get the thread name. This is fine unless you need the thread name in the logs, like some of the  [autest](https://github.com/apache/trafficserver/blob/a49c20acb710939893c23e50f77608b0815470d5/tests/gold_tests/timeout/quic_poll_timeout.test.py#L68).

With this change, symbols gets detected and the proper macro variable is set and [used](https://github.com/apache/trafficserver/blob/a49c20acb710939893c23e50f77608b0815470d5/src/tsutil/DbgCtl.cc#L345-L353).

master/quic Ci was reporting this to be an issue on the above autest.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -416,6 +416,7 @@ if(HAVE_IOURING AND USE_IOURING)
   set(TS_USE_LINUX_IO_URING 1)
 endif(HAVE_IOURING AND USE_IOURING)
 
+list(APPEND CMAKE_REQUIRED_LIBRARIES pthread)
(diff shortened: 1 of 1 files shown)


JosiahWI (approved):
The problem seems to have been that the pthread library with the needed symbols was not getting linked to the executable created by the check as it should have been.

I wondered whether we should use the `FindThreads` module provided by CMake, but [their pthread check program](https://github.com/Kitware/CMake/blob/master/Modules/FindThreads.cmake#L78) does not check for the same symbols that we do. This fix looks good.

PR #10987, opened 2024-01-15
Move some logic from CacheVC to Stripe and AggregateWriteBuffer

JosiahWI:
This breaks up some of `handleRead` into smaller methods and moves the logic for copying out of the aggregate write buffer to the classes responsible for that buffer.

There was a check to make sure that data was only copied from the aggregate write buffer if the requested data was available. I left the old check because it avoids allocating a new buffer if the data is unavailable, and I also duplicated it in `Stripe::copy_from_aggregate_write_buffer` for safety. I thought it would be better to have it in and take it out if it negatively impacts performance than not to have it and find out it would have been useful. I have not benchmarked this.

The order of the parameters of `Stripe::copy_from_aggregate_write_buffer` and `AggregateWriteBuffer::copy_from` was chosen to reflect the order of the `memcpy` parameters. This does not mean I like the way `memcpy` chose to do it, I just thought it might help consistency. :)

This is part of a chain of PRs to remove all the getters I added to `Stripe`, in order to encapsulate the aggregate write buffer properly. This PR removes `Stripe::get_agg_buffer`.

### DIFF
--- a/include/iocore/cache/CacheVC.h
+++ b/include/iocore/cache/CacheVC.h
@@ -141,6 +141,7 @@ struct CacheVC : public CacheVConnection {
 
   int handleReadDone(int event, Event *e);
   int handleRead(int event, Event *e);
+  bool check_ram_cache();
   int do_read_call(CacheKey *akey);
   int handleWrite(int event, Event *e);
   int handleWriteLock(int event, Event *e);
--- a/include/iocore/cache/CacheVC.h
+++ b/include/iocore/cache/CacheVC.h
@@ -141,9 +141,9 @@ struct CacheVC : public CacheVConnection {
 
   int handleReadDone(int event, Event *e);
   int handleRead(int event, Event *e);
-  bool check_ram_cache();
-  bool check_last_open_read_call();
-  bool check_aggregation_buffer();
+  bool load_from_ram_cache();
+  bool load_from_last_open_read_call();
+  bool load_from_aggregation_buffer();
   int do_read_call(CacheKey *akey);
   int handleWrite(int event, Event *e);
   int handleWriteLock(int event, Event *e);
--- a/include/iocore/cache/CacheVC.h
+++ b/include/iocore/cache/CacheVC.h
@@ -142,6 +142,7 @@ struct CacheVC : public CacheVConnection {
   int handleReadDone(int event, Event *e);
   int handleRead(int event, Event *e);
   bool check_ram_cache();
+  bool check_last_open_read_call();
   int do_read_call(CacheKey *akey);
   int handleWrite(int event, Event *e);
   int handleWriteLock(int event, Event *e);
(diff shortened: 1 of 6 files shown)


src/iocore/cache/CacheVC.cc
```diff
@@ -535,6 +519,40 @@ LramHit: {
...
+{
+  if (!dir_agg_buf_valid(this->stripe, &this->dir)) {
+    return false;
+  }
+
+  this->buf = new_IOBufferData(iobuffer_size_to_index(this->io.aiocb.aio_nbytes, MAX_BUFFER_SIZE_INDEX), MEMALIGNED);
+  char *doc = this->buf->data();
+  [[maybe_unused]] bool success = this->stripe->copy_from_aggregate_write_buffer(doc, dir, this->io.aiocb.aio_nbytes);
+  // We already confirmed that the copy was valid, so it should not fail.
+  assert(success);
```
masaori335, line 552: `ink_assert`?

include/iocore/cache/CacheVC.h
```diff
@@ -141,6 +141,9 @@ struct CacheVC : public CacheVConnection {
 
   int handleReadDone(int event, Event *e);
   int handleRead(int event, Event *e);
+  bool check_ram_cache();
```
masaori335, line 144: Can we make this a const function?

src/iocore/cache/CacheVC.cc
```diff
@@ -535,6 +519,40 @@ LramHit: {
...
+{
+  if (*this->read_key == this->stripe->first_fragment_key && dir_offset(&this->dir) == this->stripe->first_fragment_offset) {
+    this->buf = this->stripe->first_fragment_data;
+    return true;
+  }
+  return false;
+}
+
+bool
+CacheVC::check_aggregation_buffer()
```
masaori335, line 542: I'm not good at naming, but this function does more than check, allocate new buffer and copy. Any other idea?

src/iocore/cache/CacheVC.cc
JosiahWI: Good catch.

include/iocore/cache/CacheVC.h
JosiahWI: Great idea!

src/iocore/cache/CacheVC.cc
JosiahWI: Maybe all those methods could be named something like `load_from_aggregation_buffer`. What do you think of that name? Do you have any ideas?

masaori335: `load_from_aggregation_buffer` makes sense 👍

include/iocore/cache/CacheVC.h
JosiahWI: However, they are not `const` methods, because they create or assign to the CacheVC buffer.

JosiahWI: The methods I added to `Stripe` and `AggregateWriteBuffer` should have been marked `const`, though. This has now been fixed.

masaori335: I missed that they're changing member variables. Let's leave it.

For the `Stripe` and `AggregateWriteBuffer`, marking `const` looks good.

masaori335 (approved):
Looks good.

PR #10988, opened 2024-01-16
header_freq: Fix msg lock issues

cmcfarlen:


JosiahWI:
Did clang-format get applied to the bundled yaml-cpp code accidentally?

maskit:
It's unfortunate that the commit log doesn't have the original commit hash, which could be added by `git cherry-pick -x`.

bneradt (approved):

bneradt:
> It's unfortunate that the commit log doesn't have the original commit hash, which could be added by `git cherry-pick -x`.

True. I usually `cherry-pick -x`. But I'm fine with replicating that in the squash commit if you simply give the commit on master you cherry-picked this from, @cmcfarlen .

bryancall:
The commit hash for #10917 is 78ccc61c9c803c39b3a874588fe2b3b2d52e9a0d

PR #10992, opened 2024-01-17
Restore IntrusiveHashMap insert code from tscore (ats92)

traeak:
This removes an extra snippet of code added with the libswoc implementation of IntrusiveHashMap which causes a noticeable performance hit (expansion of the critical section) when inserting items into the session pool.  Specific test case used saw on average an increase of about 6x in the HttpSessionManager critical sections.

The change introduced with libswoc (not documented in any checkin) appends the new item to the bucket list, which was not done with the tscore (ats92) version.

### DIFF
--- a/lib/swoc/include/swoc/IntrusiveHashMap.h
+++ b/lib/swoc/include/swoc/IntrusiveHashMap.h
@@ -514,15 +514,6 @@ IntrusiveHashMap<H>::insert(value_type *v) {
     if (spot != bucket->_v) {
       mixed_p = true; // found some other key, it's going to be mixed.
     }
-    if (spot != limit) {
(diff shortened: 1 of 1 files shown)


ywkaras (changes requested):
Before Alan's passing, the code in the lib/swoc tree was copied from labeled version in https://github.com/SolidWallOfCode/libswoc .  We have to decide where the maintained source is going to be, to make PRs against it.

traeak:
[approve ci debian]

traeak:
[approve ci autest]

bneradt (approved):
I verified that this block was added in libswoc and wasn't in tscore. I'm sure this was added for a reason, but it since it introduces serious origin side connection problems, I'm OK with backing it out at least for now.

bryancall (approved):

PR #10993, opened 2024-01-18
Refactor sync_cache_dir_on_shutdown

JosiahWI:
This moves the logic for syncing the stripe buffer to disk on shutdown into the `Stripe` class, since that class manages the buffer. It also does away with the temporary buffer used to write to disk, instead writing to disk directly from the stripe's buffer. The temporary buffer may have had some performance advantage or something some time ago. I don't see a reason for it now.

This also adds an `is_empty()` method to `AggregateWriteBuffer`, and uses it in the new `Stripe::shutdown` method as well as a few other places where appropriate.


masaori335 (approved):
Looks good to me. This removes unnecessary `memcpy`.

PR #10995, opened 2024-01-18
traffic_top: avoid abort when rpc node is not available.

brbzull0:
This way, `traffic_top` will just display the error when the node is not avialable.

### DIFF
--- a/src/traffic_top/stats.h
+++ b/src/traffic_top/stats.h
@@ -278,7 +278,7 @@ public:
     lookup_table.insert(make_pair("client_dyn_ka", LookupItem("Dynamic KA", "ka_total", "ka_count", 3)));
   }
 
-  void
+  bool
   getStats()
   {
     if (_url == "") {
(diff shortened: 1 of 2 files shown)


maskit:
So traffic_top stops anyway? It's better than aborting, but I wonder if it'd be useful if traffic_top could keep running. If I restart traffic_server while I'm watching the status on traffic_top, I have to run traffic_top again. It's not a big deal though.

brbzull0:
> So traffic_top stops anyway? It's better than aborting, but I wonder if it'd be useful if traffic_top could keep running. If I restart traffic_server while I'm watching the status on traffic_top, I have to run traffic_top again. It's not a big deal though.

I actually prefer that(what you propose).  With  some sort of status like `[offline]` or something like this.

maskit:
Yeah, showing that kind of status would make sense.

maskit (approved):
This is definitely an improvement. I'm fine with merging this as it is.

PR #10996, opened 2024-01-18
Move session pool metrics calls outside critical section

traeak:
Based on internal benchmarking this can cut time spent in the session pool critical section by up to 15%.

### DIFF
--- a/src/proxy/http/HttpSessionManager.cc
+++ b/src/proxy/http/HttpSessionManager.cc
@@ -419,6 +419,7 @@ HttpSessionManager::_acquire_session(sockaddr const *ip, CryptoHash const &hostn
 {
   PoolableSession *to_return = nullptr;
   HSMresult_t retval         = HSM_NOT_FOUND;
+  bool acquired              = false;
 
   // Extend the mutex window until the acquired Server session is attached
   // to the SM. Releasing the mutex before that results in race conditions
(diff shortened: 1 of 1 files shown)


traeak:
[approve ci autest]

src/proxy/http/HttpSessionManager.cc
```diff
@@ -491,6 +494,11 @@ HttpSessionManager::_acquire_session(sockaddr const *ip, CryptoHash const &hostn
       }
     }
   }
+
+  if (acquired) {
+    Metrics::Gauge::decrement(http_rsb.pooled_server_connections);
```
bneradt, line 499: Can you help me think through this? It seems like the decrement and increment calls are reversed in this patch, but maybe I'm not thinking through the logic correctly. Why do we decrement the stat when we acquire a session, but increment when we release it?

bneradt (changes requested):

src/proxy/http/HttpSessionManager.cc
traeak: i agree the logic is confusing.  The metric keeps track of the number of sessions waiting in the pool.
* A transaction acquiring the session removes it from the session pool. 
* A transaction releasing the session adds it to the session pool

traeak: also note i removed the the check for when a session is removed from the pool.  It could instead be an assert because a session is required to be in both the fqdn and ip hash maps.

bneradt (approved):
Looks good.

PR #10997, opened 2024-01-18
Unify Dynamic Table implementation

### LINKED ISSUE #10056: Unify DynamicTables in HPACK and QPACK
There is a lot of similarity in the both Dynamic Tables, and using `MIMEHdr` doesn't seem like a great idea (lots of overhead). We should implement a DynamicTable that can be used for the both HPACK and QPACK without using `MIMEHdr`.

maskit:
This closes #10056.

We have had two Dynamic Table implementations, one for HPACK and the other for QPACK. This PR basically removes the one for HPACK, and move the one for QPACK into `proxy/hdr` so that they can use the same implementation.

While I'm testing it with HPACK, I found a couple of bugs in the new dynamic table implementation and I fixed them (we couldn't find it because QPACK has never worked and we disabled it on the docs server).

The one for HPACK uses HTTPHdr structure under the hood, and it has been troublesome because of the performance and its specialized interface which doesn't suit the Dynamic Table use case.

### DIFF
--- a/src/proxy/hdrs/unit_tests/test_XPACK.cc
+++ b/src/proxy/hdrs/unit_tests/test_XPACK.cc
@@ -125,4 +125,186 @@ TEST_CASE("XPACK_String", "[xpack]")
       REQUIRE(memcmp(actual, i.raw_string, actual_len) == 0);
     }
   }
+
+  SECTION("Dynamic Table")
+  {
+    constexpr uint16_t MAX_SIZE = 128;
+    XpackDynamicTable dt(MAX_SIZE);
+    XpackLookupResult result;
+    const char *name  = nullptr;
+    size_t name_len   = 0;
+    const char *value = nullptr;
+    size_t value_len  = 0;
+
+    // Check the initial state
+    REQUIRE(dt.size() == 0);
+    REQUIRE(dt.maximum_size() == MAX_SIZE);
+    REQUIRE(dt.is_empty());
+    REQUIRE(dt.count() == 0);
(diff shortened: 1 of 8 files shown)


src/proxy/hdrs/unit_tests/test_XPACK.cc
```diff
@@ -125,4 +125,186 @@ TEST_CASE("XPACK_String", "[xpack]")
...
+    result = dt.lookup(1, &name, &name_len, &value, &value_len);
+    REQUIRE(result.match_type == XpackLookupResult::MatchType::NONE);
+    result = dt.lookup(MAX_SIZE - 1, &name, &name_len, &value, &value_len);
+    REQUIRE(result.match_type == XpackLookupResult::MatchType::NONE);
+    result = dt.lookup(MAX_SIZE, &name, &name_len, &value, &value_len);
+    REQUIRE(result.match_type == XpackLookupResult::MatchType::NONE);
+    result = dt.lookup(MAX_SIZE + 1, &name, &name_len, &value, &value_len);
+    REQUIRE(result.match_type == XpackLookupResult::MatchType::NONE);
+
+    // Insdert one entry
```
bneradt, line 155: "Insdert"

bneradt (changes requested):

bneradt (approved):
This makes sense generally. 

@maskit : can you let me know when this lands? Let's get this on docs as soon as it does to get some coverage there.

bneradt (approved):

PR #11000, opened 2024-01-23
Doc: storage.yaml

masaori335:
Draft PR for discuss details of making `volume.config` and `storage.config` into one YAML file.

Most of the docs were written by @SolidWallOfCode last year and I made some changes from his draft. 

Differences are

- Name the config file, I propose `storage.yaml` instead of `cache.yaml`
- Make the volume id type `integer`, because it's referred from `hosting.config` ( we can change it `string` later, but I'd like to leave it as is for now )
- Remove percentage support from `span[].size`

I'll open another PR for implementation.


doc/admin-guide/files/storage.yaml.en.rst
```diff
@@ -0,0 +1,482 @@
...
+     spans:         #
+       - id:        # identifier / name of the span
+         path:      # path to storage
+         size:      # size in bytes, required for file system storage, optional for raw device
+         hash_seed: # optional, used to isolate lookup from path changes
+     volumes:       # optional
+       - id:        # identifier [1-255]
+         size:      # optional, size in percentage
+         scheme:    # optional, default to "http"
+         ram-cache: # optional, default to "true"
```
masaori335, line 46: Should be `ram_cache` to follow the naming convention.

doc/admin-guide/files/storage.yaml.en.rst
```diff
@@ -0,0 +1,482 @@
...
+   If this is not sufficient then the :arg:`hash_seed` key should be used to create a more permanent
+   assignment table. An example would be
+
+   .. code-block: yaml
+
+   cache:
+     spans:
+       - id: "span.0"
+         path: "/dev/sde"
+         hash-seed: "cache.disk.0"
```
masaori335, line 280: `hash_seed`

doc/admin-guide/files/storage.yaml.en.rst
```diff
@@ -0,0 +1,482 @@
...
+Format
+======
+
+The format of the :file:`storage.yaml` file is a series of lines of the form
+
+.. code-block:: yaml
+
+   cache:           # file level key
+     spans:         #
+       - id:        # identifier / name of the span
```
masaori335, line 38: This is very bike-shedding, but calling this `name` is better than `id`?

PR #11011, opened 2024-01-25
Fix esi-fetcher-plugin libswoc include issue

phongn:
We discovered some linking issues with use of an external libswoc; this fully qualifies inclusion so that internal or external should be usable.

### DIFF
--- a/plugins/esi/fetcher/CMakeLists.txt
+++ b/plugins/esi/fetcher/CMakeLists.txt
@@ -16,6 +16,6 @@
 #######################
 
 add_library(fetcher STATIC HttpDataFetcherImpl.cc)
-target_link_libraries(fetcher PUBLIC esi-common libswoc)
-target_include_directories(fetcher PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}")
+target_link_libraries(fetcher PUBLIC esi-common libswoc::libswoc)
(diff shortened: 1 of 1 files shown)


plugins/esi/fetcher/CMakeLists.txt
```diff
@@ -16,6 +16,6 @@
 #######################
 
 add_library(fetcher STATIC HttpDataFetcherImpl.cc)
-target_link_libraries(fetcher PUBLIC esi-common libswoc)
-target_include_directories(fetcher PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}")
+target_link_libraries(fetcher PUBLIC esi-common libswoc::libswoc)
+target_include_directories(fetcher PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}" "${libswoc_INCLUDE_DIRS}")
```
JosiahWI, line 20: We should not be both linking the target and explicitly dumping its `INCLUDE_DIRS` into `target_include_directories`, as linking the target should already do this.

JosiahWI (changes requested):

JosiahWI (commented):
Looks good.

cmcfarlen (approved):

JosiahWI (approved):

PR #10857, opened 2023-11-28
Coverity 1497378: Use of 32-bit time_t in CacheVC

bryancall:
In `CacheVC::set_pin_in_cache()` `time_pin` of type `time_t` is being assigned to a `pin_in_cache` which was an `uint32_t`.

PR #11013, opened 2024-01-25
Revert "Coverity 1497378: Use of 32-bit time_t in CacheVC (#10857)"

bneradt:
On 64 bit systems, time_t (which is an unsigned long) is generally a 64 bit value. We therefore may not change the type of pin_in_cache from an explicit 32 bit value to a 64 bit one without introducing a cache inconsistency in ATS 10. Since the community is a hard pass on introducing an inconsistency for this release, we have to revert this change.

This reverts commit 6b896f43bef38c42d574d8bc89293766e0dbc94f.

### DIFF
--- a/include/iocore/cache/CacheVC.h
+++ b/include/iocore/cache/CacheVC.h
@@ -273,7 +273,7 @@ struct CacheVC : public CacheVConnection {
   Event *trigger;
   CacheKey *read_key;
   ContinuationHandler save_handler;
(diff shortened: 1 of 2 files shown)


masaori335:
The `pin_in_cache` is a member of `CacheVC`. Does the type change affect the cache compatibility?

masaori335 (approved):
This revert is harmless. I'll check how this breaks the compatibility later.

PR #11017, opened 2024-01-27
stale_response: RFC 5861 Cache-Control Directives

bneradt:
This adds a new experimental plugin called stale_response.so. It implements the stale-while-revalidate and stale-if-error Cache-Control directives that are described in RFC 5861.

---

# For Review

The rendered doc for this new plugin is visible here (until CI cleans it up):
https://ci.trafficserver.apache.org/job/Github_Builds/job/docs/4268/artifact/output/11017/docbuild/html/admin-guide/plugins/stale_response.en.html

### DIFF
--- a/doc/admin-guide/plugins/stale_response.en.rst
+++ b/doc/admin-guide/plugins/stale_response.en.rst
@@ -1,0 +1,118 @@
+.. Licensed to the Apache Software Foundation (ASF) under one
+   or more contributor license agreements.  See the NOTICE file
+   distributed with this work for additional information
+   regarding copyright ownership.  The ASF licenses this file
+   to you under the Apache License, Version 2.0 (the
+   "License"); you may not use this file except in compliance
+   with the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing,
+   software distributed under the License is distributed on an
+   "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+   KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+   under the License.
+
+
+.. include:: ../../common.defs
+
+.. _admin-plugins-stale_response:
+
+
+Stale Response Plugin
+*********************
+
+This plugin adds ``stale-while-relavidate`` and ``stale-if-error``
+``Cache-Control`` directive functionality.  The ``stale-while-revalidate``
+directive specifies a number of seconds that an item can be stale and used as a
+cached response while |TS| revalidates the cached item. The ``stale-if-error``
+directive specifies how stale a cached response can be and still be used as a
+response if the origin server replies with a 500, 502, 503, or 504 HTTP status.
+The plugin currently only supports the ``stale-if-error`` directive in
+responses. For more details, see `RFC 5861
+<https://www.rfc-editor.org/rfc/rfc5861>`_ for the specification of these
+directives.
+
+Building and Installation
+*************************
+
+The Stale Response plugin is an experimental plugin. To build it, pass
+``-DBUILD_EXPERIMENTAL_PLUGINS=ON`` to the ``cmake`` command when building |TS|.
+By default, that will build the ``stale_response.so`` plugin and install it in
+the ``libexec/trafficserver`` directory.
+
+Configuration
+*************
+
+The Stale Response plugin supports being used as either a global plugin or as a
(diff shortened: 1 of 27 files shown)


doc/admin-guide/plugins/stale_response.en.rst
```diff
@@ -0,0 +1,118 @@
...
+
+Building and Installation
+*************************
+
+The Stale Response plugin is an experimental plugin. To build it, pass
+``-DBUILD_EXPERIMENTAL_PLUGINS=ON`` to the ``cmake`` command when building |TS|.
+By default, that will build the ``stale_response.so`` plugin and install it in
+the ``libexec/trafficserver`` directory.
+
+Configuraiton
```
randall, line 46: Spelling. Configuration

bneradt: Good catch!

bneradt:
[approve ci]

plugins/experimental/stale_response/ts_wrap.h
```diff
@@ -0,0 +1,42 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+
+#pragma once
+
+#include "ts/ts.h"
+
+#if TS_VERSION_MAJOR >= 10
```
cmcfarlen, line 28: This will always be true.

plugins/experimental/stale_response/BodyData.h
```diff
@@ -0,0 +1,118 @@
...
+/*-----------------------------------------------------------------------------------------------*/
+inline BodyData::BodyData()
+{
+  chunk_list.reserve(c_chunk_add_count);
+}
+
+/*-----------------------------------------------------------------------------------------------*/
+inline bool
+BodyData::addChunk(const char *start, int64_t size)
+{
```
cmcfarlen, line 84: could debug_assert check valid arguments here

plugins/experimental/stale_response/BodyData.h
```diff
@@ -0,0 +1,118 @@
...
+  chunk_list.emplace_back(size, std::vector<char>(start, start + size));
+  total_size += size;
+  return true;
+}
+
+/*-----------------------------------------------------------------------------------------------*/
+inline bool
+BodyData::getChunk(uint32_t chunk_index, const char **start, int64_t *size) const
```
cmcfarlen, line 92: and here

plugins/experimental/stale_response/CacheUpdate.cc
```diff
@@ -0,0 +1,374 @@
...
+/*-----------------------------------------------------------------------------------------------*/
+BodyData *
+intercept_check_request(StateInfo *state)
+{
+  uint32_t newKey      = 0;
+  BodyData *pBodyFound = nullptr;
+  TSHttpTxn txnp       = state->txnp;
+  uint32_t oldKey      = state->req_info->key_hash;
+
+#if (TS_VERSION_NUMBER >= 7000000)
```
cmcfarlen, line 274: This is always true

cmcfarlen (changes requested):
Could you go through and remove the conditional compiles for pre-10 versions?  Since this will live in the tree now, those are not necessary.  ts_wrap.h can be simplified.

Otherwise, this looks just dandy to me.

plugins/experimental/stale_response/ts_wrap.h
bneradt: My intention was to leave this version conditional incase someone wants to backport it, either in ASF or internal. We in Yahoo, for instance, need it internally for 9.x. Does that make sense? I'm convincible if you'd still like me to remove this.

My comment here applies to comparison of major version to 10 and above. Your other comments about older versions I'll definitely take out. Those were here because this is an ASF port of an old internal plugin.

plugins/experimental/stale_response/CacheUpdate.cc
bneradt: Good catch. I missed this when cleaning these out.

plugins/experimental/stale_response/BodyData.h
bneradt: Good idea.

plugins/experimental/stale_response/BodyData.h
bneradt: Good idea times 2.

cmcfarlen (approved):

zwoop (approved):

PR #11018, opened 2024-01-29
Cleanup: Rename key_to_vol

masaori335:
Part of https://github.com/apache/trafficserver/issues/10628.

Rename `key_to_vol` to `key_to_stripe` because it's returning `Stripe *`.


src/iocore/cache/Cache.cc
```diff
@@ -1810,7 +1810,7 @@ rebuild_host_table(Cache *cache)
 
 // if generic_host_rec.vols == nullptr, what do we do???
```
brbzull0, line 1811: nit: gen_host_rec`.stripes` instead of `.vol`?

brbzull0 (approved):
Thanks :+1:

src/iocore/cache/Cache.cc
masaori335: Nice catch! Done.

brbzull0 (approved):

PR #11021, opened 2024-01-29
Ensure connection retry attempts can reach the config specified value

### LINKED ISSUE #11006: [9.2.x] Child not marking parent down when parent is not running
I noticed that when the parent is not running, the child is unable to mark the parent down. Based on default configs, the child should retry the connection to the parent three times ([proxy.config.http.connect_attempts_rr_retries](https://docs.trafficserver.apache.org/admin-guide/files/records.config.en.html#proxy-config-http-connect-attempts-rr-retries)) then if it is not successful, mark the parent down. However at the moment it is unable to try the connection three times and instead returns the 502 error on the first retry. 

To reproduce,

* Set up ats with a parent/child config
* Make an call to ATS when the parent is not running

jasmine-nahrain:
With the current logic, when `s->current.attempts = 1`, the if statement on line 3695 will always be false because 0 % anything is 0.
Removing the -1 so the if statement will only be false when the retry attempts number is reached.
Should close #11006

### DIFF
--- a/proxy/http/HttpTransact.cc
+++ b/proxy/http/HttpTransact.cc
@@ -3692,7 +3692,7 @@ HttpTransact::handle_response_from_parent(State *s)
       s->current.attempts++;
 
       // Are we done with this particular parent?
-      if ((s->current.attempts - 1) % s->txn_conf->per_parent_connect_attempts != 0) {
+      if (s->current.attempts % s->txn_conf->per_parent_connect_attempts != 0) {
         // No we are not done with this parent so retry
         HTTP_INCREMENT_DYN_STAT(http_total_parent_switches_stat);
         s->next_action = how_to_open_connection(s);


masaori335:
This is a 9.2.x specific issue and the master doesn't have it because of some refactoring, right?

bryancall:
This is related to #9620 that landed on master.

proxy/http/HttpTransact.cc
```diff
@@ -3692,7 +3692,7 @@ HttpTransact::handle_response_from_parent(State *s)
       s->current.attempts++;
 
       // Are we done with this particular parent?
-      if ((s->current.attempts - 1) % s->txn_conf->per_parent_connect_attempts != 0) {
+      if (s->current.attempts % s->txn_conf->per_parent_connect_attempts != 0) {
```
ywkaras, line 3695: Seems to match ATS 10:
https://github.com/apache/trafficserver/blob/f180ea2b6f25f6922c881b60d70cef68db1acac5/src/proxy/http/HttpTransact.cc#L3617

ywkaras (approved):

PR #11024, opened 2024-01-30
Reuse arena for header compression

maskit:
Having an arena in each function is wasteful and it's effectively allocating memory at all the places on every call. One arena for each connection should be enough.

### DIFF
--- a/include/proxy/http2/HPACK.h
+++ b/include/proxy/http2/HPACK.h
@@ -126,6 +126,9 @@ public:
   uint32_t size() const;
   void update_maximum_size(uint32_t new_size);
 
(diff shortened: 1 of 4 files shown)


brbzull0 (commented):
Changes looks good, just wondering if we should land this now or after https://github.com/apache/trafficserver/pull/10979

bryancall:
@brbzull0 It is ok to approve and land this PR before #10979

brbzull0 (approved):
Thanks!

brbzull0:
> @brbzull0 It is ok to approve and land this PR before #10979

sure! I didn't meant to block it! apporved! Thanks.

PR #11026, opened 2024-01-31
Move Doc struct to Doc.h

JosiahWI:
This splits out the `Doc` struct to its own header. The reason for doing this is to use the struct in `AggregateWriteBuffer.h` without introducing a direct circular dependency on `P_CacheVol.h`, but it seems good even as a standalone change so I'm PRing it separately. I moved some `#define`s that went along with the `Doc` struct as well.

The second commit in this PR adds a `self_type` alias to `Doc`. This might not have been necessary; I was looking for various things to clean up and did it while I was messing around.


masaori335 (dismissed):
Looks good.

masaori335:
[approve ci autest]

JosiahWI:
I heard some concerns from @cmcfarlen about putting moving this to a "public" header when it was in a "private" one before.

cmcfarlen (approved):

PR #11028, opened 2024-02-01
docs: Fix pipfile package compatible release clause.

brbzull0:
For some reason I have this failing locally(rocky8), looking at the docs, this seems the right way to specify the dependency.

```
>= V.N, == V.*
```
as we use a wildcard, then `==` seems the correct way.

https://peps.python.org/pep-0440/#compatible-release


masaori335 (approved):
Nice catch!

PR #11030, opened 2024-02-01
Allows the set-body to run as a pseudo remap hook

zwoop:
It also eliminates the hook support for read-response, which is too late for this implementaiton of the set-body.


ezelkow1 (approved):

zwoop:
Cherry picked to 9.2.x

PR #11031, opened 2024-02-01
Fixing remap acl rules for @action=allow

bneradt:
If a user specifies an `@action=allow` remap.config ACL rule, then the implication is that requests with methods not in the allow list would be denied. Before this patch, allow ACL rules would just never deny. This fixes that behavior so that allow ACL rules that match on IP but not on method deny.

### DIFF
--- a/src/proxy/http/remap/UrlRewrite.cc
+++ b/src/proxy/http/remap/UrlRewrite.cc
@@ -399,7 +399,7 @@ UrlRewrite::ReverseMap(HTTPHdr *response_header)
 void
 UrlRewrite::PerformACLFiltering(HttpTransact::State *s, url_mapping *map)
 {
-  if (unlikely(!s || s->acl_filtering_performed || !s->client_connection_enabled)) {
+  if (unlikely(!s || s->acl_filtering_performed || !s->client_connection_allowed)) {
     return;
   }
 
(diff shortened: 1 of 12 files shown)


src/proxy/http/remap/UrlRewrite.cc
```diff
@@ -411,43 +411,50 @@ UrlRewrite::PerformACLFiltering(HttpTransact::State *s, url_mapping *map)
...
+          method_matches = rp->standard_method_lookup[method_wksidx];
         } else if (!rp->nonstandard_methods.empty()) {
-          match = false;
+          method_matches = false;
         } else {
           int method_str_len;
           const char *method_str = s->hdr_info.client_request.method_get(&method_str_len);
-          match                  = rp->nonstandard_methods.count(std::string(method_str, method_str_len));
+          method_matches         = rp->nonstandard_methods.count(std::string(method_str, method_str_len));
         }
```
bneradt, line 429: The other changes in this patch are test, comment, or variable rename changes. This is where the functional change happens. It simply does this:

1. Distinguish between method matches (`method_matches`) and IP matches `ip_matches`).
2. If `ip_matches`, then the rule applies and `method_matches` will apply based upon whether it is a deny or allow rule.

Without this change, we don't enforce allow rules for non-matched methods. Which defeats the purpose of an allow rule.

bneradt:
[approve ci autest]

bneradt:
After playing with this more, this got replaced by:
https://github.com/apache/trafficserver/pull/11033

This PR is therefore no longer necessary.

PR #11034, opened 2024-02-02
Don't statically link libtsutil.a into core plugins.

ywkaras:
Two equivalent libraries are now created, libtsutil.a and libtsutil_link_dummy.so, with the same code and symbol content.  Non-core plugins can be linked with libtsutil_link_dummy.so, with the --no-undefined ld option.  The drawback is, for dlopen() to successfully load the plugin, it will require the presence of libtsutil_link_dummy.so.  But, the symbols in libtsutil_link_dummy.so don't seem to be used, because they are satisfied by symbols from core TS.  I tested this (on Red Hat 8) by substituting a random .so for libtsutil_link_dummy.so, before successfully loading the xdebug.so plugin.

I checked for an option in ld to resolve symbols for the link in a .so, without creating a dependency on it.  But I could not find any such option.

### DIFF
--- a/cmake/add_atsplugin.cmake
+++ b/cmake/add_atsplugin.cmake
@@ -19,7 +19,7 @@ set(CMAKE_SHARED_LIBRARY_PREFIX "")
 
 function(add_atsplugin name)
   add_library(${name} MODULE ${ARGN})
-  target_link_libraries(${name} PRIVATE ts::tsapi ts::tsutil)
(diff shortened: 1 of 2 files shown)


ywkaras:
[approve ci autest]

ywkaras:
[approve ci format]

ywkaras:
[approve ci centos]

ywkaras:
[approve ci autest]

bneradt:
[approve ci autest]

ywkaras:
[approve ci autest]

PR #11039, opened 2024-02-05
Make linking libraries to plugins optional

cmcfarlen:


cmake/add_atsplugin.cmake
```diff
@@ -19,7 +19,14 @@ set(CMAKE_SHARED_LIBRARY_PREFIX "")
 
 function(add_atsplugin name)
   add_library(${name} MODULE ${ARGN})
-  target_link_libraries(${name} PRIVATE ts::tsapi ts::tsutil)
+  if(LINK_PLUGINS)
```
bryancall, line 22: Only do this if we are also building with BUILD_SHARED_LIBS

bryancall (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11041, opened 2024-02-05
Add basic check for pcre2

dragon512:
Allows this to be reported correctly when we are configuring cmake.

once @bryancall adds official pcre2 fixes, this check will want to be tweaked to allow us to pick prce1 or pcre2 but not both at the same time.

This should help with the work with txn_box @brbzull0 is working on and the official pcre2  support @bryancall is working on

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -246,6 +246,7 @@ if(LibLZMA_FOUND)
 endif()
 
 find_package(PCRE REQUIRED)
+find_package(PCRE2 COMPONENTS 8BIT)
 
 include(CheckOpenSSLIsBoringSSL)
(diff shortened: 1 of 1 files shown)


cmcfarlen:
PCRE2 is a cmake project and support find_package directly.  We don't need to provide our own find package script for this.  It should work with just `find_package(pcre2 REQUIRED)`

dragon512:
> PCRE2 is a cmake project and support find_package directly. We don't need to provide our own find package script for this. It should work with just `find_package(pcre2 REQUIRED)`

Ok... let me change this and test that this work as stated as reports out which pcre2 is being used

bneradt:
[approve ci autest]

cmcfarlen (approved):

PR #11046, opened 2024-02-07
Add autest to make sure H2 grace shutdown is working as intended

duke8253:
This PR addresses a few issues:

~- If a `HTTP2_SESSION_EVENT_FINI` is received after a `HTTP2_SESSION_EVENT_SHUTDOWN_INIT`, the `GOAWAY` frame with the correct `last_stream_id` isn't sent out.~
- ~Correctly set and remove the `Connection: close` header for HTTP/2 (this header is only used as a internal flag to indicate a grace shutdown is needed).~
- ~Initiate a grace shutdown when status is 429.~
- Add test to make sure grace shutdown is working as intended.

### DIFF
--- a/tests/gold_tests/h2/http2_close_connection.test.py
+++ b/tests/gold_tests/h2/http2_close_connection.test.py
@@ -1,0 +1,68 @@
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import os
+
+Test.Summary = '''
+Check whether the HTTP/2 grace shutdown works as intended.
+'''
+
+Test.SkipUnless(Condition.HasProxyVerifierVersion('2.8.0'))
+
+pv_server = Test.MakeVerifierServerProcess("pv_server", "http2_close_connection.yaml")
+
+ts = Test.MakeATSProcess('ts', select_ports=True, enable_tls=True)
+
+ts.addDefaultSSLFiles()
+ts.Disk.ssl_multicert_config.AddLine("dest_ip=* ssl_cert_name=server.pem ssl_key_name=server.key")
+ts.Disk.records_config.update(
+    {
+        "proxy.config.http.server_ports": f"{ts.Variables.port} {ts.Variables.ssl_port}:ssl",
+        'proxy.config.ssl.server.cert.path': f'{ts.Variables.SSLDir}',
+        'proxy.config.ssl.server.private_key.path': f'{ts.Variables.SSLDir}',
+        'proxy.config.ssl.client.verify.server.policy': 'PERMISSIVE',
+        'proxy.config.diags.debug.enabled': 3,
+        'proxy.config.diags.debug.tags': 'http',
+        'proxy.config.exec_thread.autoconfig.enabled': 0,
+        'proxy.config.exec_thread.limit': 4,
+        'proxy.config.ssl.client.alpn_protocols': 'h2,http/1.1',
+        'proxy.config.http.server_session_sharing.pool': 'thread',
+        'proxy.config.http.server_session_sharing.match': 'ip,sni,cert',
+    })
+
+ts.Disk.remap_config.AddLines([f'map / https://127.0.0.1:{pv_server.Variables.https_port}'])
+
+Test.PrepareTestPlugin(os.path.join(Test.Variables.AtsTestPluginsDir, 'http2_close_connection.so'), ts)
+
+tr = Test.AddTestRun()
+tr.Processes.Default.StartBefore(pv_server)
+tr.Processes.Default.StartBefore(ts)
+tr.AddVerifierClientProcess(
+    "pv_client",
+    "http2_close_connection.yaml",
+    http_ports=[ts.Variables.port],
+    https_ports=[ts.Variables.ssl_port],
+    other_args='--thread-limit 1')
+tr.Processes.Default.ReturnCode = 0
+
+tr.Processes.Default.Streams.All += Testers.ContainsExpression(
+    'Equals Success: Key: "1", Content Data: "body", Value: "server_test_1"', 'Response check')
+tr.Processes.Default.Streams.All += Testers.ContainsExpression(
+    'Received GOAWAY frame with last stream id 2147483647, error code 0', 'initial GOAWAY frame with last stream id set to max')
+tr.Processes.Default.Streams.All += Testers.ContainsExpression(
+    'Received GOAWAY frame with last stream id 1, error code 0', 'updated GOAWAY frame with last stream id set to 1')
+
+pv_server.Streams.All += Testers.ExcludesExpression('server_test_2', 'Only one response should be sent')
--- a/tests/gold_tests/h2/http2_close_connection.yaml
+++ b/tests/gold_tests/h2/http2_close_connection.yaml
@@ -1,0 +1,100 @@
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+#
+# This replay file assumes that caching is enabled and
+# proxy.config.http.cache.ignore_client_cc_max_age is set to 0 so that we can
+# test max-age in the client requests.
+#
+
+
+meta:
+  version: '1.0'
+sessions:
+- protocol:
+  - name: http
+    version: 2
+  - name: tls
+    sni: test_sni
+  - name: tcp
+  - name: ip
+    version: 4
+  transactions:
+  - client-request:
+      frames:
+      - HEADERS:
+          headers:
+            fields:
+            - [:method, GET]
+            - [:scheme, https]
+            - [:authority, example.data.com]
+            - [:path, /a/path]
+            - [Content-Type, text/html]
+            - [uuid, 1]
+
+    server-response:
+      frames:
+      - HEADERS:
+          headers:
+            fields:
+            - [:status, 200]
+            - [Content-Type, text/html]
+            - [Content-Length, '13']
+      - DATA:
+          delay: 4s
+          content:
+            encoding: plain
+            data: server_test_1
+            size: 13
+
+    proxy-response:
+      content:
+        encoding: plain
+        data: server_test_1
+        verify: {as: equal}
+
+  - client-request:
+      delay: 4s
+      frames:
+      - HEADERS:
+          headers:
+            fields:
+            - [:method, GET]
+            - [:scheme, https]
+            - [:authority, example.data.com]
+            - [:path, /b/path]
+            - [Content-Type, text/html]
+            - [uuid, 2]
+
+    server-response:
+      frames:
+      - HEADERS:
+          headers:
+            fields:
+            - [:status, 200]
+            - [Content-Type, text/html]
+            - [Content-Length, '13']
+      - DATA:
+          content:
+            encoding: plain
+            data: server_test_2
+            size: 13
+
+    proxy-response:
+      content:
+        encoding: plain
+        data: server_test_2
+        verify: {as: equal}
--- a/tests/tools/plugins/CMakeLists.txt
+++ b/tests/tools/plugins/CMakeLists.txt
@@ -37,6 +37,7 @@ add_autest_plugin(user_args user_args.cc)
 add_autest_plugin(async_engine async_engine.c)
 add_autest_plugin(hook_tunnel_plugin hook_tunnel_plugin.cc)
 add_autest_plugin(tunnel_transform tunnel_transform.cc)
+add_autest_plugin(http2_close_connection http2_close_connection.cc)
 
 target_link_libraries(continuations_verify PRIVATE OpenSSL::SSL)
 target_link_libraries(ssl_client_verify_test PRIVATE OpenSSL::SSL)
--- a/tests/tools/plugins/http2_close_connection.cc
+++ b/tests/tools/plugins/http2_close_connection.cc
@@ -1,0 +1,92 @@
+/** @file
+
+  Test adding continuation from same hook point
+
+  @section license License
+
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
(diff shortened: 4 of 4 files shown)


src/proxy/http2/Http2Stream.cc
```diff
@@ -802,16 +802,13 @@ Http2Stream::update_write_request(bool call_update)
...
-        const char *value = field->value_get(&len);
-        if (memcmp(HTTP_VALUE_CLOSE, value, HTTP_LEN_CLOSE) == 0) {
-          SCOPED_MUTEX_LOCK(lock, _proxy_ssn->mutex, this_ethread());
-          if (connection_state.get_shutdown_state() == HTTP2_SHUTDOWN_NONE) {
-            connection_state.set_shutdown_state(HTTP2_SHUTDOWN_NOT_INITIATED, Http2ErrorCode::HTTP2_ERROR_NO_ERROR);
-          }
+      // Schedule session shutdown if response header has "Connection: close",
+      // then remove "Connection: close" which is a protocol violation for HTTP/2.
+      if (!this->_send_header.is_keep_alive_set()) {
+        this->_send_header.field_delete(MIME_FIELD_CONNECTION, MIME_LEN_CONNECTION);
```
maskit, line 808: Do we really need to delete the header here? I think it (and other headers that are invalid for H2) will be deleted by `VersionConverter` later.

src/proxy/http2/Http2ClientSession.cc
```diff
@@ -191,16 +191,22 @@ Http2ClientSession::main_event_handler(int event, void *edata)
     break;
 
   case VC_EVENT_ACTIVE_TIMEOUT:
-  case VC_EVENT_INACTIVITY_TIMEOUT:
+  case VC_EVENT_INACTIVITY_TIMEOUT: {
+    if (this->connection_state.get_shutdown_state() == HTTP2_SHUTDOWN_NONE) {
+      this->connection_state.set_shutdown_state(HTTP2_SHUTDOWN_INITIATED, Http2ErrorCode::HTTP2_ERROR_NO_ERROR);
```
maskit, line 196: I may not understand the issue, but the state should not be changed directly here. It should be set by HTTP2_SESSION_EVENT_SHUTDOWN_INIT event handler. Changing the state at multiple places is going to be super confusing.

src/proxy/http2/Http2Stream.cc
duke8253: I think it does get deleted, added that mostly for testing purpose. I'll remove that.

duke8253:
> > If a HTTP2_SESSION_EVENT_FINI is received after a HTTP2_SESSION_EVENT_SHUTDOWN_INIT, the GOAWAY frame with the correct last_stream_id isn't sent out.
> > If a HTTP/2 session timed out, no grace shutdown is triggered.
> 
> Aren't these immediate close? Why do we want to initiate grace shutdown on timeout?

We got a complaint from a property saying they don't have a way to determine whether a connection is closed, since the HTTP/2 doesn't have the `Connection: close` header, sending a `GOAWAY` when that happens should give them what they want. 

About that timeout part, was doing some testing and wasn't sure whether we need that, left it in there to get some comments.

maskit:
Ok, I wonder why the property can't check if the connection is closed on TCP layer, but let's leave it aside.

The RFC says:

> A server that is attempting to gracefully shut down a connection SHOULD send an initial GOAWAY frame with the last stream identifier set to 231-1 and a [NO_ERROR](https://datatracker.ietf.org/doc/html/rfc9113#NO_ERROR) code. This signals to the client that a shutdown is imminent and that initiating further requests is prohibited. After allowing time for any in-flight stream creation (at least one round-trip time), the server MAY send another GOAWAY frame with an updated last stream identifier. This ensures that a connection can be cleanly shut down without losing requests.

and

> If a connection terminates without a GOAWAY frame, the last stream identifier is effectively the highest possible stream identifier.

So, a client cannot expect a GOAWAY. A connection close without GOAWAY can happen. I don't think current behavior of ATS is wrong.

That said,  the RFC also says:

> An endpoint MAY send multiple GOAWAY frames if circumstances change. For instance, an endpoint that sends GOAWAY with [NO_ERROR](https://datatracker.ietf.org/doc/html/rfc9113#NO_ERROR) during graceful shutdown could subsequently encounter a condition that requires immediate termination of the connection. The last stream identifier from the last GOAWAY frame received indicates which streams could have been acted upon.

I'd take this as "A server can cancel graceful shutdown and send a GOAWAY without waiting for the 1 RTT of grace time, to close a connection immediately". This may be just a different way of saying what you do on this PR, but I think this is what we should do, and I'd take a different approach.

In most cases, `send_goaway_frame()` is called right before we schedule `HTTP2_SESSION_EVENT_FINI` event. So if ATS receives `HTTP2_SESSION_EVENT_FINI`, GOAWAY should be sent. I don't think we need to forcibly progress the shutdown process (we can simply cancel it) to send the second GOAWAY. The only exception is `Http2ClientSession::do_io_close`. It schedules the event without calling `send_goaway_frame()`. But if the event is scheduled in `do_io_close`, it's probably too late to send something anyway.

TLDR; Isn't canceling graceful shutdown process enough?

src/proxy/http2/Http2ClientSession.cc
shinrich: Will sending this with the SHUTDOWN_CONT event below trigger SHUTDOWN_INIT?  Or are you suggesting that this case should send the SHUTDOWN_INIT event instead and rely on that flow to set the appropriate state?

src/proxy/http2/Http2ConnectionState.cc
```diff
@@ -1434,6 +1451,14 @@ Http2ConnectionState::main_event_handler(int event, void *edata)
 
     ink_assert(this->fini_received == false);
     this->fini_received = true;
+
+    if (shutdown_state == HTTP2_SHUTDOWN_INITIATED) {
+      if (shutdown_cont_event) {
+        shutdown_cont_event->cancel();
+      }
+      this->_shutdown_cont();
+    }
```
shinrich, line 1460: So cleaning up the in progress shutdown events, since we just shutdown cleanly here?

src/proxy/http2/Http2ClientSession.cc
maskit: I was suggesting scheduling SHUTDOWN_INIT event instead. But if canceling grace shutdown works we don't need it.

src/proxy/http2/Http2ConnectionState.cc
duke8253: Yes, since we already started a grace shutdown, I think we should finish the whole process.

duke8253:
@maskit forgot to mention one thing, when doing a `curl` command, if no `GOAWAY` frame is sent, it reports `connection left intact` even when ATS says the H2 session is destroyed. I'm not sure that is the correct behavior, or should we even care about that?

maskit:
Closing a connection without sending GOAWAY is not polite, but I don't think it's a violation. Depends on why the connection was closed.

> Endpoints SHOULD always send a GOAWAY frame before closing a connection so that the remote peer can know whether a stream has been partially processed or not. For example, if an HTTP client sends a POST at the same time that a server closes a connection, the client cannot know if the server started to process that POST request if the server does not send a GOAWAY frame to indicate what streams it might have acted on.

> An endpoint might choose to close a connection without sending a GOAWAY for misbehaving peers.

include/proxy/http2/Http2Stream.h
```diff
@@ -80,6 +80,7 @@ class Http2Stream : public ProxyTransaction
   void set_expect_send_trailer() override;
   bool expect_receive_trailer() const override;
   void set_expect_receive_trailer() override;
+  void set_close_connection(HTTPHdr &hdr) const override;
```
maskit, line 83: Unused

maskit: Ah, nvm. I just realized it's override.

maskit: Does this relay Connection header value between a request and a response? I think that should not happen.

```
Client <-- (H2) -- ATS <-- (H1) -- Origin
```

When an origin server returns `Connection: close`, ATS should not close the connection between the client and ATS, right? It's just the origin wants to close the connection between ATS and the origin. The client should be allowed to make another request, which can be served from the cache, on the same connection.

duke8253: This sets the `Connection: close`, but it doesn't decide whether the connection will be closed. `HttpTransact::handle_response_keep_alive_headers` actually decides whether that close will remain in the response or be removed. But before we never set it in the first place, so all the logic there does nothing for h2 connections.

maskit: Ok, I think I was confused by this line. It seemed like to modify a client response based on a server response, but it doesn't.
https://github.com/apache/trafficserver/blob/cbe91698e2054ddb0ac33d34a92e270fa9194909/src/proxy/http/HttpTransact.cc#L6978

Http1Transaction and Http2Stream do the same thing in set_close_connection. It's not conditional at all. I assume Http3Transaction should do the same. Now I wonder why we don't modify the `heads` directly in HttpTransact. What's the reason for letting transactions to set the header? I think we should remove this confusing function.

duke8253: This was added in #8178 , maybe at the time we had some issues with our H2 streams and this header was causing trouble?

maskit: So we are going back to the original behavior (do the same thing for H1 and H2). Are you sure you don't have the issue that was resolved by #8178, with this change?

If we don't have to to differentiate the behavior, I prefer the original code that modifies `heads` directly in HttpTransact.

shinrich: The intent with #8178 was to not drain the session if the HttpSM logic sets the connection close due to detected transaction failures.  Shutting down the connection in HTTP/1 makes sense since the TCP connection may be polluted in the case of bad request bodies. However, this is not an issue with HTTP/2. An error with a request body should not require shutting down the session due to cross-contamination.

Explicitly setting the Connection: close header in the HTTP/2 processing logic when w want to close the session seems reasonable if it makes our client's lives easier. But blindly interpreting Connection: Close passed from the state machine processing will cause HTTP/2 sessions to shut down needlessly in some cases.

src/proxy/http2/Http2ConnectionState.cc
```diff
@@ -1530,17 +1555,7 @@ Http2ConnectionState::main_event_handler(int event, void *edata)
...
-    // ..., the server can send another GOAWAY frame with an updated last stream identifier
-    if (shutdown_reason == Http2ErrorCode::HTTP2_ERROR_MAX) {
-      shutdown_reason = Http2ErrorCode::HTTP2_ERROR_NO_ERROR;
-    }
-    send_goaway_frame(latest_streamid_in, shutdown_reason);
-    // Stop creating new streams
-    SCOPED_MUTEX_LOCK(lock, this->session->get_mutex(), this_ethread());
-    this->session->set_half_close_local_flag(true);
+    this->_shutdown_cont();
   } break;
```
maskit, line 1559: Looks like we no longer need a scope here. Let's remove the braces.

maskit:
@duke8253 What's the status of this PR? Is it ready for review?

JosiahWI:
The AuTest dns_host_down failed:
```
file /tmp/sandbox/dns_host_down/ts/log/error.log : DNS lookup should fail - Failed
        Reason: Contents of /tmp/sandbox/dns_host_down/ts/log/error.log did not contains expression: "DNS Error: no valid server http://resolve.this.com/"
        ```

**Comment by JosiahWI** [comment]
[approve ci rocky autest]

**Comment by duke8253** [comment]
@maskit It's ready for another review.

File: src/proxy/http2/Http2ConnectionState.cc (hunk 3 above)
Comment (maskit) [reply src/proxy/http2/Http2ConnectionState.cc:1523]: I still don't get why we need to call `finish_grace_shutdown()` here.
When we schdule `HTTP2_SESSION_EVENT_FINI`, we send `GOAWAY` with `latest_streamid_in`, right? Doesn't this send an extra `GOAWAY`?

- GOAWAY with the big number for graceful shutdown
- GOAWAY with latest_streamid_in at the place we schedule `HTTP2_SESSION_EVENT_FINI`
- GOAWAY with latest_streamid_in here in the `HTTP2_SESSION_EVENT_FINI` handler .

Comment (duke8253) [reply src/proxy/http2/Http2ConnectionState.cc:1523]: Because when we get a `HTTP2_SESSION_EVENT_FINI` we set the handler to `Http2ConnectionState::state_closed`, which just does `shutdown_cont_event = nullptr`, so the GOAWAY with `latest_streamid_in` never gets sent out.

Comment (maskit) [reply src/proxy/http2/Http2ConnectionState.cc:1523]: We only want to send two GOAWAY frames at most, A) one with the big number, and B) one with `latest_streamid_in`, right?

B is sent when we schedule `HTTP2_SESSION_EVENT_FINI` event. For example:
https://github.com/apache/trafficserver/blob/c49e60f769db8bdd2b662a2fa1fd2f90108a47dd/src/proxy/http2/Http2ConnectionState.cc#L1451-L1455

So when we reach this line, which is in `HTTP2_SESSION_EVENT_FINI` handler, B is already sent (or it's in the send buffer at least). Calling `finish_grace_shutdown` would send third GOAWAY.

Comment (duke8253) [reply src/proxy/http2/Http2ConnectionState.cc:1523]: yes, but that's just one case, but not for all of them. That's why I check for the `shutdown_state == HTTP2_SHUTDOWN_INITIATED` to make sure that we only send a second GOAWAY with `latest_streamid_in` when we initiated a grace shutdown.

Comment (maskit) [reply src/proxy/http2/Http2ConnectionState.cc:1523]: The only two places where we don't send GOAWAY prior to scheduling `HTTP2_SESSION_EVENT_FINI` are `Http2ClientSession::do_io_close` and `Http2ServerSession::do_io_close`. And those places seem to be too late to send frames anyway because `connection_state.is_state_closed()` is true if `HTTP2_SESSION_EVENT_FINI` is scheduled there.

Even if I miss a place that schedules `HTTP2_SESSION_EVENT_FINI` without sending GOAWAY, or even if we add new code like that, checking `shutdown_state == HTTP2_SHUTDOWN_INITIATED` doesn't make sense to me. How does it prevent sending a third `GOAWAY`? It's unrelated to whether a GOWAY with `latest_streamid_in` is sent. Are you saying the change ensures that there will be at least one GOAWAY frame with `latest_streamid_in` and there may be two of them in some cases, and that's fine?

I hate to add flags just to correct behavior forcibly, but it would be more straightforward solution if we add `bool is_the_last_goaway_sent` and set `true` in `Http2ConnectionState::send_goaway_frame` if the stream id is the big number. Then we can use it to ensure that ATS always send just one GOAWAY frame with `latest_streamid_in`.

Comment (duke8253) [reply src/proxy/http2/Http2ConnectionState.cc:1523]: The only way we get an event `HTTP2_SESSION_EVENT_SHUTDOWN_INIT` is when `shutdown_state == HTTP2_SHUTDOWN_NOT_INITIATED`, and that is the trigger for a grace shutdown since we only schedule that event when checking for the state like this. 
```
if (this->connection_state.get_shutdown_state() == HTTP2_SHUTDOWN_NOT_INITIATED) {
  send_connection_event(&this->connection_state, HTTP2_SESSION_EVENT_SHUTDOWN_INIT, this);
}
```

And from testing I've done so far, I was never able to get the second GOAWAY frame as described in the RFC when a grace shutdown is initiated, and that's why I created this PR. But I can move `this->_finish_grace_shutdown();` two lines above where it is at now, since `shutdown_cont_event` is only scheduled during a grace shutdown.

Comment (duke8253) [reply src/proxy/http2/Http2ConnectionState.cc:1523]: Maybe some other changes in the past few months have fixed this in a different way then, I just checked with Proxy Verifier and it is also getting the second GOAWAY frame now. So it's your call then, do you want to add a flag to make sure the second GOAWAY gets sent during a grace shutdown, or I can remove this part of the change and just add the test to check things are working as intended.

Comment (maskit) [reply src/proxy/http2/Http2ConnectionState.cc:1523]: Okay, nice. Then the current code works as I expect. I would not add the flag, because I think something is wrong if the flag has to work.

Let's just add the test. Thank you for writing the test.

Comment (duke8253) [reply src/proxy/http2/Http2ConnectionState.cc]: Sounds good.

File: tests/gold_tests/h2/http2_close_connection.test.py [hunk 6]
```diff
@@ -0,0 +1,71 @@
+'''
+Abort HTTP/2 connection using RST_STREAM frame.
```
maskit, line 2: Please update the comment and the summary (it looks truncated).

maskit (approved):

cmcfarlen:
Cherry-picked to v10.0.x to avoid conflicts with another PR

PR #11049, opened 2024-02-07
XPACK: Fix a ats_malloc/delete mismatch

bneradt:
An ASan run showed that in XpackDynamicTable we used ats_malloc for the table entries, but then delete it in the destructor. This fixes that mismatch.

### DIFF
--- a/src/proxy/hdrs/XPACK.cc
+++ b/src/proxy/hdrs/XPACK.cc
@@ -225,7 +225,7 @@ XpackDynamicTable::XpackDynamicTable(uint32_t size) : _maximum_size(size), _avai
 XpackDynamicTable::~XpackDynamicTable()
 {
   if (this->_entries) {
-    delete this->_entries;
+    ats_free(this->_entries);
     this->_entries = nullptr;
   }
 }


src/proxy/hdrs/XPACK.cc
```diff
@@ -225,7 +225,7 @@ XpackDynamicTable::XpackDynamicTable(uint32_t size) : _maximum_size(size), _avai
 XpackDynamicTable::~XpackDynamicTable()
 {
   if (this->_entries) {
-    delete this->_entries;
+    ats_free(this->_entries);
```
bneradt, line 228: The malloc, for reference, is a few lines above in the constructor here:
https://github.com/apache/trafficserver/pull/11049/files#diff-c9759c90bdcbd117f1d5b745a67412066eff8162ec490431323162d74dcc908cR220

```cpp
  this->_entries = static_cast<struct XpackDynamicTableEntry *>(ats_malloc(sizeof(struct XpackDynamicTableEntry) * size));
```

maskit (approved):

PR #11050, opened 2024-02-07
Add auto_option for ImageMagick and add support for specifying a COMP…

cmcfarlen:
…ONENT to auto_option


bryancall (dismissed):

cmake/AutoOptionHelpers.cmake
```diff
@@ -37,18 +37,26 @@ function(_REGISTER_AUTO_OPTION _NAME _FEATURE_VAR _DESCRIPTION _DEFAULT)
...
   else()
     set(STRICTNESS REQUIRED)
   endif()
 
   foreach(PACKAGE_NAME ${_PACKAGE_DEPENDS})
-    find_package(${PACKAGE_NAME} ${STRICTNESS})
-    if(NOT ${PACKAGE_NAME}_FOUND)
-      set(${_FEATURE_VAR} FALSE)
-      break()
+    if(NOT ${_COMPONENT} STREQUAL OFF)
```
JosiahWI, line 48: What about:
```suggestion
    if(${_COMPONENT})
```

`"OFF"`, `"0"`, `"FALSE"` and `""` should all have falsey values, I think.

cmake/AutoOptionHelpers.cmake
```diff
@@ -37,18 +37,26 @@ function(_REGISTER_AUTO_OPTION _NAME _FEATURE_VAR _DESCRIPTION _DEFAULT)
   )
 endfunction()
 
-macro(_CHECK_PACKAGE_DEPENDS _OPTION_VAR _PACKAGE_DEPENDS _FEATURE_VAR)
+macro(_CHECK_PACKAGE_DEPENDS _OPTION_VAR _PACKAGE_DEPENDS _FEATURE_VAR _COMPONENT)
```
JosiahWI, line 40: This would be more flexible as a list of components.

bryancall (dismissed):

cmake/AutoOptionHelpers.cmake
```diff
@@ -128,11 +140,17 @@ macro(auto_option _FEATURE_NAME)
     message(FATAL_ERROR "Invalid auto_option default ${ARG_DEFAULT}")
   endif()
 
+  if(ARG_COMPONENT)
+    set(COMPONENT ${ARG_COMPONENT})
+  else()
+    set(COMPONENT OFF)
```
JosiahWI, line 146: I'm not sure about setting it to `OFF` here; it doesn't represent an option.

JosiahWI (changes requested):

cmake/AutoOptionHelpers.cmake
cmcfarlen: I'm pretty sure I tried this and couldn't get it to work.

cmake/AutoOptionHelpers.cmake
cmcfarlen: agreed, but I couldn't get it to work providing two lists

cmake/AutoOptionHelpers.cmake
cmcfarlen: If I don't set it to off, the condition picking the COMPONENTS path would not work

cmcfarlen:
[approve ci]

bryancall (approved):

PR #11053, opened 2024-02-07
Fix SSLAPIHooks::g_ssl_hooks memory leak

bneradt:
It's a one-time global, but ASan reported this as a leak because we new'd it but didn't free it. Solving this by wrapping it in a unique_ptr.

### DIFF
--- a/include/iocore/net/SSLAPIHooks.h
+++ b/include/iocore/net/SSLAPIHooks.h
@@ -27,6 +27,8 @@
 
 #include "ts/apidefs.h"
 
+#include <memory>
+
 class TSSslHookInternalID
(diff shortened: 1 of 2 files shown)


bneradt:
The implementation `SSLApiHooks` is based upon `HttpAPIHooks` which also does a global new. The problem with my patch is that the init is called once, but the unique_ptr destructor is called from multiple contexts when plugins are involved. This results in multiple frees, which crashes. I spent about an hour looking into fixing this in various ways - I thought making a singleton class might help. However that gets freed multiple times too. I need to move onto other issues. For now I'm giving up.

PR #11055, opened 2024-02-07
clean up namespace of plugin API

dragon512:
This closes open namespaces we have in the plugin API.
This is the first part needed to clean up the code and make it easier to untie some circular dependences we have in the code

The changes are mainly adding a using namespace, or the namespace value to a given variable in the source

A little change was made to break up the API code to make it easier to read and build.

A define was added to make it easier to build plugins. ie the define is part of building the code ATS code, the plugins would build as is.

As a note it should be seen that the api plugin we have is actually needed to build ats, previous conversation we suggested that it was not a dependence for trafficserver itself. It should be much more clear now to see what the relationships are


ywkaras:
Looks fine.  How does it help with circular dependencies?

dragon512:
> Looks fine. How does it help with circular dependencies?

The goal is to tighten out the namespaces to make it easier to see what depends on what. This change is a big set of small changes, so I wanted to get this in first so hopefully we can untangle the dependency issues with a set of smaller PRs

dragon512:
[approve ci autest]

cmcfarlen (approved):

PR #11063, opened 2024-02-08
Remove header files generated by autotools automatically

maskit:
We removed this code on #10794 when we removed the support for autotools on master, but this is actually still helpful if people switch back and forth between 9.x and master.

The original code only removed `ink_config.h` and `ink_autoconf.h` but it should probably remove `apidefs.h` as well.


bneradt (approved):
Thank you very much. :)

PR #11064, opened 2024-02-09
Fix 2 HTTP/2 trailing header memory leaks

bneradt:
This addresses two leaks discovered while testing gRPC: one in which the pre-trailer tunnel has to be deallocated when the trailing tunnel is set up, the other in which the header_block memory for the initial headers has to be cleaned up when the new trailing headers are being processed.


shinrich (approved):
Looks like reasonable cleanups.

PR #11065, opened 2024-02-09
traffic_top: Allow traffic_top to be running while TS is down.

brbzull0:
We can now have `traffic_top` running even when TS is down,  there is now a status message to know that `traffic_top` is connecting to the host.

traffic_top2

Note the :arrow_up:  `connecting`  string at the bottom of the panel. Once connected, the hostname will be set in the bottom panel.


masaori335 (approved):
cool! I'm a big fan of this change.

PR #11066, opened 2024-02-09
remove tsapi::c namespace

dragon512:


include/ts/remap_version.h
```diff
@@ -43,5 +43,5 @@
   } while (0)
 
 #ifndef ATS_BUILD
```
cmcfarlen, line 45: Could remove the ATS_BUILD lines as well

cmcfarlen (changes requested):
ATS_BUILD is not needed.

include/ts/remap_version.h
dragon512: I was keeping it as a placeholder for when I add the namespaces back

dragon512: Still I will remove that...

cmcfarlen (approved):

Issue #9783, opened 2023-06-06
H3 crash HQSession::~HQSession


bryancall:
```
(gdb) bt full
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
        set = {__val = {0, 11014569033467807663, 3253139319627950568, 173988353035902449, 1, 13979173243358019584, 579005069656919567, 283686952306183, 9006311663777206111,
            4383123015962362557, 7951664624561075015, 985160869559618686, 0, 0, 0, 0}}
        pid = <optimized out>
        tid = <optimized out>
        ret = <optimized out>
#1  0x00007f8305540859 in __GI_abort () at abort.c:79
        save_stage = 1
        act = {__sigaction_handler = {sa_handler = 0x0, sa_sigaction = 0x0}, sa_mask = {__val = {0 <repeats 16 times>}}, sa_flags = 0, sa_restorer = 0x7f82f4120420}
        sigs = {__val = {32, 0 <repeats 15 times>}}
#2  0x00007f83062fdff1 in ink_abort (message_format=0x7f830636b040 "%s:%d: failed assertion `%s`") at ink_error.cc:99
        ap = {{gp_offset = 32, fp_offset = 48, overflow_arg_area = 0x7f83028a7d80, reg_save_area = 0x7f83028a7cc0}}
#3  0x00007f83062f8e9d in _ink_assert (expression=0x5581803b5240 "this->_transaction_list.head == nullptr", file=0x5581803b5230 "Http3Session.cc", line=42) at ink_assert.cc:37
No locals.
#4  0x000055818029739c in HQSession::~HQSession (this=0x7f82f40e93c0, __vtt_parm=0x5581805552e0 <VTT for Http3Session+8>, __in_chrg=<optimized out>) at Http3Session.cc:42
No locals.
#5  0x0000558180297d30 in Http3Session::~Http3Session (this=0x7f82f40e93c0, __in_chrg=<optimized out>, __vtt_parm=<optimized out>) at Http3Session.cc:179
No locals.
#6  0x0000558180297d7c in Http3Session::~Http3Session (this=0x7f82f40e93c0, __in_chrg=<optimized out>, __vtt_parm=<optimized out>) at Http3Session.cc:184
No locals.
#7  0x000055818029065c in Http3App::~Http3App (this=0x7f82f41249a0, __in_chrg=<optimized out>) at Http3App.cc:65
No locals.
[...]

Issue #9392, opened 2023-02-08
NetVC EOS/TIMEOUT is not propagated to H3 session


maskit:
EOS and TIMEOUT events (or events that notifie EOL of NetVC) are not propagated from QuicNetVC to H3Session. The session instance is deleted after all when QuicNetVC is freed, but that's not what happens on H2 session.

bryancall:
![image](https://github.com/apache/trafficserver/assets/216749/f6b6bf52-b0c3-4cd2-b704-569ba3cdc884)

PR #11071, opened 2024-02-13
http3: Propagate events from QUICNetVC

maskit:
This is for #9392 and #9783. The crash (assertion failure) happens because transactions are not deleted due to lack of event propagation. This PR add a mechanism to propagate events such as INACTIVE_TIMEOUT from QUICNetVC to Http3Transaction through Http3Session.

The event handlers may need to handle other events and may also need to do different things for each event, but this change should prevent the crash at least.

### DIFF
--- a/include/proxy/http3/Http3Session.h
+++ b/include/proxy/http3/Http3Session.h
@@ -37,8 +37,6 @@ public:
   virtual ~HQSession();
 
   // Implement VConnection interface
-  VIO *do_io_read(Continuation *c, int64_t nbytes = INT64_MAX, MIOBuffer *buf = nullptr) override;
-  VIO *do_io_write(Continuation *c = nullptr, int64_t nbytes = INT64_MAX, IOBufferReader *buf = 0, bool owner = false) override;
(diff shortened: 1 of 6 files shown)


brbzull0 (commented):
Even though this is defintelly needed, have you try to run the `via` test with quiche enabled?

I am running this patch and getting the same assert as in one of the [issues](https://github.com/apache/trafficserver/issues/9783#issuecomment-1841599316).

```
Fatal: ../src/proxy/http3/Http3Session.cc:43: failed assertion `this->_transaction_list.head == nullptr`
```

maskit:
I guess I thought it's done when I fixed one case, but it was obviously incomplete. The events were propagated to Http3Transaction but not HttpSM. The last commit added code for it.

Also, while testing the via test, I found another issue and fixed it on another PR (#11080). With the both changes, the via test passes on my box.

brbzull0 (approved):
Looks good!
Thanks!!

cmcfarlen:
Cherry-picked to v10.0.x

PR #11072, opened 2024-02-13
Handle EVENT_INTERVAL in HttpSM::state_raw_http_server_open and HttpSM::state_http_server_open

duke8253:
In the #9366 we stopped handling `EVENT_INTERVAL` in `HttpSM::state_raw_http_server_open` and `HttpSM::state_http_server_open`, but we've been seeing some unexpected aborts recently. This adds back the handling for that event.


shinrich (approved):
Seems reasonable to add back in this support.

cmcfarlen:
Cherry-picked to v10.0.x

Issue #10580, opened 2023-10-10
Invalid HTTP/1.1 chunk sizes are accepted and forwarded


kenballus:
ATS's chunked message body validation has problems dealing with malformed chunk sizes. When a chunk size matches the following regular expression: `\d+[^\d]+` (i.e. some digits followed by some non-digits) ATS ignores the non-digits and forwards them as-is instead of either normalizing them away or rejecting the request.

Here's an example request that demonstrates the bug:
```
POST / HTTP/1.1\r\n
Host: whatever\r\n
Transfer-Encoding: chunked\r\n
\r\n
1these-bytes-never-get-validated\r\n
Z\r\n
0\r\n
\r\n
```

Here's what ATS forwards to its backend:
```
POST / HTTP/1.1\r\n
Host: whatever\r\n
Client-ip: 172.18.0.1\r\n
X-Forwarded-For: 172.18.0.1\r\n
Via: http/1.1 traffic_server[94a0796c-32b5-413c-8bd6-b2ad4990c576] (ApacheTrafficServer/10.0.0)\r\n
Transfer-Encoding: chunked\r\n
\r\n
1these-bytes-never-get-validated\r\n
Z\r\n
0\r\n
\r\n
```

Note that the malformed chunk size made it through the transformation. Other reverse proxies (Apache, Caddy, Envoy, HAProxy, nghttpx, LiteSpeed, Pound, Squid, Traefik, Varnish, and more) reject the request. The specification defines that `chunk-size = 1*HEXDIG`, so this is a correct behavior. Another arguably correct behavior would be to delete the invalid characters upon re-emission (this is what H2O does).

kenballus:
This bug is exploitable for request smuggling to backend servers that interpret chunk sizes as their longest valid prefix. There are many such servers.

maskit:
Thank you for the report. I'm working on fixing it. If you have security concerns, please contact us by email.
https://apache.org/security/
security@trafficserver.apache.org

Issue #10477, opened 2023-09-20
ATS accepts and forwards carriage returns within chunk-ext BWS


kenballus:
RFCs 9110 and 9112 define the `chunk-ext` rule (and its dependencies) as follows:
```
chunk-ext = *( BWS ";" BWS chunk-ext-name [ BWS "=" BWS chunk-ext-val ] )
BWS = OWS
OWS = *( SP / HTAB )
```

Thus, a `chunk-ext` (which may follow a chunk size in a chunked HTTP message body) may begin with whitespace. The bug in ATS is that it accepts and forwards requests that contain carriage returns within that whitespace, even though only spaces and tabs should be permitted. To verify this for yourself, try sending the following request to an instance of ATS:
```
POST / HTTP/1.1\r\n
Transfer-Encoding: chunked\r\n
\r\n
1\r\r;chunk-ext-name\r\n
Z\r\n
0\r\n
\r\n
```
Note the invalid `\r\r` before the chunk extension name. ATS then forwards something this to the backend:
```
POST / HTTP/1.1\r\n
Client-ip: 172.18.0.1\r\n
X-Forwarded-For: 172.18.0.1\r\n
Via: http/1.1 traffic_server[90916a9b-0928-413b-8028-37facb5c2957] (ApacheTrafficServer/10.0.0)\r\n
Host: echo\r\n
Transfer-Encoding: chunked\r\n
\r\n
1\r\r;chunk-ext-name\r\n
Z\r\n
0\r\n
\r\n
```
Note that the invalid carriage returns made it through ATS.

This is of concern because there's a popular HTTP server that misinterprets carriage returns in this location.

PR #11073, opened 2024-02-13
Fix a bug in parsing chunked messages

maskit:
This is for #10477 and #10580.

### DIFF
--- a/src/proxy/http/HttpTunnel.cc
+++ b/src/proxy/http/HttpTunnel.cc
@@ -176,10 +176,16 @@ ChunkedHandler::read_size()
           break;
         }
       } else if (state == CHUNK_READ_SIZE_START) {
-        if (ParseRules::is_lf(*tmp)) {
+        if (ParseRules::is_cr(*tmp)) {
+          // Skip it
+        } else if (ParseRules::is_lf(*tmp) &&
+                   bytes_used <= 2) { // bytes_used should be 2 if it's CRLF, but permit a single LF as well
           running_sum = 0;
           num_digits  = 0;
           state       = CHUNK_READ_SIZE;
+        } else { // Unexpected character
+          state = CHUNK_READ_ERROR;
+          done  = true;
         }
       }
       tmp++;
--- a/src/proxy/http/HttpTunnel.cc
+++ b/src/proxy/http/HttpTunnel.cc
@@ -157,8 +157,8 @@ ChunkedHandler::read_size()
           }
         } else {
           // We are done parsing size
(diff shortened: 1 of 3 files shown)


maskit:
[approve ci autest]

src/proxy/http/HttpTunnel.cc
```diff
@@ -157,8 +157,9 @@ ChunkedHandler::read_size()
           }
         } else {
           // We are done parsing size
-          if (num_digits == 0 || running_sum < 0) {
-            // Bogus chunk size
+          if ((num_digits == 0 || running_sum < 0) ||                               /* Bogus chunk size */
+              (!ParseRules::is_cr(*tmp) && !ParseRules::is_ws(*tmp) && *tmp != ';') /* Unexpected character */
+          ) {
```
ywkaras, line 162: Taking a quick look at this, it seems like it would accept "100 junk  \n" as the chunk size of a 256 byte chunk.

maskit: Yeah, a malformed chunk extension would pass the check. I ignored almost everything that follows a chunk size, but it should not allow reading random data as a chunk size.

bneradt: Just making sure I understand: why is `;` an expected character? I don't ever recall seeing a `;` after the chunk header size.

bneradt: Oh, should we also not be surprised by a LF? We seem to be lenient on that elsewhere in this loop.

bneradt (changes requested):
Looks good in general. Just a couple questions to make sure I understand the patch.

src/proxy/http/HttpTunnel.cc
maskit: The semicolon is for `chunk-ext`.
```
  chunk          = chunk-size [ chunk-ext ] CRLF
                   chunk-data CRLF

  chunk-ext      = *( BWS ";" BWS chunk-ext-name
                      [ BWS "=" BWS chunk-ext-val ] )
```
https://www.rfc-editor.org/rfc/rfc9112#name-chunk-extensions

`BWS` is  Bad WhiteSpace, and it's basically optional whitespace.
https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.3-5

As for LF, I forgot to loosen the rule here. I pushed a commit to allow LF.

bneradt: Sounds good. Thank you for the explanation.

bneradt (approved):
lgtm

zwoop:
Cherry-picked to v9.2.x

PR #11076, opened 2024-02-15
ESI plugin:  make maximum document size configurable.

ywkaras:
Adds the --max-doc-size <number-of-bytes> command line option.


shukitchan (approved):
It looks fine. Thanks for the change.

randall (approved):

PR #11077, opened 2024-02-15
Add ENABLE_CLANG_TIDY option

masaori335:
CMake has a property called `<LANG>_CLANG_TIDY`.  This adds a function, `clang_tidy_check`, that set the property to the target if `ENABLE_CLANG_TIDY` is true. If it's enabled, `clang-tidy` check the rule in the `.clang-tidy` file when the target is compiled. 

https://cmake.org/cmake/help/v3.29/prop_tgt/LANG_CLANG_TIDY.html

The option is slightly different from what we used to have with auto tools, `make clang-tidy`, but this looks straight forward with cmake.

### DIFF
--- a/cmake/ClangTidy.cmake
+++ b/cmake/ClangTidy.cmake
@@ -28,14 +28,17 @@
 # }
 # ```
 
+if(ENABLE_CLANG_TIDY)
+  find_program(
+    CLANG_TIDY_EXE
+    NAMES "clang-tidy"
+    HINTS ${CLANG_TIDY_PATH}
+  )
+endif()
+
+
 function(clang_tidy_check target)
   if(ENABLE_CLANG_TIDY)
-    find_program(
-      CLANG_TIDY_EXE
-      NAMES "clang-tidy"
-      HINTS ${CLANG_TIDY_PATH}
-    )
-
     set_target_properties(${target} PROPERTIES CXX_CLANG_TIDY "${CLANG_TIDY_EXE};${CLANG_TIDY_OPTS};")
   endif()
 endfunction()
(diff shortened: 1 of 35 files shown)


masaori335:
This only has `tscore` and `tsutil` for the `clang_tidy_check` target. If this approach looks good, I'll add more targets.

cmake/ClangTidy.cmake
```diff
@@ -0,0 +1,41 @@
...
+# }
+# ```
+
+function(clang_tidy_check target)
+  if(ENABLE_CLANG_TIDY)
+    find_program(
+      CLANG_TIDY_EXE
+      NAMES "clang-tidy"
+      HINTS ${CLANG_TIDY_PATH}
+    )
```
cmcfarlen, line 37: You can move this find_program (also with a surrounding if) outside of the function and still use the CLANG_TIDY_EXE inside the function.

src/tscore/CMakeLists.txt
```diff
@@ -98,6 +98,9 @@ add_library(
 )
 add_library(ts::tscore ALIAS tscore)
 
+include(ClangTidy)
```
cmcfarlen, line 101: You can put this include in the top CMakeLists.txt

masaori335: that's nice!

cmake/ClangTidy.cmake
masaori335: It'll run when this file is included?

masaori335: Seems so. Done.

bryancall (approved):

PR #11085, opened 2024-02-20
doc: max_rst_stream_frames_per_minute defaults to 200

bneradt:
This tweaks our documentation to list the default value of proxy.config.http2.max_rst_stream_frames_per_minute to be the correct 200 default value rather than 14, which was probably copy and pasted from the max_settings_frames defaults.


cmcfarlen:
Cherry-picked to v10.0.x

zwoop:
Cherry-picked to v9.2.x

PR #11090, opened 2024-02-23
IPRange.h: resolve compiler error about dangling pointer to temporary

traeak:
This should resolve the compiler error in some of the branch builds and on my local machine with gcc 13.

### DIFF
--- a/lib/swoc/include/swoc/IPRange.h
+++ b/lib/swoc/include/swoc/IPRange.h
@@ -1985,7 +1985,13 @@ IPRangeView::clear() -> self_type & {
 
 inline bool
 IPRangeView::empty() const {
-  return AF_INET6 == _family ? _raw._6->empty() : AF_INET == _family ? _raw._4->empty() : true;
+  if (AF_INET6 == _family) {
+    return _raw._6->empty();
+  } else if (AF_INET == _family) {
+    return _raw._4->empty();
+  } else {
+    return true;
+  }
 }
 
 inline bool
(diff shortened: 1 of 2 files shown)


plugins/experimental/txn_box/plugin/src/ip_space.cc
```diff
@@ -907,10 +907,10 @@ Mod_ip_space::operator()(Context &ctx, feature_type_for<IP_ADDR> addr)
   }
   Feature value{FeatureView::Literal("")};
   if (active._space) {
-    auto &&[range, payload] = *active._space->space.find(addr);
-    active._row             = range.empty() ? nullptr : &payload;
-    active._addr            = addr;
-    active._drtv            = drtv;
+    auto [range, payload] = *active._space->space.find(addr);
```
bneradt, line 910: This will call the copy constructor for both the range and payload. If `&&` doesn't work, does `&` work? That will avoid a copy.

lib/swoc/include/swoc/IPRange.h
```diff
@@ -1985,7 +1985,13 @@ IPRangeView::clear() -> self_type & {
...
 inline bool
 IPRangeView::empty() const {
-  return AF_INET6 == _family ? _raw._6->empty() : AF_INET == _family ? _raw._4->empty() : true;
+  if (AF_INET6 == _family) {
+    return _raw._6->empty();
+  } else if (AF_INET == _family) {
+    return _raw._4->empty();
+  } else {
+    return true;
+  }
```
bneradt, line 1994: Nice.

bneradt (approved):
Thanks for trying to make compilers happy.

plugins/experimental/txn_box/plugin/src/ip_space.cc
traeak: let me try more variations on this.

traeak: the documentation of swoc/IPRange.h const_iterator with the auto && [r, p] seems to indicate that the compiler should take care of the danging references.  Apparently that's not always true.

PR #11094, opened 2024-02-23
NOT_WRITE_AVAIL: schedule event to resend data frames

bneradt:
When sending DATA frames, we check that we are not exceeding our write
buffer's high water mark. If so, we don't write the frame.
Unfortunately, before this patch, no event was scheduled to come back
and try to resend any unwritten frames. This led to dropped frames at
times. This patch adds a scheduled event to later re-try sending the
unsent frames.

This patch also updates the grpc.test.py to send 100 messages instead of
1 which, at least on my local system, reproduced this issue.

---

I'm running this on a box in Yahoo production and it is running fine.


src/proxy/http2/Http2ConnectionState.cc
```diff
@@ -2014,24 +2024,39 @@ Http2ConnectionState::update_initial_local_rwnd(Http2WindowSize new_size)
...
+{
+  Http2StreamDebug(session, stream->get_id(), "Scheduling sending data frames");
+
+  SCOPED_MUTEX_LOCK(lock, this->mutex, this_ethread());
+
+  if (!_data_scheduled) {
+    _data_scheduled = true;
+
+    SET_HANDLER(&Http2ConnectionState::main_event_handler);
+    this_ethread()->schedule_in((Continuation *)this, HRTIME_MSECOND, HTTP2_SESSION_EVENT_DATA);
```
bneradt, line 2056: I chose to schedule this for 1ms in the future. Does this sound reasonable to people? I figure time should be allotted for the buffer to drain. But maybe schedule_imm is better?

shinrich: I don't think it matter that much.  But your reasoning to leave a bit of time to drain seems sound.  Since we failed the first time, this isn't going to be the fastest path anyway.

bneradt: Good point - we're already delayed behind the buffer, so sleeping the ms is probably reasonable. Thanks @shinrich .

shinrich (approved):
Looks like splitting the scheduling tracking is a good approach.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11097, opened 2024-02-26
Python3.8: fix list type hints

bneradt:
The `list` type hint works for recent Python versions, but not for older ones. This uses `typing.List` instead as that is backwards compatible.


bneradt:
[approve ci autest]

cmcfarlen:
Cherry-picked to v10.0.x

PR #11104, opened 2024-02-27
fix hostdb.host_file.interval to be set at reconfig time

cmcfarlen:
This fixes an issue where shortening the check interval would require waiting for the current interval to expire before the value was updated.


bneradt (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11106, opened 2024-02-27
Add libswoc unit tests

bneradt:
Before this PR, we had pulled in libswoc's production source code into ATS. But we had not done the same with the libswoc unit tests. This ports libswoc unit_tests into ATS and ties it into our cmake system so that they are built and run as a part of ATS unit tests.


lib/swoc/include/swoc/IntrusiveHashMap.h
```diff
@@ -514,6 +514,15 @@ IntrusiveHashMap<H>::insert(value_type *v) {
...
     }
+    if (spot != limit) {
+      // If an equal key was found, walk past those to insert at the upper end of the range.
+      do {
+        spot = H::next_ptr(spot);
+      } while (spot != limit && H::equal(key, H::key_of(spot)));
+      if (spot != limit) { // something not equal past last equivalent, it's going to be mixed.
+        mixed_p = true;
+      }
+    }
```
bneradt, line 525: This puts back what we reverted because origin side session pooling was broken by it. See #10992 where we pulled this out.

To address the functional concerns, see my patch below in HttpSessionManager.cc

lib/swoc/include/swoc/TextView.h
```diff
@@ -1384,7 +1384,7 @@ TextView::suffix(int n) const noexcept {
 inline TextView
 TextView::suffix_at(char c) const {
   self_type zret;
-  if (auto n = this->rfind(c); n != npos) {
+  if (auto n = this->rfind(c); n != npos && n + 1 < this->size()) {
```
bneradt, line 1387: This addresses a compiler warning from Centos that zret.assign referenced the byte after data() if n pointed to the last byte in the string.

bneradt:
[approve ci autest]

traeak (approved):
These all look fine and the reverting of the IPRange and adjustments to HttpSessionManager look fine.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11108, opened 2024-02-28
Use EMERGENCY instead of FATAL for some certificate loading errors

lzx404243:
@ywkaras  suggested in #10958 that it'd be better to use `EMERGENCY` for these errors. 

This PR updates that for both the client and server cert load failures. The corresponding Autest is updated also.


maskit (dismissed):

lzx404243:
@maskit Do you mind approving this again? Missed an Autest update previously...

maskit (approved):

PR #11112, opened 2024-02-29
Coverity:  trying fixing some of the issues related to an unitialized variant member.

brbzull0:
Trying to fix some of the following CID issues:

1534738
1534732
1534699
1534727
1534717

not sure if this will work, but let's try.


ywkaras:
I will be mean to Damiam about this.

brbzull0:
> I will be mean to Damiam about this.

well, got used to already! this is sort of a shoot in the Dark! the issue seems to be inside the `std::variant` intialization. If this does not work, then I'll dig deeper.

ywkaras:
None of there Coverity issues look valid.  If this doesn't work, we could just use the `#ifdef __COVERITY__`.

cmcfarlen:
@brbzull0 Should we merge this?

brbzull0:
> @brbzull0 Should we merge this?

@cmcfarlen done! Thanks

cmcfarlen:
Cherry-picked to v10.0.x

PR #11118, opened 2024-03-01
autests change perl shasum to cksum

traeak:
The 'shasum' application used in h2/httpbin and h2/http2_priority tests come from the perl-Digest-SHA.
This replaces that with a more standard sha1sum, distributed as part of coreutils of most systems.
osx only provides the very basic cksum without installing coreutils, so change over to that.


maskit:
I wonder if there's another way that don't require commands, because this would require me to install coreutils to my Mac.

traeak:
> I wonder if there's another way that don't require commands, because this would require me to install coreutils to my Mac.

doc/ext/plantuml_fetch.sh
tests/prepare_proxy_verifier.sh
doc/ext/plantuml_fetch.sh

look for sha1sum then fallback to shasum (sometimes shasum -a 256).

maskit (approved):

PR #11120, opened 2024-03-01
cmake: Link pcre2 to src/api

maskit:
I get this error without the change.
```
[ 69%] Building CXX object src/api/CMakeFiles/tsapi.dir/InkAPI.cc.o
In file included from /Users/mkitajo/src/github.com/trafficserver/src/api/InkAPI.cc:32:
In file included from /Users/mkitajo/src/github.com/trafficserver/include/tscore/Encoding.h:23:
In file included from /Users/mkitajo/src/github.com/trafficserver/include/tscore/Diags.h:38:
In file included from /Users/mkitajo/src/github.com/trafficserver/include/tscore/DiagsTypes.h:45:
/Users/mkitajo/src/github.com/trafficserver/include/tsutil/Regex.h:32:10: fatal error: 'pcre2.h' file not found
#include <pcre2.h>
         ^~~~~~~~~
1 error generated.
```

### DIFF
--- a/src/api/CMakeLists.txt
+++ b/src/api/CMakeLists.txt
@@ -23,7 +23,7 @@ set(TSAPI_PUBLIC_HEADERS ${PROJECT_SOURCE_DIR}/include/ts/ts.h ${PROJECT_SOURCE_
                          ${PROJECT_SOURCE_DIR}/include/ts/TsException.h ${PROJECT_BINARY_DIR}/include/ts/apidefs.h
 )
 
-target_link_libraries(tsapi PRIVATE libswoc::libswoc yaml-cpp::yaml-cpp PCRE::PCRE OpenSSL::SSL)
+target_link_libraries(tsapi PRIVATE libswoc::libswoc yaml-cpp::yaml-cpp PCRE::PCRE PkgConfig::PCRE2 OpenSSL::SSL)
 set_target_properties(tsapi PROPERTIES PUBLIC_HEADER "${TSAPI_PUBLIC_HEADERS}")
 
 # Items common between api and other ts libraries


cmcfarlen:
Can we remove the PCRE::PCRE lib now?

maskit:
> Can we remove the PCRE::PCRE lib now?

I thought we can but unfortunately no. I get errors at different places if I remove it. It looks like we still have code that expects pcre1.

For example:
https://github.com/apache/trafficserver/blob/9e0b359c632190d01a948a9121d486fe15e8df61/include/proxy/ControlMatcher.h#L100-L106

bryancall:
Tes, we still have some places that use PCRE1 directly.  Those need to be converted to use the Regex class.

"I get this error without the change." - What change are you talking about?

bryancall:
src/api has a dependency on tsutil.  ts::tsutil should be added to target_link_libraries

maskit:
@cmcfarlen  Which one do we want here? `tscore`, `tsutil` or `PkgConfig::PCRE2`? Any of them works for me.

I wonder how this passed CI.

cmcfarlen:
Since tsutil/Regex.h includes the pcre2.h header file then there should be a PUBLIC target_link_libraries on the tsutil target.  This will transitively propagate to any target that depends on tsutil.

bryancall (approved):

maskit:
The reason I see the error may be this setting. Without the setting, OpenSSL which is installed by Homebrew is used despite having a path for BoringSSL in `OPENSSL_ROOT_DIR`.
```
“CMAKE_IGNORE_PATH”: “/opt/homebrew/include”
```

Issue #11114, opened 2024-02-29
Crash in Http3Transaction::state_stream_open: ink_assert(default_handler != (HttpSMHandler) nullptr);


bneradt:
Docs is running HTTP/3. We see this crash with some regularity on the box (maybe once an hour or so):

```gdb
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff66ba859 in __GI_abort () at abort.c:79
#2  0x000055555611e6e2 in ink_abort (message_format=0x555556b74840 "%s:%d: failed assertion `%s`") at /home/bneradt/src/trafficserver_10/src/tscore/ink_error.cc:99
#3  0x0000555556118556 in _ink_assert (expression=0x555556be3e60 "default_handler != (HttpSMHandler) nullptr", file=0x555556bdff80 "/home/bneradt/src/trafficserver_10/src/proxy/http/HttpSM.cc", line=2620) at /home/bneradt/src/trafficserver_10/src/tscore/ink_assert.cc:35
#4  0x0000555556208317 in HttpSM::main_handler (this=0x626000057100, event=105, data=0x0) at /home/bneradt/src/trafficserver_10/src/proxy/http/HttpSM.cc:2620
#5  0x00005555560b6417 in Continuation::handleEvent (this=0x626000057100, event=105, data=0x0) at /home/bneradt/src/trafficserver_10/include/iocore/eventsystem/Continuation.h:228
#6  0x0000555556aa2d35 in HQTransaction::_signal_event (this=0x62600004b100, event=105) at /home/bneradt/src/trafficserver_10/src/proxy/http3/Http3Transaction.cc:353
#7  0x0000555556aa5e16 in Http3Transaction::state_stream_open (this=0x62600004b100, event=105, edata=0x0) at /home/bneradt/src/trafficserver_10/src/proxy/http3/Http3Transaction.cc:506
#8  0x00005555560b6417 in Continuation::handleEvent (this=0x62600004b100, event=105, data=0x0) at /home/bneradt/src/trafficserver_10/include/iocore/eventsystem/Continuation.h:228
#9  0x0000555556a9cdc7 in HQSession::main_event_handler (this=0x619000143c80, event=105, edata=0x61b00002a2e0) at /home/bneradt/src/trafficserver_10/src/proxy/http3/Http3Session.cc:167
[...]

PR #11121, opened 2024-03-01
http3: Propagate edata as well

maskit:
I think this fixes https://github.com/apache/trafficserver/issues/11114.

It seems like `edata` should be a VIO when HttpSM receives INACTIVITY_TIMEOUT event (I confirmed this on H1 connection).


brbzull0 (approved):
Looks good.

PR #11122, opened 2024-03-04
rate_limit plugin: Decouple metrics and queue config

masaori335:
Prior to the change, rate_limit plugin metrics was created only if `queue` config is specified. However, it looks like `metrics` config doesn't require `queue` config.

- `selector.sni.queue`
https://docs.trafficserver.apache.org/en/latest/admin-guide/plugins/rate_limit.en.html#cmdoption-rate-limit-arg-queue

- `selector.sni.metrics`
https://docs.trafficserver.apache.org/en/latest/admin-guide/plugins/rate_limit.en.html#cmdoption-rate-limit-arg-metrics

### DIFF
--- a/plugins/experimental/rate_limit/limiter.h
+++ b/plugins/experimental/rate_limit/limiter.h
@@ -95,16 +95,16 @@ public:
       if (queue["max_age"]) {
         _max_age = std::chrono::seconds(queue["max_age"].as<uint32_t>());
       }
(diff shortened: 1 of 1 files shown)


traeak (dismissed):
Looks like a reasonable bug fix.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11124, opened 2024-03-04
test_ip: add iterator temporary for gcc release builds

traeak:
gcc release has issues with expanding iterators
```
auto &&[r, p] = *(cspace.find(a1));
```

introducing a temporary for the iterator gets around this.


bneradt (approved):

cmcfarlen:
Cherry-picked to v10.0.x

Issue #11123, opened 2024-03-04
traffic_ctl config match is confused with metrics


masaori335:
`proxy.process.http.*` are metrics, but it match with `traffic_ctl config match http` by somehow.

```
/opt/ats-asf-master-debug/bin/traffic_ctl config match http
proxy.process.http.background_fill_bytes_aborted: Unrecognized configuration value. Record is a configuration name/value but is not registered
proxy.process.http.background_fill_bytes_completed: Unrecognized configuration value. Record is a configuration name/value but is not registered
proxy.process.http.background_fill_current_count: Unrecognized configuration value. Record is a configuration name/value but is not registered
proxy.process.http.background_fill_total_count: Unrecognized configuration value. Record is a configuration name/value but is not registered
proxy.process.http.broken_server_connections: Unrecognized configuration value. Record is a configuration name/value but is not registered
proxy.process.http.cache_deletes: Unrecognized configuration value. Record is a configuration name/value but is not registered
proxy.process.http.cache_hit_fresh: Unrecognized configuration value. Record is a configuration name/value but is not registered
proxy.process.http.cache_hit_ims: Unrecognized configuration value. Record is a configuration name/value but is not registered
proxy.process.http.cache_hit_mem_fresh: Unrecognized configuration value. Record is a configuration name/value but is not registered
proxy.process.http.cache_hit_revalidated: Unrecognized configuration value. Record is a configuration name/value but is not registered
proxy.process.http.cache_hit_rww: Unrecognized configuration value. Record is a configuration name/value but is not registered
proxy.process.http.cache_hit_stale_served: Unrecognized configuration value. Record is a configuration name/value but is not registered
proxy.process.http.cache_lookups: Unrecognized configuration value. Record is a configuration name/value but is not registered
proxy.process.http.cache_miss_changed: Unrecognized configuration value. Record is a configuration name/value but is not registered
proxy.process.http.cache_miss_client_no_cache: Unrecognized configuration value. Record is a configuration name/value but is not registered
proxy.process.http.cache_miss_cold: Unrecognized configuration value. Record is a configuration name/value but is not registered
proxy.process.http.cache_miss_ims: Unrecognized configuration value. Record is a configuration name/value but is not registered
proxy.process.http.cache_miss_client_not_cacheable: Unrecognized configuration value. Record is a configuration name/value but is not registered
proxy.process.http.milestone.cache_open_read_begin: Unrecognized configuration value. Record is a configuration name/value but is not registered
[...]

PR #11127, opened 2024-03-05
Fix for issue 11123.  Do not return metrics when a config record is requested.

brbzull0:
fixes https://github.com/apache/trafficserver/issues/11123

### DIFF
--- a/src/records/RecCore.cc
+++ b/src/records/RecCore.cc
@@ -560,19 +560,23 @@ RecLookupMatchingRecords(unsigned rec_type, const char *match, void (*callback)(
     return REC_ERR_FAIL;
   }
 
-  // First find the new metrics, this is a bit of a hack, beacuse we still use the old
-  // librecords callback with a "pseudo" record.
-  RecRecord tmp;
+  if ((rec_type & (RECT_PROCESS | RECT_NODE | RECT_PLUGIN))) {
+    // First find the new metrics, this is a bit of a hack, because we still use the old
+    // librecords callback with a "pseudo" record.
+    RecRecord tmp;
 
-  tmp.rec_type  = RECT_ALL;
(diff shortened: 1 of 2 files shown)


src/records/RecCore.cc
```diff
@@ -560,18 +560,20 @@ RecLookupMatchingRecords(unsigned rec_type, const char *match, void (*callback)(
...
-
-  tmp.rec_type  = RECT_ALL;
-  tmp.data_type = RECD_INT;
-
-  for (auto &&[name, val] : ts::Metrics::instance()) {
-    if (regex.match(name.data()) >= 0) {
-      tmp.name         = name.data();
-      tmp.data.rec_int = val;
-      callback(&tmp, data);
+  if (REC_TYPE_IS_STAT(rec_type)) {
```
brbzull0, line 563: This part only cares about the stats.

src/records/RecCore.cc
```diff
@@ -560,18 +560,20 @@ RecLookupMatchingRecords(unsigned rec_type, const char *match, void (*callback)(
...
+    tmp.data_type = RECD_INT;
+
+    for (auto &&[name, val] : ts::Metrics::instance()) {
+      if (regex.match(name.data()) >= 0) {
+        tmp.name         = name.data();
+        tmp.data.rec_int = val;
+        callback(&tmp, data);
+      }
     }
   }
```
brbzull0, line 578: all metrics will be handled here, so maybe we can just return from here and not lookup on the `g_records` which seems to hold only configs.

masaori335: Sounds reasonable for metrics lookup.

zwoop (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11129, opened 2024-03-05
cleanup cript example. add missing function

cmcfarlen:
This should fix the build issues with macOS and also allow the example cript plugin to pass verify.


zwoop (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11142, opened 2024-03-06
Fixed asan issue with global SSLAPIHooks not being deleted on shutdown

bryancall:
I didn't see where `g_ssl_hooks` was being deleted on shutdown and it was reporting this as a leak in asan.

I also saw in the unit tests `init_global_ssl_hooks()` was being called multiple times.  I added a check to make sure `g_ssl_hooks` was not being overridden.

### DIFF
--- a/src/iocore/net/SSLAPIHooks.cc
+++ b/src/iocore/net/SSLAPIHooks.cc
@@ -25,15 +25,6 @@
 
 SSLAPIHooks *g_ssl_hooks = nullptr;
 
-namespace
-{
-// structure to delete the global hooks
-struct GlobalHooksDeleter {
-  ~GlobalHooksDeleter() { delete g_ssl_hooks; }
-};
-} // namespace
-GlobalHooksDeleter g_hooks_deleter;
-
 void
 init_global_ssl_hooks()
 {
--- a/src/iocore/net/SSLAPIHooks.cc
+++ b/src/iocore/net/SSLAPIHooks.cc
@@ -25,8 +25,19 @@
 
 SSLAPIHooks *g_ssl_hooks = nullptr;
 
+namespace
+{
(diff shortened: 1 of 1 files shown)


bryancall:
[approve ci autest]

brbzull0:
[approve ci autest]

JosiahWI (dismissed):
More information in the comment on why alternatives did not work might be even better, but this looks good to me (if it works).

bryancall:
[approve ci autest]

JosiahWI:
Was the root problem the multiple allocations of `g_ssl_hooks` overwriting the old ones and leaking them?

JosiahWI:
[approve ci autest]

bneradt:
@JosiahWI and I have talked about this leaked resource in the past. Note that this comment, and another related global hook, has had this comment forever in the corresponding header file:

https://github.com/apache/trafficserver/blob/0fef5a3450a4aaf21a465300d0e15c2d4d43f5fa/include/iocore/net/SSLAPIHooks.h#L57

bryancall:
> Was the root problem the multiple allocations of `g_ssl_hooks` overwriting the old ones and leaking them?

The problem was with deleting `g_ssl_hooks` on shutdown.

I saw that we were calling init_global_ssl_hooks() multiple time in autest, so I put the nullptr check there.  I also assumed we also needed to delete the global, but it looks like something might already be cleaning it up?  According to @bneradt comment it should be leaking, but I am not seeing that myself.

JosiahWI:
If `init_global_ssl_hooks()` was called multiple times without the null check, that would probably leak memory.

JosiahWI (approved):
Approving, but a non-Yahoo person will need to approve as well.

bryancall:
Closing - better fix is in #11224

PR #11155, opened 2024-03-11
Remove support for session id based session resumption

duke8253:
As discussed in the email, since session id bases session resumption is barely used anymore (0.02% of the traffic on our production boxes), and we removed the session share plugin that shares session ids across boxes in a colo, I'm removing the support for session id based session resumption. 

Also added some new autests, and fixed up the documentation.


duke8253:
Closing this as discussed, new PR is #11157

PR #11156, opened 2024-03-12
txn_box - Fix ASAN issue on txn_box string trie.

brbzull0:
Use `swoc::MemArena` instead of manual allocation.
Beside the benefits of using the arena this change helps to mitigate a leak that was found while building with ASAN.
The issue was on desruction of the trie which should only happen when the application stops.


brbzull0:
@moonchen ping!  :rescue_worker_helmet:

moonchen (approved):

brbzull0:
> @moonchen ping! ⛑️

Thanks Mo! :bow:

cmcfarlen:
Cherry-picked to v10.0.x

PR #11160, opened 2024-03-13
Test failure checks for Stripe::add_writer

JosiahWI:
This adds a new data-driven unit test to check all 32 conditions for the behavior of `Stripe::add_writer` and fixes an error in the docstring for that method.

There is no test to confirm whether or not the connection was added to the queue because that cannot be determined easily. I think that should be tested at the point where the write to the aggregation buffer occurs.

### DIFF
--- a/src/iocore/cache/unit_tests/test_Stripe.cc
+++ b/src/iocore/cache/unit_tests/test_Stripe.cc
@@ -1,0 +1,159 @@
+/** @file
+
+  A brief file description
+
+  @section license License
+
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+
+#include "main.h"
+
+#include <array>
+#include <ostream>
+
+// Required by main.h
+int cache_vols            = 1;
+bool reuse_existing_cache = false;
+
+struct AddWriterBranchTest {
+  int initial_buffer_size{};
+  int agg_len{};
+  int header_len{};
+  int write_len{};
+  int readers{};
+  bool result{};
+};
+
+std::array<AddWriterBranchTest, 32> add_writer_branch_test_cases = {
+  {
+   {0, 0, 0, 0, 0, true},
+   {0, 0, 0, 0, 1, true},
+   {0, 0, 0, 1, 0, true},
+   {0, 0, 0, 1, 1, true},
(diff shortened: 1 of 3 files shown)


bneradt:
Marking this comment and test only change for backport to 10.0.x since it provides test benefit without a risk to production.

src/iocore/cache/unit_tests/test_Stripe.cc
```diff
@@ -0,0 +1,159 @@
...
+   {AGG_SIZE + cache_config_agg_write_backlog, 0, MAX_FRAG_SIZE + 1 - sizeof(Doc), 1, 1, false},
+   {AGG_SIZE + cache_config_agg_write_backlog, AGG_SIZE + 1, 0, 0, 0, false},
+   {AGG_SIZE + cache_config_agg_write_backlog, AGG_SIZE + 1, 0, 0, 1, false},
+   {AGG_SIZE + cache_config_agg_write_backlog, AGG_SIZE + 1, 0, 1, 0, false},
+   {AGG_SIZE + cache_config_agg_write_backlog, AGG_SIZE + 1, 0, 1, 1, false},
+   {AGG_SIZE + cache_config_agg_write_backlog, AGG_SIZE + 1, MAX_FRAG_SIZE + 1 - sizeof(Doc), 0, 0, false},
+   {AGG_SIZE + cache_config_agg_write_backlog, AGG_SIZE + 1, MAX_FRAG_SIZE + 1 - sizeof(Doc), 0, 1, false},
+   {AGG_SIZE + cache_config_agg_write_backlog, AGG_SIZE + 1, MAX_FRAG_SIZE + 1 - sizeof(Doc), 1, 0, false},
+   {AGG_SIZE + cache_config_agg_write_backlog, AGG_SIZE + 1, MAX_FRAG_SIZE + 1 - sizeof(Doc), 1, 1, false},
+   }
```
masaori335, line 76: I'm wondering why this array is not formatted by clang-format 🤔

src/iocore/cache/unit_tests/test_Stripe.cc
```diff
@@ -0,0 +1,159 @@
...
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+
+#include "main.h"
+
+#include <array>
+#include <ostream>
+
+// Required by main.h
```
masaori335, line 29: what is required by math.h?

masaori335 (approved):
This covers all cases of the if statement in the `Stripe::add_writer(CacheVC *vc)`
https://github.com/apache/trafficserver/blob/e1e8e434d3b3b27561a14bc09fec3d37c4ec0dc2/src/iocore/cache/Stripe.cc#L937-L940

src/iocore/cache/unit_tests/test_Stripe.cc
JosiahWI: `main.h`, not `math.h`. I copied that comment and the two LoC below it verbatim from another test; I think those symbols are declared in main.h as globals and the test has to define them.

src/iocore/cache/unit_tests/test_Stripe.cc
JosiahWI: How can you tell that it isn't? Perhaps the file is accidentally whitelisted somehow? I will look into this before merging.

JosiahWI: I ran the clang-format locally with the `--target clang-format` option to `cmake`, and I confirmed that local changes to `test_Stripe.cc` do get formatted correctly. I think this is enough evidence that the formatter is running on the file; but maybe it is ignoring that array for some reason. I will merge for now, but I will make a note of this.

cmcfarlen:
Cherry-picked to v10.0.x

Issue #11153, opened 2024-03-11
records.yaml validation on loading


masaori335:
`RecordElement` has a member for validation of value called `regex`, it's used when a config is given through `traffic_ctl config set` command. However, it looks like it's not used when the records.yaml (records.config for 9.2.x) is loaded.

https://github.com/apache/trafficserver/blob/master/src/records/RecordsConfig.cc#L39C14-L39C27

Link: #11110

brbzull0:
@masaori335  I can look at this is none else is already looking at it.

PR #11161, opened 2024-03-14
records.yaml - Make sure we validate the input value with the configured check expr.

brbzull0:
This was missing from the validation.
If the validation fails it will use the default value and print out a warning message. Stopping ATS from start because of this could be quite hard, but we can discuss it.

```
[Mar 14 13:15:36.962] traffic_server NOTE: records parsing completed.
[Mar 14 13:15:36.963] traffic_server NOTE:  4: proxy.config.http.connect_ports - Validity Check failed. '^(\*|[[:digit:][:space:]-]+)$' againts 'xyz'.  Default value will be used.
```

I had to refactor a bit some code to make it available.

This as is wont work on 9.2 as the code is different.

fixes https://github.com/apache/trafficserver/issues/11153


src/records/RecordsConfig.cc
```diff
@@ -805,7 +805,7 @@ static const RecordElement RecordsConfig[] =
   ,
   //  # default the ram cache size to AUTO_SIZE (-1)
   //  # alternatively: 20971520 (20MB)
-  {RECT_CONFIG, "proxy.config.cache.ram_cache.size", RECD_INT, "-1", RECU_RESTART_TS, RR_NULL, RECC_STR, "^-?[0-9]+$", RECA_NULL}
+  {RECT_CONFIG, "proxy.config.cache.ram_cache.size", RECD_INT, "-1", RECU_RESTART_TS, RR_NULL, RECC_STR, "^-?[0-9]+[A-Za-z]{0,}$", RECA_NULL}
```
serrislew, line 808: where does it evaluate the unit suffix to convert it into the appropriate int? does it check for invalid units? Seems like in the docs `proxy.config.cache.ram_cache_cutoff` also supports unit suffix. Should that regex be included as well?

brbzull0: Thanks for having a look at this!

>where does it evaluate the unit suffix to convert it into the appropriate int? does it check for invalid units?

It should be handled [here](https://github.com/apache/trafficserver/blob/4681638c043a23834cec85a2d55a0cddd071881f/src/tscore/ParseRules.cc#L51). 
> Seems like in the docs proxy.config.cache.ram_cache_cutoff also supports unit suffix. Should that regex be included as well?

yes, and probably there are others as well. This particular line was changed because it was detected at `records.yaml` loading time.

Worse case scenario, it will be detected in the `ink_atoi`.  I can probably put another PR and make sure the regex are in place for records like this.

serrislew (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11163, opened 2024-03-14
Adds TSHttpTxnErrorBodyGet() corresponding to the Set

zwoop:
One use case here (a separate PR) is that a plugin like header_rewrite  (or Cripts / TxnBox) can now check if the error body has been set in a plugin agnostic way (so can set it it one plugin, check in another). For header_rewrite, this actually would allow the set-body operator to work even when the request is read from cache or going to origin (thanks Jasmine).


bneradt (approved):
lgtm

cmcfarlen:
Cherry-picked to v10.0.x

PR #11167, opened 2024-03-18
Restore alignment for structs stored in cache

cmcfarlen:
PR #9481 changed the alignment on structs that are stored in cache.  This breaks the cache.

### DIFF
--- a/include/proxy/hdrs/HTTP.h
+++ b/include/proxy/hdrs/HTTP.h
@@ -480,19 +480,19 @@ class HTTPHdr : public MIMEHdr
 public:
   HTTPHdrImpl *m_http = nullptr;
   mutable URL m_url_cached;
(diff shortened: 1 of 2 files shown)


bneradt (approved):
Looks good.

Out of curiosity, were there warnings or the like due to the misalignment?

cmcfarlen:
> Looks good.
> 
> Out of curiosity, were there warnings or the like due to the misalignment?

Release assertions running 10.0.0 with caches created with 9.2
```
Fatal: ...src/proxy/hdrs/HdrHeap.cc:921: failed assertion `m_writeable == false`
```

cmcfarlen:
I've tested this PR in production and it no longer hits the asserts.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11169, opened 2024-03-18
set compile definition to turn on maxminddb

cmcfarlen:
This was causing errors to show up in the logs since no geo service was available.


JosiahWI:
I see that there's also a `TS_USE_HRW_GEOIP`. Is that something unrelated?

bryancall (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11172, opened 2024-03-19
Support rate limit on active connections instead of txn

zwoop:


ywkaras (approved):

shukitchan (approved):
Took me a while to trace through the loop. But it actually check out nicely. 
Good one.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11177, opened 2024-03-21
Version update: libswoc 1.5.11

bneradt:
I pulled in our ATS libswoc patches into apache/trafficserver-libswoc and cut a 1.5.11 release with that. This updates just the version we have internally to match that release. The code between both ATS libswoc and trafficserver-libswoc are now identical with
1.5.11


ezelkow1 (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11178, opened 2024-03-21
Get Stripe::aggWrite into a test harness

JosiahWI:
This demonstrates how to set up and perform a working write to disk on a Stripe. The data is written to a tmpfile that the test holds the file stream pointer to, so it should be possible to inspect and verify the written data. The test does not do this, but having this example may be useful. This commit also adds a comment to describe the purpose of some macros in CacheWrite.cc.

### DIFF
--- a/src/iocore/cache/CacheWrite.cc
+++ b/src/iocore/cache/CacheWrite.cc
@@ -26,6 +26,8 @@
 #include "iocore/cache/AggregateWriteBuffer.h"
 #include "iocore/cache/CacheEvacuateDocVC.h"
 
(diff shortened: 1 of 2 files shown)


JosiahWI:
There's also still a case (good thing the OSX CI caught it!) where there is an out-of-bounds memory write.

bneradt:
[approve ci rocky]

bneradt:
[approve ci rocky]

masaori335 (dismissed):
Pretty interesting approach. Thanks for adding more tests.

bneradt:
[approve ci]

masaori335 (approved):

PR #11180, opened 2024-03-21
Fixes for SRV records handling in hostdb rewrite

cmcfarlen:
Production testing uncovered bugs in SRV record handling.

### DIFF
--- a/src/iocore/hostdb/HostDBInfo.cc
+++ b/src/iocore/hostdb/HostDBInfo.cc
@@ -65,6 +65,17 @@ HostDBInfo::assign(IpAddr const &addr) -> self_type &
   return *this;
 }
 
+/** Assign SRV record and name
+ *
+ * @param srv Pointer to SRV to assign.
+ * @param name Pointer to hostname for the record
+ *
+ * @note This function assumes that name is stored near to the this pointer of
+ * the assigned instance within a uint16_t worth of bytes.  This invariant must
+ * be adhered to by the caller.
+ *
+ * @todo Refactor handling of SRV records
+ */
 auto
 HostDBInfo::assign(SRV const *srv, char const *name) -> self_type &
 {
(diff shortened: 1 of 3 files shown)


src/iocore/hostdb/HostDBInfo.cc
```diff
@@ -73,7 +73,7 @@ HostDBInfo::assign(SRV const *srv, char const *name) -> self_type &
   data.srv.srv_priority = srv->priority;
   data.srv.srv_port     = srv->port;
   data.srv.key          = srv->key;
-  data.srv.srv_offset   = reinterpret_cast<char const *>(this) - name;
+  data.srv.srv_offset   = name - reinterpret_cast<char const *>(this);
```
bneradt, line 76: Now that you figured this out, can you please add a comment explaining this pointer arithmetic? Or is there a comment elsewhere? The calling logic which loops over `rr_info` is not trivial for me to think through.

bneradt (changes requested):

src/iocore/hostdb/HostDBInfo.cc
cmcfarlen: I'll add a comment.  The bigger issue here is that the code assumes that the pointer parameter passed in is one that can be recreated from a 16-bit offset value.  You couldn't, for example, pass in a pointer into the stack.  This is a huge design flaw.  While this change corrects the math, it doesn't address the underlying issue.

src/iocore/hostdb/HostDBInfo.cc
```diff
@@ -65,6 +65,17 @@ HostDBInfo::assign(IpAddr const &addr) -> self_type &
   return *this;
 }
 
+/** Assign SRV record and name
+ *
+ * @param srv Pointer to SRV to assign.
+ * @param name Pointer to hostname for the record
```
bneradt, line 71: nit: Can you please make these both `@param[in]` to show they are input variables. Along these lines, I suggest: `Pointer to the SRV from which to assign.`

bneradt (changes requested):
Looks great. I just have a small request for a tweak in the doxygen comment.

bneradt (approved):

src/iocore/hostdb/HostDBInfo.cc
bneradt: Thank you!

cmcfarlen:
Cherry-picked to v10.0.x

PR #11164, opened 2024-03-15
RegexContext instance created in the heap for each thread is leaked.

ywkaras:
Damien found by running an ASAN build that the RegexContext instance created in the heap for each thread is leaked. The thread_local RegexContextCleanup variables are never constructed becasue they are never referenced in any thread.

The solution is a different approach to avoiding deadlock when loading plugins with statically allocated DbgCtl instances.  Construction of thread_locals is avoided while the DbgCtl registry mutex is locked.

PR #11184, opened 2024-03-25
Fixed asan leak issues with RegexContext

bryancall:


src/tsutil/Regex.cc
```diff
@@ -105,12 +120,20 @@ class RegexContext
 thread_local RegexContext *RegexContext::_regex_context = nullptr;
 
 //----------------------------------------------------------------------------
-namespace
+
+RegexContextCleanup::~RegexContextCleanup()
 {
```
moonchen, line 125: I think you need to acquire `_mutex` here, since it's modifying `_contexts`, right?

ywkaras:
If you don't copy in these comments, https://github.com/apache/trafficserver/pull/11164/files#diff-8b89b132c9c9eb8fa6f58b050dd7bd3c47d55fd217d00239f4e13e21b270d679R161 , somebody may say "oh, I can just make these thread_local!" six months from now.

src/tsutil/Regex.cc
bryancall: The mutex is not really needed, since it is called atexit, but I added it anyways.

ywkaras:
Unfortunately, there is a race condition, if an object with static storage uses a DbgCtl object for the first time in its destructor.  (That requires checking if its tag matches the enabled tag regex.)  I think @cmcfarlen (persumably working in league with Satan) was able to generate a scenario like that.

ywkaras:
Brian, were you going to change this so that all match requests will return false after the cleanup object is destroyed?

ywkaras (changes requested):
Brian, were you going to change this so that all match requests will return false after the cleanup object is destroyed?

cmcfarlen (dismissed):

src/tsutil/Regex.cc
```diff
@@ -265,11 +305,19 @@ Regex::exec(std::string_view subject) const
...
+  // check if there is a compiled regex
   if (_code == nullptr) {
     return 0;
   }
+
+  // get the RegexContext instance - should only be null when shutting down
+  RegexContext *regex_context = RegexContext::get_instance();
+  if (regex_context == nullptr) {
+    return false;
```
ywkaras, line 316: Seems like this should return 0.

bryancall: Fixed, copy-and-paste issue.

cmcfarlen (approved):

ywkaras (approved):

src/tsutil/DbgCtl.cc
```diff
@@ -150,14 +150,28 @@ DbgCtl::_new_reference(char const *tag)
...
+  // 1.  Thread 1 calls a DbgCtl constructor, which locks the registry mutex, but then is suspended.
+  // 2.  Thread 2 calls dlopen() for a plugin, locking the runtime mutex.  It then executes the constructor for a
+  //     statically allocated DbgCtl object, which blocks on locking the registry mutex.
+  // 3.  Thread 1 resumes, and calls member functions of the derived class of DebugInterface.  If this causes the
+  //     the construction of a thread_local variable with a non-trivial destructor, _cxa_thread_atexit() will be called
+  //     to set up a call of the variable's destructor at thread exit.  The call to _cxa_thread_atexit() will block on
+  //     the runtime mutex (held by Thread 2).  So Thread 1 holds the registry mutex and is blocked waiting for the
+  //     runtime mutex.  And Thread 2 holds the runtime mutex and is blocked waiting for the registry mutex.  Deadlock.
+  //
+  // This deadlock is avoided by having the thread_local variable register its destruction in a non-thread_local class.
```
ywkaras, line 174: ??? I thought this was already present.

bryancall:
[approve ci autest]

PR #11188, opened 2024-03-27
http2: change stream state to closed from open if END_STREAM is sent and received

masaori335:
This corrects the HTTP/2 Stream State transition of origin server side HTTP/2 connection. I'm not sure this makes any impact or not, but this is consistent with client side HTTP/2 connection.

https://github.com/apache/trafficserver/blob/29872c258edd21fb05bec51e5667df2deae2e969/src/proxy/http2/Http2Stream.cc#L402-L408

https://datatracker.ietf.org/doc/html/rfc9113#name-stream-states


masaori335:
[approve ci autest]
[approve ci fedora]

masaori335:
[approve ci fedora]

masaori335:
[approve ci autest]

bneradt (approved):

PR #11192, opened 2024-03-27
Allow Cripts to be used directly as @plugin

zwoop:
An example compiler script would be (for macOS):

```
#!/bin/sh

SOURCE=$1
DEST=$2
CXXFLAGS="-std=c++20 -I/opt/homebrew/include -I/opt/ats/include"

exec clang++ -undefined dynamic_lookup -shared -fPIC ${CXXFLAGS} -L/opt/ats/lib -lcripts -o $DEST $SOURCE
```

### DIFF
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -4901,6 +4901,20 @@ Plug-in Configuration
    Enables (``1``) or disables (``0``) the dynamic reload feature for remap
    plugins (`remap.config`). Global plugins (`plugin.config`) do not have dynamic reload feature yet.
 
+.. ts:cv:: CONFIG proxy.config.plugin.compiler_path STRING ""
+
+   Specifies an option compiler tool path for compiling plugins. This tool should
+   be an executable, which takes two arguments:
+
+   === ======================================================================
+   Arg Description
+   === ======================================================================
+   1   This is the path to the source file, which should be compiled
+   2   This is the path to the DSO file, which will be created and loaded
+   === ======================================================================
+
+   The script should exit with a status code of ``0`` if the compilation was successful.
+
 .. ts:cv:: CONFIG proxy.config.plugin.vc.default_buffer_index INT 8
    :reloadable:
    :overridable:
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -4903,7 +4903,7 @@ Plug-in Configuration
 
 .. ts:cv:: CONFIG proxy.config.plugin.compiler_path STRING ""
 
-   Specifies an option compiler tool path for compiling plugins. This tool should
+   Specifies an optional compiler tool path for compiling plugins. This tool should
    be an executable, which takes two arguments:
 
    === ======================================================================
--- a/src/proxy/http/remap/PluginFactory.cc
+++ b/src/proxy/http/remap/PluginFactory.cc
@@ -124,6 +124,14 @@ PluginFactory::setRuntimeDir(const fs::path &runtimeDir)
   return *this;
 }
 
+PluginFactory &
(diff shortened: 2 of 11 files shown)


zwoop:
[approve ci autest]

doc/admin-guide/files/records.yaml.en.rst
```diff
@@ -4901,6 +4901,20 @@ Plug-in Configuration
    Enables (``1``) or disables (``0``) the dynamic reload feature for remap
    plugins (`remap.config`). Global plugins (`plugin.config`) do not have dynamic reload feature yet.
 
+.. ts:cv:: CONFIG proxy.config.plugin.compiler_path STRING ""
+
+   Specifies an option compiler tool path for compiling plugins. This tool should
```
bneradt, line 4906: `option` -> `optional`

tools/cripts/compiler.sh
```diff
@@ -0,0 +1,97 @@
...
+
+if [ ! -f "$SOURCE" ]; then
+    echo "Source file $SOURCE does not exist"
+    exit 1
+fi
+
+if [ "${DEST_FILE}" != "${SOURCE_FILE}.so" ]; then
+    echo "Destination file name must match source file name with .so extension"
+    exit 1
+fi
```
bneradt, line 66: So if the user passes in `cripts.cc`, `DEST_FILE` should be `cripts.cc.so`? Maybe we should remove the extension from `$SOURCE_FILE` via `${SOURCE_FILE%.*}`.

tools/cripts/compiler.sh
```diff
@@ -0,0 +1,97 @@
...
+
+cache_file=""
+if [ -d "$CACHE_DIR" ]; then
+    cache_tree="${CACHE_DIR}${SOURCE_DIR}"
+    cache_file="${cache_tree}/${DEST_FILE}"
+
+    [ -d "$cache_tree" ] || mkdir -p "$cache_tree"
+
+    if [ ! "$SOURCE" -nt "$cache_file" ]; then
```
bneradt, line 75: Perhaps: `[ "$SOURCE" -ot "$cache_file" ]` (i.e., use `-ot`).

src/proxy/http/remap/PluginFactory.cc
```diff
@@ -175,6 +183,20 @@ PluginFactory::getRemapPlugin(const fs::path &configPath, int argc, char **argv,
...
 
+      // Special case for Cripts
+      if (!runtimePath.string().ends_with(".so")) {
+        if (_compilerPath.empty()) {
+          error.assign("compiler path not set for compiling plugins");
+          return nullptr;
+        }
+
+        // Add .so to the source file, so e.g. example.cc.so. ToDo: libswoc doesn't allow appending to the extension.
+        std::string newPath = runtimePath.string();
```
bneradt, line 194: Oh, I suppose we will compile it to `example.cc.so`. It shouldn't be too hard to remove the extension after the last `.` right?

bneradt (changes requested):
Looks great. I just have a few suggestions to tweak things.

tools/cripts/compiler.sh
zwoop: I did that intentional for a few reasons, but primarily to not make assumptions what extension (if any) people will use, and also assure consistent naming. You can now see exactly what file was compiled. 

This is sort of inline with how cmake does it as well :)

bneradt: Sounds reasonable.

doc/admin-guide/files/records.yaml.en.rst
zwoop: Fixed

tools/cripts/compiler.sh
zwoop: Done

bneradt (approved):

bneradt (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11193, opened 2024-03-28
Fix logic around srv records at other code path

cmcfarlen:
This fixes the code path for when SRV result is immediate


bneradt (approved):
Makes sense.

cmcfarlen:
I did verify this fix in production.  Thanks for the review!

cmcfarlen:
Cherry-picked to v10.0.x

PR #11195, opened 2024-03-28
CMakePresets: branch-quiche: use boringssl nuraft

bneradt:
The standard quiche build uses boringssl, so the NuRaft for the STEK plugin must be the NuRaft built against boringssl.


cmcfarlen (approved):

cmcfarlen:
Cherry-picked to v10.0.x

Issue #11170, opened 2024-03-19
heap use after free in QUICStreamAdapter::stream via HQTransaction::get_transaction_id


bneradt:
With the latest 10.0.x (https://github.com/apache/trafficserver/commit/a1665e58a67f2dafeeee54badcb503b479b53a97) built and installed on docs, I see the following heap-use-after-free reported by ASan:

```
=================================================================                                                                                                                                                                                                                                                             
==14752==ERROR: AddressSanitizer: heap-use-after-free on address 0x613000032068 at pc 0x55921eaaa705 bp 0x7ff0ae074a60 sp 0x7ff0ae074a50                                                                                                                                                                                      
READ of size 8 at 0x613000032068 thread T5 ([ET_NET 3])                                                                                                                                                                                                                                                                       
    #0 0x55921eaaa704 in QUICStreamAdapter::stream() /home/bneradt/src/trafficserver_10/include/iocore/net/quic/QUICStreamAdapter.h:37                                                                                                                                                                                        
    #1 0x55921eac0a23 in HQTransaction::get_transaction_id() const /home/bneradt/src/trafficserver_10/src/proxy/http3/Http3Transaction.cc:211                                                                                                                                                                                 
    #2 0x55921eabae7a in HQSession::get_transaction(unsigned long) /home/bneradt/src/trafficserver_10/src/proxy/http3/Http3Session.cc:85                                                                                                                                                                                      
    #3 0x55921eab1144 in Http3App::_handle_bidi_stream_on_read_ready(int, VIO*) /home/bneradt/src/trafficserver_10/src/proxy/http3/Http3App.cc:291                                                                                                                                                                            
    #4 0x55921eaaf159 in Http3App::main_event_handler(int, Event*) /home/bneradt/src/trafficserver_10/src/proxy/http3/Http3App.cc:149                                                                                                                                                                                         
    #5 0x55921e0cd6ea in Continuation::handleEvent(int, void*) /home/bneradt/src/trafficserver_10/include/iocore/eventsystem/Continuation.h:228                                                                                                                                                                               
    #6 0x55921ea42964 in EThread::process_event(Event*, int) /home/bneradt/src/trafficserver_10/src/iocore/eventsystem/UnixEThread.cc:162                                                                                                                                                                                     
[...]

PR #11196, opened 2024-03-28
Don't delete QUIC stream until both read and write complete

maskit:
This is for #11170.

Looks like it happens when write_compete happens before read_complete, and there was a FIXME comment about it. This PR fixes it.

### DIFF
--- a/src/proxy/http3/Http3App.cc
+++ b/src/proxy/http3/Http3App.cc
@@ -153,11 +153,10 @@ Http3App::main_event_handler(int event, Event *data)
     break;
   case VC_EVENT_READ_COMPLETE:
     adapter->clear_read_complete_event(data);
-    // Calling read_ready handlers because there's no need to do different things
     if (is_bidirectional) {
-      this->_handle_bidi_stream_on_read_ready(event, vio);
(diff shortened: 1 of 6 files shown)


src/proxy/http3/Http3App.cc
```diff
@@ -282,6 +281,17 @@ Http3App::_handle_uni_stream_on_read_ready(int /* event */, VIO *vio)
   }
 }
 
+void
+Http3App::_handle_uni_stream_on_read_complete(int event, VIO *vio)
+{
+  this->_handle_uni_stream_on_read_ready(event, vio);
+
+  QUICStreamVCAdapter *adapter = static_cast<QUICStreamVCAdapter *>(vio->vc_server);
```
brbzull0, line 289: would it make sense to check for `adapter != nullptr` or maybe an assert?

maskit: That's already checked in `Http3App::main_event_handler`.

brbzull0 (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11198, opened 2024-03-29
Add default toolchain file and adjust RelWithDebInfo flags

cmcfarlen:
Cmake requires changes to the build types and flags to be done in a toolchain file.  This PR introduces that and uses it in the top level presets.  If you aren't using presets (you should use presets) then you must specify the toolchain file on the command line if you want the updated flags: (i.e. `cmake -B build -S. -DCMAKE_TOOLCHAIN_FILE=cmake/default.toolchain.cmake`)

With presets you can just say `cmake --preset reldbg` and you'll be good.  You can also base a user preset on `default` or other that inherits `default` and that will also work.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -11,8 +11,9 @@
       "displayName": "Default build with ninja",
       "description": "Default build using Ninja generator",
       "generator": "Ninja",
-      "binaryDir": "${sourceDir}/build-default",
+      "binaryDir": "${sourceDir}/build-${presetName}",
       "cacheVariables": {
+        "CMAKE_TOOLCHAIN_FILE": "cmake/default.toolchain.cmake",
         "CMAKE_BUILD_TYPE": "Debug",
         "CMAKE_COMPILE_WARNING_AS_ERROR": "ON"
       }
@@ -38,14 +39,26 @@
     },
     {
       "name": "release",
+      "inherits": ["default"],
       "displayName": "Release build",
       "description": "Release build with Ninja generator",
       "generator": "Ninja",
-      "binaryDir": "${sourceDir}/build-release",
       "cacheVariables": {
         "CMAKE_BUILD_TYPE": "Release",
-        "CMAKE_COMPILE_WARNING_AS_ERROR": "OFF",
-        "CMAKE_INSTALL_PREFIX": "/opt/ats",
+        "CMAKE_COMPILE_WARNING_AS_ERROR": "ON",
+        "BUILD_EXPERIMENTAL_PLUGINS": "ON"
+      }
+    },
+    {
+      "name": "reldbg",
+      "inherits": ["default"],
+      "displayName": "Release build with debug info",
+      "description": "Release build with Ninja generator and debug info",
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/build-reldbg",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "RelWithDebInfo",
+        "CMAKE_COMPILE_WARNING_AS_ERROR": "ON",
         "BUILD_EXPERIMENTAL_PLUGINS": "ON"
       }
     },
@@ -65,7 +78,6 @@
       "displayName": "development",
       "description": "Development Presets",
       "inherits": ["default"],
-      "binaryDir": "${sourceDir}/build-${presetName}",
       "generator": "Ninja",
       "cacheVariables": {
         "CMAKE_COLOR_DIAGNOSTICS": "ON",
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -45,7 +45,7 @@
       "generator": "Ninja",
       "cacheVariables": {
         "CMAKE_BUILD_TYPE": "Release",
-        "CMAKE_COMPILE_WARNING_AS_ERROR": "ON",
+        "CMAKE_COMPILE_WARNING_AS_ERROR": "OFF",
         "BUILD_EXPERIMENTAL_PLUGINS": "ON"
       }
     },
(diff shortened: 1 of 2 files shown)


CMakePresets.json
```diff
@@ -11,8 +11,9 @@
       "displayName": "Default build with ninja",
       "description": "Default build using Ninja generator",
       "generator": "Ninja",
-      "binaryDir": "${sourceDir}/build-default",
+      "binaryDir": "${sourceDir}/build-${presetName}",
```
JosiahWI, line 14: Why is there an undocumented bugfix here?

CMakePresets.json
```diff
@@ -38,14 +39,26 @@
...
       "displayName": "Release build",
       "description": "Release build with Ninja generator",
       "generator": "Ninja",
-      "binaryDir": "${sourceDir}/build-release",
       "cacheVariables": {
         "CMAKE_BUILD_TYPE": "Release",
-        "CMAKE_COMPILE_WARNING_AS_ERROR": "OFF",
-        "CMAKE_INSTALL_PREFIX": "/opt/ats",
+        "CMAKE_COMPILE_WARNING_AS_ERROR": "ON",
+        "BUILD_EXPERIMENTAL_PLUGINS": "ON"
```
JosiahWI, line 49: This changes the behavior of the release build configuration. I find the change to enable -Werror particularly intriguing; was the previous behavior incorrect? Why is this change done here?

CMakePresets.json
```diff
@@ -38,14 +39,26 @@
...
+    {
+      "name": "reldbg",
+      "inherits": ["default"],
+      "displayName": "Release build with debug info",
+      "description": "Release build with Ninja generator and debug info",
+      "generator": "Ninja",
+      "binaryDir": "${sourceDir}/build-reldbg",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "RelWithDebInfo",
+        "CMAKE_COMPILE_WARNING_AS_ERROR": "ON",
```
JosiahWI, line 61: This flag and `COMPILE_WARNING_AS_ERROR` property were both added in CMake 3.24. They will have no effect on any CI builds running an earlier version. I think this should be documented somewhere, as it is a potential weak link in our CI if we forget about it.

JosiahWI (changes requested):
What are the reasons for using a toolchain instead of setting the flag directly in the presets?

I was able to reproduce and verify the desired behavior *without* a toolchain on CMake 3.20 by following these steps:
  * Checkout `cmcfarlen/normalize-release-debinfo`.
  * Edit CMakePresets.json to remove the toolchain file from defaults.
  * Create a new build directory by running `cmake --preset reldbg`.
  * Run `cmake --build build --target tscore -v`.
  * Verify that the optimization flag `-O2` was passed to the compiler.
  * Edit CMakePresets.json to add `"CMAKE_CXX_FLAGS_RELWITHDEBINFO": "-O3",` to the `reldbg` preset.
   * Rerun `cmake --preset reldbg` without removing the `build-reldbg` directory.
   * Run `grep "CXX_FLAGS" build-reldbg/CMakeCache.txt`.
   * Confirm by the `grep` output that the `CMAKE_CXX_FLAGS_RELWITHDEBINFO` is now set to `"-O3"` in the cache for this build tree.
   * Rerun `cmake --build build --target tscore -v`.
   * Confirm that the `-O3` flag is now passed to the compiler, and the `-O2` flag is not passed.

CMakePresets.json
cmcfarlen: Oh, I tried to homogenize these but maybe they weren't supposed to be the same? idk, I'll clean it up.

CMakePresets.json
cmcfarlen: This just prevents duplication of binaryDir in the presets that inherit default.

CMakePresets.json
```diff
@@ -65,7 +78,6 @@
       "displayName": "development",
       "description": "Development Presets",
       "inherits": ["default"],
-      "binaryDir": "${sourceDir}/build-${presetName}",
```
cmcfarlen, line 68: moved this line up to default

cmcfarlen:
> What are the reasons for using a toolchain instead of setting the flag directly in the presets?
> 
> I was able to reproduce and verify the desired behavior _without_ a toolchain on CMake 3.20 by following these steps:
> 
> * Checkout `cmcfarlen/normalize-release-debinfo`.
> * Edit CMakePresets.json to remove the toolchain file from defaults.
> * Create a new build directory by running `cmake --preset reldbg`.
> * Run `cmake --build reldbg --target tscore -v`.
> * Verify that the optimization flag `-O2` is passed to the compiler.
> * Edit CMakePresets.json to add `"CMAKE_CXX_FLAGS_RELWITHDEBINFO": "-O3",` to the `reldbg` preset.
> * Rerun `cmake --preset reldbg` without removing the `build-reldbg` directory.
> * Run `grep "CXX_FLAGS" build-reldbg/CMakeCache.txt`.
> * Confirm by the `grep` output that the `CMAKE_CXX_FLAGS_RELWITHDEBINFO` is now set to `"-O3"` in the cache for this build tree.
> * Rerun `cmake --build reldbg --target tscore -v`.
> * Confirm that the `-O3` flag is now passed to the compiler, and the `-O2` flag is not passed.

In my testing, when setting this outside of toolchain, I was getting what I set pre-pended to the default, so the value was `-O3 -g -DNDEBUG -O2 -g -DNDEBUG`. I could not get it to replace it without toolchain.

JosiahWI:
How exactly did you set it for testing that resulted in the values getting prepended to the default ones? If you set the `CXXFLAGS` environment variable, that will set the `CMAKE_CXX_FLAGS` CMake variable, which will be __prepended__ to `CMAKE_CXX_FLAGS_RELWITHDEBINFO`. See https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS.html#variable:CMAKE_%3CLANG%3E_FLAGS.

PR #11165, opened 2024-03-15
Fix a build issue with autest plugins

moonchen:
- Make the logic for autest plugins match ats plugins.
- Allow undefined symbols on Mac builds, to build missing_mangled_definition.

PR #11200, opened 2024-03-30
Apple Clang: missing_mangled_definition undefined symbol

bneradt:
The missing_mangled_definition tests a scenario in which there is a missing symbol. The verify_global_plugin test exercises this mistake. This patch tells Apple clang that this missing symbol is intentional so that the build works as expected.


moonchen:
I think this is a duplicate of #11165

bneradt:
Closing since, as @moonchen , his #11165 takes care of this.

Issue #11197, opened 2024-03-29
Decrypting private SSL keys is broken due to missing pass phrase callback in SSLPrivateKeyHandler


kjm33:
Apache Traffic Server - traffic_server - 9.2.3 - (build # 101116 on Oct 11 2023 at 16:11:14)

/etc/trafficserver/ssl_multicert.config
dest_ip=*   ssl_cert_name=/etc/pki/tls/certs/cert.pem ssl_key_name=/etc/pki/tls/private/encrypted.key ssl_key_dialog=exec:/tmp/show_key

/tmp/show_key
#!/bin/sh
/usr/bin/echo SECRET_KEY

journalctl -xeu trafficserver.service
Mar 29 06:10:59 XXX traffic_manager[1432308]: Enter PEM pass phrase:

Errors:
[Mar 29 06:10:59.151] traffic_server DEBUG: <SSLUtils.cc:1024 (SSLPrivateKeyHandler)> (ssl_load) failed to load server private key (-----BEGIN RSA PRIVATE KEY-----
Proc-Type: 4,ENCRY) from /etc/pki/tls/private/encrypted.key
[Mar 29 06:10:59.151] traffic_server ERROR: SSL::22753122588160:error:0909006C:PEM routines:get_name:no start line:crypto/pem/pem_lib.c:745:Expecting: TRUSTED CERTIFICATE
[Mar 29 06:10:59.151] traffic_server ERROR: SSL::22753122588160:error:2807106B:UI routines:UI_process:processing error:crypto/ui/ui_lib.c:545:while reading strings
[Mar 29 06:10:59.151] traffic_server ERROR: SSL::22753122588160:error:0906406D:PEM routines:PEM_def_callback:problems getting password:crypto/pem/pem_lib.c:59
[Mar 29 06:10:59.151] traffic_server ERROR: SSL::22753122588160:error:0906A068:PEM routines:PEM_do_header:bad password read:crypto/pem/pem_lib.c:434

Root cause:
Missing pass phrase callback parameter (passwd_cb) for PEM_read_bio_PrivateKey function in SSLPrivateKeyHandler function.
I guess that was broken in https://github.com/apache/trafficserver/commit/7dbb6cb1881aa344edb1ab59dcbe3342171ea0a7#diff-cb865c0bc65fb8ef103a206282b78a50e0c1c9e93ca6713322ba9df9d921e6c9 because loading the same key with 
[...]

PR #11201, opened 2024-04-01
Restore private key passphrase behavior

moonchen:
We accidentally removed the callbacks for private key passphrases. This patch restores the old behavior.


moonchen:
Fixes #11197

brbzull0:
[approve ci rocky]

moonchen:
> What testing was done of this? Would be nice to have Au test coverage.

Good call.  I added an autest and found out that the old code was broken.

ywkaras (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11204, opened 2024-04-02
Don't send "None" as a server name

maskit:
"None" is sent as a server name if `args` doesn't have `sni`. I don't think it's intentional.


brbzull0 (approved):
make sense.
Thanks

PR #11208, opened 2024-04-03
Clean up some of Cripts strings with modern libSWOC

zwoop:
With the introduction of libswoc into ATS 10, we can now rely on the newer versions of TextView, rather than the crippled old versions in ATS v9.2.x.

### DIFF
--- a/include/cripts/Lulu.hpp
+++ b/include/cripts/Lulu.hpp
@@ -70,11 +70,10 @@ public:
   constexpr StringViewMixin() = default;
   constexpr StringViewMixin(const char *s) { _value = mixin_type(s, strlen(s)); }
   constexpr StringViewMixin(const char *s, mixin_type::size_type count) { _value = mixin_type(s, count); }
-  constexpr StringViewMixin(Cript::string_view &str) { _value = str; }
+  constexpr StringViewMixin(const Cript::string_view &str) { _value = str; }
 
   virtual self_type &operator=(const mixin_type str) = 0;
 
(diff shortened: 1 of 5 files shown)


zwoop:
[approve ci autest]

zwoop:
[approve ci autest]

include/cripts/Lulu.hpp
```diff
@@ -74,7 +74,6 @@ template <typename ChildT> class StringViewMixin
```
ywkaras, line 74: Seems like this should be:
```
constexpr StringViewMixin(const Cript::string_view &str) { _value = str; }
```
or just pass by value.

zwoop: Probably. Alan kinda convinced me that these gets optimized in funny ways, and I also couldn't get a good answer if it should be passed by reference or not.

ywkaras: The short rule of thumb (since C++17) is, `f(T v)` and `f(const T &v)` are generally the same if the function is inlined and doesn't change `v`.  With `T g();`, `f(g())` works for both definitions of `f`, because C++ allows a const reference to refer to an unnamed temporary.  But C++ doesn't allow a non-const reference to refer to an unnamed temporary.

ywkaras:
[approve ci autest]

ywkaras (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11209, opened 2024-04-04
http2: close stream on sending RST_STREAM

masaori335:
When ATS receives HTTP/2 frame, sometimes stream error happens - e.g. malformed request. In such case, ATS sends RST_STREAM frame and make the stream state closed, but the stream is still in the `Http2ConnectionState::stream_list` and treat as active stream. It's possible that this makes miscalculation of current active stream number.


bneradt (approved):

cmcfarlen:
Cherry-picked to v10.0.x

zwoop:
Cherry-picked to v9.2.x

PR #11210, opened 2024-04-04
http2: Set SSN active on create_stream

masaori335:
Prior to the change, if a HTTP/2 stream is open on a session, but it doesn't make new transaction by some reason, the inactivity timeout of the connection never get fired. 

Current ATS v10.0.x cancels inactivity timeout of proxy session in the `Http2ConnectionState::create_stream` (introduced by 7d5e673309b985d8e699328d58b55e2df89e861b). We need to get back the session inactivity timeout when all streams are gone, but there is a activity check of proxy session. To pass the check, we need to make the proxy session active.

https://github.com/apache/trafficserver/blob/66940f5768641b22dd36e7ae172fa72dcacdcb11/src/proxy/http2/Http2ConnectionState.cc#L1994-L1998


bneradt:
[approve ci autest]

bneradt (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11211, opened 2024-04-04
Fix RelWithDebInfo to be release with debug info

cmcfarlen:


CMakeLists.txt
```diff
@@ -15,6 +15,8 @@
 #
 #######################
 
+set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-O3 -g -DNDEBUG")
```
JosiahWI, line 18: Could we add a comment explaining why we introduced this? I'm familiar with how these seemingly little, insignificant settings accumulate in the build until nobody can decipher it anymore.

JosiahWI (changes requested):
Looks good, but could we add a brief comment explaining this was done for consistency between build types?

JosiahWI (approved):
Thanks, looks good.

PR #11220, opened 2024-04-06
Fix parent proxy connection stat overcounting

bneradt:
The proxy.process.http.current_parent_proxy_connections and proxy.process.http.total_parent_proxy_connections stats are each incremented in create_server_txn, which is a fine place to increment them, but this assumes that each server transaction is associated with a single origin connection. This is not the case when sessions are reused. This fixes the counting by only incrementing these stats for the very first transaction in the session.


masaori335 (approved):
Looks reasonable.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11222, opened 2024-04-08
Terminate H2 connection by COMPRESSION_ERROR

masaori335:
[RFC 9113 4.3. Field Section Compression and Decompression](https://datatracker.ietf.org/doc/html/rfc9113#name-field-section-compression-a) says below. However, current implementation doesn't follow this ( returning stream error without HPACK decoding )

> An endpoint receiving [HEADERS](https://datatracker.ietf.org/doc/html/rfc9113#HEADERS), [PUSH_PROMISE](https://datatracker.ietf.org/doc/html/rfc9113#PUSH_PROMISE), or [CONTINUATION](https://datatracker.ietf.org/doc/html/rfc9113#CONTINUATION) frames needs to reassemble field blocks and perform decompression even if the frames are to be discarded. A receiver MUST terminate the connection with a [connection error](https://datatracker.ietf.org/doc/html/rfc9113#ConnectionErrorHandler) ([Section 5.4.1](https://datatracker.ietf.org/doc/html/rfc9113#ConnectionErrorHandler)) of type [COMPRESSION_ERROR](https://datatracker.ietf.org/doc/html/rfc9113#COMPRESSION_ERROR) if it does not decompress a field block.

As a quick fix, replacing stream error with connection error if it happens before HPACK decoding. For the long term, it'd be better to trying HPACK decoding and returning stream error.

### DIFF
--- a/src/proxy/http2/Http2ConnectionState.cc
+++ b/src/proxy/http2/Http2ConnectionState.cc
@@ -320,6 +320,13 @@ Http2ConnectionState::rcv_headers_frame(const Http2Frame &frame)
       stream     = this->create_stream(stream_id, error);
       new_stream = true;
       if (!stream) {
+        // Terminate the connection with COMPRESSION_ERROR because we don't decompress the field block in this HEADERS frame.
+        // TODO: try to decompress to keep HPACK Dynamic Table in sync.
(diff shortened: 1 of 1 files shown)


bneradt (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11223, opened 2024-04-08
find PkgConfig earlier

cmcfarlen:
I found that some combinations of setting will error on pkg-config checks.  This will find pkgconfig earlier to prevent that from happening.  (i.e. set ENABLE_MAGICK=false)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,6 +75,7 @@ execute_process(
 
 # Options
 include(AutoOptionHelpers)
+find_package(PkgConfig)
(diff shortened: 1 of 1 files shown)


JosiahWI:
This does not require PkgConfig to be present (and shouldn't) but many plugins do. Will `ENABLE_MAGICK=true` work if PkgConfnig is *not* found?

JosiahWI (approved):
PkgConfig is by the top-level CMakeLists.txt to find PCRE2. Thank you for fixing this, @cmcfarlen!

cmcfarlen:
Cherry-picked to v10.0.x

PR #11224, opened 2024-04-08
Manage storage for ssl hooks

cmcfarlen:


src/iocore/net/SSLAPIHooks.cc
```diff
@@ -22,11 +22,11 @@
...
-SSLAPIHooks *g_ssl_hooks = nullptr;
-
-void
-init_global_ssl_hooks()
+SSLAPIHooks *
+SSLAPIHooks::instance()
 {
-  g_ssl_hooks = new SSLAPIHooks;
+  static auto hooks = std::make_unique<SSLAPIHooks>();
+  return hooks.get();
```
ywkaras, line 31: Why not just?
```
static SSLAPIHooks hooks;
return &hooks;
```

cmcfarlen: It was created with new previously so did not want to change that, but did want to auto manage the pointer.

ywkaras (approved):
Uncle

cmcfarlen:
Cherry-picked to v10.0.x

PR #11228, opened 2024-04-08
branch-autest: use openssl-quic

bneradt:


bryancall:
With this change does it mean that CI wont build with BoringSSL and QUIC?

traeak:
This looks okay from the technical perspective.  I'm don't know enough about the different ways to implement h3.

traeak (approved):

bneradt:
I'll close this PR and work with Damian on using quiche built against openssl. We'll need an update to the build_h3_tools.sh script to build this type of quiche (it currently just builds quiche against boringssl).

PR #11237, opened 2024-04-10
cleanup asan odr violations

cmcfarlen:
linking ts::tsutil in a plugin with create ODR violations with DbgCtl globals.

The fix in compress.cc will still leak at shutdown, but the asan complaint was in the verify plugin which doesn't call TSPluginInit.   I'm not sure how to run code at plugin "unload" to clean up.

### DIFF
--- a/plugins/background_fetch/CMakeLists.txt
+++ b/plugins/background_fetch/CMakeLists.txt
@@ -16,5 +16,5 @@
 #######################
 
 add_atsplugin(background_fetch background_fetch.cc configs.cc headers.cc rules.cc)
-target_link_libraries(background_fetch PRIVATE libswoc::libswoc ts::tsutil)
+target_link_libraries(background_fetch PRIVATE libswoc::libswoc)
 verify_global_plugin(background_fetch)
--- a/plugins/compress/compress.cc
+++ b/plugins/compress/compress.cc
@@ -70,7 +70,7 @@ const int BROTLI_LGW               = 16;
 
 static const char *global_hidden_header_name = nullptr;
 
-static TSMutex compress_config_mutex = TSMutexCreate();
+static TSMutex compress_config_mutex = nullptr;
 
 // Current global configuration, and the previous one (for cleanup)
 Configuration *cur_config  = nullptr;
@@ -1017,6 +1017,7 @@ void
 TSPluginInit(int argc, const char *argv[])
 {
   const char *config_path = nullptr;
+  compress_config_mutex   = TSMutexCreate();
 
   if (argc > 2) {
     fatal("the compress plugin does not accept more than 1 plugin argument");
--- a/plugins/experimental/sslheaders/CMakeLists.txt
+++ b/plugins/experimental/sslheaders/CMakeLists.txt
@@ -16,7 +16,8 @@
 #######################
 
 add_library(sslhdr STATIC expand.cc util.cc)
-target_link_libraries(sslhdr PRIVATE OpenSSL::SSL libswoc::libswoc)
+target_link_libraries(sslhdr PRIVATE OpenSSL::SSL libswoc::libswoc # transitive
+)
 set_target_properties(sslhdr PROPERTIES POSITION_INDEPENDENT_CODE TRUE)
 
 if(BUILD_TESTING)
--- a/plugins/experimental/sslheaders/CMakeLists.txt
+++ b/plugins/experimental/sslheaders/CMakeLists.txt
@@ -16,7 +16,7 @@
 #######################
 
 add_library(sslhdr STATIC expand.cc util.cc)
-target_link_libraries(sslhdr PRIVATE OpenSSL::SSL ts::tsutil)
+target_link_libraries(sslhdr PRIVATE OpenSSL::SSL)
 set_target_properties(sslhdr PROPERTIES POSITION_INDEPENDENT_CODE TRUE)
 
 if(BUILD_TESTING)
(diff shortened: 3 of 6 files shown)


plugins/experimental/sslheaders/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@
 #######################
 
 add_library(sslhdr STATIC expand.cc util.cc)
-target_link_libraries(sslhdr PRIVATE OpenSSL::SSL ts::tsutil)
```
JosiahWI, line 19: Should it also be removed on L24?

plugins/experimental/tls_bridge/CMakeLists.txt
```diff
@@ -17,5 +17,5 @@
 
 add_atsplugin(tls_bridge tls_bridge.cc)
 
-target_link_libraries(tls_bridge PRIVATE ts::tsutil libswoc::libswoc)
```
JosiahWI, line 20: Isn't the underlying issue that `tls_bridge` depends on this? This will fix the error, but it won't fix the problem.

plugins/compress/compress.cc
```diff
@@ -70,7 +70,7 @@ const int BROTLI_LGW               = 16;
 
 static const char *global_hidden_header_name = nullptr;
 
-static TSMutex compress_config_mutex = TSMutexCreate();
+static TSMutex compress_config_mutex = nullptr;
```
JosiahWI, line 73: Was this change (deferring the initialization to `TSPluginInit`) supposed to be included?

plugins/background_fetch/CMakeLists.txt
```diff
@@ -16,5 +16,5 @@
 #######################
 
 add_atsplugin(background_fetch background_fetch.cc configs.cc headers.cc rules.cc)
-target_link_libraries(background_fetch PRIVATE libswoc::libswoc ts::tsutil)
```
JosiahWI, line 19: Isn't the underlying issue that `background_fetch` depends on this? Maybe it could get all the needed utilities directly from libswoc?

plugins/remap_stats/CMakeLists.txt
```diff
@@ -17,5 +17,4 @@
 
 add_atsplugin(remap_stats remap_stats.cc)
 
-target_link_libraries(remap_stats PRIVATE ts::tscore)
```
JosiahWI, line 20: It seems `tscore/ink_config.h` is included by remap_stats.cc but is not actually needed. We can maybe remove it. It also includes `tscore/ink_defs.h` for the `unlikely` definition; should we duplicate the definition in the plugin to avoid the dependency?

plugins/experimental/sslheaders/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@
...
+target_link_libraries(sslhdr PRIVATE OpenSSL::SSL libswoc::libswoc)
```
JosiahWI, line 19: It seems swoc isn't included anywhere. Why is it needed here?

JosiahWI (changes requested):

plugins/compress/compress.cc
JosiahWI: Addressed in PR description.

plugins/remap_stats/CMakeLists.txt
JosiahWI: I say here "not actually needed", but "is needed and a feature flag is missing that should be there" is another possibility.

cmcfarlen: unfortunately lots of plugins include stuff from tscore, but they still cannot link tscore.  Removing includes is fine if they aren't used, but outside of scope for this PR.

plugins/experimental/sslheaders/CMakeLists.txt
cmcfarlen: something includes libswoc headers

plugins/experimental/sslheaders/CMakeLists.txt
cmcfarlen: Its fine to link in unit tests, this PR is to fix ODR violations in plugins

plugins/experimental/tls_bridge/CMakeLists.txt
JosiahWI: It uses `tscore/Regex.h`; maybe it could use PCRE directly as the slice plugin does.

plugins/experimental/sslheaders/CMakeLists.txt
JosiahWI: It looks like it isn't needed, but maybe that can be addressed in a separate PR.

plugins/experimental/sslheaders/CMakeLists.txt
JosiahWI: There may be an incorrect dependency somewhere else.

plugins/experimental/tls_bridge/CMakeLists.txt
```diff
@@ -17,5 +17,5 @@
...
+target_link_libraries(tls_bridge PRIVATE libswoc::libswoc PCRE::PCRE)
```
JosiahWI, line 20: Please mark `PCRE::PCRE` linkage as transitive so it can be found for cleanup later.

JosiahWI (approved):
Thanks.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11238, opened 2024-04-10
header_rewrite: add debug to accompany parse_line errors

traeak:
If there's an error while parsing a line of a header rewrite file TSError is invoked.  This PR also adds a Dbg() call to help figure out which header rewrites actually have the error in the case that errors go to diags.log and debug goes to another file.

Since parser.cc is used in a catch test the debug has to be put in functions that call into it.


shukitchan (approved):
Thanks. That will be useful.

PR #11239, opened 2024-04-10
Fix asan leak errors for cachevol test

cmcfarlen:


cmcfarlen:
[approve ci autest]

src/iocore/cache/unit_tests/test_CacheVol.cc
```diff
@@ -365,6 +369,8 @@ class CacheVolTest : public CacheInit
   cache_init_success_callback(int event, void *e) override
   {
     // Test
+    ClearCacheVolList(&cp_list, cp_list_len);
```
JosiahWI, line 372: I'm unfamiliar with this code; why is this called both here and also on line 224?

cmcfarlen: Because not calling it here causes a memory leak.  Turns out the structures this cleans up are initialized before this callback is called.  Global mutable state is bad mkay?

JosiahWI: Ah, so the structures are cleaned up after initialization completes because they are not needed anymore. That makes sense now. Thank you!

brbzull0 (approved):
Looks good to me!
Thanks.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11240, opened 2024-04-10
IP Allow: document when rules are applied

bneradt:
I have noticed when discussing the IP Allow feature that there is some confusion over when various types of rules are applied. This PR documents the three stages in which any given rule could be applied.


zwoop (approved):
I still think this is all weird, but sounds like we didn't do exactly what was agreed a number of years ago. My expectations would have been that this in remap.config would always be honored, but doesn't seem it is:

```
.deactivatefilter ip_allow
```

cmcfarlen:
Cherry-picked to v10.0.x

PR #11241, opened 2024-04-11
CID 1534718: txn_box - Fix invalid use of signed index.

brbzull0:
It seems that the code was trying to deduce and store the kv separator from
a query string, this does not make sense as the separator char is known beforehand(used to split the kv), so instead, we just leave the default sep.

### DIFF
--- a/plugins/experimental/txn_box/plugin/src/query.cc
+++ b/plugins/experimental/txn_box/plugin/src/query.cc
@@ -51,12 +51,13 @@ static inline constexpr TextView ARG_NOCASE = "nc";
 static inline constexpr TextView ARG_REVERSE = "rev";
 
 struct QPair {
-  using self_type = QPair;
+  static constexpr char Sep = '=';
+  using self_type           = QPair;
 
   TextView name;
   TextView value;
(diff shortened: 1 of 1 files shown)


plugins/experimental/txn_box/plugin/src/query.cc
```diff
@@ -88,12 +88,9 @@ query_take_qpair(TextView &qs)
   // If there's anything left, look for kv pair.
   if (qs) {
     auto v{qs.clip_prefix_of([](char c) { return c != '&' && c != ';'; })};
-    auto k = v.take_prefix_at('=');
+    auto k = v.take_prefix_at(QPair::Sep);
     QPair zret{k, v};
     zret.elt_sep = elt_sep;
-    if (v.begin() > k.end()) {
-      zret.kv_sep = v[-1];
```
brbzull0, line 95: I think this can be ignored, it is using the same separator to find(`take_prefix_at`) and to place it here, so probably is the same just to leave the default which will not be overwritten.

cmcfarlen (approved):

cmcfarlen:
Cherry-picked to v10.0.x

Issue #10681, opened 2023-10-26
ATS 10 docs: unexpected tunnel_handler event


bneradt:
I noticed this on docs today:

```gdb
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff714c859 in __GI_abort () at abort.c:79
#2  0x0000555555effb85 in ink_abort (message_format=0x55555606b484 "%s:%d: failed assertion `%s`") at ink_error.cc:99
#3  0x0000555555efbbd3 in _ink_assert (expression=0x555555fb3418 "event == HTTP_TUNNEL_EVENT_DONE || event == VC_EVENT_INACTIVITY_TIMEOUT", file=0x555555fb12da "HttpSM.cc", line=3006) at ink_assert.cc:37
#4  0x0000555555b16eb9 in HttpSM::tunnel_handler (this=0x7fffe802d040, event=104, data=0x7fffe8037300) at HttpSM.cc:3006
#5  0x0000555555b152ce in HttpSM::main_handler (this=0x7fffe802d040, event=104, data=0x7fffe8037300) at HttpSM.cc:2666
#6  0x0000555555aa04ef in Continuation::handleEvent (this=0x7fffe802d040, event=104, data=0x7fffe8037300) at /home/maskit/trafficserver/iocore/eventsystem/I_Continuation.h:228
#7  0x0000555555c25843 in Http2Stream::main_event_handler (this=0x7fffe8037050, event=104, edata=0x7fffa0000cd0) at Http2Stream.cc:232
#8  0x0000555555aa04ef in Continuation::handleEvent (this=0x7fffe8037050, event=104, data=0x7fffa0000cd0) at /home/maskit/trafficserver/iocore/eventsystem/I_Continuation.h:228
#9  0x0000555555f90677 in EThread::process_event (this=0x7ffff41aa010, e=0x7fffa0000cd0, calling_code=104) at UnixEThread.cc:150
#10 0x0000555555f908c7 in EThread::process_queue (this=0x7ffff41aa010, NegativeQueue=0x7fffef9b0170, ev_count=0x7fffef9b0124, nq_count=0x7fffef9b0120) at UnixEThread.cc:185
#11 0x0000555555f90bd6 in EThread::execute_regular (this=0x7ffff41aa010) at UnixEThread.cc:241
[...]

PR #11242, opened 2024-04-11
HttpSM::tunnel_handler: Handle WRITE events

bneradt:
Multiple HttpSM handlers have logic to handle write events after an EOS is sent. This applies that logic to the HttpSM::tunnel_handler as well, avoiding an unneccesary assertion.

Fixes: #10681

### DIFF
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -2980,6 +2980,12 @@ HttpSM::tunnel_handler(int event, void *data)
 {
   STATE_ENTER(&HttpSM::tunnel_handler, event);
 
+  // If we had already received EOS, just go away. We would sometimes see
+  // a WRITE event appear after receiving EOS from the server connection
+  if ((event == VC_EVENT_WRITE_READY || event == VC_EVENT_WRITE_COMPLETE) && server_entry->eos) {
+    return 0;
+  }
+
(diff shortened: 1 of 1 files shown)


bneradt:
[approve ci]

bneradt:
[approve ci autest]

src/proxy/http/HttpSM.cc
```diff
@@ -2980,6 +2980,12 @@ HttpSM::tunnel_handler(int event, void *data)
 {
   STATE_ENTER(&HttpSM::tunnel_handler, event);
 
+  // If we had already received EOS, just go away. We would sometimes see
+  // a WRITE event appear after receiving EOS from the server connection
+  if ((event == VC_EVENT_WRITE_READY || event == VC_EVENT_WRITE_COMPLETE) && server_entry->eos) {
```
masaori335, line 2985: As your pointed out, other event handlers have a guard like this - e.g. `HttpSM::state_read_server_response_header`. 

https://github.com/apache/trafficserver/blob/14d5b2ce401446a13b2fef4ab8dbc7cf890acf25/src/proxy/http/HttpSM.cc#L1873-L1877

So this should be good as a workaround fix. However, I prefer to move the state that ignore events when HttpSM got EOS. Because HttpSM is a state machine. It should not controlled by a bunch of flags .

masaori335 (approved):
Let's fix the crash.

bneradt:
> Let's fix the crash.

Thank you @masaori335

Issue #11131, opened 2024-03-05
CI txn_box ramp autest failure


bneradt:
The [ramp](https://github.com/apache/trafficserver/blob/85a4cb96b8c08449403929f35a08ef680901c484/tests/gold_tests/pluginTest/txn_box/ramp/ramp.test.py) test is a relatively new txn_box autest that is failing with some frequency in CI.

Here is a recent example:
https://ci.trafficserver.apache.org/view/Github/job/Github_Builds/job/autest/19066/

The failure is due to a failed HttpSM assertion:

```gdb
[Mar  5 17:38:02.479] traffic_server NOTE: Traffic Server is running unprivileged, not switching to user 'nobody'
Fatal: ../src/proxy/http/HttpSM.cc:2979: failed assertion `event == HTTP_TUNNEL_EVENT_DONE || event == VC_EVENT_INACTIVITY_TIMEOUT`
NOTE: using the environment variable TS_RUNROOT
[Mar  5 17:38:02.767] traffic_crashlo NOTE: crashlog started, target=16830, debug=false syslog=true, uid=1200 euid=1200
[Mar  5 17:38:02.768] traffic_crashlo NOTE: logging to 0x1cb1370
[Mar  5 17:38:02.769] traffic_crashlo ERROR: wrote crash log to /tmp/sandbox/ramp/ts/log/crash-2024-03-05-173802.log
traffic_server: received signal 6 (Aborted)
traffic_server - STACK TRACE: 
/tmp/sandbox/ramp/ts/bin/traffic_server(crash_logger_invoke(int, siginfo_t*, void*)+0xce)[0x881b86]
/lib64/libc.so.6(+0x3e9a0)[0x7fa0b885e9a0]
/lib64/libc.so.6(+0x90834)[0x7fa0b88b0834]
/lib64/libc.so.6(raise+0x1e)[0x7fa0b885e8ee]
/lib64/libc.so.6(abort+0xdf)[0x7fa0b88468ff]
/tmp/sandbox/ramp/ts/bin/traffic_server(ink_warning(char const*, ...)+0x0)[0x8b7538]
/tmp/sandbox/ramp/ts/bin/traffic_server[0x8b3c4e]
/tmp/sandbox/ramp/ts/bin/traffic_server(HttpSM::tunnel_handler(int, void*)+0x12a)[0x915cfc]
/tmp/sandbox/ramp/ts/bin/traffic_server(HttpSM::main_handler(int, void*)+0x254)[0x914486]
[...]

PR #11243, opened 2024-04-11
txn_box ramp autest reliability improvements

bneradt:
This updates the ramp test to actually ramp traffic to a dedicated
"staging" server. It also reduces the number of transactions to 100.
Both of these changes makes the test more reliable against spurious
connection issues that caused the test to be flakey, especially in CI.
    
Fixes: #11131

### DIFF
--- a/tests/gold_tests/pluginTest/txn_box/ramp/ramp.replay.yaml
+++ b/tests/gold_tests/pluginTest/txn_box/ramp/ramp.replay.yaml
@@ -7,7 +7,9 @@ meta:
       select:
       - lt: 300
         do:
-          ua-req-host: "ex.two"
+        - ua-req-url-host: "ex.two"
+        - ua-req-url-port: {server_port}
+        - ua-req-scheme: "http"
 
   blocks:
   - base_request: &base_request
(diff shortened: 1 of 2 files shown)


brbzull0 (commented):
Looks good to me.
We need a non-Y approval though.

Thanks Brian!

ywkaras:
So you feel we can dispense with checking the number of lines in the transaction log?

bneradt:
> So you feel we can dispense with checking the number of lines in the transaction log?

I don't think we did though. We imported that code but didn't call it. Maybe we should call it?

ywkaras:
> > So you feel we can dispense with checking the number of lines in the transaction log?
> 
> I don't think we did though. We imported that code but didn't call it. Maybe we should call it?

Hmm the exec'ed code just defines a class and a function, and doesn't use them.

bneradt:
> > > So you feel we can dispense with checking the number of lines in the transaction log?
> > 
> > 
> > I don't think we did though. We imported that code but didn't call it. Maybe we should call it?
> 
> Hmm the exec'ed code just defines a class and a function, and doesn't use them.

Exactly. Other tests do use the functionality, but this test didn't.

moonchen (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11244, opened 2024-04-11
Regex: fix memory leak if the match buffer is too small

bryancall:
Other versions of pcre2 needed a different buffer size for matches and Regex was using `malloc()`, but not keeping track of when do `free()`.

I changed the way we fall back to using `malloc()` and pcre2 will keep track of which `free()` function to use when freeing.

I also added a unit test that will force the use of `malloc()` for the matches buffer and increased the calculation of the internal buffer, so older versions of pcre2 wont need to `malloc()`.

### DIFF
--- a/src/tsutil/Regex.cc
+++ b/src/tsutil/Regex.cc
@@ -21,13 +21,13 @@
   limitations under the License.
  */
 
-#include "tsutil/Regex.h"
+#include <tsutil/Regex.h>
+#include <tsutil/Assert.h>
 
 #define PCRE2_CODE_UNIT_WIDTH 8
 #include <pcre2.h>
 
 #include <array>
-#include <assert.h>
 #include <vector>
 #include <mutex>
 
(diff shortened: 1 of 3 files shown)


src/tsutil/Regex.cc
```diff
@@ -120,6 +120,10 @@ RegexMatches::RegexMatches(uint32_t size)
     &RegexMatches::malloc, [](void *, void *) -> void {}, static_cast<void *>(this));
 
   _match_data = pcre2_match_data_create(size, ctx);
+  if (_match_data == nullptr) {
```
cmcfarlen, line 123: Since the idea for having built-in storage is to prevent heap allocations when the passed in `size` is <= `DEFAULT_SIZE`, can we debug assert that `!(_match_data == nullptr && size == DEFAULT_SIZE)`?  This will alert devs when the expected size increases and this would allocate for the default.

cmcfarlen (changes requested):
I'd like to debug assert that heap allocation doesn't happen when size == DEFAULT_SIZE.

src/tsutil/Regex.cc
bryancall: Yes, I think having a debug assert is a good idea - I will update the PR to include that

bryancall:
rebased from master

src/tsutil/Regex.cc
```diff
@@ -172,7 +172,14 @@ RegexMatches::RegexMatches(uint32_t size)
   pcre2_general_context *ctx = pcre2_general_context_create(
     &RegexMatches::malloc, [](void *, void *) -> void {}, static_cast<void *>(this));
 
-  _MatchData::set(_match_data, pcre2_match_data_create(size, ctx));
+  pcre2_match_data *match_data = pcre2_match_data_create(size, ctx);
+  if (match_data == nullptr) {
+    // buffer was too small, allocate from heap
+    debug_assert("Buffer too small, allocating from heap");
```
cmcfarlen, line 178: You need a first parameter here for this to abort, even if it's just `false`.

bryancall: yeah, don't know what I was thinking - fixed it

cmcfarlen (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11246, opened 2024-04-12
Avoid including pcre2.h in Regex.h.

ywkaras:


brbzull0:
[approve ci autest]

ywkaras:
[approve ci autest]

include/tsutil/Regex.h
```diff
@@ -28,14 +28,11 @@
...
-#include <pcre2.h>
-
 /// @brief Match flags for regular expression evaluation.
 enum REFlags {
-  RE_CASE_INSENSITIVE = PCRE2_CASELESS,  ///< Ignore case (default: case sensitive).
-  RE_UNANCHORED       = PCRE2_MULTILINE, ///< Unanchored (DFA defaults to anchored).
-  RE_ANCHORED         = PCRE2_ANCHORED,  ///< Anchored (Regex defaults to unanchored).
+  RE_CASE_INSENSITIVE = 0x00000008u, ///< Ignore case (default: case sensitive).
+  RE_UNANCHORED       = 0x00000400u, ///< Unanchored (DFA defaults to anchored).
+  RE_ANCHORED         = 0x80000000u, ///< Anchored (Regex defaults to unanchored).
```
bneradt, line 35: These values seem pretty magical now without the PCRE2 names. Can you add a comment above the enum saying these are taken directly from pcre2.h but without the include to avoid the header.

src/tsutil/Regex.cc
```diff
@@ -23,11 +23,18 @@
...
+#include <pcre2.h>
+
 #include <array>
 #include <assert.h>
 #include <vector>
 #include <mutex>
 
+static_assert(RE_CASE_INSENSITIVE == PCRE2_CASELESS, "Update RE_CASE_INSERSITIVE for current PCRE2 version.");
+static_assert(RE_UNANCHORED == PCRE2_MULTILINE, "Update RE_MULTILINE for current PCRE2 version.");
+static_assert(RE_ANCHORED == PCRE2_ANCHORED, "Update RE_ANCHORED for current PCRE2 version.");
```
bneradt, line 36: Nice.

include/tsutil/Regex.h
```diff
@@ -63,19 +60,23 @@ class RegexMatches
...
+
+  struct _MatchData;
+  class _MatchDataPtr
+  {
+    friend struct _MatchData;
+
+  private:
+    void *_ptr = nullptr;
+  };
+  _MatchDataPtr _match_data;
```
bneradt, line 79: A comment here would be helpful to future readers. Let's explain that this effectively wraps a void* so that we can avoid requiring the pcre2.h include for the user of the Regex API. (If I'm understanding it correctly.)

bneradt (changes requested):

bneradt (commented):
Looks good to me.

We'll need a non-Yahoo approver. Might be good for @bryancall to look at this too.

masaori335 (approved):
Hiding the 3rd party library from `tsutil/Regex.h` makes sense. This also fixes the tls_bridge plugin build issue.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11252, opened 2024-04-12
CMakePresets: add opentelemetry_ROOT with CURL_ROOT

bneradt:
CURL_ROOT is only useful if opentelemetry_ROOT is set. Having the former without the latter causes a warning. This will address the warning and also get more otel_tracer build coverage.


masaori335 (approved):
Looks reasonable.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11258, opened 2024-04-15
wasm plugin: Add build support

bneradt:
# Draft status

#11257 fixes a build issue in wasm that has to land before building the wasm plugin will succeed. I'll mark this as a draft until that lands.

## Description

Initial commits of plugin/experimental/wasm on master had automake build support. When we transitioned to cmake we temporarilly dropped the automake files, leaving cmake support to a future commit. This is that future commit.


cmcfarlen (approved):

cmcfarlen:
Cherry-picked to v10.0.x

Issue #11265, opened 2024-04-18
Crash on ActivityCop of Http2ConnectionState


masaori335:
```
#0  0x00007f2f2f3d81d9 in waitpid () from /lib64/libpthread.so.0
#1  0x00005575273576a9 in crash_logger_invoke(int, siginfo_t*, void*) (signo=11, info=0x7f2f245fd0f0, ctx=0x7f2f245fcfc0) at src/traffic_server/Crash.cc:168
#2  <signal handler called>
#3  0x00007f2b03ead0e0 in ?? ()
#4  0x000055752748598b in stop (this=0x7f2c320cf0f0) at include/iocore/net/NetTimeout.h:229
#5  Http2ConnectionState::destroy() (this=0x7f2c320cf080) at /src/proxy/http2/Http2ConnectionState.cc:1358
#6  0x0000557527476fcd in Http2CommonSession::common_free(ProxySession*) (this=0x7f2c320cf078, ssn=<optimized out>) at /src/proxy/http2/Http2CommonSession.cc:133
#7  0x0000557527494bd4 in Http2ClientSession::free() (this=0x7f2c320ced60) at /src/proxy/http2/Http2ClientSession.cc:62
#8  0x00005575274954c6 in Http2ClientSession::main_event_handler(int, void*) (this=0x7f2c320ced60, event=<optimized out>, edata=<optimized out>) at /src/proxy/http2/Http2ClientSession.cc:249
#9  0x0000557527656c8a in handleEvent (this=<optimized out>, event=105, data=0x111c8448) at /include/iocore/eventsystem/Continuation.h:228
#10 read_signal_and_update(int, UnixNetVConnection*) (event=105, vc=0x7f2b23deb600) at /src/iocore/net/UnixNetVConnection.cc:87
#11 0x00005575276579b6 in UnixNetVConnection::mainEvent(int, Event*) (this=0x7f2b23deb600, event=<optimized out>, e=<optimized out>) at /src/iocore/net/UnixNetVConnection.cc:1115
#12 0x000055752764454e in InactivityCop::check_inactivity(int, Event*) (this=<optimized out>, event=<optimized out>, e=0x7f2f2e3adb60) at /src/iocore/net/UnixNet.cc:139
#13 0x0000557527690d79 in handleEvent (this=<optimized out>, event=2, data=0x7f2f2e3adb60) at /include/iocore/eventsystem/Continuation.h:228
[...]

PR #11266, opened 2024-04-18
Add nullptr guards in ActivityCop functions

masaori335:
Possible fix of #11265. I still don't get root cause of crash, but checking `_event` might work as workaround. IMO, ASan build will find the root cause.

### DIFF
--- a/include/iocore/net/NetTimeout.h
+++ b/include/iocore/net/NetTimeout.h
@@ -219,14 +219,18 @@ template <class T, class List>
 inline void
 ActivityCop<T, List>::start()
 {
-  _event = this_ethread()->schedule_every(this, HRTIME_SECONDS(_freq));
+  if (_event == nullptr) {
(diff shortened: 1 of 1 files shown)


bneradt (approved):

include/iocore/net/NetTimeout.h
```diff
@@ -219,14 +219,18 @@ template <class T, class List>
...
+  }
 }
 
 template <class T, class List>
 inline void
 ActivityCop<T, List>::stop()
 {
-  _event->cancel();
+  if (_event != nullptr) {
+    _event->cancel();
```
maskit, line 232: We won't be able to restart this? Why don't you set `nullptr` after cancelling the event?

masaori335: Sure

bneradt (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11270, opened 2024-04-19
tools: Dual build for openssl and boringssl H3 tools.

brbzull0:
Build both boringssl and openssl tools which can be used to build ATS with either SSL vendor. This can be used if you want to use quiche either with boring or openssl.

**BoringSSL**
Curl uses boringssl, not quiche as it seems very experimental, I did try this out and I wasn't happy with the outcome.

**OpenSSL**
curl and all the tools uses openssl/quictls

**Why 3 scripts instead of 1?**

I did split a single script into two to be able to run either of them individually, after you want to create a fresh set of tools and you may just need to run one. On CI  we can just keep running the existing `build_h3_tools.sh`.

If no `BASE` is set, default folder will be `/opt/h3-tools-boringssl` and `/opt/h3-tools-openssl` respectively. If `BASE` is set, then `h3-tools-boringssl` and `h3-tools-openssl` will be added to the passed `BASE`.

Any feedback is welcome. Thanks


brbzull0:
> You can build ngtcp2, curl, etc with BoringSSL if you want. 

Yes, that was sort of the plan, sorry I should have this as Draft from the beginning, I am planning to change some builds to use either bssl or ossl. 

We had some conversation with some of the guys about having a single scripts, two or something like that, so this is like the introduction to 3 scripts :face_with_spiral_eyes: .

Thanks for the comments :+1:

brbzull0:
[approve ci autest]

tools/build_boringssl_h3_tools.sh
```diff
@@ -0,0 +1,260 @@
+#!/usr/bin/env bash
+#
+#  Simple script to build OpenSSL and various tools with H3 and QUIC support
```
brbzull0, line 3: I'll fix this

tools/build_boringssl_h3_tools.sh
```diff
@@ -0,0 +1,271 @@
...
+  --disable-examples \
+  ${ENABLE_APP}
+${MAKE} -j ${num_threads}
+sudo ${MAKE} install
+sudo chmod -R a+rX ${BASE}
+cd ..
+
+# Then curl
+echo "Building curl ..."
+[ ! -d curl ] && git clone --depth 1 -b curl-8_7_0 https://github.com/curl/curl.git
```
bneradt, line 254: We're currently using `curl-8_7_1`. Might as well stick with that unless you have concerns with that version.

bneradt (commented):
Looks fine to me. Thank you for working on this @brbzull0 .

maskit (approved):
Looks good. Thanks.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11271, opened 2024-04-19
rate_limit: Adds a --rate option for RPS limits

zwoop:


plugins/experimental/rate_limit/limiter.cc
```diff
@@ -0,0 +1,82 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "limiter.h"
+
+BucketManager gBucketMngr;
```
ywkaras, line 21: Why not make `BucketManager` a namespace, and put its member variables into an anonymous namespace in this file?

plugins/experimental/rate_limit/limiter.h
```diff
@@ -24,16 +24,129 @@
...
+      return _count.load(std::memory_order_acquire);
+    }
+
+    bool
+    consume()
+    {
+      uint32_t val = _count.load(std::memory_order_acquire);
+
+      while (val > 0) {
+        if (_count.compare_exchange_weak(val, val - 1, std::memory_order_release, std::memory_order_acquire)) {
```
masaori335, line 68: It looks like the `val` can be changed by other thread. In such case, this `compare_exchange_weak` is failed and this loop keeps running. Is it bette to load the `_count` again in this loop?

Another approach is simple `fetch_sub(1, std::memory_order_relaxed)`? It doesn't need loop, I think.

masaori335: NVM, Chris pointed out when it's failed, the `val` is loaded again automatically.

randall (approved):

PR #11273, opened 2024-04-19
HTTP 418: "I'm a teapot" reason string

bneradt:
Add reason string "I'm a teapot" support for HTTP response status 418. ATS core will almost certainly not use this, but plugins may for testing purposes. It's handy for those circumstances to add the response reason support for the 418 status.

---
# For Review

Unfortunately, c macro [stringification](https://gcc.gnu.org/onlinedocs/gcc-4.1.2/cpp/Stringification.html) does not support the ``'`` character in strings. So the vast majority of this review is simply converting away from the reliance upon stringification to, instead, explicitly quoted reason phrases. The only new line added is the `418` status containing `I'm a teapot`.


bneradt:
[approve ci autest]

bneradt:
[approve ci autest]

masaori335:
There were discussions about `418 I'm a teapot` at httpbis. Now, RFC 9110 says `418` is "unused" as HTTP Status Code.

https://www.rfc-editor.org/rfc/rfc9110.html#section-15.5.19
https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml

Personally, I love the humor of `Hyper Text Coffee Pot Control Protocol`, but we don't need to add the reason string to ATS as of 2024, because it's not the `Hyper Text Transfer Protocol`.

maskit:
I'd say relying on reason strings is a bad custom even for testing because such a test is not runnable on H2 and H3. I don't think we should open wide the door for making tests that won't work for H2/3.

maskit:
I guess there's no problem on runtime (assuming the change on this PR has no impact to performance), but what's the motivation for having the reason string if we agree on that we shouldn't use it? We'd have to check whether future PRs don't use the reason string.

PR #11277, opened 2024-04-22
CMake. Fix yamlcpp version definition when an external yaml lib is set.

brbzull0:


cmcfarlen (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11279, opened 2024-04-22
Convert records to use Regex wrapper instead of pcre directly

bryancall:


JosiahWI:
AuTest "ramp" failed.

cmcfarlen (approved):
This is good cleanup.  I wonder though if some of the callers of this could instead just re-use a static Regex object rather than compile for each call.  See `recordIPCheck` for an example.  This `recordRegexCheck` function could have an override that takes a Regex or the caller could just call exec directly.

PR #11280, opened 2024-04-23
cleanup: Resolve some of dependencies on private headers (records)

maskit:
I could keep RecordsUtils.h and convert.h under include/ to avoid referring to  a parent directory like `#include "../common/RecordsUtils.h"`, but they are only used in src/mgmt/ after all. I'm not sure if we want to avoid the cross-submodule reference.

### DIFF
--- a/include/records/RecordsConfig.h
+++ b/include/records/RecordsConfig.h
@@ -23,6 +23,7 @@
 
 #pragma once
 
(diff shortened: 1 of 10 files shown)


bneradt (approved):

brbzull0 (approved):
Make sense. Thanks.

PR #11285, opened 2024-04-23
clang-format v18 + modified configs

zwoop:


include/tscore/ink_memory.h
```diff
@@ -103,9 +103,10 @@ int ats_mlock(caddr_t addr, size_t len);
 
 void *ats_track_malloc(size_t size, uint64_t *stat);
 void *ats_track_realloc(void *ptr, size_t size, uint64_t *alloc_stat, uint64_t *free_stat);
-void ats_track_free(void *ptr, uint64_t *stat);
+void  ats_track_free(void *ptr, uint64_t *stat);
 
-static inline size_t __attribute__((const)) ats_pagesize()
+static inline size_t __attribute__((const))
+ats_pagesize()
```
ywkaras, line 109: This seems pretyy rando.

include/mgmt/rpc/jsonrpc/JsonRPCManager.h
```diff
@@ -244,8 +244,8 @@ class JsonRPCManager
       InternalHandler(const RPCRegistryInfo *info, TSRPCHandlerOptions const &opt) : _regInfo(info), _options(opt) {}
       /// Sets the handler.
       template <class T, class F> void set_callback(F &&t);
-      explicit operator bool() const;
-      bool operator!() const;
+      explicit                         operator bool() const;
+      bool                             operator!() const;
```
brbzull0, line 248: methods too?  this seems odd to me. I  like this for members but for  methods is a bit weird.

src/tscore/unit_tests/test_Extendible.cc
```diff
@@ -94,8 +94,18 @@ class C : public B, public Extendible<C>
...
+  operator[](F field) const
+  {
+    return ext::get(*this, field);
+  }
+  template <typename F>
+  decltype(auto)
+  operator[](F field)
+  {
+    return ext::set(*this, field);
+  }
```
ywkaras, line 108: This  just seems moody.

src/iocore/net/UnixUDPNet.cc
```diff
@@ -504,33 +502,31 @@ UDPNetProcessorInternal::read_multiple_messages_from_net(UDPNetHandler *nh, UDPC
...
   sockaddr_in6 toaddr[MAX_RECEIVE_MSG_PER_CALL];
-  int toaddr_len = sizeof(toaddr);
+  int          toaddr_len = sizeof(toaddr);
 
   size_t total_bytes_read{0};
 
-  static const size_t cmsg_size
-  {
-    CMSG_SPACE(sizeof(int))
+  static const size_t cmsg_size{CMSG_SPACE(sizeof(int))
```
bryancall, line 518: This seems a little odd

include/tscore/ink_memory.h
zwoop: Yah, it seems clang-format v18 actually fixed this where the function ought to start where it does. So, this is correct, and the previous clang-format was wrong.

include/mgmt/rpc/jsonrpc/JsonRPCManager.h
zwoop: Yeh, I don't see adoption for this specifically :/

src/iocore/net/UnixUDPNet.cc
zwoop: Yeh, this looks like a bug probably ?

zwoop: Actually no, this looks like it actually fixed a bug in previous versions of clang-format, this is an initializer, and not a function:

```
  static const size_t cmsg_size{CMSG_SPACE(sizeof(int))
#ifdef IP_PKTINFO
                                + CMSG_SPACE(sizeof(struct in_pktinfo))
#endif
#if defined(IPV6_PKTINFO) || defined(IPV6_RECVPKTINFO)
                                + CMSG_SPACE(sizeof(struct in6_pktinfo))
#endif
#ifdef IP_RECVDSTADDR
                                + CMSG_SPACE(sizeof(struct in_addr))
#endif
#ifdef UDP_GRO
                                + CMSG_SPACE(sizeof(uint16_t))
#endif
  };
```

include/iocore/cache/AggregateWriteBuffer.h
```diff
@@ -88,19 +88,19 @@ class AggregateWriteBuffer
...
-  void reset_buffer_pos();
-  int get_bytes_pending_aggregation() const;
-  void add_bytes_pending_aggregation(int size);
+  char                                    *get_buffer();
+  int                                      get_buffer_pos() const;
+  void                                     add_buffer_pos(int size);
+  void                                     seek(int offset);
+  void                                     reset_buffer_pos();
+  int                                      get_bytes_pending_aggregation() const;
+  void                                     add_bytes_pending_aggregation(int size);
```
bneradt, line 97: Stuff like this can be helped with more concise type aliases.

bneradt (approved):

bneradt (approved):

PR #11286, opened 2024-04-24
txn_box autests: prefix each with 'txn_box'

bneradt:
I noticed the following warning in our CI:

```
Warning: overiding test tls with test in /var/tmp/trafficserver/tests/gold_tests/pluginTest/txn_box/basic
Warning: overiding test basic with test in /var/tmp/trafficserver/tests/gold_tests/pluginTest/txn_box/basic
Warning: overiding test redirect with test in /var/tmp/trafficserver/tests/gold_tests/pluginTest/txn_box/basic
```

This indicates that our new txn_box autests have naming conflicts with previously created ATS autests ("basic", "tls", "redirect", to be specific). This patch disambiguates these tests by prefixing each txn_box autest with "txn_box". In addition to resolving the naming conflicts, this also identifies the tests as txn_box for users seeing the tests run and allows regex matching on them.


brbzull0 (commented):
Thanks Brian.

maskit:
This change is fine, but it sounds like a design issue in AuTest...

I actually get errors when I run autest because of too long path lengths. I don't remember which ones had the issue, but I had to rename (shorten) the filenames to run them. If we have to have prefixes in filenames where we already have the "prefixes" in the directory names, it's so unfortunate.

maskit (approved):

bneradt:
Thanks for the review @maskit .

> This change is fine, but it sounds like a design issue in AuTest...

That's a good point. Maybe I can submit a patch so autest itself can keep the tests separate. At least it warns. :) 

> I actually get errors when I run autest because of too long path lengths.

I'm familiar with this. The pathing issue isn't actually an autest framework problem. Some tests create sockets and the OS limits the path length of those. I generally get around this by using `--sandbox /tmp/sb`. `/tmp/sb` is short enough that it avoids the path length issues.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11287, opened 2024-04-24
Update XPACK.cc to fix use-of-uninitialized-pointer-field case

shukitchan:
Fuzzing reveals a use-of-uninitialized-pointer-field case

https://oss-fuzz.com/testcase-detail/6285050312196096

I think this should fix the problem

### DIFF
--- a/src/proxy/hdrs/XPACK.cc
+++ b/src/proxy/hdrs/XPACK.cc
@@ -218,6 +218,7 @@ XpackDynamicTable::XpackDynamicTable(uint32_t size) : _maximum_size(size), _avai
 {
   XPACKDebug("Dynamic table size: %u", size);
   this->_entries      = static_cast<struct XpackDynamicTableEntry *>(ats_malloc(sizeof(struct XpackDynamicTableEntry) * size));
+  memset(this->_entries, 0, sizeof(struct XpackDynamicTableEntry) * size);
   this->_entries_head = size - 1;
   this->_entries_tail = size - 1;
 }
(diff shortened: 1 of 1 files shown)


maskit:
I'm not too sure if we should do this, the memset. Use of uninitialized field implies reading an entry that does not exist. I think we should check index numbers instead so such access does not happen. Touching the entire buffer would suppress the report, but it can hide real issues.

shukitchan:
Let me give it more thoughts

bneradt:
[approve ci autest]

shukitchan:
@maskit 
I double checked and i think it should be the fix we should do. 

If we start from the beginning and insert an oversized entry, _make_space() will always look at the tail entry. So memset to 0 will make we fail gracefully. That's the problem the fuzz test is having.

If we already have entries and now add an oversized entry, we need _make_space() to at least clear out all entries first before giving up. So we can't have a precondition check for an oversized entry. That is implied in the unit test - https://github.com/apache/trafficserver/blob/master/src/proxy/hdrs/unit_tests/test_XPACK.cc#L318

maskit:
Shouldn't we call `is_empty()` at the beginning of `_make_space()` and return `false` if the table is empty?

shukitchan:
> Shouldn't we call `is_empty()` at the beginning of `_make_space()` and return `false` if the table is empty?

I agree. I just made a change for that.

shukitchan:
[approve ci fedora]

shukitchan:
[approve ci autest]

maskit (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11296, opened 2024-04-26
Fedora 40: autest output updates for new curl and openssl

bneradt:
These patches update some autest output expectations for the newer version of curl and openssl s_client. Also, curl strangely used to accept `--resolv` instead of `--resolve`, but 8.7 no longer does. To address this, a test is updated to correctly pass `--resolve`.

After landing this patch, all autests should now work with fedora:40.


tests/gold_tests/tls/test-0rtt-s_client.py
```diff
@@ -69,7 +69,8 @@ def main():
         data = b''
         for line in lines:
             line += b'\n'
-            if line.startswith(bytes('SSL_connect:', 'utf-8')) or \
+            if line.startswith(bytes('Connecting to', 'utf-8')) or \
+                    line.startswith(bytes('SSL_connect:', 'utf-8')) or \
```
duke8253, line 73: Looks good, OpenSSL 3.2 has a new output we need to process.

duke8253 (approved):

moonchen (approved):

PR #11301, opened 2024-04-29
Update XPACK.cc - initialize the table after allocation to avoid bad access

shukitchan:
Let's determine if we need this after fixing #11287 and whether fuzzing complains on other code path


shukitchan:
[approve ci autest]

maskit:
Let's say the fuzzer doesn't find anything at the moment. How do we know index numbers are checked appropriately at every place, after we make changes?

Also, the zeros set by memset are only available until actual entries overwrite them.

I still don't get the benefit of doing this.

shukitchan:
I think the actual benefit is that it is a precaution. if you forget to initialize an entry before using it accidentally, it will be less harmful.

maskit:
> if you forget to initialize an entry before using it accidentally, it will be less harmful.

Until the buffer is overwritten. An argument would be that we should zero-clear an entry when we remove it as well. That would make sense, but I'm not sure if it worth. I think letting the fuzzer find the errors, and/or having more unit tests is safer.

shukitchan:
Let's not worry about this for now till the fuzzer reports another problem related to this part.

PR #11308, opened 2024-05-01
Fix case where 206 could be returned for range revalidates

cmcfarlen:
If a client asks for a range request for a cached object that when re-validated the server returns an error, then ATS would return a 206.


duke8253 (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11310, opened 2024-05-02
Print the hash of secret data

maskit:
Printing just the first 50 byte is not very helpful in some cases (e.g. certificate and the key are concatenated)

### DIFF
--- a/src/iocore/net/SSLSecret.cc
+++ b/src/iocore/net/SSLSecret.cc
@@ -26,6 +26,7 @@
 #include "P_SSLConfig.h"
 
 #include <utility>
+#include <openssl/evp.h>
 
 namespace
 {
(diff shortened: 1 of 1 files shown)


src/iocore/net/SSLSecret.cc
```diff
@@ -35,6 +36,20 @@ DbgCtl dbg_ctl_ssl_secret_err{"ssl_secret_err"};
 
 } // end anonymous namespace
 
+static void
+get_hash_str(const std::string &input, char hash_str[EVP_MAX_MD_SIZE * 2], unsigned int *hash_len)
+{
+  unsigned char hash[EVP_MAX_MD_SIZE];
```
bneradt, line 42: Maybe initialize the arrays with `= {0};` ?

bneradt (approved):
Looks good. Thanks.

src/iocore/net/SSLSecret.cc
maskit: Hmmm, I don't think that helps something.

cmcfarlen:
Cherry-picked to v10.0.x for #11464

PR #11317, opened 2024-05-03
Send EOS event when QUIC connection becomes draining state

maskit:


src/iocore/net/QUICNetVConnection.cc
```diff
@@ -502,10 +511,8 @@ QUICNetVConnection::is_handshake_completed() const
 void
 QUICNetVConnection::net_read_io(NetHandler *nh, EThread *lthread)
 {
-  if (quiche_conn_is_readable(this->_quiche_con)) {
-    SCOPED_MUTEX_LOCK(lock, this->mutex, this_ethread());
-    this->handleEvent(QUIC_EVENT_PACKET_READ_READY, nullptr);
-  }
+  SCOPED_MUTEX_LOCK(lock, this->mutex, this_ethread());
+  this->handleEvent(QUIC_EVENT_PACKET_READ_READY, nullptr);
```
maskit, line 515: Checking `quiche_conn_is_readable` was wrong because the function checks whether there are readable *streams*. There may be something need to be processed on the connection level.

src/iocore/net/QUICNetVConnection.cc
```diff
@@ -180,6 +180,10 @@ QUICNetVConnection::state_handshake(int event, Event *data)
     break;
   }
 
+  if (this->closed != 1 && (quiche_conn_is_closed(this->_quiche_con) || quiche_conn_is_draining(this->_quiche_con))) {
```
bneradt, line 183: Not directly related to this change, but should `closed` be a boolean? It seems like this would read better that way.

bneradt (approved):

src/iocore/net/QUICNetVConnection.cc
maskit: I wish it were that simple. There is a place that sets it to -1 in UnixNetVConnection.

bneradt: Ah, got it. Thanks.

brbzull0 (approved):
Looks good.

I have tested this, I can now run load test on h3 again. Thanks

cmcfarlen:
Cherry-picked to v10.0.x

PR #11319, opened 2024-05-03
Coverity 1518564: Out-of-bounds write in fq_pacing plugin

bryancall:
I believe this is a false-positive and `TSfread` shouldn't return a value larger than the buffer size.    However, we should just use `open, read, close` instead of our hacky buffered file reading in our APIs.  We are only reading 4 bytes anyways.


zwoop (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11320, opened 2024-05-03
Adds a run-plugin operator to HRW

zwoop:
An example use case is to conditionally enable / run the rate_limit.so plugin for some requests:

```
cond %{CLIENT-HEADER:X-Leif} ="yes"
     run-plugin rate_limit.so "--limit=300 --error=429"
```

### DIFF
--- a/doc/admin-guide/plugins/header_rewrite.en.rst
+++ b/doc/admin-guide/plugins/header_rewrite.en.rst
@@ -775,6 +775,14 @@ changing the remapped destination, ``<part>`` should be used to indicate the
 component that is being modified (see `URL Parts`_). Currently the only valid
 parts for rm-destination are QUERY, PATH, and PORT.
 
+run-plugin
+~~~~~~~~~~~~~~
+::
+
+  run-plugin <plugin-name>.so "<plugin-argument> ..."
+
+This allows to run an existing remap plugin, conditionally, from within a
+header rewrite rule.
 
 set-header
 ~~~~~~~~~~
@@ -1350,3 +1358,13 @@ the client where the requested data was served from.::
    cond %{HEADER:ATS-SRVR-UUID} ="" [OR]
    cond %{CACHE} ="hit-fresh"
    set-header ATS-SRVR-UUID %{ID:UNIQUE}
+
+Apply rate limiting for some select requests
+------------------------------------
+
+This rule will conditiionally, based on the client request headers, apply rate
+limiting to the request.::
+
+   cond %{REMAP_PSEUDO_HOOK} [AND]
+   cond %{CLIENT-HEADER:Some-Special-Header} ="yes"
+   run-plugin rate_limit.so "--limit=300 --error=429"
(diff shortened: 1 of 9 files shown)


doc/admin-guide/plugins/header_rewrite.en.rst
```diff
@@ -775,6 +775,14 @@ changing the remapped destination, ``<part>`` should be used to indicate the
...
 parts for rm-destination are QUERY, PATH, and PORT.
 
+run-plugin
+~~~~~~~~~~~~~~
+::
+
+  run-plugin <plugin-name>.so "<plugin-argument> ..."
+
+This allows to run an existing remap plugin, conditionally, from within a
+header rewrite rule.
```
ywkaras, line 785: Can this run on any hook other than the remap hook?

ywkaras:
Testing schmesting I guess.

shukitchan (dismissed):

shukitchan:
actually there is something wrong with the test.

but basically with the plugin factory, you can achieve this anywhere else, like Lua plugin or txn_box

ywkaras:
I'll be on vacation for 2 weeks after today, so I'll I have to leave any further review to others.  Looks OK from a quick read through.

doc/admin-guide/plugins/header_rewrite.en.rst
zwoop: No it can only run in a remap hook, but the plugin it runs can setup other hooks.

zwoop:
[approve ci]

zwoop:
> Testing schmesting I guess.

Yeh, well... I can't reproduce the build problem on either my Mac or  my Linux box, but seems like it may be that the header_rewrite_test target needs a score dependency?

zwoop:
[approve ci]

plugins/header_rewrite/operators.cc
```diff
@@ -1103,3 +1106,73 @@ OperatorSetHttpCntl::exec(const Resources &res) const
...
+  require_resources(RSRC_CLIENT_REQUEST_HEADERS); // Need this for the txnp
+}
+
+void
+OperatorRunPlugin::exec(const Resources &res) const
+{
+  TSReleaseAssert(_plugin != nullptr);
+
+  if (res._rri && res.txnp) {
+    _plugin->doRemap(res.txnp, res._rri);
```
shukitchan, line 1176: Just curious. should we pass on the return status of the doRemap() function somewhere?

zwoop: Yeh, so we kinda butchered this a bit in all the refactoring and dealings with "chaining" remap plugins. In most cases, this return code is obsoleted, and we don't' really use it. So, I made a choice to ignore it in HRW, whereas in Cripts, the caller can chose to look at it and do something with it (albeit, I think that's unlikely).

I'm not sure the stop / continue works any more either, but almost no plugin that's in use uses that.

shukitchan (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11330, opened 2024-05-08
UDP-Net: Add support to use SO_TXTIME if available.

brbzull0:
This change uses `SO_TXTIME` to set a hint to schedule transmission of specific packets at specific times, we use the hint given by the `quiche` library when `quiche_conn_send` is invoked.

No confuration needed, cmake will detect if available and use it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -407,6 +407,15 @@ if(TS_USE_HWLOC)
   )
 endif()
 
+check_symbol_exists(SO_TXTIME "sys/socket.h" SO_TXTIME_FOUND)
+set(CMAKE_EXTRA_INCLUDE_FILES "linux/net_tstamp.h")
+check_type_size("struct sock_txtime" STRUCT_SOCK_TXTIME_FOUND)
+unset(CMAKE_EXTRA_INCLUDE_FILES)
+
+if(SO_TXTIME_FOUND AND STRUCT_SOCK_TXTIME_FOUND)
+  set(HAVE_SO_TXTIME TRUE)
+endif()
+
 option(USE_IOURING "Use experimental io_uring (linux only)" 0)
 if(HAVE_IOURING AND USE_IOURING)
   message(STATUS "Using io_uring")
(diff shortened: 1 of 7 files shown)


brbzull0:
No configuration added, it just uses the insfrastructure if available... which kind of add some extra `#ifdef` around :vomiting_face:  . We can add a config flag if we really want it.

CMakeLists.txt
```diff
@@ -402,6 +402,15 @@ if(TS_USE_HWLOC)
   )
 endif()
 
+check_symbol_exists(SO_TXTIME "sys/socket.h" SO_TXTIME_FOUND)
+set(CMAKE_EXTRA_INCLUDE_FILES "linux/net_tstamp.h")
+check_type_size("struct sock_txtime" STRUCT_SOCK_TXTIME_FOUND)
+unset(CMAKE_EXTRA_INCLUDE_FILES)
+
+if(SO_TXTIME_FOUND AND STRUCT_SOCK_TXTIME_FOUND)
```
maskit, line 410: Are there platforms that don't have either of these? Checking the both is fine but I'm just curious.

brbzull0: I did check on FreeBSD, `net_tstamp.h` wasn't  available, neither was `SO_TXTIME` .  Are they available in Mac?

brbzull0:
[approve ci debian]

CMakeLists.txt
maskit: No, macOS doesn't have neither of the two. I agree that checking the both is more robust. I just wondered what are used if there's a platform that has only one of the two.

bneradt:
[approve ci autest]

maskit (approved):
Looks good. Thanks.

Were you able to see any performance difference?

brbzull0:
> Looks good. Thanks.
> 
> Were you able to see any performance difference?

Not really, but I am using an Intel NUC with 8 cores, it didn't get worse though, I see very little difference between runs but   not worth mentioning it. Need to try this on a better lab machine, I'll post some results when I try it.

Thanks.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11332, opened 2024-05-08
Enable ASan on CI Rocky preset

masaori335:


masaori335:
[approve ci]

masaori335:
Following @bneradt's advice, enabling ASan on rocky instead of adding another preset.

masaori335:
gcc-toolset-11 doesn't like UBSan :(
```
/opt/rh/gcc-toolset-11/root/usr/bin/ld: cannot find -lubsan
```

CMakePresets.json
```diff
@@ -151,7 +151,9 @@
         "quiche_ROOT": "/opt/h3-tools-boringssl/quiche",
         "CMAKE_INSTALL_PREFIX": "/tmp/ats-quiche",
         "opentelemetry_ROOT": "/opt",
-        "ENABLE_QUICHE": "ON"
+        "ENABLE_QUICHE": "ON",
+        "CMAKE_CXX_FLAGS_DEBUG": "-g -fsanitize=address",
+        "CMAKE_C_FLAGS_DEBUG": "-g -fsanitize=address"
```
bneradt, line 156: We should be able to use `"ENABLE_ASAN": "ON",` instead of setting the flags. That would be preferable. Can you also use `ENABLE_ASAN` for the other places in here that set those `-fsanitize=address` flags?

bneradt (changes requested):
Thanks for setting on `ci-rocky`. It will be good to get this coverage in PR CI.

CMakePresets.json
masaori335: Sure. For the other places, I'll open another PR.

masaori335:
[approve ci rocky]

bneradt (approved):

PR #11336, opened 2024-05-09
Coverity 1534725: Uninitialized scalar variable

serrislew:
Coverity complaining `col._row_size` not initialized but that's only when `col._type` hits default case (which comment implies shouldn't happen). Can also add `col._row_size = 0` to just default case


ezelkow1 (approved):

bneradt (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11346, opened 2024-05-13
Handle EOS events during server open

duke8253:
We saw a bunch of aborts happening due to EOS not being handled during `HttpSM::state_http_server_open`, and after some discussion we believe this should be handled the same as a `VC_EVENT_ERROR` case. This is now deployed globally for us.

### DIFF
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -1088,6 +1088,7 @@ HttpSM::state_raw_http_server_open(int event, void *data)
     break;
   }
   case VC_EVENT_ERROR:
+  case VC_EVENT_EOS:
   case NET_EVENT_OPEN_FAILED:
     t_state.current.state = HttpTransact::OPEN_RAW_ERROR;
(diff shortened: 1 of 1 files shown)


cmcfarlen:
Other places with `VC_EVENT_ERROR` but not `VC_EVENT_EOS`

```
src/iocore/net/QUICNextProtocolAccept.cc|44 col 8| case VC_EVENT_ERROR:
src/proxy/http/HttpSM.cc|1232 col 8| case VC_EVENT_ERROR:
src/proxy/http/HttpSM.cc|1808 col 8| case VC_EVENT_ERROR:
src/proxy/Transform.cc|634 col 10| case VC_EVENT_ERROR:
src/proxy/Transform.cc|793 col 10| case VC_EVENT_ERROR:
```

cmcfarlen (approved):
I do wonder if these other places should handle EOS, but this PR looks fine.

cmcfarlen:
Cherry-picked to v10.0.x

ezelkow1:

PR #11347, opened 2024-05-13
Add tests for writing documents to the cache

JosiahWI:
This tests that the correct document data is written to cache by aggWrite. It also covers some fields of the Doc struct used for reading the data: the magic value, the length and header length, and the optional checksum.

This commit also fixes the use of an uninitialized variable (header.phase) reported by valgrind by initializing it to the same default value used in the Stripe::_clear_init.


masaori335 (approved):
Looks good

PR #11348, opened 2024-05-13
Restore private key passphrase behavior

moonchen:
We accidentally removed the callbacks for private key passphrases. This patch restores the old behavior.  Also add autest.


maskit:
#11201

PR #11352, opened 2024-05-14
Allow query parameters to erase() 'all but'

zwoop:
The optional boolean flag argument reverses the list semantics (erase all but the listed query parameters). E.g.

```
do_remap()
{
  borrow url = Client::URL::get();

  url.query.erase({"foo", "bar"}, true);
}
```

Which will remove all query parameters, except **foo** and  **bar**. Alternatively, as a convenience, keep() was also added:

```
do_remap()
{
  borrow url = Client::URL::get();

  url.query.keep({"foo", "bar"});
}
```

### DIFF
--- a/include/cripts/Urls.hpp
+++ b/include/cripts/Urls.hpp
@@ -426,20 +426,19 @@ public:
     Cript::string      operator+=(Cript::string_view add);
     Parameter          operator[](Cript::string_view param);
     void               erase(Cript::string_view param);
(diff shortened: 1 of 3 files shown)


brbzull0:
[approve ci autest]

bneradt (approved):

cmcfarlen:
Cherry-picked to v10.0.x

Issue #11343, opened 2024-05-13
Proposition for removing the global suppression `-Wno-stringop-overflow` from the build


freak82:
Hi there,

As far as I checked the only place where this warning is raised is here:
```
/z/x3me-ocn/src/iocore/net/OCSPStapling.cc:1166:12: warning: ‘char* strncat(char*, const char*, size_t)’ specified bound 1 equals source length [-Wstringop-overflow=]
 1166 |     strncat(url->end(), "/", 1);
``` 
Here is the code in question:
```
  if (url->buf()[url->size() - 1] != '/') {                                     
    strncat(url->end(), "/", 1);                                                
    url->fill(1);                                                                   
  } 
```  
Note that the `url` buffer size is calculated above in such a way that there is a guarantee that it's big enough for this concatenation. 

[This type of GCC warning](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=83404) is a bit bogus.
However, I think with a slight change the code here can be made better/safer and at the same time the global suppression of this warning can be removed.
Although it's not very obvious (IMO) from the man page of `strncat`, the last `size_t` argument is supposed to specify [the remaining space in the destination buffer](https://stackoverflow.com/questions/6903997/how-can-i-use-strncat-without-buffer-overflow-concerns).
So, I think this code is better written as:
```
  if (url->buf()[url->size() - 1] != '/') {                                     
    strncat(url->end(), "/", url->write_avail());                                                
    url->fill(1);                                                                   
  } 
``` 
or
```
 if (url->buf()[url->size() - 1] != '/') {                                     
      written = ink_strlcat(url->end(), "/", url->write_avail());          
      url->fill(written);                                                                
  } 
``` 

What do you think?

maskit:
The proposed code looks good, and I'm +1 on removing the suppression.  I'd go with `ink_strlcat` because use of strncat isn't common in ATS codebase. Would you make a Pull Request?

freak82:
I'm OK to do a pull request but I'm not sure against which branch to do it?

maskit:
A Pull Request against `master` would be great. We'll take care of other branches.

PR #11355, opened 2024-05-15
Remove the `-Wno-stringop-overflow` warning suppression

freak82:
The  `-Wstringop-overflow` warning was caused by single usage of `strncat` function.
The problem was that GCC complained for possibly incorrect usage of `strncat` due to the last argument being equal to the source string length. The last argument of strncat is supposed to be the remaining free space in the destination buffer.
Replaced the usage of `strncat` with the ATS function `ink_strlcat`.
More info about this can be found in [this ATS issue](https://github.com/apache/trafficserver/issues/11343)

### DIFF
--- a/src/iocore/net/UnixUDPNet.cc
+++ b/src/iocore/net/UnixUDPNet.cc
@@ -97,7 +97,7 @@ UDPPacket::new_UDPPacket(struct sockaddr const *to, ink_hrtime when, Ptr<IOBuffe
   p->p.segment_size = segment_size;
 #ifdef HAVE_SO_TXTIME
   if (send_at_hint) {
-    memcpy(&p->p.send_at, &send_at_hint, sizeof(struct timespec));
+    memcpy(&p->p.send_at, send_at_hint, sizeof(struct timespec));
   }
 #endif
   return p;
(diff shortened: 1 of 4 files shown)


freak82:
I apologize for including the fix about the `memcpy` in `src/iocore/net/UnixUDPNet.cc` in this pull request.
I intended to do separate pull request for this change but ended up pushing the change in the same branch and thus it got into this pull request.
If needed I will close this pull request and do 2 separate ones.

src/iocore/net/UnixUDPNet.cc
```diff
@@ -97,7 +97,7 @@ UDPPacket::new_UDPPacket(struct sockaddr const *to, ink_hrtime when, Ptr<IOBuffe
   p->p.segment_size = segment_size;
 #ifdef HAVE_SO_TXTIME
   if (send_at_hint) {
-    memcpy(&p->p.send_at, &send_at_hint, sizeof(struct timespec));
+    memcpy(&p->p.send_at, send_at_hint, sizeof(struct timespec));
```
brbzull0, line 100: Ohh!! this a very good catch.
Must have done some refactor before pushing this and let this slide!! 

Please split this into a different PR just in case.

brbzull0 (changes requested):

src/iocore/net/UnixUDPNet.cc
freak82: Ok. Will close this PR and make new ones.

PR #11357, opened 2024-05-15
Remove the need of `-Wno-stringop-overflow` warning suppression

freak82:
The problem was that GCC complained for possibly incorrect usage of `strncat` due to the last argument being equal to the source string length. The last argument of `strncat` is supposed to be the remaining free space in the destination buffer.
Replaced the usage of `strncat` with the ATS function `ink_strlcat`.
More info about this can be found in [this issue](https://github.com/apache/trafficserver/issues/11343)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -600,7 +600,7 @@ add_compile_options(-pipe -Wall -Wextra -Wno-unused-parameter)
 add_compile_options(
   "$<$<COMPILE_LANGUAGE:CXX>:-Wno-noexcept-type;-Wsuggest-override;-Wno-vla-extension;-fno-strict-aliasing>"
 )
-add_compile_options("$<$<CXX_COMPILER_ID:GNU>:-Wno-format-truncation;-Wno-stringop-overflow>")
+add_compile_options("$<$<CXX_COMPILER_ID:GNU>:-Wno-format-truncation>")
 
 if(NOT EXTERNAL_YAML_CPP)
   include(subproject_version)
--- a/lib/swoc/unit_tests/CMakeLists.txt
+++ b/lib/swoc/unit_tests/CMakeLists.txt
@@ -39,7 +39,6 @@ if(CMAKE_COMPILER_IS_GNUCXX)
             -Werror
             -Wno-unused-parameter
             -Wno-format-truncation
-            -Wno-stringop-overflow
             -Wno-invalid-offsetof
   )
   # stop the compiler from complaining about unused variable in structured binding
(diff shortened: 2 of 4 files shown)


ezelkow1:
[approve ci]

JosiahWI:
```
In function 'size_t str_concat(char*, size_t, const char*, size_t)',
    inlined from 'TSHttpStatus S3Request::authorizeV2(S3Config*)' at ../../../plugins/s3_auth/s3_auth.cc:920:25:
../../../plugins/s3_auth/s3_auth.cc:761:12: error: 'char* strncat(char*, const char*, size_t)' specified bound 1 equals source length [-Werror=stringop-overflow=]
  761 |     strncat(dst, src, to_copy);
      |     ~~~~~~~^~~~~~~~~~~~~~~~~~~
In function 'size_t str_concat(char*, size_t, const char*, size_t)',
    inlined from 'TSHttpStatus S3Request::authorizeV2(S3Config*)' at ../../../plugins/s3_auth/s3_auth.cc:926:25:
../../../plugins/s3_auth/s3_auth.cc:761:12: error: 'char* strncat(char*, const char*, size_t)' specified bound 1 equals source length [-Werror=stringop-overflow=]
  761 |     strncat(dst, src, to_copy);
      |     ~~~~~~~^~~~~~~~~~~~~~~~~~~
      ```
      
      It looks like there is a possible similar incorrect usage of `strncat` in the s3_auth plugin that the CentOS build is now complaining about.

**Comment by freak82** [comment]
Hmm, I built all of the plugins and there was no other complains from GCC 13.2 on Debian.
I can see that the `s3_auth` plugin has been built. It has not been skipped.
Most likely this is due to different GCC versions.

It seems to me that in the `plugins/s3_auth/s3_auth.cc ` the GCC analyzer is wrong about this warning:
It most likely get confused from this usage `str_concat(&left[loff], (left_size - loff), "/", 1);` and because in the `str_concat` there is `std::min(dst_len, src_len)` for the `strncat` size argument.

Maybe it'll be better if I change all of the remaining 5 usages of `strncat` in
```
plugins/s3_auth/s3_auth.cc (1)
plugins/experimental/url_sig/url_sig.cc (4)
```
with `ink_strlcat`?

Or the suppression can be left as it is?

maskit:
I don't mind leaving the warning suppression. Sounds like the change is enough for you, and I'm not going to have you fix all the places. We can remove the warning suppression later when we are ready to do so. But if you are willing to make additional changes for the plugin code, please use [TSstrlcat](https://docs.trafficserver.apache.org/developer-guide/api/functions/TSmalloc.en.html#c.TSstrlcat), which is exactly the same as ink_strlcat but it's exported for plugins.

freak82:
I replaced the `strncat` usage in the `s3_auth` plugin with `TSstrlcat`.
Tested the build again with both GCC 11, which was giving the warning in the plugin, and with GCC 13.

maskit (approved):
Thank you!

cmcfarlen:
Cherry-picked to v10.0.x

Issue #5165, opened 2019-03-14
Use SO_INCOMING_CPU with bound ports


zwoop:
When we migrate to use SO_REUSEPORT and bind the listening ports on all ET_NET threads, we should use SO_INCOMING_CPU too. It should be optional and perhaps somehow associated with the feature we already have of CPU affinity for threads.

bryancall:
SO_INCOMING_CPU support was added in #11366 (first released in the 10.0.0 release; in the current 10.1.2 release). It's an optional socket option (`net.sock_option_flag_in`) that binds to the listening thread's CPU affinity, alongside the per-thread SO_REUSEPORT path, exactly as this issue requested. Closing as done.

PR #11366, opened 2024-05-20
Add SO_INCOMING_CPU support to net.sock_option_flag_in

masaori335:
Address #5165.

Add a option (128) for `SO_INCOMING_CPU` to the `proxy.config.net.sock_option_flag_in`.

# Background

There was a bug with SO_REUSEPORT since v4.6. But it's fixed by v6.2 and some distros backported it.
Now we can use SO_INCOMING_CPU, SO_REUSEPORT and CPU Affinity 4 for perfect locality.

https://github.com/torvalds/linux/commit/b261eda84ec136240a9ca753389853a3a1bccca2

# Ref.

        SO_INCOMING_CPU (gettable since Linux 3.19, settable since Linux 4.4)

              Sets or gets the CPU affinity of a socket.  Expects an
              integer flag.

                  int cpu = 1;
                  setsockopt(fd, SOL_SOCKET, SO_INCOMING_CPU, &cpu,
                             sizeof(cpu));

              Because all of the packets for a single stream (i.e., all
              packets for the same 4-tuple) arrive on the single RX
              queue that is associated with a particular CPU, the
              typical use case is to employ one listening process per RX
              queue, with the incoming flow being handled by a listener
              on the same CPU that is handling the RX queue.  This
              provides optimal NUMA behavior and keeps CPU caches hot.
              
https://www.man7.org/linux/man-pages/man7/socket.7.html

### DIFF
--- a/src/iocore/net/Connection.cc
+++ b/src/iocore/net/Connection.cc
@@ -27,10 +27,12 @@
   Commonality across all platforms -- move out as required.
 
 **************************************************************************/
-#include "tscore/ink_platform.h"
-#include "tscore/ink_defs.h"
 
 #include "P_Net.h"
+
+#include "tscore/ink_defs.h"
+#include "tscore/ink_platform.h"
+#include "tscore/ink_sock.h"
 
 #ifdef SO_ACCEPTFILTER
 #include <sys/param.h>
@@ -49,6 +51,7 @@ DbgCtl dbg_ctl_http_tproxy{"http_tproxy"};
 DbgCtl dbg_ctl_proxyprotocol{"proxyprotocol"};
 DbgCtl dbg_ctl_iocore_net_server{"iocore_net_server"};
 DbgCtl dbg_ctl_connection{"connection"};
+DbgCtl dbg_ctl_iocore_thread{"iocore_thread"};
 
 } // end anonymous namespace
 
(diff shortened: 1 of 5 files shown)


src/iocore/net/Connection.cc
```diff
@@ -238,6 +241,23 @@ Server::setup_fd_for_listen(bool non_blocking, const NetProcessor::AcceptOptions
 #endif
   }
 
+#ifdef SO_INCOMING_CPU
+  if (opt.sockopt_flags & NetVCOptions::SOCK_OPT_INCOMING_CPU) {
+    int      cpu     = 0;
+    EThread *ethread = this_ethread();
+
+#ifdef TS_USE_HWLOC
+    cpu = ethread->hwloc_obj->os_index;
```
masaori335, line 250: The `hwloc_obj::os_index` looks "OS-provided physical index number", but hwloc has any better API?
https://github.com/open-mpi/hwloc/blob/63a8288d31a1baf67a909466aba9a022c78ca7b1/include/hwloc.h#L491

masaori335:
[approve ci freebsd]
[approve ci debian]

masaori335:
[approve ci freebsd]

masaori335:
[approve ci debian]

masaori335:
[approve ci autest]

masaori335:
[approve ci debian]

bryancall (approved):

masaori335:
@cmcfarlen I'd request to 10.0.x backport. But, this is a new feature, so we can wait for 10.1.x, if you want.

cmcfarlen:
Cherry-picked to v10.0.x

Issue #11358, opened 2024-05-15
Event handles with VC_EVENT_ERROR but not VC_EVENT_EOS


cmcfarlen:
While reviewing #11346, I went through and found other places where the EOS event was not handled but ERROR was:

```
src/iocore/net/QUICNextProtocolAccept.cc|44 col 8| case VC_EVENT_ERROR:
src/proxy/http/HttpSM.cc|1232 col 8| case VC_EVENT_ERROR:
src/proxy/http/HttpSM.cc|1808 col 8| case VC_EVENT_ERROR:
src/proxy/Transform.cc|634 col 10| case VC_EVENT_ERROR:
src/proxy/Transform.cc|793 col 10| case VC_EVENT_ERROR:
```

I wonder, should we add EOS to these switch statements as well?

maskit:
I'm not sure whether QUICNextProtocolAccept.cc receives EOS, but we can add it.

Is the list complete? I wonder why SSLNextProtocolAccept.cc isn't on the list.

cmcfarlen:
Its likely not complete, just me staring at the code

PR #11368, opened 2024-05-20
Handle VC_EVENT_EOS in more places

bneradt:
This adds the handling of VC_EVENT_EOS to a few handlers that did handle VC_EVENT_ERROR but not EOS. This is a follow up from a review comment in #11346. See:

https://github.com/apache/trafficserver/pull/11346#issuecomment-2112949224

### DIFF
--- a/src/iocore/net/QUICNextProtocolAccept.cc
+++ b/src/iocore/net/QUICNextProtocolAccept.cc
@@ -41,6 +41,7 @@ quic_netvc_cast(int event, void *edata)
     return dynamic_cast<QUICNetVConnection *>(ptr.vc);
   case VC_EVENT_INACTIVITY_TIMEOUT:
   case VC_EVENT_READ_COMPLETE:
+  case VC_EVENT_EOS:
   case VC_EVENT_ERROR:
     ptr.vio = static_cast<VIO *>(edata);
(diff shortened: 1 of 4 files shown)


bneradt:
I'm having second thoughts about the Tranform.cc - it already has extensive default handling which actually checks for `m_closed` which I think is generally set on EOS. For now I'll revert that one since I'm not sure about it.

bneradt:
Adding @maskit for review of the QUIC change.

maskit:
I commented this on #11358, but shouldn't `ssl_netvc_cast` in SSLNextProtocolAccept.cc have it as well if we add it to `quic_netvc_cast`?

bneradt:
> I'm not sure whether QUICNextProtocolAccept.cc receives EOS, but we can add it.
> 
> Is the list complete? I wonder why SSLNextProtocolAccept.cc isn't on the list.

Ah, I had not seen that comment. Good catch.

cmcfarlen:
Cherry-picked to v10.0.x

ezelkow1:
Do we still need this on 9.2.x?

bneradt:
> Do we still need this on 9.2.x?

Adding these events to the error handlers should be pretty safe and it can prevent crashes in some scenarios. I'd say we backport it. That's why I marked it for 9.2.x. As 9.2.x release manager, feel free to override me. :)

Issue #11375, opened 2024-05-21
Is this missed to forward argument in `Http2CommonSession::add_url_to_pushed_table`?


freak82:
Hi there,

The `url_len` argument of `Http2CommonSession::add_url_to_pushed_table` [function](https://github.com/apache/trafficserver/blob/d31a40fecb19f80ae8b7ddbb2ff0efecc57bf7f9/src/proxy/http2/Http2CommonSession.cc#L459) is not passed to the `emplace` call. 
```
void
Http2CommonSession::add_url_to_pushed_table(const char *url, int url_len)
{
  // Delay std::unordered_set allocation until when it used
  if (_h2_pushed_urls == nullptr) {
    this->_h2_pushed_urls = new std::unordered_set<std::string>();
    this->_h2_pushed_urls->reserve(Http2::push_diary_size);
  }

  if (_h2_pushed_urls->size() < Http2::push_diary_size) {
    _h2_pushed_urls->emplace(url);
  }
}
```
This means that it'll always search for the '\0' terminator when emplacing the string in `_h2_pushed_urls`.
After checking the usage of `add_url_to_pushed_table`, it seems to be that the `url_len` should also be forwarded to the `emplace`. 
Is this correct observation or not?

maskit:
Yes, it looks like a bug.

PR #11382, opened 2024-05-22
Pass url_len when emplacing in Http2CommonSession::_h2_pushed_urls

freak82:
The signature of the function in question is
```
void
Http2CommonSession::add_url_to_pushed_table(const char *url, int url_len)
```
The given `url` is used when emplacing/inserting to `_h2_pushed_urls` collection but the given `url_len` is not used and this may change the actually inserted URL because it'll search for the `\0` terminator instead of inserting only `url_len` characters/bytes.
And the `\0` terminator may not even be present in some cases (the `TSHttpTxnServerPush` calls the above function i.e. it may be called by different plugins).
I've opened [this issue](https://github.com/apache/trafficserver/issues/11375) about this and it resulted in this pull request.
Note that the problem with `url_len` is reported by the compiler if/when the warning for unused parameters is enabled.

### DIFF
--- a/src/proxy/http2/Http2CommonSession.cc
+++ b/src/proxy/http2/Http2CommonSession.cc
@@ -466,7 +466,7 @@ Http2CommonSession::add_url_to_pushed_table(const char *url, int url_len)
   }
 
   if (_h2_pushed_urls->size() < Http2::push_diary_size) {
-    _h2_pushed_urls->emplace(url);
+    _h2_pushed_urls->emplace(url, url_len);
   }
 }
 


ywkaras:
Good find!  It looks like `add_url_to_pushed_table` is only used in `TSHttpTxnServerPush` (in `src/api/InkAPI.cc`).  As a part of this PR, I suggest you add `sdk_assert(url && url_len > 0)` in `TSHttpTxnServerPush`.  Some TS API functions that accept strings with lengths work like this:  https://github.com/apache/trafficserver/blob/4f74b0222b5f6b77cf7a39a56bc8807595c35bc9/src/api/InkAPI.cc#L1292  So, someone might assume they can pass `-1` as the length.

freak82:
My bad. I'm not very accustomed to this pattern yet.
Do you think I should add the `sdk_assert` in `TSHttpTxnServerPush` or I should add the check with the `if (len < 0) len = strlen(...)`?
I'm asking because the `TSHttpTxnServerPush` function calls
```
url_obj.parse(url, url_len);
```
and the `url_obj` also does this check internally i.e. it's prepared for `-1` length and thus maybe the it's expected that `TSHttpTxnServerPush` can also take `-1` length and should be able to handle it instead of asserting.

ywkaras:
I'm not sure what the consensus is about this.  I asked in the slack channel:  https://the-asf.slack.com/archives/CHQ1FJ9EG/p1716395850701069

freak82:
Updated the pull request.

Overall the situation with this "maybe-negative-length" pattern is a bit confusing (at least for me).
I mean, if there this usage:
```
void some_exported_API(const char* str, int len)
{
  internal_fun_1(str, len);
  
  internal_fun_2(str, len);
  
  internal_fun_3(str, len);
}
```
Is the check of `if (len < 0) len = strlen(str)` supposed to happen at the highest level i.e. in the `some_exported_API` or every internal function is supposed to have this check (but then the if-check or the call to `strlen` will be potentially repeated multiple times).
Or the internal functions are just supposed to assert the `len >= 0`? But then why just don't get it as `size_t`.
Anyhow, the these are just my thoughts. It's what it's.

ywkaras (approved):

PR #11386, opened 2024-05-22
Coverity 1545947: Resource leak in header_rewrite plugin

bryancall:


plugins/header_rewrite/header_rewrite.cc
```diff
@@ -199,23 +199,22 @@ RulesConfig::parse_config(const std::string &fname, TSHttpHookID default_hook, c
     }
 
     // If we are at the beginning of a new condition, save away the previous rule (but only if it has operators).
-    if (p.is_cond() && add_rule(rule)) {
-      rule = nullptr;
+    if (p.is_cond() && add_rule(rule.get())) {
+      rule.release();
```
maskit, line 203: It's unfortunate that we still have to call `release()`. I wondered if we can `std::move` the `rule` and call `release()` in `add_rule` but it looks like the following code relies on whether `rule` becomes nullptr.

maskit (approved):

bryancall:
[approve ci centos]

plugins/header_rewrite/header_rewrite.cc
bryancall: I wanted to check to return value of `add_rule()` before I released the pointer.   `add_rule()` checks to see if the `rule` is not a nullptr and has an operator before assigning the pointer and returning true..

cmcfarlen:
Cherry-picked to v10.0.x

PR #11392, opened 2024-05-24
Add ts::hex and ts::unhex

cmcfarlen:


zwoop:
Ship it.

zwoop:
For this use case at least, this is at least as fast as Boost, sometimes even faster.

zwoop (approved):

include/tsutil/StringConvert.h
```diff
@@ -0,0 +1,71 @@
...
+#include <charconv>
+#include <stdexcept>
+#include <string>
+#include <system_error>
+
+namespace ts
+{
+
+inline std::string
+hex(const std::string_view input)
```
bneradt, line 30: I was initially confused, assuming `hex` converted from base 10 or some such to hex. Maybe comment with something like:

```cpp
/** Convert values into their hex ascii equivalents.
 * @note This is a reimplementation of boost's hex.
 * @return The series of hex values corresponding to @a input
 */
```

include/tsutil/StringConvert.h
```diff
@@ -0,0 +1,71 @@
...
+    } else {
+      throw std::runtime_error(std::make_error_code(err).message().c_str());
+    }
+  }
+
+  return result;
+}
+
+inline std::string
+unhex(const std::string_view input)
```
bneradt, line 49: Maybe doc with:

```cpp
/** The inverse function of hex above.
 * @note This is a reimplementation of boost's unhex.
 */

bneradt (changes requested):

bneradt:
Looks good, other than the request for documentation. I verified in godbolt that hex this seems to act like boost's.

bneradt (approved):

cmcfarlen:
Cherry-picked to v10.0.x

Issue #11377, opened 2024-05-21
Is there interest to get rid of `-Wno-unused-parameter` warning suppression?


freak82:
Hi there,

So, I did the above experiment for myself and there are lots of places which need to be fixed.
However, the warnings also revealed few places with potential issues/bugs and I've opened issues here about them. You can find them by my user, I guess.
Everything else seems to be purely mechanical work and I can do it if there is a desire for this.
I just believe that warning suppression is bad because the warnings sometimes point to real issues/bugs and we should use any help from the compiler and the other tools to catch bugs as early as possible (preferably at compile time, instead at runtime).
If there is a desire for this it needs to be decided how you want the changes.
I mean that in the project currently there are 5 types of suppressing this warning:
1. `ATS_UNUSED` after the parameter name
Drawbacks of this solution:
- `tscore/ink_defs.h` needs to be included in order of the above macro to be visible
- it's easy to forget to remove this tag if the parameter becomes used in the future and this potentially may confuse some readers of the code

2. `/* <parameter-name> ATS_UNUSED */` or just `/* <parameter-name> */`
Drawbacks of this solution:
- the solution which includes `ATS_UNUSED` in the commented part become a bit longer and too much (IMO) but is probably better for the readers? 

3. `[[maybe_unused]]`
Drawbacks of this solution:
- it's easy to forget to remove this attribute if the parameter becomes used in the future and this potentially may confuse some readers of the code

4. `(void)<parameter-name>` at the beginning of the function body
Drawbacks of this solution:
- it's easy to forget to remove this suppression if the parameter becomes used in the future and the function is longer and this potentially may confuse some readers of the code
- the unused parameters are not visible from the function signature
- this way of changing will require more manual work compared to some of the above solutions.

5. Just removing the name of the parameter
Drawbacks of this solution:
- it becomes unclear in some cases what this parameter is supposed to be

Note that all of the above are currently used in the ATS code as far as I checked. The 5-th way is probably the least used but there are few places still.
I'd personally go with 1, 2 or 3 and I'm leaning towards 2.

maskit:
> I just believe that warning suppression is bad because the warnings sometimes point to real issues/bugs and we should use any help from the compiler and the other tools to catch bugs as early as possible (preferably at compile time, instead at runtime).

I completely agree. And I feel like many of the inline suppressions in the ATS codebase are unnecessary (could be avoided), although some places may really need it.

[...]

PR #11394, opened 2024-05-25
Fix unused parameters in `tsutil`

freak82:
I got the liberty of slightly modifying the code in `tsutil/unit_tests/test_PostScript.cc`, instead of using `[[maybe_unused]]`, just to be consistent with the other test functions.

These changes are related to the discussion [here](https://github.com/apache/trafficserver/issues/11377).

fixes https://github.com/apache/trafficserver/issues/11377

### DIFF
--- a/include/tsutil/ts_unit_parser.h
+++ b/include/tsutil/ts_unit_parser.h
@@ -94,7 +94,7 @@ inline UnitParser::UnitParser(UnitParser::Units &&units, bool unit_required_p) n
 }
 
 inline UnitParser::self_type &
-UnitParser::unit_required(bool flag)
+UnitParser::unit_required([[maybe_unused]] bool flag)
 {
   _unit_required_p = false;
   return *this;
--- a/src/tsutil/DbgCtl.cc
+++ b/src/tsutil/DbgCtl.cc
@@ -338,7 +338,7 @@ struct DiagTimestamp {
 };
 
 swoc::BufferWriter &
-bwformat(swoc::BufferWriter &w, swoc::bwf::Spec const &spec, DiagTimestamp const &ts)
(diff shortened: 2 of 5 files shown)


freak82:
I can't figure out how the test failure is related to my changes.
I mean, it gives error in: `Run: Test proxy.config.http2.write_size_threshold: Failed`.
There in the `http2_write_threshold.test.py` script there is:
```
 24 class TestGrpc():                                                               
 25     """Test proxy.config.http2.write_size_threshold and its associated timeout."""
 26                                                                                 
 27     def __init__(self, description: str, write_threshold: int, write_timeout: int) -> None:
``` 
but at the end of the file it's called like this with a floating point value instead of an integer for the `write_threshold`:
```
test = TestGrpc("Test proxy.config.http2.write_size_threshold", 0.5, 10)
```
Is it possible that the test fails due to this types mismatch?

brbzull0:
[approve ci autest]

brbzull0:
> I can't figure out how the test failure is related to my changes. I mean, it gives error in: `Run: Test proxy.config.http2.write_size_threshold: Failed`. There in the `http2_write_threshold.test.py` script there is:
> 
> ```
>  24 class TestGrpc():                                                               
>  25     """Test proxy.config.http2.write_size_threshold and its associated timeout."""
>  26                                                                                 
>  27     def __init__(self, description: str, write_threshold: int, write_timeout: int) -> None:
> ```
> 
> but at the end of the file it's called like this with a floating point value instead of an integer for the `write_threshold`:
> 
> ```
> test = TestGrpc("Test proxy.config.http2.write_size_threshold", 0.5, 10)
> ```
> 
> And this value is later set in the trafficserver config `proxy.config.http2.write_size_threshold`.
> 
> Is it possible that the test fails due to the types mismatch here?

transient error.

cmcfarlen:
Cherry-picked to v10.0.x

Issue #11388, opened 2024-05-22
AddressSanitizer: heap-buffer-overflow on address in XpackDynamicTableStorage::write


bneradt:
Running ATS 10 (https://github.com/apache/trafficserver/commit/399d04ceace14fae66dc615d96ff93557b55ad00 specifically) in production under ASan, we see the following ASan buffer overflow error:

[asan_overflow.txt](https://github.com/apache/trafficserver/files/15408600/asan_overflow.txt)

```
=================================================================
==3888081==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x625008bc3900 at pc 0x7f52c03d0a63 bp 0x7f52a25a0da0 sp 0x7f52a25a0548
WRITE of size 6238 at 0x625008bc3900 thread T30 ([ET_NET 28])
    #0 0x7f52c03d0a62 in __interceptor_memcpy (/lib64/libasan.so.8+0x70a62) (BuildId: 71dbf393857c775be459ab5583ba7b5fcbd9c884)
    #1 0xb38b3e in XpackDynamicTableStorage::write(char const*, unsigned int, char const*, unsigned int) /sd/workspace/src/git.ouryahoo.com/Edge/build/_scm/trafficserver10.0_asan/src/proxy/hdrs/XPACK.cc:545
    #2 0xb38b3e in XpackDynamicTable::insert_entry(char const*, unsigned long, char const*, unsigned long) /sd/workspace/src/git.ouryahoo.com/Edge/build/_scm/trafficserver10.0_asan/src/proxy/hdrs/XPACK.cc:351
    #3 0xb3936a in XpackDynamicTable::insert_entry(std::basic_string_view<char, std::char_traits<char> >, std::basic_string_view<char, std::char_traits<char> >) /sd/workspace/src/git.ouryahoo.com/Edge/build/_scm/trafficserver10.0_asan/src/proxy/hdrs/XPACK.cc:366
    #4 0x9ce105 in HpackIndexingTable::add_header_field(HpackHeaderField const&) /sd/workspace/src/git.ouryahoo.com/Edge/build/_scm/trafficserver10.0_asan/src/proxy/http2/HPACK.cc:367
[...]

Issue #10082, opened 2023-07-24
Return status 500 instead of abort


maskit:
`ink_release_assert` is used at over 400 places. Although having core files is helpful for debugging, it kills all connections abruptly  and it's not nice to clients. It'd be nice if we could just return status 500 and save connections/requests that don't face any issues.

jpeach:
`ink_release_assert` should be used for internal invariants (i.e. when you do something that never fails and you want to document that it never fails). If `ink_release_assert` ever fails, it ought to be a serious bug.

maskit:
Yes I'd use it like that, but some of `ink_release_assert` are casually used like "I want some hint (coredump) if this check fails".

bryancall:
I am closing this as a stale enhancement request with no clear path forward. The suggestion is to make process-aborting `ink_release_assert` checks instead fail the affected transaction with a 500 so unrelated connections are not killed. As noted in the discussion, `ink_release_assert` is intended for internal invariants that should never fail and that indicate a serious bug when they do, so converting them wholesale is not straightforward. The behavior is unchanged on current master (`include/tscore/ink_assert.h` still routes `ink_release_assert` through `_ink_assert`, which never returns, and there are over 400 call sites), and no implementation or design proposal has materialized since 2023. If a concrete, scoped plan emerges for converting specific transaction-level invariants into recoverable error responses, please open a new issue with those specifics and we can pursue it there.

PR #11396, opened 2024-05-26
XPACK: update_maximum_size for storage

bneradt:
This updates the XpackDynamicTableStorage memory allocated when XpackDynamicTable's update_maximum_size is called. Without this expansion, XpackDynamicTableStorage's memory is potentially overrun with future insertions of header fields.

Fixes: #11388

---

I ran this patch in production and verified that with this patch ATS 10 is now stable at Yahoo.

### DIFF
--- a/src/proxy/hdrs/XPACK.cc
+++ b/src/proxy/hdrs/XPACK.cc
@@ -29,6 +29,7 @@
 #include "tscore/Diags.h"
 #include "tscore/ink_memory.h"
 #include "tsutil/LocalBuffer.h"
+#include <cstdint>
 
 namespace
 {
@@ -412,11 +413,15 @@ XpackDynamicTable::update_maximum_size(uint32_t new_max_size)
   if (used < new_max_size) {
     this->_maximum_size = new_max_size;
     this->_available    = new_max_size - used;
+    this->_expand_storage_size(new_max_size);
     return true;
   }
 
+  // used is larger than or equal to new_max_size. This means that _maximum_size
+  // is shrinking and we need to evict entries to get the used space below
+  // new_max_size.
   bool ret = this->_make_space(used - new_max_size);
-  // Size update must suceeds
+  // Size update must succeed.
   if (ret) {
     this->_available    = new_max_size - (this->_maximum_size - this->_available);
     this->_maximum_size = new_max_size;
@@ -476,6 +481,18 @@ XpackDynamicTable::count() const
   }
 }
 
+void
+XpackDynamicTable::_expand_storage_size(uint32_t new_storage_size)
+{
+  ExpandCapacityContext context{this->_storage, new_storage_size};
+  uint32_t              i   = this->_calc_index(this->_entries_tail, 1);
+  uint32_t              end = this->_calc_index(this->_entries_head, 1);
+  for (; i != end; i = this->_calc_index(i, 1)) {
+    auto &entry  = this->_entries[i];
+    entry.offset = context.copy_field(entry.offset, entry.name_len + entry.value_len);
+  }
+}
+
 bool
 XpackDynamicTable::_make_space(uint64_t required_size)
 {
@@ -523,11 +540,10 @@ XpackDynamicTable::_calc_index(uint32_t base, int64_t offset) const
 // DynamicTableStorage
 //
 
-XpackDynamicTableStorage::XpackDynamicTableStorage(uint32_t size) : _head(size * 2 - 1), _tail(size * 2 - 1)
+XpackDynamicTableStorage::XpackDynamicTableStorage(uint32_t size)
+  : _overwrite_threshold(size), _capacity{size * 2}, _head{_capacity - 1}
 {
-  this->_data_size           = size * 2;
-  this->_data                = reinterpret_cast<uint8_t *>(ats_malloc(this->_data_size));
-  this->_overwrite_threshold = size;
+  this->_data = reinterpret_cast<uint8_t *>(ats_malloc(this->_capacity));
 }
 
 XpackDynamicTableStorage::~XpackDynamicTableStorage()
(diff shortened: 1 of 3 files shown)


src/proxy/hdrs/XPACK.cc
```diff
@@ -339,14 +339,22 @@ XpackDynamicTable::insert_entry(const char *name, size_t name_len, const char *v
...
   if (required_size > this->_available) {
     if (!this->_make_space(required_size - this->_available)) {
       // We can't insert a new entry because some stream(s) refer an entry that need to be evicted or the header is too big to
       // store. This is fine with HPACK, but not with QPACK.
       return {UINT32_C(0), XpackLookupResult::MatchType::NONE};
     }
   }
+  // Protect against overrun: ensure that our storage has space for the new
+  // header.
+  if (this->_storage.will_overrun(required_storage_size)) {
```
maskit, line 353: Looks redundant and I don't think it can be true. How does this happen?

src/proxy/hdrs/XPACK.cc
```diff
@@ -559,7 +578,50 @@ XpackDynamicTableStorage::write(const char *name, uint32_t name_len, const char
...
+XpackDynamicTableStorage::update_maximum_size(uint32_t new_maximum_size)
+{
+  if (new_maximum_size <= this->_capacity) {
+    // We never shrink our memory for the data storage because we don't support
+    // arbitrary eviction from it. The XpackDynamicTable class keeps track of
+    // field sizes and therefore can evict properly.
+    return true;
+  }
+  uint32_t const new_data_size = new_maximum_size * 2;
+  uint8_t       *new_data      = reinterpret_cast<uint8_t *>(ats_malloc(new_data_size));
```
maskit, line 616: Why don't you use `realloc()`? If we are lucky data copy (memcpy) doesn't happen.

bneradt:
The crash happens from the test_HPACK unit test in which storage is verifying there is enough space to erase (storage did not erase before this patch):

```
#6  0x00000000004d84c8 in XpackDynamicTableStorage::erase (this=0xfffff5108250, to_erase=963) at /home/bneradt/src/ts_asf_master_fix_xpack_overrun/src/proxy/hdrs/XPACK.cc:584
584       ink_release_assert(to_erase <= this->size());

(gdb) f 7
#7  0x00000000004d7b1c in XpackDynamicTable::_make_space (this=0xfffff5108228, required_size=18) at /home/bneradt/src/ts_asf_master_fix_xpack_overrun/src/proxy/hdrs/XPACK.cc:520
520         this->_storage.erase(storage_freed);
(gdb) l
515       // Evict
516       if (freed > 0) {
517         XPACKDbg("Evict entries: from %u to %u", this->_entries[this->_calc_index(this->_entries_tail, 1)].index,
518                  this->_entries[tail - 1].index);
519         this->_available += freed;
520         this->_storage.erase(storage_freed);
521         this->_entries_tail = tail - 1;
522         XPACKDbg("Available size: %u", this->_available);
523       }
524
(gdb) p freed
$7 = 995
(gdb) p storage_freed
$8 = 963
(gdb) p this->_storage
$9 = {
  _overwrite_threshold = 4096,
  _data = 0x525000000100 ":content-length0set-cookieACOOKIE=C8ctADEzMC4xMjkuNjguNzMtMjY4MDEwOTgyNC4zMDI1OTY1NgAAAAAAAAAEAAAAmLwAAJEelVCHHpVQ9r0AALkelVCSHpVQ+r0AAJ0elVCdHpVQ970AALwelVC8HpVQAQAAAHhHAAC8HpVQhx6VUAAAAAA-; path=/; "...,
  _capacity = 8192,
  _head = 1657,
  _tail = 845
}
(gdb) p 1657-845
$10 = 812
```

So there's only 812 bytes to erase, but 963 bytes are being erased.

src/proxy/hdrs/XPACK.cc
bneradt: It may be should be redundant, but this is the overrun that this patch addresses. Before this patch, this overrun happens because `_make_space` didn't expand _storage before this patch. Per your observation, that expansion is being added here.

In any case, rather than risk an overrun, I'm trying to add a defensive check. I understand your comment though. With the patch, it should be redundant. Maybe I should make this a release assert instead.

src/proxy/hdrs/XPACK.cc
bneradt: Thanks, that's a good idea.

src/proxy/hdrs/XPACK.cc
maskit: I think a *release* assert should be avoided. If `will_overrun` returns true, something is wrong and it's unrecoverable state, so the connection has to be closed anyway, however, we should not abandon all the other established connections. I'm fine with a regular (non-release) assert.

src/proxy/hdrs/XPACK.cc
```diff
@@ -559,7 +578,47 @@ XpackDynamicTableStorage::write(const char *name, uint32_t name_len, const char
 }
 
 void
-XpackDynamicTableStorage::erase(uint32_t name_len, uint32_t value_len)
+XpackDynamicTableStorage::erase(uint32_t to_erase)
+{
+  // Ensure that _tail is not overcoming _head.
+  ink_release_assert(to_erase <= this->size());
```
maskit, line 584: For the same reason above, I think we should avoid a release assert. See https://github.com/apache/trafficserver/issues/10082

src/proxy/hdrs/XPACK.cc
```diff
@@ -559,7 +578,47 @@ XpackDynamicTableStorage::write(const char *name, uint32_t name_len, const char
...
 {
-  this->_tail = (this->_tail + (name_len + value_len)) % this->_data_size;
+  if (new_maximum_size <= this->_capacity) {
+    // We never shrink our memory for the data storage because we don't support
+    // arbitrary eviction from it. The XpackDynamicTable class keeps track of
+    // field sizes and therefore can evict properly.
+    return true;
+  }
+  uint32_t const new_data_size = new_maximum_size * 2;
+  this->_data = reinterpret_cast<uint8_t *>(ats_realloc(this->_data, new_data_size));
```
maskit, line 616: Actually, it's not this simple. It doesn't work if ` _head < _tail`. Writing new data will overwrite data at the tail too early.

We may have to create a new Storage and copy all the entries into the new storage to update the offsets.

src/proxy/hdrs/XPACK.cc
bneradt: I appreciate the concern about release asserts, but guarding overruns is one place I suggest we should use them. Overruns cause mysterious crashes anyway (if we're lucky - otherwise we may be sending sensitive data to a host, 
 leading to a security issue), but it can be very difficult to find the culprit because the crash happens after the fact with little or no evidence as to who corrupted the memory. A release_assert at least makes the crash happen deterministically at the point of where the overrun will happen.

src/proxy/hdrs/XPACK.cc
bneradt: I tried to account for `_head < _tail` with the conditional check below this. Won't `_tail` only ever be past `_head` if `_tail` is still pointing at the end of the allocated memory? Otherwise I don't think `_tail` should ever overcome `_head`.

But I clearly am missing something because one of the unit tests is failing in this area. Maybe your comment here is the reason.

src/proxy/hdrs/XPACK.cc
maskit: I still think we shouldn't use release assert here, and I'd print logs and return an error so H2 impl can send GOAWAY frame with COMPRESSION_ERROR. But let's revisit this on #10082.

src/proxy/hdrs/XPACK.cc
bneradt: Thank you for explaining this further to me on slack!

Yes, you are write that things are more complicated after space is expanded. As you suggested, I now call `insert_entry` for each active entry after the resize so that the offsets are appropriate in the dynamic table and the entries are laid out appropriately in the storage. I added a unit test for this as well.

maskit (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11398, opened 2024-05-29
Fixes Bundles errors, and better checks

zwoop:
The critical fix here is the check for errors, somehow in the translation from internal to upstream, this check got wrong. This is so severe, that this must be fixed in v10.0.x.

Also, I added some better tracking and checks around the Bundles instantiation, assuring that each bundle is only activated once.

### DIFF
--- a/src/cripts/CMakeLists.txt
+++ b/src/cripts/CMakeLists.txt
@@ -45,8 +45,10 @@ set(CRIPTS_PUBLIC_HEADERS
     ${PROJECT_SOURCE_DIR}/include/cripts/Urls.hpp
     ${PROJECT_SOURCE_DIR}/include/cripts/UUID.hpp
     ${PROJECT_SOURCE_DIR}/include/cripts/Configs.hpp
-    ${PROJECT_SOURCE_DIR}/include/cripts/Bundles/Common.hpp
-    ${PROJECT_SOURCE_DIR}/include/cripts/Bundles/LogsMetrics.hpp
+)
+
+set(CRIPTS_BUNDLE_HEADERS ${PROJECT_SOURCE_DIR}/include/cripts/Bundles/Common.hpp
+                          ${PROJECT_SOURCE_DIR}/include/cripts/Bundles/LogsMetrics.hpp
 )
 
 add_library(cripts SHARED ${CPP_FILES})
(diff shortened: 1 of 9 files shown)


cmcfarlen:
[approve ci autest]

src/cripts/CMakeLists.txt
```diff
@@ -45,8 +45,10 @@ set(CRIPTS_PUBLIC_HEADERS
     ${PROJECT_SOURCE_DIR}/include/cripts/Urls.hpp
     ${PROJECT_SOURCE_DIR}/include/cripts/UUID.hpp
     ${PROJECT_SOURCE_DIR}/include/cripts/Configs.hpp
-    ${PROJECT_SOURCE_DIR}/include/cripts/Bundles/Common.hpp
-    ${PROJECT_SOURCE_DIR}/include/cripts/Bundles/LogsMetrics.hpp
+)
+
+set(CRIPTS_BUNDLE_HEADERS ${PROJECT_SOURCE_DIR}/include/cripts/Bundles/Common.hpp
+                          ${PROJECT_SOURCE_DIR}/include/cripts/Bundles/LogsMetrics.hpp
```
JosiahWI, line 51: Shouldn't the paths be quoted? (also above)
If `PROJECT_SOURCE_DIR` contains spaces and doesn't already have quotes, this will break.

src/cripts/CMakeLists.txt
```diff
@@ -64,6 +66,9 @@ set_target_properties(
 
 install(TARGETS cripts PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/cripts)
 
+# The PUBLIC_HEADER target can't handle include files in a subdirectory, so do those manually.
+install(FILES ${CRIPTS_BUNDLE_HEADERS} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/cripts/Bundles)
```
JosiahWI, line 70: Shouldn't the path be quoted?

JosiahWI (changes requested):

src/cripts/CMakeLists.txt
zwoop: It wasn't in the install target above (old code), so I assumed you didn't need it. I can certainly add the quotes.

src/cripts/CMakeLists.txt
zwoop: They weren't quoted before, and I just moved them.

src/cripts/CMakeLists.txt
JosiahWI: There are other paths that might need quoting, so I will make a PR for this separately across all CMakeLists.txt files.

zwoop:
[approve ci autest]

cmcfarlen (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11399, opened 2024-05-29
Fix race conditions in Regex.cc.

ywkaras:
These changes should prevent (low-frequency) crashes that are being seen in Yahoo Prod.


ywkaras:
[approve ci format]

cmcfarlen:
Should this be added to 10.0.x project as well for cherry-pick?

ywkaras:
> Should this be added to 10.0.x project as well for cherry-pick?

Debatable.  Any crashes would happen at shutdown, and seem infrequent (with limited testing in Yahoo Prod).

ywkaras:
This is where the segment violation is occurring in Yahoo Prod:  https://github.com/apache/trafficserver/blob/5f52c374dfcc3dc223a7c23bd885bac64ec2398f/src/tsutil/Regex.cc#L114

cmcfarlen (changes requested):
Could RegexContext be automatic within the thread? i.e.

```
RegexContext* get_instance() {
  thread_local RegexContext ctx;
  return &ctx;
}
```

Why is the global vector of contexts necessary at all?

IIRC this was originally done because some Debug statements in destructors could come after the shutdown of the thread's regex context.  But if we use Dbg in those locations, then this isn't a problem anymore?

ywkaras:
Yes, it would not be a logic error in creating automatic, function scope instances of RegexContext when one is needed.  The issue is whether the performance hit is tolerable.

ywkaras:
> Could also look at how Metrics storage lifetime outlasts the threads and see if this would work for _Data lifetime:
> 
> https://github.com/apache/trafficserver/blob/master/src/tsutil/Metrics.cc#L34-L42

Because of https://github.com/apache/trafficserver/blob/0a53513961cbe55a6bfde32be208c0e04a70b0f1/src/tsutil/DbgCtl.cc#L161 , RegexContext::get_instance() cannot cause the construction of a thread_local variable.  It can only use thread_locals that are initialized by zero of const initialization.  The Metrics code is not limited by that restriction.

I guess I should look into changing DbgCtl:::_new_reference() so it doesn't call tag_activated() with its mutex locked.

ywkaras:
Closing this because I think https://github.com/apache/trafficserver/pull/11416 is better.

PR #11400, opened 2024-05-30
Update Http3Frame.cc - fix Use-of-uninitialized-value error

shukitchan:
Fuzzing reveals a use-of-uninitialized-pointer-field case

https://oss-fuzz.com/testcase-detail/6159793621368832

I think this should fix the problem

### DIFF
--- a/src/proxy/http3/Http3Frame.cc
+++ b/src/proxy/http3/Http3Frame.cc
@@ -285,7 +285,7 @@ Http3SettingsFrame::Http3SettingsFrame(const uint8_t *buf, size_t buf_len, uint3
       this->_error_reason = reinterpret_cast<const char *>("invalid SETTINGS frame");
       break;
     }
-    
+
     size_t   value_len  = QUICVariableInt::size(buf + len);
     uint64_t value      = QUICIntUtil::read_QUICVariableInt(buf + len, buf_len - len);
     len                += value_len;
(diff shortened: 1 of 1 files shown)


src/proxy/http3/Http3Frame.cc
```diff
@@ -280,10 +280,20 @@ Http3SettingsFrame::Http3SettingsFrame(const uint8_t *buf, size_t buf_len, uint3
     size_t   id_len  = QUICVariableInt::size(buf + len);
     uint16_t id      = QUICIntUtil::read_QUICVariableInt(buf + len, buf_len - len);
     len             += id_len;
+    if (len > buf_len) {
```
maskit, line 283: It should probably be `len >= buf_len`. Having only id is malformed, and current code (even with your fix) can cause buffer-over-run on `size_t value_len = QUICVariableInt::size(buf + len);`.

maskit (changes requested):
Setting `_error_code` and `_error_reason` is fine. Just the condition is a bit loose.

src/proxy/http3/Http3Frame.cc
shukitchan: Yes. I think even the reading is dangerous. I should give up before the reading

shukitchan:
[approve ci autest]

maskit (approved):
Looks good. Thanks.

cmcfarlen:
Cherry-picked to v10.0.x

Issue #11395, opened 2024-05-25
Missed to pass `length` parameter in the tscpp api - `HeaderField::append`


freak82:
Hi there,

The [HeaderField::append](https://github.com/apache/trafficserver/blob/b46f5d061af455dc75956346da3308788381b30d/src/tscpp/api/Headers.cc#L310) function overload doesn't pass the input length parameter to the `TSMimeHdrFieldValueStringInsert`.
Instead it passes `-1` which means that the function will use the full string (assuming that it's null terminated) instead just `length` number of characters.
```
bool
HeaderField::append(const char *value, int length)
{
  return (TSMimeHdrFieldValueStringInsert(iter_.state_->mloc_container_->hdr_buf_, iter_.state_->mloc_container_->hdr_loc_,
                                          iter_.state_->mloc_container_->field_loc_, -1, value, -1) == TS_SUCCESS);
}
```
Should I do a pull request for this where the `length` is passed accordingly to the `TSMimeHdrFieldValueStringInsert` function?

freak82:
Just a bit of clarification about these issues for missed to pss parameters.
I've patches which allow the `-Wno-unused-parameter` warning suppression to be removed.
I have intention to do pull requests for them subsystem by subsystem. The problem is that for some of  them I think that the unused parameter is a possible bug, like this one, and the correct solution is not just to put `[[maybe_unused]]` in front of it.
On the other hand the correct fix may introduce bugs in some user/plugins level code due to the [Hyrym's Law.](https://www.hyrumslaw.com/) where the client code already relies on the undocumented (possibly wrong) behavior of the API.

PR #11401, opened 2024-05-30
Pass unused length parameter in HeaderField::append function

freak82:
The parameter is passed to the TSMimeHdrFieldValueStringInsert instead of the current -1.
The latter causes `strlen(value)` instead `length` bytes to be inserted.
Related [issue](https://github.com/apache/trafficserver/issues/11395).


bryancall (approved):

JosiahWI:
Fixes #11395.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11402, opened 2024-05-30
Fix unused parameters in the tscpp source code

freak82:
This is related to the discussion [here](https://github.com/apache/trafficserver/issues/11377)


cmcfarlen (approved):

cmcfarlen (changes requested):
For function parameters that are always unused, we should just remove the parameter name (or comment the name out with /* */).

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-unused

If conditionally unused due to constexpr or a macro we can use `[[maybe unused]]` to avoid the warning.

src/tscpp/api/utils_internal.cc
```diff
@@ -260,7 +260,7 @@ utils::internal::invokePluginForEvent(GlobalPlugin *plugin, TSHttpTxn ats_txn_ha
 }
 
 void
-utils::internal::invokePluginForEvent(GlobalPlugin *plugin, TSHttpAltInfo altinfo_handle, TSEvent event)
+utils::internal::invokePluginForEvent(GlobalPlugin *plugin, TSHttpAltInfo altinfo_handle, [[maybe_unused]] TSEvent event)
```
freak82, line 263: There is an assertion in this function which uses the `event` argument.
That's why it's with `[[maybe_unused]]` instead of like the other ones.

freak82:
I updated this pull request to use `/* name ATS_UNUSED */` as it was decided in the discussion [here](https://github.com/apache/trafficserver/pull/11406).

JosiahWI:
@freak82 It looks like the length parameter to `Headers::count` is no longer unused after your fix #11405 was applied. Could you update this PR accordingly?

freak82:
@JosiahWI 
The PR is updated.

JosiahWI:
[approve ci rocky]

JosiahWI (approved):
Thanks for the cleanup, looks good!

cmcfarlen (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11405, opened 2024-05-31
Use the specified `length` characters of the key in `Headers::count`

freak82:
The `length` parameters in the `Headers::count` was not taken into account. This lead to the case insensitive comparison being executed on the whole `key`, assuming it's null terminated, than on `length` number of bytes.
This change is related to the discussion [here](https://github.com/apache/trafficserver/issues/11377): the unused `length` parameter is reported when the warning suppression is removed.

I tried to keep to minimum the related changes in the other classes in the `Headers.cc` although the whole functionality there could be done in a better way (IMO):
- the needless string copying here and there can be removed
- the const correctness of the functions can be improved/fixed
- the iterator interface can be improved
- operators between the classes can be reused and/or generated by the compiler (in C++ 20).
- some functions/operators for implict casting should be removed (IMO) because this "convenience" have the price of hiding heap allocations

Last, this check in `Headers::count`
```
if ((*it).name() == key_name)
```
can be done like this:
```
if (*it == key_name)
```
generates a bit better assembly for the whole function: 10 lines shorter where GCC just seem to shuffle some more registers around. 
On the whole I preferred to keep it the related changes to as few as possible because the core issue leading to this is the unused `length` parameter.
Not all of the added operators in the `Headers` functionality are needed for this fix but I added them for consistency.

### DIFF
--- a/include/tscpp/api/Headers.h
+++ b/include/tscpp/api/Headers.h
@@ -91,6 +91,12 @@ public:
    * Case insensitive comparison of this HeaderFieldName
    * @return true if the two strings are equal.
    */
+  bool operator==(std::string_view field_name);
+
+  /**
+   * Case insensitive comparison of this HeaderFieldName
+   * @return true if the two strings are equal.
+   */
   bool operator==(const std::string &field_name);
 
   /**
@@ -98,6 +104,12 @@ public:
    * @return true if the two strings are not equal.
    */
   bool operator!=(const char *field_name);
+
+  /**
+   * Case insensitive comparison of this HeaderFieldName
+   * @return true if the two strings are not equal.
+   */
+  bool operator!=(std::string_view field_name);
 
   /**
    * Case insensitive comparison of this HeaderFieldName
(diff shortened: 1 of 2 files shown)


maskit:
The fix for `Headers::count` looks fine. However, I am not sure if we should add the operators. Because 1) this C++ API will be deprecated on version 10.0, which is going to be released in very near future (hopefully within 2 months), and b) this is a bug fix and it should be back ported to old versions. This change wouldn't break any compatibility, so technically we can add the operators to old versions, but adding anything new to 8.1.x at this time is very unusual and we'd still have to go through the API review process.

Although the additions makes perfect sense, I think you could achieve the same thing without the operators by calling `name()->c_str()` and `name()->length()` if you wanted to avoid any change to the API. I'd go that way to make it a simple bug fix.

But it's not a judgement call. If you think we should have the operators, you can start the API review process on the mailing list and request for feedback.
https://cwiki.apache.org/confluence/display/TS/API+Review+Process

freak82:
Yeah, I read somewhere that the C++ API is going to be deprecated but wasn't sure for the best approach because doing the case insensitive comparison outside would mean that I duplicate the cpp API functionality.
Anyway, I think you have a point and I'll do it as you suggested.
I'll do it without the API change then but I'll have time for this at the end of this week.

maskit:
> doing the case insensitive comparison outside would mean that I duplicate the cpp API functionality.

Yes, that is unfortunate. If the API were not going to be deprecated the decision could have been hard.

> I'll do it without the API change then but I'll have time for this at the end of this week.

Thanks. No rush.

freak82:
I removed the added tscpp API operators.

maskit (approved):
Thanks!

ezelkow1:

maskit:
Yes, this is a bug fix, and I think it's safe to back port.

PR #11408, opened 2024-05-31
Fix XpackDynamicTable::_make_space return value

bneradt:
Callers expect `_make_space` to return true if it was able to free the amount of given space, but it rather returns true if the amount of space available exceeds the passed in value. This fixes the function to return the intended result.


maskit (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11410, opened 2024-06-01
Http2ConnectionState::restart_receiving: check for closed peer

bneradt:
I noticed a crash in obtaining the peer's session window in which the session's netvc was nullptr. Digging into the backtrace, this seems to have occurred because restart_receiving was called for a closed session. This patch addresses this by handling this situation.

---

I ran this over the weekend in production and the box was stable.

### DIFF
--- a/src/proxy/http2/Http2CommonSession.cc
+++ b/src/proxy/http2/Http2CommonSession.cc
@@ -419,6 +419,9 @@ Http2CommonSession::do_process_frame_read(int event, VIO *vio, bool inside_frame
 bool
 Http2CommonSession::_should_do_something_else()
 {
+  if (this->get_proxy_session()->is_peer_closed()) {
+    return false;
+  }
   if (this->_interrupt_reading_frames) {
     this->_interrupt_reading_frames = false;
     return true;
--- a/src/proxy/http2/Http2ConnectionState.cc
+++ b/src/proxy/http2/Http2ConnectionState.cc
@@ -1905,6 +1905,10 @@ Http2ConnectionState::restart_streams()
 void
 Http2ConnectionState::restart_receiving(Http2Stream *stream)
 {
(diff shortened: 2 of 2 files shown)


maskit (approved):

cmcfarlen:
We are also seeing a crash in this area.  I wonder though if the check for close should come sooner.  In our case at least, `restart_receiving` is called after `_should_do_something_else` returns true.  Should the check for closed be in that function instead?

cmcfarlen:
```    if (this->_should_do_something_else()) {
      if (this->_reenable_event == nullptr) {
        this->connection_state.restart_receiving(nullptr);
        vio->disable();
        this->_reenable_event = this->get_mutex()->thread_holding->schedule_in(this->get_proxy_session(), HRTIME_MSECONDS(1),
                                                                               HTTP2_SESSION_EVENT_REENABLE, vio);
        return 0;
      }
    }
```
This is where this function is called leading up to the crash.

bneradt:
> We are also seeing a crash in this area. I wonder though if the check for close should come sooner. In our case at least, `restart_receiving` is called after `_should_do_something_else` returns true. Should the check for closed be in that function instead?

Yes, what you're describing is the exact crash I'm addressing.

I'm definitely open to your feedback. I placed the check here in `restart_receiving` because I figured that would catch other code paths that can land there. Note that `restart_receiving` is called from a variety of locations, not just after `_should_do_something_else()`. Clearly we don't want to restart receiving data on a closed connection, so I thought it would be a wider and safer scope to catch it there.

bneradt:
[approve ci autest]

maskit (approved):

bneradt:
[approve ci autest]

cmcfarlen:
Cherry-picked to v10.0.x

PR #11416, opened 2024-06-04
Avoid race conditions in Regex.cc.

ywkaras:


ywkaras:
[approve ci autest]

src/tsutil/DbgCtl.cc
```diff
@@ -150,49 +150,55 @@ DbgCtl::_new_reference(char const *tag)
...
-  auto res = d.map.insert(new_elem);
+    debug_assert(res.second);
+
+    new_tag_data = &*res.first;
+  }
+  new_tag_data->second = p && p->debug_tag_activated(tag);
 
-  debug_assert(res.second);
+  // It is important that debug_tag_activated() is NOT called while the ra object exists, and the registry mutex is
+  // locked.  There is a mutex in the C/C++ runtime that both dlopen() and _cxa_thread_atexit() lock while running.
```
cmcfarlen, line 188: :+1: Good comment update here and good idea moving this outside of the lock as a solution to the deadlock.

PR #11417, opened 2024-06-04
Adds a bridge betwen Cripts and HRW operators

zwoop:
This also adds some new mechanisms to some Cripts core, and updates documentation as well. 

Note: This does not handle all HRW conditions yet, rather, I think we can add them as people have need for them.

### DIFF
--- a/src/cripts/Bundles/Headers.cc
+++ b/src/cripts/Bundles/Headers.cc
@@ -22,7 +22,7 @@
 
 namespace
 {
-enum HeaderTargets {
+enum HeaderTargets : uint8_t {
   NONE,
   CLIENT_REQUEST,
   CLIENT_RESPONSE,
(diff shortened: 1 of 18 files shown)


src/cripts/Bundles/Headers.cc
```diff
@@ -0,0 +1,183 @@
...
+namespace Bundle
+{
+const Cript::string Headers::_name = "Bundle::Headers";
+
+bool
+Headers::validate(std::vector<Cript::Bundle::Error> &errors) const
+{
+  bool failed = false;
+  // errors.emplace_back("dscp must be between 0 and 63", name(), "dscp");
+  // failed = true;
```
cmcfarlen, line 60: This function seems unfinished.

cmcfarlen (approved):

src/cripts/Bundles/Headers.cc
zwoop: fixed / removed, it was a remnant.

zwoop:
[approve ci autest]

cmcfarlen:
Cherry-picked to v10.0.x

PR #11419, opened 2024-06-05
Fix all Sphinx doc generation warnings

bneradt:
This fixes all docs Sphinx warnings and turns on the `-W` flag to sphinx so that the build will fail if any warnings get introduced in the future.

Many of these changes involve transitioning many of the domains from :c: to :cpp:. Note there are still uses for the :c: domain since some directives we use are only supported there, such as our macros. But many functions and enumerations that used to be c are now cpp.


bneradt:
[approve ci autest ubuntu]

bneradt:
[approve ci autest ubuntu]

cmcfarlen (approved):

PR #11421, opened 2024-06-06
Add Remap ACL table test

mattyw:
This change adds a table test that tests every combination of ACLs, namely:

1. remap.config inline rule
2. named ACLs in remap.config.
3. ip_allow.yaml.

It tests the expected response for a GET request (explicity denied or allowed in the test) and a POST (implicitly denied or allowed by the GET rule)

The current test implementation is optimised for minimal code changes. Because of this a new server is started for each test. Because of this the new tests alone will add 1m20s to the total autest testing time.

This could be optimised if we feel that this style of test is worth considering.

### DIFF
--- a/tests/gold_tests/remap/remap_acl.test.py
+++ b/tests/gold_tests/remap/remap_acl.test.py
@@ -18,7 +18,13 @@ Verify remap.config acl behavior.
 #  limitations under the License.
 
 import os
+import io
 import re
+import pathlib
+import inspect
+import tempfile
+from yaml import load, dump
+from yaml import CLoader as Loader
 from typing import List, Tuple
 
 Test.Summary = '''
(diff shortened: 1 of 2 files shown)


tests/gold_tests/remap/remap_acl.test.py
```diff
@@ -296,3 +329,267 @@ def _configure_client(self, tr: 'TestRun') -> None:
...
+                if method == "GET":
+                    transaction["proxy-response"]["status"] = get_proxy_response
+                elif method == "POST":
+                    transaction["proxy-response"]["status"] = post_proxy_response
+                else:
+                    raise Exception(
+                        "Expected to find GET or POST request, found %s", method
+                    )
+    with open(replay_file, "w") as f:
+        f.write(dump(data))
```
mattyw, line 380: This is also less than ideal: We create a number of temp files to support the 27 replay files we need.

mattyw:
This is now being worked on in #11433

PR #11424, opened 2024-06-06
Modernize ja3 fingerprint encoding implementation 

JosiahWI:
This PR makes the following changes:

  * This reorders some imports in ja3_fingerprint.cc.
  * This adds unit tests for the behavior of custom_get_ja3_prefixed.
  * This rewrites custom_get_ja3_prefixed as two separate functions,
    encode_word_buffer and encode_dword_buffer. The new implementations
    both offer a 20% performance improvement over the old one, measured on a
    release build with -O3 using Catch2 benchmarks and a 10 byte buffer that
    included GREASE values.
  * This adds a function encode_integer_buffer for encoding the buffer of TLS
    extensions.
  * This documents the new functions and moves them to a new ja3:: namespace
    in a separate source file.
  * This puts the steps of custom_get_ja3 in the right order, improves its
    variable names, and uses some std::string methods instead of + for string
    concatenation.

### DIFF
--- a/plugins/ja3_fingerprint/ja3_fingerprint.cc
+++ b/plugins/ja3_fingerprint/ja3_fingerprint.cc
@@ -16,17 +16,9 @@
   limitations under the License.
  */
 
-#include <cstdlib>
-#include <cstdio>
-#include <cstring>
-#include <cstdlib>
 #include <getopt.h>
 #include <netinet/in.h>
 #include <arpa/inet.h>
-
-#include <string>
-#include <unordered_set>
-#include <memory>
 
 #include "ts/apidefs.h"
 #include "ts/ts.h"
@@ -39,6 +31,13 @@
 #include <openssl/ssl.h>
 #include <openssl/md5.h>
 #include <openssl/opensslv.h>
+
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <memory>
+#include <string>
+#include <unordered_set>
 
 // Get 16bit big endian order and update pointer
 #define n2s(c, s) ((s = (((unsigned int)(c[0])) << 8) | (((unsigned int)(c[1])))), c += 2)
--- a/plugins/ja3_fingerprint/ja3_fingerprint.cc
+++ b/plugins/ja3_fingerprint/ja3_fingerprint.cc
@@ -1,6 +1,9 @@
-/** @ja3_fingerprint.cc
+/** @file ja3_fingerprint.cc
+ *
   Plugin JA3 Fingerprint calculates JA3 signatures for incoming SSL traffic.
+
   @section license License
+
   Licensed to the Apache Software Foundation (ASF) under one
   or more contributor license agreements.  See the NOTICE file
   distributed with this work for additional information
@@ -8,12 +11,15 @@
   to you under the Apache License, Version 2.0 (the
   "License"); you may not use this file except in compliance
   with the License.  You may obtain a copy of the License at
+
       http://www.apache.org/licenses/LICENSE-2.0
+
   Unless required by applicable law or agreed to in writing, software
   distributed under the License is distributed on an "AS IS" BASIS,
   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
   See the License for the specific language governing permissions and
   limitations under the License.
+
  */
 
 #include "ja3_utils.h"
@@ -107,7 +113,7 @@ custom_get_ja3(SSL *ssl)
 
   // Get cipher suites
   len = SSL_client_hello_get0_ciphers(ssl, &buf);
-  result.append(ja3::encode_dword_buffer(buf, len));
+  result.append(ja3::encode_word_buffer(buf, len));
   result.push_back(',');
 
   // Get extensions
@@ -121,14 +127,14 @@ custom_get_ja3(SSL *ssl)
   // Get elliptic curves
   if (SSL_client_hello_get0_ext(ssl, 0x0a, &buf, &len) == 1) {
     // Skip first 2 bytes since we already have length
-    result.append(ja3::encode_dword_buffer(buf + 2, len - 2));
+    result.append(ja3::encode_word_buffer(buf + 2, len - 2));
   }
   result.push_back(',');
 
   // Get elliptic curve point formats
   if (SSL_client_hello_get0_ext(ssl, 0x0b, &buf, &len) == 1) {
     // Skip first byte since we already have length
-    result.append(ja3::encode_word_buffer(buf + 1, len - 1));
+    result.append(ja3::encode_byte_buffer(buf + 1, len - 1));
   }
   return result;
 }
--- a/plugins/ja3_fingerprint/ja3_fingerprint.cc
+++ b/plugins/ja3_fingerprint/ja3_fingerprint.cc
@@ -16,6 +16,8 @@
   limitations under the License.
  */
 
+#include "ja3_utils.h"
+
 #include <getopt.h>
 #include <netinet/in.h>
 #include <arpa/inet.h>
@@ -37,10 +39,6 @@
 #include <cstring>
 #include <memory>
 #include <string>
-#include <unordered_set>
-
-// Get 16bit big endian order and update pointer
-#define n2s(c, s) ((s = (((unsigned int)(c[0])) << 8) | (((unsigned int)(c[1])))), c += 2)
 
 const char            *PLUGIN_NAME = "ja3_fingerprint";
 static DbgCtl          dbg_ctl{PLUGIN_NAME};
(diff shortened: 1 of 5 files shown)


JosiahWI:
[approve ci format]

JosiahWI:
[approve ci]

plugins/ja3_fingerprint/ja3_fingerprint.cc
```diff
@@ -126,52 +94,43 @@ getIP(sockaddr const *s_sockaddr, char res[INET6_ADDRSTRLEN])
 }
 
 static std::string
-custom_get_ja3(SSL *s)
+custom_get_ja3(SSL *ssl)
```
bneradt, line 97: Just checking: can this not be const? I'm assuming the APIs probably take them non-const which prevents that, but it seems a shame since the plugin should function read-only on this stuff.

plugins/ja3_fingerprint/ja3_utils.cc
```diff
@@ -0,0 +1,102 @@
+/** @ja3_utils.cc
+  Plugin JA3 Fingerprint calculates JA3 signatures for incoming SSL traffic.
+  @section license License
```
bneradt, line 3: Probably good to space this out a bit, like here:
https://github.com/apache/trafficserver/blob/master/src/proxy/http/HttpSM.cc#L3

plugins/ja3_fingerprint/ja3_utils.cc
```diff
@@ -0,0 +1,102 @@
+/** @ja3_utils.cc
```
bneradt, line 1: I know this is copied from the other ja3_fingerprint.cc header file, but I think it's supposed to be `@file ja3_utils.cc`. Can you fix it in the ja3_fingerprint.cc file too?

plugins/ja3_fingerprint/ja3_utils.cc
```diff
@@ -0,0 +1,102 @@
...
+{
+
+// GREASE table as in ja3
+static std::unordered_set<std::uint16_t> const GREASE_table = {0x0a0a, 0x1a1a, 0x2a2a, 0x3a3a, 0x4a4a, 0x5a5a, 0x6a6a, 0x7a7a,
+                                                               0x8a8a, 0x9a9a, 0xaaaa, 0xbaba, 0xcaca, 0xdada, 0xeaea, 0xfafa};
+
+static constexpr std::uint16_t
+from_big_endian(unsigned char lowbyte, unsigned char highbyte)
+{
+  return (static_cast<std::uint16_t>(lowbyte) << 8) + highbyte;
```
bneradt, line 34: Bitwise or (`|`) maybe? Could be slightly more efficient.

plugins/ja3_fingerprint/ja3_utils.cc
```diff
@@ -0,0 +1,102 @@
...
+    it += 2;
+  }
+  if (it < buf + len) {
+    // Benchmarks show that reserving buf.size() - 1 space in the string here
+    // would have no impact on performance. Since the string may not even need
+    // that much due to GREASE values present in the buffer, we don't do it.
+    result.append(std::to_string(from_big_endian(it[0], it[1])));
+    it += 2;
+    for (; it < buf + len; it += 2) {
+      auto value{from_big_endian(it[0], it[1])};
```
bneradt, line 75: might as well `const`

plugins/ja3_fingerprint/ja3_utils.cc
```diff
@@ -0,0 +1,102 @@
...
+  }
+  return result;
+}
+
+std::string
+encode_dword_buffer(unsigned char const *buf, int const len)
+{
+  std::string result;
+  auto        it{buf};
+  while (it < buf + len && ja3_should_ignore(from_big_endian(it[0], it[1]))) {
```
bneradt, line 65: I suggest adding some parens to help the reader a bit:

```cpp
while ((it < (buf + len)) && ja3_should_ignore(from_big_endian(it[0], it[1]))) {
```

plugins/ja3_fingerprint/ja3_utils.cc
```diff
@@ -0,0 +1,102 @@
...
+    it += 2;
+  }
+  if (it < buf + len) {
```
bneradt, line 68:
```cpp
if (it < (buf + len)) {

File: plugins/ja3_fingerprint/ja3_utils.cc [hunk 8]
```diff
@@ -0,0 +1,102 @@
...
+  std::string result;
+  if (len > 0) {
+    // Benchmarks show that reserving space in the string here would cause
+    // a 40% increase in runtime for a buffer with 10 elements... so we
+    // don't do it.
+    result.append(std::to_string(buf[0]));
+    std::for_each(buf + 1, buf + len, [&result](unsigned char i) {
+      result.push_back('-');
+      result.append(std::to_string(i));
+    });
```
bneradt, line 55: Nice use of algorithms.

plugins/ja3_fingerprint/ja3_utils.h
```diff
@@ -0,0 +1,65 @@
...
+ * joined with the '-' character. Any GREASE values in the buffer will be
+ * ignored.
+ *
+ * @param buf The buffer to encode. This should be a big-endian SSL buffer
+ *  of 16bit values.
+ * @param len The length of the buffer. If the length is zero, buf will not
+ *  be dereferenced.
+ * @return The string-encoded ja3 representation of the buffer.
+ */
+std::string encode_dword_buffer(unsigned char const *buf, int const len);
```
bneradt, line 49: I know you're not introducing this terminology here, but isn't this the typical names of things?

* `byte`: 8 bits
* `word`: 16 bits
* `dword`: 32 bits

plugins/ja3_fingerprint/ja3_utils.h
```diff
@@ -0,0 +1,65 @@
...
+ * joined with the '-' character. Any GREASE values in the buffer will be
+ * ignored.
+ *
+ * @param buf The buffer to encode. The buffer underlying the span should be
+ *  an SSL buffer of ints.
+ * @param len The length (number of values) in the buffer. If the length is
+ *  zero, buf will not be dereferenced.
+ * @return The string-encoded ja3 representation of the buffer.
+ */
+std::string encode_integer_buffer(int const *buf, int const len);
```
bneradt, line 63: Thanks for right side const. :)

Copy by value parameters aren't typically made const though (`int len`).

bneradt (changes requested):
Looks great. I just have a few small questions and comments. Thank you @JosiahWI .

plugins/ja3_fingerprint/ja3_utils.cc
```diff
@@ -0,0 +1,104 @@
...
+}
+
+std::string
+encode_word_buffer(unsigned char const *buf, int const len)
+{
+  std::string result;
+  if (len > 0) {
+    // Benchmarks show that reserving space in the string here would cause
+    // a 40% increase in runtime for a buffer with 10 elements... so we
+    // don't do it.
```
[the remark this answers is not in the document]
bneradt: It's too bad std::string doesn't have a constructor that pre-allocates the reserved bytes. That might have helped with this?

plugins/ja3_fingerprint/ja3_utils.h
JosiahWI: This was a precaution to make sure the compiler knows `len` won't change, so that duplicated expressions of `buf + len` won't result in recomputing the result.

JosiahWI: It may not be typical, and I didn't originally have it, but when I gave some thought to it I realized it still gives the compiler potentially useful information that could increase performance.

plugins/ja3_fingerprint/ja3_utils.h
JosiahWI: Thank you for telling me I used the wrong terms. I will need to fix these.

plugins/ja3_fingerprint/ja3_utils.cc
JosiahWI: It's also more consistent for the reader, because this is in the context of bit operations.

plugins/ja3_fingerprint/ja3_fingerprint.cc
JosiahWI: The APIs take them non-const. This is indeed a shame.

plugins/ja3_fingerprint/ja3_utils.cc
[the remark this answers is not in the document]
JosiahWI: I'm not sure whether that would have helped. This result was very surprising to me, and I can't explain it.

plugins/ja3_fingerprint/ja3_utils.cc
JosiahWI: Could we leave the `ja3_utils.cc` part off completely as in the core, and keep just the `@file` like everywhere else?

JosiahWI: https://www.doxygen.nl/manual/commands.html#cmdfile

Seems it should be done as you said.

plugins/ja3_fingerprint/ja3_utils.h
bneradt: Gotcha. It won't hurt anythying.

bneradt (approved):
Looks great. Thanks for the updates and the new tests.

ezelkow1 (approved):

PR #11426, opened 2024-06-06
Make Cripts contexts use Proxy Allocator, and cleanup

zwoop:
In addition:

* Some docs updates
* Makes the immutable URLs (such as remap From-URL) read-only.
* Don't reset() those immutable URLs, once setup, it can be reused between all hooks inside the transaction.

Fwiw, the reset to reset() URLs between hooks is because other plugins may modify URLs, and then the header heaps etc. can reallocate / move, and then it's hilarious to debug ...


cmcfarlen (dismissed):

JosiahWI:
The AuTests prefetch_simple and proxy_serve_stale failed.

JosiahWI:
[approve ci autest]

JosiahWI:
The AuTest tunnel_transform failed (client returned exit code 35).

JosiahWI:
[approve ci autest]

JosiahWI:
The AuTest proxy_protocol failed on the run "Proxy Protocol v1 without blind tunneling on TLS connection to origin" because the default process returned exit code 1 instead of 0.

JosiahWI:
[approve ci autest]

JosiahWI:
[approve ci autest]

cmcfarlen (approved):

cmcfarlen:
Cherry-picked to v10.0.x

Issue #11431, opened 2024-06-08
Intermittent CI failures of ja3_fingerprint


JosiahWI:
After investigating the CI failures, I discovered that they are due to a race condition between the `DUMP_HEADER` macro and some other debug logging in `HttpTransact`. The debug output can be interleaved because the `DUMP_HEADER` macro makes separate, unsynchronized calls to `fprintf(stderr, ...)`, and the AuTest fails because traffic.out doesn't match the gold file. I intend to fix this by batching everything in one call to `fprintf` instead of separate ones.

PR #11432, opened 2024-06-08
Fix race condition causing interleaved logs 

JosiahWI:
The output of the `DUMP_HEADER` macro was being written in multiple calls to
fprintf() that raced with other logging, resulting in interleaved output that
intermittently failed the ja3_fingerprint AuTest on CI. Whether other AuTest
failures were related to the same issue is unknown at this point.

This fixes the race condition by batching the calls to a single fprintf() call.
This also refactors the macro to two inline functions. In a benchmark using an
expanded version of the headers in the HttpTransact tests, a 60% speedup was
measured over the macro if all the header data fit in the buffer. In the worst
case that the buffer could only hold a byte at a time, the speedup was 80%.
The benchmark was compiled with -O3 and stderr was piped to /dev/null.

These were the headers used for the benchmark:
```
  header input1[] = {
    {"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA", "111111111111111111111111"},
    {"BBB",                               "222"                     },
    {"CCC",                               "333"                     },
    {"DDD",                               "333"                     },
    {"EEE",                               "333"                     },
    {"FFF",                               "333"                     },
    {"GGG",                               "333"                     },
    {"HHH",                               "333"                     },
    {"III",                               "333"                     },
    {"JJJ",                               "333"                     },
    {"KKK",                               "333"                     },
  };
```

This fixes https://github.com/apache/trafficserver/issues/11431.

### DIFF
--- a/include/proxy/http/HttpTransact.h
+++ b/include/proxy/http/HttpTransact.h
@@ -69,8 +69,8 @@ s_dump_header(HTTPHdr const *hdr, std::string &out)
     // the use of tmp is precautionary to make sure our logic
     // doesn't break in case it does change in some circumstance.
     int tmp{offset};
-    done     = hdr->print(b, 4096, &used, &tmp);
-    offset  += used;
+    done    = hdr->print(b, 4096, &used, &tmp);
+    offset += used;
     out.append(b, used);
   } while (0 == done);
 }
--- a/include/proxy/http/HttpTransact.h
+++ b/include/proxy/http/HttpTransact.h
@@ -69,17 +69,16 @@ s_dump_header(HTTPHdr const *hdr, std::string &out)
     // the use of tmp is precautionary to make sure our logic
     // doesn't break in case it does change in some circumstance.
     int tmp{offset};
-    done     = hdr->print(b, 4095, &used, &tmp);
+    done     = hdr->print(b, 4096, &used, &tmp);
     offset  += used;
-    b[used]  = '\0';
-    out.append(b);
+    out.append(b, used);
   } while (0 == done);
 }
 
 inline void
 dump_header(DbgCtl const &ctl, HTTPHdr const *hdr, std::int64_t sm_id, std::string_view description)
 {
-  if (true) {
+  if (ctl.on()) {
     std::string output;
     output.append("+++++++++ ");
     output.append(description);
--- a/include/proxy/http/HttpTransact.h
+++ b/include/proxy/http/HttpTransact.h
@@ -88,6 +88,8 @@ dump_header(DbgCtl const &ctl, HTTPHdr const *hdr, std::int64_t sm_id, std::stri
     output.push_back('\n');
     if (hdr->valid()) {
       s_dump_header(hdr, output);
+    } else {
+      output.append("Invalid header!\n");
     }
     // We make a single call to fprintf so that the output does not get
     // interleaved with output from other threads performing I/O.
--- a/include/proxy/http/HttpTransact.h
+++ b/include/proxy/http/HttpTransact.h
@@ -25,6 +25,7 @@
 
 #include <cstddef>
 
+#include "tsutil/DbgCtl.h"
 #include "tscore/ink_assert.h"
 #include "tscore/ink_platform.h"
 #include "iocore/hostdb/HostDB.h"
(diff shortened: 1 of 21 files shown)


include/proxy/http/HttpTransact.h
```diff
@@ -45,31 +46,55 @@
...
+    // The buffer offset is taken non-const and it is apparently
+    // modified in some code path, but in my testing it does
+    // not change, it seems. Since we manually bump the offset,
+    // the use of tmp is precautionary to make sure our logic
+    // doesn't break in case it does change in some circumstance.
+    int tmp{offset};
+    done     = hdr->print(b, 4095, &used, &tmp);
+    offset  += used;
+    b[used]  = '\0';
+    out.append(b);
```
bneradt, line 75: A small thing, but rather than null terminate `b` and append it, `std::string::append` takes a length which handles this for us:

```cpp
    done     = hdr->print(b, 4095, &used, &tmp);
    offset  += used;
    out.append(b, used);
```

include/proxy/http/HttpTransact.h
```diff
@@ -45,31 +46,55 @@
...
+  } while (0 == done);
+}
+
+inline void
+dump_header(DbgCtl const &ctl, HTTPHdr const *hdr, std::int64_t sm_id, std::string_view description)
+{
+  if (true) {
```
bneradt, line 82:
```cpp
  if (true) {
```

JosiahWI:
LSan detected a memory leak in the ImageMagick plugin.

JosiahWI:
[approve ci rocky]

src/api/InkAPI.cc
```diff
@@ -1518,17 +1518,18 @@ TSMimeHdrCopy(TSMBuffer dest_bufp, TSMLoc dest_obj, TSMBuffer src_bufp, TSMLoc s
 void
 TSMimeHdrPrint(TSMBuffer bufp, TSMLoc obj, TSIOBuffer iobufp)
 {
+  // The bufp argument is no longer used, but the assert is left in to verify
+  // the preconditon so that the API contract is not inadvertantly changed.
   sdk_assert(sdk_sanity_check_mbuffer(bufp) == TS_SUCCESS);
```
bneradt, line 1523: Adding @zwoop to get his input. @JosiahWI's work here cleans up things and removes this unused `bufp` parameter. Well, it becomes unused after this cleanup. Do we want to change the `TSMimeHdrPrint` accordingly? Maybe add a new function without `bufp` with a deprecation on this old one with the intention have b removed in the 11 release?

Regardless, we do not need to hold up this PR. We can handle that as a separate PR.

bneradt (dismissed):
Looks good. It's a big bonus that this makes things more efficient when debug logging is on.

JosiahWI:
The regression test PARENT_SELECTION failed.

JosiahWI:
[approve ci fedora]

bneradt (commented):
Looks good to me. I'll let a non-Yahoo reviewer chip in.

include/proxy/http/HttpTransact.h
```diff
@@ -45,31 +46,54 @@
...
+    offset += used;
+    out.append(b, used);
+  } while (0 == done);
+}
+
+inline void
+dump_header(DbgCtl const &ctl, HTTPHdr const *hdr, std::int64_t sm_id, std::string_view description)
+{
+  if (ctl.on()) {
+    std::string output;
```
zwoop, line 82: I don' know what we did before, but do we want to dump the "header" here (including the SM ID) if the header isn't valid ? I don't know why it'd be valid, but we do check for that.

src/api/InkAPI.cc
zwoop: If we can get this in for 10.0.0 @cmcfarlen ?) I'd say fix the API now and break compatibility. This is really just for debugging as I understand things, so this is not going to break any thing critical. Obviously fix any plugins in the core accordingly.

That change would need to go to the mailing list as well, but we're certainly at liberty to break compatibility for 10.0.0. And again, this doesn't seem like an API worthy of making two versions of :).

include/proxy/http/HttpTransact.h
JosiahWI: I‘d like to make that PR separately, since it constitutes a non-bugfix behavioral change.

cmcfarlen (approved):

JosiahWI:
Rebased and did the changes requested by @zwoop.

bneradt (commented):
Looks good to me.

JosiahWI:
The AuTest bad_chunked_encoding failed with a _very_ interesting message:

```
        Reason: Contents of /tmp/sandbox/bad_chunked_encoding/_output/bad_chunked_encoding-verifier-server2/stream.stdout.txt contained expression
     file /tmp/sandbox/bad_chunked_encoding/_output/bad_chunked_encoding-verifier-server2/stream.stdout.txt : Verify that the body never got through. - Failed
        Reason: Contents of /tmp/sandbox/bad_chunked_encoding/_output/bad_chunked_encoding-verifier-server2/stream.stdout.txt contains expression: "abc"
           Details:
             Via: http/1.1 traffic_server[9ecabccd-0556-4866-b183-7e2b1407c2fd] (ApacheTrafficServer/10.1.0) : 17
             Via: http/1.1 traffic_server[9ecabccd-0556-4866-b183-7e2b1407c2fd] (ApacheTrafficServer/10.1.0) : 35
             Via: http/1.1 traffic_server[9ecabccd-0556-4866-b183-7e2b1407c2fd] (ApacheTrafficServer/10.1.0) : 53
             Via: http/1.1 traffic_server[9ecabccd-0556-4866-b183-7e2b1407c2fd] (ApacheTrafficServer/10.1.0) : 71
             Via: http/1.1 traffic_server[9ecabccd-0556-4866-b183-7e2b1407c2fd] (ApacheTrafficServer/10.1.0) : 95
             Via: http/1.1 traffic_server[9ecabccd-0556-4866-b183-7e2b1407c2fd] (ApacheTrafficServer/10.1.0) : 119
 ```

JosiahWI:
[approve ci autest]

zwoop (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11438, opened 2024-06-10
Fail earlier if Netcat is not installed

JosiahWI:
Some AuTests require Netcat as a client, but do not get skipped if it is missing. ~~This updates the tests so that they get skipped if Netcat is not installed.~~

EDIT: This updates CMakeLists.txt and autest.sh to fail appropriately if Netcat is missing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -515,6 +515,7 @@ find_package(resolv)
 if(ENABLE_DOCS OR ENABLE_AUTEST)
   find_package(Python3 REQUIRED)
   find_program(PipEnv pipenv REQUIRED)
+  find_program(NETCAT_PROGRAM nc REQUIRED)
 endif()
 
 if(ENABLE_DOCS)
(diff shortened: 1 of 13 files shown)


bneradt (commented):
Looks good to me. I double-checked that each of these does indeed use `nc` or calls a script that does.

As usual, I'll let a non-Yahoo reviewer approve.

JosiahWI:
We may not want to merge this because it would cause CI to pass without indicating an issue if the `nc` command were missing, and we need those tests to run.

JosiahWI:
The AuTest ja3_fingerprint failed.

JosiahWI:
[approve ci autest]

JosiahWI:
I've repurposed the PR to have the CMake build and autest.sh check for netcat and fail if it is missing, rather than skipping tests based on its presence.

bneradt (commented):
Looks good. Thanks for giving this `nc` requirement explicit messaging.

cmcfarlen (approved):

PR #11441, opened 2024-06-11
Update magick plugin to OpenSSL 1.1.1 API

JosiahWI:
The functions `EVP_MD_CTX_create` and `EVP_MD_CTX_destroy` were renamed to `EVP_MD_CTX_new` and `EVP_MD_CTX_free` in OpenSSL 1.1.0. This renames them in the magick plugin accordingly. This also replaces the individual calls to `EVP_DigestVerifyUpdate` and `EVP_DigestVerifyFinal` with a single call to `EVP_DigestVerify` which is available in OpenSSL 1.1.1.

### DIFF
--- a/plugins/experimental/magick/magick.cc
+++ b/plugins/experimental/magick/magick.cc
@@ -237,12 +237,13 @@ verify(const byte *const msg, const size_t mlen, const byte *const sig, const si
     return false;
   }
 
-  const int rc = EVP_DigestVerify(evp.context, sig, slen, msg, mlen);
-  if (1 == rc) {
+  if (const int rc = EVP_DigestVerify(evp.context, sig, slen, msg, mlen); 1 == rc) {
     return true;
-  } else if (0 == rc) {
-    return false;
   } else {
+    // The OpenSSL 1.1.1 API distinguishes between a verification failure and a
+    // more serious error with a specific return value for the former, but we
+    // have collapsed them into one case because we don't do any special
+    // handling for serious errors.
     ssl_error();
     return false;
   }
--- a/plugins/experimental/magick/magick.cc
+++ b/plugins/experimental/magick/magick.cc
@@ -203,6 +203,20 @@ struct EVPKey {
   }
 };
 
+/** Remove the last error from this thread's error queue and print it.
+ */
+static void
+ssl_error()
+{
+  if (unsigned long error_code{ERR_get_error()}; 0 != error_code) {
+    if (char const *reason{ERR_lib_error_string(error_code)}; NULL == reason) {
+      Dbg(dbg_ctl, "SSL error: error code %lu", error_code);
+    } else {
+      Dbg(dbg_ctl, "SSL error: %s", reason);
+    }
+  }
+}
+
 bool
 verify(const byte *const msg, const size_t mlen, const byte *const sig, const size_t slen, EVP_PKEY *const pkey)
 {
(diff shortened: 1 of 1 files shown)


JosiahWI:
LSan detected a leak in the magick plugin and the AuTest prefetch_simple failed:

```
   Run: 2-tr: Failed
     Starting TestRun 2-tr : No Issues found - Passed
        Reason: Started!
     Process: Default: Failed
       Test : Checking that ReturnCode == 0 - Passed
          Reason: Returned Value: 0 == 0
       file /tmp/sandbox/prefetch_simple/_output/2-tr-Default/stream.stdout.txt : Checking that /tmp/sandbox/prefetch_simple/_output/2-tr-Default/stream.stdout.txt matches prefetch_simple.gold - Failed
          Reason: File differences
           Gold File : /home/jenkins/workspace/Github_Builds/autest/src/tests/gold_tests/pluginTest/prefetch/prefetch_simple.gold
           Data File : /tmp/sandbox/prefetch_simple/_output/2-tr-Default/stream.stdout.txt
             GET http://domain.in/texts/demo-1.txt HTTP/1.1
             GET http://domain.in/texts/demo-2.txt HTTP/1.1
             GET http://domain.in/texts/demo-3.txt HTTP/1.1
           - GET http://domain.in/texts/demo-4.txt HTTP/1.1
```

JosiahWI:
[approve ci rocky autest]

plugins/experimental/magick/magick.cc
```diff
@@ -226,19 +226,17 @@ verify(const byte *const msg, const size_t mlen, const byte *const sig, const si
...
-
   ERR_clear_error();
 
   {
-    const int rc = EVP_DigestVerifyFinal(evp.context, sig, slen);
-    return 1 == rc;
+    const int rc = EVP_DigestVerify(evp.context, sig, slen, msg, mlen);
+    if (1 == rc) {
+      return true;
+    } else if (0 == rc) {
```
maskit, line 235: If we check this case let's print debug log.

JosiahWI: Is my use of `Dbg` for this purpose good? I wonder whether we should do this for the API call to load the RSA certificate as well. The current implementation does not even handle such a failure in production. It continues to operate without the cert, and will probably cause everything to fail to verify, if it doesn't do something worse.

JosiahWI:
LSan detected a leak in the magick plugin again.

JosiahWI:
[approve ci rocky]

plugins/experimental/magick/magick.cc
JosiahWI: It does debug asserts and it's probably assuming that it should never fail to load its RSA certificate.

JosiahWI:
[approve ci rocky]

JosiahWI:
[approve ci rocky]

bneradt (commented):
Looks reasonable to me.

plugins/experimental/magick/magick.cc
maskit: It just looked like doing exactly same thing as the `else` clause doesn't make sense. Current code still makes me wonder why the error is silently ignored in this case. Leaving a note/comment on the source code may be enough. I'm also fine with letting the `else` clause handle alll the error case.

maskit (approved):

bneradt (approved):

PR #11433, opened 2024-06-10
Change remap ACL matching policy to 'Match on IP and Method'

masaori335:
Introduce a new config `proxy.config.url_remap.acl_matching_policy` to change the behavior of remap ACL filters.

- `0` (default) employs the "first explicit match wins" policy, this means ATS continue processing ACL filters if it's implicit match and stop processing only if it's explicit match.
- `1` employs "first any match wins" policy, this means the first ACL filter is used if it exists. Please note that action is flipped  for implicit matches.

For AuTest
- Tests in the `remap_acl.test.py` configured to use "first any match wins" policy to minimize change for now.
- `remap_acl_2.test.py` covers all cases. ( I grabbed code from #11421 )

If all we agree with this approach, I'll add docs and cleanup AuTests.

----

 Co-authored-by: @mattyw

PR #11448, opened 2024-06-14
Doc: Clarification of ACL filters

masaori335:
Describe ACL filters with https://github.com/apache/trafficserver/pull/11433. Separated PR to do not block the PR by documentation change.


doc/admin-guide/files/remap.config.en.rst
```diff
@@ -518,20 +514,95 @@ mapping rules. (It is activated before any mappings and is never
...
    map ...
    .activatefilter ip_allow
 
-Note this entirely disables IP Allow checks for those remap rules.
+Note this entirely disables :file:`ip_allow.yaml` checks for those remap rules.
+
+Evaluation Order and Matching Policy
+------------------------------------
+
+ATS evaluates multiple ACL filters in below order.
```
mattyw, line 544:
```suggestion
ATS evaluates multiple ACL filters in the following order:
```

doc/admin-guide/files/remap.config.en.rst
```diff
@@ -518,20 +514,95 @@ mapping rules. (It is activated before any mappings and is never
...
+
+1. Special "deny all" filter in :file:`ip_allow.yaml`
+2. In-line Filter in :file:`remap.config`
+3. Named Filter in :file:`remap.config`
+4. Filters in :file:`ip_allow.yaml`
+
+When a ACL filter is found, ATS stop processing ACL filters depending on mathcing policy configured by
```
mattyw, line 551:
```suggestion
When an ACL filter is found, ATS stops processing ACL filters depending on the matching policy configured by
```

mattyw (commented):
Docs look great @masaori335. I've only added nitpick suggestions

doc/admin-guide/files/remap.config.en.rst
```diff
@@ -518,20 +514,95 @@ mapping rules. (It is activated before any mappings and is never
...
 second mapping.
 
-Implict IPAllow filter
-======================
+Special Filter and ip_allow Named Filter
+----------------------------------------
+
+If :file:`ip_allow.yaml` has a "deny all" filter, it is treated as a special filter that is applied before remapping for
+optimizaion. To control this for specific remap rules, a named filter called `ip_allow` is pre-defined. This named filter is
+activated implicitly in default. To stop applying the special rule, disable the `ip_allow` filter as shown below.
```
zwoop, line 522: Should that be "by default" ?

doc/admin-guide/files/remap.config.en.rst
```diff
@@ -518,20 +514,95 @@ mapping rules. (It is activated before any mappings and is never
...
+3. Named Filter in :file:`remap.config`
+4. Filters in :file:`ip_allow.yaml`
+
+When an ACL filter is found, ATS stops processing ACL filters depending on the mathcing policy configured by
+:ts:cv:`proxy.config.url_remap.acl_matching_policy`.
+
+First Explicit Match Wins Policy
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This is the default matching policy. ATS stops processing ACL filters only if an explicit match of IP address and HTTP method is
```
zwoop, line 557: But it can be an explicit match without an IP right? Like I can match on method only? And that can still match right?

masaori335: Right. If there're no `@src_ip`, it's the same as `@src_ip=ALL` that matches with any IP.

doc/admin-guide/files/remap.config.en.rst
masaori335: Do you mean "deactivate" the ip_allow named filter by default? 

A tricky behavior is if this is deactivated, step 1 and 4 is disabled. Should we disable only step 1  by default?

> 1. Special "deny all" filter in :file:`ip_allow.yaml`
> 2. In-line Filter in :file:`remap.config`
> 3. Named Filter in :file:`remap.config`
> 4. Filters in :file:`ip_allow.yaml`

doc/admin-guide/files/remap.config.en.rst
```diff
@@ -477,7 +475,7 @@ If no IP address is specified for ``@src_ip``, ``@src_ip_category``, or
 can be explicitly stated with ``@src_ip=all``.
 
 Named Filters
-=============
+-------------
```
bneradt, line 478: Good fixes for these.

doc/admin-guide/files/remap.config.en.rst
```diff
@@ -518,20 +514,95 @@ mapping rules. (It is activated before any mappings and is never
...
+1. Special "deny all" filter in :file:`ip_allow.yaml`
+2. In-line Filter in :file:`remap.config`
+3. Named Filter in :file:`remap.config`
+4. Filters in :file:`ip_allow.yaml`
+
+When an ACL filter is found, ATS stops processing ACL filters depending on the mathcing policy configured by
+:ts:cv:`proxy.config.url_remap.acl_matching_policy`.
+
+First Explicit Match Wins Policy
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
bneradt, line 555: I appreciate what you're trying to say here, but distinguishing the two behaviors as "explicit" and "implicit" is probably not helpful to the user. Nor do we want to say that the one is explicit, and the other is first match wins: if you think about it, both only apply to the first match. The real difference between these is what the match criteria is:

* In the new way added in 10, the ACL filters match based upon IP only, as they do for ip_allow.yaml.
* In the old way they match differently than ip_allow: they have to match both IP and method. 

I think this will be the most helpful framework for the user to understand and choose between the two policies. Thus I suggest something like this in these sections:

```
Match on IP and Method Policy
~~~~~~~~~~~~~~~~~~~~~~~

With this policy, ACL filters, named or inline, only take effect if both
IP and method match the incoming request. This policy is useful for
organizations that want ACL rules to additively allow or deny specific
methods in addition to other ACL filters and ip_allow.yaml rules.
Consider a filter like the following:

map http://www.example.com/ http://internal.example.com/ @action=deny @method=POST

The implicit @src_ip is all client IP addresses, so this filter will
match on any POST request matched by this remap rule from any client and
its action will be to deny such POST requests. For all other methods,
the filter will not take effect, thus allowing other active ACL filters
or an ip_allow.yaml rule to determine the action to take for any other
transaction.

...

Match on IP Policy 
~~~~~~~~~~~~~~

With this policy, ACL filters match solely based upon IP, meaning that
ACL filters match like ip_allow.yaml rules. This policy is useful for
organizations that want to have ACL filters behave like ip_allow rules
specific to remap targets. Thus, consider a filter like the following:

map http://www.example.com/ http://internal.example.com/ @action=allow @method=GET @method=HEAD

The implicit @src_ip is all client IP address, so this filter will apply
to all transactions matching this remap rule. Again, like an analogously
crafted ip_allow allow rule, this will allow ``GET`` and ``HEAD`` requests
while denying all other methods to the ``internal.example.com`` origin.
No other ACL filters or ip_allow rules will apply for this target.

This policy is new to ATS 10.

```

doc/admin-guide/files/records.yaml.en.rst
```diff
@@ -3545,6 +3545,18 @@ URL Remap Rules
...
+.. ts:cv:: CONFIG proxy.config.url_remap.acl_matching_policy INT 0
+   :reloadable:
+
+   This controls matching policy of ACL filters in :file:`remap.config`. See :ref:`acl-filters` for more details.
+
+   ===== =========================
+   Value Description
+   ===== =========================
+   ``0`` First Explicit Match Wins
+   ``1`` First Filter Wins
```
bneradt, line 3557: Given my below comments, I would describe these as:

```
``0`` ACL filters match on both IP and method.
``1`` ACL filters match on IP.
```

doc/admin-guide/files/remap.config.en.rst
```diff
@@ -518,20 +514,95 @@ mapping rules. (It is activated before any mappings and is never
...
+
+====== ================ ================ ================ ================ =============
+Method In-line Filter   Named Filter 1   Named Filter 2   ip_allow.yaml    result
+====== ================ ================ ================ ================ =============
+GET    deny  (explicit) \-               \-               \-               denied  (403)
+HEAD   allow (implicit) deny  (explicit) \-               \-               denied  (403)
+POST   allow (implicit) allow (implicit) allow (explicit) \-               allowed (200)
+PUT    allow (implicit) allow (implicit) deny  (implicit) deny  (explicit) denied  (403)
+DELETE allow (implicit) allow (implicit) deny  (implicit) allow (implicit) allowed (200)
+====== ================ ================ ================ ================ =============
```
bneradt, line 588: I really like these tables you made. This will be very helpful to users. I'm sorry if I come across as harping on the explicit/implicit terminology, but I don't prefer it here either, but for different reasons than the policy naming. For this "match on ip and method" policy we don't want to say, that the inline filter implicitly allows HEAD requests, for example. It does not in fact allow them, but rather leaves it to other filters or rules to decide. It's more accurate to say it doesn't apply to HEAD requests. How about this tweak?

```
====== ================ ================ ================ ================ =============
Method In-line Filter   Named Filter 1   Named Filter 2   ip_allow.yaml    result
====== ================ ================ ================ ================ =============
GET    deny             \-               \-               \-               denied  (403)
HEAD   N/A              deny             \-               \-               denied  (403)
POST   N/A              N/A              allow            \-               allowed (200)
PUT    N/A              N/A              N/A              deny             denied  (403)
DELETE N/A              N/A              N/A              allow            allowed (200)
====== ================ ================ ================ ================ =============
```

doc/admin-guide/files/remap.config.en.rst
```diff
@@ -518,20 +514,95 @@ mapping rules. (It is activated before any mappings and is never
...
+
+====== ================ ================ ================ ================ =============
+Method In-line Filter   Named Filter 1   Named Filter 2   ip_allow.yaml    result
+====== ================ ================ ================ ================ =============
+GET    deny  (explicit) \-               \-               \-               denied  (403)
+HEAD   allow (implicit) deny  (explicit) \-               \-               allowed (200)
+POST   allow (implicit) \-               allow (explicit) \-               allowed (200)
+PUT    allow (implicit) \-               \-               deny  (explicit) allowed (200)
+DELETE allow (implicit) \-               \-               \-               allowed (200)
+====== ================ ================ ================ ================ =============
```
bneradt, line 604: I suggest:

```

If the IP match policy is applied, then given the example remap rule
above, all requests to ``internal.example.com`` are filtered like an
ip_allow rule applies to them that denies GET requests and allows all
other methods:

====== ================ ================ ================ ================ =============
Method In-line Filter   Named Filter 1   Named Filter 2   ip_allow.yaml    result
====== ================ ================ ================ ================ =============
GET    deny             \-               \-               \-               denied  (403)
HEAD   allow            \-               \-               \-               allowed (200)
POST   allow            \-               \-               \-               allowed (200)
PUT    allow            \-               \-               \-               allowed (200)
DELETE allow            \-               \-               \-               allowed (200)
====== ================ ================ ================ ================ =============
```

bneradt (changes requested):
Good docs. The example rules and filters with the table will be helpful to users.

doc/admin-guide/files/remap.config.en.rst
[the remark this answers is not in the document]
masaori335: Sounds like we need to talk which config should be default. 

I'd prefer compatible with old ATS versions. New behavior as default seems too painful on upgrading to me.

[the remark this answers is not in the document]
bneradt: Adding the configuration shouldn't be too bad, right? There's already configuration change requirements involved with ATS 10. We changed the entire format of the file to YAML. :) I can't image setting this to the non-default value is unreasonable.

We need to rather decide which is the proper default for users and make that the default.

---

While writing this comment, the thought occurs to me that maybe we should take a slightly different approach to this. Maybe it makes sense for us to build on top of your configuration switch and make it work like this:

* **Legacy ACL matching.** This keeps ACL rule matching the way ATS 9 and prioror matches.
* **Enhanced ACL filters.** With this enabled, `@action=deny` and `@action=allow` function like ip_allow.yaml rules (i.e., the way my change makes them). But in addition to those actions, we can add `@action=add_deny` and `@action=add_allow`. These would allow users to add specific methods to deny or allow lists, like ATS 9 behaves currently behaves with `@action=deny` and `@action=allow`. This way enhanced offers a super set of functionality.

Going this route, we would make the "enhanced" feature the default because it is unambiguously superior. But since it is backwards incompatible, the legacy switch is added (non-default) to ease transition to ATS 10.

This will also make things easier for users to transition to the new behavior because remap.config would simply be changed with a few sed commands to transition `@action=deny` to `@action=add_deny` and `@action=allow` to `@action=add_allow`.

doc/admin-guide/files/remap.config.en.rst
```diff
@@ -518,20 +514,95 @@ mapping rules. (It is activated before any mappings and is never
...
 
-Note this entirely disables IP Allow checks for those remap rules.
+Note this entirely disables :file:`ip_allow.yaml` checks for those remap rules.
+
+Evaluation Order and Matching Policy
+------------------------------------
+
+ATS evaluates multiple ACL filters in the following order:
+
+1. Special "deny all" filter in :file:`ip_allow.yaml`
```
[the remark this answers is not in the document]
masaori335: Add a note at the end of this subsection.

doc/admin-guide/files/remap.config.en.rst
```diff
@@ -518,20 +514,95 @@ mapping rules. (It is activated before any mappings and is never
...
+   # ip_allow.yaml
+   ip_allow:
+      - apply: in
+        ip_addrs: 198.51.100.0/24
+        action: deny
+        method: PUT
+
+   # remap.config
+   .definefilter named-filter-1 @action=deny @method=HEAD
+   .definefilter named-filter-2 @actio==allow @method=POST
```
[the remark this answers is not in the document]
masaori335: Nice catch! Fixed.

doc/admin-guide/files/remap.config.en.rst
masaori335: PTAL

doc/admin-guide/files/remap.config.en.rst
masaori335: PTAL. I left specified actions in subsequent filters to high light they're ignored.

doc/admin-guide/files/remap.config.en.rst
masaori335: Beside the naming of the policies, descriptions are clear. I'd use the same example to clarify the differences.

doc/admin-guide/files/remap.config.en.rst
masaori335: With Brians's comment, the new description clarifies the absence of `@src_ip` matches with any IP.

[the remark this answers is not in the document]
jasmine-nahrain: I would suggest keeping the default behaviour the same as 9.x only because the human errors that could occur by changing the default are pretty high. Changing the default would make the assumption administrates will read the new docs thoroughly and update their configs to replicate the same behaviour. 

If we implicitly allow methods that are not specified, we are creating behaviour that is not visible to the administrators. They may not think like a software engineer and investigate in a way that any of us would which leaves big gaps of vulnerabilities that could open up. 

I think keeping the newer behaviour as an optional enhancement for the right people to try would make more sense. They would be updating their configs knowing the implications of the change rather than just trying to make it all run again.

doc/admin-guide/files/remap.config.en.rst
masaori335: After I considered naming for a while, I start calling them "Match on IP and Method Policy" and "Match on IP only Policy".

doc/admin-guide/files/remap.config.en.rst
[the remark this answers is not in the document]
bneradt: Thanks for the thoughts, @masaori335 and @jasmine-nahrain .

I would be content with:
1. Make it the default for ATS 10, but deprecate it.
2. Remove it for ATS 11

Does that sound fine?

bneradt (approved):
Looks great. Thanks for adding these docs. We've talked offline about other ideas, but let's move forward with what you have here.

Thank you for adding the "more realistic" example to the match on ip only section. That worked out well.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11450, opened 2024-06-14
Fix use of uninitialized stack memory in records

### LINKED ISSUE #11449: Records parses uninitialized data for IP addresses
This issue was detected by valgrind.

This is the relevant part of the code, with comments added to show where things go wrong:
```cpp
void
RecHttpLoadIpAddrsFromConfVar(const char *value_name, swoc::IPRangeSet &addrs)
{
  // We declare a fixed-size buffer to load the config content into. Good so far.
  char value[1024];

  // We load the config content into the buffer. It may not fill the whole buffer, but we don't know the
  // actual content length. The unused part of the buffer is uninitialized.
  if (REC_ERR_OKAY == RecGetRecordString(value_name, value, sizeof(value))) {
    Debug("config", "RecHttpLoadIpAddrsFromConfVar: parsing the name [%s] and value [%s]", value_name, value);
    // We load the value into a swoc::TextView. This TextView constructor uses the array length as the buffer size...
    // unfortunately, we don't really mean to use the whole buffer.
    // EDIT: A previous version of this comment said the resulting buffer may be missing a null terminator, but this was false.
    swoc::TextView text(value);
    while (text) {
      auto token = text.take_prefix_at(',');
      // Now we start parsing for IPs... if the config content is empty we still parse the entire buffer. If it were to
      // hypothetically contain a leftover IP address range somehow we would likely load it.
      if (swoc::IPRange r; r.load(token)) {
        Debug("config", "RecHttpLoadIpAddrsFromConfVar: marking the value [%.*s]", int(token.size()), token.data());
        addrs.mark(r);
      }
    }
  }
  Debug("config", "RecHttpLoadIpMap: parsed %zu IpMap entries", addrs.count());
}
```

JosiahWI:
Fixes #11449.

~~This adds an overload of `RecGetRecordString` that takes an out-parameter for the size of the content copied into the out buffer. It updates `RecHTTPLoadIpAddrsFromConfVars` to use this parameter to get the buffer size to use when initializing the TextView for parsing, so that we only parse the actual content and ignore the uninitialized part of the buffer.~~

This passes the length of the value read to the TextView constructor so that it only parses the actual content and ignores the uninitialized part of the buffer.

### DIFF
--- a/src/records/RecHttp.cc
+++ b/src/records/RecHttp.cc
@@ -33,6 +33,7 @@
 #include "tscore/ink_inet.h"
 #include "swoc/BufferWriter.h"
 #include "swoc/bwf_ip.h"
+#include <cstring>
 #include <string_view>
 #include <unordered_set>
 
@@ -150,7 +151,7 @@ RecHttpLoadIpAddrsFromConfVar(const char *value_name, swoc::IPRangeSet &addrs)
 
   if (REC_ERR_OKAY == RecGetRecordString(value_name, value, sizeof(value))) {
     Debug("config", "RecHttpLoadIpAddrsFromConfVar: parsing the name [%s] and value [%s]", value_name, value);
-    swoc::TextView text(value);
+    swoc::TextView text(value, std::strlen(value));
     while (text) {
       auto token = text.take_prefix_at(',');
       if (swoc::IPRange r; r.load(token)) {
--- a/src/records/RecHttp.cc
+++ b/src/records/RecHttp.cc
@@ -146,11 +146,12 @@ RecHttpLoadIp(char const *name)
 void
 RecHttpLoadIpAddrsFromConfVar(const char *value_name, swoc::IPRangeSet &addrs)
 {
-  char value[1024];
+  char        value[1024];
+  std::size_t value_size{};
 
-  if (REC_ERR_OKAY == RecGetRecordString(value_name, value, sizeof(value))) {
+  if (REC_ERR_OKAY == RecGetRecordString(value_name, value, sizeof(value), value_size)) {
     Debug("config", "RecHttpLoadIpAddrsFromConfVar: parsing the name [%s] and value [%s]", value_name, value);
-    swoc::TextView text(value);
+    swoc::TextView text(value, value_size);
     while (text) {
       auto token = text.take_prefix_at(',');
       if (swoc::IPRange r; r.load(token)) {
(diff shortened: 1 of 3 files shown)


JosiahWI:
[approve ci rocky]

bneradt (commented):
Looks good to me. I'll let a non-Yahoo review and approve.

JosiahWI:
This fix is incorrect and introduces a buffer overflow vulnerability because `strlcpy` returns the total length of the source string, not the destination string.

JosiahWI:
[approve ci rat]

JosiahWI:
Should be good now.

JosiahWI:
[approve ci]

src/records/RecHttp.cc
```diff
@@ -150,7 +151,7 @@ RecHttpLoadIpAddrsFromConfVar(const char *value_name, swoc::IPRangeSet &addrs)
 
   if (REC_ERR_OKAY == RecGetRecordString(value_name, value, sizeof(value))) {
     Debug("config", "RecHttpLoadIpAddrsFromConfVar: parsing the name [%s] and value [%s]", value_name, value);
-    swoc::TextView text(value);
+    swoc::TextView text(value, std::strlen(value));
```
bneradt, line 154: Probably better to favor `strnlen(value, 1024)`.

bneradt (changes requested):
Thank you for finding this and working on it.

src/records/RecHttp.cc
JosiahWI: Can we use this even though it's part of the C standard and not part of the C++ standard? https://en.cppreference.com/w/cpp/string/byte/strlen

The rest of records uses `strlen(str)` after calling `RecGetRecordString` and this is safe because it's guaranteed to always null-terminate the buffer.

bneradt: True. `strlen` should be fine.

ezelkow1 (approved):

ywkaras:
Both GCC and CLang seem to be very good at evaluation strlen for string constants at runtime:  https://godbolt.org/z/fqnd5Pb4n .
So, we should change TextView and remove the complicated attempts to avoid calling strlen.

ywkaras:
The template overload for C string literals doesn't seem to work anyway:  https://godbolt.org/z/5cTWxYq9v

cmcfarlen:
Cherry-picked to v10.0.x for #11464

PR #11451, opened 2024-06-15
Make NetDbg available on releaes builds

maskit:
Autest quic_poll_timeout fails on release builds because the test relies on the debug output.


maskit:
@ywkaras @zwoop My understanding is that the debug tag check is faster than before and this change does not affect performance as long as the debug tags are disabled.

JosiahWI:
[approve ci rocky]

ywkaras (approved):

PR #11454, opened 2024-06-17
records.yaml -  Rename 'ts' yaml tag to 'records'.

brbzull0:
This apply to `records.yaml` and `conf_remap` plugin that uses input files.


zwoop (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11459, opened 2024-06-18
Clarify VCONN_CLOSE serialization in developer docs

JosiahWI:
Plugins may want to use both TLS and HTTP hooks without a mutex. An example of this is the ja3_fingerprint, which has a possible race condition between the `VCONN_CLOSE` hook and HTTP hooks according to the current documentation. However, these events happen on the same thread and are thus serialized. This updates the documentation to clarify that the current behavior is guaranteed.


JosiahWI:
[approve ci rocky]

ezelkow1 (approved):

PR #11461, opened 2024-06-19
Fix pipelined chunked parsing.

bneradt:
For chunked bodies, we overly aggressively consumed bytes from the client side reader buffer, causing us to drop a potentially future pipelined request after the chunked body. This fixes how we consume chunked body bytes so that a pipelined request is not dropped.

---

Making a draft for now while I figure out how to address the buffered request use case that the `body_buffer` test fails on.


bneradt:
A separate future PR will address this.

PR #11466, opened 2024-06-21
Fix unused function parameters in the plugins functionality

freak82:
Use `/* name ATS_UNUSED */` for fully unused parameters.
Use `[[maybe_unused]]` attribute for conditionally unused parameters used in asserts.

This pull request is part of the effort for [removing](https://github.com/apache/trafficserver/issues/11377 ) the `-Wno-unused-parameter` warning suppression.

PR #11467, opened 2024-06-21
Update access_control to OpenSSL 1.1.1 HMAC API

JosiahWI:
This updates the `cryptoMessageDigestGet` implementation for OpenSSL 1.1.1 ~~and removes the unused `outLen` parameter~~.

### DIFF
--- a/plugins/experimental/access_control/utils.cc
+++ b/plugins/experimental/access_control/utils.cc
@@ -194,11 +194,13 @@ cryptoErrStr(char *buffer, size_t bufferLen)
  * @param dataLen message length
  * @param key ptr to a counted string containing the key (secret)
  * @param keyLen key length
- * @param out ptr to where to store the digest (must be at least MAX_MSDIGEST_BUFFER_SIZE)
+ * @param out ptr to where to store the digest
+ * @param outLen length of the out buffer (must be at least MAX_MSGDIGEST_BUFFER_SIZE)
  * @return the number of character actually written to the buffer.
  */
 size_t
-cryptoMessageDigestGet(const char *digestType, const char *data, size_t dataLen, const char *key, size_t keyLen, char *out)
+cryptoMessageDigestGet(const char *digestType, const char *data, size_t dataLen, const char *key, size_t keyLen, char *out,
+                       size_t /* outLen ATS_UNUSED */)
 {
   const EVP_MD *md  = nullptr;
   unsigned int  len = 0;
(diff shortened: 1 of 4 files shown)


JosiahWI:
Waiting to rebase this on #11466.

JosiahWI:
The `HAVE_HMAC_CTX_NEW` define can't be removed yet because the s3_auth plugin also uses it. After that plugin is updated it can be removed.

plugins/experimental/access_control/utils.cc
```diff
@@ -217,32 +209,10 @@ cryptoMessageDigestGet(const char *digestType, const char *data, size_t dataLen,
...
-    AccessControlError("failed to create EVP message digest context: %s", cryptoErrStr(buffer, sizeof(buffer)));
-    goto err;
-  }
-
-  if (!HMAC_Update(ctx, reinterpret_cast<const unsigned char *>(data), dataLen)) {
-    AccessControlError("failed to update the signing hash: %s", cryptoErrStr(buffer, sizeof(buffer)));
-    goto err;
+  if (!HMAC(md, reinterpret_cast<const unsigned char *>(key), keyLen, reinterpret_cast<const unsigned char *>(data), dataLen,
+            reinterpret_cast<unsigned char *>(out), &len)) {
+    AccessControlError("failed to get the signing hash: %s", cryptoErrStr(buffer, sizeof(buffer)));
```
bneradt, line 214: Nice change. Since we have all the data here in one place, no need to use the `HMAC_Update` funcionality.

bneradt (commented):
Nice cleanup.

maskit:
We may actually want to keep `outlen`. Although `HMAC()` is available on OpenSSL 3.x, we will want to use `EVP_Q_mac()` instead to use `LIBCTX`.

> HMAC() uses the default OSSL_LIB_CTX. Use [EVP_Q_mac(3)](https://www.openssl.org/docs/man3.3/man3/EVP_Q_mac.html) instead if a library context is required.

And `EVP_Q_mac` receives `outsize`.
```
unsigned char *EVP_Q_mac(OSSL_LIB_CTX *libctx, const char *name, const char *propq,
                         const char *subalg, const OSSL_PARAM *params,
                         const void *key, size_t keylen,
                         const unsigned char *data, size_t datalen,
                         unsigned char *out, size_t outsize, size_t *outlen);
```

JosiahWI:
Good foresight. Thanks for pointing this out, @maskit. I won't take this out of draft until I've put the parameter back.

maskit (approved):

PR #10967, opened 2024-01-05
Add auto options for all experimental plugins

JosiahWI:
Auto options were added for the experimental plugins with special dependencies in #10741. This adds options for the remaining experimental plugins.

PR #11469, opened 2024-06-21
Restore webp_transform to build

JosiahWI:
A bug was introduced in #10967 causing us not to register the auto_option for webp_transform if the ImageMagick::Magick++ target already exists. None of the other non-experimental plugins have auto_options at this time, so this demotes webp_transform to be always build if its dependency is found, in line with those other plugins.

### DIFF
--- a/plugins/CMakeLists.txt
+++ b/plugins/CMakeLists.txt
@@ -65,20 +65,10 @@ endif()
 # find_package() call to ensure the module doesn't try
 # to create the same target twice.
 if(NOT TARGET ImageMagick::Magick++)
-  auto_option(
-    WEBP_TRANSFORM
-    FEATURE_VAR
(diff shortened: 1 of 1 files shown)


JosiahWI:
The motivation for the hacky workaround attempt that had less than desirable side effects was a bug in CMake's FindImageMagick module. Coincidentally, I made out the patch for that bug to CMake just this morning before becoming aware that webp_transform had been broken by the workaround. 

https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9617

Unfortunately we won't be able to depend on the CMake version containing the fix, once they release it, for some time, probably.

bneradt (commented):
Makes sense. Thanks for the fix.

cmcfarlen (approved):

PR #11470, opened 2024-06-21
Skip verify imagemagick plugins for asan builds

cmcfarlen:


shukitchan (approved):
Looks trivial

bneradt (approved):
We should probably make our reasoning here explicit in case someone questions it in the future: generally speaking we want to be careful about removing tests simply because they are failing. Annoying failing tests have resulted in important production fixes (like the TSDbgCtl deadlock we worked on together last summer, for instance).

But in this case, we find the asan report to be questionable and it's for a startup/shutdown invocation of ATS which is not an actual production server invocation of ATS. For these reasons, we think it is reasonable to quiet this nuisance test.

Thank you @cmcfarlen and @JosiahWI for working on patches to make CI more reliable.

cmcfarlen:
I agree we need to be careful disabling things like this.  In this case, we are only skipping for ASAN builds and only skipping the plugin verify, which doesn't exercise any code, only confirms the .so was linked correctly.  We aren't skipping any unit tests or really any code testing here.  Its just due to some strangeness around libmagick.  These tests will certainly happen on production builds that enable testing.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11471, opened 2024-06-22
Fix last left unused parameters in the project

freak82:
Fix unused parameters, using /`* name ATS_UNUSED */`, in `mgmt` and `proxy` and few others left here and there.

This pull request is the last part of the effort for [removing](https://github.com/apache/trafficserver/issues/11377) the `-Wno-unused-parameter` warning suppression.

### DIFF
--- a/plugins/header_rewrite/matcher.h
+++ b/plugins/header_rewrite/matcher.h
@@ -187,7 +187,7 @@ private:
   }
 
   bool
-  test_reg(const std::string &t, bool nocase = false) const
+  test_reg(const std::string &t) const
   {
     int ovector[OVECCOUNT];
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -598,7 +598,7 @@ add_subdirectory(lib)
 # Keep this after lib because lib is made up of third party libraries, so if
 # there are warnings, we can't do anything about it.
 # Note: -DCMAKE_COMPILE_WARNING_AS_ERROR=ON will turn warnings into errors.
-add_compile_options(-pipe -Wall -Wextra -Wno-unused-parameter)
+add_compile_options(-pipe -Wall -Wextra)
 add_compile_options(
   "$<$<COMPILE_LANGUAGE:CXX>:-Wno-noexcept-type;-Wsuggest-override;-Wno-vla-extension;-fno-strict-aliasing>"
 )
--- a/include/proxy/HostStatus.h
+++ b/include/proxy/HostStatus.h
@@ -208,7 +208,7 @@ public:
 
   // Deletes instance.
   int
-  mainEvent(int event, Event *e)
+  mainEvent(int /* event ATS_UNUSED */, Event * /* e ATS_UNUSED */)
   {
     sync_task();
 
--- a/include/proxy/ParentRoundRobin.h
+++ b/include/proxy/ParentRoundRobin.h
@@ -42,7 +42,7 @@ public:
   ParentRoundRobin(ParentRecord *_parent_record, ParentRR_t _round_robin_type);
   ~ParentRoundRobin() override;
   pRecord *
-  getParents(ParentResult *result) override
+  getParents(ParentResult * /* result ATS_UNUSED */) override
   {
     return parents;
   }
--- a/include/proxy/http/remap/RemapPlugins.h
+++ b/include/proxy/http/remap/RemapPlugins.h
@@ -36,7 +36,10 @@
  **/
 struct RemapPlugins : public Continuation {
   RemapPlugins() = default;
-  RemapPlugins(HttpTransact::State *s, URL *u, HTTPHdr *h, host_hdr_info *hi) : _s(s), _request_url(u), _request_header(h) {}
+  RemapPlugins(HttpTransact::State *s, URL *u, HTTPHdr *h, host_hdr_info * /* hi ATS_UNUSED */)
+    : _s(s), _request_url(u), _request_header(h)
+  {
+  }
 
   // Some basic setters
   void
--- a/include/proxy/logging/LogFilter.h
+++ b/include/proxy/logging/LogFilter.h
@@ -414,7 +414,7 @@ findPatternFromParamName(const char *lookup_query_param, const char *pattern)
  * replace param value whose name contains pattern with same count 'X' of original value str length
 ---------------------------------------------------------------------------*/
 static void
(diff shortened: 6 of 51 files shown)


plugins/header_rewrite/matcher.h
```diff
@@ -187,7 +187,7 @@ template <class T> class Matchers : public Matcher
   }
 
   bool
-  test_reg(const std::string &t, bool nocase = false) const
+  test_reg(const std::string &t) const
```
freak82, line 190: There is `_nocase` member flag which is used internally and the `nocase` parameter is completely ignored here as well in the `Matchers::test` function where the `test_reg` function is called.
That's why I got the liberty to remove this parameter. 
Was this the right decision?

freak82:
This is strange. I built in debug and release but the builds were successful. 
There is some difference.
It's a bit embarrassing to use the CI infrastructure to catch these compile time errors.

freak82:
The reason for missing these warnings was that `ENABLE_CRIPTS` was not set for my build.

freak82:
And I build now with clang because GCC seems to not report unused parameters in some unused functions (like the Trie copy constructor).

freak82:
So, Clang started reporting `unused-parameters` in the `lib` directory but I was told to not do changes there because there is 3rd party libraries.
```
/z/trafficserver/lib/swoc/include/swoc/Vectray.h:235:112: error: unused parameter 'alloc' [-Werror,-Wunused-parameter]
  235 | template <typename T, size_t N, class A> Vectray<T, N, A>::Vectray(Vectray::size_type n, allocator_type const &alloc) : Vectray() {
```
Should I add the `-Wno-unused-parameter` to some `CMakeLists.txt` there or do something else?

lib/swoc/include/swoc/Vectray.h
```diff
@@ -232,7 +232,7 @@ template <typename T, size_t N, class A = std::allocator<T>> class Vectray {
 
 template <typename T, size_t N, typename A> Vectray<T, N, A>::Vectray() {}
 
-template <typename T, size_t N, class A> Vectray<T, N, A>::Vectray(Vectray::size_type n, allocator_type const &alloc) : Vectray() {
+template <typename T, size_t N, class A> Vectray<T, N, A>::Vectray(Vectray::size_type n, allocator_type const &) : Vectray() {
```
freak82, line 235: I was told to not change code in the `lib` directory because it contains 3rdparty libraries.
However, this was the only error reported from there and I don't think it could have been silenced via `CMakeLists.txt` configuration because this is a template class which is not built in the library but directly included and used from outside.
I think, it was either changing this piece of code or leaving the warning suppression at the highest level.

freak82:
I'm going to set back the warning suppression as there are more knobs which need to turned on.
I don't want to turn the CI into a sandbox for finding the next missing feature which I haven't enabled while building.

JosiahWI:
The AuTest tls_sni_with_port failed.

```
Process: server2.wow.com: Failed
       Setting up : Copying '/home/jenkins/workspace/Github_Builds/autest/src/build/proxy-verifier-v2.10.1/linux-amd64' to '/tmp/sandbox/tls_sni_with_port/server2.wow.com/bin' - Passed
       Setting up : recycling port: 61178, queue size: 993 - Passed
       Setting up : recycling port: 61179, queue size: 992 - Passed
       Setting up : recycling port: 61180, queue size: 991 - Passed
       Setting up : Copying '/home/jenkins/workspace/Github_Builds/autest/src/tests/gold_tests/autest-site/../../tools/proxy-verifier/ssl/server.pem' to '/tmp/sandbox/tls_sni_with_port/server2.wow.com/server.pem' - Passed
       Setting up : Copying '/home/jenkins/workspace/Github_Builds/autest/src/tests/gold_tests/autest-site/../../tools/proxy-verifier/ssl/ca.pem' to '/tmp/sandbox/tls_sni_with_port/server2.wow.com/ca.pem' - Passed
       Setting up : Copying 'tls_sni_with_port.replay.yaml' to '/tmp/sandbox/tls_sni_with_port/server2.wow.com/tls_sni_with_port.replay.yaml' - Passed
       Test : Checking that ReturnCode == 0 - Passed
          Reason: Returned Value: 0 == 0
          Reason: Returned value: 2.0000126361846924 < 600.0
       file /tmp/sandbox/tls_sni_with_port/_output/0-tr-server2.wow.com/stream.all.txt : request was remaped to server three - Failed
          Reason: Contents of /tmp/sandbox/tls_sni_with_port/_output/0-tr-server2.wow.com/stream.all.txt did not contains expression: "Received an HTTP/1 Content-Length body of 16 bytes for key conn_remapped"
```

JosiahWI:
[approve ci autest]

JosiahWI:
The AuTest proxy_protocol failed.

JosiahWI:
[approve ci autest]

lib/swoc/include/swoc/Vectray.h
JosiahWI: Would you make a special PR for the libswoc change? I think there's value in keeping a separation between the core changes and the libswoc changes in case people want to work with the libswoc commit history.

plugins/header_rewrite/matcher.h
JosiahWI: It looks like the API is designed so that `Matchers::set(data, COND_NOCASE)` should be called to enable that option. I think removing the parameter is was a good decision.

lib/swoc/include/swoc/Vectray.h
freak82: Done.

freak82:
Thanks.
I was hoping initially that these will be the last changes needed.
However, it turned out that I'm missing some build knobs which enable certain features/subsystems.
I need to figure out all of the possible build knobs, to enable them explicitly and cleanup the last standing unused parameters.

JosiahWI:
The main ones are probably building with QUIC enabled (requires an alternate SSL implementation, tools/build_h3_tools can build that stuff) and making sure all plugins are built (some are turned off if dependencies are missing).

JosiahWI:
A very useful technique for debugging build configurations is to use the CI presets, for example: `cmake --preset ci-rocky`. This should enable all the same knobs used in CI. But this will still skip some experimental plugins I think if dependencies are missing, and you'll need to have run the h3_build_tools script so that custom-built dependencies are installed in the right locations where it expects them.

freak82:
That's really useful. Thank you.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11472, opened 2024-06-24
Add set-custom-body config item to header_rewrite

jasmine-nahrain:
Adding config item set-custom-body to header rewrite. 
The config takes in a url. When triggered, there will be a secondary call the the specified endpoint to get a response body from. What the client will see is the response body from the second call.
The second call goes through as an internal ATS request and goes gets remapped. 
The status of the original transaction will remain in all error status code cases. OK status codes will appear as a 500 error code due to the nature of TSHttpTxnErrorBodySet.

e.g.
set_custom_body.conf
```
cond %{READ_RESPONSE_HDR_HOOK} [AND]
cond %{STATUS} = 404
set-custom-body https://example.com/404
```
remap.config
```
map /home http://example.com/home @plugin=header_rewrite.so @pparam=set_custom_body.conf
map http://example.com/404 http://example.com/404
```
With the above configs, when /home returns a 404 status code, a second call will be made to /404 through ATS internal request
The response that is sent to the client is 
- The status code from the original transaction (/home)
- The response body from the second transaction (/404)


JosiahWI:
[approve ci rocky]

JosiahWI:
[approve ci autest]

plugins/header_rewrite/operators.cc
```diff
@@ -30,6 +30,91 @@
 #include "operators.h"
 #include "ts/apidefs.h"
 
+namespace
+{
+const unsigned int LOCAL_IP_ADDRESS = 0x0100007f;
+const unsigned int MAX_SIZE         = 256;
+const int          LOCAL_PORT       = 8080;
+
+static int
```
ywkaras, line 39: It doesn't need to be static.  Putting it in an anonymous namespace makes it file scope.

plugins/header_rewrite/operators.cc
```diff
@@ -30,6 +30,91 @@
...
+  } break;
+  case OperatorSetBodyFrom::TS_EVENT_FETCHSM_TIMEOUT: {
+    Dbg(pi_dbg_ctl, "OperatorSetBodyCustom: Timeout getting custom body");
+    TSHttpTxnReenable(http_txn, TS_EVENT_HTTP_CONTINUE);
+  } break;
+  case TS_EVENT_HTTP_TXN_CLOSE: {
+    TSContDestroy(cont);
+    TSHttpTxnReenable(http_txn, TS_EVENT_HTTP_CONTINUE);
+  } break;
+  case TS_EVENT_HTTP_SEND_RESPONSE_HDR:
```
ywkaras, line 80: What's the purpose of executing the continuation on this hook (since nothing is done)?

tests/gold_tests/pluginTest/header_rewrite/header_rewrite_set_body_from.test.py
```diff
@@ -0,0 +1,84 @@
...
+        tr.Processes.Default.StartBefore(self.ts)
+        tr.Processes.Default.Streams.stderr = "gold/header_rewrite-set_body_from_headers.gold"
+        tr.Processes.Default.Streams.stdout = "gold/header_rewrite-set_body_from_body.gold"
+        tr.StillRunningAfter = self.server
+
+    def run(self):
+        self.runTraffic()
+
+
+HeaderRewriteSetBodyFromTest().run()
```
ywkaras, line 84: Could you add another test run where the fetch of the body URL fails?

plugins/header_rewrite/operators.cc
jasmine-nahrain: This pauses the original transaction to allow time to get the response from the second endpoint. The second endpoint will reenable the first once it has completed.

JosiahWI:
The AuTest transaction_data_sink failed.

```
     file /tmp/sandbox/transaction_data_sink/ts/log/traffic.out : The expected HTTP/2 response body was dumped. - Failed
        Reason: Contents of /tmp/sandbox/transaction_data_sink/ts/log/traffic.out did not contains expression: ""http2_response_body_dumped""
```

JosiahWI:
[approve ci autest]

plugins/header_rewrite/operators.cc
```diff
@@ -1219,3 +1304,60 @@ OperatorRunPlugin::exec(const Resources &res) const
...
+
+    TSHttpTxnHookAdd(res.txnp, TS_HTTP_SEND_RESPONSE_HDR_HOOK, fetchCont);
+    TSHttpTxnHookAdd(res.txnp, TS_HTTP_TXN_CLOSE_HOOK, fetchCont);
+
+    TSFetchEvent event_ids;
+    event_ids.success_event_id = TS_EVENT_FETCHSM_SUCCESS;
+    event_ids.failure_event_id = TS_EVENT_FETCHSM_FAILURE;
+    event_ids.timeout_event_id = TS_EVENT_FETCHSM_TIMEOUT;
+
+    struct sockaddr_in addr;
```
JosiahWI, line 1349: Include `<sys/socket.h>`?

JosiahWI (changes requested):

plugins/header_rewrite/operators.cc
```diff
@@ -1219,3 +1304,60 @@ OperatorRunPlugin::exec(const Resources &res) const
...
+  // we want the arg since body only takes one value
+  _value.set_value(p.get_arg());
+  require_resources(RSRC_SERVER_RESPONSE_HEADERS);
+  require_resources(RSRC_RESPONSE_STATUS);
+}
+
+void
+OperatorSetBodyFrom::initialize_hooks()
+{
+  add_allowed_hook(TS_HTTP_READ_RESPONSE_HDR_HOOK);
```
ywkaras, line 1322: I notice that your test scenario is using header_rewrite as a remap plugin.  Would this operator ever be useful when header_rewrite is used as a global plugin?

plugins/header_rewrite/operators.cc
```diff
@@ -30,6 +30,91 @@
...
+  case TS_EVENT_HTTP_TXN_CLOSE: {
+    TSContDestroy(cont);
+    TSHttpTxnReenable(http_txn, TS_EVENT_HTTP_CONTINUE);
+  } break;
+  case TS_EVENT_HTTP_SEND_RESPONSE_HDR:
+    // Do nothing
+    // The transaction is reenabled with the FetchSM transaction
+    break;
+  default:
+    TSWarning("Warning: handleFetchEvents got unknown event: %d", event);
```
ywkaras, line 85: I think this is generally treated as at least an error:  https://github.com/apache/trafficserver/blob/36bbb0ee900d152637daac35044e24b6c4308c21/plugins/s3_auth/s3_auth.cc#L1013

tests/gold_tests/pluginTest/header_rewrite/header_rewrite_set_body_from.test.py
```diff
@@ -0,0 +1,90 @@
...
+        # Request/response for original failed transaction
+        fail_1_request_header = {"headers": "GET /fail HTTP/1.1\r\nHost: www.example.com\r\n\r\n"}
+        self.server.addResponse("sessionfile.log", fail_1_request_header, response_header)
+
+        # Request/response for custom body transaction
+        fail_2_request_header = {"headers": "GET /404_fail HTTP/1.1\r\nHost: www.example.com\r\n\r\n"}
+        fail_2_response_header = {"headers": "HTTP/1.1 502 \r\nConnection: close\r\n\r\n", "body": "Fail\n"}
+        self.server.addResponse("sessionfile.log", fail_2_request_header, fail_2_response_header)
+
+        # Request/response for original Successful transaction
```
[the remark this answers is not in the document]
jasmine-nahrain: Successful meaning the second transaction was successful. I can update the naming if too confusing.

plugins/header_rewrite/operators.cc
jasmine-nahrain: Yes - this works with both remap and plugin configs. I have just added some tests to check this aswell.

plugins/header_rewrite/operators.h
```diff
@@ -478,3 +478,24 @@ class OperatorRunPlugin : public Operator
...
 };
+
+class OperatorSetBodyFrom : public Operator
+{
+public:
+  OperatorSetBodyFrom() { Dbg(pi_dbg_ctl, "Calling CTOR for OperatorSetBodyFrom"); }
+
+  // noncopyable
+  OperatorSetBodyFrom(const OperatorSetBodyFrom &) = delete;
+  void operator=(const OperatorSetBodyFrom &)      = delete;
```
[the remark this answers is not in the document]
jasmine-nahrain: What do you mean?

tests/gold_tests/pluginTest/header_rewrite/header_rewrite_set_body_from.test.py
```diff
@@ -0,0 +1,90 @@
...
+        success_1_request_header = {"headers": "GET /success HTTP/1.1\r\nHost: www.example.com\r\n\r\n"}
+        self.server.addResponse("sessionfile.log", success_1_request_header, response_header)
+
+        # Request/response for custom body transaction
+        success_2_request_header = {"headers": "GET /404 HTTP/1.1\r\nHost: www.example.com\r\n\r\n"}
```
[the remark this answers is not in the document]
jasmine-nahrain: I have updated the naming to hopefully be more clear. E.g. the first transaction returns 404 so we want a 404 response body.

plugins/header_rewrite/operators.h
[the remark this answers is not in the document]
JosiahWI: Thought about it some more - this is probably fine. Whenever copy constructors are changed it’s good to think about move constructors too.

tests/gold_tests/pluginTest/header_rewrite/rules/rule_set_body_from_remap.conf
```diff
@@ -0,0 +1,24 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+cond %{READ_RESPONSE_HDR_HOOK}
+cond %{CLIENT-URL:PATH} = "remap_success" [OR]
+cond %{CLIENT-URL:PATH} = "200"
+set-body-from http://www.example.com/404.html
+
+cond %{READ_RESPONSE_HDR_HOOK}
```
ywkaras, line 22: Does the set-body-from operator _require_ this hook condition in order to work properly?  If so, maybe you should mention that in the documentation.

tests/gold_tests/pluginTest/header_rewrite/header_rewrite_set_body_from.test.py
```diff
@@ -0,0 +1,160 @@
...
+        self.ts.Disk.remap_config.AddLine(
+            """\
+             map http://www.example.com/remap_success http://127.0.0.1:{0}/remap_success @plugin=header_rewrite.so @pparam={1}/rule_set_body_from_remap.conf
+             map http://www.example.com/200 http://127.0.0.1:{0}/200 @plugin=header_rewrite.so @pparam={1}/rule_set_body_from_remap.conf
+             map http://www.example.com/remap_fail http://127.0.0.1:{0}/remap_fail @plugin=header_rewrite.so @pparam={1}/rule_set_body_from_remap.conf
+             map http://www.example.com/plugin_success http://127.0.0.1:{0}/plugin_success
+             map http://www.example.com/plugin_fail http://127.0.0.1:{0}/plugin_fail
+             map http://www.example.com/404.html http://127.0.0.1:{0}/404.html
+             map http://www.example.com/502 http://127.0.0.1:{0}/502
+             """.format(self.server.Variables.Port, Test.RunDirectory))
```
ywkaras, line 80: I suggest you also have a remap for a body URL like this:
```

Test.GetTcpPort("bad_port")
self.ts.Disk.remap_config.AddLine(f'map http://www.example.com/plugin_no_server http://127.0.0.1::{Test.Variables.bad_port}/plugin_no_server")
```
in order to verify proper behavior if it's not possible to connect to the server that provided the body that's set by the operator.

tests/gold_tests/pluginTest/header_rewrite/header_rewrite_set_body_from.test.py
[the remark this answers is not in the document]
JosiahWI: Thanks, much easier for me to understand now!

tests/gold_tests/pluginTest/header_rewrite/rules/rule_set_body_from_remap.conf
```diff
@@ -0,0 +1,24 @@
...
+cond %{CLIENT-URL:PATH} = "remap_fail"
+set-body-from http://www.example.com/fail
```
ywkaras, line 24: So, in all the test cases, we still do the original request to upstream, then entirely throw it away.  Wouldn't more realistic test cases be with conditions checking response status or response headers?

If the intent is to always discard the original response, wouldn't it be better for the operator to execute in the remap pseudo hook, and set an error status there, to block the original request?

ywkaras: statichit is an example of a plugin that blocks the request to upstream by setting an error status in the DoRemap funciton:  https://github.com/apache/trafficserver/blob/0d004b3a8f8a63cde8e2e67e9117f2def968a312/plugins/statichit/statichit.cc#L639

plugins/header_rewrite/operators.cc
```diff
@@ -30,6 +30,91 @@
...
+  case TS_EVENT_HTTP_TXN_CLOSE: {
+    TSContDestroy(cont);
+    TSHttpTxnReenable(http_txn, TS_EVENT_HTTP_CONTINUE);
+  } break;
+  case TS_EVENT_HTTP_SEND_RESPONSE_HDR:
+    // Do nothing
+    // The transaction is reenabled with the FetchSM transaction
+    break;
+  default:
+    TSError("Warning: handleFetchEvents got unknown event: %d", event);
```
ywkaras, line 85: So in diags.log this will look like:
ERROR: Warning: ...
huh???

tests/gold_tests/pluginTest/header_rewrite/rules/rule_set_body_from_remap.conf
jasmine-nahrain: It does not. I am using the path to differentiate the tests.

tests/gold_tests/pluginTest/header_rewrite/rules/rule_set_body_from_remap.conf
jasmine-nahrain: I have written the tests to replicate what happens in production. Do you have an example of another test that only checks the response status?

plugins/header_rewrite/operators.cc
```diff
@@ -1219,3 +1304,60 @@ OperatorRunPlugin::exec(const Resources &res) const
...
+    return;
+  }
+
+  char req_buf[MAX_SIZE];
+  int  req_buf_size = 0;
+  if (createRequestString(_value.get_value(), req_buf, &req_buf_size) == TS_SUCCESS) {
+    TSCont fetchCont = TSContCreate(handleFetchEvents, TSMutexCreate());
+    TSContDataSet(fetchCont, static_cast<void *>(res.txnp));
+
+    TSHttpTxnHookAdd(res.txnp, TS_HTTP_SEND_RESPONSE_HDR_HOOK, fetchCont);
```
ywkaras, line 1341: Suppose there is another continuation for the transaction, that is either after the continuation calling this exec func on the READ_RESPONSE_HDR_HOOK, or before fetchCont on the SEND_RESPONSE_HDR_HOOK  And, suppose this continuation, like fetchEvent for the SEND_RESPONSE_HDR evert, does not call TxnReenable().  Seems like there could be a race condition, where the URL fetch body done event reenables for that other continuation by mistake.

Could you add a return value to the operator exec funtions, which would control whether the continuation running the exec functions called TxnReenable()?  That way, no other continuations of transaction hooks would be run until the fetch of the body URL finished.

jasmine-nahrain: I see what you are saying but I dont see how the exec func returning something would help. I think a return value would only have impact on header_rewrite rather than all plugins.

tests/gold_tests/pluginTest/header_rewrite/rules/rule_set_body_from_remap.conf
jasmine-nahrain: The status is set from that initial request.

plugins/header_rewrite/operators.cc
ywkaras: This is what I'm proposing, for the exec() function to return a flag that controls whether reenable is called:  https://github.com/ywkaras/trafficserver/commit/0ef4d24b905953f88d866b3c11cd5cfb88e55f89

Unfortunately, for this approach to work when the operator is run in the remap pseudo hook, the call to TSFetchUrl() would have to be deferred to the POST_REMAP hook.

doc/admin-guide/plugins/header_rewrite.en.rst
```diff
@@ -704,6 +704,30 @@ set-body
...
+  set-body-from <URL>
+
+Will call ``<URL>`` (see URL in `URL Parts`_) to retrieve a custom error response
+and set the body with the result. Triggering this rule on an OK transaction will
+send a 500 status code to the client with the desired response. If this is triggered
+on any error status code, that original status code will be sent to the client.
+
+An example config would look like
+
+   cond %{READ_RESPONSE_HDR_HOOK}
```
ywkaras, line 720: Should the doc say this condition is required for this new operator?

plugins/header_rewrite/operators.cc
jasmine-nahrain: I see. I think it would be better to seperate these changes into 2 PRs because the change you are proposing impacts all of header rewrite.

ywkaras: OK, I can put up a follow-on PR once this is merged.

ywkaras:
[approve ci autest]

tests/gold_tests/pluginTest/header_rewrite/header_rewrite_set_body_from.test.py
```diff
@@ -0,0 +1,156 @@
...
+        Test where set-body-from request succeeds and returns 200 OK
+        Triggered from remap file
+        This is tested because right now, TSHttpTxnErrorBodySet will change OK status codes to 500 INKApi Error
+        Ideally, this would not occur.
+        '''
+        tr = Test.AddTestRun()
+        tr.Processes.Default.Command = (
+            'curl -s -v --proxy 127.0.0.1:{0} "http://www.example.com/200"'.format(self.ts.Variables.port))
+        tr.Processes.Default.ReturnCode = 0
+        tr.Processes.Default.Streams.All = "gold/header_rewrite-set_body_from_200.gold"
```
ywkaras, line 142: Testing against Streams.All seems to be making this test flaky/fragile.  I think you need your gold file to be matched against Streams.stdout, which I think would just contain the body.  You can test the response HTTP headers like this:  https://github.com/apache/trafficserver/blob/84206a8c089e56efc59fd08f2e3b693c413aaf03/tests/gold_tests/pluginTest/s3_auth/s3_auth_config.test.py#L60 .

JosiahWI (commented):
Looks good. Let's fix the AuTest so it isn't flaky.

JosiahWI (dismissed):

ywkaras (approved):

ywkaras:
This PR introduces this issue:  https://github.com/apache/trafficserver/issues/11549

PR #11473, opened 2024-06-24
Remove unused defines from P_Connection.h

JosiahWI:
It seems these have been superceded by NetVCOptions. Having them in the source code is confusing, since they no longer have any effect.

### DIFF
--- a/src/iocore/net/Connection.cc
+++ b/src/iocore/net/Connection.cc
@@ -46,8 +46,6 @@
 // set in the OS
 // #define RECV_BUF_SIZE            (1024*64)
 // #define SEND_BUF_SIZE            (1024*64)
-#define FIRST_RANDOM_PORT 16000
-#define LAST_RANDOM_PORT  32000
 
 namespace
 {
--- a/src/iocore/net/Connection.cc
+++ b/src/iocore/net/Connection.cc
@@ -43,10 +43,6 @@
 #include <sys/linker.h>
 #endif
 
-// set in the OS
-// #define RECV_BUF_SIZE            (1024*64)
-// #define SEND_BUF_SIZE            (1024*64)
-
 namespace
 {
 DbgCtl dbg_ctl_http_tproxy{"http_tproxy"};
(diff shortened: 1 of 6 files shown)


src/iocore/net/Connection.cc
```diff
@@ -46,8 +46,6 @@
 // set in the OS
 // #define RECV_BUF_SIZE            (1024*64)
 // #define SEND_BUF_SIZE            (1024*64)
```
cmcfarlen, line 48: Could probably remove these lines as well

src/iocore/net/P_Connection.h
```diff
@@ -58,20 +58,8 @@ struct NetVCOptions;
...
-#define BIND_RANDOM_PORT     true
-#define BIND_ANY_PORT        false
-#define ENABLE_MC_LOOPBACK   true
-#define DISABLE_MC_LOOPBACK  false
-#define BC_NO_CONNECT        true
-#define BC_CONNECT           false
-#define BC_NO_BIND           true
-#define BC_BIND              false
+#define NON_BLOCKING true
+#define BLOCKING     false
```
cmcfarlen, line 62: These are only used for NetAccept::do_listen.  Perhaps there is a cleaner way than macros?

JosiahWI: How about splitting `NetAccept::do_listen` into `NetAccept::do_blocking_listen` and `NetAccept::do_non_blocking_listen` or something similar? The defines are being passed as boolean arguments to the method, so basically someone compensated for `NetAccept::do_listen(false)` being incomprehensible. I quite like how it reads, actually, but I think a more explicit API might be cleaner. I'm worried about the verbosity of splitting it into two methods, though... the macros make it very easy to understand with very little code.

JosiahWI: It's still suboptimal since `NetAccept::do_listen` called into other APIs that had the same annoyance. I kept it as `NetAccept::do_listen_impl` and wrapped it in `NetAccept::do_listen` and `NetAccept::do_blocking_listen` so that the public API will at least be more presentable.

JosiahWI:
[approve ci autest clang-analyzer fedora ubuntu]

cmcfarlen (approved):

PR #11475, opened 2024-06-25
Remove unused `alloc` parameter in swoc/Vectray functionality

freak82:
This pull request is part of the effort for [removing](https://github.com/apache/trafficserver/issues/11377 ) the `-Wno-unused-parameter` warning suppression.

I was told to not change code in the lib directory because it contains 3rdparty libraries.
However, this was the only error reported from there and I don't think it could have been silenced via `CMakeLists.txt` configuration because this is a template class which is not built in the library but directly included and used from outside.
I think, it was either changing this piece of code or leaving the warning suppression at the highest level (in the main `CMakeLists.txt`).

This warning is reported only by Clang but not by GCC according to my build tests.

### DIFF
--- a/lib/swoc/include/swoc/Vectray.h
+++ b/lib/swoc/include/swoc/Vectray.h
@@ -232,7 +232,7 @@ protected:
 
 template <typename T, size_t N, typename A> Vectray<T, N, A>::Vectray() {}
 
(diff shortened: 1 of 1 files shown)


JosiahWI (approved):

JosiahWI:
We are maintaining the libswoc we use, so we can make fixes to it as necessary, and this is the correct repository to make those fixes in.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11479, opened 2024-06-26
Add UnixSocket class based on SocketManager

JosiahWI:
## UnixSocket Class

  The new class is a thin wrapper for a Unix socket file descriptor. It does
  not manage the resource itself, although this would be a good idea for a
  future improvement.

## Removed Functions

  Some functions are not applicable to sockets and do not belong in a utility
  for managing sockets. The functions open, pwrite, lseek and fsync were not
  transferred to the new class. The function socket() was replaced by a
  UnixSocket constructor.

## API Changes

  The file descriptor no longer needs to be passed because it is remembered
  by the UnixSocket object. Some parameters that were unused are removed
  from the new interface. There is no longer a socket() function; instead
  a UnixSocket constructor with the same arguments is provided.

## Implementation Changes

  Other than small cleanups like changing uint64_t to std::uint64_t and
  constifying parameters, all implementations are directly copied from
  the old SocketManager ones except for client_fastopen_supported. It
  was refactored to improve naming and make the use of the bitmask obvious.

## Performance Considerations

  Since the class is a thin wrapper, there should be no performance overhead
  from the use of a class. All methods that were declared inline in a header
  for UnixSocket are also declared inline in the UnixSocket.h header. There
  could be a small increase in compile times because there is no separate
  P_UnixSocket.h header, but UnixSocket.h is private to inkevent so far, so
  the compilation time difference should be negligible.

### DIFF
--- a/src/iocore/eventsystem/UnixSocket.h
+++ b/src/iocore/eventsystem/UnixSocket.h
@@ -89,16 +89,16 @@ private:
   int sock_fd{-1};
 };
 
-TS_INLINE
+inline
 UnixSocket::UnixSocket(int fd) : sock_fd{fd} {}
 
-TS_INLINE
+inline
 UnixSocket::UnixSocket(int domain, int type, int protocol)
 {
   this->sock_fd = socket(domain, type, protocol);
 }
 
-TS_INLINE std::int64_t
+inline std::int64_t
           UnixSocket::read(void *buf, int size) const
 {
   std::int64_t r;
@@ -112,7 +112,7 @@ TS_INLINE std::int64_t
   return r;
 }
 
-TS_INLINE int
+inline int
 UnixSocket::recv(void *buf, int size, int flags) const
 {
   int r;
@@ -124,7 +124,7 @@ UnixSocket::recv(void *buf, int size, int flags) const
   return r;
 }
 
-TS_INLINE int
+inline int
 UnixSocket::recvfrom(void *buf, int size, int flags, struct sockaddr *addr, socklen_t *addrlen) const
 {
   int r;
@@ -137,7 +137,7 @@ UnixSocket::recvfrom(void *buf, int size, int flags, struct sockaddr *addr, sock
   return r;
 }
 
-TS_INLINE int
+inline int
 UnixSocket::recvmsg(struct msghdr *m, int flags) const
 {
   int r;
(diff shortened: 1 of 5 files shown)


cmcfarlen:
imo we should NOT use TS_INLINE for new code.  Just use `inline` where appropriate.

maskit:
What's the responsibility of SocketManager? Do we still need it?

JosiahWI:
@maskit Nope, although we have to find a home for the `open` function before we get rid of it - it's used only in the cache `Store` so maybe it could go there. I left it in as a compatability layer to protect the rest of ATS from the changes; I will follow this up with (a) separate PR(s) to remove `SocketManager`. We will have to update inknet, inkdns, inkproxy,  and inkhostdb for the new API. Since this PR is already a sizable chunk for review, I didn't want to go touching every other subsystem at the same time.

src/iocore/eventsystem/UnixSocket.h
```diff
@@ -0,0 +1,291 @@
...
+  int res;
+  do {
+    if (unlikely((res = ::shutdown(this->sock_fd, how)) < 0)) {
+      res = -errno;
+    }
+  } while (res == -EINTR);
+  return res;
+}
+
+// This does not belong here. ^^
```
JosiahWI, line 279: Oops I forgot about this. 😂

maskit:
An option for `open` function is to have a constructor that receives the three parameters instead of an fd. But because we don't use exceptions, we'd need a function to check whether UnixSocket has a valid fd.

maskit:
> The open function is used to open a file by pathname. The socket function is used to create a socket.

You are right. I somehow mixed up the two, and I meant `socket()` on the previous comment. Functions for files need their own home.

src/iocore/eventsystem/UnixSocket.h
```diff
@@ -0,0 +1,290 @@
...
+private:
+  int sock_fd{-1};
+};
+
+inline UnixSocket::UnixSocket(int fd) : sock_fd{fd} {}
+
+inline UnixSocket::UnixSocket(int domain, int type, int protocol)
+{
+  this->sock_fd = socket(domain, type, protocol);
+}
```
JosiahWI, line 97: We need a way to check whether the socket is valid.

JosiahWI:
I have added a function to check whether it was able to get the socket.

maskit (approved):

PR #11488, opened 2024-06-27
cripts: rename get/set macros to Get/Set

cmcfarlen:
The lowercase macros interfere with boringssl headers when used in the build.

### DIFF
--- a/include/cripts/Context.hpp
+++ b/include/cripts/Context.hpp
@@ -100,8 +100,8 @@ private:
 } // namespace Cript
 
 // This may be weird, but oh well for now.
-#define get()               _get(context)
-#define set(_value)         _set(context, _value)
+#define Get()               _get(context)
+#define Set(_value)         _set(context, _value)
 #define update()            _update(context)
 #define runRemap()          _runRemap(context)
 #define activate()          _activate(context->p_instance)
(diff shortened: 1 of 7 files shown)


include/cripts/Context.hpp
```diff
@@ -100,8 +100,8 @@ class Context
 } // namespace Cript
 
 // This may be weird, but oh well for now.
-#define get()               _get(context)
-#define set(_value)         _set(context, _value)
+#define Get()               _get(context)
```
maskit, line 103: Can we do this?
```c
#define Response::get()    Response::_get(context)
```

zwoop: No, C++ doesn't allow this (I tried that long ago, it's a good idea, but no bueno).

```
/tmp/foo.cc:3:17: warning: ISO C99 requires whitespace after the macro name [-Wc99-extensions]
#define Response::get() Response::_get(context)
                ^
1 warning generated.
```

zwoop (approved):
This is good. For consistency, we agreed to change all public define / methods like this to Pascal case, rather than camel case. Consistency is important, particularly for this type of "scripting" API.

I'll take care of the rest from here.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11489, opened 2024-06-28
Remove the `-Wno-unused-parameter` warning suppression

freak82:
This removes the `-Wno-unused-parameter` warning suppression from the main `CMakeLists.txt` of the project.

I built the project with different presets: `ci-debian`, `ci-ubuntu`, `ci-rocky`, `ci-clang-analyzer` with `GCC 13.2`.
Also built it with `Clang 17`. Built it in both debug and release modes.

Seems like the warning is no longer needed. Let's see if I'm missing something.

This pull request is hopefully the final part of the effort for [removing](https://github.com/apache/trafficserver/issues/11377) the `-Wno-unused-parameter` warning suppression.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -597,7 +597,7 @@ add_subdirectory(lib)
 # Keep this after lib because lib is made up of third party libraries, so if
 # there are warnings, we can't do anything about it.
 # Note: -DCMAKE_COMPILE_WARNING_AS_ERROR=ON will turn warnings into errors.
-add_compile_options(-pipe -Wall -Wextra -Wno-unused-parameter)
+add_compile_options(-pipe -Wall -Wextra)
 add_compile_options(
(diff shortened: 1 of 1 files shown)


JosiahWI:
As I cautioned about earlier, some experimental plugins are skipped if dependencies are missing. It looks like those are the remaining culprits. CMake should print a report at the end of its configuration showing which plugins are on or off, and you can enable the ones that are off (e.g. `-DENABLE_MAXMIND_ACL=ON`) to require them so that CMake tells you which dependencies are missing.

freak82:
Understood. I'll close this pull request as the things are not ready yet.

JosiahWI:
Feel free to re-open it with the remaining fixes once they're ready. I think we'll be able to remove the warning suppression after the experimental plugins are fixed.

Issue #11446, opened 2024-06-13
HTTP/3 benchmark results


bryancall:
```
**http2load**
finished in 65.00s, 25725.23 req/s, 29.04MB/s
requests: 1543514 total, 1543514 started, 1543514 done, 1543514 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 1543477 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 1.70GB (1827287672) total, 289.53MB (303596678) headers (space savings 35.34%), 1.88GB (2023121920) data
UDP datagram: 2186563 sent, 2028845 received
                     min         max         mean         sd        +/- sd
time for request:     1.73ms       2.21s     29.36ms     19.76ms    78.47%
time for connect:        0us         0us         0us         0us     0.00%
time to 1st byte:        0us         0us         0us         0us     0.00%
req/s           :     165.41      350.58      257.25       29.79    64.00%

**dstat**
You did not select any stats, using -cdngy by default.
----total-usage---- -dsk/total- ---net/lo-- -net/total- ---paging-- ---system--
usr sys idl wai stl| read  writ| recv  send: recv  send|  in   out | int   csw
 77   9  13   0   0|   0  4111k| 532B  532B:8215k  112M|   0     0 |3735k  640k
 89   7   3   0   0|   0   456k| 512B  512B:7116k   85M|   0     0 |3437k  610k
 92   6   2   0   0|   0   140k| 511B  511B:5983k   60M|   0     0 |2680k  495k
 94   5   1   0   0|   0  1639B| 533B  533B:5033k   44M|   0     0 |2351k  314k
 95   4   1   0   0|   0  6552k| 511B  511B:4398k   34M|   0     0 |2149k  146k
 96   4   1   0   0|   0    32M| 512B  512B:3940k   29M|   0     0 |2012k  152k
 96   4   0   0   0|   0  1639B| 512B  512B:3651k   25M|   0     0 |1916k  159k
 96   4   0   0   0|   0     0 | 511B  511B:3396k   22M|   0     0 |1840k  167k
 96   3   0   0   0|   0    46k| 511B  511B:3180k   20M|   0     0 |1754k  168k
 97   3   0   0   0|   0  2458B| 512B  512B:3030k   19M|   0     0 |1711k  171k
 97   3   0   0   0|   0  6580k| 512B  512B:2892k   18M|   0     0 |1699k  175k
[...]

PR #11490, opened 2024-06-28
H3: Make sure we remove streams when needed.

brbzull0:
Streams weren’t deleted when needed. Stream list was increasing in size as items were never removed from it, generating a very large list and making finding items slow as we needed to walk over a large list.
This fixes the logic to check if the stream is ready to be removed, if so, then the stream is correctly removed from the list.

The main issue was the cmp in `is_readable`. Beside this I made another change to  not call `find_stream` twice when creating a stream(probably could be in a separate pr).

Notes
-------------
Tests done on a small nuc with:

`Intel(R) Core(TM) i7-6770HQ CPU @ 2.60GHz 32G mem 4 Cores 8T`

**Build in Release mode.**

### So, with the patch:

```
h2load -n 400000 -c 20 --alpn-list=h3 https://fedora:4443/cache/123000
```
Numbers from h2load are very crappy, but based on the hardware used, that's not the main point. 
```
finished in 72.81s, 5493.49 req/s, 645.26MB/s
requests: 400000 total, 400000 started, 400000 done, 400000 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 400000 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 45.88GB (49266000000) total, 59.89MB (62800000) headers (space savings 35.39%), 45.82GB (49200000000) data
UDP datagram: 1297880 sent, 2561723 received
time for request:     1.04ms     32.30ms      3.47ms       866us    72.19%
time for connect:    17.34ms     37.90ms     30.31ms      7.39ms    65.00%
time to 1st byte:    39.80ms     44.56ms     42.17ms      1.34ms    65.00%
req/s           :     274.68      300.91      288.35       10.60    40.00%
```

```
Samples: 1M of event 'cycles:P', Event count (approx.): 970950082837
Overhead  Symbol
  10.24%  [.] __memmove_avx_unaligned_erms
   5.14%  [.] _aesni_ctr32_ghash_6x
   5.09%  [.] freelist_new(_InkFreeList*)
   3.22%  [.] <alloc::string::String as core::fmt::Write>::write_str
   2.61%  [.] core::fmt::write
   2.55%  [.] _int_malloc
   1.91%  [.] <std::io::buffered::bufwriter::BufWriter<W> as std::io::Write>::write_all
   1.81%  [.] freelist_free(_InkFreeList*, void*)
   1.48%  [.] serde_json::ser::format_escaped_str
   1.36%  [.] quiche::Connection::send_single
   1.33%  [k] syscall_return_via_sysret
   1.31%  [k] rep_movs_alternative
   1.18%  [.] IOBufferBlock::clear()
   1.10%  [.] (anonymous namespace)::build_iovec_block_chain(unsigned int, long, Ptr<IOBufferBlock>&, iovec*) [clone .constprop.0]
   1.03%  [.] _int_free
   1.03%  [k] entry_SYSCALL_64_after_hwframe
   1.03%  [.] realloc
   1.01%  [.] core::fmt::Formatter::pad_integral
   0.97%  [.] <core::time::Duration as core::fmt::Debug>::fmt::fmt_decimal
   0.94%  [.] core::fmt::num::imp::<impl core::fmt::Display for usize>::fmt

```
`QUICStream::id()` and `QUICStreamManager::find_stream`:

```
Samples: 1M of event 'cycles:P', Event count (approx.): 970950082837
Overhead  Symbol
   0.05%  [.] QUICStream::id() const  -      -
```

```
Samples: 1M of event 'cycles:P', Event count (approx.): 970950082837
Overhead  Symbol
   0.04%  [.] QUICStreamManager::find_stream(unsigned long)  -      -
```

### Without the patch:

Could not use the same 400k request as the test before because the process
ran out of memory. even with 200k it started swapping, so at least we can see that the two functions mentioned up above are on the top list.

```
finished in 52.08s, 3840.39 req/s, 451.09MB/s
requests: 200000 total, 200000 started, 200000 done, 200000 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 200000 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 22.94GB (24633000000) total, 29.95MB (31400000) headers (space savings 35.39%), 22.91GB (24600000000) data
UDP datagram: 795553 sent, 1271448 received
time for request:     1.51ms    706.84ms      5.00ms      6.24ms    99.68%
time for connect:    16.42ms     38.93ms     28.44ms      7.43ms    60.00%
time to 1st byte:    31.85ms     43.99ms     37.77ms      4.06ms    60.00%
req/s           :     192.05      215.89      200.38        9.53    80.00%

```

```
Samples: 1M of event 'cycles:P', Event count (approx.): 791082557670
Overhead  Symbol
>> 23.45%  [.] QUICStream::id() const <<<<<<<< HERE
   7.26%  [.] __memmove_avx_unaligned_erms
   3.61%  [.] freelist_new(_InkFreeList*)
   3.09%  [.] _aesni_ctr32_ghash_6x
   1.91%  [.] <alloc::string::String as core::fmt::Write>::write_str
   1.88%  [k] __irqentry_text_end
   1.64%  [.] _int_malloc
   1.61%  [.] core::fmt::write
   1.39%  [.] freelist_free(_InkFreeList*, void*)
   1.24%  [k] error_entry
   1.20%  [.] <std::io::buffered::bufwriter::BufWriter<W> as std::io::Write>::write_all
   0.96%  [k] syscall_return_via_sysret
>> 0.95%  [.] QUICStreamManager::find_stream(unsigned long) <<<<<<< HERE
   0.93%  [.] serde_json::ser::format_escaped_str
   0.91%  [.] IOBufferBlock::clear()
   0.87%  [.] (anonymous namespace)::build_iovec_block_chain(unsigned int, long, Ptr<IOBufferBlock>&, iovec*) [clone .constprop.0]
   0.85%  [.] realloc
```
Fixes: https://github.com/apache/trafficserver/issues/11446

### DIFF
--- a/src/iocore/net/QUICNetVConnection.cc
+++ b/src/iocore/net/QUICNetVConnection.cc
@@ -645,8 +645,8 @@ QUICNetVConnection::_handle_read_ready()
     QUICConVDebug("stream %" PRIu64 " is readable\n", s);
     QUICStream *stream = static_cast<QUICStream *>(this->_stream_manager->find_stream(s));
     if (stream == nullptr) {
-      this->_stream_manager->create_stream(s);
-      stream = static_cast<QUICStream *>(this->_stream_manager->find_stream(s));
+      [[maybe_unused]] QUICConnectionError err;
+      stream = this->_stream_manager->create_stream(s, err);
     }
     stream->receive_data(this->_quiche_con);
   }
@@ -662,8 +662,8 @@ QUICNetVConnection::_handle_write_ready()
     while (quiche_stream_iter_next(writable, &s)) {
       QUICStream *stream = static_cast<QUICStream *>(this->_stream_manager->find_stream(s));
       if (stream == nullptr) {
-        this->_stream_manager->create_stream(s);
-        stream = static_cast<QUICStream *>(this->_stream_manager->find_stream(s));
+        [[maybe_unused]] QUICConnectionError err;
+        stream = this->_stream_manager->create_stream(s, err);
       }
       stream->send_data(this->_quiche_con);
     }
--- a/src/iocore/net/quic/QUICStreamManager.cc
+++ b/src/iocore/net/quic/QUICStreamManager.cc
@@ -30,7 +30,20 @@
 
 QUICStreamManager::QUICStreamManager(QUICContext *context, QUICApplicationMap *app_map) : _context(context), _app_map(app_map) {}
 
-QUICStreamManager::~QUICStreamManager() {}
+QUICStreamManager::~QUICStreamManager()
+{
+  // We attempt to remove any stream that's left on the list.
+  if (!stream_list.empty()) {
+    QUICStream *stream = stream_list.head;
+    QUICStream *next   = nullptr;
+    while (stream) {
+      next = stream->link.next;
+      stream_list.remove(stream);
+      delete stream;
+      stream = next;
+    }
+  }
+}
 
 void
 QUICStreamManager::set_default_application(QUICApplication *app)
@@ -89,15 +102,15 @@ QUICStreamManager::find_stream(QUICStreamId stream_id)
   return nullptr;
 }
 
-QUICConnectionErrorUPtr
-QUICStreamManager::create_stream(QUICStreamId stream_id)
+QUICStream *
+QUICStreamManager::create_stream(QUICStreamId stream_id, QUICConnectionError &err)
 {
   QUICStream *stream = new QUICStream(this->_context->connection_info(), stream_id);
   this->stream_list.push(stream);
 
   QUICApplication *application = this->_app_map->get(stream_id);
   application->on_stream_open(*stream);
-  return nullptr;
+  return stream;
 }
 
 QUICConnectionErrorUPtr
--- a/src/iocore/net/quic/QUICStreamManager.cc
+++ b/src/iocore/net/quic/QUICStreamManager.cc
@@ -35,9 +35,8 @@ QUICStreamManager::~QUICStreamManager()
   // We attempt to remove any stream that's left on the list.
   if (!stream_list.empty()) {
     QUICStream *stream = stream_list.head;
-    QUICStream *next   = nullptr;
     while (stream) {
-      next = stream->link.next;
+      QUICStream *next = stream->link.next;
       stream_list.remove(stream);
       delete stream;
       stream = next;
@@ -134,7 +133,6 @@ QUICStreamManager::delete_stream(QUICStreamId stream_id)
   application->on_stream_close(*stream);
 
   stream_list.remove(stream);
-
   delete stream;
 
   return nullptr;
--- a/src/iocore/net/quic/QUICStreamVCAdapter.cc
+++ b/src/iocore/net/quic/QUICStreamVCAdapter.cc
@@ -337,7 +337,7 @@ QUICStreamVCAdapter::is_readable()
 bool
 QUICStreamVCAdapter::is_writable()
 {
-  return this->stream().direction() != QUICStreamDirection::RECEIVE && _write_vio.nbytes != -1;
+  return this->stream().direction() != QUICStreamDirection::RECEIVE && _write_vio.nbytes != _read_vio.ndone;
 }
 
 int
--- a/src/iocore/net/quic/QUICStreamVCAdapter.cc
+++ b/src/iocore/net/quic/QUICStreamVCAdapter.cc
@@ -331,7 +331,7 @@ QUICStreamVCAdapter::reenable(VIO * /* vio ATS_UNUSED */)
 bool
 QUICStreamVCAdapter::is_readable()
 {
-  return this->stream().direction() != QUICStreamDirection::SEND && _read_vio.nbytes == _read_vio.ndone;
+  return this->stream().direction() != QUICStreamDirection::SEND && _read_vio.nbytes != _read_vio.ndone;
 }
 
 bool
--- a/include/iocore/net/quic/QUICStreamManager.h
+++ b/include/iocore/net/quic/QUICStreamManager.h
@@ -46,7 +46,8 @@ public:
   uint32_t    stream_count() const;
   QUICStream *find_stream(QUICStreamId stream_id);
 
-  QUICConnectionErrorUPtr create_stream(QUICStreamId stream_id);
+  QUICStream *create_stream(QUICStreamId stream_id, QUICConnectionError &err);
+
   QUICConnectionErrorUPtr create_uni_stream(QUICStreamId new_stream_id);
   QUICConnectionErrorUPtr create_bidi_stream(QUICStreamId new_stream_id);
   QUICConnectionErrorUPtr delete_stream(QUICStreamId new_stream_id);


src/iocore/net/QUICNetVConnection.cc
```diff
@@ -645,8 +645,8 @@ QUICNetVConnection::_handle_read_ready()
     QUICConVDebug("stream %" PRIu64 " is readable\n", s);
     QUICStream *stream = static_cast<QUICStream *>(this->_stream_manager->find_stream(s));
     if (stream == nullptr) {
-      this->_stream_manager->create_stream(s);
-      stream = static_cast<QUICStream *>(this->_stream_manager->find_stream(s));
+      [[maybe_unused]] QUICConnectionError err;
```
brbzull0, line 648: Eventually we will check for `err`.

src/iocore/net/quic/QUICStreamManager.cc
```diff
@@ -89,15 +102,15 @@ QUICStreamManager::find_stream(QUICStreamId stream_id)
   return nullptr;
 }
 
-QUICConnectionErrorUPtr
-QUICStreamManager::create_stream(QUICStreamId stream_id)
+QUICStream *
+QUICStreamManager::create_stream(QUICStreamId stream_id, QUICConnectionError &err)
```
brbzull0, line 106: This is just to avoid calling `find_stream` twice when creating a stream, no much effect on performance.

src/iocore/net/QUICNetVConnection.cc
ywkaras: Maybe you should check it with ink_assert for now.

src/iocore/net/quic/QUICStreamManager.cc
```diff
@@ -30,7 +30,20 @@
...
+  // We attempt to remove any stream that's left on the list.
+  if (!stream_list.empty()) {
+    QUICStream *stream = stream_list.head;
+    QUICStream *next   = nullptr;
+    while (stream) {
+      next = stream->link.next;
+      stream_list.remove(stream);
+      delete stream;
+      stream = next;
+    }
```
ywkaras, line 44: You could just do:
```
    QUICStream *streamp = stream_list.head;
    stream_list.clear();
    while (streamp) {
      QUICKStream *next = streamp->link.next;
      delete streamp;
      streamp = next;
    }
```

brbzull0: thanks Walt.
why is it better than the other?

src/iocore/net/QUICNetVConnection.cc
maskit: We'd never discard a value that we use on the following normal flow. Returning an error with `[[nodiscard]]` attribute would be nicer in general, although I'm not sure what error can happen in `create_stream`.

src/iocore/net/quic/QUICStreamVCAdapter.cc
```diff
@@ -331,7 +331,7 @@ QUICStreamVCAdapter::reenable(VIO * /* vio ATS_UNUSED */)
 bool
 QUICStreamVCAdapter::is_readable()
 {
-  return this->stream().direction() != QUICStreamDirection::SEND && _read_vio.nbytes == _read_vio.ndone;
```
maskit, line 334: This is fine, but didn't you mention `is_writable` as well?

src/iocore/net/quic/QUICStreamManager.cc
maskit: Accessing items in the list after calling `clear` *looks* odd. I guess it works fine for our DLL, but it doesn't look like a common way (it can be problematic if we switch the container).

ywkaras: We could put the clear after the while loop.  But I can't get my code to wash Damian car, the way he seems to want.  It's not a necessary change.

ywkaras: It seem like this destructor is only called when the connection is closed.  What if there was a connection that lasted a long time, with many streams opened and closed?  Can we remove the stream object from the list at the same time that the stream is closed?

maskit: I'd call this a safety net or the last resort. Streams are (supposed to be) removed from the list when they are closed cleanly. This cleanup should be necessary only if a connection is closed abruptly (e.g. connection errors, active timeout, etc).

src/iocore/net/QUICNetVConnection.cc
brbzull0: The only error could be that there is no app for that particular stream id, but seems unlikely. The only reason I kept the error was because it was in the original API.

We can just rely on the returned vale, either a new stream or  `nullptr` and remove the error from this API.

src/iocore/net/quic/QUICStreamVCAdapter.cc
brbzull0: I did, but from the tests I made, all streams seems to be removed base on the `is_readable`. Every time a stream is checked for `is_writable` the `write_vio.nbytes` have `INT64_MAX` as value. 

Does it make sense to have `is_writable` performing the same(keeping the direction right direction check) logic than `is_readable`, I mean, as long as they are different, means there are write capabilities on the stream.

maskit: The `-1` doesn't make sense anyway, nothing sets `-1`. Doing a similar check (`nbytes != ndone`) would make some sense, though it sounds like we have no ways to verify it at the moment. But I don't mind leaving it for now. I was just wondering because you mentioned it on Slack and no change was made for it. There is Unidirectional Stream and it's used for QPACK. We'll perhaps see an issue there once we enabled Dynamic Table.

src/iocore/net/QUICNetVConnection.cc
maskit: The API has that signature because it was designed before we start using Quiche. The old implementation had to check the number of streams open and return an error if the number exceeds the limit. And I thought returning a QUICConnectionError would be handy because the error code should be always the same if stream creation fails.

src/iocore/net/quic/QUICStreamVCAdapter.cc
brbzull0: Agreed with `-1`. On my tests, `write_vio.nbytes` was always written after we check for `is_writable`.

I'll change the check for `nbytes != ndone`  just to at least not checking for some value that it is never set.

src/iocore/net/QUICNetVConnection.cc
brbzull0: How strong you feel about the old API? If I put the old API back, then we won't be able to get the stream when creating a new one which will involve `find_stream` after creating it(as before).

maskit: Oh, I didn't mean we should keep the original API. I was just telling the background. The only debatable point would be which should be the return type of create_stream, the error or the stream. But it's not a huge deal. I'm fine with the current API.

maskit (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11491, opened 2024-06-28
Fix unused parameters in conditionally built experimental plugins

freak82:
Fix the warnings for unused parameters when all of the conditionally compiled plugins are enabled.

Clang-format did a bit weird formatting in two places.

### DIFF
--- a/plugins/experimental/stek_share/state_machine.h
+++ b/plugins/experimental/stek_share/state_machine.h
@@ -38,7 +38,7 @@ public:
   ~STEKShareSM() {}
 
   nuraft::ptr<nuraft::buffer>
-  pre_commit(const uint64_t log_idx, nuraft::buffer &data) override
+  pre_commit(const uint64_t /* log_idx ATS_UNUSED */, nuraft::buffer & /* data ATS_UNUSED */) override
   {
     return nullptr;
   }
@@ -89,21 +89,21 @@ public:
   }
 
   void
-  commit_config(const uint64_t log_idx, nuraft::ptr<nuraft::cluster_config> &new_conf) override
+  commit_config(const uint64_t log_idx, nuraft::ptr<nuraft::cluster_config> & /* new_conf ATS_UNUSED */) override
   {
     // Nothing to do with configuration change. Just update committed index.
     last_committed_idx_ = log_idx;
   }
 
   void
-  rollback(const uint64_t log_idx, nuraft::buffer &data) override
+  rollback(const uint64_t /* log_idx ATS_UNUSED */, nuraft::buffer & /* data ATS_UNUSED */) override
   {
     // Nothing to do here since we don't have pre-commit.
   }
 
   int
-  read_logical_snp_obj(nuraft::snapshot &s, void *&user_snp_ctx, uint64_t obj_id, nuraft::ptr<nuraft::buffer> &data_out,
-                       bool &is_last_obj) override
+  read_logical_snp_obj(nuraft::snapshot            &s, void            *&/* user_snp_ctx ATS_UNUSED */, uint64_t /* obj_id ATS_UNUSED */,
+                       nuraft::ptr<nuraft::buffer> &data_out, bool &is_last_obj) override
   {
     // Dbg(dbg_ctl, "read snapshot %lu term %lu object ID %lu", s.get_last_log_idx(), s.get_last_log_term(), obj_id);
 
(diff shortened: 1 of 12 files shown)


plugins/experimental/stek_share/state_machine.h
```diff
@@ -89,21 +89,21 @@ class STEKShareSM : public nuraft::state_machine
...
-  rollback(const uint64_t log_idx, nuraft::buffer &data) override
+  rollback(const uint64_t /* log_idx ATS_UNUSED */, nuraft::buffer & /* data ATS_UNUSED */) override
   {
     // Nothing to do here since we don't have pre-commit.
   }
 
   int
-  read_logical_snp_obj(nuraft::snapshot &s, void *&user_snp_ctx, uint64_t obj_id, nuraft::ptr<nuraft::buffer> &data_out,
-                       bool &is_last_obj) override
+  read_logical_snp_obj(nuraft::snapshot            &s, void            *&/* user_snp_ctx ATS_UNUSED */, uint64_t /* obj_id ATS_UNUSED */,
```
freak82, line 105: Probably the `*&` the pointer to reference syntax confuses the `clang-format` here and it did this weird formatting.

plugins/experimental/wasm/ats_context.cc
```diff
@@ -535,8 +535,8 @@ Context::getBuffer(WasmBufferType type)
 }
 
 WasmResult
-Context::httpCall(std::string_view target, const Pairs &request_headers, std::string_view request_body,
-                  const Pairs &request_trailers, int timeout_millisconds, uint32_t *token_ptr)
+Context::httpCall(std::string_view /* target ATS_UNUSED */, const Pairs &request_headers, std::string_view             request_body,
```
freak82, line 538: I'm not sure why `clang-format` left this space before the `request_body`.

plugins/experimental/wasm/lib/src/signature_util.cc
```diff
@@ -55,7 +55,7 @@ template <size_t N> constexpr std::array<uint8_t, N> hex2pubkey(const char (&hex
 
 namespace proxy_wasm {
 
-bool SignatureUtil::verifySignature(std::string_view bytecode, std::string &message) {
+bool SignatureUtil::verifySignature([[maybe_unused]] std::string_view bytecode, [[maybe_unused]] std::string &message) {
```
shukitchan, line 58: library files are coming from another repo. I would prefer to make no changes on these files. Thanks.

freak82: Reverted.

JosiahWI (approved):
Thank you for the fixes, looks good!

JosiahWI:
It's too bad `clang-format` is confused by those things. It's possible a future release will contain fixes and then it'll look better.

freak82:
I've few  general questions how to proceed with the removing of the removing of the `-Wno-unused-params` warning suppression.
I'm not sure where is the appropriate place to ask them. So, I'm posting them here.
I mean:
- I ran `tools/build_h3_tools.sh` then build the project in few different presents - `rocky`, `ubuntu`, `debian`, `clang-analyzer`
- I built the project with all plugins enabled
So, my questions are:
1. Is there some other way of project configuration which I should run before trying to remove the global warning suppression?
2. How should I proceed with the unused parameters in `plugins/experimental/wasm/lib/src/signature_util.cc`? I mean, as I understood, it's not OK to change the code there because it comes from 3rd party libraries. Is it OK then to add the warning suppression locally in the `plugins/experimental/wasm/lib/CMakeLists.txt`? Or should I do something else? Because it'll start complaining when/if the global suppression is removed.

cmcfarlen:
Hello @freak82 

I think I would try to remove the flag where it is set globally and then add it back where the cmake build goes into external code (i.e. lib/CMakeLists.txt).  CMake will only change the compiler flags for configs "at and below" where you set them, so this would not affect the parent or sibling directories, but would for any subdirectories.  I do wonder if we should keep the external code in the wasm plugin where it is or move that into the lib directory.  That is a question for @shukitchan I believe.  

Generally if you have questions, you can ask them in the apache slack in the traffic-server channel.  We can get you an invite for that if you don't have one.

Thanks so much for these cleanup contributions.  We really appreciate your help!

cmcfarlen:
Cherry-picked to v10.0.x

PR #11494, opened 2024-07-01
records.yaml - Some more cleanup from the records.yaml root key rename.

brbzull0:


bryancall (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11495, opened 2024-07-01
Use "git rev-parse --absolute-git-dir" in tools scripts.

ywkaras:
The alternate way of getting this, that we are currently using, doesn't work in cmake version 3.28.

### DIFF
--- a/tools/autopep8.sh
+++ b/tools/autopep8.sh
@@ -39,8 +39,8 @@ function main() {
     pip install -q virtualenv
   fi
 
-  GIT_COMMON_DIR=$(cd $(dirname $0) && git rev-parse --path-format=absolute --git-common-dir)
-  AUTOPEP8_VENV=${AUTOPEP8_VENV:-${GIT_COMMON_DIR}/fmt/autopep8_${AUTOPEP8_VERSION}_venv}
+  GIT_DIR=$(git rev-parse --absolute-git-dir)
+  AUTOPEP8_VENV=${AUTOPEP8_VENV:-${GIT_DIR}/fmt/autopep8_${AUTOPEP8_VERSION}_venv}
   if [ ! -e ${AUTOPEP8_VENV} ]
   then
     virtualenv ${AUTOPEP8_VENV}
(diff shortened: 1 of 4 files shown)


JosiahWI (changes requested):
Seems like a good improvement, but could we change the variable name?

tools/autopep8.sh
```diff
@@ -39,7 +39,7 @@ function main() {
     pip install -q virtualenv
   fi
 
-  GIT_COMMON_DIR=$(cd $(dirname $0) && git rev-parse --path-format=absolute --git-common-dir)
+  GIT_COMMON_DIR=$(git rev-parse --absolute-git-dir)
```
JosiahWI, line 42: The variable name is no longer correct, as this will correspond to `$GIT_DIR` and **not** to `$GIT_COMMON_DIR`. I think this should be fine, but could confirm that and update the name to reflect this?


> --absolute-git-dir
Like --git-dir, but its output is always the canonicalized absolute path.

JosiahWI (changes requested):

tools/autopep8.sh
ywkaras: So is it a cosmetic issue, that we follow some convention for naming this issue?  Or, do you mean we want to make it overridable from the process environment, like we do for AUTOPEP8_VENV on the next line?

JosiahWI: It's a readability issue. It's like the difference between `PROJECT_SOURCE_DIR` and `CMAKE_SOURCE_DIR` in CMake. They're not the same thing and in edge cases the difference matters. Your commit introduces a behavioral change to the script, and the name of the `GIT_COMMON_DIR` variable is lying about what itself. It's not the git common dir anymore.

JosiahWI: > GIT_COMMON_DIR
If this variable is set to a path, non-worktree files that are normally in $GIT_DIR will be taken from this path instead. Worktree-specific files such as HEAD or index are taken from $GIT_DIR. See [gitrepository-layout[5]](https://git-scm.com/docs/gitrepository-layout) and [git-worktree[1]](https://git-scm.com/docs/git-worktree) for details. This variable has lower precedence than other path variables such as GIT_INDEX_FILE, GIT_OBJECT_DIRECTORY…​

ywkaras: But, given that it's not being put into the environment, is GIT_DIR also not an appropriate name?

JosiahWI: I would be fine with `GIT_DIR`. Would shadowing the environment variable be harmful? The previous name was shadowing one, so it would be consistent with the current pattern to simply rename to `GIT_DIR` I think.

JosiahWI (commented):
Thank you for changing the variable name. Looks good to me and makes the scripts easier to follow. Will let a non-Yahoo person approve.

JosiahWI (commented):
Dismiss pending change request as it has been addressed.

JosiahWI (approved):
I think GitHub has a small issue. Giving explicit approval to dismiss the change request. PR still needs approval from a non-Yahoo person.

cmcfarlen (approved):
This works fine on macOS

cmcfarlen:
Cherry-picked to v10.0.x

PR #11499, opened 2024-07-01
multiplexer: fix consume of too many bytes

bneradt:
IOBufferReader::consume contractually assumes that all callers of it will not consume more bytes than read_avail for the buffer chain. This adds a debug assertion for this and fixes multiplexer so that it doesn't violate this invariant.

### DIFF
--- a/plugins/multiplexer/fetcher.h
+++ b/plugins/multiplexer/fetcher.h
@@ -213,6 +213,8 @@ template <class T> struct HttpTransaction {
             self->t_.header(self->parser_.buffer_, self->parser_.location_);
             self->parsingHeaders_ = false;
           }
+          // Parsing headers will indirectly read from our reader. Update available accordingly.
+          available = TSIOBufferReaderAvail(self->in_->reader);
         }
         if (!self->parsingHeaders_) {
           if (self->chunkDecoder_ != NULL) {
--- a/src/iocore/eventsystem/P_IOBuffer.h
+++ b/src/iocore/eventsystem/P_IOBuffer.h
@@ -551,6 +551,7 @@ IOBufferReader::is_read_avail_more_than(int64_t size)
 TS_INLINE void
 IOBufferReader::consume(int64_t n)
 {
+  ink_assert(read_avail() >= n);
   start_offset += n;
   if (size_limit != INT64_MAX) {
     size_limit -= n;


src/iocore/eventsystem/P_IOBuffer.h
```diff
@@ -551,6 +551,7 @@ IOBufferReader::is_read_avail_more_than(int64_t size)
 TS_INLINE void
 IOBufferReader::consume(int64_t n)
 {
+  ink_assert(read_avail() >= n);
```
masaori335, line 554: It looks like this function should handle this case, a). handle it as error case or b). consume only `read_avail()`. This assertion is a good start 👍

masaori335 (dismissed):
Look good.

plugins/multiplexer/fetcher.h
```diff
@@ -219,14 +220,17 @@ template <class T> struct HttpTransaction {
             available = self->chunkDecoder_->decode(self->in_->reader);
             if (available == 0) {
               self->t_.data(self->in_->reader, available);
```
masaori335, line 222: I'm not familiar with this plugin, so I'd clarify a bit. What this `t_.data()` does?

Is it ok, calling `TSIOBufferReaderAvail` *after* the `t_.data()`? The order of the call should be like this?
```
              available = TSIOBufferReaderAvail(self->in_->reader);
              self->t_.data(self->in_->reader, available);
              TSIOBufferReaderConsume(self->in_->reader, available);
```

masaori335: It looks like if the `available` is larger than actual `read_avail()`, the `length += l;` will be problem.
https://github.com/apache/trafficserver/blob/7bea1d655e92ee62a6ffd997a457c9338c3f60ef/plugins/multiplexer/dispatch.cc#L185

src/iocore/eventsystem/P_IOBuffer.h
bneradt: I'm open to suggestions, but I don't think there's a good way to handle this, not here in consume. There's not enough context. In production, we don't want to spend the time in the `read_avail()`. If needed the caller should do that as a part of fulfilling its contract for this API. The debug assert here seems like a reasonable way to catch issues in debug builds.

plugins/multiplexer/fetcher.h
bneradt: That `length` is used for a statistic. And it is fine to increment it by `l` here because the `available` value passed in is correct at that point in the code.

The problem is that the `data()` call, after the increment mentioned, may read from the reader. Which is fine, but then `available` in the calling scope obviously must change in order for future calls to `data()` to be accurate. Thus the update to `available` after the call to `data()`.

serrislew: > available value passed in is correct at that point in the code

How can you verify that it is always correct at this point in the code? Does it make sense to do some verification of `l` against the buffer reader `r`?

bneradt: I suppose we can add a debug assertion. To me the code reads that the implied contract is that `l` is less than or equal to the number of available bytes in the reader.

I'll add a debug assertion.

Thanks for the review.

bneradt: Actually, digging into this further, I notice that the issue with multiplexer is not `data()` at all but pulling bytes from the reader for parsing the headers. I've updated (the much simpler) patch.

Thanks again for the review @serrislew !

serrislew (approved):

cmcfarlen:
Cherry-picked to v10.0.x

ezelkow1:
do we still need this on 9.2?

bneradt:
> do we still need this on 9.2?

This one liner is pretty valuable since it keeps multiplexer from potentially reading past the end of a buffer. We've never observed an issue in production, but better safe than sorry.

bneradt:
9.2.x PR:
https://github.com/apache/trafficserver/pull/11965

PR #11501, opened 2024-07-02
Fix for OpenSSL 3.2+.

ywkaras:
It seems BIO user data is no longer initialized to nullptr.


src/iocore/net/SSLNetVConnection.cc
```diff
@@ -212,9 +212,9 @@ SSLNetVConnection::_make_ssl_connection(SSL_CTX *ctx)
       BIO *bio = BIO_new(const_cast<BIO_METHOD *>(BIO_s_fastopen()));
       BIO_set_fd(bio, this->get_socket(), BIO_NOCLOSE);
 
-      if (this->options.f_tcp_fastopen) {
-        BIO_set_conn_address(bio, this->get_remote_addr());
-      }
+      BIO_set_conn_address(bio, this->options.f_tcp_fastopen ?
+                                  const_cast<void *>(static_cast<const void *>(this->get_remote_addr())) :
+                                  const_cast<void *>(static_cast<const void *>(nullptr)));
```
ywkaras, line 217: The casts were required to compile for Rocky Linux.

maskit:
> It seems BIO user data is no longer initialized to nullptr.

Yeah, and it's initialized to something else. This change probably cause a memory leak and  frees something that should not be freed by OpenSSL, if TFO is enabled.
https://github.com/openssl/openssl/blob/77d716d3be95317ade5a7b8dda0b4946522c5a34/crypto/bio/bss_sock.c#L94

ywkaras:
[approve ci autest]

maskit:
How about using `BIO_set_ex_data`?

ywkaras:
> How about using `BIO_set_ex_data`?

It looks like it's deprecated in OpenSSL 3:  https://www.openssl.org/docs/man3.2/man3/BIO_set_ex_data.html

maskit:
Ones for BIO are not, right?

ywkaras:
> Ones for BIO are not, right?

Ah, I see now, it's replaced by CRYPTO_new/set_ex_data in 3.0+.  Maybe useable with some ninja #ifdefs.

src/iocore/net/BIO_fastopen.cc
```diff
@@ -21,12 +21,72 @@
...
+#if !defined(OPENSSL_IS_BORINGSSL)
+
+class ExData
+{
+public:
+  static int
+  idx()
+  {
+    static int idx_ = []() -> int {
+      int i = CRYPTO_get_ex_new_index(CRYPTO_EX_INDEX_BIO, 0, nullptr, _new, _dup, _free);
```
maskit, line 45: I don't understand the necessity of `ExData` class in the first place, but even if it's needed, I don't see a reason to use `CRYPTO_get_ex_new_index` instead of `BIO_get_ex_new_index`.

maskit:
What would be the issue if we simply use `BIO_get_ex_data` instead of `BIO_get_data`?

src/iocore/net/BIO_fastopen.cc
```diff
@@ -21,12 +21,72 @@
...
+  _dup(CRYPTO_EX_DATA * /* to */, const CRYPTO_EX_DATA * /* from */, _Type_from_d /* from_d */, int /* idx */, long /* argl */,
+       void * /* argp */)
+  {
+    ink_assert(false);
+    return 0;
+  }
+};
+
+#undef BIO_set_data
+#define BIO_set_data(BIO, PTR) ink_assert(BIO_set_ex_data((BIO), ExData::idx(), (PTR)) == 1)
```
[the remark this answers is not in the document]
ywkaras: These definitions only apply within this file (BIO_fastopen.cc).  Boringssl does not provide BIO_set/get_ex_data.  It seemed like the best approach to avoid further #ifdefs for OPENSSL_IS_BORINGSSL in this file.

src/iocore/net/BIO_fastopen.cc
ywkaras: OK I will switch to BIO_get_ex_new_index, apparently it's not deprecated.  The ExData class is for scoping and to handle C++ order of initialization fiasco.

ywkaras:
> What would be the issue if we simply use `BIO_get_ex_data` instead of `BIO_get_data` regardless of the SSL library and its version.

Not supported in Boring:
```
wkaras /opt/boringssl
$ grep 'BIO_.*_ex' $(findsrc)
wkaras /opt/boringssl
$ 
```

ywkaras:
[approve ci autest]

src/iocore/net/BIO_fastopen.cc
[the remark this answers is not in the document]
maskit: https://commondatastorage.googleapis.com/chromium-boringssl-docs/bio.h.html#ex_data-functions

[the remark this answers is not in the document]
ywkaras: It's only been supported for about 8 months:  https://github.com/google/boringssl/commit/2139aba2e3e28cd1cdefbd9b48e2c31a75441203 .   This commit does not seem to be in the version of BoringSSL we are using in the Rocky CI job.  I guess we could add a CMake check for it.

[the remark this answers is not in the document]
maskit: Bummer. I update BoringSSL once in a while so I didn't think it's unavailable.

I'd rather update the BoringSSL requirement if that's the reason. Oldest commit we currently require is https://boringssl.googlesource.com/boringssl/+/a1843d660b47116207877614af53defa767be46a
but I don't think there's a big reason to stick with the commit.
https://lists.apache.org/thread/d6ymfjhpmjogb5stnz92hwhhps909809


What do you think? Would it make the code simple?

src/iocore/net/BIO_fastopen.cc
```diff
@@ -167,18 +249,18 @@ fastopen_bread(BIO *bio, char *out, int outsz)
...
 {
+#if !defined(OPENSSL_IS_BORINGSSL) && (OPENSSL_VERSION_MAJOR >= 3)
   switch (cmd) {
-  case BIO_C_SET_CONNECT:
-    // We only support BIO_set_conn_address(), which sets a sockaddr.
-    ink_assert(larg == 2);
-    BIO_set_data(bio, ptr);
-    return 0;
+  case BIO_C_SET_FD:
+    BIO_set_shutdown(bio, int(larg));
```
maskit, line 255: Shutdown??

src/iocore/net/BIO_fastopen.cc
```diff
@@ -167,18 +249,18 @@ fastopen_bread(BIO *bio, char *out, int outsz)
 static long
 fastopen_ctrl(BIO *bio, int cmd, long larg, void *ptr)
 {
+#if !defined(OPENSSL_IS_BORINGSSL) && (OPENSSL_VERSION_MAJOR >= 3)
```
maskit, line 252: It'd be nice if we could avoid using `OPENSSL_IS_BORINGSSL`. Otherwise I'd have to add `OPENSSL_IS_AWSLC`. What's the context of this check?

ywkaras: The OPENSSL_IS_BORINGSSL is already used in many places in ATS code.  It is controlled by this cmake code:  https://github.com/apache/trafficserver/blob/master/cmake/CheckOpenSSLIsBoringSSL.cmake .

src/iocore/net/BIO_fastopen.cc
ywkaras: I had to add that code to avoid triggering this assert:  https://github.com/apache/trafficserver/blob/68a5ee56deee9c76c0204f046883f2752f1d7c1e/src/iocore/net/BIO_fastopen.cc#L100

ywkaras:
[approve ci osx]

src/iocore/net/BIO_fastopen.cc
ywkaras: https://github.com/google/boringssl/blob/b34976cae99f8d1b864dbab31e20fc00d06acb09/crypto/bio/fd.c#L230

https://github.com/google/boringssl/blob/b34976cae99f8d1b864dbab31e20fc00d06acb09/crypto/bio/bio.c#L234

src/iocore/net/BIO_fastopen.cc
maskit: That doesn't answer my question. I know the check exists and you are not the only person who use it. I had to add one for AWS-LC because `OPENSSL_IS_BORINGSSL` is used. If everything is done by checks for technical requirements (i.e. not by the library name), I didn't have to do that.

My question is why this block is only needed by OpenSSL 3+.

ywkaras: It's necessary to avoid triggering the ink_assert(), see below.

src/iocore/net/BIO_fastopen.cc
maskit: Ok, I misunderstood what "shutdown" means here, and now setting it makes sense. But why only on OpenSSL 3+?

The odd thing is that it looks like `BIO_meth_get_ctrl` returns the ctrl function for Socket BIO, and it sets the shutdown flag. How does that not happen?

src/iocore/net/BIO_fastopen.cc
ywkaras: I wan't aware we also have to support a third TLS library.

src/iocore/net/BIO_fastopen.cc
ywkaras: I need to retest this.  Looking at the openssl code, it shouldn't be necessary.

src/iocore/net/BIO_fastopen.cc
maskit: As the community we don't have the consensus to support AWS-LC and other variations. It's just I don't think checking the library name is not a great idea in general where there are many forks of SSL library. That's why I was asking just the context so that I can find a better way later.

src/iocore/net/BIO_fastopen.cc
ywkaras: My testing was bogus, that change isn't necessary.

ywkaras:
[approve ci format]

maskit (changes requested):
Overall looks good, but this does not build with a BoringSSL which has `BIO_get_ex_new_index`.
```
/Users/mkitajo/src/github.com/trafficserver/src/iocore/net/BIO_fastopen.cc:48:15: error: no matching function for call to 'BIO_get_ex_new_index'
      int i = BIO_get_ex_new_index(0, nullptr, _new, _dup, _free);
              ^~~~~~~~~~~~~~~~~~~~
/Users/mkitajo/opt/boringssl/include/openssl/bio.h:819:20: note: candidate function not viable: no known conversion from 'void (void *, void *, CRYPTO_EX_DATA *, int, long, void *)' (aka 'void (void *, void *, crypto_ex_data_st *, int, long, void *)') to 'CRYPTO_EX_unused *' (aka 'int *') for 3rd argument
OPENSSL_EXPORT int BIO_get_ex_new_index(long argl, void *argp,
                   ^
/Users/mkitajo/src/github.com/trafficserver/src/iocore/net/BIO_fastopen.cc:59:24: error: use of undeclared identifier 'CRYPTO_set_ex_data'
    ink_release_assert(CRYPTO_set_ex_data(ad, idx_, nullptr) == 1);
                       ^
/Users/mkitajo/src/github.com/trafficserver/src/iocore/net/BIO_fastopen.cc:65:24: error: use of undeclared identifier 'CRYPTO_set_ex_data'
    ink_release_assert(CRYPTO_set_ex_data(ad, idx_, nullptr) == 1);
                       ^
3 errors generated.
```

src/iocore/net/BIO_fastopen.cc
```diff
@@ -21,12 +21,117 @@
...
+#error
+#elif defined(OPENSSL_IS_BORINGSSL) || defined(OPENSSL_IS_AWSLC)
+#define BORINGLIKE 1
+#else
+#define BORINGLIKE 0
+#endif
+
+#if defined(HAVE_BIO_GET_EX_NEW_INDEX) && defined(HAVE_BIO_GET_EX_DATA) && defined(HAVE_BIO_SET_EX_DATA)
+
+// Thwart boringssl's thwarting of ex data for custom BIOs.  Boring fails to export these functions.
```
maskit, line 43: I agree that BoringSSL's API is broken, but do we need to use these on code for BoringSSL?

ywkaras: What alternative do you propose?

maskit: Remove theses exports, and use empty callbacks for BoringSSL.

maskit: I don't quite understand why OpenSSL says this though. BoringSSL doesn't say this.

> When setting exdata, the pointer specified with a particular index is saved, and returned on a subsequent "get" call. If the application is going to release the data, it must make sure to set a NULL value at the index, to avoid likely double-free crashes.

bneradt:
[approve ci]

src/iocore/net/BIO_fastopen.cc
```diff
@@ -21,12 +21,117 @@
...
+      ink_release_assert(i >= 0);
+      return i;
+    }();
+    return idx_;
+  }
+
+private:
+#if BORINGLIKE
+  static const CRYPTO_EX_unused _new{};
+  // Hopefully Boring intializes ex data pointers to nullptr by default.
```
maskit, line 66: How about setting `nullptr` in `fastopen_create` by `BIO_set_ex_data`? I'm not sure if this is really possible. It could be too early to call `BIO_set_ex_data`.

maskit: Another option is setting nullptr after making the BIO on SSLNetVC side. It'd be a silly requirement if this BIO is publicly available for others though.

ywkaras: I made a change in SSLNetVConnection.cc to initialize the ex data to either the destination address or nullptr for all BIOs using the fastopen methods.

src/iocore/net/BIO_fastopen.cc
ywkaras: I don't know why either.  In our case, the ex data is not a pointer to an allocated resource, so there is no risk of a double deallocation or a leak.  I only clear the ex data pointer to know it was free'd when looking at a coredump.

src/iocore/net/BIO_fastopen.cc
```diff
@@ -21,12 +21,116 @@
...
+      int i = BIO_get_ex_new_index(0, nullptr, _new, _dup, _free);
+      ink_release_assert(i >= 0);
+      return i;
+    }();
+    return idx_;
+  }
+
+private:
+#if BORINGLIKE
+  static const CRYPTO_EX_unused _new{};
```
maskit, line 65:
```
[ 80%] Building CXX object src/iocore/net/CMakeFiles/inknet.dir/BIO_fastopen.cc.o
/Users/mkitajo/src/github.com/trafficserver/src/iocore/net/BIO_fastopen.cc:56:15: error: no matching function for call to 'BIO_get_ex_new_index'
      int i = BIO_get_ex_new_index(0, nullptr, _new, _dup, _free);
              ^~~~~~~~~~~~~~~~~~~~
/Users/mkitajo/opt/boringssl/include/openssl/bio.h:819:20: note: candidate function not viable: no known conversion from 'const CRYPTO_EX_unused' (aka 'const int') to 'CRYPTO_EX_unused *' (aka 'int *') for 3rd argument
OPENSSL_EXPORT int BIO_get_ex_new_index(long argl, void *argp,
                   ^
1 error generated.
```

```suggestion
  static constexpr CRYPTO_EX_unused *_new{nullptr};
```

ywkaras:
[approve ci autest]

maskit (approved):

cmcfarlen:
Cherry-picked to v10.0.x

Issue #11498, opened 2024-07-01
File configs/records.yaml.default.in needs s/ts/records/


cmcfarlen:
This file still has the old `ts` top level key.

PR #11502, opened 2024-07-02
records.yaml - Some more cleanup after renaming the root node.

brbzull0:
fixes https://github.com/apache/trafficserver/issues/11498


cmcfarlen (approved):
Thanks!

PR #11507, opened 2024-07-02
wasm plugin: Initialize SanitizationConfig  

shukitchan:


shukitchan:
@traeak can you please take a look?
the /lib contents all come from the proxy-wasm project and I would like to keep them untouched for now. 

Can the fix be done in this way and be ok? Thanks.

traeak (approved):
These look fine and will isolate the changes.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11511, opened 2024-07-03
Prioritize the library path for SSL library

maskit:
SSL library in `OPENSSL_ROOT_DIR` is not be used/linked if PCRE and another SSL library are in the system default directory, because the linker finds the one in the system default directory first. The change tweak the library path priority so that linker can pick the one in `OPENSSL_ROOT_DIR`.

This is not a perfect solution. If a user wants to use the SSL library in the system default directory and PCRE in somewhere else, the user will face the same issue but for PCRE. But using a custom PCRE for ATS is uncommon, so it wouldn't likely be an issue.

### DIFF
--- a/src/api/CMakeLists.txt
+++ b/src/api/CMakeLists.txt
@@ -23,6 +23,7 @@ set(TSAPI_PUBLIC_HEADERS ${PROJECT_SOURCE_DIR}/include/ts/ts.h ${PROJECT_SOURCE_
                          ${PROJECT_SOURCE_DIR}/include/ts/TsException.h ${PROJECT_BINARY_DIR}/include/ts/apidefs.h
 )
 
+# OpenSSL needs to be listed in before PCRE and other libraries that can be found in the system default lib directory (See # #11511)
 target_link_libraries(tsapi PRIVATE libswoc::libswoc yaml-cpp::yaml-cpp OpenSSL::SSL PCRE::PCRE PkgConfig::PCRE2)
 set_target_properties(tsapi PROPERTIES PUBLIC_HEADER "${TSAPI_PUBLIC_HEADERS}")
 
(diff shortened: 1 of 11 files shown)


src/api/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ set(TSAPI_PUBLIC_HEADERS ${PROJECT_SOURCE_DIR}/include/ts/ts.h ${PROJECT_SOURCE_
                          ${PROJECT_SOURCE_DIR}/include/ts/TsException.h ${PROJECT_BINARY_DIR}/include/ts/apidefs.h
 )
 
-target_link_libraries(tsapi PRIVATE libswoc::libswoc yaml-cpp::yaml-cpp PCRE::PCRE PkgConfig::PCRE2 OpenSSL::SSL)
+target_link_libraries(tsapi PRIVATE libswoc::libswoc yaml-cpp::yaml-cpp OpenSSL::SSL PCRE::PCRE PkgConfig::PCRE2)
```
JosiahWI, line 26: The link order appears arbitrary and I think someone not thinking about it wouldn't realize its importance (this PR to fix it exists, after all). Could we add a comment so someone doesn't accidentally cause a regression?

JosiahWI (changes requested):

src/api/CMakeLists.txt
maskit: It turned out that there are several places potentially problematic. I'm not sure where would be a good place to put the comment on. I updated the documentation for plugin development for now.

JosiahWI: I think it'd still be good to have a comment here since this is in core and not covered by the plugin documentation.

What about the plugins that link `OpenSSL::Crypto`? I suspect those should also be updated.

JosiahWI:
What if we create our own PCRE target that depends on OpenSSL? Would that ensure the link order is correct everywhere?

src/api/CMakeLists.txt
maskit: > What about the plugins that link OpenSSL::Crypto? I suspect those should also be updated.

Good point. I updated some more places for it. Thanks.

And I added the comment here. Better than not having any.

JosiahWI (approved):
If we have libraries in multiple directories all seen by the linker, it will pick the first one. Seems there's no easy solution to this for all cases, but this fix will patch it for the common case of having OpenSSL installed in multiple locations. Thanks for making the build more robust!

cmcfarlen:
Cherry-picked to v10.0.x

PR #11512, opened 2024-07-03
Fix quick_server autest to verify server response

bneradt:
The quick_server.test.py autest incorrectly sent a literal '\r\n' four bytes instead of the expected 2 bytes. The test still largely functioned as needed, but didn't verify that the expected respone made it to the client. This updates the test to verify that the 200 OK response makes it to the client as expected. ATS behaves as expected, so there is no production change needed.

### DIFF
--- a/tests/gold_tests/slow_post/quick_server.py
+++ b/tests/gold_tests/slow_post/quick_server.py
@@ -67,9 +67,9 @@ def send_response(sock: socket.socket, abort_early: bool) -> None:
     if abort_early:
         response = "HTTP/1."
     else:
-        response = (r"HTTP/1.1 200 OK\r\n"
-                    r"Content-Length: 0\r\n"
-                    r"\r\n")
+        response = ("HTTP/1.1 200 OK\r\n"
+                    "Content-Length: 0\r\n"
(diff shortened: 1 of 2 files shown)


tests/gold_tests/slow_post/quick_server.py
```diff
@@ -67,9 +67,9 @@ def send_response(sock: socket.socket, abort_early: bool) -> None:
     if abort_early:
         response = "HTTP/1."
     else:
-        response = (r"HTTP/1.1 200 OK\r\n"
-                    r"Content-Length: 0\r\n"
-                    r"\r\n")
+        response = ("HTTP/1.1 200 OK\r\n"
+                    "Content-Length: 0\r\n"
+                    "\r\n")
```
bneradt, line 72: This is the key change: it was a mistake to use raw strings here.

maskit (approved):

cmcfarlen:
Cherry-picked to v10.0.x

ezelkow1:

bneradt:
> Do we still need this for 9.2?

I think we might as well fix the test and keep the test in sync with master.

PR #11517, opened 2024-07-08
Try TCP_FASTOPEN socket option with mptcp

serrislew:
With support for TCP_FASTOPEN in newer kernels, allow ATS to try to use TCP_FASTOPEN with mptcp enabled

### DIFF
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -5075,8 +5075,8 @@ Sockets
         TCP_NOTSENT_LOWAT (64)
         INCOMING_CPU (128)
 
-   Note: If MPTCP is enabled, TCP_NODELAY is only supported on Linux kernels 5.17+. TCP_FASTOPEN
-   and TCP_NOTSENT_LOWAT socket options are currently not supported.
+   Note: If MPTCP is enabled, TCP_NODELAY is only supported on Linux kernels 5.17+ and TCP_FASTOPEN is
+   only supported on Linux kernels 6.2+. TCP_NOTSENT_LOWAT socket option is currently not supported.
 
 .. note::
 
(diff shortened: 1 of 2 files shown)


JosiahWI:
Where is `TCP_FASTOPEN` defined?

```
% grep -Ir "TCP_FASTOPEN"
./doc/admin-guide/files/records.yaml.en.rst:        TCP_FASTOPEN (8)
./doc/admin-guide/files/records.yaml.en.rst:   Note: If MPTCP is enabled, TCP_NODELAY is only supported on Linux kernels 5.17+. TCP_FASTOPEN
./doc/admin-guide/files/records.yaml.en.rst:        TCP_FASTOPEN (8)
./src/proxy/http/HttpProxyServerMain.cc:#ifdef TCP_FASTOPEN
./src/iocore/net/Connection.cc:#ifdef TCP_FASTOPEN
./src/iocore/net/Connection.cc:      Warning("[Server::listen] TCP_FASTOPEN socket option not valid on MPTCP socket level");
./src/iocore/net/Connection.cc:    } else if (safe_setsockopt(fd, IPPROTO_TCP, TCP_FASTOPEN, (char *)&opt.tfo_queue_length, sizeof(int))) {
```

We don't define it anywhere ourselves. Is it defined in a system header?

serrislew:
yes @JosiahWI  it's defined in the system header files `tcp.h`, like all of the other TCP socket options

JosiahWI (commented):
The [POSIX specification](https://pubs.opengroup.org/onlinepubs/9699919799/) doesn't list `EOPNOTSUPP` as a case where `setsockopt` (our `safe_setsockopt` wraps it in a retry on interrupt) shall or may fail. I don't see that listed as a possible error code for a `setsockopt` call by [FreeBSD](https://man.freebsd.org/cgi/man.cgi?query=setsockopt&sektion=2) or [Apple](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/setsockopt.2.html) either. Do we need to check it? Is it just to be defensive?

JosiahWI:
@serrislew

JosiahWI (approved):
Thank you for adding support for fast open with mptcp, looks good!

cmcfarlen:
Cherry-picked to v10.0.x

PR #11519, opened 2024-07-08
Deprecate the support for URL param segment

maskit:
This is a PR for the change I proposed on the lists. See the archive for the detail.
https://lists.apache.org/thread/sgxf150vpsmorclg4p3504o0kn7yfhzy

This keeps the variables for params in `URLImpl` class and its marshaling functions, because removing those might cause cache data corruption.

### DIFF
--- a/include/ts/ts.h
+++ b/include/ts/ts.h
@@ -667,6 +667,9 @@ TSReturnCode TSUrlFtpTypeSet(TSMBuffer bufp, TSMLoc offset, int type);
     argument. Note: the returned string is not guaranteed to be
     null-terminated.
 
+    This function is deprecated and returns empty string.
+    Plugins that need "params" can call TSUrlPathGet to get a whole path string to parse it.
+
     @param bufp marshal buffer containing the URL.
     @param offset location of the URL.
     @param length of the returned string.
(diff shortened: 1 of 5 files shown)


src/proxy/hdrs/URL.cc
```diff
@@ -551,8 +551,22 @@ URLImpl::set_path(HdrHeap *heap, const char *value, int length, bool copy_string
 void
 URLImpl::set_params(HdrHeap *heap, const char *value, int length, bool copy_string)
```
bryancall, line 552: Why do we still have set_params if we don't support params anymore?

include/ts/ts.h
```diff
@@ -683,6 +686,9 @@ const char *TSUrlHttpParamsGet(TSMBuffer bufp, TSMLoc offset, int *length);
     copies the string to within bufp, so you can modify or delete
     value after calling TSUrlHttpParamsSet().
 
+    This function is deprecated. The value passed will be internally appended to the path portion.
+    Thus, TSUrlHttpParamsGet will not return the value set by this function.
```
bryancall, line 690: I would like to see us remove the TSUrlHttpParamsGet() instead of passing back an empty string.   People should change the logic in their plugins based on the change instead of it continuing to build and possibly discovering an issue later in production.

src/proxy/hdrs/URL.cc
maskit: To keep compatibility as possible as we can, since we only deprecate them at this time.

include/ts/ts.h
maskit: Although that is a valid option, that would make the migration to 10.0 hard (plugins have to be changed to upgrade ATS). And we'd have to change the Lua API, Cripts, and url_sig plugin.

Everybody has to change the logic in their plugins at some point, if they use Get/SetParams, but this deprecation gives grace period.

What issues would be found? The APIs are not working as expected in the first place. And this change fixes a common mistake under the hood without fixing plugins.

bryancall (approved):

maskit:
We'll land this as the first step, and like announced on the dev list, remove the APIs for 10.0.

cmcfarlen:
Cherry-picked to v10.0.x

Issue #11468, opened 2024-06-21
traffic_ctl: Timeout connection to jsonrpc socket


cmcfarlen:
I had a situation where traffic_server was unhappy but some tools were trying to run traffic_ctl.  These all blocked on the connect to the jsonrpc socket and ended up with tons of traffic_ctl instances (and too many open fd errors).  Could/should we add a timeout to that connect?  The tool running traffic_ctl was also misbehaving, but this would be nice to have.

PR #11523, opened 2024-07-09
Tools: JSON-RPC - Change the Unix Domain Socket client to retry and wait when connection to the server.

brbzull0:
Very simple mechanism to wait and retry when the connect to the server socket is temporarily unavailable.
This should not block on `connect`. This logic is simpler than adding a `select` to wait on the `fd` as most of the time it will tell us it's ready after the first `EAGAIN` by  the `connect` but it does not mean that we can just write to the unix domain socket as the enpoint may not be ready.

Eventually if needed I can add some polling on the `read` to avoid hangung up there, but it's unlikely to happen.

Fixes: https://github.com/apache/trafficserver/issues/11468

### DIFF
--- a/include/shared/rpc/IPCSocketClient.h
+++ b/include/shared/rpc/IPCSocketClient.h
@@ -31,6 +31,7 @@
 
 namespace shared::rpc
 {
+using namespace std::chrono_literals;
 /// The goal of this class is abstract the Unix Socket implementation and provide a JSONRPC Node client for Tests and client's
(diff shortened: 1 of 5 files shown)


brbzull0:
`test_jsonrpcserver` is failing. Fixing it.

cmcfarlen (approved):
Thanks!

cmcfarlen:
Cherry-picked to v10.0.x

PR #11524, opened 2024-07-09
Fix unused parameters in the test plugins

freak82:
- Use `/* name ATS_UNUSED */` for fully unused parameters in C++ files 
- Use `[[maybe_unused]]` for conditionally unused parameters in C++ files 
- Use `name ATS_UNUSED` for unused parameters in C files

I wasn't sure if the use `(void)param` inside the functions or `ATS_UNUSED` for the C files which are compiled as such.
There can't be used `/* name ATS_UNUSED */` because this requires C2X build.
I decided to use `ATS_UNUSED` because it's closer to the way the C++ files are fixed.

### DIFF
--- a/src/tscore/hugepages.cc
+++ b/src/tscore/hugepages.cc
@@ -61,7 +61,7 @@ ats_hugepage_enabled()
 }
 
 void
-ats_hugepage_init(int enabled)
+ats_hugepage_init([[maybe_unused]] int enabled)
 {
 #ifdef MAP_HUGETLB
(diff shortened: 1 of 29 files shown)


JosiahWI (approved):
Thanks, looks good. This introduces a dependency on tscore to some AuTest plugins, but these are purposed for testing ATS and are therefore closely related to it, so the coupling seems acceptable to me. Many of the test plugins already have this dependency.

PR #11526, opened 2024-07-09
Do not retry to open cache store on ENOMEM

JosiahWI:
This replaces the `SocketManager::open` function with a new `safe_open` function in tscore that retries the call if it is interrupted a signal, but does not retry it if the kernel runs out of memory or buffers for a FIFO cannot be allocated. The change is split into three commits with descriptive comments.

### DIFF
--- a/include/tscore/ink_file.h
+++ b/include/tscore/ink_file.h
@@ -80,10 +80,7 @@
 //
 #define INK_FILEPATH_TRUENAME 0x20
 
-namespace
-{
 inline constexpr mode_t DEFAULT_OPEN_MODE{0644};
-} // end anonymous namespace
 
 int safe_open(char const *path, int oflag = O_RDWR | O_NDELAY | O_CREAT, mode_t mode = DEFAULT_OPEN_MODE);
 
--- a/include/tscore/ink_file.h
+++ b/include/tscore/ink_file.h
@@ -31,7 +31,7 @@
 
 #pragma once
 
-#include "tscore/ink_config.h"
+#include "tscore/ink_platform.h"
 
 #include <cstdio>
 #include <sys/types.h>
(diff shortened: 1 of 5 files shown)


JosiahWI:
[approve ci freebsd]

src/tscore/ink_file.cc
```diff
@@ -42,6 +42,21 @@ using ioctl_arg_t = union {
...
+ * This is a wrapper for the POSIX open() call that will retry the call if a
+ * transient error occurs.
+ */
+int
+safe_open(char const *path, int oflag, mode_t mode)
+{
+  int r{-1};
+  do {
+    r = open(path, oflag, mode);
+  } while ((r < 0) && (errno == EINTR));
```
JosiahWI, line 56: The `ENOBUFS` error code isn't checked because it's not documented as a possible error condition for `open()`.

JosiahWI:
[approve ci freebsd]

include/tscore/ink_file.h
```diff
@@ -80,6 +80,13 @@
 //
 #define INK_FILEPATH_TRUENAME 0x20
 
+namespace
```
ywkaras, line 83: Combining an anonymous namespace with `inline` is redundant.

src/tscore/ink_file.cc
```diff
@@ -42,6 +42,21 @@ using ioctl_arg_t = union {
...
+  return r;
+}
+
```
ywkaras, line 59: I'm not sure why this was originally inlined in the header.  But remember that functions declared with inline are not necessarily inlined.  If the compiler considers it non-optimal to actually inline, if will put the function in a vague linkage section, which prevents multiple copies of it from being present in the linked executable:  https://gcc.gnu.org/onlinedocs/gcc/Vague-Linkage.html .

JosiahWI: I think we wouldn't usually put definitions in the header file unless we think they may be useful to inline. Should we keep this in the header file with the `inline` declaration, then?

ywkaras: Personally I'm fine with it not being inline.

ywkaras:
Looks good to me, just needs a non-Yahoo approval.

maskit (approved):

PR #11534, opened 2024-07-10
Proxy Verifier: Update to version v2.12.0

bneradt:
This new Proxy Verifier release includes the following that will be helpful for our tests:

* --poll-timeout: a new command line argument to specify what the socket timeout should be. The previous hard-coded value, and the current default value, is 5000 milliseconds.
* close-on-goaway: a new session directive for HTTP/2 connections that directs whether the verifier client will close the connection on a GOAWAY frame after processing outstanding streams, or continue to try to send the other specified streams from the replay file. The former behavior is the default as that is more in keeping with the RFC.
* HTTP/2 Expect: 100-Continue fixes.


maskit (approved):

cmcfarlen:
Cherry-picked to v10.0.x

ezelkow1:

bneradt:
> Do we still need this for 9.2?

I think things are easier in CI and with other backports if we keep these in sync with master. It's a test only change, so should be safe.

PR #11537, opened 2024-07-11
JSONRPC Client - Fix OSX/FreeBSD write socket issue.

brbzull0:
Tested on FreeBSD as I don't have OSX

### DIFF
--- a/include/shared/rpc/IPCSocketClient.h
+++ b/include/shared/rpc/IPCSocketClient.h
@@ -85,6 +85,7 @@ struct IPCSocketClient {
   }
 
 protected:
+  ssize_t            _safe_write(int fd, const char *buffer, int len);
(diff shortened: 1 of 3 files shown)


maskit:
I assume it actually called `safe_write` in ink_sock.h. This is why I use `this->`.

maskit:
This implies that we cannot use safe_write in ink_sock.cc for other use cases either. That would be a trap.

Could you provide the detail of the issue? I feel like we should fix the one in ink_sock.cc (maybe with ifdef if needed).

maskit (approved):
I have a concern but approve this not to block other PRs

JosiahWI:
Merging with Damian's permission.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11538, opened 2024-07-11
Remove dead locking logic in `HttpCacheSM`

JosiahWI:
This has no behavioral change. The two fields removed from `HttpCacheSM` were checked in a few places, but were never changed after their initial value.


bneradt (approved):
Pretty straightforward. Thanks for removing unused code.

ezelkow1 (approved):

ezelkow1:
[approve ci]

PR #11542, opened 2024-07-12
Reuse prepared cache write on redirected request

JosiahWI:
Fixes https://github.com/apache/trafficserver/issues/9275.

When there is no parent proxy and ATS is following a redirect, it will do
a cache lookup on the redirected request. If that lookup results in a cache
miss, it will start to prepare a new cache write, which is wrong. The state
machine intentionally leaves the write open so we can re-use the connection;
this is possible to ascertain from the code and from comments to that effect.

We should only open a new write if we don't already have one, when following
a redirect.

### DIFF
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -8043,9 +8043,8 @@ HttpSM::set_next_state()
   }
 
   case HttpTransact::SM_ACTION_CACHE_ISSUE_WRITE: {
-    ink_assert((cache_sm.cache_write_vc == nullptr) || t_state.redirect_info.redirect_in_process);
+    ink_assert(cache_sm.cache_write_vc == nullptr);
     HTTP_SM_SET_DEFAULT_HANDLER(&HttpSM::state_cache_open_write);
-
     do_cache_prepare_write();
     break;
   }
--- a/tests/gold_tests/redirect/redirect_to_same_origin_on_cache.test.py
+++ b/tests/gold_tests/redirect/redirect_to_same_origin_on_cache.test.py
@@ -1,0 +1,140 @@
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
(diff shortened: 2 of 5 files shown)


tests/gold_tests/redirect/redirect_to_same_origin_on_cache.test.py
```diff
@@ -0,0 +1,153 @@
...
+                'proxy.config.http.number_of_redirections': 1,
+            })
+
+        ts.Disk.remap_config.AddLine(f"map oof.com http://oof.backend.com:{server_one.Variables.http_port}")
+
+        # In this manual configuration of the cache, we were able to trigger
+        # an ink_assert during cache lookup.
+        #ts.Disk.storage_config.AddLine("storage 128M")
+        #ts.Disk.hosting_config.AddLine("hostname=* volume=1")
+        #ts.Disk.volume_config.AddLine("volume=1 scheme=http size=100% ramcache=false")
```
JosiahWI, line 128: Uncommenting these three lines of configuration will reproduce an ATS crash via an `ink_assert` in the cache lookup, which is unmitigated by this fix so far. It is possible to mitigate by skipping the cache lookup altogether when following a redirect, but we likely don't want to go back to origin every time we follow a redirect, if we could already have the new resource in cache.

src/proxy/http/HttpSM.cc
```diff
@@ -8043,9 +8043,8 @@ HttpSM::set_next_state()
   }
 
   case HttpTransact::SM_ACTION_CACHE_ISSUE_WRITE: {
-    ink_assert((cache_sm.cache_write_vc == nullptr) || t_state.redirect_info.redirect_in_process);
```
JosiahWI, line 8046: Am I missing something? This looks like the assertion went off and someone tweaked it so it would stop going off. Issuing a new write while an existing one is in progress is always going to be bad. This bug could've been verified and fixed ages ago if this blaring alarm hadn't been silenced.

JosiahWI:
[approve ci autest]

vmamidi (approved):

cmcfarlen:
Cherry-picked to v10.0.x

bryancall:
We need to close issue #10955 after cherry picking this PR to the 9.2.x branch.

PR #11546, opened 2024-07-13
CMakePresets: build for autest in Debug

bneradt:
Build ATS for autests using debug mode so debug-level assert checking is done. This can also help with debugging autest cores too since optimization will be off.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -54,6 +54,7 @@
       "inherits": ["default"],
       "binaryDir": "${sourceDir}/build-autest",
       "cacheVariables": {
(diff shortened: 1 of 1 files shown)


JosiahWI (commented):
I like this change. During my local debugging of redirection following errors, I discovered that one of my AuTests to reproduce it triggered debug assertions. I think it'd be good to have the assertions checked in CI - that's what they're there for.

bneradt:
[approve ci]

bneradt:
Looks like Debug is already picked up from `default`. So this change isn't needed.

PR #11536, opened 2024-07-11
Http/2: Fix a crash with empty bodied Expect: 100-continue

bneradt:
This fixes a crash in processing an empty-bodied HTTP/2 request with an Expect: 100-continue header. Such requests are technically non-RFC compliant, but we should not crash if they happen. With this patch, such requests will be handled as an otherwise compliant Expect: 100-Continue request. That is, the request will be forwarded to the origin, the origin's 100 Continue will be proxied to the client, and then the origin's 200 OK will also be proxied to the client.

Fixes #9857

---

This crash was introduced with the HTTP/2 to origin feature and therefore is not present in 9.x or before.

PR #11548, opened 2024-07-15
setup_server_read_response_header: remove assertion

bneradt:
This addresses an intermittent assertion failure associated with Expect:
100-Continue via the h2origin.test.py autest introduced via #11536.
Since the 100 Continue is followed immediately by 200 OK, it is possible
that both are processed at once if the Expect: 100-Continue request has
no body. In these cases, the assertion will fail because
state_read_server_response_header will have processed the entire
response. Since that is now possible, we might as well just remove the
assertion. In theory, it can be kept for the case when there is no
response bytes available to read, but in those situations the assertion
isn't really doing anything because there will be no complicated logic
from state_read_server_response_header and we set the server_entry->vc
earlier in the function.


maskit (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11545, opened 2024-07-12
Check incoming protocol list length first

maskit:
If the length is 0, there's no need to check the other list.

PR #11550, opened 2024-07-15
Update `UnixSocket` API for inkdns

JosiahWI:
This includes a number of updates to `UnixSocket` to make the API easier to use and more encompassing, with the goal of replacing `SocketManager`, and properly abstracting on top of ink_sock.h. The first and last commit refactor inkdns to depend on the new API, and remove its dependencies on both `SocketManager` and ink_sock.h.

### DIFF
--- a/include/iocore/eventsystem/UnixSocket.h
+++ b/include/iocore/eventsystem/UnixSocket.h
@@ -50,7 +50,7 @@
 bool transient_error();
 
 struct UnixSocket {
-  int sock_fd{NO_SOCK};
+  int fd{NO_SOCK};
 
   UnixSocket(int fd);
 
@@ -103,17 +103,17 @@ struct UnixSocket {
   static bool client_fastopen_supported();
 };
 
-inline UnixSocket::UnixSocket(int fd) : sock_fd{fd} {}
+inline UnixSocket::UnixSocket(int fd) : fd{fd} {}
 
 inline UnixSocket::UnixSocket(int domain, int type, int protocol)
 {
-  this->sock_fd = socket(domain, type, protocol);
+  this->fd = socket(domain, type, protocol);
 }
 
 inline bool
 UnixSocket::ok() const
 {
-  return NO_SOCK != this->sock_fd;
+  return NO_SOCK != this->fd;
 }
 
 inline std::int64_t
@@ -121,7 +121,7 @@ UnixSocket::read(void *buf, int size) const
 {
   std::int64_t r;
   do {
-    r = ::read(this->sock_fd, buf, size);
+    r = ::read(this->fd, buf, size);
     if (likely(r >= 0)) {
       break;
     }
@@ -135,7 +135,7 @@ UnixSocket::recv(void *buf, int size, int flags) const
 {
   int r;
   do {
-    if (unlikely((r = ::recv(this->sock_fd, static_cast<char *>(buf), size, flags)) < 0)) {
+    if (unlikely((r = ::recv(this->fd, static_cast<char *>(buf), size, flags)) < 0)) {
       r = -errno;
     }
   } while (r == -EINTR);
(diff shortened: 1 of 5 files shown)


JosiahWI:
The AuTest proxy_protocol failed because it couldn't acquire a session to connect to origin with, same as on #11545 but on a different run.

JosiahWI:
The other AuTest failures appear to be related to my change.

src/iocore/dns/DNSConnection.cc
```diff
@@ -209,7 +190,7 @@ DNSConnection::connect(sockaddr const *addr, Options const &opt)
   return 0;
 
 Lerror:
-  if (fd != NO_FD) {
+  if (this->sock.ok()) {
```
JosiahWI, line 193: This is wrong, it's missing a negation.

JosiahWI: Nope, I misread, this was done correctly the first try.

bneradt:
[approve ci autest rocky]

include/iocore/eventsystem/UnixSocket.h
```diff
@@ -49,24 +49,27 @@
...
+   * Call ok() to determine whether this call succeeded. If the call
    * failed, errno will be set to indicate the error.
    *
-   * @see has_socket
+   * @see ok
    */
   UnixSocket(int domain, int ctype, int protocol);
 
-  bool has_socket() const;
+  bool ok() const;
```
brbzull0, line 66: I would probably prepend it with `is_`  => `is_ok()`

JosiahWI:
Rebased.

include/iocore/eventsystem/UnixSocket.h
```diff
@@ -49,24 +49,27 @@
 
 bool transient_error();
 
-class UnixSocket
-{
-public:
+struct UnixSocket {
+  int fd{NO_SOCK};
```
moonchen, line 53: I think this member should stay private.  If it needs to be readable, you can use an accessor.  I don't think we want anyone outside this class to write this value, right?

include/iocore/eventsystem/UnixSocket.h
```diff
@@ -92,34 +95,33 @@ class UnixSocket
...
-  int sock_fd{NO_SOCK};
 };
 
-inline UnixSocket::UnixSocket(int fd) : sock_fd{fd} {}
+inline UnixSocket::UnixSocket(int fd) : fd{fd} {}
 
 inline UnixSocket::UnixSocket(int domain, int type, int protocol)
 {
-  this->sock_fd = socket(domain, type, protocol);
+  this->fd = socket(domain, type, protocol);
```
moonchen, line 110: I think we should remove all the instances of `this->` unless a local variable is shadowing a class member.

moonchen (changes requested):
Thank you for cleaning this up.  Besides the comments, I have some questions for this class:

1. Should this class have ownership of the fd?  If so
  1. I think it should be RAII,  e.g. implement a destructor that closes the socket.
  1. I think it should not be copy-constructible.

I'm okay with leaving the ownership question out of this PR, since you're not functionally changing the class here.

include/iocore/eventsystem/UnixSocket.h
JosiahWI: There aren't supposed to be any invariants here... there actually is one that I didn't think of, because `is_ok()` doesn't check for all invalid values. Having to go through a getter is just an inconvenience if there is no gain from encapsulation.

JosiahWI:
Are you suggesting RAII for `UnixSocket`?

include/iocore/eventsystem/UnixSocket.h
JosiahWI: We've decided to leave this because leave the file is new and doesn't have a different existing practice.

include/iocore/eventsystem/UnixSocket.h
JosiahWI: Keeping this private will make it easier to switch to RAII in the future. Done.

moonchen (approved):

Issue #11553, opened 2024-07-16
Errata severity string is not being render properly.


brbzull0:
It's using the raw value instead, this means that the library is not detecting the settings we [have](https://github.com/apache/trafficserver/blob/master/src/tsutil/ts_diags.cc#L28) for it.

```
[Jul 16 15:07:25.668] traffic_server NOTE: records parsing completed.
[Jul 16 15:07:25.668] traffic_server WARNING: We have found the following issues when reading the records node:
 5:   'records' root key not present or no fields to read.
  'records' root key not present or no fields to read.
  'records' root key not present or no fields to read.
  'records' root key not present or no fields to read.
  4: Ignoring field 'invalid_record' at Line 114. Not registered and Unknown tag type '?'
```
`4` should be `Warn` and `5` should be `Error`

PR #11554, opened 2024-07-16
Errata: Make sure the initialization of the errata members are in place.

brbzull0:
It seems the compiler/linker can omit a unit(in a static lib) if none of the functions in they are actually called. 
This commit adds a function call to initialize the `swoc::Errata` settings.
Without this the Errata is not properly initialized and runs with the defaults values.

with the change:
```
[Jul 16 16:06:30.658] traffic_server WARNING: We have found the following issues when reading the records node:
 Error:   'records' root key not present or no fields to read.
  'records' root key not present or no fields to read.
  'records' root key not present or no fields to read.
  'records' root key not present or no fields to read.
  Warn: Ignoring field 'invalid_record' at Line 114. Not registered and Unknown tag type '?'
 

```
Check the issue and the details before this patch:

fixes: https://github.com/apache/trafficserver/issues/11553

### DIFF
--- a/src/tsutil/ts_diags.cc
+++ b/src/tsutil/ts_diags.cc
@@ -25,9 +25,10 @@
 #include "tsutil/ts_diag_levels.h"
 #include "tsutil/ts_errata.h"
 
-static const bool INITIALIZED = []() -> bool {
+void
+Initialize_Errata_Settings()
+{
   swoc::Errata::DEFAULT_SEVERITY = ERRATA_ERROR;
   swoc::Errata::FAILURE_SEVERITY = ERRATA_WARN;
   swoc::Errata::SEVERITY_NAMES   = swoc::MemSpan<swoc::TextView const>(Severity_Names.data(), Severity_Names.size());
-  return true;
-}();
+}
--- a/src/traffic_server/traffic_server.cc
+++ b/src/traffic_server/traffic_server.cc
@@ -694,6 +694,8 @@ initialize_process_manager()
 
 extern void initializeRegistry();
 
+extern void Initialize_Errata_Settings();
+
 static void
 initialize_file_manager()
 {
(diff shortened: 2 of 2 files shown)


src/tsutil/ts_diags.cc
```diff
@@ -25,7 +25,13 @@
 #include "tsutil/ts_diag_levels.h"
 #include "tsutil/ts_errata.h"
 
-static const bool INITIALIZED = []() -> bool {
+// It seems the linker will omit this unit and the INITIALIZED lambda if it detects
+// that no other function is being called from this unit. So the following function
+// it's just a ugly and dirty way to force the linker and compiler to include this
+// and actually invoke the lambda down below.
+void Do_Not_Omit_Errata_Initialization() {};
```
ywkaras, line 32: I think it would a avoid unnecessary complexity to just have a function to do initialization for Errata, rather than doing the initialization in the dummy INITIALIZED variable.  I don't think we can be confident that the initialization of SEVERITY_NAMES within libswoc is done with constant initialization.  So, it could happen *after* the initialization of INITIALIZED.

brbzull0: Exactly, this should happen early on, and the best way is using the `INITIALIZED`, I did try calling a function to do this, but I did it probably too late in the flow. I can try just right after we start the `main` in `traffic_server.cc`

ywkaras: But, if the three variables are themselves used during non-local initialization, there is no guarantee (due to Initialization Fiasco) that they will be initialized before they are used.  Or, as I said, that the dynamic initialization of SEVERITY_NAMES will happen before INITIALIZED is initialized.

brbzull0:
Update:
Do the actual initialization when calling the init function.

ywkaras (approved):

ywkaras:
Oops need a non-Yahoo approve of this.

brbzull0:
> Oops need a non-Yahoo approve of this.

Too late :cry: . I think this is a safe change though.

ywkaras:
I will testify against you in court in exchange for immunity.

brbzull0:
I always knew you'll flip :dolphin: with no doubt!!

cmcfarlen:
Shame!

Cherry-picked to v10.0.x

PR #11555, opened 2024-07-16
ACL filter behavior configuration: deprecate legacy mode

bneradt:
This renames the ACL filter behavior as "legacy" and "modern" and
documents that the legacy behavior, while the default for 10.x, is
deprecated and will be removed in 11.x

---

# For Review

Note that this is a purely "superficial" naming and documentation PR. No behavior changes with this PR.


bneradt:
The addition of `add_allow` and `add_deny` actions are handled as a separate PR here:
https://github.com/apache/trafficserver/pull/11568

That seems non-controversial. Once that lands, I'll rebase this policy rename and deprecation messaging on top of that.

bneradt:
[approve ci]

bneradt:
[approve ci autest]

doc/admin-guide/files/records.yaml.en.rst
```diff
@@ -3559,17 +3559,25 @@ URL Remap Rules
...
-   ===== =============================
+   ===== =================================
    Value Description
-   ===== =============================
-   ``0`` Match on IP and Method Policy
-   ``1`` Match on IP only Policy
-   ===== =============================
+   ===== =================================
+   ``0`` Legacy (9.x and below) behavior.
+   ``1`` Modern (10.x and above) behavior.
```
masaori335, line 3571: I still don't have motivation to rename them to `Legacy` and `Modern`. I don't see why we can't keep it as is which describes the behavior well.

bneradt: Thank you for sharing this. It's good to discuss this.

Here's my logic: in changing the names I'm aiming for us to communicate to the end user what I think is the most important aspect of this configuration: which is the "old" compatibility mode and which is the "new" incompatibility mode. In other words, for most users what will be the most crucial aspect is "Which configuration will allow me to use my current remap.config without breaking things?" I chose the names "legacy" and "modern" because those words seem more noble than "old" versus "new". :) But I'm open to other suggestions.

My hesitation with "Match on IP and Method" and "Match on IP Only" is that, while they are literally accurate, aside from you, Matty, and me who have worked very closely with this system recently, I'm guessing those names will not likely be readily obvious in their meaning to most users. If we presented those names to a random ops person at Apple, do you think they'll know which one is the 9.x behavior confidently? They might be able to reason through it, but it might add a bit of a cognitive load. By the way, I'm not blaming you for these names, I did after all collaborate with you on them.

Having said that, I might be wrong. This really isn't a big deal to me. I'm glad you added the configuration in the first place, that will really help this release. Maybe the original names will help users consider the actual behaviors - that's their purpose. Either way we'll clearly document which is the compatibility mode and which is the new, which is default, and what our 11.x plans are.

So I'll leave it up to you. If after reading the above, you still want the original names, I'll happily leave it. Or if you have other naming suggestions, I'll consider that too.

Thank you for the review.

masaori335: Thanks for explanation. I got your point.

One thing I'd like to clarify, sorry for repeating what I pointed out on ML, is that "Match on IP and Method Policy" is pretty similar to the behavior of 9.2.x and older, so almost all of the old configs will work as is. However, it employs "first match win" policy between ACL filters in remap.config and ip_allow.yaml. Then, strictly speaking, we can't say it's 100% compatible with old configs. I'd still say calling it "Legacy" is a kind of mis-leading. I wrote this in the "Note" of the Match on IP and Method Policy docs. I welcome to make it more clear, but I prefer to keep calling it "Match on IP and Method Policy".

https://docs.trafficserver.apache.org/en/10.0.x/admin-guide/files/remap.config.en.html#match-on-ip-and-method-policy

doc/admin-guide/files/remap.config.en.rst
```diff
@@ -675,18 +764,18 @@ PURGE  \-             \-             \-             deny             denied  (40
...
-With the "Match on IP only Policy", the in-line filter works like an :file:`ip_allow.yaml` rule applies to all requests to
-``www.example.com`` that denies ``POST`` requests and implicitly allows all other methods:
+With the "Modern" policy, the in-line filter works like an :file:`ip_allow.yaml` rulei: it applies to all requests to
+``www.example.com`` and denies ``POST`` requests while allowing all other methods:
 
 ====== ================ ============== ============== ============= =============
 Method In-line Filter   Named Filter 1 Named Filter 2 ip_allow.yaml result
 ====== ================ ============== ============== ============= =============
-GET    allow (implicit) \-             \-             \-            allowed (200)
+GET    allow            \-             \-             would allow   allowed (200)
```
masaori335, line 773: What do you mean by `would allow` and `would deny` in this table?

doc/admin-guide/files/records.yaml.en.rst
bneradt: Sounds good. I'll revert to current names then tomorrow.

doc/admin-guide/files/remap.config.en.rst
```diff
@@ -675,18 +764,18 @@ PURGE  \-             \-             \-             deny             denied  (40
 PUSH   \-             \-             \-             deny             denied  (403)
 ====== ============== ============== ============== ================ =============
 
-With the "Match on IP only Policy", the in-line filter works like an :file:`ip_allow.yaml` rule applies to all requests to
-``www.example.com`` that denies ``POST`` requests and implicitly allows all other methods:
+With the "Modern" policy, the in-line filter works like an :file:`ip_allow.yaml` rulei: it applies to all requests to
```
bneradt, line 767: `rulei` is not a word

doc/admin-guide/files/remap.config.en.rst
bneradt: With the Match on IP Only policy, only the inline ACL rule takes effect. None of the named ACL filters nor the ip_allow rule is queried. So I thought it helpful to make it clear in the table that the other rules are not actually denying or allowing something, rather they "would allow" if they were in effect. Note before, we had some `allow` and `deny` entries while those filters were not in effect. So I wanted to indicate that _had_ they run they would make that impact, thus "would deny" or "would allow". 

I suppose we should do one of two things:

* Change all of the other entries to `\-` other than the in-line filter column and the result column.
* Fill in all the other entries with "would allow/deny" because, if they were in effect, with Match on IP Only policy, they would completely specify the behavior.

I guess my vote would be to change all to `\-`.

doc/admin-guide/files/records.yaml.en.rst
masaori335: Thank you!

bneradt: I'll replace this PR with one that does not do the name changes:
https://github.com/apache/trafficserver/pull/11633

bneradt:
Closing since this will be replaced with the non-name change version:
https://github.com/apache/trafficserver/pull/11633

bneradt:
> Closing since this will be replaced with the non-name change version: #11633

After working on the set_allow and set_deny documentation, I've come to the realization that we really should stick with this rename opening this again. Opening as a separate PR though after a rebase and branch rename:
https://github.com/apache/trafficserver/pull/11635

PR #11557, opened 2024-07-16
Fix large post bodies in transform plugins

### LINKED ISSUE #11556: Large POST Bodies Cause Failure with Transform Plugin
When we run traffic with small post bodies (e.g., 1024 bytes), the traffic passes through our transform plugin without issue. 

When the post body is larger (e.g. 4096), the curl request fails and this message shows up in error.log "Request transformation failed to set content length".

The key difference between the working and failing case is that in the failing case, the post body is being read over multiple packets. In the working case, the post body and the request header are read together.

We first saw this in 9.1.3, but it reproduces in 10.  I have a code change to make this work for me. I will post that, but I may be missing how to set the data argument for the transform.

shinrich:
As discussed in the issue, this PR fixes the POST stall with our plugin.

Should I be setting the output length of the post body from the plugin? If so how? 

Should we pull the content length out of the client request header instead of leaving it unset?

If this seems like a reasonable approach, I'll add an autest case.

This closes #11556

### DIFF
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -1131,17 +1131,13 @@ HttpSM::state_request_wait_for_transform_read(int event, void *data)
       //   data as well as fix up the request header
       t_state.hdr_info.transform_request_cl = size;
       t_state.hdr_info.server_request.value_set_int64(MIME_FIELD_CONTENT_LENGTH, MIME_LEN_CONTENT_LENGTH, size);
-      setup_server_send_request_api();
-      break;
-    } else {
-      // No content length from the post.  This is a no go
-      //  since http spec requires content length when
-      //  sending a request message body.  Change the event
-      //  to an error and fall through
-      event = VC_EVENT_ERROR;
-      Log::error("Request transformation failed to set content length");
     }
-  // FALLTHROUGH
+    // If the post body is more than one packet, we may not have the total body size set.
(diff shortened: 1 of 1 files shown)


src/proxy/http/HttpSM.cc
```diff
@@ -1131,17 +1131,13 @@ HttpSM::state_request_wait_for_transform_read(int event, void *data)
...
-      //  to an error and fall through
-      event = VC_EVENT_ERROR;
-      Log::error("Request transformation failed to set content length");
     }
-  // FALLTHROUGH
+    // If the post body is more than one packet, we may not have the total body size set.
+    // Continue on with the transform_request_cl set to the default value of -1. If the transform
+    // changes the content lenght of the post body, the data argument must be specified even after
+    // the first packet
+    setup_server_send_request_api();
```
vmamidi, line 1139: I think we shouldn't do POST transformation if we do not have TE/CL headers for POST request.

shinrich: The header does have TE/CL headers in the case where I ran into this. But this size data was not being pass through the data argument to the call back. 

I guess this is something that the transform needs to set? But I couldn't find examples of transforms setting this value, or documentation describing how to set this value. I think this needs to be passed through if the transform changes the size of the post body.  Otherwise, the client side values will pass over. 

I could change this logic to validate that the TE/CL headers have been set on the client side even if a new size is not set through the callback data argument.

PR #11562, opened 2024-07-17
Rename `Stripe` to `StripeSM`

JosiahWI:
The real goal here is to split the continuation logic out of `Stripe` into `StripeSM` to help with the re-partitioning of methods into smaller .cc files. I'm going to do this 'backwards', by changing the existing class to `StripeSM` and extracting logic back out into `Stripe`. This rename is quite large, so I'm doing it as its own PR first to make sure I don't miss anything. I will restore the `Stripe` class again, but many of the references in the code will still need to point to `StripeSM`.


masaori335 (approved):
`Stripe` is pretty big class and it has several aspect. Decoupling event handlers to `StripeSM` seems like good idea.

Anyway,  this PR seems just renaming and harmless.

PR #11563, opened 2024-07-17
Fix server names for post-early-return

bneradt:
The post-early-return.test.py server* names are all copy and pasted as "server1" which means they are not configured correctly. This patch updates those names so they are distinct.


JosiahWI:
The AuTest chunked_encoding_disabled failed.

JosiahWI:
[approve ci autest]

cmcfarlen:
Cherry-picked to v10.0.x

PR #11565, opened 2024-07-18
Extract Stripe superclass from StripeSM

JosiahWI:
The `StripeSM` class retains all the evacuation data and the mutex.

### DIFF
--- a/src/iocore/cache/P_CacheDir.h
+++ b/src/iocore/cache/P_CacheDir.h
@@ -33,6 +33,7 @@
 // aio
 #include "iocore/aio/AIO.h"
 
+class Stripe;
 class StripeSM;
 struct InterimCacheVol;
 struct CacheVC;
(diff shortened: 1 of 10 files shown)


src/iocore/cache/P_CacheDir.h
```diff
@@ -266,29 +267,28 @@ struct CacheSync : public Continuation {
...
+int      dir_probe(const CacheKey *, StripeSM *, Dir *, Dir **);
+int      dir_insert(const CacheKey *key, StripeSM *stripe, Dir *to_part);
+int      dir_overwrite(const CacheKey *key, StripeSM *stripe, Dir *to_part, Dir *overwrite, bool must_overwrite = true);
+int      dir_delete(const CacheKey *key, StripeSM *stripe, Dir *del);
+int      dir_lookaside_probe(const CacheKey *key, StripeSM *stripe, Dir *result, EvacuationBlock **eblock);
+int      dir_lookaside_insert(EvacuationBlock *b, StripeSM *stripe, Dir *to);
+int      dir_lookaside_fixup(const CacheKey *key, StripeSM *stripe);
+void     dir_lookaside_cleanup(StripeSM *stripe);
+void     dir_lookaside_remove(const CacheKey *key, StripeSM *stripe);
+void     dir_free_entry(Dir *e, int s, Stripe *stripe);
```
cmcfarlen, line 279: It looks odd that some of these take `StripeSM*` and some take `Stripe*`. It looks like it's because the mutex is checked which is part of the `Continuation` base class.  Does it seem worth changing the signatures on these functions? Could these be part of the Stripe interface rather than bare functions?

cmcfarlen (commented):
What is the incentive to do this? Are there plans to use the split out Stripe class on its own in some way? It seems like without the lock from Continuation, this can't be used on its own.

Maybe consider making Stripe or some superclass of Stripe an interface to clean that up and have StripeSM implement the aspects that integrate with Continuations and locking.

src/iocore/cache/P_CacheDir.h
JosiahWI: I'd like to keep these separate from the `Stripe` interface; as I understand it they are essentially methods on an imaginary cache directory object that we don't yet have a class to represent.

JosiahWI:
This is an code organization change. The long term incentive is separation of concerns, and the short term goal is to split up massive source files into smaller, separate source files. In this case, we're separating the data structure from the event handling. The separation is still subpar because the event handling methods touch the data structure too much, but it's a start. This will be followed by a PR to move all the `StripeSM` event handling into a new `StripeSM.cc` source file.

src/iocore/cache/P_CacheDir.h
JosiahWI: Maybe we could remove some of the assertions, but it will need careful examination of the code and I think it should be done in a separate PR for safety.

cmcfarlen (approved):

PR #11568, opened 2024-07-18
add_allow and add_deny ACL filter actions

bneradt:
This adds the add_allow and add_deny filter actions. These actions behave just like the allow and deny actions from pre-10 ACL filters, but makes the difference of their behavior from ip_allow.yaml allow and deny actions explicit. The allow and deny ACL filter actions are also kept, but their behavior depends upon which
proxy.config.url_remap.acl_matching_policy is set:

* If the policy is 0 (the default), then the allow and deny ACL filter actions act as synonyms for add_allow and add_deny since their behavior is compatible with the pre-10 release.
* If the policy is 1 (the new 10 behavior), then the allow and deny ACL filter actions behave like their ip_allow.yaml counterparts.

These new actions supplement the additive nature of ACL filters to the proxy.config.url_remap.acl_matching_policy 1 configured value, while also allowing users to explicitly specify this additive nature when the ACL matching policy is set to 0.

---

Note that this PR is a subset of the #11555 in that it simply adds the `add_allow` and `add_deny` actions while not renaming the `proxy.config.url_remap.acl_matching_policy` policies nor using any deprecation language.

### DIFF
--- a/doc/admin-guide/files/remap.config.en.rst
+++ b/doc/admin-guide/files/remap.config.en.rst
@@ -447,6 +447,29 @@ In-line filters can be created to control access of specific remap lines. The ma
 is very similar to that of :file:`ip_allow.yaml`, with slight changes to
 accommodate remap markup.
 
+Actions
+~~~~~~~
+
+As is the case with :file:`ip_allow.yaml` rules, each ACL filter takes one of a number of actions. They are specified as
+``@action=<action>``, such as ``@action=add_allow``. There are four possible actions:
+
+- ``allow``: This behaves like the ``allow`` action in :file:`ip_allow.yaml` in which a list of allowed methods are
+  provided. Any request with a method in the list is allowed, while any request with a method not in the list is denied.
+  The exception to this is if :ts:cv:`proxy.config.url_remap.acl_matching_policy` is set to ``0``. In this case, the
+  ``allow`` action is a synonym for ``add_allow``, described below.
+- ``add_allow``: This action adds a list of allowed methods to whatever other methods are allowed in a subsequently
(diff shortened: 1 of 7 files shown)


doc/admin-guide/files/remap.config.en.rst
```diff
@@ -447,6 +447,29 @@ In-line filters can be created to control access of specific remap lines. The ma
...
+
+As is the case with :file:`ip_allow.yaml` rules, each ACL filter takes one of a number of actions. They are specified as
+``@action=<action>``, such as ``@action=add_allow``. There are four possible actions:
+
+- ``allow``: This behaves like the ``allow`` action in :file:`ip_allow.yaml` in which a list of allowed methods are
+  provided. Any request with a method in the list is allowed, while any request with a method not in the list is denied.
+  The exception to this is if :ts:cv:`proxy.config.url_remap.acl_matching_policy` is set to ``0``. In this case, the
+  ``allow`` action is a synonym for ``add_allow``, described below.
+- ``add_allow``: This action adds a list of allowed methods to whatever other methods are allowed in a subsequently
+  matched ACL filter or :file:`ip_allow.yaml` rule. Thus, if an ``add_allow`` ACL filter specifies the ``POST`` method,
```
masaori335, line 461: aside naming of the action, I suggest below to contrast the "first match wind policy".

```
This action adds a list of allowed methods to whatever specified methods are denied in a subsequently
matched ACL filter or :file:`ip_allow.yaml` rule.
```

masaori335 (approved):
Implementation wise, this looks good. The name of actions, if nobody get brilliant idea, we can go with this. Let's move forward.

cmcfarlen:
Cherry-picked to v10.0.x

Issue #11552, opened 2024-07-16
Client connection may time out in proxy_protocol AuTest


JosiahWI:
## Default Process
```
[DEBUG]: Configured logging at level DEBUG
[INFO]: Loading replay data from "/tmp/sandbox/proxy_protocol/pp-out-client-4/proxy_protocol_out.replay.yaml".
    [INFO]: Replay file parsing took: 0 milliseconds.
[INFO]: Parsed 1 transaction in 1 session.
[DEBUG]: Poll timeout set to 5000ms.
[DEBUG]: Initialize TLS
    [DEBUG]: Writing TLS secrets to: /tmp/sandbox/proxy_protocol/pp-out-client-4/tls_secrets.txt
  [DEBUG]: Finished TLSSession::init
[DEBUG]: Initialize H2
  [DEBUG]: Finished H2Session::init
[DEBUG]: Initialize H3
    [DEBUG]: QUIC log files will be written to /tmp/sandbox/proxy_protocol/pp-out-client-4/qlog_directory
  [DEBUG]: Finished H3Session::init
[DEBUG]: Client initialization complete.
[DEBUG]: Starting session "/tmp/sandbox/proxy_protocol/pp-out-client-4/proxy_protocol_out.replay.yaml":24 protocol=https.
[DEBUG]: Connecting via TLS.
    [DEBUG]: Proxy TLS verification result: 18 (not X509_V_OK).
[ERROR]: SSL_read timed out waiting to TLS content after 5000 milliseconds milliseconds.
        [DEBUG]: Sent the following HTTP/1 request headers for key 1:
GET /get/request HTTP/1.1
uuid: 1
host: pp.origin.com

      [DEBUG]: Reading response header.
      [ERROR]: Failed to find a well-formed, completed HTTP response: PARSE_INCOMPLETE
    [ERROR]: Failed HTTP/1 transaction with key: 1
[INFO]: 1 transaction in 1 session (reuse 1) in 5018 milliseconds (0.0 / millisecond).
```

## ATS
```
+++++++++ Proxy's Request +++++++++
-- State Machine Id: 0
GET /get/request HTTP/1.1
uuid: 1
host: backend.pp.origin.com:61990
Client-ip: 127.0.0.1
X-Forwarded-For: 127.0.0.1
Via: https/1.1 traffic_server[1c94b2e8-4158-4da2-9cf5-cf165e58480b] (ApacheTrafficServer/10.1.0)

[...]

PR #11574, opened 2024-07-18
Dispatch WRITE_COMPLETE when TLS handshake completes

bneradt:
The SSLNetVConnection.cc logic already does this, so this is a port over
of that event dispatch functionality when ntodo is 0 for
UnixNetVConnection.cc. @josiahwi and I investigated a periodic failure
in proxy_protocol.test.py and we noticed that the issue was due to the
handshake completing within a single call of sslStartHandshake. In those
cases, the ConnectingEntry handler wasn't called back and the connection
would hang without progressing with the rest of the request after the
handshake completed. This ensures that the ConnectingEntry handler is
called in these circumstances.

---

# Review Comments

## SSLNetVConnection Reference
For reference, here is the corresponding dispatch from SSLNetVConnection that does this callback as well:

https://github.com/apache/trafficserver/blob/5f29d55f3cff865f2178d597199d72eb747ac80b/src/iocore/net/SSLNetVConnection.cc#L690-L694

This patch simply pulls in that callback for UnixNetVConnection's TLS handshake handling logic.

## Testing

The proxy_protocol.test.py test demonstrates this infrequently in ASF CI. I have never seen this locally on my mac. I've run this test in CI using my access to the CI vms dozens of times and have not seen a failure. I have been able to reproduce the error fairly consistently before this patch, so I feel this addresses the issue well.

### DIFF
--- a/src/iocore/net/UnixNetVConnection.cc
+++ b/src/iocore/net/UnixNetVConnection.cc
@@ -418,6 +418,12 @@ write_to_net_io(NetHandler *nh, UnixNetVConnection *vc, EThread *thread)
       if (vc->write.enabled) {
         nh->write_ready_list.in_or_enqueue(vc);
       }
+      // If this was driven by a zero length read, signal complete when
+      // the handshake is complete. Otherwise set up for continuing read
+      // operations.
+      if (s->vio.ntodo() <= 0) {
+        vc->readSignalDone(VC_EVENT_WRITE_COMPLETE, nh);
+      }
     } else {
       write_reschedule(nh, vc);
     }


JosiahWI:
Fixes #11552.

src/iocore/net/UnixNetVConnection.cc
```diff
@@ -418,6 +418,12 @@ write_to_net_io(NetHandler *nh, UnixNetVConnection *vc, EThread *thread)
       if (vc->write.enabled) {
         nh->write_ready_list.in_or_enqueue(vc);
       }
+      // If this was driven by a zero length read, signal complete when
+      // the handshake is complete. Otherwise set up for continuing read
+      // operations.
+      if (s->vio.ntodo() <= 0) {
+        vc->readSignalDone(VC_EVENT_READ_COMPLETE, nh);
```
maskit, line 425: `VC_EVENT_READ_COMPLETE`? Not `VC_EVENT_WRITE_COMPLETE`?

bneradt: Great catch! I copied from the read operation of the SSLNetVConnection, but this is for the `do_io_write`. Let me try out the patch with WRITE_COMPLETE and verify that addresses the failure as well. It almost certainly will since the ConnectingEntry handles both the read and write complete events in the same switch case block.

maskit (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11582, opened 2024-07-19
Use VIO reader and writer APIs

JosiahWI:
These APIs are already provided and make the code more succinct. This updates the code everywhere to consistently use these convenience APIs. The continuation API usage is not changed because `VIO::set_continuation` has some additional behavior, so the usage is not as straightforward.


bneradt (commented):
Thanks for the cleanup. It's pretty straightforward. I'll let a non-Yahoo reviewer look at this for official review/approval.

serrislew (approved):

PR #11583, opened 2024-07-19
remap_action_to_add: helper script for new ACL actions

bneradt:
This helper script converts 9.2.x @action=allow and @action=deny to
@action=add_allow and @action=add_deny in a provided input remap.config
filename. This prepares a 9.2.x remap.config file to be compatible with
the new 10.x ACL action behavior.

### DIFF
--- a/tools/remap/convert_remap_actions_to_10x
+++ b/tools/remap/convert_remap_actions_to_10x
@@ -1,0 +1,172 @@
+#!/usr/bin/env python3
+'''Convert allow and deny @actions to add_allow and add_deny.'''
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import argparse
+import os
+import shutil
+import sys
+import tempfile
+from typing import Optional, TextIO, Tuple
+
+help = '''Convert allow and deny @actions to add_allow and add_deny.
+
+This script is used to convert a pre-10.x remap.config file to be functionally
+equivalent to a 10.x remap.config file. In 10.x, the pre-10.x @action=allow and
+@action=deny actions are renamed to @action=add_allow and @action=add_deny,
+respectively. This script will convert the former to the latter. For further
+details, see the remap.config documentation.
+'''
+
+
+def convert_line(input_line: str) -> str:
+    '''Convert an input line to an output line.
+
+    This function modifies the input line for that of the new 9.2.x format.
+
+    :param input_line: The line to convert.
+    :return: The converted line.
+
+    :examples:
+        >>> convert_line('# nothing to convert')
+        '# nothing to convert'
+        >>> convert_line('map http://example.com http://backend.example.com')
+        'map http://example.com http://backend.example.com'
+
+        >>> convert_line('map http://example.com http://backend.example.com @action=allow @method=GET')
+        'map http://example.com http://backend.example.com @action=add_allow @method=GET'
+        >>> convert_line('map http://example.com http://backend.example.com @action=deny @method=PUT')
+        'map http://example.com http://backend.example.com @action=add_deny @method=PUT'
+
+        # Verify that add_allow and add_deny are not converted.
+        >>> convert_line('map http://example.com http://backend.example.com @action=add_deny @method=PUT')
+        'map http://example.com http://backend.example.com @action=add_deny @method=PUT'
+        >>> convert_line('map http://example.com http://backend.example.com @action=add_allow @method=GET')
+        'map http://example.com http://backend.example.com @action=add_allow @method=GET'
+
+        # Comments should be converted too.
+        >>> convert_line('# Using @action=allow is nice.')
+        '# Using @action=add_allow is nice.'
+        >>> convert_line('# map http://example.com http://backend.example.com @action=allow @method=GET')
+        '# map http://example.com http://backend.example.com @action=add_allow @method=GET'
+    '''
+    output_line = input_line.replace('@action=allow', '@action=add_allow')
+    output_line = output_line.replace('@action=deny', '@action=add_deny')
+    return output_line
+
+
+def copy_input_file(input: str, no_backup: bool) -> str:
+    '''Create a copy of the input file.
+
+    :param input: The input file path from which to create a copy.
+
+    :param no_backup: Do not create a backup file from input. Rather simply
+    create a temporary file that will be later removed. Otherwise, create a
+    backup file of the format <input>.action_to_add.bak.<num>, where <num> is
+    one greater than the greatest value in the directory where @a input exists.
+
+    :return: The path to the copied file.
+    '''
+    if no_backup:
+        copy_path = tempfile.NamedTemporaryFile(delete=False).name
+    else:
+        input_dir = os.path.dirname(input)
+        input_dir = '.' if input_dir == '' else input_dir
+        input_base = os.path.basename(input)
+        # Find the largest <num> in input_dir/input_base.action_to_add.bak.<num>.
+        max_num = 0
+        for file in os.listdir(input_dir):
+            if file.startswith(f'{input_base}.action_to_add.bak.'):
+                try:
+                    num = int(file.split('.')[-1])
+                except ValueError:
+                    # The suffix is not a number. Ignore this file.
+                    continue
+                max_num = max(max_num, num)
+        copy_path = os.path.join(input_dir, f'{input_base}.action_to_add.bak.{max_num + 1}')
+    shutil.copyfile(input, copy_path)
+    return copy_path
+
+
+def prepare_files(input: str, output: Optional[str], no_backup: bool) -> Tuple[str, str]:
+    '''Prepare the input and output files.
+
+    :param input: The input file path.
+    :param output: The output file path.
+    :param no_backup: Do not create a backup of the input file when <input> is modified.
+
+    :return: A tuple containing the input and output filenames, in that order.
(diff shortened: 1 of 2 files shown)


tools/remap/convert_remap_actions_to_10x
```diff
@@ -0,0 +1,172 @@
...
+    if input == output:
+        input_path = copy_input_file(input, no_backup)
+        output_path = input
+    elif output is None:
+        input_path = copy_input_file(input, no_backup)
+        output_path = input
+    else:
+        input_path = input
+        output_path = output
+    return input_path, output_path
```
JosiahWI, line 132: Good use of the syntax to return multiple values as a tuple.

tools/remap/convert_remap_actions_to_10x
```diff
@@ -0,0 +1,172 @@
...
+    :return: A tuple containing the input and output filenames, in that order.
+    The input file will contain the content from which to copy, the output file
+    will reference a file whose content should be overridden. The caller should
+    delete the input filename if @a no_backup is true. Otherwise the input
+    filename will be the path to the backup file.
+    '''
+    if not os.path.exists(input):
+        raise FileNotFoundError(f'Input file does not exist: {input}')
+
+    if input == output:
```
JosiahWI, line 123: Uses a temporary file to make the changes before replacing the output. Very good. Comment might not hurt but not necessary either.

tools/remap/convert_remap_actions_to_10x
```diff
@@ -0,0 +1,172 @@
...
+        input_base = os.path.basename(input)
+        # Find the largest <num> in input_dir/input_base.action_to_add.bak.<num>.
+        max_num = 0
+        for file in os.listdir(input_dir):
+            if file.startswith(f'{input_base}.action_to_add.bak.'):
+                try:
+                    num = int(file.split('.')[-1])
+                except ValueError:
+                    # The suffix is not a number. Ignore this file.
+                    continue
```
JosiahWI, line 100: Testing `if num.isdigit()` would be shorter and more explicit. No need for try/except here.

tools/remap/convert_remap_actions_to_10x
```diff
@@ -0,0 +1,172 @@
...
+    elif output is None:
+        input_path = copy_input_file(input, no_backup)
+        output_path = input
+    else:
+        input_path = input
+        output_path = output
+    return input_path, output_path
+
+
+def parse_args() -> argparse.Namespace:
```
JosiahWI, line 135: Good use of argparse to support command line arguments.

tools/remap/convert_remap_actions_to_10x
```diff
@@ -0,0 +1,172 @@
...
+    :return: The process exit code.
+    '''
+
+    args = parse_args()
+    input_path, output_path = prepare_files(args.input, args.output, args.no_backup)
+
+    with open(input_path, 'r') as fd_in, open(output_path, 'w') as fd_out:
+        for input_line in fd_in:
+            output_line = convert_line(input_line)
+            fd_out.write(output_line)
```
JosiahWI, line 163: This is fine, but it would be faster to read everything and then write everything, rather than reading and writing one line at a time.

tools/remap/convert_remap_actions_to_10x
```diff
@@ -0,0 +1,172 @@
...
+def main() -> int:
+    '''Convert the input remap.config file.
+
+    :return: The process exit code.
+    '''
+
+    args = parse_args()
+    input_path, output_path = prepare_files(args.input, args.output, args.no_backup)
+
+    with open(input_path, 'r') as fd_in, open(output_path, 'w') as fd_out:
```
JosiahWI, line 160: The 'r' mode is the default, it can be omitted, but if you want to leave it in to be explicit I don't mind.

tools/remap/convert_remap_actions_to_10x
```diff
@@ -0,0 +1,172 @@
...
+        for input_line in fd_in:
+            output_line = convert_line(input_line)
+            fd_out.write(output_line)
+    if args.no_backup:
+        os.remove(input_path)
```
JosiahWI, line 165: The backup won't get removed if an exception occurs after creating the file but before getting here, and that would be annoying. 😠

tools/remap/convert_remap_actions_to_10x
```diff
@@ -0,0 +1,172 @@
...
+    return 0
```
JosiahWI, line 166: Good idea to return exit code for `sys.exit`.

tools/remap/convert_remap_actions_to_10x
```diff
@@ -0,0 +1,172 @@
...
+
+
+if __name__ == '__main__':
+    import doctest
+    doctest.testmod()
```
JosiahWI, line 171: A bit messy to always run doctest. The user has no way to disable this functionality. :/

JosiahWI (changes requested):
Thanks for the script. It passes `python3 -m mypy remap_action_to_add` static type linting, and has good docstrings and unit tests. I have some other thoughts for consideration. Also tested with a local file to make sure the replacement works as advertised.

tools/remap/convert_remap_actions_to_10x
```diff
@@ -0,0 +1,172 @@
...
+        '# Using @action=add_allow is nice.'
+        >>> convert_line('# map http://example.com http://backend.example.com @action=allow @method=GET')
+        '# map http://example.com http://backend.example.com @action=add_allow @method=GET'
+    '''
+    output_line = input_line.replace('@action=allow', '@action=add_allow')
+    output_line = output_line.replace('@action=deny', '@action=add_deny')
+    return output_line
+
+
+def copy_input_file(input: str, no_backup: bool) -> str:
```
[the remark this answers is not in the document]
bneradt: I see, that's cool. Is that done with the special `/` parameter?

tools/remap/convert_remap_actions_to_10x
```diff
@@ -0,0 +1,172 @@
...
+
+def convert_line(input_line: str) -> str:
+    '''Convert an input line to an output line.
+
+    This function modifies the input line for that of the new 9.2.x format.
+
+    :param input_line: The line to convert.
+    :return: The converted line.
+
+    :examples:
```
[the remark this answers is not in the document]
bneradt: I typically indented these for visual offset, but I can unindent since that makes doctest happier.

However, as I'm guessing you saw, the tests still don't run with that addressed. The doctests do run via the main, so it seems like some kind of bug in doctest in my opinion (at least a usability bug).

tools/remap/convert_remap_actions_to_10x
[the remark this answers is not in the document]
JosiahWI: Yes.

tools/remap/convert_remap_actions_to_10x
bneradt: The `copy_input_file` doc says it will create a temporary. Would you suggest another comment here?

tools/remap/convert_remap_actions_to_10x
bneradt: Cool.

tools/remap/convert_remap_actions_to_10x
bneradt: Isn't it better to read a line at a time to reduce resource usage? These files can be rather large. Yahoo's are 10's of thousands of entries.

tools/remap/convert_remap_actions_to_10x
bneradt: I didn't know 'r' is the default so thanks for the comment. But I would prefer to leave it to keep it explicit.

tools/remap/convert_remap_actions_to_10x
bneradt: I'd feel comfortable removing it if `python3 -m doctest convert_remap_actions_to_10x` worked. I've never run in this way and I seem to not be able to make it work:

```shell
╰─➤  cat d
#!/usr/bin/env python3

def add_one(i: int) -> int:
    '''Add one to the input.

    :param i: The input to increment.
    :return: One ore than @a i.
    :example:
    >>> add_one(2)
    3
    >>> add_one(0)
    1
    '''

    return i + 1

n = add_one(0)
print(n)
╭─[31/07/24 6:14:53] bneradt@lima-rancher-desktop  /tmp 
╰─➤  python3 -m doctest -v d
Trying:
    add_one(2)
Expecting:
    3
**********************************************************************
File "d", line 9, in d
Failed example:
    add_one(2)
Exception raised:
    Traceback (most recent call last):
      File "/usr/lib64/python3.12/doctest.py", line 1368, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest d[0]>", line 1, in <module>
        add_one(2)
        ^^^^^^^
    NameError: name 'add_one' is not defined
Trying:
    add_one(0)
Expecting:
    1
    '''
**********************************************************************
File "d", line 11, in d
Failed example:
    add_one(0)
Exception raised:
    Traceback (most recent call last):
      File "/usr/lib64/python3.12/doctest.py", line 1368, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest d[1]>", line 1, in <module>
        add_one(0)
        ^^^^^^^
    NameError: name 'add_one' is not defined
**********************************************************************
1 items had failures:
   2 of   2 in d
2 tests in 1 items.
0 passed and 2 failed.
***Test Failed*** 2 failures.
```

tools/remap/convert_remap_actions_to_10x
JosiahWI: Sure, if the resource usage is a concern then I understand trading off performance.

tools/remap/convert_remap_actions_to_10x
JosiahWI: Nah.

tools/remap/convert_remap_actions_to_10x
JosiahWI: What about `python3 -m doctest -v convert_remap_actions_to_10x`?

JosiahWI: Looks like that just adds a verbose flag.

JosiahWI: I'm fine with leaving it.

bneradt: Just to record it, you found that this is due to the script not having a .py extension. I'd prefer to leave the doctest in the code than to add a .py extension. Scripts shouldn't need to advertise their implementation.

bneradt:
Thanks for the review @JosiahWI

cmcfarlen (approved):
Just rubber stamping @JosiahWI Thanks!

cmcfarlen:
Cherry-picked to v10.0.x

PR #11590, opened 2024-07-22
Remove paramset in url_sig

ezelkow1:
This removes the ParamsSet call within url_sig which was just clearing out the ATS parsed params anyway

### DIFF
--- a/plugins/experimental/url_sig/url_sig.cc
+++ b/plugins/experimental/url_sig/url_sig.cc
@@ -890,7 +890,6 @@ allow:
     if (*new_path) {
       TSUrlPathSet(rri->requestBufp, rri->requestUrl, new_path, strlen(new_path));
     }
-    TSUrlHttpParamsSet(rri->requestBufp, rri->requestUrl, nullptr, 0);
   }
 
   TSfree((void *)current_url);


JosiahWI (approved):
Looks good.

plugins/experimental/url_sig/url_sig.cc
```diff
@@ -890,7 +890,6 @@ TSRemapDoRemap(void *ih, TSHttpTxn txnp, TSRemapRequestInfo *rri)
     if (*new_path) {
       TSUrlPathSet(rri->requestBufp, rri->requestUrl, new_path, strlen(new_path));
     }
-    TSUrlHttpParamsSet(rri->requestBufp, rri->requestUrl, nullptr, 0);
```
maskit, line 893: Can we really remove this line? The original behavior may mean to remove "matrix parameters", which is everything that follows a semicolon in a URL.

ezelkow1: Im trying to create an autest for this first before merging. I believe what it was doing was packing it all into 'new_path', setting the Url path, and then removing the matrix params afterwards.

So with the old implementation would doing a ParamsSet end up re-evaluating the path to locate the params again before removing them? I was figuring that what it was doing was setting the Path with any included new path params and then remove ATS' stored path params so that it would not end up being duplicated

maskit: I didn't follow the entire logic. It just doesn't look like equivalent even if we set nullptr. If `new_path` contains parameters it may be fine, but what if we don't set `new_path` here?

ezelkow1: alright let me think about it some more. Pinged some around here but no one really knows how this code was meant to work any more so its a bit of reverse engineering going on

PR #11599, opened 2024-07-23
Remove pathset for urlsig

ezelkow1:
Added 2 tests, one with a bad path param and one with a properly encoded path param. ATS returns 403 and 200 properly

Added path param checks for success and failure, shuffled tests for correct logging output

If there is a spurious path param in a request to urlsig it's going to 403 because it will not match the signature provided. So that should be why it is ok to remove the ParamsSet here since if there are no path params then the new path does not need to be set and params dont need to be updated. If there are path params, and they are good, then the new path will be set which includes the params. If there are path params but they are bad then it is going to 403 anyway


maskit (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11601, opened 2024-07-23
Add an autest for s-maxage Cache-Control response directive

bneradt:
This adds an autest to verify that ATS caching funcionality behaves correctly when handling the s-maxage Cache-Control response directive. No production changes are needed because ATS passes all the added tests.


cmcfarlen (approved):

cmcfarlen:
Cherry-picked to v10.0.x

ezelkow1:
do we still need this on 9.2?

bneradt:
> do we still need this on 9.2?

I think we might as well. The patch is test-only, so it's safe and the coverage wouldn't hurt.

PR #11602, opened 2024-07-23
Set -Werror=no-stringop-overflow for boringssl

### LINKED ISSUE #11592: Cannot build static boringssl using build_h3_tools with GCC13
The boringssl ticket for the issue is at https://issues.chromium.org/issues/42290598. It has been fixed in boringssl as of June 2024, ~~but ngtcp2 does not build with that recent version of boringssl because the symbol it uses to detect whether the provided library is boringssl is no longer there~~. We could work around it on our end by building with `-Wno-error=stringop-overflow`. There is no ngtcp2 ticket for building with a newer boringssl yet.

Edit: The ngtcp2 build issue may be unrelated to the version change. It reproduces with our current commit checkout of boringssl as well. Tested: ngtcp2 does fail with the new version, I can eventually look in to it.

JosiahWI:
Fixes #11592.

### DIFF
--- a/tools/build_boringssl_h3_tools.sh
+++ b/tools/build_boringssl_h3_tools.sh
@@ -157,8 +157,8 @@ cmake \
   -DGO_EXECUTABLE=${GO_BINARY_PATH} \
   -DCMAKE_INSTALL_PREFIX=${BASE}/boringssl \
   -DCMAKE_BUILD_TYPE=Release \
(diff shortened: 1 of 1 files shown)


brbzull0 (commented):
Nice thanks.

maskit (approved):
This is fine, but we don't have a reason to continue using the old commit.

JosiahWI:
@maskit makes a good point. Now that I have figured out other issues on my platform I will test with a newer boringssl.

cmcfarlen:
Cherry-picked to v10.0.x

Issue #11566, opened 2024-07-18
☂ Remove the TSUrlHttpParamsGet and TSUrlHttpParamsSet APIs for 10.x


bryancall:
**Note:** Make sure to mark all PRs with an "Incompatible" Label.

**TSUrlHttpParamsGet**
- [x] cookie remap (@bryancall) [#11579]
- [x]  stale reqponse (@bneradt) [#11572]
- [x] header rewrite (@zwoop) [#11570]
- [x]  lua (@shukitchan) [#11580]
- [x] regex_remap (@zwoop) [#11571]
- [x] s3 auth (@maskit) [#11586]
- [x] Cripts (@zwoop) [#11569]

**TSUrlHttpParamsSet**
- [x] Cripts (@zwoop)
- [x] lua (@shukitchan) [#11580]
- [x] url sig (@ezelkow1) [#11599]

**Overall**
- [x] remove docs for TSUrlHttpParamsGet [#11587]
- [x] remove APIs [#11624]

bryancall:
@maskit is going to send email to the dev@ mailing list talking about our decision to remove these APIs for 10.x

shukitchan:
yeah. I will do my bit before end of the week

maskit:
All the todo were done.

PR #11606, opened 2024-07-23
Run regression tests for InkAPI

maskit:
Apparently we lost `--disable-tests` build option when we switched to cmake. And because of that, regression tests for InkAPI are not run since the switch.

We may want to add back the build option, but this PR just include the tests for now.

### DIFF
--- a/src/api/CMakeLists.txt
+++ b/src/api/CMakeLists.txt
@@ -16,7 +16,10 @@
 #######################
 
 # plugin api *only*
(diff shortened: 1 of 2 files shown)


bneradt (changes requested):
This seems to fix the building of InkAPIInternal.cc:

https://github.com/maskit/trafficserver/pull/3

maskit:
Regression tests for `TSUrlHttpParamsSet&Get` failed. We need to finish #11566 first.

bneradt (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11609, opened 2024-07-24
Restore returning string metrics from traffic_ctl match

cmcfarlen:
PR #11127 prevented metrics being returned for config records, but removed returning string metrics.

This PR restores string metrics.

### DIFF
--- a/src/records/RecCore.cc
+++ b/src/records/RecCore.cc
@@ -576,8 +576,7 @@ RecLookupMatchingRecords(unsigned rec_type, const char *match, void (*callback)(
         callback(&tmp, data);
       }
     }
-    // all done for metrics
-    return REC_ERR_OKAY;
+    // Fall through to return any matching string metrics
   }
 
   num_records = g_num_records;


src/records/RecCore.cc
```diff
@@ -576,8 +576,7 @@ RecLookupMatchingRecords(unsigned rec_type, const char *match, void (*callback)(
         callback(&tmp, data);
       }
     }
-    // all done for metrics
-    return REC_ERR_OKAY;
+    // Fall through to return any matching string metrics
```
brbzull0, line 579: so, the `g_records` is still holding metrics??

cmcfarlen: It holds string metrics, yes

cmcfarlen: For example:
```
traffic_ctl metric get proxy.process.version.server.long
```

works but:
```
traffic_ctl metric match proxy.process.version.server.*
```

does not

brbzull0 (approved):
Thanks.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11611, opened 2024-07-24
Parse frame types consisting of multiple bytes

### LINKED ISSUE #11589: ATS 10 HTTP/3 assertion failure on docs server
We noticed this crash on our docs server. It happened on an HTTP/3 QUIC connection.

#### (from traffic.out)
```
Fatal: /home/bneradt/src/trafficserver_10/src/proxy/http3/Http3Frame.cc:59: failed assertion `ret != 1`
```

## Traceback
```
Thread 3814260, [ET_NET 6]:
0    0x00007fcbdaa03510 __interceptor_waitpid.part.0 + 0x40
1    0x00005633aa59c1c1 crash_logger_invoke(int, siginfo_t*, void*) + 0x12a
2    0x00007fcbd9c06420 funlockfile + 0x60
3    0x00007fcbd9a4300b gsignal + 0xcb
4    0x00007fcbd9a22859 abort + 0x12b
5    0x00005633aa613382 ink_abort(char const*, ...) + 0x126
6    0x00005633aa60d1f6 _ink_assert(char const*, char const*, int) + 0x3b
7    0x00005633aafcc03e Http3Frame::type(unsigned char const*, unsigned long) + 0x152
8    0x00005633aafd0644 Http3FrameFactory::fast_create(IOBufferReader&) + 0xf6
9    0x00005633aafa6963 Http3FrameDispatcher::on_read_ready(unsigned long, Http3StreamType, IOBufferReader&, unsigned long&) + 0x90d
10   0x00005633aaf9c134 Http3Transaction::_process_read_vio() + 0x47a
11   0x00005633aaf96ed8 HQTransaction::do_io_read(Continuation*, long, MIOBuffer*) + 0x1fc
12   0x00005633aa6f6122 HttpSM::attach_client_session(ProxyTransaction*) + 0x904
13   0x00005633aacc90c3 ProxyTransaction::new_transaction(bool) + 0x65d
14   0x00005633aaf877c8 Http3App::_handle_bidi_stream_on_read_ready(int, VIO*) + 0x412
15   0x00005633aaf857dc Http3App::main_event_handler(int, Event*) + 0x47c
16   0x00005633aa5aa84b Continuation::handleEvent(int, void*) + 0x1e3
17   0x00005633aaf194d5 EThread::process_event(Event*, int) + 0x347
18   0x00005633aaf19a29 EThread::process_queue(Queue<Event, Event::Link_link>*, int*, int*) + 0x1d5
19   0x00005633aaf19fc0 EThread::execute_regular() + 0x3d6
20   0x00005633aaf1aa6c EThread::execute() + 0x30c
21   0x00005633aaf1770d spawn_thread_internal(void*) + 0x14e
22   0x00007fcbd9bfa609 start_thread + 0xd9
23   0x00007fcbd9b1f353 clone + 0x43
```

#### src/http3/Http3Frame.cc
```cpp
Http3FrameType
Http3Frame::type(const uint8_t *buf, size_t buf_len)
{
  uint64_t type              = 0;
  size_t   type_field_length = 0;
  int      ret               = QUICVariableInt::decode(type, type_field_length, buf, buf_len);
  // [ANNOTATION] This is the failing assertion.
  ink_assert(ret != 1);
```
  
#### src/iocore/net/quic/QUICIntUtil.cc
```cpp
int
QUICVariableInt::decode(uint64_t &dst, size_t &len, const uint8_t *src, size_t src_len)
{
  if (src_len < 1) {
    return -1;
  }
  len = 1 << (src[0] >> 6);
  if (src_len < len) {
    // [ANNOTATION] This is where the unexpected return value originates.
    return 1;
  }
```

### Analysis

Is it correct to assert for `ret != 1` here, or did we intend to assert for `ret != -1`, for the error condition?

JosiahWI:
The variable integer for the frame type can have up to 8 bytes. Without this patch, ATS uses a 1 byte array to store the frame type, and we experience an assertion failure if we don't have enough bytes to decode the variable integer. This increases the array size to 8 bytes.

Fixes #11589.

I have done a few tests in different scenarios locally and I believe this change is safe, but I'm keeping in draft while I consider whether any other tests should be committed to ensure correctness of the multi-byte frame type handling.

### DIFF
--- a/src/proxy/http3/test/test_Http3FrameDispatcher.cc
+++ b/src/proxy/http3/test/test_Http3FrameDispatcher.cc
@@ -29,6 +29,17 @@
 
 TEST_CASE("Http3FrameHandler dispatch", "[http3]")
 {
+  Http3FrameDispatcher  http3FrameDispatcher;
+  Http3MockFrameHandler handler;
+  Http3ProtocolEnforcer enforcer;
+  http3FrameDispatcher.add_handler(&handler);
+  http3FrameDispatcher.add_handler(&enforcer);
+
+  MIOBuffer      *buf    = new_MIOBuffer(BUFFER_SIZE_INDEX_512);
+  IOBufferReader *reader = buf->alloc_reader();
+  uint64_t        nread  = 0;
+  Http3ErrorUPtr  error  = Http3ErrorUPtr(nullptr);
+
   SECTION("Test good case")
   {
     uint8_t input[] = {// 1st frame (HEADERS)
@@ -37,17 +48,6 @@ TEST_CASE("Http3FrameHandler dispatch", "[http3]")
                        0x00, 0x04, 0xaa, 0xbb, 0xcc, 0xdd,
                        // 3rd frame (incomplete)
                        0xff};
-
-    Http3FrameDispatcher  http3FrameDispatcher;
-    Http3MockFrameHandler handler;
-    Http3ProtocolEnforcer enforcer;
-    http3FrameDispatcher.add_handler(&handler);
-    http3FrameDispatcher.add_handler(&enforcer);
-
-    MIOBuffer      *buf    = new_MIOBuffer(BUFFER_SIZE_INDEX_512);
-    IOBufferReader *reader = buf->alloc_reader();
-    uint64_t        nread  = 0;
-    Http3ErrorUPtr  error  = Http3ErrorUPtr(nullptr);
 
     buf->write(input, sizeof(input));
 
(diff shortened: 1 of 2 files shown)


src/proxy/http3/test/test_Http3FrameDispatcher.cc
```diff
@@ -59,9 +59,32 @@ TEST_CASE("Http3FrameHandler dispatch", "[http3]")
...
+
+    buf->write(input, sizeof(input));
+
+    // Initial state
+    CHECK(handler.total_frame_received == 0);
+    CHECK(nread == 0);
+
+    error = http3FrameDispatcher.on_read_ready(0, Http3StreamType::UNKNOWN, *reader, nread);
+    CHECK(!error);
+    CHECK(handler.total_frame_received == 0);
```
maskit, line 83: Shouldn't this be 1?

JosiahWI: Good catch.

src/proxy/http3/test/test_Http3FrameDispatcher.cc
```diff
@@ -225,6 +261,45 @@ TEST_CASE("control stream tests", "[http3]")
...
+
+  http3FrameDispatcher.add_handler(&handler);
+
+  MIOBuffer      *buf    = new_MIOBuffer(BUFFER_SIZE_INDEX_512);
+  IOBufferReader *reader = buf->alloc_reader();
+  uint64_t        nread  = 0;
+  Http3ErrorUPtr  error  = Http3ErrorUPtr(nullptr);
+
+  uint8_t input[] = {
+    0x40, 0x04, // Type
```
maskit, line 279: This is an interesting input. If this were for QUIC frame type, it'd be protocol violation.

RFC 9000 says:
> Values do not need to be encoded on the minimum number of bytes necessary, with the sole exception of the Frame Type field; see [Section 12.4](https://datatracker.ietf.org/doc/html/rfc9000#frames).

> The Frame Type field uses a variable-length integer encoding (see [Section 16](https://datatracker.ietf.org/doc/html/rfc9000#integer-encoding)), with one exception. To ensure simple and efficient implementations of frame parsing, a frame type MUST use the shortest possible encoding.

I don't see a similar limitation on RFC 9114 (HTTP/3).

maskit (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11613, opened 2024-07-24
Remove unused MIOBuffer constructor, add explicit to another.

ywkaras:


include/iocore/eventsystem/IOBuffer.h
```diff
@@ -1236,9 +1236,7 @@ class MIOBuffer
 
   const char *_location = nullptr;
 
-  MIOBuffer(void *b, int64_t bufsize, int64_t aWater_mark);
-  // cppcheck-suppress noExplicitConstructor; allow implicit conversion
-  MIOBuffer(int64_t default_size_index);
```
JosiahWI, line 1241: What's the reason for making this explicit now?

JosiahWI: Makes sense. It's strange someone suppressed the linter check before instead of fixing it.

JosiahWI (commented):
Looks good to me. Needs a non-Yahoo approval.

include/iocore/eventsystem/IOBuffer.h
ywkaras: I would guess that code was written before the `explicit` keyword was available in whatever compiler they were using to build ATS at the time.

cmcfarlen (approved):
Nice cleanup

PR #11614, opened 2024-07-24
traffic_crashlog: use SIGCONT for parent death

cmcfarlen:
With PDEATHSIG set to SIGKILL, the traffic_crashlog was not exiting when traffic_server exited (normally).  Changing the signal to SIGCONT is a little nicer and allows traffic_crashlog to realize that traffic_server has exited and then terminate normally.

I'm not sure why SIGKILL wasn't working


JosiahWI:
This bug caused our docs server to run out of memory resulting in a docs/CI outage. It would be good to get this in.

JosiahWI (approved):
Thanks for the fix. I wonder if `SIGKILL` not working is related to this somehow:

> The signals SIGKILL and SIGSTOP cannot be caught, blocked, or ignored.
 - signal(7) man page

cmcfarlen:
Cherry-picked to v10.0.x

PR #11619, opened 2024-07-28
ACL filters: set_allow/set_deny

bneradt:
Add set_allow and set_deny ACL filter action names, replacing allow and deny as the previous names. These action names are required instead of the previous allow/deny actions with the new MATCH_ON_IP_ONLY matching policy. This replacement allows the new policy to detect if someone updates their policy to the new MATCH_ON_IP_ONLY without explicitly choosing either the new add_allow/add_deny action behavior or the previous action behavior now named set_allow/set_deny.

---

# For Review

I plan to update documentation of this as a separate PR after #11555 lands to avoid inevitable conflicts.


masaori335 (approved):
Looks good. This is basically adding `set_allow` and `set_deny` action. Also, no impact to the 10.0.x behavior with default config.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11628, opened 2024-07-30
Extract `PreservationTable` from `StripeSM`

JosiahWI:
This extracts `PreservationTable` as a superclass. The new class represents the responsibility of tracking documents that need to be preserved to avoid being overwritten, such as pinned documents. Docstrings are added to better document how this process works. The `StripeSM::evac_range` method is left with that class because it is not part of the evacuation data structure despite the name. It should be refactored and renamed in a future PR. The documentation has been updated.

### Related changes also included
 - The visibility of the methods has been changed. Those that can be private are now private.
 - The `force_evacuate_head` method no longer returns a value. We didn't use the return value anywhere.
 - Some private methods of the new class have been renamed to better reflect what they do.

### DIFF
--- a/src/iocore/cache/P_CacheVol.h
+++ b/src/iocore/cache/P_CacheVol.h
@@ -28,6 +28,7 @@
 #include "P_CacheStats.h"
 #include "P_RamCache.h"
 #include "AggregateWriteBuffer.h"
+#include "PreservationTable.h"
 #include "Stripe.h"
 
 #include "iocore/eventsystem/EThread.h"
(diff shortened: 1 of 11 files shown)


src/iocore/cache/P_CacheVol.h
```diff
@@ -73,34 +66,7 @@ struct DiskStripe;
...
-  Dir dir;
-  Dir new_dir;
-  // we need to have a list of evacuationkeys because of collision.
-  EvacuationKey       evac_frags;
-  CacheEvacuateDocVC *earliest_evacuator;
-  LINK(EvacuationBlock, link);
-};
-
-class StripeSM : public Continuation, public Stripe
+class StripeSM : public Continuation, public Stripe, public PreservationTable
```
cmcfarlen, line 69: Does StripeSM need to inherit PreservationTable or can it be aggregate?   It seems like the few public members could be either internalized or proxied.  I prefer more hasA to isA, especially since StripeSM is already spread across multiple classes.

cmcfarlen (commented):
This looks great.  I wonder if StripeSM could haveA PreservationTable instead.  The places that use this interface outside of StateSM might be possible to bring into the class.

src/iocore/cache/P_CacheVol.h
JosiahWI: I don't think it's presently an issue for it to be `isA`, but I'm also thinking it should be `hasA`. My reason was that I want to add a constructor, but bringing in some logic from outside `StateSM` is another great reason. The question is whether to do it immediately in this PR. I didn't do it yet because the added delegation is more complicated than doing the superclass and minimizing the changes minimizes risk, but I think it'd be fine to do it now if you think that's better.

cmcfarlen: I don't feel strongly about it, but I do think it would help clean up some boundaries.  Leaving the more impactful refactoring to another PR sounds smart to me as well.

cmcfarlen (approved):
Nice cleanup

PR #11632, opened 2024-07-31
Remove H3 frame sequence number tracking

### LINKED ISSUE #11622: `Http3FrameDispatcher` counts sequence numbers incorrectly
They are *mostly* counted correctly, but there is a case that isn't. This can be reproduced by adding a protocol enforcer to the "padding should not be interpreted as a data frame" test in test_Http3FrameDispatcher.cc.

```cpp
  uint8_t input[] = {
    0x40, 0x04, // Type
    0x03,       // Length
    0x06,       // Identifier
    0x44, 0x00, // Value
  };
```

This gets read in one byte at a time; if read all at once, the bug doesn't manifest. The padding byte in the frame type causes something to be miscounted, and the protocol enforcer will get called with a sequence number of `-1`.  Because this is a SETTINGS  frame on a CONTROL stream and the sequence number isn't 0, it generates an `H3_FRAME_UNEXPECTED` error.

maskit:
This closes #11622.

The frame sequence was only used for `ProtocolEnforcer` and there is no other use case planned. This PR removes the sequence number tracking, which was reset on every READ_READY, and use a private flag in ProtocolEnforcer to do the same check for SETTINGS frame.

### DIFF
--- a/src/proxy/http3/test/test_Http3FrameDispatcher.cc
+++ b/src/proxy/http3/test/test_Http3FrameDispatcher.cc
@@ -258,44 +258,32 @@ TEST_CASE("control stream tests", "[http3]")
     CHECK(nread == sizeof(input));
   }
 
-  free_MIOBuffer(buf);
-}
+  SECTION("padding should not be interpreted as a DATA frame", "[http3]")
+  {
+    uint8_t input[] = {
+      0x40, 0x04, // Type
+      0x03,       // Length
+      0x06,       // Identifier
+      0x44, 0x00, // Value
+    };
 
-// This test needs to run without an enforcer due to a frame counting bug.
-// Add a ProtocolEnforcer handler to reproduce.
-TEST_CASE("padding should not be interpreted as a DATA frame", "[http3]")
-{
-  Http3FrameDispatcher  http3FrameDispatcher;
-  Http3MockFrameHandler handler;
+    // Initial state
+    CHECK(handler.total_frame_received == 0);
+    CHECK(nread == 0);
 
-  http3FrameDispatcher.add_handler(&handler);
+    int total_nread{};
+    for (uint8_t *it{input}; it < input + sizeof(input); ++it) {
+      buf->write(it, 1);
+      error        = http3FrameDispatcher.on_read_ready(0, Http3StreamType::CONTROL, *reader, nread);
+      total_nread += nread;
+      CHECK(!error);
+    }
 
-  MIOBuffer      *buf    = new_MIOBuffer(BUFFER_SIZE_INDEX_512);
-  IOBufferReader *reader = buf->alloc_reader();
-  uint64_t        nread  = 0;
-  Http3ErrorUPtr  error  = Http3ErrorUPtr(nullptr);
+    CHECK(handler.total_frame_received == 1);
+    CHECK(total_nread == 6);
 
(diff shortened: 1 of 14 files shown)


src/proxy/http3/test/test_Http3FrameDispatcher.cc
```diff
@@ -258,45 +258,33 @@ TEST_CASE("control stream tests", "[http3]")
...
-    error        = http3FrameDispatcher.on_read_ready(0, Http3StreamType::CONTROL, *reader, nread);
-    total_nread += nread;
-    CHECK(!error);
+    free_MIOBuffer(buf);
   }
 
-  CHECK(handler.total_frame_received == 1);
-  CHECK(total_nread == 6);
-
   free_MIOBuffer(buf);
```
JosiahWI, line 288: Doesn't need to be freed twice.

JosiahWI (changes requested):
Nice cleanup! Let's not free the buffer in the test twice, though.

JosiahWI:
The tests for `AggregateWriteBuffer` segfaulted on CentOs. That is unrelated to this change and I will look into it.

src/proxy/http3/test/test_Http3FrameDispatcher.cc
maskit: Good catch.

maskit:
@JosiahWI Do you want to gather information before I push a new commit and trigger new jenkins jobs?

JosiahWI:
@maskit I'm not sure I can gather anything useful from the job. I don't know how to get a stack trace. I suspect it's a race condition, but I don't know how it's related to the test or whether it's related at all. The valgrind report on my OpenSUSE container comes up completely clean.

maskit:
Ok, I'l rerun the job without pushing a new commit to see if it fails again. I'll push a new commit if it doesn't fail.

maskit:
[approve ci centos]

maskit:
All tests passed this time.

JosiahWI (approved):
Thanks for the fix!

maskit:
```
139/139 Test   #9: test_AIO ...............................***Timeout 1500.06 sec
configuration file: sample.cfg
reading disk_size = 1
reading hotset_size = 1
reading hotset_frequency = 0.5
reading run_time = 30
reading threads_per_disk = 1
reading touch_data = 1
reading seq_read_percent = 0.5
reading seq_write_percent = 0.3
reading rand_read_percent = 0.2
reading seq_read_size = 131072
reading seq_write_size = 4093
reading rand_read_size = 4096
reading write_skip = 5
reading chains = 1
reading delete_disks = 1
reading disk_path = ./aio.tst
reading num_processors = 5
Starting hotset document writing 
Finished hotset documents  [8] offset [  8192] size [131072]
Starting the aio_testing 
=================================================================
==4621==ERROR: AddressSanitizer: heap-use-after-free on address 0x6040000001b8 at pc 0x0000004530d4 bp 0x7f0593db2fa0 sp 0x7f0593db2f90
READ of size 1 at 0x6040000001b8 thread T3 ([ET_NET 2])
```

maskit:
[approve ci rocky]

cmcfarlen:
Cherry-picked to v10.0.x

PR #11464, opened 2024-06-20
Remove remaining use of obsolete debug output functions.

ywkaras:

PR #11634, opened 2024-08-01
Fix initialization issue in cache unit tests

JosiahWI:
This fixes a segfault that is likely the same one seen in CI. The diags may be accessed by the global FetchSM allocator, so we have to initialize them before that global is constructed. The name for the diags has been changed because we no longer have access to the unit test metadata at initialization.

### DIFF
--- a/src/iocore/cache/unit_tests/stub.cc
+++ b/src/iocore/cache/unit_tests/stub.cc
@@ -27,6 +27,29 @@
 
 #include "proxy/HttpAPIHooks.h"
 
+#include "tscore/BaseLogFile.h"
+#include "tscore/Diags.h"
+
+class InitDiags
+{
+public:
+  InitDiags() : base_log_file{new BaseLogFile("stderr")}
+  {
+    DiagsPtr::set(new Diags("cache unit test", "*" /* tags */, "" /* actions */, this->base_log_file));
+  }
+
+  ~InitDiags() { delete this->base_log_file; }
+
+private:
+  BaseLogFile *base_log_file;
+};
+
+// The FetchSM class allocator below uses the diags in its constructor, so we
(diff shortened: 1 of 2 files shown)


JosiahWI:
[approve ci]

src/iocore/cache/unit_tests/stub.cc
```diff
@@ -27,6 +27,29 @@
...
+
+private:
+  BaseLogFile *base_log_file;
+};
+
+// The FetchSM class allocator below uses the diags in its constructor, so we
+// force them to be initialized prior to that. This gets "leaked" at shutdown
+// to ensure that the log file remains valid if accessed by other threads
+// during static object destruction.
+InitDiags const *g_init_diags_dummy = new InitDiags{};
```
ywkaras, line 51: This is a leak isn't it?  Why not just:
```
namespace
{
InitDiags init_diags_dummy;
}
```

JosiahWI: That calls could result in a use after three of the log handle, which I think is worse. I'm considering suppressing the ASan report, but I don't like that either. There's really no way to make this work with the current design.

JosiahWI: I think a better way to fix this is to have the allocator wait for the diags to become ready, but we need to make sure this can't cause a deadlock.

ywkaras: The crash this PR solves would not happen if https://github.com/apache/trafficserver/pull/11464 got merged.  (Maybe someday @zwoop  will come out of his review coma.)  The only remaining issue will be that there will be no way to enable debug trace output during static initialization.

JosiahWI:
Will wait for #11464.

PR #11637, opened 2024-08-01
Reduce the number of checks for BoringSSL

maskit:
- Removed one unnecessary check
- Added HAVE_NATIVE_DUAL_CERT_SUPPORT


src/iocore/net/OCSPStapling.cc
```diff
@@ -1309,11 +1310,7 @@ ocsp_update()
 
 // RFC 6066 Section-8: Certificate Status Request
 int
-#if !defined(OPENSSL_IS_BORINGSSL) && !defined(OPENSSL_IS_AWSLC)
-ssl_callback_ocsp_stapling(SSL *ssl)
```
ywkaras, line 1313: This only ever compiled thanks to an evil cast:
https://github.com/openssl/openssl/blob/ed7a8bfd7409ac4a516581f1711d98a9362a70d5/include/openssl/tls1.h#L345

ywkaras (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11639, opened 2024-08-02
Remove `SocketManager`

JosiahWI:
This replaces all remaining usage of `SocketManager` with `UnixSocket` and removes the `SocketManager` namespace and all its associated code. There were no docs to update. I carefully double checked all uses of `UnixSocket::is_ok()` as it is very easy to add or forget a negation.


JosiahWI:
The HTTP/2 unit test failures are likely due to a real issue.

cmcfarlen (approved):

Issue #11607, opened 2024-07-24
traffic_ctl : update ``--records``  output


brbzull0:
It seems we do not need this anymore as in ATS10 we do not support this format anymore.

```
$ traffic_ctl config diff --records                                                                            
CONFIG proxy.config.diags.debug.enabled INT 1 # default: 0                                                                                    
CONFIG proxy.config.diags.debug.tags STRING udp # default: http|dns 
CONFIG proxy.config.quic.qlog.file_base STRING /tmp/ats-stream-test # default: null
CONFIG proxy.config.udp.enable_gso INT 0 # default: 1

```

we can have `YAML` or `JSON` output when `--records` is requested.

PR #11650, opened 2024-08-05
traffic_ctl: Display `YAML` format when `--records` option is set (instead of the legacy `records.config` style)

brbzull0:
This PR changes the `traffic_ctl config` output when `--records` is requested.
`--records` was used to display the output of a record  in the `records.config` configuration style:

```bash
$ traffic_ctl config get  proxy.config.diags.show_location proxy.config.diags.logfile_perm proxy.config.diags.logfile.filename --records

CONFIG proxy.config.diags.show_location INT 2
CONFIG proxy.config.diags.logfile_perm STRING rw-r--r--
CONFIG proxy.config.diags.logfile.filename STRING diags.log
```
As we do not support `records.config` style anymore, it now make sense to have just YAML, so the same command will now display:

```bash
$ traffic_ctl config get  proxy.config.diags.show_location proxy.config.diags.logfile_perm proxy.config.diags.logfile.filename --records
records:
  diags:
    logfile:
      filename: diags.log
    logfile_perm: rw-r--r--
    show_location: 2
```

For things like `traffic_ctl config diff --records` it will add the default value as **standard** YAML comment.

*Note: This is the main reason why this PR uses `YAML::Emitter` and not just the `YAML::Node` to build up the yaml doc, it would have been very easy just to use the `YAML::Node` but it cannot keep track of the comments which for things like the `diff` option is key.*

``` bash
$ traffic_ctl config diff --records
records:
  diags:
    debug:
      tags: quic|udp  # default: http|dns
    show_location: 2  # default: 1
  http:
    server_ports: 8080 4443:ssl 4443:quic  # default: 8080 8080:ipv6
  quic:
    initial_max_streams_bidi_in: 5  # default: 100
    initial_max_streams_bidi_out: 100000  # default: 100
  udp:
    enable_gso: 0  # default: 1
    threads: 1  # default: 0
```
if no `--records` option is passed, the behavior remain unchanged.

fixes: https://github.com/apache/trafficserver/issues/11607

### DIFF
--- a/doc/appendices/command-line/traffic_ctl.en.rst
+++ b/doc/appendices/command-line/traffic_ctl.en.rst
@@ -129,6 +129,13 @@ traffic_ctl alarm
 traffic_ctl config
 ------------------
 
+   Manipulate configuration records.
+
+   .. program:: traffic_ctl config
+   .. option:: --records
+
+   Display the config output in YAML format. This out can be used directly into ATS if needed.
+
 .. program:: traffic_ctl config
 .. option:: defaults [--records]
 
--- a/include/records/RecYAMLDecoder.h
+++ b/include/records/RecYAMLDecoder.h
@@ -56,9 +56,9 @@ swoc::Errata RecYAMLConfigFileParse(const char *path, RecYAMLNodeHandler handler
 /// @brief This function parses the YAML root node ("records") and convert each field
 ///        into a record style object.
 ///
-/// As we keep the internal records without a change we should rebuild each records name
+/// As we keep the internal records without a change we should rebuild each record name
 /// from a YAML structure, this function parses the yaml and while walking down to the scalar
-/// node it builds the record name, this where the handler gets called.
+/// node it builds up the record name, once finishes the handler gets called.
 ///
 /// Example:
 ///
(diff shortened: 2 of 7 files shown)


brbzull0:
Will fix CI:

```
../src/traffic_ctl/PrintUtils.h:85:1: error: 'const char* rec_sourceof(int)' defined but not used [-Werror=unused-function]
   85 | rec_sourceof(int rec_source)
      | ^~~~~~~~~~~~
../src/traffic_ctl/PrintUtils.h:73:1: error: 'const char* rec_labelof(int)' defined but not used [-Werror=unused-function]
   73 | rec_labelof(int rec_class)
      | ^~~~~~~~~~~
../src/traffic_ctl/PrintUtils.h:42:1: error: 'const char* rec_updateof(int)' defined but not used [-Werror=unused-function]
   42 | rec_updateof(int rec_updatetype)
      | ^~~~~~~~~~~~
../src/traffic_ctl/PrintUtils.h:29:1: error: 'const char* rec_accessof(int)' defined but not used [-Werror=unused-function]
   29 | rec_accessof(int rec_access)
```

cmcfarlen (approved):
Worth adding a test or two?

brbzull0:
> Worth adding a test or two?

definitely yes.  I'll add them.
Thanks Chris.

brbzull0:
> Worth adding a test or two?

@cmcfarlen : done -> https://github.com/apache/trafficserver/pull/11740

PR #11651, opened 2024-08-05
Disable mptcp if proper header files are not available

serrislew:
Currently, to log whether a connection used mptcp, we [check the socket option](https://github.com/apache/trafficserver/blob/2adc60f98bf49553c41245b260cf0a1cb44aae65/src/iocore/net/P_UnixNetVConnection.h#L306), which requires specific libraries to be available. This is not done when enabling mptcp in our configurations. Therefore, it is possible that we enable mptcp but our logging says we are not using mptcp so that leads to confusion

### DIFF
--- a/src/records/RecHttp.cc
+++ b/src/records/RecHttp.cc
@@ -98,11 +98,13 @@ mptcp_supported()
   ats_scoped_fd fd(::open("/proc/sys/net/mptcp/enabled", O_RDONLY));
   int           value = 0;
 
+#if defined(HAVE_STRUCT_MPTCP_INFO_SUBFLOWS) && defined(MPTCP_INFO) && MPTCP_INFO == 1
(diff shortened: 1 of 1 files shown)


src/records/RecHttp.cc
```diff
@@ -98,11 +98,13 @@ mptcp_supported()
   ats_scoped_fd fd(::open("/proc/sys/net/mptcp/enabled", O_RDONLY));
```
maskit, line 98: Do we need to try to open this regardless of the conditions?

serrislew: You're right. I will include it in the conditional, thanks!

bneradt (approved):

cmcfarlen:
Cherry-picked to v10.0.x

Issue #11549, opened 2024-07-15
New header_rewrite set-body-from operator could hypothetically lock up Transaction State Machine


ywkaras:
The execution of the set-body-operator initiates an internal URL fetch:
https://github.com/apache/trafficserver/blob/0d8a6aced92f5c8dc5cef3113e34268cc99dfc01/plugins/header_rewrite/operators.cc#L1353

This execution function runs in the READ_RESPONSE_HDR Txn hook.  The operator then runs a continuation on the SEND_RESPONSE_HDR Txn hook, that does not call TSHttpTxnReenable().  The intent is that the hook processing is reenabled once the internal URL fetch completes, here:
https://github.com/apache/trafficserver/blob/0d8a6aced92f5c8dc5cef3113e34268cc99dfc01/plugins/header_rewrite/operators.cc#L66

But, it is not clear there is a mechanism that will prevent the internal fetch from being completed before the operator's continuation runs on the SEND_RESPONSE_HDR.  Also, if other continuations on a Txn hook also uses indirect reenabling, the reenable for the fetch completing could effectively reenable the wrong continuation.

PR #11653, opened 2024-08-06
Allow header_rewrite rules to defer reenabling

jasmine-nahrain:
#11549


plugins/header_rewrite/header_rewrite.cc
```diff
@@ -523,6 +548,8 @@ TSRemapDoRemap(void *ih, TSHttpTxn rh, TSRemapRequestInfo *rri)
     if (rule->eval(res)) {
       OperModifiers rt = rule->exec(res);
 
+      ink_assert((rt & OPER_NO_REENABLE) == 0);
```
ywkaras, line 551: Generally speaking, ink_assert() should not be used in plugin code.  But, since header_rewrite already include core ATS headers, go with the flow I guess.

plugins/header_rewrite/operators.cc
```diff
@@ -64,14 +64,17 @@ handleFetchEvents(TSCont cont, TSEvent event, void *edata)
       TSWarning("[%s] Successful set-custom-body fetch did not result in any content", __FUNCTION__);
     }
     TSHttpTxnReenable(http_txn, TS_EVENT_HTTP_ERROR);
```
ywkaras, line 66: Don't you need to remove the reenables?  Let them be done in `deferRuleDone()`?

plugins/header_rewrite/header_rewrite.cc
```diff
@@ -42,8 +42,27 @@ namespace header_rewrite_ns
 DbgCtl dbg_ctl{PLUGIN_NAME_DBG};
 DbgCtl pi_dbg_ctl{PLUGIN_NAME};
 
-std::once_flag initHRWLibs;
-PluginFactory  plugin_factory;
+std::once_flag   initHRWLibs;
+PluginFactory    plugin_factory;
+std::atomic<int> defer{0};
```
ywkaras, line 47: Unfortunately, I don't think it can work to have a global defer counter.  A per-transaction defer counter is needed.

ywkaras:
The best way I can think of to have a per-transaction pointer to a `defer` counter in the heap is with a user pointer.  (The `authproxy` plugin, for example, uses one.)
```
wkaras ~/LOCAL_REPOS/TS/plugins/authproxy
$ grep -Fn 'TSUserArg' authproxy.cc
186:    opt = static_cast<AuthOptions *>(TSUserArgGet(this->txn, AuthTaggedRequestArg));
676:  return AuthTaggedRequestArg != -1 && TSUserArgGet(txn, AuthTaggedRequestArg) != nullptr;
683:  AuthOptions *opt = static_cast<AuthOptions *>(TSUserArgGet(txn, AuthTaggedRequestArg));
807:  TSReleaseAssert(TSUserArgIndexReserve(TS_USER_ARGS_TXN, "AuthProxy", "AuthProxy authorization tag", &AuthTaggedRequestArg) ==
821:  TSReleaseAssert(TSUserArgIndexReserve(TS_USER_ARGS_TXN, "AuthProxy", "AuthProxy authorization tag", &AuthTaggedRequestArg) ==
858:  TSUserArgSet(txn, AuthTaggedRequestArg, options);
```

I suggest we stick with the approach that works when only one operator per hook needs to defer the transaction reenable ( https://github.com/ywkaras/trafficserver/commit/8ac223775573068fdb57cf5418b3c05d4474ccc0 ).  We can add the approach with a defer counter in the future if we ever need it.

PR #11657, opened 2024-08-06
Fixed crash in XPACK when making space for a new entry

bryancall:
Cleaned up how the tail index in the dynamic table advances and never have it go beyond the head.

### DIFF
--- a/src/proxy/hdrs/XPACK.cc
+++ b/src/proxy/hdrs/XPACK.cc
@@ -496,24 +496,17 @@ XpackDynamicTable::_expand_storage_size(uint32_t new_storage_size)
 bool
 XpackDynamicTable::_make_space(uint64_t extra_space_needed)
 {
-  if (is_empty()) {
-    // If the table is empty, there's nothing to free.
-    return extra_space_needed == 0;
-  }
   uint32_t freed = 0;
   uint32_t tail  = this->_entries_tail;
 
-  while (extra_space_needed > freed) {
+  // Check to see if we need more space and that we have entries to evict
+  while (extra_space_needed > freed && this->_entries_head != tail) {
     tail = this->_calc_index(tail, 1); // Move to the next entry
 
     if (this->_entries[tail].ref_count) {
       break;
     }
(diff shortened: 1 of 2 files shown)


src/proxy/hdrs/XPACK.cc
```diff
@@ -501,25 +501,28 @@ XpackDynamicTable::_make_space(uint64_t extra_space_needed)
...
-      break;
-    }
+    tail = this->_calc_index(tail, 1); // Move to the next entry
+
     if (this->_entries[tail].ref_count) {
       break;
     }
     freed += this->_entries[tail].name_len + this->_entries[tail].value_len + ADDITIONAL_32_BYTES;
-    tail   = this->_calc_index(tail, 1);
+    if (this->_entries_head == tail) {
```
maskit, line 513: Shouldn't this block be at the beginning of the loop? What if `_make_space` is called where the table is empty?

maskit: Or we could also append the check to the condition for the `while` statement.

bryancall: There was an empty conditional at the top of the method.  I moved the check into the while and removed the empty check at the top.

bryancall: This however now changes the result code when the table is empty.  Before it would return true if the table was empty and now it will return false.   I have to look to see if this will cause and issue.  It does pass the unit test either way, which is worrisome.

maskit: > This however now changes the result code when the table is empty. Before it would return true if the table was empty and now it will return false.

Does it? It still returns true if the table is empty AND `extra_space_needed` is 0. It doesn't seem like changed.

bryancall: It looks like changing the result on an empty table is better since that is what we are expecting in `insert_entry()` and `update_maximum()` size we check to see if the table is not empty already.

maskit (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11658, opened 2024-08-07
header_rewrite: Allow Txn reenable to be deferred.

ywkaras:
The SetBodyFrom operator needs to defer calling Txn reenable until the fetch of the URL providing the response body completes.

### DIFF
--- a/plugins/header_rewrite/operator.h
+++ b/plugins/header_rewrite/operator.h
@@ -31,11 +31,12 @@
 
 // Operator modifiers
 enum OperModifiers {
-  OPER_NONE = 0,
-  OPER_LAST = 1,
-  OPER_NEXT = 2,
-  OPER_QSA  = 4,
-  OPER_INV  = 8,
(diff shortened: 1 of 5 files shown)


plugins/header_rewrite/operator.h
```diff
@@ -55,17 +56,22 @@ class Operator : public Statement
   OperModifiers get_oper_modifiers() const;
   void          initialize(Parser &p) override;
 
-  void
+  // Returns number of executed operators that need to defer call to TSHttpTxnReenable().  It is a fatal error if this
+  // returns more than 1.
```
jasmine-nahrain, line 60: Maybe add a comment specifying that if another rule gets created that needs to be deferred, issue #11549 should be revisited.

jasmine-nahrain (changes requested):

jasmine-nahrain (approved):

masaori335 (approved):

Issue #11649, opened 2024-08-05
validity check error when loading all records from a custom records.yaml


brbzull0:
Was doing some tests and I found the following errors, need further investigation on this.

```
[Aug  5 15:21:05.918] traffic_server NOTE: Warn:   Warn: proxy.config.diags.logfile.rolling_min_count - Validity Check failed. '^0*[1-9][0-9]*$' against '0'. Default value will be used
  Warn: proxy.config.http2.default_buffer_water_mark - Validity Check failed. '^[0-9]+$' against '-1'. Default value will be used
  Warn: proxy.config.log.rolling_min_count - Validity Check failed. '^0*[1-9][0-9]*$' against '0'. Default value will be used
  Warn: proxy.config.net.additional_accepts - Validity Check failed. '^-1|[0-9]+$' against '-1'. Default value will be used
  Warn: proxy.config.output.logfile.rolling_min_count - Validity Check failed. '^0*[1-9][0-9]*$' against '0'. Default value will be used
  Warn: proxy.config.ssl.client.version.max - Validity Check failed. '^-?[0-9]+$' against '-1'. Default value will be used
  Warn: proxy.config.ssl.client.version.min - Validity Check failed. '^-?[0-9]+$' against '-1'. Default value will be used
  Warn: proxy.config.ssl.server.version.max - Validity Check failed. '^-?[0-9]+$' against '-1'. Default value will be used
  Warn: proxy.config.ssl.server.version.min - Validity Check failed. '^-?[0-9]+$' against '-1'. Default value will be used

```

brbzull0:
Some of this are fixed in https://github.com/apache/trafficserver/pull/11654

brbzull0:
Adding some notes here


[proxy.config.http2.default_buffer_water_mark](http://fedora/admin-guide/files/records.yaml.en.html#proxy-config-http2-default-buffer-water-mark)
```
When set to -1 (default), a thread will accept connections as long as there are connections waiting in its listening queue.is equivalent to “accept all”, and setting to 0 is equivalent to “accept one”.
```
[...]

PR #11659, opened 2024-08-07
NetHandler - Use the right type to hold `additional_accepts` config value.

brbzull0:
[proxy.config.net.additional_accepts ](https://ci.trafficserver.apache.org/job/Github_Builds/job/docs/5594/artifact/output/11648/docbuild/html/admin-guide/files/records.yaml.en.html#proxy-config-net-additional-accepts) is expected to  use `-1`  but the code wasn't storing the value into a signed type, so changed from `unsigned` to `signed` as this is what's expected. 

This PR also makes sure the records validity check passes.

Original PR -> https://github.com/apache/trafficserver/pull/10098

Part of a fix for https://github.com/apache/trafficserver/issues/11649

### DIFF
--- a/include/iocore/net/NetHandler.h
+++ b/include/iocore/net/NetHandler.h
@@ -236,7 +236,7 @@ private:
   // TS_EVENT_MGMT_UPDATE event like with the Config settings above because
   // accept threads are not always on a standard NET thread with a NetHandler
   // that has TS_EVENT_MGMT_UPDATE handling logic.
-  static std::atomic<uint32_t> additional_accepts;
+  static std::atomic<int32_t>  additional_accepts;
   static std::atomic<uint32_t> per_client_max_connections_in;
 
   void _close_ne(NetEvent *ne, ink_hrtime now, int &handle_event, int &closed, int &total_idle_time, int &total_idle_count);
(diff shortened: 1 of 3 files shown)


brbzull0:
[approve ci autest]

brbzull0:
[approve ci autest]

sharkleash:
Thanks for finding this, would the behavior be the same in this case? It is definitely a bug that additional_accepts isn't a signed int, but doesn't setting an unsigned int to -1, make it equal to 2^32 - 1 (INT32_MAX - 1)? In that case, there's a section of code in NetHandler.cc that adds 1 to that, making it overflow to 0, and then sets it to 2^32 - 1 anyway. 

this is that code: 
`int config_value = additional_accepts.load(std::memory_order_relaxed) + 1;
  return (config_value > 0 ? config_value : INT32_MAX - 1);`

It's certainly a bug, but if the behavior is the same, maybe this doesn't need to go into 10.0.x?

brbzull0:
> Thanks for finding this, would the behavior be the same in this case? 

No bother, thanks for looking at it. Yes, same behavior.

> It's certainly a bug, but if the behavior is the same, maybe this doesn't need to go into 10.0.x?

Well, it partialy depends on where https://github.com/apache/trafficserver/pull/11667  lands, **if** https://github.com/apache/trafficserver/pull/11667 lands in 10  then this should also land in 10, at least part of this, to avoid:

```
traffic_server ERROR: proxy.config.net.additional_accepts: Default value consistency check failed. default value=-1 pattern=^-1|[0-9]+$.
```

Thanks.

moonchen (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11660, opened 2024-08-07
Extend lifetimes of `DbgCtl` instances in inkcache

JosiahWI:
This ensures that threads can access these instances safely at any time. It should help clean up TSan reports a little bit by preventing data races with the destructors, and use-after-free issues.


JosiahWI:
The AuTest header_rewrite_url failed because the connection timed out. These usually indicate a deadlock or similar issue.

JosiahWI:
The tagging system won't be cleaned up if we do this.

PR #11664, opened 2024-08-10
Suppress CID 1523641 - dead code in XPACK

JosiahWI:
The variable `use_huffman` is used to decide which branch to take, but is hardcoded to `true`. This has already been suppressed for cppcheck, and Coverity says we can suppress it by declaring the variable with `const`. I think `constexpr` should also work, but I couldn't confirm this.


maskit (approved):
Let's give it a a shot.

PR #11668, opened 2024-08-12
Fix a bunch of COPY_INSTEAD_OF_MOVE Coverity CIDs

JosiahWI:


brbzull0:
when you say, "a bunch" you mean it! :rofl:

src/iocore/net/SSLSessionCache.cc
```diff
@@ -353,7 +354,7 @@ SSLOriginSessionCache::insert_session(const std::string &lookup_key, SSL_SESSION
   // Create the shared pointer to the session, with the custom deleter
   std::shared_ptr<SSL_SESSION>      shared_sess(sess_ptr, SSLSessDeleter);
   ssl_curve_id                      curve = (ssl == nullptr) ? 0 : SSLGetCurveNID(ssl);
-  std::unique_ptr<SSLOriginSession> ssl_orig_session(new SSLOriginSession(lookup_key, curve, shared_sess));
+  std::unique_ptr<SSLOriginSession> ssl_orig_session(new SSLOriginSession(lookup_key, curve, std::move(shared_sess)));
```
cmcfarlen, line 357: You can avoid the temp and move if you move the construction to the argument:
```
  std::unique_ptr<SSLOriginSession> ssl_orig_session(new SSLOriginSession(lookup_key, curve, std::shared_ptr<SSL_SESSION>(sess_ptr, SSLSessDeleter)));
```

src/iocore/net/SSLUtils.cc
```diff
@@ -1677,7 +1679,8 @@ SSLMultiCertConfigLoader::_store_ssl_ctx(SSLCertLookup *lookup, const shared_SSL
   std::vector<SSLLoadingContext> ctxs = this->init_server_ssl_ctx(data, sslMultCertSettings.get());
   for (const auto &loadingctx : ctxs) {
     shared_SSL_CTX ctx(loadingctx.ctx, SSL_CTX_free);
-    if (!sslMultCertSettings || !this->_store_single_ssl_ctx(lookup, sslMultCertSettings, ctx, loadingctx.ctx_type, common_names)) {
+    if (!sslMultCertSettings ||
+        !this->_store_single_ssl_ctx(lookup, sslMultCertSettings, std::move(ctx), loadingctx.ctx_type, common_names)) {
```
cmcfarlen, line 1683: Could move the construction to the arglist here as well.  This will also avoid an unused tmp if the disjunction shorts

src/traffic_layout/engine.cc
```diff
@@ -276,7 +277,7 @@ LayoutEngine::create_runroot()
         std::string value = it.second.as<std::string>();
         auto        iter  = new_map.find(key);
         if (iter != new_map.end()) {
-          iter->second = value;
+          iter->second = std::move(value);
```
cmcfarlen, line 280: You can just remove the `value` variable as its only temp

src/traffic_layout/engine.cc
```diff
@@ -362,7 +363,7 @@ LayoutEngine::remove_runroot()
     return;
   }
 
-  std::string clean_root = path;
+  std::string clean_root = std::move(path);
```
cmcfarlen, line 366: Not sure what the intent here was.  Does this code want a copy of path? If so, your std::move is counter to that goal.  If not, then this is an unnecessary temp.

cmcfarlen (changes requested):
These sorts of optimizations seem premature to me and unnecessary, even to appease coverity.  IMO, we should try to fix the code so that move is less necessary.

plugins/experimental/stek_share/log_store.cc
```diff
@@ -90,7 +91,7 @@ STEKShareLogStore::write_at(uint64_t index, nuraft::ptr<nuraft::log_entry> &entr
   while (itr != logs_.end()) {
     itr = logs_.erase(itr);
   }
-  logs_[index] = clone;
+  logs_[index] = std::move(clone);
```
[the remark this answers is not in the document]
JosiahWI: The reason this uses a temporary is to reduce the amount of work done while holding the logs lock. Thoughts?

plugins/statichit/statichit.cc
```diff
@@ -71,7 +73,7 @@ struct StaticHitConfig {
     base_path = std::filesystem::weakly_canonical(base_path);
 
     if (std::filesystem::is_directory(base_path)) {
-      dirPath      = base_path;
+      dirPath      = std::move(base_path);
```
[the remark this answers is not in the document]
JosiahWI: I'm sure it does. I'll figure out which CID it's for and fix it, too.

src/traffic_layout/engine.cc
JosiahWI: I think the intent was to describe the data contained in the string, even though it's the same string object being modified. I think `clean_root` should be a reference - there's no need to copy or move.

plugins/statichit/statichit.cc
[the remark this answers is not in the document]
JosiahWI: Weirdly, Coverity does not have a CID for it. Maybe it has to do with the difference in types: `std::filesystem::path` vs `std::string`. I changed it anyway for consistency.

plugins/experimental/stek_share/log_store.cc
```diff
@@ -83,6 +83,7 @@ STEKShareLogStore::append(nuraft::ptr<nuraft::log_entry> &entry)
 void
 STEKShareLogStore::write_at(uint64_t index, nuraft::ptr<nuraft::log_entry> &entry)
 {
+  // We prepare what we can taking the logs lock to minimize time holding it.
```
cmcfarlen, line 86: Missing the word `before` perhaps?

cmcfarlen (approved):
might be a typo in the last commit

cmcfarlen (approved):

PR #11671, opened 2024-08-12
Add algorithm to calculate a JA4 fingerprint

JosiahWI:
This is the beginning of a new experimental ja4_fingerprint plugin. This is just the part of the algorithm described by the JA4 specification that doesn't depend on OpenSSL. This includes thorough unit tests. I intend to do the plugin and OpenSSL logic in a separate PR with AuTests.


ywkaras:
A strategic issue involved here is whether we need to reduce per-transaction memory allocation/deallocation.  Are we assuming that newer allocators with per-thread arenas make it a non-issue?

In a Yahoo internal plugin, I am using thread_local instances of this class to reduce memory allocation/deallocation:
```
class GrowBuf
{
public:
  explicit GrowBuf(unsigned capacity = 0)
  {
    _v.reserve(capacity);
  }

  unsigned size() const { return _size; }

  const unsigned char * data() const { return _v.data(); }

  void clear() { _size = 0; }

  void push_back(unsigned char b)
  {
    if (_size < _v.size()) {
      _v[_size++] = b;
    } else {
      _v.push_back(b);
      ++_size;
    }
  }

  // No copy/move.
  GrowBuf(GrowBuf const &) = delete;
  GrowBuf & operator = (GrowBuf const &) = delete;

private:
  std::vector<unsigned char> _v;
  unsigned _size{0};
};
```

JosiahWI:
I do want to reserve bytes for the strings in some cases because I know exactly how long they will be, so I'm glad you reminded me of that. For the other allocations, I would be hesitant to make any modifications to optimize allocations without benchmarking, and I'd prefer to do it in a separate PR.

JosiahWI:
The test Alternate_S_to_L_remove_S segfaulted on OSX.

JosiahWI:
[approve ci osx]

moonchen:
Thanks for making the contribution.  A suggestion if you're developing this plugin further in the future: perhaps it's worthwhile to use a struct to hold a JA4 signature, with its multiple fields of fixed length strings.  This should allow you to allocate all of those fixed length fields in one allocation and fill it out without additional allocations.

PR #11673, opened 2024-08-13
Fatal error if no udp threads but using quic

mattyw:
Fatal error if no udp threads but using quic

If proxy.config.http.server_ports contains an entry to allow quic
connections on a given port but there no udp threads ATS used to
silently ignore these connections.

Given that this configuration isn't valid we raise a fatal error.

We also raise a fatal error if quic is listed in server ports but ATS
was built without quic support.

### DIFF
--- a/src/traffic_server/traffic_server.cc
+++ b/src/traffic_server/traffic_server.cc
@@ -2215,6 +2215,21 @@ main(int /* argc ATS_UNUSED */, const char **argv)
       REC_ReadConfigInteger(num_of_udp_threads, "proxy.config.udp.threads");
     }
 
+    ats_scoped_str server_ports(255);
+    *server_ports = '\0';
+    REC_ReadConfigString(server_ports, "proxy.config.http.server_ports", 255);
+    bool quic_server_port = strstr(server_ports, "/quic/") != server_ports;
+
+#if TS_USE_QUIC == 0
+    if (quic_server_port) {
+      Fatal("proxy.config.http.server_ports listening for quic but traffic_server wasn't built with quic support.");
+    }
+#endif
+
+    if (quic_server_port && num_of_udp_threads == 0) {
+      Fatal("proxy.config.http.server_ports listening for quic but proxy.config.udp.threads is 0.");
+    }
+
     udpNet.register_event_type();
     if (num_of_udp_threads) {
       udpNet.start(num_of_udp_threads, stacksize);


src/traffic_server/traffic_server.cc
```diff
@@ -2215,6 +2215,15 @@ main(int /* argc ATS_UNUSED */, const char **argv)
       REC_ReadConfigInteger(num_of_udp_threads, "proxy.config.udp.threads");
     }
 
+    ats_scoped_str server_ports(255);
```
mattyw, line 2218: 255 chosen arbitrarily - would appreciate guidance of a better number

JosiahWI:
[approve ci]

brbzull0 (commented):
Not sure about this, but there is a `HttpProxyPort::Group` which gets filled in and have the information you need.

src/traffic_server/traffic_server.cc
```diff
@@ -2215,6 +2215,21 @@ main(int /* argc ATS_UNUSED */, const char **argv)
...
+    *server_ports = '\0';
+    REC_ReadConfigString(server_ports, "proxy.config.http.server_ports", 255);
+    bool quic_server_port = strstr(server_ports, "/quic/") != server_ports;
+
+#if TS_USE_QUIC == 0
+    if (quic_server_port) {
+      Fatal("proxy.config.http.server_ports listening for quic but traffic_server wasn't built with quic support.");
```
maskit, line 2225: This check could be done as a part of this block, if we want to have a specialized message for QUIC (the block already has code for unrecognized options).
https://github.com/apache/trafficserver/blob/fc5918e5a8bb1eb88b98689c463c3574a64e8bef/src/records/RecHttp.cc#L367-L463

src/traffic_server/traffic_server.cc
```diff
@@ -2215,6 +2215,21 @@ main(int /* argc ATS_UNUSED */, const char **argv)
...
+    }
+#endif
+
+    if (quic_server_port && num_of_udp_threads == 0) {
+      Fatal("proxy.config.http.server_ports listening for quic but proxy.config.udp.threads is 0.");
```
maskit, line 2230: I'm not sure if we should use `Fatal`. Even port configuration error is just a warning. It depends on the use case for sure, but QUIC being unavailable is not the end of the world.

src/traffic_server/traffic_server.cc
mattyw: @maskit Yeah good suggestion! I tried looking into the records code to see a good place to put this logic in but couldn't find a good place. I see if I can make progress with your suggestion here.

src/traffic_server/traffic_server.cc
mattyw: @maskit I think you might be right w.r.t warning - I bounced between fatal and warning a few times before proposing.

> but QUIC being unavailable is not the end of the world.

Well, QUIC being unavailable when you've specifically asked ATS to serve quic could be considered to be the end of the world - it would certainly be surprising enough that I'd think we'd want to catch it sooner rather than later

maskit: How you catch misconfiguration and how ATS behave while configuration may be wrong are different things. You could watch for a warning to catch misconfiguration, or you could also watch for the UDP ports. Ideally you should periodically send health check requests. Look at the code I mentioned above. You can try to enable mptcp where mptcp is unavailable on the host and that does not stop ATS. Lack of QUIC does not threaten the safety. QUIC can be unavailable even with proper configuration if middle boxes block it. I still think Fatal is too much.

PR #11678, opened 2024-08-15
Fix JA4 TLS version encoding

JosiahWI:
The JA4 specification gives a table of which versions are recognized and what strings should be generated for them. The JA4 module incorrectly takes the digits of the version instead of using the string given by the specification. This fixes it to match up with the spec.


bneradt (commented):
lgtm

moonchen (approved):

moonchen (approved):

PR #11680, opened 2024-08-15
Upgrade notes: removal of disable_h2 sni.yaml

bneradt:
This adds upgrade notes of the removal of disable_h2 from the sni.yaml file.

### DIFF
--- a/doc/release-notes/upgrading.en.rst
+++ b/doc/release-notes/upgrading.en.rst
@@ -99,7 +99,7 @@ The following incompatible changes to the configurations have been made in this 
 
 The ``records.config`` file has been renamed to :file:`records.yaml` and now it is structured in YAML format.
 
-The following :file:`records.yaml` chnages have been made:
+The following :file:`records.yaml` changes have been made:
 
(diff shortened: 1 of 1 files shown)


doc/release-notes/upgrading.en.rst
```diff
@@ -64,6 +64,8 @@ Configuration changes
 ---------------------
 The following incompatible changes to the configurations have been made in this version of ATS.
 
+The following changes have been made to the :file:`records.yaml` file:
```
maskit, line 67: Sounds a bit odd because records.yaml is technically a new config file, mentioning file names is a good idea though. Let's adjust it (if needed) on another PR that explains the transition.

maskit (approved):

doc/release-notes/upgrading.en.rst
bneradt: What do you think now? Now, via another PR, the change to `records.yaml` is mentioned before this.

maskit: Now it looks fine.

maskit (approved):

PR #11684, opened 2024-08-15
Upgrade notes: Removal of mgmt libraries

cmcfarlen:


doc/release-notes/upgrading.en.rst
```diff
@@ -49,6 +49,10 @@ removed in the next major release of ATS.
   * TSVConnArgIndexNameLookup
   * TSVConnArgIndexLookup
 
+* Removed Internal Libraries
+
+  * mgmt_p
```
bryancall, line 54: I don't think we need to document changes to internal libs, if end users won't notice these have been removed.

bryancall (approved):

maskit (approved):

PR #11705, opened 2024-08-16
Linkify records.yaml references in release notes

bneradt:
This makes records.yaml config references links to the configuration.

It also removes redundant references to connect_attempts_timeout being removed.


cmcfarlen:
It looks like other active PRs have some of these already updated

PR #11707, opened 2024-08-16
Fix whats-new links and remove redundant timeout

bneradt:
whats-new has malformed link references to non-existent links. This fixes that. Also connect_attempts_timeout changes are referenced twice in upgrading in a way that doesn't seem helpful: we say first that we removed them, then in a separate bullet we say that they are removed from default config. The latter is implied by the former.

### DIFF
--- a/doc/release-notes/upgrading.en.rst
+++ b/doc/release-notes/upgrading.en.rst
@@ -147,9 +147,6 @@ The following :file:`records.yaml` changes have been made:
   from ``PERMISSIVE`` to ``STRICT``.
 - The records.yaml entry ``proxy.config.http.keepalive_internal_vc`` has been removed.  This entry
   was previously undocumented.
-- The records.yaml entries ``proxy.config.http.parent_proxy.connect_attempts_timeout`` and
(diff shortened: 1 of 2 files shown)


maskit:
The line between upgrading and whats-new is getting blurry. I thought whats-new only have things purely new (i.e. upgrading without reading whats-new doesn't harm anything).

bneradt:
> The line between upgrading and whats-new is getting blurry. I thought whats-new only have things purely new (i.e. upgrading without reading whats-new doesn't harm anything).

Yeah, I was just fixing the formatting there. Want me to remove that section all together? It's already covered in upgrading. It seems to me that it doesn't belong in "whats new".

maskit:
Yeah, that's what implied. If we remove either of them I'd remove the one in whats-new.

bneradt:
> Yeah, that's what implied. If we remove either of them I'd remove the one in whats-new.

Gotcha. Thanks for the thought. Updated with the removal of the section.

cmcfarlen (approved):

PR #11718, opened 2024-08-19
Fix use-of-uninitialized value problem found by fuzzing

shukitchan:
Details here - https://oss-fuzz.com/testcase-detail/4793610426449920

### DIFF
--- a/src/proxy/http3/Http3Frame.cc
+++ b/src/proxy/http3/Http3Frame.cc
@@ -508,7 +508,7 @@ Http3FrameFactory::create(IOBufferReader &reader)
   ts::Http3Config::scoped_config params;
   Http3Frame                    *frame = nullptr;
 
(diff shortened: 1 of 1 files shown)


JosiahWI:
It'd be nice to have see the input that caused this. Zero-initializing the array may not be an appropriate fix, because if the data in the array is being processed past the correct number of bytes, we're probably reading an incorrect value.

JosiahWI:
I do not have authorization to the above link.  Thank you for the stack trace.

shukitchan:
The input causing this is here 
[clusterfuzz-testcase-fuzz_http3frame-4793610426449920.txt](https://github.com/user-attachments/files/16665283/clusterfuzz-testcase-fuzz_http3frame-4793610426449920.txt)

shukitchan:
Yeah, not a good idea at this point to initialize with zero.

PR #11722, opened 2024-08-20
Move `Cache::` methods to Cache.cc

JosiahWI:
This is purely a mechanical change. It also results in the complete removal of the CacheVol.cc source file, since it no longer contains anything.


src/iocore/cache/Cache.cc
```diff
@@ -43,6 +45,9 @@
 #include <system_error>
 #include <filesystem>
 
+#define SCAN_BUF_SIZE              RECOVERY_SIZE
+#define SCAN_WRITER_LOCK_MAX_RETRY 5
```
masaori335, line 49: [nitpick] `constexpr` is preferred.

masaori335 (approved):
Looks good 👍

src/iocore/cache/Cache.cc
JosiahWI: I will make a note of it - I think it'd be good to change these all at once over the whole subsystem for consistency.

PR #11723, opened 2024-08-20
QUIC: Fix a bug that ssl_multicert.config is not reloaded for QUIC

maskit:
The fix is in SSLClientCoordinator. It's simply lacks the line for QUIC.

The log output format is slightly changed because it'd look like same config file is reloaded twice if it doesn't have a tag name (The file is loaded twice for two contexts, Regular TLS (on TCP) and QUIC).


JosiahWI (dismissed):
Thanks for the fix!

@maskit Looks like the new line needs to be conditionally compiled based on whether QUIC is being built.

JosiahWI (approved):

Issue #11725, opened 2024-08-20
ATS 10 traffic_ctl doesn't get a warning when you don't have permissions


bryancall:
On a production host I am not seeing a warning when I don't have permissions and I am trying to set a config option:

```
[bcall@e5 trafficserver]$ /opt/edge/1.1/trafficserver/10.0/bin/traffic_ctl config set proxy.config.http.connect.down.policy 0
[bcall@e5 trafficserver]$
```

brbzull0:
Just tested this, the error itself is that  ATS is not respoding with the right error response:
```bash
$ traffic_ctl config set proxy.config.diags.debug.tags "http.*|dns.*"
Server Error found:
[10] Unauthorized action
- [2] Denied privileged API access for uid=105106 gid=100
```
in any case the record is **not set** unless you use `sudo` which is ok, so no harm will be done.
The issue was introduced by https://github.com/apache/trafficserver/pull/10986 few months ago.

I'll make a patch for this.

PR #10986, opened 2024-01-15
libsowc - remove legacy Errata.

brbzull0:
This is a continuation of the work started by Alan (https://github.com/apache/trafficserver/pull/10112) +
-  Update for the JSONRPC logic which wasn't complete on his original PR due to changes needed in the libswoc which are now completed.
- Make sure all autests/cpptests are green.

co-author: @SolidWallOfCode

PR #11730, opened 2024-08-21
RPC: Make sure any internal(not handlers) error gets properly set.

brbzull0:
Thus, the RPC server respond with a proper error response. 
This issue was introduced by #10986 when migrated from `ts:Errata` to `swoc::Errata`. This PR makes sure that the right error code gets set

fixes https://github.com/apache/trafficserver/issues/11725

### DIFF
--- a/src/mgmt/rpc/jsonrpc/JsonRPCManager.cc
+++ b/src/mgmt/rpc/jsonrpc/JsonRPCManager.cc
@@ -62,12 +62,11 @@ swoc::Rv<YAML::Node>    g_rpcHandlerResponseData;
 bool                    g_rpcHandlerProcessingCompleted{false};
 
 // --- Helpers
-swoc::Errata
+std::pair<swoc::Errata, error::RPCErrorCode>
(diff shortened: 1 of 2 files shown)


brbzull0:
[approve ci autest]

src/mgmt/rpc/jsonrpc/JsonRPCManager.cc
```diff
@@ -112,7 +111,7 @@ JsonRPCManager::Dispatcher::dispatch(Context const &ctx, specs::RPCRequestInfo c
   }
 
   // We have got a valid handler, we will now check if the context holds any restriction for this handler to be called.
-  if (auto errata = check_for_blockers(ctx, handler.get_options()); !errata.is_ok()) {
+  if (auto &&[errata, ec] = check_for_blockers(ctx, handler.get_options()); !errata.is_ok()) {
```
brbzull0, line 114: The error was detected by this check, but the error code was never set. This fixes that as `ec`  now holds the error.

masaori335 (approved):
Looks good.

PR #11731, opened 2024-08-21
Sever `PreservationTable` inheritance and add utility methods

JosiahWI:
This adds the following methods:

* find(Dir const &)
* acquire(Dir const&, CacheKey const&)
* release(Dir const&)

The latter two are used for reference counting readers of documents so that the documents will be preserved while they're open for reading.

### DIFF
--- a/src/iocore/cache/CacheEvacuateDocVC.cc
+++ b/src/iocore/cache/CacheEvacuateDocVC.cc
@@ -55,9 +55,8 @@ CacheEvacuateDocVC::evacuateDocDone(int /* event ATS_UNUSED */, Event * /* e ATS
   DDbg(dbg_ctl_cache_evac, "evacuateDocDone %X o %d p %d new_o %d new_p %d", (int)key.slice32(0),
        (int)dir_offset(&this->overwrite_dir), (int)dir_phase(&this->overwrite_dir), (int)dir_offset(&this->dir),
        (int)dir_phase(&this->dir));
-  int i = dir_evac_bucket(&this->overwrite_dir);
   // nasty beeping race condition, need to have the EvacuationBlock here
-  EvacuationBlock *b = this->stripe->evac_bucket_valid(i) ? this->stripe->get_preserved_dirs().find(this->overwrite_dir) : nullptr;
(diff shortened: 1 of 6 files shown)


src/iocore/cache/CacheEvacuateDocVC.cc
```diff
@@ -57,88 +57,84 @@ CacheEvacuateDocVC::evacuateDocDone(int /* event ATS_UNUSED */, Event * /* e ATS
...
+    // only overwrite the entry in the directory if its not a head.
+    if (!dir_head(&this->overwrite_dir)) {
+      // find the earliest key
+      EvacuationKey *evac = &b->evac_frags;
+      for (; evac && !(evac->key == doc->key); evac = evac->link.next) {
+        ;
+      }
+      ink_assert(evac);
+      if (!evac) {
+        goto RET;
```
masaori335, line 77: How about simply returning from here?

If calling `free_CacheEvacuateDocVC(this)` is the matter, setting defer by `ts::PostScript` in the beginning is another apporach.
```
ts::PostScript defer_free([&](){ free_CacheEvacuateDocVC(this); });
```

masaori335 (changes requested):
Overall, I love this change, thank you! Please take a look at my comments.

src/iocore/cache/CacheEvacuateDocVC.cc
JosiahWI: In other branches we don't free it. Maybe the cleanest thing for now is to simply return as you say and also duplicate the call to free.

src/iocore/cache/StripeSM.cc
```diff
@@ -1295,13 +1261,13 @@ evacuate_fragments(CacheKey *key, CacheKey *earliest_key, int force, StripeSM *s
     if (dir_head(&dir)) {
       continue;
     }
-    EvacuationBlock *b = evacuation_block_exists(&dir, stripe);
```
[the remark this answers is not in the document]
JosiahWI: Good catch!

masaori335 (approved):
Looks good!

PR #11734, opened 2024-08-22
Changes all of Cripts to be in cripts:: namespace

zwoop:
This turns out to be required, for Cripts to be able to include core include files (like HttpSM.h) directly. I would like to do that later, bypassing InkAPI in Cripts directly.

This also feels more consistent within all of Cripts.

### DIFF
--- a/doc/developer-guide/cripts/cripts-bundles.en.rst
+++ b/doc/developer-guide/cripts/cripts-bundles.en.rst
@@ -41,14 +41,14 @@ does *not* exclude doing additional hooks in the Cript itself.
 
 The following bundles are available in the core today:
 
-============================   ====================================================================
-Bundle                         Description
-============================   ====================================================================
-``Bundle::Common``             For DSCP and an overridable Cache-Control header.
-``Bundle::LogsMetrics``        Log sampling, TCPInfo  and per-remap metrics.
-``Bundle::Headers``            For removing or adding headers.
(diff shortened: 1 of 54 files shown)


maskit:
This is going to be an incompatible change. What's the plan?

zwoop:
[approve ci]

zwoop:
> This is going to be an incompatible change. What's the plan?

Yeh, incompatible. But, we also haven't declared Cripts "stable" yet, so I'm hoping no one will object for this to go into 10.1.0.

I think if for some other reason we re-spin 10.0.0, we should get this in there, but I won't -1 the current vote just for this.

zwoop:
[approve ci]

bneradt (changes requested):
In general the change looks straightforward. My initial thought, though, is that I'm a bit surprised that the namespace is `Cript` instead of `Cripts`. Everywhere our document calls the feature `Cripts`. 

> Cripts, C-Scripts, is a set of wrappers and include

> Cripts is built as a separate library

We should probably choose one or the other. Is it "Cript is a such and such..." or "Cripts is a such and such...", then name the namespace accordingly. Currently per our documentation it seems like we should call the namespace `Cripts`.

bneradt (approved):
lgtm

cmcfarlen:
Cherry-picked to v10.0.x

PR #11736, opened 2024-08-23
Add gcp auth & rename s3_auth plugin to origin_server_auth

jasmine-nahrain:
Adding GCP authentication to existing `s3_auth` plugin. As this makes the `s3_auth` plugin no longer exclusively s3. I have renamed the plugin to `origin_server_auth`. This better reflects the full functionality of the plugin.

I have changed the parameter `version` from an int to a string. Instead of specifying 2 or 4 for the version, you must specify awsv2, awsv4 or gcpv1. This should allow adding more versions/auth services easier. The old versioning is now deprecated and should be removed in a later release.

`s3_auth` is now deprecated.


duke8253 (approved):
Looks good to me.

PR #11737, opened 2024-08-23
Decouple disk from stripe

JosiahWI:
This moves the following variables out of `Stripe`:

- [x] `fd`
- [x] `path`
- [x] `disk`
- [ ] ~~`hash_text`~~

The `hash_text` is used for some logging, so it might take some more design changes or possibly changes to the logging to easily move it.

### DIFF
--- a/src/iocore/cache/StripeSM.h
+++ b/src/iocore/cache/StripeSM.h
@@ -78,6 +78,8 @@ public:
 
   Event *trigger = nullptr;
 
+  CacheDisk *disk{};
+
   OpenDir              open_dir;
   RamCache            *ram_cache = nullptr;
   DLL<EvacuationBlock> lookaside[LOOKASIDE_SIZE];
--- a/src/iocore/cache/StripeSM.h
+++ b/src/iocore/cache/StripeSM.h
@@ -66,6 +66,7 @@ class StripeSM : public Continuation, public Stripe
 {
 public:
   CryptoHash hash_id;
+  int        fd{-1};
 
   int hit_evacuate_window{};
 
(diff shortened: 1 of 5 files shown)


masaori335:
A tricky idea is adding a pointer to the `StripeSM` to the `Stripe`.
https://github.com/JosiahWI/trafficserver/pull/2/files

src/iocore/cache/StripeSM.h
```diff
@@ -65,6 +65,8 @@ class StripeSM : public Continuation, public Stripe
 {
 public:
   CryptoHash hash_id;
+  char      *path = nullptr;
+  int        fd{-1};
```
masaori335, line 69: `CacheDisk` has `path` and `fd`. I start wondering we should just refer them instead of copy it because `StripeSM` has `CacheDisk *`.
https://github.com/apache/trafficserver/blob/d0b6a4b49ba8f2738fe91606c9cea55b06220cec/src/iocore/cache/P_CacheDisk.h#L86
https://github.com/apache/trafficserver/blob/d0b6a4b49ba8f2738fe91606c9cea55b06220cec/src/iocore/cache/Cache.cc#L291

JosiahWI: Sure enough. We should definitely deduplicate that! Good find.

JosiahWI: Changes made in #11895 and #11896.

JosiahWI:
[approve ci autest 2 3]

JosiahWI:
[approve ci autest 3]

masaori335 (approved):
Looks good. Thank you!

PR #11742, opened 2024-08-27
Cleanup: Move Stripe initialization to Stripe::_init_data

masaori335:
Two cosmetic changes of `Stripe` initialization.

~1. I noticed somehow `Stripe::_init_data_internal()` is called three times, but calling once seems enough.~

~The function was renamed and moved, but this comes from the initial commit.~
https://github.com/apache/trafficserver/blob/a1651345d11f228dfb392cc79169760e80d25f06/iocore/cache/Cache.cc#L981-L983

2. `StripeSM::init` has initialization of `StripeSM` and `Stripe`. This clarifies `Stripe::_init_data` initializes `Stripe` data. ( should be move to constructor later) 

----

Update: 1. is dropped from this PR as @JosiahWI pointed out.

### DIFF
--- a/src/iocore/cache/StripeSM.cc
+++ b/src/iocore/cache/StripeSM.cc
@@ -48,7 +48,6 @@
 
 #include "tscore/InkErrno.h"
 #include "tscore/Diags.h"
-#include "tscore/hugepages.h"
 #include "tscore/ink_assert.h"
 #include "tscore/ink_hrtime.h"
 #include "tscore/List.h"
@@ -154,6 +153,7 @@ StripeSM::clear_dir()
 int
 StripeSM::init(char *s, off_t blocks, off_t dir_skip, bool clear)
 {
+  // Hash
   char        *seed_str       = disk->hash_base_string ? disk->hash_base_string : s;
   const size_t hash_seed_size = strlen(seed_str);
   const size_t hash_text_size = hash_seed_size + 32;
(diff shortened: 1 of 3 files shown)


JosiahWI:
I do not think we can call `_init_data_internal()` only once. It's an unrolled loop used to approximate or converge towards an optimal ratio of directory size to buffer size. If we only call it once, the ratios we compute will be different. That is a behavioral change and needs careful consideration. I don't think we should do it in a cleanup.

masaori335:
Ah, the start pointer is referred and changed every call. Seems like we should keep it. At least, I need to drop it from this PR. Thanks!

masaori335:
Updated. This PR has moving Stripe member initialization only. PTAL.

src/iocore/cache/StripeSM.cc
```diff
@@ -164,19 +164,15 @@ StripeSM::init(char *s, off_t blocks, off_t dir_skip, bool clear)
            static_cast<uint64_t>(dir_skip), static_cast<uint64_t>(blocks));
   CryptoContext().hash_immediate(hash_id, hash_text, strlen(hash_text));
 
-  dir_skip = ROUND_TO_STORE_BLOCK((dir_skip < START_POS ? START_POS : dir_skip));
```
JosiahWI, line 167: Where did this go? Is there a reason it can be removed?

masaori335: It's here. This method doesn't use modified `dir_skip`, right?
https://github.com/apache/trafficserver/pull/11742/files#diff-88d9e07d3781ce4d9c6307114be7856582eba25edc0d1c64ac9a73a59d00baecR285

masaori335:
[approve ci autest]

src/iocore/cache/StripeSM.cc
JosiahWI: I see now, the temporary assignment was removed and `skip` is calculated directly. Nice improvement.

src/iocore/cache/StripeSM.cc
```diff
@@ -164,19 +164,15 @@ StripeSM::init(char *s, off_t blocks, off_t dir_skip, bool clear)
...
-  path     = ats_strdup(s);
-  len      = blocks * STORE_BLOCK_SIZE;
-  ink_assert(len <= MAX_STRIPE_SIZE);
-  skip             = dir_skip;
-  prev_recover_pos = 0;
```
JosiahWI, line 172: This is already initialized to 0 in the header, so no need to repeat ourselves here, good catch.

JosiahWI (approved):
Thanks, looks good!

PR #11745, opened 2024-08-27
Cleanup: Remove server name stuff from NetVConnection

maskit:
Server name is `nullptr` unless it's indicated by TLS SNI extension. Probably not a something we want to have in `NetVConnection`.


maskit:
http2_flow_control and txn_box_multi-ramp-1-3 failed.

maskit:
[approve ci autest]

maskit:
Same tests failed.

maskit:
Tests previously failed were fixed but now remap_acl fails.

maskit:
[approve ci autest]

src/api/InkAPI.cc
```diff
@@ -7928,17 +7928,17 @@ TSVConnFdGet(TSVConn vconnp)
 const char *
 TSVConnSslSniGet(TSVConn sslp, int *length)
 {
-  char const     *server_name = nullptr;
-  NetVConnection *vc          = reinterpret_cast<NetVConnection *>(sslp);
-
-  if (vc == nullptr) {
+  if (sslp == nullptr) {
     return nullptr;
```
bryancall, line 7932: I think length should also be set to 0 if we are returning a nullptr.

bryancall (approved):

src/api/InkAPI.cc
maskit: I'll make that change separately with documentation update.

PR #11748, opened 2024-08-28
Fix performance issues in cripts::Random

zwoop:
Seems the STL random functions are not reentrant, ATS core fixes this by having thread local engines, so Cripts can just piggy back on the existing code.

Co-Author: Nathan Wang

### DIFF
--- a/include/cripts/Lulu.hpp
+++ b/include/cripts/Lulu.hpp
@@ -17,7 +17,6 @@
 */
 #pragma once
 
-#include <random>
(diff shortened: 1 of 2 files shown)


shukitchan (approved):

cmcfarlen (approved):

zwoop:
This is a strong candidate for v10.0.1, it has some serious performance implications.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11751, opened 2024-08-29
Cleanup of redundant HdrToken & WKS

zwoop:


masaori335 (commented):
This looks reasonable 👍 

Probably, `_hdrtoken_strs` and `_hdrtoken_commonly_tokenized_strs` had different meanings in the past, but current code doesn't have it anymore.

JosiahWI (approved):
Agree with Masaori. Looks good, thanks!

zwoop:
> This looks reasonable 👍
> 
> Probably, `_hdrtoken_strs` and `_hdrtoken_commonly_tokenized_strs` had different meanings in the past, but current code doesn't have it anymore.

Yeh they did, _hdrtoken_commonly_tokenized_strs was smaller / different, but at some point (I can't find when / where) they were made identical. Maybe as part of us changing the hash map / searches ?

bryancall (approved):

PR #11752, opened 2024-08-30
Fix an infinite loop in Histogram

moonchen:
The loop in Histogram::operator() is infinite if R = 7, S = 2, and sample = 512.

### DIFF
--- a/include/tsutil/Histogram.h
+++ b/include/tsutil/Histogram.h
@@ -75,7 +75,7 @@ public:
   /// Samples less than this go in the underflow range.
   static constexpr raw_type UNDERFLOW_BOUND = static_cast<raw_type>(1) << N_SPAN_BITS;
   /// Sample equal or greater than this  go in the overflow bucket.
-  static constexpr raw_type OVERFLOW_BOUND = static_cast<raw_type>(1) << (N_RANGE_BITS + N_SPAN_BITS + 1);
+  static constexpr raw_type OVERFLOW_BOUND = static_cast<raw_type>(1) << (N_RANGE_BITS + N_SPAN_BITS);
 
   /** Add @sample to the histogram.
    *
--- a/src/tscore/unit_tests/test_Histogram.cc
+++ b/src/tscore/unit_tests/test_Histogram.cc
@@ -45,7 +45,7 @@ TEST_CASE("Histogram Basic", "[libts][histogram]")
   REQUIRE(h.min_for_bucket(16) == 32);
   REQUIRE(h.min_for_bucket(17) == 40);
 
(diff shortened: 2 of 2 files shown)


bneradt (approved):

moonchen:
The infinite loop happens as follows:
```
R = 7
S = 2
N_SPAN_BITS = S = 2
N_RANGE_BITS = R = 7
UNDERFLOW_BOUND = 1 << N_SPAN_BITS = 1 << 2 = 4
OVERFLOW_BOUND = 1 << (N_RANGE_BITS + N_SPAN_BITS + 1) = 1 << (7 + 2 + 1) = 2^10 = 1024
MSB_MASK = 1 << (N_RANGE_BITS + N_SPAN_BITS - 1) = 1 << (9 - 1) = 2^8 = 256
```
If the sample is 512, `sample & mask` will equal 0 for every mask, so the loop does not terminate.

bneradt:
I see. So the key code are these constexpr:

```cpp
  static constexpr raw_type MSB_MASK = static_cast<raw_type>(1) << (N_RANGE_BITS + N_SPAN_BITS - 1);
  static constexpr raw_type OVERFLOW_BOUND = static_cast<raw_type>(1) << (N_RANGE_BITS + N_SPAN_BITS + 1);
```

Where `N_RANGE_BITS` and `N_SPAN_BITS` correspond to R (7 in your description) and S (2 in your description). Thus, before your patch:

* `MSB_MASK` is 256
* `OVERFLOW_BOUND` is 1024

Then later these constants are used here:

```cpp
  } else if (sample < OVERFLOW_BOUND) { // not overflow bucket.
    idx       -= N_SPAN_BUCKETS;        // bottom bucket in the range.
    auto mask  = MSB_MASK;              // Mask to probe for bit set.
    // Shift needed after finding the MSB to put the span bits in the LSBs.
    unsigned normalize_shift_count = N_RANGE_BITS - 1;
    // Walk the mask bit down until the MSB is found. Each span bumps down the bucket index
    // and the shift for the span bits. An MSB will be found because @a sample >= @c UNDERFLOW_BOUND
    // The MSB is not before @c MSB_MASK because @a sample < @c UPPER_BOUND
    while (0 == (sample & mask)) {
      mask >>= 1;
      --normalize_shift_count;
      idx -= N_SPAN_BUCKETS;
    }
```

Note that inner while loop is trying to find the `mask` value which matches sample. But if `mask` starts at `MSB_MASK`, which is 256, and `sample` is 512, then no amount of right shifting will ever find that match. By setting `OVERFLOW_BOUND` to be 512 instead of 1024 per your patch, you are making it so we never run that loop because `sample` of 512 will not be less than `OVERFLOW_BOUND` of 512.

Makes sense.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11759, opened 2024-09-09
http2: use exponential backoff for restarting streams when going past watermark

cmcfarlen:
In production we were seeing a huge increase in the number of events processed and tracked it down to this
`HTTP2_SESSION_EVENT_DATA` event.  This PR adds an exponential backoff to the retries to try to find a retry period that matches the read rate of the client.  The adjusted backoff will remain for the life of the connection.


bneradt (approved):

bneradt:
Thanks. This looks like a really good update.

cmcfarlen:
Cherry-picked to v10.0.x

PR #11761, opened 2024-09-09
Event loop sub-stats

sharkleash:
Introduces additional tracking of the 3 major subsections of an event loop: draining the event queue, polling for io events, and processing the ready list once populated. Each referred to as drain_queue, io_wait, and io_work.

I've been tracking these stats on top of the base event loop time stat to get a better idea of the cause in the event of increasing loop times. Increasing time to drain the event queue, for instance, would have a different cause and potentially different action, compared with if there was an increase in the time spent processing io.

New Stats:
        `LOOP_DRAIN_QUEUE_MAX, ///< Max time Draining the event queue`
        `LOOP_IO_WAIT_MAX,     ///< Min time spent IO waiting`
        `LOOP_IO_WORK_MAX,     ///< Max time spent IO processing`

### DIFF
--- a/doc/admin-guide/monitoring/statistics/core/eventloop.en.rst
+++ b/doc/admin-guide/monitoring/statistics/core/eventloop.en.rst
@@ -80,6 +80,21 @@ high then it is likely transactions are experiencing significant latency.
 
     The maximum amount of time spent in a single loop in the last 10 seconds.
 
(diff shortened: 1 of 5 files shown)


duke8253 (changes requested):
Looks good. Could you add some documentation for them? 

https://github.com/apache/trafficserver/blob/master/doc/admin-guide/monitoring/statistics/core/eventloop.en.rst

duke8253 (approved):

PR #11762, opened 2024-09-11
ja3_fingerprint: add --preserve option

bneradt:
Add the --preserve option to ja3_fingerprint such that any ja* fields already existing are not appended with new values. This is useful in situations where there are multiple proxies adding ja headers and the one of greatest interest is the outermost edge proxy (presumably, the one closest to the client).

### DIFF
--- a/plugins/ja3_fingerprint/README
+++ b/plugins/ja3_fingerprint/README
@@ -27,6 +27,11 @@ Add flag --modify-incoming if it is desired that the plugin modify the incoming
     plugin) not as a remap plugin.  This is because remap plugins by definition
     are enaged upon remap completion and by that point it is too late to
     meaningfully modify the client request headers.
+Add flag --preserve for situations where there may be multiple proxies
+    handling a request and the first one, the one closest to the client, is
+    the one whose ja value is the most useful. If --preserve is used, then ja
+    values are not appended to already existing values in the request. By
+    default, ja header field values are appened if one exists already.
 
 3. plugin.config
 In plugin.config, supply name of the plugin and any desired options. For example:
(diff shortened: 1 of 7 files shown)


plugins/ja3_fingerprint/README
```diff
@@ -27,6 +27,11 @@ Add flag --modify-incoming if it is desired that the plugin modify the incoming
     plugin) not as a remap plugin.  This is because remap plugins by definition
     are enaged upon remap completion and by that point it is too late to
     meaningfully modify the client request headers.
+Add flage --preserve for situations where there may be multiple proxies
```
JosiahWI, line 30: flage -> flag

tests/gold_tests/pluginTest/ja3_fingerprint/ja3_fingerprint.test.py
```diff
@@ -134,7 +141,11 @@ def _configure_client(self, tr: 'TestRun') -> None:
         """
         name = f'client{self._client_counter}'
         p = tr.AddVerifierClientProcess(
-            name, self._replay_file, http_ports=[self._ts.Variables.port], https_ports=[self._ts.Variables.ssl_port])
+            name,
+            self._replay_file,
+            http_ports=[self._ts.Variables.port],
+            https_ports=[self._ts.Variables.ssl_port],
+            other_args='--thread-limit 1')
```
JosiahWI, line 148: Why is `--thread-limit 1` needed?

JosiahWI (dismissed):
Looks good. Adding a `bool` flag to the end of an already long parameter list makes me sad, but it's not part of an API and not convenient to change.

tests/gold_tests/pluginTest/ja3_fingerprint/ja3_fingerprint.test.py
bneradt: I think this will address the log ordering issues we've seen in this test. It makes ATS serialize across the sessions, and this test has an HTTP/1 and HTTP/2 session. Currently there's a race condition as to which finishes first.

plugins/ja3_fingerprint/README
bneradt: Thanks for the catch.

tests/gold_tests/pluginTest/ja3_fingerprint/ja3_fingerprint.test.py
bneradt: I broke this change out as a separate PR so that it can get backported to 10.0.x.

https://github.com/apache/trafficserver/pull/11766

ezelkow1 (approved):

PR #11763, opened 2024-09-11
JSONRPC - Use dynamic and configurable buffer for incoming jsonrpc messages.

brbzull0:
With the old implementation we only allowed up to `32k` messages, this could be an issue once we start accepting full configuration files through the rpc, like injecting full `records.yaml` or even multiple files in a single message.
As there is no way to know the message size beforehand, this change lets the user configure the max size for a single request (`jsonrpc.yaml`).

**The change:**

For small messages the rpc server will preallocate 32k memory(same as before, this does not change) and this will only increase in case the message is bigger than the first allocated chunk, in  which case we will add up another 32k up  and so on until we reach the configured limit. To sum up the full configured limit will not be allocated from the start, it will be only if needed.

`MessageStorage.h` was moved from a local scope so it can be shared between client and server.

Please note that even though the rpc client uses the same buffer logic, there is no change in it.

Every autest that uses `traffic_ctl` exercises this change, so this is the main reason there is no new autest here. Also https://github.com/apache/trafficserver/pull/11740 is also meant to exercise this change.

### DIFF
--- a/include/shared/rpc/MessageStorage.h
+++ b/include/shared/rpc/MessageStorage.h
@@ -1,0 +1,104 @@
+/**
+  @section license License
+
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+#pragma once
+
+#include <sstream>
+#include <swoc/BufferWriter.h>
+
+/// @brief Simple storage to keep the jsonrpc server's response.
+///
+///        With small content it will just use the LocalBufferWriter, if the
+///        content gets bigger, then it will just save the buffer into a string
+///        and reuse the already created LocalBufferWriter. If stored data fits in the
+///        original bw, then no need to create the extra string. If message fist in the
+///        first chunk, no extra space will be allocated into the storage string.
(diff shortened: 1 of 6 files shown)


include/shared/rpc/MessageStorage.h
```diff
@@ -0,0 +1,104 @@
...
+      _content.reserve(_written + _bw.size());
+    }
+
+    _content.append(_bw.data(), _bw.size());
+    _written += _bw.size();
+
+    _bw.clear();
+  }
+
+  std::string
```
cmcfarlen, line 86: This can be a string_view correct? The storage is always backed by either _bw or _content

src/mgmt/rpc/server/IPCSocketServer.cc
```diff
@@ -219,11 +217,12 @@ IPCSocketServer::run()
...
     std::error_code ec;
     if (int fd = this->accept(ec); !ec) {
-      Client client{fd};
+      Client client{fd, _conf.incomingRequestMaxBufferSize};
+      Buffer bw;
 
       if (auto [ok, errStr] = client.read_all(bw); ok) {
-        const auto   json = std::string{bw.data(), bw.size()};
-        rpc::Context ctx;
+        const std::string &json = bw.str();
```
cmcfarlen, line 224: I think this could create a temporary and a copied string.  I suggest having str() return string_view and storing it by value here.

cmcfarlen (changes requested):

src/mgmt/rpc/server/IPCSocketServer.cc
brbzull0: I did the `const &` intentionally, to extend(from the `str()` return by value)  the lifetime of the temporary string returned by `str()` so we just 1  copy which will be stored into `json`.

I can do either way, no strong feeling on any one.

cmcfarlen: I understand extending the lifetime, but a copy seems unnecessary since the buffer is already managed by the Buffer class.   I think Buffer::str could just call flush() and then return a const ref to the internal string and you'd get the same allocations for the small case, but save a copy when reading > N bytes.  Its not a huge deal based on how often this happens.

brbzull0: no it's fine, it should be just 1 ideally, that was the point at least.

PR #11764, opened 2024-09-11
hostdb: remove latency with cache misses

cmcfarlen:
If hostdb suffers a cache miss, it would schedule a continuation to process the dns lookup.  Unfortunately, at some point in time, this event was scheduled with a delay instead of immediately.  This PR restores that and saves 20ms on a dns lookup.

Also, if the probe request could be fulfilled by parsing an IP or the hostfile, this was also delayed by the same 20ms.  This PR also moves those two checks to probe-time rather than dns lookup time (neither require IO).

coauthored by: @moonchen

### DIFF
--- a/src/iocore/hostdb/HostDB.cc
+++ b/src/iocore/hostdb/HostDB.cc
@@ -490,6 +490,38 @@ db_mark_for(IpAddr const &ip)
 }
 
 HostDBRecord::Handle
+probe_ip(HostDBHash const &hash)
+{
+  HostDBRecord::Handle result;
+
+  if (hash.is_byname()) {
+    Dbg(dbg_ctl_hostdb, "DNS %.*s", int(hash.host_name.size()), hash.host_name.data());
+    IpAddr tip;
+    if (0 == tip.load(hash.host_name)) {
+      result            = HostDBRecord::Handle{HostDBRecord::alloc(hash.host_name, 1)};
+      result->af_family = tip.family();
+      auto &info        = result->rr_info()[0];
+      info.assign(tip);
+    }
+  }
+
+  return result;
+}
+
+HostDBRecord::Handle
+probe_hostfile(HostDBHash const &hash)
+{
+  HostDBRecord::Handle result;
+  // Check if this can be fulfilled by the host file
+  //
+  if (auto static_hosts = hostDB.acquire_host_file(); static_hosts) {
+    result = static_hosts->lookup(hash);
+  }
+
+  return result;
+}
+
+HostDBRecord::Handle
 probe(HostDBHash const &hash, bool ignore_timeout)
 {
   static const Ptr<HostDBRecord> NO_RECORD;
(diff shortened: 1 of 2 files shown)


bneradt:
[approve ci autest]

bneradt:
bad_chunked_encoding test failed. rescheduling the autests.

src/iocore/hostdb/HostDB.cc
```diff
@@ -784,9 +826,9 @@ HostDBContinuation::lookup_done(TextView query_name, ts_seconds answer_ttl, SRVH
 {
   ink_assert(record);
   if (query_name.empty()) {
-    if (is_byname()) {
+    if (hash.is_byname()) {
       Dbg(dbg_ctl_hostdb, "lookup_done() failed for '%.*s'", int(hash.host_name.size()), hash.host_name.data());
-    } else if (is_srv()) {
+    } else if (hash.is_srv()) {
```
bneradt, line 831: Can you explain why these various functions are now all called via `hash`?

cmcfarlen: I moved these utility functions to `HostDBHash` so I could use them without a `HostDBContinuation`

cmcfarlen:
[approve ci autest]

cmcfarlen:
slice_prefetch failed.

src/iocore/hostdb/HostDB.cc
bneradt: Oh, I see. I saw in your patch that functions got moved, but I didn't notice it was these and that they were moved to a new class. Thank you for the explanation.

cmcfarlen:
I did some benchmarking before/after this change:

The test setup is a list of hosts with about 300 different hostnames and then 200 lines of `127.0.0.1`
The benchmark will randomize this list for each thread in the test and then perform a HostDB lookup for each line. Ideally, hostdb will perform as few actual DNS lookups as possible and realize the IP address can just be use directly.

Before this PR:
```
Hosts: 503 lookup count: 14282 thread count: 64
dns min/max: 0.016236/0.028204
imm min/max: 0.000000/0.000031
Total results: 32192 average lookup 0.008421

real    0m4.678s
user    0m0.388s
sys     0m0.471s
```

After this PR:
```
Hosts: 503 lookup count: 464 thread count: 64
dns min/max: 0.000011/0.012030
imm min/max: 0.000000/0.000433
Total results: 32192 average lookup 0.000065

real    0m0.408s
user    0m0.342s
sys     0m0.220s
``` 

Analysis:

Comparing the DNS min/max time, you can see the drop of about 20ms that comes directly from the unfortunate schedule delay.  You can also see from the lookup count that HostDB could immediately return way more results after this PR.  This is due to the fact that lookups to IP addresses are not considered misses  now (and subsequently aren't subject to the unfortunate delay).
Finally note that the average lookup time and total runtime are an order of magnitude faster.

This host list was picked to show a scenario where less than half of lookups take the IP parse path which will be a fairly bad case for the pre-PR HostDB code.  A list with all resolvable hostnames would not have shown this drastic of a difference.

bneradt (commented):
I installed this on a box in Yahoo prod and it seems to run well. I can't say that I noticed a change in overall latency, but maybe our particular pattern doesn't directly tickle this enough for me to notice in broad stroke numbers. I didn't see a performance degradation at any rate.

Thanks @moonchen and @cmcfarlen .

bneradt (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11766, opened 2024-09-12
ja3_fingerprint.test.py: address race condition

bneradt:
ja3_fingerprint.test.py uses the Proxy Verifier client and configures it with two sessions/connections. By default, Proxy Verifier will run the transactions for those two connections in parallel. Generally the first HTTP/1 session will finish before the second HTTP/2 session, but sometimes the second session finished first. That causes the test to fail because the gold file expects the other order. This patch addresses this race condition issue by simply configuring the client to run the sessions in serial rather than in parallel.


bneradt:
Marking for 10.0.x since this is a test-only change which should make the test more reliable.

zwoop (approved):

PR #11770, opened 2024-09-13
Only include TestBox.h if needed

moonchen:
This silences some linter warnings about unused includes when regression tests are disabled.


bneradt:
[approve ci autest]

PR #11771, opened 2024-09-13
Fix libswoc compile with clang 20

moonchen:
- Fix incorrect method name has_intersection (should be has_intersection_with).
- template keyword on a method call causes a warning if template arguments are not supplied.


bneradt (approved):
Thanks for the libswoc fixes.

PR #11772, opened 2024-09-13
multiplexer: support request bodies from non-POST/PUT

bneradt:
This updates the multiplexer plugin to support request bodies from non-POST and PUT methods. As long as they are Content-Length requests.

### DIFF
--- a/doc/admin-guide/plugins/multiplexer.en.rst
+++ b/doc/admin-guide/plugins/multiplexer.en.rst
@@ -24,8 +24,9 @@
 
 |Name| is a remap plug-in that allows requests to certain origins to be multiplexed (i.e.,
 duplicated and dispatched in parallel) to one or more other hosts.  The headers are copied into the
-new requests as well as POST bodies.  Optionally POST and PUT requests can be skipped via
-``pparam=proxy.config.multiplexer.skip_post_put=1``.
+new requests as well as request bodies specified via Content-Length. Chunk encoded request bodies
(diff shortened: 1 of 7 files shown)


doc/admin-guide/plugins/multiplexer.en.rst
```diff
@@ -90,54 +91,3 @@ Here are some example :file:`remap.config` configuration lines::
 #. The fourth entry will multiplex requests sent to ``http://www.example.com`` with a path of ``/d``
    to both ``host1.example.com`` and ``host2.example.com``, but POST and PUT requests will
    not be multiplexed.
-
-Implementation
-==============
-
-Parsing Chunk Encoded Data
---------------------------
```
bneradt, line 98: I'm removing this because I don't think this will be helpful to end users and may confuse them into thinking that we support multiplexing chunk encoded requests. We do parse and send them onto the origin, but we do not support multiplexing them.

plugins/multiplexer/ats-multiplexer.cc
```diff
@@ -158,7 +158,7 @@ DoRemap(const Instance &i, TSHttpTxn t)
     generateRequests(i.origins, buffer, location, requests);
     assert(requests.size() == i.origins.size());
 
-    if (is_post_or_put) {
+    if (is_post_or_put || content_length > 0) {
```
bneradt, line 161: Note that this is the only production change to support other methods that have Content-Length bodies.

randall (approved):

PR #11774, opened 2024-09-17
quic: Enable quiche debug logging only if debug tag vv_quiche is set

maskit:
Although current code prevents printing debug log if the debug tag, `vv_quiche`, is not set (or ATS debug log is disabled), quiche generates debug log internally anyway if `quiche_enable_debug_logging` is called.

ATS has to be restarted to enable quiche debug logging, but you'd never want to enable it where restarting is going to be a problem.


brbzull0 (approved):
Nice.
Thanks.

PR #11775, opened 2024-09-17
http3: Remove an unnecessary copy

maskit:
This uses more IOBufferBlocks, but it wouldn't be problematic unlike H2 because the size of Data frame is much larger than H2.


brbzull0 (approved):
LGTM.
Thanks.

PR #11778, opened 2024-09-18
Reduce the number of calls for ERR_clear_error

maskit:
Since we check the error only after the loop, clearing error on each iteration does not make sense. Unlike the `errno`, the error variable in Open/BoringSSL is actually a queue so clearing it is not a trivial operation (`memset` and `free` are called inside).

I feel like we should check whether there was an error on each iteration and break out if necessary, but that's another issue. This is for (tiny) performance improvement.


brbzull0 (approved):
make sense. Thanks.

PR #11785, opened 2024-09-24
header_rewrite: Fix condition parser

### LINKED ISSUE #11776: header_rewrite: IP range matcher crashes
For `INBOUND` and `IP` condition, the special matcher to parse IP ranges causes ATS to crash if not configured properly. Instead of crashing, ATS should fail to load the bad config. Docs should also be updated to not include whitespaces between brackets

Reproduce:
Add line to header_rewrite config
`cond %{INBOUND:REMOTE-ADDR} { 192.168.201.0/24,10.0.0.0/8 }`

maskit:
This fixes #11776.

This change allows the spaces before/after the braces. It also prevents crashes even if there are syntax errors.

### DIFF
--- a/plugins/header_rewrite/condition.cc
+++ b/plugins/header_rewrite/condition.cc
@@ -43,13 +43,23 @@ parse_matcher_op(std::string &arg)
     break;
   case '/':
     arg.erase(0, 1);
-    arg.erase(arg.length() - 1, arg.length());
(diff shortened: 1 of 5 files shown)


duke8253 (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11789, opened 2024-09-24
INKVConnInternal::do_io_*: handle null buffer

bneradt:
It's common for users of VC's to cancel io via a 0 byte, nullptr read or write on the VC. INKVConnInternal::do_io_read and
INKVConnInternal::do_io_write were not prepared to handle such cancellations. This updates them to handle this gracefully rather than crashing on a nullptr dereference. This change was found to be needed for the multiplexer plugin for handling HttpTunnel aborts.

For reference, see, for example, UnixNetVConnection::do_io_read which handles a nullptr MIOBuffer. This basically copies that logic into INKVConnInternal so it handles cancellation gracefully.

---
# For Review

Hiding whitespace when viewing the diff makes this patch more understandable.

### DIFF
--- a/src/api/InkVConnInternal.cc
+++ b/src/api/InkVConnInternal.cc
@@ -70,17 +70,21 @@ INKVConnInternal::destroy()
 VIO *
 INKVConnInternal::do_io_read(Continuation *c, int64_t nbytes, MIOBuffer *buf)
 {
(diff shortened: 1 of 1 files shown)


bneradt:
This didn't backport cleanly to 10.0.x, so I created a separate PR for it:
https://github.com/apache/trafficserver/pull/11802

PR #11791, opened 2024-09-25
Move HttpAPIHooks from proxy to tsapibackend

cmcfarlen:
This removes the circular dependency between cache and proxy libs.


src/iocore/cache/CMakeLists.txt
```diff
@@ -48,8 +48,8 @@ target_include_directories(inkcache PRIVATE ${CMAKE_SOURCE_DIR}/lib)
 
 target_link_libraries(
   inkcache
-  PUBLIC ts::aio ts::hdrs ts::inkevent ts::tscore
-  PRIVATE ts::proxy fastlz ZLIB::ZLIB
+  PUBLIC ts::aio ts::hdrs ts::inkevent ts::tscore ts::tsapibackend
+  PRIVATE fastlz ZLIB::ZLIB
```
JosiahWI, line 52: $ grep -Ir "#include \"api" src/iocore/cache/
src/iocore/cache/HttpTransactCache.cc:#include "api/APIHook.h"
src/iocore/cache/HttpTransactCache.cc:#include "api/InkAPIInternal.h"
src/iocore/cache/unit_tests/stub.cc:#include "api/InkAPIInternal.h"

Is the change to PUBLIC intentional? It's surprising.

JosiahWI (changes requested):

JosiahWI (approved):
It even has `API` in the name, so it seems appropriate to put it in tsapibackend, and this improves the dependency situation. Thank you!

PR #11792, opened 2024-09-25
Cleanup aio inline and remove P_AIO header

cmcfarlen:
This cleans up aio Inline.cc and removes the private header use in other libraries.  It also combines the AIOCallback and AIOCallbackInternal.  Since the latter was always used externally, it didn't make sense to have the separation.

There was a bit of additional fallout because P_AIO.h had a `using ts::Metrics` that spread to other source files.


cmcfarlen:
[approve ci autest]

bneradt (approved):

PR #11795, opened 2024-09-27
add cache_open_write_fail_count metric

traeak:
This is so we can get an idea of cache write lock contention.

The counter could also be bumped in `HttpSM::state_cache_open_write` in the CACHE_EVENT_OPEN_WRITE_FAILED case.


traeak:
[approve ci debian]

bryancall (approved):

PR #11801, opened 2024-10-01
Add Dockerfile in contrib

cmcfarlen:
This could be used to provide a traffic server docker image on hub.docker.com.

From the `contrib/docker/ubuntu/noble` directory:
```
$ docker build -t ats .
# make remap.config
$ cat remap.config
map http://ts:8080/foo http://httpbin.org/uuid
$ docker run -v $PWD/remap.config:/opt/etc/trafficserver/remap.config -d --name ts --rm ats
```

```
# image comes with h3 enabled curl
$ docker run -it --rm --link ts ats curl http://ts:8080/foo
{
  "uuid": "a6d4458f-4243-4d6f-87cf-878cf016e8f5"
}
```

```
docker logs ts
```

### DIFF
--- a/contrib/docker/ubuntu/noble/Dockerfile
+++ b/contrib/docker/ubuntu/noble/Dockerfile
@@ -236,4 +236,4 @@ ENV PATH="$PATH:/opt/bin"
 
 EXPOSE 8080
 
-CMD ["traffic_server", "-T", "http"]
+CMD ["traffic_server"]
--- a/contrib/docker/ubuntu/noble/Dockerfile
+++ b/contrib/docker/ubuntu/noble/Dockerfile
@@ -1,3 +1,19 @@
+#######################
+#
+#  Licensed to the Apache Software Foundation (ASF) under one or more contributor license
+#  agreements.  See the NOTICE file distributed with this work for additional information regarding
+#  copyright ownership.  The ASF licenses this file to you under the Apache License, Version 2.0
+#  (the "License"); you may not use this file except in compliance with the License.  You may obtain
+#  a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software distributed under the License
+#  is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+#  or implied. See the License for the specific language governing permissions and limitations under
+#  the License.
+#
+#######################
 FROM ubuntu:noble AS build
 
 ARG LLVM_VERSION=18
--- a/contrib/docker/ubuntu/noble/Dockerfile
+++ b/contrib/docker/ubuntu/noble/Dockerfile
@@ -1,0 +1,218 @@
+FROM ubuntu:noble AS build
+
+ARG LLVM_VERSION=18
+ARG BASE=/opt
+ARG ATS_VERSION=10.0.0
+
+RUN apt update \
+ && apt upgrade --yes \
+ && apt install --no-install-recommends --yes \
+    ca-certificates \
+    clang-${LLVM_VERSION} \
+    libc++-${LLVM_VERSION}-dev \
+    cmake \
+    ninja-build \
+    libc-ares-dev \
+    libsystemd-dev \
+    libev-dev \
+    libevent-dev \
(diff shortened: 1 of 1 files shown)


shukitchan:
Do we intend to use this to generate docker images to be stored with GitHub through a GitHub action?

contrib/docker/ubuntu/noble/Dockerfile
```diff
@@ -0,0 +1,234 @@
...
+RUN if [ `uname -m` = "arm64" -o `uname -m` = "aarch64" ]; then echo "arm64" > /arch; else echo "amd64" > /arch; fi \
+  && wget -qO- https://go.dev/dl/go1.21.6.linux-$(cat /arch).tar.gz | tar -C ${BASE} -xzf -
+
+ENV CC=clang-${LLVM_VERSION}
+ENV CXX=clang++-${LLVM_VERSION}
+ENV GO_BINARY_PATH=${BASE}/go/bin/go
+
+RUN git clone https://boringssl.googlesource.com/boringssl \
+ && cd boringssl \
+ && git checkout 45b2464158379f48cec6e35a1ef503ddea1511a6 \
```
bneradt, line 73: For these libraries, can we copy down the build_h3_tools.sh script and run that instead? It would be nice to avoid two implementations of the invocations of these builds if we can.

bneradt (changes requested):
Looks good.

contrib/docker/ubuntu/noble/Dockerfile
```diff
@@ -0,0 +1,234 @@
...
+#  the License.
+#
+#######################
+FROM ubuntu:noble AS build
+
+ARG LLVM_VERSION=18
+ARG BASE=/opt
+ARG ATS_VERSION=10.0.0
+
+RUN apt update \
```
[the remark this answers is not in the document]
cmcfarlen: Are heredocs well supported now? Last time I tried them in a Dockerfile it didn't go well depending on the docker setup

contrib/docker/ubuntu/noble/Dockerfile
cmcfarlen: I don't think it would work as-is, but maybe we could do that?

contrib/docker/ubuntu/noble/Dockerfile
```diff
@@ -0,0 +1,239 @@
...
+    libjansson4 \
+ && apt clean --yes
+
+COPY --from=build /opt /opt
+RUN chown nobody /opt/var/trafficserver /opt/var/log/trafficserver
+ENV PATH="$PATH:/opt/bin"
+
+EXPOSE 8080
+
+CMD ["traffic_server", "-T", "http"]
```
bryancall, line 239: We shouldn't run traffic server in debug mode.

bneradt:
[approve ci clang-analyzer]

contrib/docker/ubuntu/noble/Dockerfile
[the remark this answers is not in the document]
bneradt: I've never had trouble myself. I actually filed an issue against the docker github last year suggesting they document the feature more:
https://github.com/docker/docs/issues/17559

They didn't indicate that it isn't supported but rather pointed to this as evidence that it is documented (albeit in a link potentially not obvious to get to):
https://docs.docker.com/reference/dockerfile/#here-documents

contrib/docker/ubuntu/noble/Dockerfile
bneradt: It would be nice to update the script to work in Ubuntu if it's not too much trouble. That would be a win for this Dockerfile and others trying to use the script in Ubuntu.

bneradt (approved):

contrib/docker/ubuntu/noble/Dockerfile
```diff
@@ -0,0 +1,239 @@
...
+    libevent-dev \
+    zlib1g-dev \
+    cargo \
+    wget \
+    git \
+    libtool \
+    make \
+    pkg-config \
+# ATS deps
+    libxml2-dev \
```
bneradt, line 43: This comment at 42 doesn't comment out the rest of the packages in the "line" does it? Or is Docker smart enough not to extend the comment?

PR #11805, opened 2024-10-03
Coverity. Fixes for CID-1497391 and CID-1497371

brbzull0:


cmcfarlen (approved):
The Coverity issues are complaining about missing move constructors.  If you declare a move constructor or assignment operator `= default` they will do a member wise move construct or assignment.  This should silence the warnings.

PR #11811, opened 2024-10-08
Add a constructor to `Stripe` to impose invariants early

JosiahWI:
The recent commits are not signed because gpg broke on WSL after an update.

This change also revealed an issue with the `Stripe` unit tests. The write head was pointing to a hardcoded value that happened to be valid instead of the start of the document blocks. This also fixes that issue - see c39be53.

### DIFF
--- a/src/iocore/cache/StripeSM.cc
+++ b/src/iocore/cache/StripeSM.cc
@@ -109,7 +109,7 @@ struct StripeInitInfo {
   }
 };
 
-StripeSM::StripeSM(off_t blocks, off_t dir_skip) : Continuation(new_ProxyMutex()), Stripe{blocks, dir_skip}
+StripeSM::StripeSM(CacheDisk *disk, off_t blocks, off_t dir_skip) : Continuation(new_ProxyMutex()), Stripe{disk, blocks, dir_skip}
 {
   this->_preserved_dirs.evacuate_size = static_cast<int>(len / EVACUATION_BUCKET_SIZE) + 2;
   int evac_len                        = this->_preserved_dirs.evacuate_size * sizeof(DLL<EvacuationBlock>);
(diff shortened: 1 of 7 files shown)


JosiahWI:
The AuTest post-early-return failed.

src/iocore/cache/StripeSM.cc
```diff
@@ -151,36 +162,11 @@ StripeSM::clear_dir()
...
-  // Hash
-  char        *seed_str       = disk->hash_base_string ? disk->hash_base_string : s;
-  const size_t hash_seed_size = strlen(seed_str);
-  const size_t hash_text_size = hash_seed_size + 32;
-
-  hash_text = static_cast<char *>(ats_malloc(hash_text_size));
-  ink_strlcpy(hash_text, seed_str, hash_text_size);
-  snprintf(hash_text + hash_seed_size, (hash_text_size - hash_seed_size), " %" PRIu64 ":%" PRIu64 "",
-           static_cast<uint64_t>(dir_skip), static_cast<uint64_t>(blocks));
-  CryptoContext().hash_immediate(hash_id, hash_text, strlen(hash_text));
```
masaori335, line 165: Where this code goes?

JosiahWI: It moved to `Stripe::_init_hash_text`.

JosiahWI: It happened in 0412927.

masaori335: do you mean `Stripe::_init_hash_text` function? I can see `snprintf` makes the `hash_text`, but I can't find using it for `hash_id`. What I'm missing?
https://github.com/apache/trafficserver/pull/11811/files#diff-88d9e07d3781ce4d9c6307114be7856582eba25edc0d1c64ac9a73a59d00baecR102-R113

masaori335 (approved):
Looks good! Thanks for cleaning up.

PR #11814, opened 2024-10-09
Fixes to HostDB marking IPs up/down

cmcfarlen:
We have been troubleshooting and issue where ATS10 was returning more 500s than a nearby ATS9 host.  This PR fixes several related HostDB issues:

 * Set the fail count to 0 when marking an IP up
 * Restore the fail count from an old entry when migrating due to updates from DNS
 * Combine incrementing the fail count and marking the IP down to avoid a race condition
 * Cleanup to encapsulate failure state from HostDB users (i.e. don't increment fail_count directly)
 * HostDBInfo::select is now const and does not update the fail_time.  This was causing IPs to never be marked up.
 * If a HostDB record only has one IP, always select that for use/retry regardless of fail status (This restores the ATS9 behavior)
 * Add some ATS_PROBE calls to aid in troubleshooting.

Co-authors: @moonchen @maskit

### DIFF
--- a/include/iocore/hostdb/HostDBProcessor.h
+++ b/include/iocore/hostdb/HostDBProcessor.h
@@ -151,7 +151,7 @@ struct HostDBInfo {
    * If a zombie is selected the failure time is updated to make it appear down to other threads in a thread safe
    * manner. The caller should check @c last_fail_time to see if a zombie was selected.
    */
-  bool select(ts_time now, ts_seconds fail_window);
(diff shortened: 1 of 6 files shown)


cmcfarlen:
[approve ci rocky]

cmcfarlen:
[approve ci rocky]

bryancall (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11815, opened 2024-10-10
jsonrpc: Wait for the socket to be ready to read.

brbzull0:
We found out that sometimes we hit a busy loop here, so adding a timeout to make sure we do not hangup forever here.

### DIFF
--- a/src/shared/rpc/IPCSocketClient.cc
+++ b/src/shared/rpc/IPCSocketClient.cc
@@ -35,7 +35,7 @@ namespace shared::rpc
 {
 
 IPCSocketClient::self_reference
-IPCSocketClient::connect(std::chrono::milliseconds ms, int attempts)
+IPCSocketClient::connect(std::chrono::milliseconds wait_ms, int attempts)
 {
   std::string text;
   int         err, tries{attempts};
@@ -67,7 +67,7 @@ IPCSocketClient::connect(std::chrono::milliseconds ms, int attempts)
     if (errno == EAGAIN || errno == EINPROGRESS) {
       // Connection cannot be completed immediately
       // EAGAIN for UDS should suffice, but just in case.
-      std::this_thread::sleep_for(ms);
+      std::this_thread::sleep_for(wait_ms);
       err = errno;
       continue;
     } else {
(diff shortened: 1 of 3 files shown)


bneradt (commented):
lgtm. I'll let a non-Yahoo reviewer chime in.

src/shared/rpc/IPCSocketClient.cc
```diff
@@ -30,12 +30,12 @@
 #include "shared/rpc/MessageStorage.h"
 #include <tscore/ink_assert.h>
 #include <tscore/ink_sock.h>
-
+#include <iostream>
```
cmcfarlen, line 33: This header is unused.

src/shared/rpc/IPCSocketClient.cc
```diff
@@ -117,23 +117,36 @@ IPCSocketClient ::send(std::string_view data)
 }
 
 IPCSocketClient::ReadStatus
-IPCSocketClient::read_all(std::string &content)
+IPCSocketClient::read_all(std::string &content, std::chrono::milliseconds timeout_ms, int attempts)
 {
   if (this->is_closed()) {
     // we had a failure.
     return {};
```
cmcfarlen, line 124: I know it's not part of the changes, but this will return `NO_ERROR`? But the comment indicates a failure.  Could `UNKNOWN` be a better return?

src/shared/rpc/IPCSocketClient.cc
```diff
@@ -117,23 +117,36 @@ IPCSocketClient ::send(std::string_view data)
...
+      if (ret < 0 && (errno == EAGAIN || errno == EINTR)) {
+        if (auto r = read_ready(_sock, timeout_ms.count()); r <= 0) {
+          readStatus = r == 0 ? ReadStatus::TIMEOUT : ReadStatus::READ_ERROR;
+          --attempts_left;
+        }
+      } else {
+        // done!
+        break;
+      }
+    } while (ret != 0 && attempts_left > 0);
```
cmcfarlen, line 145: Should this check for readStatus timeout and error and also stop the loop? If you don't check for timeout, this function will wait for timeout_ms * attempts (10 seconds by default).  What errors would poll return that we would want to retry?

cmcfarlen (changes requested):
I am not sure this is correct.  Should you call read_ready if there were some bytes read in previous iterations of the outer loop? Shouldn't a timeout from read_ready stop the inner loop?   It seems like this will always wait until EOF or until (timeout_ms * attempts) have elapsed. The logic is a bit confusing, so maybe some comments about what is happening will clear it up.

brbzull0:
> I am not sure this is correct. Should you call read_ready if there were some bytes read in previous iterations of the outer loop? Shouldn't a timeout from read_ready stop the inner loop? It seems like this will always wait until EOF or until (timeout_ms * attempts) have elapsed. The logic is a bit confusing, so maybe some comments about what is happening will clear it up.

Hi. It was weird indeed 🤮 

I changed it a bit now, single loop, check for the data in the socket, if available we read, if not we wait a few times unless an error occurs. 
Should be simpler now.

side note:
I am planning to clean up this in a future PR, some of the `ReadError` makes no more sense since we changed the buffer approach few months back.

src/shared/rpc/IPCSocketClient.cc
```diff
@@ -117,35 +117,52 @@ IPCSocketClient ::send(std::string_view data)
...
+    } else if (r < 0) {
+      // No more tries.
+      readStatus = ReadStatus::READ_ERROR;
+      break;
+    }
+
+    nread = ::read(_sock, buf, to_read);
+    if (nread > 0) {
+      bs.save(nread);
+      continue;
```
cmcfarlen, line 151: If this doesn't break here, and there are no more bytes available, it seems this loop will wait the max time before returning.   Shouldn't we go ahead and return these bytes and if the caller needs more it can call this again?

cmcfarlen (changes requested):
Please clarify if this loop should always wait the max time for any data vs returning after reading any.

src/shared/rpc/IPCSocketClient.cc
brbzull0: Sorry, didn't clarify this, yes It will wait till  we have the data, or exausted the max time/tries.
Returning some data won't work as it will be a badly formed message, so I rather wait and see if we get the whole message before giving up. I see no benefit on making it more complicated. 

very simple protocol, and no need to retry, if any error, just run the command again and that's it(traffic_ctl or traffic_top).

cmcfarlen: The max time default is 10 seconds.  Is the socket guaranteed to close or something?  How does `read_all` know the message is complete?

brbzull0: if you have a message for instance, to fetch records, or metrics and you have more than 6k metrics, then `10s` seems rasonable as it could take a while to send all the strings over the pipe.
 
I would expect `read` return `EOF` once done, if it's done then `read_all` knows there is no more data to read(ofc).  If there is any error different than EAGAIN then this will just stop and fail.

if you think tehre is  something that I am missing, please let me know. I'll be happy to address.

brbzull0:
> Please clarify if this loop should always wait the max time for any data vs returning after reading any.

The idea is to wait till EO, or an error(ofc), returning any data could be useless as the message will be incomplete(without trying again for more data), this should be simple, if there is any error, then run `traffic_ctl` again and that should be it. 
I can add some comments to make this clear in the code.

cmcfarlen (approved):
Thank you!

brbzull0:
> Thank you!

Thank you for taking the time and provide good feedback! appreciated!

PR #11816, opened 2024-10-11
records -  Honour the minimum value for `--maxRecords`

brbzull0:
Although the default and minimum value is `2048` (as advertised) ATS will accept any value less than the advertised default, which will end up in ATS crashing. This change sets the advertised default value if the passed value is invalid or less than the default.

Not sure if this may impact someone using different default values, or max values for records?

Update after review feedback:

Changed the warning to check against the configured `--maxRecords`(either the default or the one passed by the user). If there are more records to read than the allocated, then ATS will fatal as we cannot store all records from `records.yaml`

### DIFF
--- a/src/records/P_RecCore.cc
+++ b/src/records/P_RecCore.cc
@@ -188,6 +188,10 @@ RecSetRecord(RecT rec_type, const char *name, RecDataT data_type, RecData *data,
       goto Ldone;
     }
     r1 = RecAlloc(rec_type, name, data_type);
+    if (r1 == nullptr) {
+      err = REC_ERR_FAIL;
+      goto Ldone;
+    }
     RecDataSet(data_type, &(r1->data), data);
     if (REC_TYPE_IS_STAT(r1->rec_type) && (data_raw != nullptr)) {
       r1->stat_meta.data_raw = *data_raw;
--- a/src/records/P_RecDefs.h
+++ b/src/records/P_RecDefs.h
@@ -30,7 +30,7 @@
 // We need at least this many internal record entries for our configurations and metrics.
 // This may need adjustments if we make significant additions to librecords. Note that
 // plugins are using their own metrics systems.
-#define REC_MAX_RECORDS               2048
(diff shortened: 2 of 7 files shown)


cmcfarlen:
It's a little odd that you can change the number of records allocated at all currently.  I think you'll get issues if you set it > `REC_MAX_RECORDS`.   See this here:
https://github.com/apache/trafficserver/blob/master/src/records/RecUtils.cc#L54-L57

Maybe we should either fix that or remove (reremove?) the ability to change how many are allocated?

brbzull0:
> It's a little odd that you can change the number of records allocated at all currently. I think you'll get issues if you set it > `REC_MAX_RECORDS`. See this here: https://github.com/apache/trafficserver/blob/master/src/records/RecUtils.cc#L54-L57
> 
> Maybe we should either fix that or remove (reremove?) the ability to change how many are allocated?

Thanks Chris, I was looking for feedback here too.

Agreed it is odd, the whole point is to avoid crashing. I did some tests before putting this PR testing out after passing this max, I know some people just change the `REC_MAX_RECORDS` on their own build. The above check happens when IT'S actually reads the `records.yaml` and not by us setting this manually I believe.

I did some tests also with a large `records.yaml`(and also changing the `REC_MAX_RECORDS` to avoid the warning) and it was fine, which makes me think we can actually(as you suggested)  remove the hard max from the code and have it as default instead, with the possibility ofc to change it. I believe this max is a legacy restriction from the old traffic_manager stuff(for metrics and records).

I'll make some changes here(and add some AuTest) and we can keep discussing.

Thanks.

brbzull0:
[approve ci autest]

brbzull0:
Update after review feedback:

Changed the warning to check against the configured `--maxRecords`(either the default or the one passed by the user). If there are more records to read than the allocated, then ATS will fatal as we cannot store all records from `records.yaml`

zwoop (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11827, opened 2024-10-22
check if client response is still valid on txn close

sharkleash:
It's sometimes possible that the client may have gone away already before Cripts reaches the TxnClose hook. 

In the TxnClose hook, Cripts attempts to read the response status when determining which metric to increment, but if the client response object is already invalid, this will create a false assert/failure and subsequent ATS abort down the line in TSHttpHdrStatusGet.

This fix ensures that the response is still valid before completing the hook.


zwoop (approved):

bneradt (approved):

cmcfarlen:
Cherry-picked to 10.0.0

PR #11843, opened 2024-11-05
chunked body autests: avoid UUID colision

### LINKED ISSUE #11841: AuTest bad chunked encoding exclude collides with hash output
There's a hash in the `Via` header in the response that can randomly contain the string "abc"... the string that the AuTest makes sure is excluded from the response, intending to make sure it isn't in the response body. To fix it, we can exclude the headers from the check (could be difficult) or choose a value for the body that won't be present in a hash.

bneradt:
The ATS via header string contains a UUID which is a set of hex values. Since some of the chunked encoding tests used a chunked body of abc, this would sometimes coincidentally cause a failure when the autest found abc, not from the body, but from the UUID. This avoids that by including non-hex values in the test body.

Fixes #11841


randall (approved):

cmcfarlen:
Cherry-picked to v10.0.x

ezelkow1:
Do we still want this on 9.2?

PR #11844, opened 2024-11-05
Add metrics for negotiated group with client

masaori335:
Like TLS ciphers.

```
proxy.process.ssl.group.user_agent.P-256
proxy.process.ssl.group.user_agent.P-384
proxy.process.ssl.group.user_agent.P-521
proxy.process.ssl.group.user_agent.X25519
proxy.process.ssl.group.user_agent.P-224
proxy.process.ssl.group.user_agent.X448
proxy.process.ssl.group.user_agent.X25519MLKEM768
```

The `SSL_get_negotiated_group` is introduced by OpenSSL 3.0.0 (and BoringSSL), so this doesn't work with OpenSSL 1.1.1 and older.

### DIFF
--- a/src/iocore/net/SSLUtils.cc
+++ b/src/iocore/net/SSLUtils.cc
@@ -1086,6 +1086,28 @@ ssl_callback_info(const SSL *ssl, int where, int ret)
       }
       Metrics::Counter::increment(it->second);
     }
+
+#if defined(OPENSSL_IS_BORINGSSL)
+    uint16_t group_id = SSL_get_group_id(ssl);
+    if (group_id != 0) {
+      const char *group_name = SSL_get_group_name(group_id);
+      if (auto it = tls_group_map.find(group_name); it != tls_group_map.end()) {
+        Metrics::Counter::increment(it->second);
+      } else {
+        Warning("Unknown TLS Group");
+      }
+    }
+#elif defined(SSL_get_negotiated_group)
+    int nid = SSL_get_negotiated_group(const_cast<SSL *>(ssl));
+    if (nid != NID_undef) {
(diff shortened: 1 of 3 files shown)


src/iocore/net/SSLUtils.cc
```diff
@@ -1086,6 +1086,20 @@ ssl_callback_info(const SSL *ssl, int where, int ret)
...
+
+#if defined(OPENSSL_IS_BORINGSSL) || defined(SSL_get_negotiated_group)
+    // TLS Group
+#ifdef OPENSSL_IS_BORINGSSL
+    int nid = SSL_get_negotiated_group(ssl);
+#elif defined(SSL_get_negotiated_group)
+    int nid = SSL_get_negotiated_group(const_cast<SSL *>(ssl));
+#endif
+    if (nid != NID_undef) {
+      if (auto it = tls_group_map.find(nid); it != tls_group_map.end()) {
```
maskit, line 1098: We should have OTHER, otherwise we'd see a mysterious drop when SSL libraries and clients support a new group.
https://github.com/apache/trafficserver/pull/9623

And the table could be autogenerated if the library used has `SSL_get_all_group_names`. Only users who use SSL library without the function would be affected if new groups are added.

masaori335: Yeh, I was thinking the same thing. Will do them.

masaori335: BoringSSL has an API to get NID from group name?

maskit: No, it doesn't. Look at `cipher_map`. Its key is a string, not NID. It may not be great in terms of calculating the hash, but if we really care about it, I suppose we could use the pointer for the const char.

duke8253 (approved):

Issue #11836, opened 2024-10-31
When there's no response from the Origin, the escalate plugin does not function correctly.


incheoloh:
I'm using the 'escalate' plugin. 
When the origin responses like 2xx or 5xx, it works well with redirecting to second-origin. 
However, when there's no response from origin due to its unavailability, the 'escalate' plugin doesn't work as expected. 
I believe TS_HTTP_READ_RESPONSE_HDR_HOOK is not called if there's no response from Origin.

Is there any way to make a redirection to second-origin when there's no response from Origin?

incheoloh:
It would be great if there was a hook to handle the case when parent or origin doesn't respond. For example, TS_HTTP_READ_RESPONSE_ERROR_HDR_HOOK.

PR #11847, opened 2024-11-06
escalate plugin: handle origin not reachable

bneradt:
Update the escalate plugin to handle dispatching to the failover server if the original server is down.

This also adds a new autest for the escalate plugin.

Fixes: #11836


bneradt:
[approve ci autest]

cmcfarlen (approved):

PR #11850, opened 2024-11-08
Don't call getsockname on every packet

maskit:
`UnixUDPConnection` should have a local copy of the local address so no need to bother Kernel.

### DIFF
--- a/src/iocore/net/UnixUDPNet.cc
+++ b/src/iocore/net/UnixUDPNet.cc
@@ -592,7 +592,7 @@ UDPNetProcessorInternal::read_multiple_messages_from_net(UDPNetHandler *nh, UDPC
       return;
     }
 
-    toaddr.ss_family = AF_UNDEF;
+    toaddr.ss_family = AF_UNSPEC;
     if (mhdr.msg_controllen > 0) {
       for (auto cmsg = CMSG_FIRSTHDR(&mhdr); cmsg != nullptr; cmsg = CMSG_NXTHDR(&mhdr, cmsg)) {
         if (get_ip_address_from_cmsg(cmsg, &toaddr[packet_num])) {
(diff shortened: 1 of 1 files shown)


src/iocore/net/UnixUDPNet.cc
```diff
@@ -456,6 +455,10 @@ UDPNetProcessorInternal::read_single_message_from_net(UDPNetHandler *nh, UDPConn
       }
 #endif
     }
+    // Use the local address if none of the cmsg to get the destination address is unavailable
```
JosiahWI, line 458: > if none ... is unavailable

That wording seems a little strange; maybe it's a mistake? Otherwise, it might be clearer to say "if any ... is available".

maskit: I wanted to say "If we couldn't get the address from any cmsgs".

JosiahWI:
@maskit What's the reason for changing the socket storage type from `sockaddr_in6` to `sockaddr_storage`?

maskit:
Because it seemed more appropriate type, although sockkaddr_in6 is technically large enough for our use case.

JosiahWI (approved):
Thanks. It's nice not to waste resources calling `getsockname` when the `DNSConnection` already has a local copy. It looks like that local copy should always exist since it's created in `UDPBind`.

I'm sad to see the introduction of `reinterpret_cast`s, but it seems `sockaddr_storage` is designed for this kind of purpose, and the casts are guaranteed by POSIX in most cases not to violate strict aliasing, which is a nice bonus even though we don't enforce strict aliasing requirements.

maskit:
> It looks like that local copy should always exist since it's created in UDPBind.

This may not be true on outgoing QUIC connections, although we currently don't support it.

JosiahWI:
I think it's set on the QUIC connections as well.

PR #11859, opened 2024-11-13
TransformTerminus::do_io_*: handle null buffer

bneradt:
This is a very similar change as #11789. With this patch, TransformTerminus properly supports a 0 byte read to cancel VIO. Without this patch, the logic naively scheduled an event which called the handler which would crash on what was often a canceled continuation or otherwise invalid continuation.

### DIFF
--- a/include/proxy/TransformInternal.h
+++ b/include/proxy/TransformInternal.h
@@ -31,6 +31,7 @@ class TransformTerminus : public VConnection
 {
 public:
   TransformTerminus(TransformVConnection *tvc);
+  ~TransformTerminus() override;
 
   int handle_event(int event, void *edata);
 
@@ -49,6 +50,12 @@ public:
   int                   m_deletable;
   int                   m_closed;
   int                   m_called_user;
+
+private:
+  Event *_read_event     = nullptr;
+  bool   _read_disabled  = true;
+  Event *_write_event    = nullptr;
+  bool   _write_disabled = true;
 };
 
 class TransformVConnection : public TransformVCChain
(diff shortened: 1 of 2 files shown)


bneradt:
[approve ci autest freebsd]

include/proxy/TransformInternal.h
```diff
@@ -49,6 +50,12 @@ class TransformTerminus : public VConnection
   int                   m_deletable;
   int                   m_closed;
   int                   m_called_user;
+
+private:
+  Event *_read_event     = nullptr;
+  bool   _read_disabled  = false;
+  Event *_write_event    = nullptr;
+  bool   _write_disabled = false;
```
cmcfarlen, line 58: Is it safer for the disabled flags to default to true?

cmcfarlen (dismissed):
lgtm if my question's answer is  "no"

include/proxy/TransformInternal.h
bneradt: That's a good point. My logic was to default this new "feature" I'm adding to off.

The initial values shouldn't make a difference, since the events we're protecting only happen during the do_io_* function calls which will explicitly set these to true or false, but I see your logic for defaulting the flags to disabled since the buffer will be nullptr. I'll run a few quick internal tests in production to verify I'm not missing something and update the PR if it seems fine.

Thanks for the review!

bneradt: Updated.

cmcfarlen (approved):
Thanks!

PR #11863, opened 2024-11-14
Adds a HTTP-CNTL condition to header_rewrite

zwoop:
For example, this can be useful to avoid doing computationally expensive HRW rules if logging is turned off, if the computation is only used when producing logs.

```
cond %{SEND_RESPONSE_HDR_HOOK} [AND]
cond %{HTTP-CNTL:LOGGING}
   # Do expensive stuff here
```

I also cleaned up docs a bit, the operators and conditions are supposed to be in alphabetic order.


bneradt (approved):
lgtm

PR #11864, opened 2024-11-14
s3_auth does not retry after a file read error

### LINKED ISSUE #11831: s3_auth does not retry after a file read error
If s3_auth attempts to load its configuration from file and fails due to a file system error, it will not retry on its own.  This can cause credentials to expire without being renewed.

jasmine-nahrain:
Currently, if a config file is invalid or cannot be opened/read, ATS does not retry loading the config. As a result, if there is something wrong, ATS will not retry which can lead to 400 errors until there is manual intervention.

This adds an exponential retry to loading and validating the config. 

This fixes #11831

### DIFF
--- a/plugins/origin_server_auth/origin_server_auth.cc
+++ b/plugins/origin_server_auth/origin_server_auth.cc
@@ -30,6 +30,7 @@
 #include <cstdlib>
 #include <climits>
 #include <cctype>
+#include <cmath>
 
 #include <fstream> /* std::ifstream */
 #include <string>
--- a/plugins/origin_server_auth/origin_server_auth.cc
+++ b/plugins/origin_server_auth/origin_server_auth.cc
@@ -420,6 +420,12 @@ public:
     return _conf_reload_count++;
   }
 
+  int
+  incr_invalid_file_count()
+  {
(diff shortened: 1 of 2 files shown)


plugins/origin_server_auth/origin_server_auth.cc
```diff
@@ -1087,9 +1100,12 @@ config_reloader(TSCont cont, TSEvent /* event ATS_UNUSED */, void *edata)
   S3Config *file_config = gConfCache.get(s3->conf_fname());
 
   if (!file_config || !file_config->valid()) {
-    TSError("[%s] invalid configuration. Check mandatory fields.", PLUGIN_NAME);
+    TSError("[%s] invalid configuration. Check mandatory fields. Scheduling reload", PLUGIN_NAME);
+    long delay = pow(s3->incr_invalid_file_count(), 2);
```
masaori335, line 1105: Do you mean power of 2?  Please use bitwise left shift `1 << s3->incr_invalid_file_count()`.

bneradt (approved):

ezelkow1:

jasmine-nahrain:
No

PR #11865, opened 2024-11-14
Adds a fragment-size= option to volume.config

zwoop:


JosiahWI:
Would it make sense to remove the `cache_config_` global for this, or mark it const? The purposes seem to be redundant to me. 

Maybe it would be good to introduce a type to hold cache options for initialization at some point; doesn't need to be in this PR though.

JosiahWI:
The new configuration is per-volume. The `cache_config_` is the default for all volumes. We need both.

zwoop:
> Would it make sense to remove the `cache_config_` global for this, or mark it const? The purposes seem to be redundant to me.

Well, I don't think we can (easily) eliminate cache_config_target_fragment_size, it's a reloadable configuration, and it has to have a "destination" to write to in the callback.

What we could do is to simplify target_fragment_size() by always setting the StripeSMs size to either cache_config_target_fragment_size or the volume.config specified value. Since the cache is a bit of a mystery, I was worried about making too many changes ... But I could do this one (it avoids a "comparison" at runtime).

zwoop:
Actually, I take that back, the records.config value ("the default") being reloadable, we can't set that in the stripe value, because then the stripes are fixed for the runtime of the process. So I think we have to leave this as-is.

src/iocore/cache/StripeSM.h
```diff
@@ -168,7 +168,7 @@ class StripeSM : public Continuation, public Stripe
    *
    * @see START_POS
    */
-  StripeSM(CacheDisk *disk, off_t blocks, off_t dir_skip, int avg_obj_size = -1);
+  StripeSM(CacheDisk *disk, off_t blocks, off_t dir_skip, int avg_obj_size = -1, int fragment_size = -1);
```
JosiahWI, line 171: Looks like the docstring is out of date.

src/iocore/cache/Stripe.h
```diff
@@ -115,7 +117,7 @@ class Stripe
    *
    * @see START_POS
    */
-  Stripe(CacheDisk *disk, off_t blocks, off_t dir_skip, int avg_obj_size = -1);
+  Stripe(CacheDisk *disk, off_t blocks, off_t dir_skip, int avg_obj_size = -1, int fragment_size = -1);
```
JosiahWI, line 120: This could make take the vol config directly, especially if we are likely to add more configuration parameters in the future.

JosiahWI (changes requested):
I have a few ideas, but it looks good. It might also be good to exercise the new config option in some some AuTest. The only conceptual concern I have about this change is that previously, all volumes should have had the same fragment size, which I think would let us make some assumptions about moving fragments between volumes. I have no evidence that we do this anywhere, but we could have in theory. Now that volumes can have different fragment sizes, we can no longer make those assumptions, and if we wanted to move data between volumes, we would have to recalculate the fragments. Do you think there's any real concern there, or are we fine as far as that's concerned?

src/iocore/cache/Stripe.h
JosiahWI: On second thought, that may not be great, because each config is used for an entire cache span (volume), so the dependencies would get messier. I think it'd be nice to have a config type for stripes at some point, which could maybe be part of the volume config, but let's leave it as is for now.

src/iocore/cache/StripeSM.h
zwoop: Updated.

JosiahWI (approved):

PR #11866, opened 2024-11-15
Change the interpretation of cache fragment size

zwoop:
This shouldn't change any code, but it fixes two things:

1. When the setting was set to 0, we'd set the fragment size to 1MB - sizeof(Doc), which is 1MB-72. However, when we use the fragment size, we subtract sizeof(Doc) again, in `target_fragment_size().` The value returned by this helper function is what it ends up using when creating a fragment to write.
2. Also confusing is that if you try to set a too large fragment size, it reverts to 1MB, rather than the maximum (capped) value of 4MB.

I intentionally left MAX_FRAG_SIZE as it is (4MB - 72), because this constant is used all over the place and expects to be this adjusted value.

### DIFF
--- a/src/iocore/cache/Cache.cc
+++ b/src/iocore/cache/Cache.cc
@@ -860,8 +860,11 @@ ink_cache_init(ts::ModuleVersion v)
              REC_ERR_FAIL);
   REC_ReadConfigInt32(cache_config_target_fragment_size, "proxy.config.cache.target_fragment_size");
 
-  if (cache_config_target_fragment_size == 0 || cache_config_target_fragment_size - sizeof(Doc) > MAX_FRAG_SIZE) {
+  if (cache_config_target_fragment_size == 0) {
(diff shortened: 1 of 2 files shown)


JosiahWI:
Are you saying that the reason we have to subtract `sizeof(Doc)` from the fragment size before comparing with the maximum value is that `MAX_FRAG_SIZE` does not include the `sizeof(Doc)` and we don't want to change it because it could break other things?

zwoop:
Yes it's used all over the place already in cache computations, so I'm not going to change that define.

But, the previous code reading the records.config would potentially have caused us to deduct the fragment size by 144 instead of 72, because we subtract 72 before using it in the cache code.

Now, this likely never happened to anyone, it would only have been wrong if you set the setting to 0 or a value above 4MB.

JosiahWI (approved):
Awesome, thanks for the fix.

PR #11868, opened 2024-11-15
Coverity: Fix fox CID-1550411

brbzull0:


cmcfarlen (approved):
This is fine, it doesn't look like any callers check the return value though.

brbzull0:
> This is fine, it doesn't look like any callers check the return value though.

exactly, they don't.

PR #11855, opened 2024-11-12
Fix unchecked return value of initgroups()

maskit:

PR #11872, opened 2024-11-15
Invoke initgroups() iff we got enough privileges (#11869)

### LINKED ISSUE #11869: ATS 9.2.6 fails to start as root
As a side effect of https://github.com/apache/trafficserver/pull/11855 ATS is now unable to start as root:

[traffic_manager uses ImpersonateUser() ](https://github.com/apache/trafficserver/blob/de3e58726f6f58a95266a41eaa31ff3c6fc14336/src/traffic_manager/traffic_manager.cc#L1084)to drop privileges triggering the following syscalls:
```
450941 setgroups(1, [123])              = 0
450941 setresgid(-1, 123, -1)           = 0
450941 setresuid(-1, 118, -1)           = 0
```

afterwards traffic_server attempts to drop privileges as well but setgroups fails due to lack of permissions:
```
450952 setgroups(1, [123])              = -1 EPERM (Operation not permitted)
```

jvgutierrez:
Follow up of https://github.com/apache/trafficserver/pull/11855, that rendered unusable ATS as root when spawned via traffic_manager. As mentioned on #11869, traffic_manager drops privileges when spawned as root and as consequence by the time that traffic_server attempts to call initgroups() it gets an EPERM error. 

Fixes #11869

### DIFF
--- a/src/tscore/ink_cap.cc
+++ b/src/tscore/ink_cap.cc
@@ -156,8 +156,10 @@ impersonate(const struct passwd *pwd, ImpersonationLevel level)
 #endif
 
   // Always repopulate the supplementary group list for the new user.
-  if (initgroups(pwd->pw_name, pwd->pw_gid) != 0) {
-    Fatal("switching to user %s, failed to initialize supplementary groups ID %ld", pwd->pw_name, (long)pwd->pw_gid);
+  if (geteuid() == 0) { // check that we have enough rights to call initgroups()
+    if (initgroups(pwd->pw_name, pwd->pw_gid) != 0) {
+      Fatal("switching to user %s, failed to initialize supplementary groups ID %ld", pwd->pw_name, (long)pwd->pw_gid);
+    }
   }
 
   switch (level) {


ezelkow1 (dismissed):

maskit (dismissed):
The change is necessary for 9.2.x, which uses traffic_manager. I don't think we need this for 10.x but it's probably nice to keep the code consistent.

src/tscore/ink_cap.cc
```diff
@@ -156,8 +156,10 @@ impersonate(const struct passwd *pwd, ImpersonationLevel level)
 #endif
 
   // Always repopulate the supplementary group list for the new user.
-  if (initgroups(pwd->pw_name, pwd->pw_gid) != 0) {
-    Fatal("switching to user %s, failed to initialize supplementary groups ID %ld", pwd->pw_name, (long)pwd->pw_gid);
+  if (geteuid() != 0) { // check that we have enough rights to call initgroups()
```
cmcfarlen, line 159: This should be `geteuid() == 0`?

jvgutierrez: :sweat_smile: you're totally right, fixed :)

maskit (approved):

midchildan:
I believe this fix will reintroduce the problem in #11855. traffic_manager spawns traffic_server with an unprivileged effective UID and privileged real UID. With this change, traffic_server will keep root's supplementary groups because its EUID won't equal 0.

midchildan:
My bad, I got mixed up with a separate workaround I had for https://github.com/apache/trafficserver/issues/11869.

src/tscore/ink_cap.cc
```diff
@@ -156,8 +156,10 @@ impersonate(const struct passwd *pwd, ImpersonationLevel level)
 #endif
 
   // Always repopulate the supplementary group list for the new user.
-  if (initgroups(pwd->pw_name, pwd->pw_gid) != 0) {
-    Fatal("switching to user %s, failed to initialize supplementary groups ID %ld", pwd->pw_name, (long)pwd->pw_gid);
+  if (geteuid() == 0) { // check that we have enough rights to call initgroups()
+    if (initgroups(pwd->pw_name, pwd->pw_gid) != 0) {
```
midchildan, line 160: I think `initgroups()` should be called at the end of `impersonate()` too so that the supplementary groups would be repopulated when elevating privileges instead of dropping them.

jvgutierrez: that would require to add `CAP_SETGID` to `RestrictCapabilities()`? maybe it's worth moving this discussion to another PR/issue and solve #11869 with this one

ezelkow1:
cherry picked to 9.2.x

cmcfarlen:
Cherry-picked to v10.0.x

PR #11885, opened 2024-11-27
Add better logging for origin_server_auth

### LINKED ISSUE #10773: expired s3_auth token -- improve logging
In the s3_auth plugin for v4 signatures, there is an undocumented feature that allows for an expiration configuration. The plugin goes through considerable effort to reload a nearly expired token (https://github.com/apache/trafficserver/blob/6e00344df420db067e2b3cddbe8a7a48b7559700/plugins/s3_auth/s3_auth.cc#L1037), but is largely silent about doing so, which hampers operational ability to detect and remediate the issue. I'd suggest:

- In the cases mentioned in code comments (1 hour before, 15min before, and at expiration) logging that the token is about to expire as more than debug level.
- the log should have the token's name so that an operator can find the specific issue
- there appears to be an error log printed ~10min after the token has expired. I think it would be appropriate to continue logging every 10 times -- eg, https://github.com/apache/trafficserver/blob/6e00344df420db067e2b3cddbe8a7a48b7559700/plugins/s3_auth/s3_auth.cc#L1084 instead of "==10", perhaps s3->incr_conf_reload_count() % 10 == 0

jasmine-nahrain:
This fixes #10773 

- Adds the absolute file path to logs where relevant
- Changes logs related to scheduling a reload to TSNote (from debug)

### DIFF
--- a/plugins/origin_server_auth/origin_server_auth.cc
+++ b/plugins/origin_server_auth/origin_server_auth.cc
@@ -230,12 +230,12 @@ public:
     /* Check mandatory parameters first */
     if (versions::gcpv1 == _version && (!_token || !(_token_len > 0))) {
       Dbg(dbg_ctl, "version = %s; keyid = %s", versionString(), _keyid);
-      TSWarning("[%s] missing mandatory configs for version: %s", PLUGIN_NAME, versionString());
+      TSWarning("[%s] missing mandatory configs for version: %s in file: %s", PLUGIN_NAME, versionString(), _conf_fname);
       return false;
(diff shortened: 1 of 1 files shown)


brbzull0 (approved):
Looks good.
Thanks.

cmcfarlen:
removed milestone because this PR on 10.0.x #11914

ezelkow1:
Do we still want this for 9.2? I dont see the origin_server_auth plugin on there

jasmine-nahrain:
It was decided that because this involved renaming the plugin, we wont backport the changes.

PR #11890, opened 2024-12-04
Support splice in http blind tunnel

YIHONG-JIN:
1. Add TS_USE_LINUX_SPLICE as a compilation option.
2. Make MIOBuffer and MIOBufferReader polymorphic classes making their member function virtual.
3. Create PipeIOBuffer and PipeIOBufferReader as derived classes, encapsulating Linux pipe.
4. Use dynamic_cast to enable logic switch in state machines and continuations.

Documentations:
[ATS_splice_runbook.md](https://github.com/user-attachments/files/18005014/ATS_splice_runbook.md)
[ATS Performance Benchmark.pdf](https://github.com/user-attachments/files/18072370/ATS.Performance.Benchmark.pdf)
[ATS_splice_design_doc.pdf](https://github.com/user-attachments/files/18005025/ATS_splice_design_doc.pdf)


YIHONG-JIN:
Passes unit test and manual integration test on Debian 12. Performance benchmark shows that splice could improve maximum blind tunnel throughput from 300 MB/s to 575 MB/s and reduce latency by 40% for MB level payload on C6in.large EC2 instance. Feel free to benchmark this CR.

src/iocore/net/UnixNetVConnection.cc
```diff
@@ -508,6 +508,81 @@ UnixNetVConnection::net_read_io(NetHandler *nh)
     read_disable(nh, this);
     return;
   }
+#if TS_USE_LINUX_SPLICE
```
moonchen, line 511: This function is getting quite long, with two independent code paths to read from socket to buffer.  I would prefer using polymorphism to handle the different socket-to-buffer copies, but at least we should split this function up so that it's more readable.

src/iocore/eventsystem/P_IOBuffer.h
```diff
@@ -656,6 +656,15 @@ new_MIOBuffer_internal(const char *location, int64_t size_index)
 TS_INLINE void
 free_MIOBuffer(MIOBuffer *mio)
 {
+#if TS_USE_LINUX_SPLICE
+  // check if mio is PipeIOBuffer using dynamic_cast
+  PipeIOBuffer *pipe_mio = dynamic_cast<PipeIOBuffer *>(mio);
```
moonchen, line 661: Relying on a runtime type check with dynamic_cast to select the proper deallocation mechanism is generally considered an anti-pattern.  Instead, ClassAllocator has a `Destruct_on_free_` parameter that allows you to use a virtual destructor to clean up properly depending on the underlying type.

moonchen (changes requested):
Thank you for the PR.  Overall I think this is a clever idea to introduce zero-copy in a way that fits the existing architecture.  I found a few small issues that I hope you can address before merging.

YIHONG-JIN:
Thanks for the comments @moonchen. I will start to resolve the comments

src/proxy/http/HttpSM.cc
```diff
@@ -7274,6 +7274,32 @@ HttpSM::setup_blind_tunnel(bool send_response_hdr, IOBufferReader *initial)
   //  header buffer into new buffer
   client_request_body_bytes += from_ua_buf->write(_ua.get_txn()->get_remote_reader());
 
+#if TS_USE_LINUX_SPLICE
+  MIOBuffer *from_ua_pipe_buf = new_PipeIOBuffer(BUFFER_SIZE_INDEX_32K);
```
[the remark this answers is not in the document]
YIHONG-JIN: Yes, it is the same as the MIOBuffer size we use for blind tunnel. The default linux pipe size is 16 pages so I am trying to save memory by requesting only 8 pages here.  However, the system admin still need to lift pipe-user-pages-soft limit to avoid exceptions.

src/iocore/net/UnixNetVConnection.cc
```diff
@@ -508,6 +508,81 @@ UnixNetVConnection::net_read_io(NetHandler *nh)
...
+  PipeIOBuffer *pipe_buffer = dynamic_cast<PipeIOBuffer *>(buf.writer());
+  if (pipe_buffer) {
+    // Use splice_to to transfer data from socket directly to pipe with SPLICE_F_MORE hint
+    int64_t to_splice = std::min(ntodo, pipe_buffer->write_avail());
+    if (to_splice > 0) {
+      r = con.sock.splice_to(pipe_buffer->fd[1], to_splice, SPLICE_F_MOVE | SPLICE_F_NONBLOCK);
+
+      if (r <= 0) {
+        // Temporary Unavailable, Non-Blocking I/O
+        if (r == -EAGAIN || r == -ENOTCONN) {
```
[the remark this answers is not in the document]
YIHONG-JIN: It is actually impossible to get EAGAIN here **because the pipe is at capacity**. We will disable the read vio after each successful read and only reenable it when its corresponding pipe is empty again. 

However, it is possible that we get EAGAIN because the socket is somehow unavailable. In that case, we wait for next epoll edge trigger same as the logic without zero copy

src/iocore/net/UnixNetVConnection.cc
YIHONG-JIN: I have considered using polymorphism here but it looks impossible without significant refactor because of exposure of low level io operations in this function. Will split this function up for now

src/iocore/eventsystem/P_IOBuffer.h
YIHONG-JIN: The PipeIOBuffer is now managed by pipeIOAllocator instead of ioAllocator. I introduced pipeIOAllocator because it is a new class with additional attributes. The ClassAllocator appears to be designed for concrete classes rather than abstract classes or interfaces.

1. I didn’t move the cleanup logic to the virtual destructor because the existing logic for MIOBuffer doesn’t use it. Instead, it manually handles cleanup in free_MIOBuffer with the following steps:
```
mio->_writer = nullptr;  
mio->dealloc_all_readers();  
```
Interestingly, the destructor already performs this same cleanup, but for some reason, the existing code avoids relying on it. If anyone knows why, I’d appreciate the insight.

2. Even if we moved the cleanup logic to the virtual destructor, we would still need a runtime type check since true polymorphism is not yet in place. For example, when dealing with a PipeIOBuffer, we would need to use
```
THREAD_FREE(mio, pipeIOAllocator, this_thread());
```
instead of 
```
THREAD_FREE(mio, ioAllocator, this_thread());
```
to make sure the correct allocator is used.

YIHONG-JIN: Will keep it as it was until figure out how to implement real polymorphism here. More background info required.

YIHONG-JIN:
Slightly reorganized the file structure to decouple PipieIOBuffer from original IOBuffer.

src/iocore/eventsystem/P_PipeIOBuffer.h
```diff
@@ -0,0 +1,503 @@
...
+//////////////////////////////////////////////////////////////////
+//
+//  class PipeIOBufferReader --
+//         inline functions definitions
+//
+//////////////////////////////////////////////////////////////////
+TS_INLINE char *
+PipeIOBufferReader::start()
+{
+  throw std::runtime_error("Not applicable for PipeIOBufferReader");
```
masaori335, line 40: Hmm, this PR is pretty interesting as concept. However, overriding bunch of methods and throwing runtime errors like this is indicating something is wrong with class design. ( Also, please note that ATS doesn't use Exception in these fundamental code )

YIHONG-JIN: I’m aware of this and have discussed it with Susan. The ultimate solution would be to extract a pure abstract class named IOBuffer, making MIOBuffer and PipeIOBuffer both derived classes of IOBuffer. This would provide a cleaner separation of concerns and eliminate the need for unnecessary overrides. However, this would require significant engineering effort, and I may not have enough resources to invest in this at the moment.

The reason I explicitly throw errors in overridden methods is to prevent developers from incorrectly calling MIOBuffer-specific methods on inappropriate instances, which could lead to undefined behavior. This ensures that such issues are caught early rather than silently causing unintended consequences.

I’m open to discussing alternative approaches if there’s a way to improve the class design while keeping the effort reasonable.

YIHONG-JIN:
The latest revision is to resolve the following comments: 

1. https://github.com/apache/trafficserver/pull/11890#discussion_r1925531050
2. https://github.com/apache/trafficserver/pull/11890#discussion_r1925533339

moonchen (approved):

moonchen:
[approve ci]

moonchen:
[approve ci]

moonchen:
@YIHONG-JIN can you take a look at the CI failures?  Looks like clang-format and license header checks found some problems.

YIHONG-JIN:
Thanks! Will take a look at CI failures this weekend. @moonchen

bryancall:
[approve ci]

YIHONG-JIN:
Took a look. Looks like format is fine this time. @bryancall Is there a way to read the log of failed tests? Looks like I don't have permissions to do that

bryancall:
[approve ci]

bryancall:
[approve ci]

bryancall:
@YIHONG-JIN sorry for the late response, we went through all the open PRs yesterday in our meeting.

You can go to the URLs below and look at the console output.  For example here is the run for Ubuntu:
https://ci.trafficserver.apache.org/job/Github_Builds/job/ubuntu/6069/
And here is the Ubuntu build output:
https://ci.trafficserver.apache.org/job/Github_Builds/job/ubuntu/6069/console

And I am seeing the errors:
```
/usr/bin/ccache /usr/lib/llvm-12/bin/clang++ -DDEBUG -DPACKAGE_NAME="\"Apache Traffic Server\"" -DPACKAGE_VERSION=\"10.2.0\" -D_DEBUG -Dlinux -I/home/jenkins/workspace/Github_Builds/ubuntu/src/include -I/home/jenkins/workspace/Github_Builds/ubuntu/src/build/include -I/home/jenkins/workspace/Github_Builds/ubuntu/src/lib/swoc/include -I/home/jenkins/workspace/Github_Builds/ubuntu/src/lib/yamlcpp/include -isystem /home/jenkins/workspace/Github_Builds/ubuntu/src/lib/systemtap -D_FORTIFY_SOURCE=2 -fPIE -fstack-protector -pthread -g -std=c++20 -Wno-invalid-offsetof -pipe -Wall -Wextra -Wno-noexcept-type -Wsuggest-override -Wno-vla-extension -fno-strict-aliasing -Wno-deprecated-declarations -mcx16 -Werror -MD -MT src/iocore/net/CMakeFiles/inknet.dir/UnixNetVConnection.cc.o -MF src/iocore/net/CMakeFiles/inknet.dir/UnixNetVConnection.cc.o.d -o src/iocore/net/CMakeFiles/inknet.dir/UnixNetVConnection.cc.o -c /home/jenkins/workspace/Github_Builds/ubuntu/src/src/iocore/net/UnixNetVConnection.cc
../src/iocore/net/UnixNetVConnection.cc:643:3: error: unknown type name 'PipeIOBuffer'
  PipeIOBuffer *pipe_buffer = dynamic_cast<PipeIOBuffer *>(buf.writer());
  ^
../src/iocore/net/UnixNetVConnection.cc:643:44: error: unknown type name 'PipeIOBuffer'
  PipeIOBuffer *pipe_buffer = dynamic_cast<PipeIOBuffer *>(buf.writer());
                                           ^
../src/iocore/net/UnixNetVConnection.cc:649:20: error: no member named 'splice_to' in 'UnixSocket'
      r = con.sock.splice_to(pipe_buffer->fd[1], to_splice, SPLICE_F_MOVE | SPLICE_F_NONBLOCK);
          ~~~~~~~~ ^
3 errors generated.
```

PR #11895, opened 2024-12-05
Cleanup: Remove disk path copy from Stripe

masaori335:
`Stripe::path` has copy of `CacheDisk::path`, but `Stripe` has pointer of `CacheDisk` and it doesn't change. 

This is part of making `Stripe` smaller and copyable.


JosiahWI (approved):
Nice cleanup to reduce duplication.

PR #11896, opened 2024-12-05
Cleanup: Remove fd copy from Strip

masaori335:
Pretty similar change to the #11895 but for `fd`.

### DIFF
--- a/src/iocore/cache/StripeSM.cc
+++ b/src/iocore/cache/StripeSM.cc
@@ -154,7 +154,7 @@ StripeSM::clear_dir()
   size_t dir_len = this->dirlen();
   this->_clear_init();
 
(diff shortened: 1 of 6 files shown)


src/iocore/cache/StripeSM.cc
```diff
@@ -214,7 +214,7 @@ StripeSM::handle_dir_clear(int event, void *data)
     if (!op->ok()) {
       Warning("unable to clear cache directory '%s'", hash_text.get());
       disk->incrErrors(op);
-      fd = -1;
+      disk->fd = -1;
```
masaori335, line 217: @moonchen Would you take a look at this change? When `CacheDisk` got error, can we make the `fd = -1` or we should keep it?

masaori335:
Hmm, this breaks bunch of Cache tests. I'll leave it as is for now.

PR #11899, opened 2024-12-05
Fix records length for log roll and normalize

ezelkow1:
The current lengths for 2 of the log roll fields and normalize_ae are off by 1 which causes a warning when loading and then falls back to the default value


ezelkow1:
[approve autest 1]

ezelkow1:
[approve ci autest 1]

shukitchan (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11902, opened 2024-12-06
Update ssl.session_cache.enabled to be session_cache.value

### LINKED ISSUE #11901: ssl.session_cache.enabled unknown
While bringing up ats10 and scrubbing through all the 'unknown' fields that have been removed and show up in traffic.out I noticed that the ssl.session_cache.enabled comes up as unknown. This is still shown in the ATS 10 docs so there appears to be an inconsistency somewhere, either the field isnt working or its been removed and the docs should be updated

ezelkow1:
Fixes #11901

### DIFF
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -3879,7 +3879,7 @@ SSL Termination
   Setting a value less than or equal to ``0`` effectively disables
   SSL session cache for the origin server.
 
-.. ts:cv:: CONFIG proxy.config.ssl.session_cache.enabled INT 2
+.. ts:cv:: CONFIG proxy.config.ssl.session_cache.value INT 2
 
    Enables the SSL session cache:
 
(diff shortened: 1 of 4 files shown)


maskit:
I thought the implementation should be changed instead because the documentation is what users see and new users who didn't use ATS 9 would be required to update their settings. But, unfortunately, convert2yaml.py disagrees.
https://github.com/apache/trafficserver/blob/dc59c4c1fc8243f60601127442ea6835c434ffdd/tools/records/convert2yaml.py#L48-L49

I think the best way is to support the both names (`.value` and `.enabled`) for now, and introduce new name (`.mode`) to address the inconsistency between session_cache and origin_session_cache.

ezelkow1:
yea I did notice that the upgrading to 10 doc does mention some of these. So there is a reference there for previous users:
https://docs.trafficserver.apache.org/en/10.0.x/release-notes/upgrading.en.html

bryancall:
It looks like there are other configuration options named proxy.*.enabled and this is the only one with proxy.*.value.  It would be great if this was consistent with other settings.  @maskit suggested supporting both configuration option names for now.  I would lean towards this as a solution too.

maskit:
We can introduce `mode` later if we want. Sooner would be better but I think 10.1 is reasonable (and we should deprecate the current ones). For now, we should address the inconsistency between the documentation and the implementation.

If we support the both names (for this particular setting), which one wins if the both `enabled` and `value` exist and have different values? Should it be an error?

PR #11908, opened 2024-12-12
Coverity. `swoc` Make Name Binding global variables a singleton. 

brbzull0:
This is an attempt to fix some ordering creation issues for some global variables within  lib swoc. 
The is to get `swoc::bwf::Global_Names` initialized first.

[#1534740](https://scan6.scan.coverity.com/#/project-view/58063/10191?selectedIssue=1534740) and more...

### DIFF
--- a/lib/swoc/include/swoc/bwf_base.h
+++ b/lib/swoc/include/swoc/bwf_base.h
@@ -467,7 +467,7 @@ protected:
  * This nameset is used if no other is provided. Therefore bindings added to this nameset will be
  * available in the default formatting use.
  */
-extern ExternalNames Global_Names;
+extern ExternalNames& Global_Names();
 
 // --------------- Implementation --------------------
 /// --- Spec ---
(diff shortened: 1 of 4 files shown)


brbzull0:
[approve ci debian]

brbzull0:
`rocky` issue here is that ci is using an external libswoc and this changes the library objects, so it fails.

brbzull0:
[approve autest 1]

brbzull0:
@bneradt you think we should patch `libswoc` first?

brbzull0:
`fedora` is failing because it uses an external `libswoc`.

@bneradt  what do you think? should I patch `libswoc` repo first in this case?

lib/swoc/include/swoc/bwf_base.h
```diff
@@ -467,7 +467,7 @@ template <typename T> class ContextNames : public NameMap<BufferWriter &(BufferW
  * This nameset is used if no other is provided. Therefore bindings added to this nameset will be
  * available in the default formatting use.
  */
-extern ExternalNames Global_Names;
+extern ExternalNames& Global_Names();
```
cmcfarlen, line 470: You can drop `extern` here since its a function.

cmcfarlen (approved):
Seems good if it helps with the order.

brbzull0:
[approve ci rocky]

bneradt:
[approve ci rocky]

cmcfarlen:
Cherry-picked to v10.0.x

PR #11911, opened 2024-12-13
Make error if remap ACL has more than one @action

masaori335:
One remap line or named filter takes only one `@action=`. However, prior to the change, ATS doesn't have check for this on loading remap.config.

### DIFF
--- a/src/proxy/http/remap/RemapConfig.cc
+++ b/src/proxy/http/remap/RemapConfig.cc
@@ -471,6 +471,7 @@ remap_validate_filter_args(acl_filter_rule **rule_pp, const char **argv, int arg
     Dbg(dbg_ctl_url_rewrite, "[validate_filter_args] new acl_filter_rule class was created during remap rule processing");
   }
 
+  bool action_flag  = false;
   bool ip_is_listed = false;
   for (i = 0; i < argc; i++) {
     unsigned long ul;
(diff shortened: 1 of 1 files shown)


src/proxy/http/remap/RemapConfig.cc
```diff
@@ -632,6 +633,13 @@ remap_validate_filter_args(acl_filter_rule **rule_pp, const char **argv, int arg
     }
 
     if (ul & REMAP_OPTFLG_ACTION) { /* "action=" option */
+      if (action_flag) {
+        std::string_view err = "Only one @action= is available";
```
bneradt, line 637: `available` -> `allowed per remap ACL`

bneradt (changes requested):
Looks good. I just have a suggestion to rephrase.

src/proxy/http/remap/RemapConfig.cc
masaori335: Thank you for suggestion. Updated.

bneradt (approved):

masaori335:
For 9.2.x and 10.0.x, making Warning instead of Error for compatibility. Thus, 10.1.0 and greater will make it Error.

cmcfarlen:
Cleared milestone for #11925

PR #11913, opened 2024-12-13
Restore speed bucket metrics with more buckets

cmcfarlen:
This restores the origin server and user agent speed buckets that were removed in 10.0.0 but have been sorely missed.
The new buckets are 200M, 400M, 800M and the new catch all bucket is 1G


maskit:
[approve ci freebsd]

bneradt (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11919, opened 2024-12-17
Deduplicate StripeSM evacuate hit window computation

JosiahWI:
The same expression is duplicated in 3 places, and 2 of them have a comment which I borrowed the name from. This allows us to make `Stripe::data_blocks` protected.


masaori335 (approved):
Looks good.

PR #11926, opened 2024-12-19
Group directory related `Stripe` fields into struct

JosiahWI:
This is all a mechanical cleanup change, broken into four individual commits. I intend to follow this up with another PR to make all the directory operations methods on this struct instead of standalone functions, with the objective of breaking the cyclic dependency between the directory functions and the stripe.

It is interesting to note that the `Stripe::direntries` method existed, but the expression it computed was used inline in multiple places where the method could have been called. I went ahead and changed those places to call the method, since the use of the method clarifies what quantity is being computed.

I did not move the `Stripe::headerlen` and `Stripe::dirlen` methods because the concept of rounding to a store block seems to be specific to the stripe, and the directory functions do not depend on those methods.

### DIFF
--- a/src/iocore/cache/CacheDir.cc
+++ b/src/iocore/cache/CacheDir.cc
@@ -403,7 +403,7 @@ dir_clean_vol(Stripe *stripe)
 void
 dir_clear_range(off_t start, off_t end, Stripe *stripe)
 {
-  for (off_t i = 0; i < stripe->directory.buckets * DIR_DEPTH * stripe->directory.segments; i++) {
+  for (off_t i = 0; i < stripe->directory.entries(); i++) {
     Dir *e = dir_index(stripe, i);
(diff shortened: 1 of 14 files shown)


masaori335:
> I intend to follow this up with another PR to make all the directory operations methods on this struct instead of standalone functions, with the objective of breaking the cyclic dependency between the directory functions and the stripe.

You're going to move all `dir_*` functions into the new `Directory` struct?

JosiahWI:
That's correct. Right now, those functions take the entire Stripe structure as an argument, but the only fields they need to do their job are those related to the directory searching/bookkeeping.

masaori335:
Sounds good. I'll approve this after conflicts are resolved.

JosiahWI:
Rebased on master.

masaori335 (approved):
Looks good, it's all mechanical change.

PR #11927, opened 2024-12-21
Fix documentation to control caching for compressed range requests

vmamidi:


Clendenin (approved):

doc/admin-guide/plugins/compress.en.rst
```diff
@@ -107,7 +107,9 @@ range-request
 -------------
 
 When set to ``true``, causes |TS| to compress responses to Range Requests.
-Disabled by default. Setting this to true while setting cache to false leads to delivering corrupted content.
+Disabled by default.
+
+WARNING! Do NOT set this to true if cache is set to false. This combination will deliver corrupt content.
```
mlibbey, line 112: Is this really

```
 if :ts:cv:`proxy.config.http.cache.http` is set to false (0)
 ```

**Comment by bneradt** [comment]
In the PR/issue meeting today, we discussed the plugin issuing an ERROR message if this body-corrupting configuration combination is detected.

File: doc/admin-guide/plugins/compress.en.rst (hunk 1 above)
Comment (masaori335) [reply doc/admin-guide/plugins/compress.en.rst:112]: This `cache` here is a configuration of compress plugin. https://docs.trafficserver.apache.org/en/10.0.x/admin-guide/plugins/compress.en.html#cache

Config like below generates corrupted content. ( compressed partial content with status code 200 OK )
```
cache false
range-request true
```

Also, I want to mention a thing, even if this `cache` config is `true`, and `range-request` is `false` like below,
```
cache true
range-request false
```
once compressed content is cached, response to the Range request will be "partial content of compressed contents". 

IMO, if a client send both of `Accept-Encoding` and `Range` header, we should choose one of them as a workaround. ( I'll file an issue and PR for this )

masaori335: The second case, cache true and range-request false, is covered by #11975

masaori335:
Sorry Vijay, doc is updated. 

> Do NOT set this to none (or true) if the cache config is set to false. This combination will deliver corrupted content.

https://docs.trafficserver.apache.org/en/latest/admin-guide/plugins/compress.en.html#range-request

PR #11939, opened 2025-01-06
Turn off ccache for coverity build

ezelkow1:
I think ccache might be messing with coverity results/building so want to disable it for those runs


JosiahWI (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #11941, opened 2025-01-07
Configure autests to use microserver 1.0.8

bneradt:
microserver 1.0.8 has updates to run with Python 3.13 while still being backwards compatible.


dragon512 (approved):

cmcfarlen:
Cherry-picked to 10.0.0

ezelkow1:

PR #11936, opened 2025-01-02
Do not allow extra CRs in chunks

maskit:

PR #11942, opened 2025-01-07
Do not allow extra CRs in chunks (#11936)

maskit:
* Do not allow extra CRs in chunks

* Renumber test uuid

* Add test cases and fix an oversight

* Use prefix increment

(cherry picked from commit f5f2256c00abbfd02c22fbae3937da1c7bd8a34f)

### DIFF
--- a/proxy/http/HttpTunnel.cc
+++ b/proxy/http/HttpTunnel.cc
@@ -136,6 +136,7 @@ ChunkedHandler::read_size()
 {
   int64_t bytes_used;
   bool done = false;
+  int cr    = 0;
 
   while (chunked_reader->read_avail() > 0 && !done) {
     const char *tmp   = chunked_reader->start();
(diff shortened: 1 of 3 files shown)


bneradt:
[approve ci autest 0]

maskit:
The keyword, "abc", should not appear on the origin server log, because requests that contains the keyword should be blocked by ATS, and the test origin server does not send the keyword. But the keyword accidentally appears because the response data which I added for testing response parser has "abc" where it should be "def".

https://github.com/apache/trafficserver/blob/d9195ac85c4958355e5202006d2495989b959652/tests/gold_tests/chunked_encoding/bad_chunked_encoding.test.py#L184-L186

The test passes if I fix the test case (I confirmed this locally), but that makes me wonder why the tests passed on master.

maskit:
The test on master is slightly different. It checks "abcwxyz" instead of "abc". Thus, "abc" was not caught.
https://github.com/apache/trafficserver/pull/11843

maskit:
tls_verify4 failed

ezelkow1 (approved):

PR #11943, opened 2025-01-07
THREAD_FREE: rely less on macros

moonchen:
Move as much of the implementation of THREAD_FREE as possible into a type-safe function.  This was already done for THREAD_ALLOC.

### DIFF
--- a/include/iocore/eventsystem/ProxyAllocator.h
+++ b/include/iocore/eventsystem/ProxyAllocator.h
@@ -84,17 +84,23 @@ void thread_freeup(Allocator &a, ProxyAllocator &l);
 
 #endif
 
-#define THREAD_FREE(_p, _a, _tin)                                                                  \
-  do {                                                                                             \
-    ::_a.destroy_if_enabled(_p);                                                                   \
-    Thread *_t = (_tin);                                                                           \
-    if (_t && !cmd_disable_pfreelist) {                                                            \
-      *(char **)_p    = (char *)_t->_a.freelist;                                                   \
-      _t->_a.freelist = _p;                                                                        \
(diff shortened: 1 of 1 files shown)


bneradt (changes requested):
Good change.

Do you think `thread_free` should be marked inline?

include/iocore/eventsystem/ProxyAllocator.h
```diff
@@ -84,17 +84,23 @@ void thread_freeup(Allocator &a, ProxyAllocator &l);
...
+
+template <class T>
+void
+thread_free(T *p, Allocator &global_allocator, ProxyAllocator &thread_allocator, Thread *tin)
+{
+  global_allocator.destroy_if_enabled(p);
+  if (tin && !cmd_disable_pfreelist) {
+    *(reinterpret_cast<char **>(p)) = reinterpret_cast<char *>(thread_allocator.freelist);
+    thread_allocator.freelist       = p;
+    thread_allocator.allocated++;
```
bneradt, line 99: nit: let's change to prefix increment.

bneradt (changes requested):

include/iocore/eventsystem/ProxyAllocator.h
moonchen: Changed to prefix increment

moonchen:
> Good change.
> 
> Do you think `thread_free` should be marked inline?

I don't know.  We don't need it to comply with ODR since the function is a template function.  My understanding is that modern compilers are already smart about whether to inline the function, so it's hard to say whether adding the keyword will help with performance.

moonchen:
[approve ci autest]

PR #11945, opened 2025-01-08
Make directory operations methods on `Directory`

JosiahWI:
These operations are working with directory data. Right now, they operate on data of type `Stripe` or `StripeSM`, but that was because those data structures contained the directory data they needed until recently. Most of the other data there is irrelevant. 

This PR moves those operations (see commits for the list) into the `Directory` structure to reduce the dependence on the irrelevant parts of `Stripe` and `StripeSM`.

The `StripeSMs` are needed in some cases for debugging, metrics, and an intriguing invocation of `Stripe::dir_valid` from `dir_probe`; the `Directory` structure will need to be adjusted to have access to a few things for those purposes, but the details of that are still a work in progress.


bryancall:
[approve ci autest 2]

JosiahWI:
Rebased on master.

JosiahWI:
I have not removed the dependency of `Directory` on `StripeSM` or `Stripe`, but after looking at this multiple times, I'm not sure it's a good idea to try to force that. It would be nice to have a cleaner dependency structure, but we should aim to simplify things, not add complexity. This PR cleans up some call sites and makes it clearer what data is required to carry out each operation. Therefore, I think it is a good change as-is.

src/iocore/cache/Cache.cc
```diff
@@ -342,7 +342,7 @@ Cache::open_read(Continuation *cont, const CacheKey *key, CacheFragType type, st
   CacheVC      *c     = nullptr;
   {
     CACHE_TRY_LOCK(lock, stripe->mutex, mutex->thread_holding);
-    if (!lock.is_locked() || (od = stripe->open_read(key)) || dir_probe(key, stripe, &result, &last_collision)) {
+    if (!lock.is_locked() || (od = stripe->open_read(key)) || stripe->directory.probe(key, stripe, &result, &last_collision)) {
```
masaori335, line 345: Note for later refactoring: this can be like this?
```
stripe->dir_probe(key, &result, &last_collision)
```

src/iocore/cache/CacheDir.cc
```diff
@@ -217,41 +217,41 @@ dir_bucket_loop_check(Dir *start_dir, Dir *seg)
...
+      directory->free_entry(dir_bucket_row(bucket, l), s);
     }
   }
 }
 
 // break the infinite loop in directory entries
 // Note : abuse of the token bit in dir entries
 int
-dir_bucket_loop_fix(Dir *start_dir, int s, Stripe *stripe)
+dir_bucket_loop_fix(Dir *start_dir, int s, Directory *directory)
```
masaori335, line 237: Can we make this function a member of `Directory` too?

src/iocore/cache/CacheDir.cc
```diff
@@ -304,14 +304,14 @@ check_dir(Stripe *stripe)
 }
 
 inline void
-unlink_from_freelist(Dir *e, int s, Stripe *stripe)
+unlink_from_freelist(Dir *e, int s, Directory *directory)
```
masaori335, line 307: Similar to above. Would you share me what's the policy of which function should be moved to `Directory` or not here?

masaori335 (approved):
This just moves `Dir` operators into `Directory` for cleanup. Probably, we need more refactoring in this area, but this seems good to me.

src/iocore/cache/CacheDir.cc
JosiahWI: I looked for the functions that were operations on the directory, and I might have missed this one.

src/iocore/cache/CacheDir.cc
JosiahWI: Definitely, I missed this opportunity. Very easy improvement for further refactoring. Thanks for catching it.

JosiahWI: I think `dir_init_segment` can be a method of `Directory` as well.

src/iocore/cache/CacheDir.cc
JosiahWI: Yeah, this should be moved. I see what happened. I only moved the public interface, and not any helper functions. Further refactoring should make all these helper functions private methods on `Directory`.

PR #11951, opened 2025-01-13
Add unit tests for remap rules. Fix release-assert crasher

cmcfarlen:
This PR adds some tests for remap rule parsing and fixes a potential crasher when multiple `.definefilter` lines use the same filter name.


include/proxy/http/remap/RemapConfig.h
```diff
@@ -76,6 +76,7 @@ struct BUILD_TABLE_INFO {
 };
 
 const char *remap_parse_directive(BUILD_TABLE_INFO *bti, char *errbuf, size_t errbufsize);
+bool        remap_parse_config_bti(const char *path, BUILD_TABLE_INFO *bti);
```
cmcfarlen, line 79: This declaration moved out of anonymous namespace so it can be tested.

src/proxy/http/remap/RemapConfig.cc
```diff
@@ -146,6 +143,15 @@ process_filter_opt(url_mapping *mp, const BUILD_TABLE_INFO *bti, char *errStrBuf
...
       }
+      if (auto rule = *rpp; rule) {
+        // If no IP addresses are listed, treat that like `@src_ip=all`.
+        if (rule->src_ip_valid == 0 && rule->src_ip_cnt == 0) {
+          src_ip_info_t *ipi       = &rule->src_ip_array[rule->src_ip_cnt];
+          ipi->match_all_addresses = true;
+          rule->src_ip_cnt++;
+          rule->src_ip_valid = 1;
+        }
+      }
```
cmcfarlen, line 154: Moved this logic to when the filter rules are merged during a map declaration.  This avoids the need to check if a src_ip was already added since this happens outside of filter definition.

bneradt (dismissed):
The unit tests will be really helpful for testing multiple filter interactions. I assume we'll eventually add method and src_ip verification across multiple defines for the same name. But this is a good foundation for future PRs to do that.

cmcfarlen:
Turns out these tests found a legit memory leak

src/proxy/http/remap/RemapConfig.cc
```diff
@@ -100,6 +97,14 @@ BUILD_TABLE_INFO::BUILD_TABLE_INFO()
...
 {
   this->reset();
+
+  // clean up any leftover named filter rules
+  auto *rp = rules_list;
+  while (rp != nullptr) {
+    auto *tmp = rp->next;
+    delete rp;
+    rp = tmp;
+  }
```
bneradt, line 107: Awesome.

Let's not do this in this PR because what you're doing here with delete appropriately fits the current setup, but it would be good to change these over to shared_ptrs. (I think they are shared across rules so unique_ptr won't work?).

bneradt (approved):

src/proxy/http/remap/RemapConfig.cc
cmcfarlen: They aren't actually shared, but copied into the actual mappings as they are created.  This code is very pointer-heavy and is over due for a general update to C++ pass.  I do plan on expanding this work soon, but wanted to get these couple fixes that will need to be back ported.  Thanks for the review!

PR #11955, opened 2025-01-14
ci-ubuntu preset: use clang

bneradt:
Update the ci-ubuntu preset to explicitly use clang. Currently CI configures clang via setting CC and CXX environment variables. This will configure clang across other environments more consistently.


cmcfarlen (approved):
Thanks!

cmcfarlen:
Cherry-picked to v10.0.x

PR #11956, opened 2025-01-14
Fix a test case for malformed chunk header

maskit:
This fixes the typo which found on #11942.


bneradt (approved):
Good catch.

cmcfarlen:
Cherry-picked to 10.0.0

PR #11957, opened 2025-01-15
Add unit tests for remap rules. Fix release-assert crasher (#11951) 

cmcfarlen:
Backports #11951 and adjusts the test to account for the warning for specifying multiple actions in the same rule.


cmcfarlen:
[approve ci rocky]

PR #11965, opened 2025-01-16
multiplexer: fix consume of too many bytes (#11499)

ezelkow1:
IOBufferReader::consume contractually assumes that all callers of it will not consume more bytes than read_avail for the buffer chain. This adds a debug assertion for this and fixes multiplexer so that it doesn't violate this invariant.

(cherry picked from commit 6774984a3f71f7473092d1af7ba99a004e8890d9)


bneradt (approved):

bneradt:
[approve ci fedora]

PR #11966, opened 2025-01-16
9.2.x backport autests (#11601) and (#11512)

ezelkow1:
9.2.x backport autests (#11601) and (#11512)

PR #11967, opened 2025-01-16
Fix quick_server autest to verify server response (#11512)

bneradt:
The quick_server.test.py autest incorrectly sent a literal '\r\n' four bytes instead of the expected 2 bytes. The test still largely functioned as needed, but didn't verify that the expected respone made it to the client. This updates the test to verify that the 200 OK response makes it to the client as expected. ATS behaves as expected, so there is no production change needed.

(cherry picked from commit 168e23ac22b9dd614aeb009d78a0e2e2f6a66b7b)


bneradt:
Replaced by:
https://github.com/apache/trafficserver/pull/11966

PR #11975, opened 2025-01-20
[compress] Add range request control option

masaori335:
Prior to the change, the `range-request` config controls compress plugin to skip compression if `Range` request. However, this makes unexpected behavior once the object is cached without `Range` request. 

~To fix the case, this removes `Range` header if client send both of `Accept-Encoding` and `Range`  header.~

Update: add new options to choose `Accept-Encoding` header or `Range` header if a client sends both headers.

- `ignore-range` (or `true`) : Remove `Range` header
- `no-compression`: Remove `Accept-Encoding` header
- `none` (or `false`): do nothing

### DIFF
--- a/plugins/compress/compress.cc
+++ b/plugins/compress/compress.cc
@@ -94,7 +94,7 @@ handle_range_request(TSMBuffer req_buf, TSMLoc req_loc, HostConfiguration *hc)
     return;
   }
 
-  TSMLoc range_hdr_field = TSMimeHdrFieldFind(req_buf, req_loc, TS_MIME_FIELD_RANGE, TS_MIME_LEN_RANGE);
+  TSMLoc         range_hdr_field = TSMimeHdrFieldFind(req_buf, req_loc, TS_MIME_FIELD_RANGE, TS_MIME_LEN_RANGE);
   ts::PostScript range_defer([&]() -> void { TSHandleMLocRelease(req_buf, req_loc, range_hdr_field); });
   if (range_hdr_field == TS_NULL_MLOC) {
     return;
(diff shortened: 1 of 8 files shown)


masaori335:
Update: add a new option to remove `Accept-Encoding` header. Now, we can choose `Accept-Encoding` header or `Range` header if a client sends both headers.

masaori335:
[approve ci autest 3of4]

plugins/compress/compress.cc
```diff
@@ -76,6 +77,57 @@ static TSMutex compress_config_mutex = nullptr;
...
+{
+  TSMLoc accept_encoding_hdr_field =
+    TSMimeHdrFieldFind(req_buf, req_loc, TS_MIME_FIELD_ACCEPT_ENCODING, TS_MIME_LEN_ACCEPT_ENCODING);
+  if (accept_encoding_hdr_field == TS_NULL_MLOC) {
+    return;
+  }
+
+  TSMLoc range_hdr_field = TSMimeHdrFieldFind(req_buf, req_loc, TS_MIME_FIELD_RANGE, TS_MIME_LEN_RANGE);
+  if (range_hdr_field == TS_NULL_MLOC) {
+    return;
```
bryancall, line 96: don't you need to release the accept_encoding_hdr_field before returning?

bryancall (changes requested):

plugins/compress/compress.cc
masaori335: Thanks for remind me that! Will do.

masaori335: Done. Please take another look.

bryancall (approved):

PR #11978, opened 2025-01-22
Codeql updates

ezelkow1:
Need to update the actions to newer versions and include cmake and pcre updates

It's possible this needs more updates but these are a first check


masaori335 (approved):
Thanks for update.

PR #11984, opened 2025-01-24
Cleanup of iocore

moonchen:
**For iocore source files and header files:**

* Make headers self-contained.

Header files should work correctly regardless of the context of where it is included.  This means the header should #include all of its dependencies.

* Remove unused #includes.

Source and header files should only #include headers that are required. This reduces compile time, expresses clear and explicit dependencies, reduces namespace pollution, enhances modularity, and decouples components.

* Break up aggregate headers.

Some header files used to include a number of other header files for the sole purpose of packaging them into one #include directive.  This increases compilation times, creates tight coupling and hidden dependencies, and introduces namespae pollution.  Furthermore, this creates potential for circular dependencies.

* Fix linter warnings

Use C++ style casts.  Pass by reference when appropriate.

**What this change is not:**

This does not get us all the way to "Include What You Use" (IWYU), although it is a step in that direction.

This change does not address the issue of "private (P_*)" headers.

This change only targets iocore.  Changes to other files are mainly for the purpose of fixing compilation.


include/iocore/cache/HttpTransactCache.h
```diff
@@ -30,8 +30,9 @@
  ****************************************************************************/
 #pragma once
 
-#include "tscore/ink_platform.h"
+#include "proxy/hdrs/HTTP.h"
```
masaori335, line 33: We made an effort to avoid including `proxy/` from `iocore/cache/` before. 
https://github.com/apache/trafficserver/commit/5932a30e3f10956abf9d88cf4e57ee28bc20f1f4

moonchen: I agree that we should avoid including `proxy/` from `iocore/`, but as it stands today, HTTPHdr is dereferenced in HttpTransactCache.cc, for example, [here](https://github.com/moonchen/trafficserver/blob/b7b16275d210577b08d90db870570f718b9eecc7/src/iocore/cache/HttpTransactCache.cc#L200).

masaori335: I guess `HttpTransactCache.cc` can include the `proxy/` headers by itself and this `HttpTransactCache.h` header file doesn't need to include?

masaori335: As talked at bug scrub, it looks like this just uncovered existing circular dependency. Let's fix it after shipping this.

bneradt (approved):
Looks good. 90% is include changes, and the rest seem pretty straightforward. I'm surprised by how many un-bracketed one line conditionals you had to address here. But I'm glad that's addressed now.

bneradt:
[approve ci autest 1]

bneradt:
[approve ci autest 1]

PR #11986, opened 2025-01-27
[compress] Make warning if compress plugin has cache and range-request config

masaori335:
As discussed on https://github.com/apache/trafficserver/pull/11927, combination of below config of compress plugin might cause an issue. This makes a warning if compress plugin is configured to use the combination.
```
cache false
range-request true
```


bneradt (approved):

cmcfarlen:
Cherry-picked to 10.0.0

PR #12000, opened 2025-01-29
Fix a build issue from iocore cleanup

moonchen:
The build of test_net was broken on Mac when the build mode is set to release.

**Root cause:**

We support two kinds of linkers.  One kind will only search for symbols that are to the left of the object being linked.  Another kind will search the list of libraries repeatedly until the no new undefined references are created.  Both are sensitive to the order of libraries on the command line.

In order to work with both, we have to carefully order our dependencies in cmake.  See the build rule for traffic_server for a working list of dependencies.

In this case, libinknet_stub.cc provides some symbols that are also provided by ts::proxy, in order to fix an issue of cyclic dependency between ts::inknet and ts::proxy.  However, this causes the link to have duplicate symbols when libraries are not ordered correctly, on linkers that repeat through the list.

**References:**

https://stackoverflow.com/questions/45135/why-does-the-order-in-which-libraries-are-linked-sometimes-cause-errors-in-gcc

### DIFF
--- a/src/iocore/net/CMakeLists.txt
+++ b/src/iocore/net/CMakeLists.txt
@@ -124,18 +124,32 @@ if(BUILD_TESTING)
   )
   if(CMAKE_LINK_GROUP_USING_RESCAN_SUPPORTED OR CMAKE_CXX_LINK_GROUP_USING_RESCAN_SUPPORTED)
     # Use link groups to solve circular dependency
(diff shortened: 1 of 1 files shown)


bneradt (approved):

maskit (approved):
Works for me.

PR #12007, opened 2025-01-31
Adding a 103 Early Hints Autest

bneradt:
This adds an autest to cover 103 early hints behavior. The test passes and found no issues.


bneradt:
[approve ci autest 3]

maskit (approved):
Thanks for adding tests.

Nothing is wrong with this change, but we might not want to forward 103 response on H1.

> Note: For compatibility and security reasons, it is recommended to [only send HTTP 103 Early Hints responses over HTTP/2 or later](https://www.rfc-editor.org/rfc/rfc8297#section-3) unless the client is known to handle informational responses correctly.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/103

maskit:
I created a separate issue about status 103 on H1.
https://github.com/apache/trafficserver/issues/12012

PR #12009, opened 2025-01-31
Adds the %{GROUP} condition to HRW, and else clause

zwoop:
The syntax may not be beautiful, but it's inline with how HRW works today, and this solves a significant problem with lack of ()'s in the syntax.

Note that there's an implicit GROUP  / () around the outermost conditions. This has no impact on performance, but it moves some of the evaluation logic into the `ConditionGroup` class.

This also adds an optional else clause to the conditions and operators.

### DIFF
--- a/plugins/header_rewrite/conditions.h
+++ b/plugins/header_rewrite/conditions.h
@@ -659,3 +659,68 @@ protected:
 private:
   TSHttpCntlType _http_cntl_qual = TS_HTTP_CNTL_LOGGING_MODE;
 };
+
+class ConditionGroup : public Condition
+{
+public:
+  ConditionGroup() { Dbg(dbg_ctl, "Calling CTOR for ConditionGroup"); }
+
+  ~ConditionGroup() override
+  {
+    Dbg(dbg_ctl, "Calling DTOR for ConditionGroup");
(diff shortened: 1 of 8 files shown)


ezelkow1 (approved):

ezelkow1 (dismissed):

plugins/header_rewrite/conditions.h
```diff
@@ -659,3 +659,68 @@ class ConditionHttpCntl : public Condition
...
+  bool
+  closes() const
+  {
+    return _end;
+  }
+
+  void
+  append_value(std::string & /* s ATS_UNUSED */, const Resources & /* res ATS_UNUSED */) override
+  {
+    TSReleaseAssert(!"%{GROUP} should never be used as a condition value!");
```
cmcfarlen, line 697: Should this release assert?  Maybe an error instead?

zwoop: > Should this release assert? Maybe an error instead?

Changed to a debug release, and a TSError. Note that this is a runtime "error", so it won't fail configuration loading / reloading here. That'd require a different change / refactoring I think.

ezelkow1 (approved):

PR #12019, opened 2025-02-07
install libswoc with versioned filenames

cmcfarlen:
This installs the libswoc shared libraries with the versions and typical symlinks.  Before this was only installing a non-standard file `libswoc-1.5.12.so`

```
libswoc.1.5.12.so
libswoc.1.so -> libswoc.1.5.12.so
libswoc.so -> libswoc.1.so
```


bneradt:
[approve ci autest 1]

bneradt (approved):

JosiahWI (approved):

PR #12024, opened 2025-02-08
Use ats_free instead of free in LogFlushData destructor

hnakamur:
When m_logfile->m_file_format in LogData is LOG_FILE_ASCII or LOG_FILE_PIPE, m_data in LogData is allocated with ats_malloc in LogFile::write_ascii_logbuffer3.


cmcfarlen:
Cherry-picked to v10.0.x

PR #12026, opened 2025-02-11
Static link opentelemetry-cpp libraries to otel_tracer plugin

hnakamur:
Without this, `cmake --build build --target test` causes an error that failes to load otel_tracer.so because of undefined symbol `_ZN6google8protobuf7Message19CopyWithSourceCheckERS1_RKS1_`.

Also update instructions for building and installing dependency libraries using CMake.

Also update opentelemetry-cpp version to v1.19.0.

### DIFF
--- a/cmake/Findopentelemetry.cmake
+++ b/cmake/Findopentelemetry.cmake
@@ -20,57 +20,240 @@
 # This will define the following variables
 #
 #     opentelemetry_FOUND
-#     opentelemetry_LIBRARY
+#     opentelemetry_common_LIBRARY
+#     opentelemetry_exporter_in_memory_LIBRARY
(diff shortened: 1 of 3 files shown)


hnakamur:
Could you wait for reviewing this pull request?
I found this change causes an error in tests on macOS. I am working to fix it.

hnakamur:
I think it's ready to be reviewd now.
I verified the `verify_global_otel_tracer` test passes on the following environments:

* Ubuntu 24.04, CMake 3.28.3
* Ubuntu 22.04, CMake 3.22.1
* macOS 15.3.1, CMake 3.31.5
* FreeBSD 14.2, CMake 3.31.1

shukitchan:
[approve ci autest 1]

cmcfarlen (approved):

PR #12031, opened 2025-02-11
Add support for Unix Domain Socket server

cmcfarlen:
Initial support for listening for connections on unix domain sockets.

If used with proxy protocol v2, much of the functionality (including ip acts) can be maintained
without proxy protocol, ip acls are replaced with a match-all acl.

If the exec_thread.listen feature is enabled, then each thread will listen on a UDS with the path suffixed with the exec thread id.


include/tscore/ink_inet.h
```diff
@@ -295,6 +312,26 @@ ats_ip_are_compatible(sockaddr const *lhs, ///< Address to test.
   return lhs->sa_family == rhs;
 }
 
+/// @return the path length including the \0
+inline int
+ats_unix_path_len(sockaddr_un *s)
+{
+  return strnlen(s->sun_path, TS_UNIX_SIZE);
+}
```
bneradt, line 320: I don't think `strnlen` includes the null terminator in its count.

bneradt (changes requested):

cmcfarlen:
[approve ci autest 1]

bneradt (approved):
lgtm

davidbirdsong:
Can anyone comment on how or when this would get included in a release? I can't find any info on the release process. Recent release tags don't have this as an ancestor--unless my git fu is wrong.

PR #12032, opened 2025-02-12
Add TSVConnPPInfoGet

maskit:
https://lists.apache.org/thread/nzr39cdvf7277crkf0gf82b7201qxzb8

The API is only tested with PROXY protocol version 1 that doesn't support TLV. I could use Proxy Verifier to use PROXY protocol version 2, but it doesn't seem like to support sending TLV. The PROXY protocol parser itself is tested by unit tests.


bryancall:
[approve ci autest 2]

include/ts/ts.h
```diff
@@ -3077,3 +3077,28 @@ TSRalloc(size_t count = 1 /**< Number of instances of T to allocate storage for.
...
+   @return @c TS_SUCCESS if the requested info is supported, TS_ERROR otherwise
+
+*/
+TSReturnCode TSVConnPPInfoGet(TSVConn vconn, uint16_t key, const char **value, int *length);
+
+/**
+   Return the particular PROXY protocol info requested.
+
+   @param vconn the vconection pointer
+   @param key the requested PROXY protocol info. One of TSVConnPPInfoKey or TLV type ID
```
bneradt, line 3098: This one has to be a `TSVConnPPInfoKey` value, right? Should it take a `TSVConnPPInfoKey` type rather than a `uint16_t`?

bneradt (changes requested):

include/ts/ts.h
maskit: No, because it can be a non-standard value.

https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt
```2.2.8. Reserved type ranges

The following range of 16 type values is reserved for application-specific
data and will be never used by the PROXY Protocol. If you need more values
consider extending the range with a type field in your TLVs.

        #define PP2_TYPE_MIN_CUSTOM    0xE0
        #define PP2_TYPE_MAX_CUSTOM    0xEF

This range of 8 values is reserved for temporary experimental use by
application developers and protocol designers. The values from the range will
never be used by the PROXY protocol and should not be used by production
functionality.

        #define PP2_TYPE_MIN_EXPERIMENT 0xF0
        #define PP2_TYPE_MAX_EXPERIMENT 0xF7

The following range of 8 values is reserved for future use, potentially to
extend the protocol with multibyte type values.

        #define PP2_TYPE_MIN_FUTURE    0xF8
        #define PP2_TYPE_MAX_FUTURE    0xFF
```

bneradt (approved):

cmcfarlen:

PR #12033, opened 2025-02-12
Update the cmake hwloc processing unit support test to use the hwloc include directory

bryancall:


phongn (approved):

JosiahWI:
Do we not need to provide the `hwloc` library path? Does the check not have to perform the linking step?

JosiahWI (approved):
We do not need to link libhwloc for the check, because it doesn't depend on any symbols other than the define that would be in the hwloc header.

PR #12039, opened 2025-02-17
access_control plugin: Generate Session Cookie by "exp=0"

masaori335:
Add generating Session Cookie support to the Access Control plugin. It's triggered by setting `exp=0` in the `TokenRespHdr` header in the response from origin server.

https://docs.trafficserver.apache.org/en/10.0.x/admin-guide/plugins/access_control.en.html

### DIFF
--- a/tests/gold_tests/pluginTest/access_control/replays/access_control.replay.yaml
+++ b/tests/gold_tests/pluginTest/access_control/replays/access_control.replay.yaml
@@ -1,0 +1,75 @@
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+sessions:
+- protocol: [{ name: tls, sni: test }]
+
+  transactions:
+  # 1: Regular request
+  # ```
+  # payload='sub=frogs-in-a-well&exp=1577836800&nbf=1514764800&iat=1514160000&tid=1234567890&kid=key1&st=HMAC-SHA-256&md='
+  # signature=$(echo -n ${payload} | openssl dgst -sha256 -hmac "PEIFtmunx9")
+  # TokenRespHdr=${payload}${signature}
+  # ```
+  - client-request:
+      method: "GET"
+      version: "1.1"
+      url: /test-1/
+      headers:
+        fields:
+          - [ uuid, 1]
+          - [ Host, example.com ]
+
+    server-response:
+      status: 200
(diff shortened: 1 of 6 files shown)


tests/gold_tests/pluginTest/access_control/replays/access_control.replay.yaml
```diff
@@ -0,0 +1,75 @@
...
+      headers:
+        fields:
+          - [ TokenRespHdr, sub=frogs-in-a-well&exp=0&nbf=1514764800&iat=1514160000&tid=1234567890&kid=key1&st=HMAC-SHA-256&md=e67dce2fcfaf169dde0fe0e6ecf14bea7658bffbe3f1b30c22b048432e9cc41f ]
+          - [ Content-Length, 1024 ]
+
+    proxy-response:
+      status: 200
+      headers:
+        fields:
+          - [ set-cookie, { value: "TokenCookie=c3ViPWZyb2dzLWluLWEtd2VsbCZleHA9MCZuYmY9MTUxNDc2NDgwMCZpYXQ9MTUxNDE2MDAwMCZ0aWQ9MTIzNDU2Nzg5MCZraWQ9a2V5MSZzdD1ITUFDLVNIQS0yNTYmbWQ9ZTY3ZGNlMmZjZmFmMTY5ZGRlMGZlMGU2ZWNmMTRiZWE3NjU4YmZmYmUzZjFiMzBjMjJiMDQ4NDMyZTljYzQxZg; path=/; Secure; HttpOnly", as: equal } ]
```
masaori335, line 75: Does proxy-verifier has support of "NOT contains" check? Here, I need to check absence of the `Expires=` attribute.

brbzull0: There is a `include` feature available so so wonder if using `not` you could do:

```yaml
  - [ Set-Cookie, { value: [Expires] , not: includes } ]
```
Using `[]` as per the examples, not sure if for a single value we can just use the string.

From the docs


`includes`:

``` 
Helpful for multi-value fields. Specifies that the set of value strings exist in the header field
values for a particular header name in any order. 
Note that each value is matched against the set like ``contains``, so each value is a substring match
```


`not`

```
Proxy Verifier supports inverting the result of any rule by using not instead of as. 
```

masaori335: Thank you! I didn't noticed these. I'll try it.

masaori335: Somehow, my verifier-client doesn't like the `includes` directive. Proxy-Verifier version is `2.12.0`.
```
    [INFO]: No meta node ("meta") at "/Users/masaori/src/github.com/apache/trafficserver-asf-master-work-6/tests/_sandbox/access_control/verifier-client/access_control.replay.yaml":16.
            [ERROR]: Failed to parse response at line 71
              [ERROR]: Field rule at line 74 has an invalid directive (includes).
```              

In this case, `{ value: Expires=, not: contains }` seems work ( I checked it failed if the header has the `Expires=` attribute ).

masaori335:
[approve ci debian]

masaori335:
[approve ci]

tests/gold_tests/pluginTest/access_control/replays/access_control.replay.yaml
brbzull0: it looks like we may need another release of PV, the `includes`  [commit](https://github.com/yahoo/proxy-verifier/commit/17dd57a3fd91fff0323a6d8a625b180854a64a8e) is not part of any tag, just master



cc: @bneradt

brbzull0 (approved):
looks good. thanks.

PR #12050, opened 2025-02-20
Fix up logging for write-fail-5/collapsed-forwarding

ezelkow1:
The logic here changed so that it no longer loops for all write retries before retrying to the read. So the logging no longer worked correctly. Changing so that as long as it has attempted a write, but wound up with some sort of HIT, then that should be a CF success since it jumped back to the read state to get that


ezelkow1:
[approve ci autest 1]

bneradt (approved):

cmcfarlen:
Cherry-picked to v10.0.x

PR #12059, opened 2025-02-28
Cleanup: Make Cache functions const

masaori335:
I found that many `struct Cache` functions don't touch member variables.


brbzull0 (approved):
Make sense. Thanks

PR #12063, opened 2025-02-28
Clarifies that parent.config is to be replaced with strategies

mlibbey:


jrushford:
Hi Evan, I would think so.  Withe the next 10 release announcement, include that parent. Config is deprecated.JohnOn Feb 28, 2025, at 4:27 PM, Evan Zelkowitz ***@***.***> wrote:﻿
  Should we put in a version in there so people known with 10.1 its deprecated and 'dead' in 11? other than that 👍—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you are subscribed to this thread.Message ID: ***@***.***>



ezelkow1 left a comment (apache/trafficserver#12063)
Should we put in a version in there so people known with 10.1 its deprecated and 'dead' in 11? other than that 👍

bryancall (approved):

PR #12065, opened 2025-03-03
Fix memory leak in ink_event_system_init

hnakamur:
The memory for chunk_sizes_string is allocated in
REC_ConfigReadString
RecGetRecordString_Xmalloc
RecGetRecord_Xmalloc
RecDataSet
ats_strdup
_xstrdup
ats_malloc

So it must be freed with ats_free.


cmcfarlen (approved):
Good catch!

PR #12067, opened 2025-03-03
Use ip command as a fallback in redirect_actions test

hnakamur:
Without this workaround, if the hostname is set to a loopback address redirect_actions test fails like below:

```
       file /home/jenkins/autest_work/sandbox/redirect_actions/_output/9-tr-Default/stream.stdout.txt : Checking that /home/jenkins/autest_work/sandbox/redirect_actions/_output/9-tr-Default/stream.stdout.txt matches /home/jenkins/trafficserver/tests/gold_tests/redirect/generated_test_data/9-tr.gold - Failed
          Reason: File differences
           Gold File : /home/jenkins/trafficserver/tests/gold_tests/redirect/generated_test_data/9-tr.gold
           Data File : /home/jenkins/autest_work/sandbox/redirect_actions/_output/9-tr-Default/stream.stdout.txt
           - HTTP/1.1 307 Temporary Redirect
           + HTTP/1.1 204 No Content
```

This error happens since the address is a loopback so it is not rejected as expected.

As a workaround, use the ip command to obtain a non-loopback IP address.

### DIFF
--- a/tests/gold_tests/redirect/redirect_actions.test.py
+++ b/tests/gold_tests/redirect/redirect_actions.test.py
@@ -53,6 +53,24 @@ try:
 except socket.gaierror:
     pass
 
+# If the hostname is set to a loopback address, use the ip command to obtain a non-loopback IP address.
+if any(ip.startswith('127.') for ip in ipv4addrs):
+    import json
+    from subprocess import check_output
+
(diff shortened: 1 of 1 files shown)


masaori335 (approved):
As a workaround, this looks good. Thanks.

PR #12072, opened 2025-03-04
Fix send 100 Continue optimization for GET

bryancall:
This fixes a bug with the proxy.config.http.send_100_continue_response feature for GET requests in which the following would happen:

1. We do not send the optimized 100 Continue out of proxy for GET requests with Exect: 100-Continue. This is reasonable since the vast majority of 100-Continue requests will be POST.
2. Later, we inspect the proxy.config.http.send_100_continue_response value and assume we did send a 100 Continue response and stripped the Expect: 100-Continue header from the proxied request. This is disasterous as it leaves the server waiting for a body which would never come because the client is waiting for a 100 Continue response which will never come.


bryancall:
[approve ci]

cmcfarlen (approved):

PR #12036, opened 2025-02-13
Create directory before chown on install (#10954)

hnakamur:

PR #12085, opened 2025-03-05
Move defaulting install prefix before layout setup

cmcfarlen:
If a build config defaults CMAKE_INSTALL_PREFIX and then installs as the root user, the build will have an error about LOGDIR not existing.  This PR fixes that by making sure that CMAKE_INSTALL_PREFIX is set to the default value before the layout cmake script sets up the install paths according to the layout.

replaces #12036


cmcfarlen:

Issue #12047, opened 2025-02-20
performance problem met in range request with 304 update scenario, any suggestions ?


wangchunpeng:
In ATS, we use the Range_request plugin and the Slice plugin to handle large file requests.  when the fragment size is set 1MB and slice range is also set 1MB，each http sub-request generates only one single document（including doc，header_info and bodydata).  When a file expires and an IMS (If-Modified-Since) request is made to the origin server, and the server responds with a 304 (Not Modified), ATS generates a separate document to store the updated header information. This results in the creation of many small files, each around a few kilobytes in size. When a subsequent request is made, it triggers two IO read operations, which reduces performance. Are there any optimization suggestions for this scenario ?

moonchen:
Hi @wangchunpeng,

As I understand it, ATS's behavior when handling a 304 Not Modified response from the origin is mostly the same whether using the Slice plugin or not. In both cases:

-     A new metadata record is created for a "header update," only one per URL regardless of slicing.
-     A subsequent client request requires two I/O reads per slice: one for the metadata and one for the body data.

Are you seeing I/O overhead that is different for slicing, specifically when handling 304?

traeak:
I think what's happening is that each slice of that asset is having to handle and store the 304 response.  In the general case I'm not sure how to get around that since individual slices of the same asset name might have different content from multiple different versions of the named asset.

A possible solution:

Use a reference slice.  That should always have the most up to date etag/last-modified "variant identifier" (or whatever it's called).

Pass this reference slice etag/last-modified "variant identifier" to subsequent calls of the CRR plugin.  If the CRR plugin encounters a STALE in-cache result (from cache read hook) but the variant identifier matches, switch to FRESH and continue.  This should stop the CRR plugin from going to parent and also should result in having a variant header being written to cache.  I think.

PR #12092, opened 2025-03-07
slice/cache_range_requests plugin: avoid subsequent slice IMS requests

traeak:
relates to issue #12047

With the slice plugin when an asset goes stale in cache and an IMS request results in a 304, the slice plugin may end up causing each range request to perform an IMS request, get a 304 and write another header into the cache for each subsequent range.

To get around this the slice plugin will take the strongest identifier from the first range request response and attach an identifier header to the cache_range_requests plugin.  If during cache lookup the CRR plugin finds that the resulting asset is stale, it will check the provided identifier against the "stale" slice and will flip it to "fresh" if the identifiers match.

The identifier will be passed as header, with Etag being preferred over Last-Modified.
X-Crr-Ident: Etag: "etag"
or
X-Crr-Ident: Last-Modified: "last modified date"

It might be worthwhile to consider reworking the slice self healing functionality to use this new header instead of the X-Crr-Ims header -- will file a separate PR for this.

I enabled this flag in the slice_selfhealing autest just to ensure it doesn't interfere.
Also found a memory leak in slice.cc regarding the large asset check code path and added extra debug for header rewrite token parsing failures.

### DIFF
--- a/doc/admin-guide/plugins/cache_range_requests.en.rst
+++ b/doc/admin-guide/plugins/cache_range_requests.en.rst
@@ -112,7 +112,6 @@ which includes information about the partial content range.  In this mode,
 all requests (include partial content) will use consistent hashing method
 for parent selection.
 
-
 X-Crr-Ims header support
 ------------------------
 
@@ -143,6 +142,35 @@ When used with the :program:`slice` plugin its `--crr-ims-header`
 option must have the same value (or not be defined) in order to work.
 
 Presence of the `--ims-header` automatically sets the `--consider-ims` option.
+
+X-Crr-Ident header support
+--------------------------
+
+.. option:: --consider-ident
+.. option:: -d
+.. option:: --ident-header=[header name] (default: X-Crr-Ident)
+.. option:: -j
+
+This supports the slice plugin which makes multiple adjancent
+range requests.  The slice plugin will record the identifier
+of the first range request (Etag, or Last-Modified in that order)
+and will add the value of that to the this header.
+
+.. code::
+
+    X-Crr-Ident: Etag: "foo"
+    X-Crr-Ident: Last-Modified: Tue, 19 Nov 2019 13:26:45 GMT
+
+If during the cache lookup hook a range request is considered stale
+the identifier from this header will be compared to the stale cache
+response and if matches the response will be marked FRESH to bypass
+visiting a parent cache with an IMS request.
+
+When used with the :program:`slice` plugin its `--crr-ident-header`
+option must have the same value (or not be defined) in order to work.
+
+Presence of the `--ident-header` automatically sets the `--consider-ident`
+option.
 
 Don't modify the Cache Key
 --------------------------
--- a/doc/admin-guide/plugins/cache_range_requests.en.rst
+++ b/doc/admin-guide/plugins/cache_range_requests.en.rst
@@ -151,20 +151,20 @@ X-Crr-Ident header support
 .. option:: --ident-header=[header name] (default: X-Crr-Ident)
 .. option:: -j
 
-This supports the slice plugin which makes multiple adjancent
-range requests.  The slice plugin will record the identifier
-of the first range request (Etag, or Last-Modified in that order)
-and will add the value of that to the this header.
+This supports the slice plugin which makes multiple adjacent range
+requests. The slice plugin will record the identifier of the first range
+request (Etag, or Last-Modified in that order) and will add the value
+to this header.
 
 .. code::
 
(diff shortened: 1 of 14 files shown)


bryancall:
[approve ci autest 1]

plugins/cache_range_requests/cache_range_requests.cc
```diff
@@ -605,6 +642,60 @@ handle_cache_lookup_complete(TSHttpTxn txnp, txndata *const txn_state)
...
+              if (!tag.empty()) {
+                TSMLoc const id_loc = TSMimeHdrFieldFind(resp_buf, resp_loc, tag.data(), tag.size());
+                if (TS_NULL_MLOC != id_loc) {
+                  int                    len = 0;
+                  char const *const      str = TSMimeHdrFieldValueStringGet(resp_buf, resp_loc, id_loc, 0, &len);
+                  std::string_view const sv(str, len);
+
+                  DEBUG_LOG("Checking cached '%.*s' against request '%.*s'", len, str, (int)svreq.size(), svreq.data());
+
+                  if (std::string_view::npos != svreq.rfind(sv)) {
```
serrislew, line 684: How does this check if the response is a MISS? Is it possible that the first block is a MISS but the second block is STALE and we use the ident header so we don't call parent or origin server since we mark it as FRESH?

doc/admin-guide/plugins/cache_range_requests.en.rst
```diff
@@ -144,6 +143,35 @@ option must have the same value (or not be defined) in order to work.
...
+
+.. code::
+
+    X-Crr-Ident: Etag: "foo"
+    X-Crr-Ident: Last-Modified: Tue, 19 Nov 2019 13:26:45 GMT
+
+If during the cache lookup hook a range request is considered stale
+the identifier from this header will be compared to the stale cache
+response and if matches the response will be marked FRESH to bypass
+visiting a parent cache with an IMS request.
```
serrislew, line 167: is it worth mentioning that this doesn't apply if the ims header is configured? we only check the identity header if we're not forced revalidating

tests/gold_tests/pluginTest/slice/slice_ident.test.py
```diff
@@ -0,0 +1,188 @@
...
+    'proxy.config.diags.debug.tags': 'slice',
+})
+
+ts.Disk.remap_config.AddLines(
+    [
+        f"map http://preload/ http://127.0.0.1:{server.Variables.Port}",
+        f'map http://slice/ http://127.0.0.1:{server.Variables.Port}' +
+        f' @plugin=slice.so @pparam=--blockbytes-test={block_bytes}',
+        f'map http://slicecustom/ http://127.0.0.1:{server.Variables.Port}' +
+        f' @plugin=slice.so @pparam=--blockbytes-test={block_bytes}' + ' @pparam=--crr-ident-header=CrrIdent',
```
serrislew, line 104: I think separate autests is good to isolate the different functionalities. Do you think it would be helpful to also have an autest that uses both CRR and slice plugin for end to end testing?

serrislew:
Though I think this config option is helpful and can reduce subsequent slice IMS requests, I'm not sure if this solves the issue tagged. Their use case doesn't seem to be slice related but also difficult to assume without understanding their situation more.

plugins/cache_range_requests/cache_range_requests.cc
traeak: the first requested slice will establish the validator.  so if the first slice ends up vacating out of cache but a subsequent slice is cached but stale this should still work as long as the first slice validator matches the stale slice's validator.

doc/admin-guide/plugins/cache_range_requests.en.rst
```diff
@@ -144,6 +143,35 @@ option must have the same value (or not be defined) in order to work.
...
+
+.. option:: --consider-ident
+.. option:: -d
+.. option:: --ident-header=[header name] (default: X-Crr-Ident)
+.. option:: -j
+
+This supports the slice plugin which makes multiple adjancent
+range requests.  The slice plugin will record the identifier
+of the first range request (Etag, or Last-Modified in that order)
+and will add the value of that to the this header.
```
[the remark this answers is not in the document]
traeak: changed and reworded.

traeak:
I agree that this doesn't directly fix their issue, but it will help prevent the situation from ever happening. I don't think the issue can truly be fixed without a more extensive redesign of the slice plugin.

traeak:
[approve ci autest 1]

serrislew (approved):

PR #12093, opened 2025-03-07
Fix up logging for write-fail-5/collapsed-forwarding (#12050)

ezelkow1:
The logic here changed so that it no longer loops for all write retries before retrying to the read. So the logging no longer worked correctly

(cherry picked from commit c82122e743258d26f05874330959f23d56d64fec)


cmcfarlen (approved):

ezelkow1:
[approve ci autest 2]

PR #12095, opened 2025-03-10
Fix a few places where inet_ntop could fail

cmcfarlen:


shukitchan:
is that because now we support unix domain socket?

shukitchan (approved):

cmcfarlen:

PR #12096, opened 2025-03-11
Slice plugin support UDS path

serrislew:
Support UDS path as client "ip" in slice plugin


traeak (approved):
These changes look good and should now happily handle clients coming in from a UDS.

cmcfarlen:

PR #12099, opened 2025-03-14
Fix a memory leak in the main function in traffic_server.cc

hnakamur:


cmcfarlen:

PR #12107, opened 2025-03-18
Fix hii log field for UDS

### LINKED ISSUE #12106: Log field hii can cause a crash if UDS is used
~Log field `chp` is for "Port number of the client’s host", and its type is integer. If UDS is used, the actual value is a path for UDS, not a number. This mismatch causes log buffer corruption and can cause a crash when ATS tries to unmarshal the buffer for following fields.~

----
Update: The problematic field was `hii`. The unmarshal function does not support UDS.

maskit:
This fixes #12106.


bneradt (approved):

cmcfarlen:

PR #12110, opened 2025-03-19
Add server_cipher_suite to sni.yaml

bneradt:
Allow the override of records.yaml configurations:

* proxy.config.ssl.server.cipher_suite
* proxy.config.ssl.server.TLSv1_3.cipher_suites

By the following new sni.yaml configurations:

* server_cipher_suite
* server_TLSv1_3_cipher_suites.

### DIFF
--- a/src/iocore/net/YamlSNIConfig.cc
+++ b/src/iocore/net/YamlSNIConfig.cc
@@ -155,6 +155,12 @@ YamlSNIConfig::Item::populate_sni_actions(action_vector_t &actions)
   if (!client_sni_policy.empty()) {
     actions.push_back(std::make_unique<OutboundSNIPolicy>(client_sni_policy));
   }
+  if (!server_cipher_suite.empty()) {
+    actions.push_back(std::make_unique<ServerCipherSuite>(server_cipher_suite));
+  }
+  if (!server_TLSv1_3_cipher_suites.empty()) {
+    actions.push_back(std::make_unique<ServerTLSv1_3CipherSuites>(server_TLSv1_3_cipher_suites));
+  }
   if (http2_buffer_water_mark.has_value()) {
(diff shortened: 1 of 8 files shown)


src/iocore/net/YamlSNIConfig.cc
```diff
@@ -442,6 +452,12 @@ template <> struct convert<YamlSNIConfig::Item> {
     if (node[TS_client_sni_policy]) {
       item.client_sni_policy = node[TS_client_sni_policy].as<std::string>();
     }
+    if (node[TS_server_cipher_suite]) {
+      item.server_cipher_suite = node[TS_server_cipher_suite].as<std::string>();
+    }
+    if (node[TS_server_TLSv1_3_cipher_suites]) {
```
maskit, line 458: Don't we need to check `TS_USE_TLS_SET_CIPHERSUITES`?

bneradt: Huh. That's a good point. I followed the lead of TS_USE_HELLO_CB for `TS_valid_tls_versions_in`, etc., which `#if` around the declarations, but they also don't check around the yaml nodes like this either.

bneradt: Updated. Thanks for catching this.

maskit: I thought this would cause a compile error but I was wrong. `TSDECL` declares the name regardless of the availability, so we can have this without the check.

bneradt: Ah, right. It's declared, just not in the `valid_sni_config_keys` list. OK. Taking the checks back out.

maskit (approved):

cmcfarlen:

PR #12111, opened 2025-03-20
1 create a docker compose file to test and build the application

makarkul:


makarkul:
Commit docker files for testing ATS

cmcfarlen (changes requested):
Thank you for contributing this docker-compose setup!

Could you please move this under the `contrib/docker` directory?

Also, I don't think it makes sense to build the 9.2.0 version by default from the master branch, since we are well into the 10.x lifecycle now.  I also see a records.yaml file here, so perhaps the Dockerfile just needs an update?  I suggest allowing the version to build be passed in as an argument.

kulkarnima:
> Thank you for contributing this docker-compose setup!
> 
> Could you please move this under the `contrib/docker` directory?
> 
> Also, I don't think it makes sense to build the 9.2.0 version by default from the master branch, since we are well into the 10.x lifecycle now. I also see a records.yaml file here, so perhaps the Dockerfile just needs an update? I suggest allowing the version to build be passed in as an argument.

Sure, will do the required change and push

bryancall:
@makarkul Can you please make the change to move these files into the directory specified above.  Thank you.

makarkul:
@bryancall @cmcfarlen - it appears that I am not in a position to complete the work on this pull request. You may want to just discard or delete this. I would need some bandwidth to fix all the issues pointed out by @cmcfarlen which I currently lack

Issue #12090, opened 2025-03-06
s3_auth crashes when using plug pparam options instead of a config file


ezelkow1:
On many of our services we pass in options to the s3_auth plugin as plugin parameters rather than as a config file. It appears that one of the 2 latest changes, either logging improvements or the config reloading, now causes the plugin to crash on startup when not using a config

ezelkow1:
Reverted
https://github.com/apache/trafficserver/commit/4194d983981d95384cf75e2f87b219da850429f5
https://github.com/apache/trafficserver/pull/11914

and that seems to have gotten rid of the crashes here

ezelkow1:
fixed in #12115

PR #12115, opened 2025-03-21
Fix logging for origin_server_auth

jasmine-nahrain:
This fixes issue #12090.


ezelkow1 (approved):
tested here with the same config that was broken and it now loads correctly

cmcfarlen:

PR #12116, opened 2025-03-21
[compress] Fix compressible-status-code config

masaori335:
The [`compressible-status-code`](https://docs.trafficserver.apache.org/en/10.0.x/admin-guide/plugins/compress.en.html#compressible-status-code) config of compress plugin doesn't work as expected. Because the set of status code has default value but it's not cleared before insert. So setting like `compressible-status-code 200,304` doesn't update the set.

https://github.com/apache/trafficserver/blob/07888c0aa85523871a9adb39b11e1075cc0a4d62/plugins/compress/configuration.h#L155-L156


bryancall (approved):

cmcfarlen:

PR #12121, opened 2025-03-24
[compress] Add an option to do not compress partial object

masaori335:
During https://github.com/apache/trafficserver/pull/11975 test, we found that new configs doesn't work as expected if other plugins (like `cache_range_request`) run first and remove `Range` header from client request.

Also, @vmamidi wants to send both of `Range` and `Accept-Encoding` headers to allow origin server to return "partial content of compressed content".

This PR clarify name of current options and add a new option to do not compress partial object.

### DIFF
--- a/plugins/compress/compress.cc
+++ b/plugins/compress/compress.cc
@@ -103,8 +103,8 @@ handle_range_request(TSMBuffer req_buf, TSMLoc req_loc, HostConfiguration *hc)
   debug("Both of Accept-Encoding and Range header are found in the request");
 
   switch (hc->range_request_ctl()) {
-  case RangeRequestCtrl::IGNORE_RANGE: {
-    debug("Remove the Range header by ignore-range config");
+  case RangeRequestCtrl::REMOVE_RANGE: {
+    debug("Remove the Range header by remove-range config");
     while (range_hdr_field) {
       TSMLoc next_dup = TSMimeHdrFieldNextDup(req_buf, req_loc, range_hdr_field);
       TSMimeHdrFieldDestroy(req_buf, req_loc, range_hdr_field);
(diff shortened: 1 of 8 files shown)


plugins/compress/compress.cc
```diff
@@ -733,6 +737,26 @@ transformable(TSHttpTxn txnp, bool server, HostConfiguration *host_configuration
     return 0;
   }
 
+  // check Partial Object is transformable
+  if (host_configuration->range_request_ctl() == RangeRequestCtrl::NO_COMPRESSION) {
```
masaori335, line 741: This block restores code prior to the PR #11975 and add another check of `Content-Range` header in server response. Because `cache_range_request` might remove `Range` header from client request.
https://github.com/apache/trafficserver/pull/11975/files#diff-a50782f349c34e95c6c8643b834aaeba685f244f78a645e85a402bbda9012fb6L680-L689

plugins/compress/configuration.cc
```diff
@@ -270,16 +270,22 @@ HostConfiguration::compression_algorithms()
...
 HostConfiguration::set_range_request(const std::string &token)
 {
-  // "true" and "false" are compatibility with old version, will be removed
-  if (token == "false" || token == "ignore-range") {
-    range_request_ctl_ = RangeRequestCtrl::IGNORE_RANGE;
-  } else if (token == "true" || token == "none") {
+  if (token == "true" || token == "none") {
     range_request_ctl_ = RangeRequestCtrl::NONE;
-  } else if (token == "no-compression") {
+  } else if (token == "false" || token == "no-compression") {
```
bryancall, line 281: Should you still support `ignore-range` for compatibility?

bryancall (changes requested):

plugins/compress/configuration.cc
masaori335: It's introduced by https://github.com/apache/trafficserver/pull/11975 recently and not released yet. ( only master and 10.1.x branch have it ). So no compatibility issue yet.

masaori335:
Rebased on the latest master.

masaori335:
[approve ci autest 1of4]

bryancall (approved):

cmcfarlen:
This was added to milestone 10.1.x via #12258

PR #12129, opened 2025-03-25
esi: Add --cache-control-policy

bneradt:
This adds the --cache-control-policy configuration option to the esi
plugin, allowing the user to choose from a variety of behaviors with
respect to Cache-Control directives.

---

Converting to a draft while we consider whether this feature is needed.


bneradt:
[approve ci]

bneradt:
It looks like this won't be needed. I'll close this and open a new PR with just the autest updates.

Issue #12119, opened 2025-03-22
info address is not actual logged in hex even though it has prefix 0x in dbg_ctl_config debug log


hnakamur:
* https://github.com/apache/trafficserver/blob/07888c0aa85523871a9adb39b11e1075cc0a4d62/src/iocore/eventsystem/ConfigProcessor.cc#L91-L92
* https://github.com/apache/trafficserver/blob/07888c0aa85523871a9adb39b11e1075cc0a4d62/src/iocore/eventsystem/ConfigProcessor.cc#L142

I think we should use `PRIx64` instead of `PRId64`:

```
  Dbg(dbg_ctl_config, "Set for slot %d 0x%" PRIx64 " was 0x%" PRIx64 " with ref count %d", id, (int64_t)info, (int64_t)old_info,
      (old_info) ? old_info->refcount() : 0);
```

```
    Dbg(dbg_ctl_config, "Release config %d 0x%" PRIx64, id, (int64_t)info);
```

I suppose it is better to use `"%p"` and avoid casts:

```
  Dbg(dbg_ctl_config, "Set for slot %d %p was p with ref count %d", id, info, old_info,
      (old_info) ? old_info->refcount() : 0);
```

```
    Dbg(dbg_ctl_config, "Release config %d %p", id, info);
```

bryancall:
I agree they should be changed to %p.

JosiahWI:
Closed by #12131.

PR #12131, opened 2025-03-26
Use %p to print memory address in hex in ConfigProcessor debug log (#12119)

hnakamur:
Closes (#12119)

### DIFF
--- a/src/iocore/eventsystem/ConfigProcessor.cc
+++ b/src/iocore/eventsystem/ConfigProcessor.cc
@@ -88,8 +88,7 @@ ConfigProcessor::set(unsigned int id, ConfigInfo *info, unsigned timeout_secs)
   idx      = id - 1;
   old_info = infos[idx].exchange(info);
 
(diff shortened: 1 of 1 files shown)


JosiahWI (changes requested):
Looks great, thanks for fixing this! Can we also fix line 142 at the same time? It has the same issue.

hnakamur:
Oh, thank you for a good catch! I fixed it at [6c8d330](https://github.com/apache/trafficserver/pull/12131/commits/6c8d330b9cc96e3b2ddc73e16ac7b785ccd31728).

JosiahWI (approved):
Thanks! I noticed on cppreference that the format generated by %p is implementation defined, but if it is known that the format is reasonable on the mainstream Linux/BSD compilers (GCC, Clang) then I think this is a good solution.

hnakamur:
Thanks for sharing! I didn't know that.
I confirmed the explanation for `%p` is "Writes an implementation defined character sequence defining a pointer." at https://en.cppreference.com/w/cpp/io/c/fprintf

PR #12132, opened 2025-03-26
Doc updates to prepare for Sphinx 8

bneradt:
This fixes issues that Sphinx 8 highlights with our footnotes and references. This patch is in preparation of updating to Sphinx 8, which will be in a separate PR.

This patch addresses:

* The txn-box class-reference.en was referenced from multiple layers of the txn-box toc tree, which Sphinx 8 allerted to being a problem. Sphinx assumes a single parent for each entry in a toc. Before I assume it silently choose one over the other. It seems not needed in dev-guide to me, so I simply removed it from there.
* txn_box/misc.en.rst had an unreferenced footnote which this patch removes.
* cache-architecture/architecture.en.rst had two broken footnote references because there wasn't a space between the previous word and the footnote. Sphinx requires this space for the footnote to work.

### DIFF
--- a/doc/admin-guide/plugins/txn_box/dev/dev-guide.en.rst
+++ b/doc/admin-guide/plugins/txn_box/dev/dev-guide.en.rst
@@ -33,11 +33,3 @@ This section is for C++ developers working on or with |TxB|.
    memory-management.en
    dev-extractor.en
    dev-directive.en
-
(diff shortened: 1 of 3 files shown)


ezelkow1 (approved):

cmcfarlen:

PR #12138, opened 2025-03-28
Stabilize autest tls_sni_with_port

maskit:
Autest tls_sni_with_port sometimes fails because Proxy Verifier outputs slightly different log depends on how a request was received.
```
Received an HTTP/1 Content-Length body of 16 bytes for key conn_remapped
```
```
Received (with headers) an HTTP/1 body of 16 bytes for key conn_remapped with content
```
The regex works for the both.


JosiahWI (approved):
Looks good!

cmcfarlen:

PR #12141, opened 2025-03-28
Fix a bug that PPv2 parser fails to read TLV

maskit:
ATS fails to read PPv2 TLV fields if the head of HTTP request follows in the same buffer.

### DIFF
--- a/src/iocore/net/ProxyProtocol.cc
+++ b/src/iocore/net/ProxyProtocol.cc
@@ -291,7 +291,7 @@ proxy_protocol_v2_parse(ProxyProtocol *pp_info, const swoc::TextView &msg)
     }
 
     if (tlv_len > 0) {
-      if (pp_info->set_additional_data(msg.substr(msg.length() - tlv_len)) < 0) {
+      if (pp_info->set_additional_data(msg.substr(total_len - tlv_len, tlv_len)) < 0) {
         return 0;
       }
     }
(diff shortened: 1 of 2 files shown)


bneradt (approved):

bryancall:
@maskit Please update the docs: https://docs.trafficserver.apache.org/en/latest/admin-guide/configuration/proxy-protocol.en.html#proxy-protocol
"The current implementation doesn’t support TLV fields of Version 2."

src/iocore/net/ProxyProtocol.cc
```diff
@@ -291,7 +291,7 @@ proxy_protocol_v2_parse(ProxyProtocol *pp_info, const swoc::TextView &msg)
     }
 
     if (tlv_len > 0) {
-      if (pp_info->set_additional_data(msg.substr(msg.length() - tlv_len)) < 0) {
+      if (pp_info->set_additional_data(msg.substr(total_len - tlv_len, tlv_len)) < 0) {
```
bneradt, line 294: Are we sure that total_len > tlv_len?

maskit: I think these lines ensure that the calculated value is always valid.
```
const size_t total_len = PPv2_CONNECTION_HEADER_LEN + len;

if (msg.size() < total_len) {
    return 0;
}

tlv_len = len - PPv2_ADDR_LEN_INET;
```

maskit:
PR for the docs
https://github.com/apache/trafficserver/pull/12146

cmcfarlen:

PR #12144, opened 2025-03-31
Fix memory leaks in HttpConfig::reconfigure

hnakamur:
It is easier to see diff (without spaces changes) at https://github.com/apache/trafficserver/pull/12144/files?w=1

### DIFF
--- a/src/proxy/http/HttpConfig.cc
+++ b/src/proxy/http/HttpConfig.cc
@@ -1080,6 +1080,7 @@ HttpConfig::reconfigure()
 
   params->outbound = m_master.outbound;
 
+  ats_free(params->proxy_hostname);
   params->proxy_hostname                           = ats_strdup(m_master.proxy_hostname);
   params->proxy_hostname_len                       = (params->proxy_hostname) ? strlen(params->proxy_hostname) : 0;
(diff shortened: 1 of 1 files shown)


cmcfarlen (approved):
Good fix!

bneradt (approved):

src/proxy/http/HttpConfig.cc
```diff
@@ -1080,6 +1080,7 @@ HttpConfig::reconfigure()
 
   params->outbound = m_master.outbound;
 
+  ats_free(params->proxy_hostname);
```
cmcfarlen, line 1083: The `params` variable is allocated with `new` on line 1077, so all of these pointers should not have anything assigned to them.  We discussed this at the review meeting and think this change is not necessary.

cmcfarlen (changes requested):
If you agree this is needed, lets go ahead and close this PR.

src/proxy/http/HttpConfig.cc
hnakamur: Oh, I'm sorry for my mistake. I agree this change is not necessary.
I'm closing this now.

PR #12147, opened 2025-04-01
Take string view arguments in cache processor

hnakamur:


cmcfarlen (approved):

cmcfarlen:
I don't think this PR changes layout of any disk-stored structs.

hnakamur:
Yes, I agree with you. This PR changes the field of `struct CacheVC` but I suppose it is not stored on disk.

I found this branch has conflicts from with changes from [Cleanup: Remove unused args from open_read and open_write by masaori335 · Pull Request #12181 · apache/trafficserver](https://github.com/apache/trafficserver/pull/12181).
I merged master to the copy of this branch, resolved conflicts and pushed it to https://github.com/hnakamur/trafficserver/tree/take_string_view_arguments_in_cache_processor2.

Can I foward this branch to https://github.com/apache/trafficserver/commit/f7e77f91f01a00704ecba6a23097be6a6e5eeb5c?

masaori335:
We're going to do "squash & merge" to merge this PR, so forwarding this branch to `f7e77f9` should be fine. Let's see what diff shows.

hnakamur:
Oops, I found I created the https://github.com/hnakamur/trafficserver/tree/take_string_view_arguments_in_cache_processor2 from the wrong branch.
I'm going to fix it later today.

src/iocore/cache/CacheVC.h
```diff
@@ -281,8 +281,7 @@ struct CacheVC : public CacheVConnection {
   int                       fragment;
   int                       scan_msec_delay;
   CacheVC                  *write_vc;
-  char                     *hostname;
-  int                       host_len;
+  std::string_view          hostname;
```
masaori335, line 284: @cmcfarlen This is part of "Region C". Can we change this? IIRC, I broke cache compatibility by changing type in this region.

masaori335 (changes requested):
Let me check this doesn't break cache compatibility.

src/iocore/cache/CacheVC.h
masaori335: @bneradt do you remember what you faced with `time_t pin_in_cache`?
https://github.com/apache/trafficserver/pull/11013

hnakamur:
@masaori335 

> Let me check this doesn't break cache compatibility.

Thank you. I'll revert the change in `struct CacheVC` if it causes cache incompatibility.

> We're going to do "squash & merge" to merge this PR, so forwarding this branch to `f7e77f9` should be fine. Please push the change, let's see what diff shows.

The merge commit https://github.com/apache/trafficserver/commit/f7e77f91f01a00704ecba6a23097be6a6e5eeb5c contains the changes in other branches which were merged to master. However the diff betwen master and f7e77f9 https://github.com/apache/trafficserver/compare/master...hnakamur:f7e77f9 contains only changes in this branch.

So I think it should be OK to forward this branch to f7e77f9.

src/iocore/cache/CacheVC.h
masaori335: If changing type of member of `CacheVC` breaks cache compatibility, I afraid this change. Because the size are different. The underlaying types are `char *` and `size_t`, I guess.
```
(lldb) p sizeof(char *)
(unsigned long) 8
(lldb) p sizeof(int)
(unsigned long) 4
(lldb) p sizeof(std::string_view)
(unsigned long) 16
```

Let me clarify the symptom, if cache compatibility is broken, we'll face warnings when new ATS read old cache from disk, right?

masaori335: I remember we did back and forth around `pin_in_cache`. A change of `Doc` obviously breaks compatibility, but change in the CacheVC was not actually?
[back and forth"](https://ell.stackexchange.com/questions/189720/back-and-forward-or-back-and-forth)

masaori335 (approved):
As we talked at bug scrub, I haven't seen warnings on start up with reading old cache on disk. Let's merge this and see!

PR #12162, opened 2025-04-04
Remove misleading log statement

serrislew:
[Doc change only] Documentation for `chi` seems to imply proxy protocol ip address is used if enabled, similar to `pps` but [code](https://github.com/apache/trafficserver/blob/master/src/proxy/logging/LogAccess.cc#L1419) only reads for client info.

Discussion in slack thread suggests better wording or removal of line. Open to ideas

### DIFF
--- a/doc/admin-guide/logging/formatting.en.rst
+++ b/doc/admin-guide/logging/formatting.en.rst
@@ -493,8 +493,7 @@ incoming/outgoing ports, and network interfaces used during transactions.
 ===== ============== ==========================================================
 Field Source         Description
 ===== ============== ==========================================================
-chi   Client         IP address of the client's host. If :ref:`Proxy Protocol <proxy-protocol>`
-                     is used, this represents the IP address of the previous hop.
+chi   Client         IP address of the client's host.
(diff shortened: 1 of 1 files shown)


serrislew:
[approve ci]

masaori335:
Sorry, it's me who added this unclear sentence. As we talked at bug scrub, "peer" seems better term here.
https://github.com/apache/trafficserver/commit/cc2979d36722fc40f0434be0cd7a431414371c6a

bneradt:
> Sorry, it's me who added this unclear sentence. As we talked at bug scrub, "peer" seems better term here. [cc2979d](https://github.com/apache/trafficserver/commit/cc2979d36722fc40f0434be0cd7a431414371c6a)

In your defense, it's not so much that what you wrote was unclear as it gets tricky for us to succinctly communicate the details when Proxy Protocol is involved with the two host possibilities. We'll probably have to be a bit verbose. Something like: `Even if Proxy Protocol is enabled, this field will report the IP of the peer rather than the client IP behind the peer.` Maybe?

bneradt (approved):

PR #12157, opened 2025-04-02
Add CurlCommand to autests (#12122)

cmcfarlen:
* Add CurlCommand to autests

* MakeCurlCommand, SpawnCurlCommand, quote spaces

* separate curl and use rundir for uds path

---------

Co-authored-by: Serris Lew <lserris@apple.com>
(cherry picked from commit 35735bdee88f73ede563fc7e5bb3ad672e6f8c11)

PR #12165, opened 2025-04-05
autest updates for recent curl and nghttp2 versions

bneradt:
This patch contains updates to the autest output expections for curl 8.12.1 and nghttp2 1.65.0. It is backwards compatible with the older versions.


bneradt:
[approve ci autest 0]

randall (approved):

cmcfarlen:
This was picked with #12157 due to a chicken-egg with this and the new versions being on CI.

PR #12167, opened 2025-04-07
Take string view in mime methods

hnakamur:


bneradt:
Here's the clang-analyzer report concerning a nullptr dereference, in case it's helpful:

https://ci.trafficserver.apache.org/job/Github_Builds/job/clang-analyzer/5704/artifact/output/12167/scan-build-2025-04-07-13-36-30-195012-dqlezm4f/index.html

hnakamur:
> Here's the clang-analyzer report concerning a nullptr dereference, in case it's helpful:
> 
> https://ci.trafficserver.apache.org/job/Github_Builds/job/clang-analyzer/5704/artifact/output/12167/scan-build-2025-04-07-13-36-30-195012-dqlezm4f/index.html

I suppose this was a false positive from clang-analyzer.

I did 'git bisect' with a script whose content is https://github.com/apache/trafficserver-ci/blob/9a6bb8b40f2a6c830d52f619fa06cda22c60abf4/jenkins/github/clang-analyzer.pipeline#L60-L91
and found the commit https://github.com/apache/trafficserver/pull/12167/commits/99a15bfab85b23535e71b1882b0dede635f1be3e was the trigger.

However I did not think the commit was actual cause after reviewing the diff.
I think clang-analyzer was confused that `string_get` may return nullptr because of `if (new_url == nullptr)` below it.
```
new_url = to_free = u->string_get(&s->arena, &new_url_len);
if (new_url == nullptr) {
  new_url = "";
}
```

I suppose that `string_get` always returns non nullptr.
So I added `assert(to_free != nullptr);` at https://github.com/apache/trafficserver/pull/12167/commits/038258a62b5a3663ea1aa68e443866da02552419 and clang-analyzer does not report nullptr dereference anymore.

If this assumption is true, the following three lines are not needed.
```
if (new_url == nullptr) {
  new_url = "";
}
```

If this assumption is false, we must guard the call to `str_free` like below:
```
if (to_free) {
  s->arena.str_free(to_free);
}
```

include/proxy/hdrs/MIME.h
```diff
@@ -1139,35 +1141,32 @@ class MIMEHdr : public HdrHeapSDKHandle
...
-  void value_set_int(const char *name, int name_length, int32_t value);
-  void value_set_uint(const char *name, int name_length, uint32_t value);
-  void value_set_int64(const char *name, int name_length, int64_t value);
-  void value_set_date(const char *name, int name_length, time_t value);
+  int              value_get_index(std::string_view name, std::string_view value) const;
+  std::string_view value_get(std::string_view name) const;
+  int32_t          value_get_int(std::string_view name) const;
+  uint32_t         value_get_uint(std::string_view name) const;
+  int64_t          value_get_int64(std::string_view name) const;
+  time_t           value_get_date(std::string_view name) const;
```
cmcfarlen, line 1149: I think for this one and a few others we could provide an override that takes string_view rather than replace the existing one.  That will avoid some of the conversions to string_view that just get converted back immediately.

include/proxy/hdrs/MIME.h
```diff
@@ -1652,7 +1636,7 @@ MIMEHdr::value_append(const char *name, int name_length, const char *value, int
 inline time_t
 MIMEHdr::get_age() const
 {
-  int64_t age = value_get_int64(MIME_FIELD_AGE, MIME_LEN_AGE);
+  int64_t age = value_get_int64(std::string_view{MIME_FIELD_AGE, static_cast<std::string_view::size_type>(MIME_LEN_AGE)});
```
cmcfarlen, line 1639: This is another example where also having the other override is cleaner.

cmcfarlen (changes requested):
I like the idea of this cleanup, but I think if we kept the old functions and just added overrides then there would be fewer incidental conversions to string_view.

include/proxy/hdrs/MIME.h
hnakamur: I thought we could change lower-level functions such as `mime_hdr_prepare_for_value_set` to take string_view once this pull request is merged.

include/proxy/hdrs/MIME.h
hnakamur: I thought we could add string_view version of mime field constants once this pull request is merged.
After that, we could deprecate `const char * MIME_FIELD_XXX` and `int MIME_LEN_XXX`.

include/proxy/hdrs/MIME.h
hnakamur: I wonder should I change lower-level functions to take string_view before this pull request?
Or should I revert removing the version taking `const char *` and `int` as you suggested?

include/proxy/hdrs/MIME.h
hnakamur: I've made the changes I said above at https://github.com/apache/trafficserver/compare/038258a62b5a3663ea1aa68e443866da02552419...hnakamur:trafficserver:take_string_view_in_mime_methods2

hnakamur: I added `c_str_view` which is similar to `std::string_view` but asserts the string is terminate with `\0`.
It is needed in src/api/InkAPIInternal.cc like:
```
TS_MIME_FIELD_ACCEPT                    = MIME_FIELD_ACCEPT.c_str();
```

The type of `TS_MIME_FIELD_ACCEPT` is `const char *` and it expected the string is terminated with `\0`.
The type of `MIME_FIELD_ACCEPT` was changed to `c_str_view` which expresses so.

include/tscore/ink_memory.h
```diff
@@ -617,3 +618,45 @@ struct ats_unique_buf_deleter {
...
+  c_str_view() : c_str_view(nullptr, 0) {}
+  explicit c_str_view(const char *data, size_type length)
+    : c_str_view{
+        std::string_view{data, length}
+  }
+  {
+  }
+  explicit c_str_view(std::string_view sv) : sv_{sv}
+  {
+    ink_assert(sv_.data() == nullptr ? sv_.length() == 0 : sv_.data()[sv_.length()] == '\0');
```
cmcfarlen, line 636: `sv_.data()[sv_.length()] == '\0'` could be considered UB if the string_view is not null terminated as this might be reading past a buffer, but since this is an assert anyway its fine if it crashes when called with a non-null terminated string_view.

cmcfarlen (changes requested):
One more change request just for commentary.  Thanks for doing this!

include/tscore/ink_memory.h
```diff
@@ -617,3 +618,45 @@ struct ats_unique_buf_deleter {
 };
 using ats_unique_buf = std::unique_ptr<uint8_t[], ats_unique_buf_deleter>;
 ats_unique_buf ats_unique_malloc(size_t size);
+
+class c_str_view
```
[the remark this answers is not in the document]
hnakamur: Thank you for your comment. I added a comment at [629bdf3](https://github.com/apache/trafficserver/pull/12167/commits/629bdf3eed355e9cf146d5db68bfc987aafc6340).

cmcfarlen (approved):

PR #12168, opened 2025-04-07
Run autests with curl UDS default

### LINKED ISSUE #12184: Domain socket path for autest should be in the trafficserver runtime directory, not the test directory
Currently, the UDS path is configured to use the test directory.  If traffic_server changes user id and deescalates privilege, then it will not have access to write to the currently configured path for the server domain socket.  Also, if an autest configure multiple traffic servers, the current path will conflict.

I recommend moving the configured UDS path to be the trafficserver runtime directory.

serrislew:
Extension of #12122 

Add `autest-uds` cmake preset to run all autests with `curl --unix-socket [ts_runtime_dir/socket]` command
- skips h2 and tls autests for uds case
- skips unrelated autests or specific test runs with `Condition.CurlIsUsingUnixDomainSocket()`
   - changes include removing `--ipv4` flag when `--unix-socket` is present, not including full url in gold file, etc
- `curl_base` string format created to just have `curl` (without uds flag) available in uds case

Resolves #12184


bneradt (changes requested):
Thanks for working on this and giving an example of adding uds testing to a test in the bigobj.test.py file.

Posting a quick observation while we continue to discuss how to move forward.

serrislew:
[approve ci]

tests/gold_tests/basic/config.test.py
```diff
@@ -15,17 +15,18 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
+import os
```
bneradt: Is this needed?

tests/gold_tests/basic/copy_config.test.py
```diff
@@ -15,18 +15,19 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
+import os
```
bneradt: Not needed? i don't see a reference to `os` in your new lines, but maybe my eyes are missing it.

tests/gold_tests/basic/copy_config.test.py
```diff
@@ -37,14 +38,14 @@
 t = Test.AddTestRun("Talk to ts1")
 t.Processes.Default.StartBefore(ts1)
 t.Processes.Default.StartBefore(ts2)
-t.MakeCurlCommand("127.0.0.1:{port}".format(port=ts1.Variables.port))
+t.MakeCurlCommand("127.0.0.1:{port}".format(port=ts1.Variables.port), uds_path=ts1.Variables.uds_path)
```
bneradt: I suggest passing ts rather than uds_path for each of these. That way the user doesn't need to think of the uds_path for each of these tests (it's easier just to pass ts) and if something else like this is overlaid in MakeCurlCommand based on ts it can grab it from there.

bneradt (changes requested):

serrislew:
[approve ci centos]

bneradt (approved):

cmcfarlen:
Added to milestone 10.1.1 via #12360

PR #12169, opened 2025-04-07
9.2.x: fix libyaml for gcc 15, and cherry-pick fedora:42 updates

bneradt:
autest updates for recent curl and nghttp2 versions (#12165)

* number_of_redirects.test.py: curl verbosity change

Change curl verbosity of number_of_redirects.test.py to -v from -vvv
since in later versions of curl, the output of -vvv no longer matches
the gold file.

* chunked_encoding_h2: make content-length search specific

Later versions of nghttp2 now include content-length headers in the
request generated by chunked_encoding_h2. Therefore the search strings
for the responses have to be more specific to look in server responses
only otherwise they see the content-length in the request and
incorrectly fail.

* compress.test.py: update for recent curl output

Recent versions of curl add a 'using HTTP/2' string to the output that
the compress.test.py test has to filter out.

* connect.test.py: accommodate curl output order changes

Recent versions of curl switch the order of the "Connected to" CONNECT
information with respect to the rest of the verbose output content. This
updates the test expectations to accommodate that.

* cripts autests updates

This allows the cripts.test.py file to work in generic environments
rather than just those that install into /tmp/ats.

* polite_hook_wait.test.py: update for recent curl output

Update the polite_hook_wait.test.py so that it can handle recent curl
output that adds a 'using HTTP/1.x' or similar strings to the output.

* test_TSHttpSsnInfo.test.py: accommodate nghttp

The latest version of nghttp2 has different output that the
test_TSHttpSsnInfo.test.py gold file has to be updated to accommodate.
This updates it so that it can work for newer and older versions of the
tool.

(cherry picked from commit 73670a3c6cf83c3eb01b9d62e15e37b3698075f7)

---

yamlcpp: fix gcc 15 build issue

We don't want to update all of yamlcpp at this time in 9.2.x. This patch
addresses the following gcc 15 build error:

```
src/emitterutils.cpp:221:11: error: 'uint16_t' was not declared in this scope
  221 | std::pair<uint16_t, uint16_t> EncodeUTF16SurrogatePair(int codePoint) {
      |           ^~~~~~~~
src/emitterutils.cpp:13:1: note: 'uint16_t' is defined in header '<cstdint>'; this is probably fixable by adding '#include <cstdint>'
   12 | #include "yaml-cpp/null.h"
  +++ |+#include <cstdint>
   13 | #include "yaml-cpp/ostream_wrapper.h"
```


ezelkow1 (approved):

bneradt:
[approve ci autest 1]

bneradt:
[approve ci autest 2]

bneradt:
[approve ci autest 2]

PR #12172, opened 2025-04-08
Cleanup - some more clang-tidy recommendation.

brbzull0:


src/traffic_server/RpcAdminPubHandlers.h
```diff
@@ -20,6 +20,8 @@
 
 #pragma once
 
+#include <iostream>
```
JosiahWI, line 23: Should this be included somewhere else? It doesn't look like the header is used in this file.

JosiahWI (changes requested):
Thanks for cleaning these up. I have one concern about the `<iostream>` include.

brbzull0:
[approve ci autest 0]

JosiahWI (dismissed):

JosiahWI:
Also needs a non-Yahoo approval yet.

randall (approved):

Issue #12173, opened 2025-04-08
header_rewrite: set-destination path corruption


bneradt:
@djcarlin noticed something interesting about the header_rewrite plugin with respect to its `set-destination` directive. He was working with a config file like this:

```
cond %{REMAP_PSEUDO_HOOK}
set-destination PATH /prefix/this/%{CLIENT-URL:PATH}
```

The documentation is not clear about whether the `set-destination`'s path argument should or should not take the `/` prefix:

https://docs.trafficserver.apache.org/en/10.0.x/admin-guide/plugins/header_rewrite.en.html#header-rewrite-plugin

It does say, "URL substring beginning with (but not including) the first / after the hostname ", but how should that be applied? Does that mean the initial `/` should or should not be provided for the `set-destination` directive. All the `cond` `PATH` flavors include the initial `/`. In any case, he tried with the initial `/` first. The result of the above config is that a mysterious space gets inserted in the path. Thus, if the client had a path of `/original/path`, the result above the above `set-destination` directive is this:

```
[Apr  8 21:44:15.441] [ET_NET 0] DIAG: (header_rewrite) Building resources, hook=TS_HTTP_LAST_HOOK
[Apr  8 21:44:15.441] [ET_NET 0] DIAG: (header_rewrite)         Adding TXN client request header buffers
[Apr  8 21:44:15.441] [ET_NET 0] DIAG: (header_rewrite)         Adding TXN client request header buffers for remap instance
[Apr  8 21:44:15.441] [ET_NET 0] DIAG: (header_rewrite) Evaluating GROUP()
[Apr  8 21:44:15.441] [ET_NET 0] DIAG: (header_rewrite) Appending '/prepend/ this/' to evaluation value
[Apr  8 21:44:15.441] [ET_NET 0] DIAG: (header_rewrite)    Using the pristine url
[Apr  8 21:44:15.441] [ET_NET 0] DIAG: (header_rewrite)    Path to match is: original/path
[Apr  8 21:44:15.441] [ET_NET 0] DIAG: (header_rewrite) OperatorSetDestination::exec() invoked with PATH: /prepend/ this/original/path
```

Note the space inserted in the final path: `/prepend/ this/original/path`

This is obviously problematic. @djcarlin found that leaving off the initial `/` in the config file resolves the issue:

```
cond %{REMAP_PSEUDO_HOOK}
set-destination PATH prefix/this/%{CLIENT-URL:PATH}
```

The result is now the expected result:

```
[...]

PR #12174, opened 2025-04-09
header_rewrite: set-destination path fix

bneradt:
This test reproduces a set-destination issue in which a space is added if an initial / (probably mistakenly) is added to the set-destination operator parameter. This patch also updates the documentation to make it explicitl that set-destination PATH values should not contain an initial '/'.

Fixes: #12173


doc/admin-guide/plugins/header_rewrite.en.rst
```diff
@@ -958,7 +958,11 @@ changing the remapped destination, ``<part>`` should be used to indicate the
 component that is being modified (see `URL Parts`_), and ``<value>`` will be
 used as its replacement. You must supply a non-zero length value, otherwise
 this operator will be an effective no-op (though a warning will be emitted to
-the logs if debugging is enabled).
+the logs if debugging is enabled). Also, note that for ``PATH`` ``<part>``
+values, a leading ``/`` is not included. That is, you will generally not want
+to start your ``PATH`` value with a leading ``/``. See the ``set-destination``
```
mlibbey, line 963: Perhaps ``` `Prepend a Path to the URL`_  ``` (instead of set-destination) so that we get a direct link to it? (and if there become multiple set-dest examples, this will be clearer).

doc/admin-guide/plugins/header_rewrite.en.rst
```diff
@@ -1541,8 +1545,17 @@ could each be tagged with a consistent name to make finding logs easier.::
     sometimes avoid setting internal ``@`` headers for passing information between hooks.
     These internal state variables are much more efficient than setting and reading headers.
 
+Prepend a Path to the URL
+-------------------------
+
+The following prepends ``/foo/bar`` to the URL path. Note that since ``PATH`` is
```
mlibbey, line 1551: This should be ``/foo/bar/`` (with the trailing slash)

doc/admin-guide/plugins/header_rewrite.en.rst
```diff
@@ -1541,8 +1545,17 @@ could each be tagged with a consistent name to make finding logs easier.::
...
+non-inclusive of the initial ``/``, the provided prefix does not start with a ``/``::
```
mlibbey, line 1552: This was confusing to me -- the literal string PATH in the example versus the concept of PATH. I think I would be less confused if we add "but, it does end with a trailing /". Eg, if the url is example.com/example/ then providing foo/bar/%{CLIENT-URL:PATH} will yield example.com/foo/bar/example/ (and foo/bar%{CLIENT-URL:PATH} will yield example.com/foo/barexample/)

doc/admin-guide/plugins/header_rewrite.en.rst
```diff
@@ -958,7 +958,11 @@ changing the remapped destination, ``<part>`` should be used to indicate the
 component that is being modified (see `URL Parts`_), and ``<value>`` will be
 used as its replacement. You must supply a non-zero length value, otherwise
 this operator will be an effective no-op (though a warning will be emitted to
-the logs if debugging is enabled).
+the logs if debugging is enabled). Also, note that for ``PATH`` ``<part>``
+values, a leading ``/`` is not included. That is, you will generally not want
+to start your ``PATH`` value with a leading ``/``. See the ``set-destination``
+example in the ``Examples`` section below for a demonstration of a ``PATH``
```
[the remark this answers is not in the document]
bneradt: Good point.

doc/admin-guide/plugins/header_rewrite.en.rst
bneradt: I think I improved things. I added an example similar to the one you provided here too. Let me know what you think and we can iterate.

bneradt:
Thank you for the thoughtful doc comments, @mlibbey . @zwoop provided feedback as well that the intention is for the tokenization issues to be address via quoting the path string. I've updated the docs to make this quoting expectation more clear too (in addition to the expectation that the path string shouldn't start with an initial `/`). 

Let me know if you have other observations after these updates.

doc/admin-guide/plugins/header_rewrite.en.rst
mlibbey: Like the doc updates!
Interestingly, this week we were troubleshooting a string concat problem, where I believe the solution was to remove the quotes (they yielded a Warning). The initial string was like
set-header X-Foo "%{HEADER:Cache-Control}, %{Age}"
... the warning went away with 
set-header X-Foo %{HEADER:Cache-Control}, %{Age}

bneradt:
I'm going to close this out. Hopefully this is less needed with the hdrw4u changes.

PR #12180, opened 2025-04-10
otel build update for GCC 15

bneradt:
When building the opentelemetry plugin against GCC 15, we get a warning from a deprecated use of ciso646 in the library. No version of opentelemetry-cpp currently addresses this. The `-Wno-cpp` silences that warning. Here is the warning, for reference:

```
In file included from /opt/include/opentelemetry/nostd/internal/absl/base/options.h:73,
                 from /opt/include/opentelemetry/nostd/internal/absl/base/config.h:66,
                 from /opt/include/opentelemetry/nostd/internal/absl/types/variant.h:45,
                 from /opt/include/opentelemetry/nostd/variant.h:52,
                 from /opt/include/opentelemetry/common/attribute_value.h:10,
                 from /opt/include/opentelemetry/sdk/trace/recordable.h:6,
                 from /opt/include/opentelemetry/sdk/trace/exporter.h:9,
                 from /opt/include/opentelemetry/exporters/otlp/otlp_http_exporter.h:7,
                 from /home/bneradt/src/ts_edge/plugins/experimental/otel_tracer/tracer_common.h:18,
                 from /home/bneradt/src/ts_edge/plugins/experimental/otel_tracer/otel_tracer.cc:40:
/usr/include/c++/15/ciso646:46:4: error: #warning "<ciso646> is deprecated in C++17, use <version> to detect implementation-specific macros" [-Werror=cpp]
   46 | #  warning "<ciso646> is deprecated in C++17, use <version> to detect implementation-specific macros"
      |    ^~~~~~~
```


cmake/Findopentelemetry.cmake
```diff
@@ -118,5 +119,6 @@ if(opentelemetry_FOUND)
         list(APPEND opentelemetry_LIBRARIES opentelemetry::${OTLIB})
       endif()
     endforeach()
+    target_compile_options(opentelemetry::opentelemetry_trace INTERFACE -Wno-cpp)
```
shukitchan, line 122: `opentelemetry::opentelemetry` or `opentelemetry::opentelemetry_trace` ?

bneradt: The trace is the specific component that had the problem, so I applied it there. I wasn't 100% sure that `opentelemetry::opentelemetry` would work for the non-Apple case.

bneradt: Yeah, I tested `opentelemetry::opentelemetry_trace` and that's not actually defined for the non-Apple condition in here. But `opentelemetry::opentelemetry_trace` does make the build happy.

bneradt:
[approve ci autest fedora]

shukitchan (approved):

cmcfarlen (approved):

bneradt:
Backported to 10.1.x with:
https://github.com/apache/trafficserver/pull/12330

cmcfarlen:
Added to milestone 10.1.0 via #12330

Issue #11700, opened 2024-08-16
10.1: Crash HttpSM::setup_cache_write_transfer while handling redirect


bneradt:
We're seeing a crash related to an interaction between cache and handling redirects:

```gdb
#0  0x0000000000587c7d in HttpSM::setup_cache_write_transfer (name=0x8e1ab5 "cache write", skip_bytes=1395, store_info=0x7fa31d298180, source_vc=0x7fadc0aedc90, c_sm=0x7fa31d29a430, this=0x7fa31d298000)
    at /sd/workspace/src/git.ouryahoo.com/Edge/build/_scm/trafficserver10.0/src/proxy/http/HttpSM.cc:6770
#1  HttpSM::perform_cache_write_action (this=this@entry=0x7fa31d298000) at /sd/workspace/src/git.ouryahoo.com/Edge/build/_scm/trafficserver10.0/src/proxy/http/HttpSM.cc:6425
#2  0x000000000059d25e in HttpSM::handle_api_return (this=0x7fa31d298000) at /sd/workspace/src/git.ouryahoo.com/Edge/build/_scm/trafficserver10.0/src/proxy/http/HttpSM.cc:1630
#3  0x000000000059731e in HttpSM::state_api_callout (this=0x7fa31d298000, event=<optimized out>) at /sd/workspace/src/git.ouryahoo.com/Edge/build/_scm/trafficserver10.0/src/proxy/http/HttpSM.cc:1491
#4  0x000000000059e4a6 in HttpSM::state_api_callback (this=this@entry=0x7fa31d298000, event=event@entry=60000, data=data@entry=0x0)
    at /sd/workspace/src/git.ouryahoo.com/Edge/build/_scm/trafficserver10.0/src/proxy/http/HttpSM.cc:1295
#5  0x00007fadce570813 in TSHttpTxnReenable (txnp=0x7fa31d298000, event=TS_EVENT_HTTP_CONTINUE) at /sd/workspace/src/git.ouryahoo.com/Edge/build/_scm/trafficserver10.0/src/api/InkAPI.cc:4935
#6  0x00007fac5029cc76 in ?? ()
#7  0x00007fa31d298000 in ?? ()
#8  0x00007fa81fcac1c0 in ?? ()
#9  0x0000000000000000 in ?? ()
```

Some interesting variables @JosiahWI asked me to look at:

```gdb
(gdb) p t_state.redirect_info.redirect_in_process
$1 = true

(gdb) f 1
[...]

PR #12182, opened 2025-04-11
Reset HttpCacheSM on following redirect

masaori335:
We face an issue that `http.current_active_client_connections` metrics keep increasing. After investigation with @cmcfarlen and @moonchen, we found `HttpSM` hangs on `state_cache_open_read` when it follows redirect and `HttpCacheSM::captive_action` was cancelled. Because `HttpCacheSM::state_cache_open_read` does nothing.

https://github.com/apache/trafficserver/blob/b1e9327069fde8207c2d5ed6178addc4e237498e/src/proxy/http/HttpCacheSM.cc#L106-L112

To fix the issue, this PR does reset `captive_action.cancelled` and other counters of `HttpCacheSM` on following redirect.

From our testing, this patch seems working, however, `captive_action` and `pending_action` related code seems pretty odd. They should be fixed.

### DIFF
--- a/include/proxy/http/HttpCacheSM.h
+++ b/include/proxy/http/HttpCacheSM.h
@@ -49,6 +49,12 @@ struct HttpCacheAction : public Action {
   {
     sm = sm_arg;
   };
+  void
+  reset()
+  {
+    cancelled = false;
+  }
+
   HttpCacheSM *sm = nullptr;
 };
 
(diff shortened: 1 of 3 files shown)


JosiahWI:
Would also be interesting to know whether this affects #11700.

JosiahWI (approved):
Very nice fix. I believe I found this bug during testing earlier, but I did not have time to confirm it or track down what caused it.

I agree we need to look at how `captive_action` and `pending_action` are handled... I think other bugs are probably also related to it.

masaori335:
For the record, the issue happens only on 10.0.x and master. Because we reverted commits from 9.2.x branch.
https://github.com/apache/trafficserver/pull/9338

cmcfarlen:

PR #12183, opened 2025-04-11
traffic_ctl - Implement `server status` command option.

brbzull0:
With the removal of `traffic_manager` we lost this command option, so now this change  makes ATS to respond the rpc call with some basic `TSSystemState` Information, like intimidation status, draining status, etc. The output is a simple json output.
Command option is read only, no change is done in the internal ATS state.

Example:
```bash
$ traffic_ctl server status | jq
{
    "initialized_done": "true",
    "is_ssl_handshaking_stopped": "false",
    "is_draining": "false",
    "is_event_system_shut_down": "false"
}
```
This also includes some autests to validate the output. Changes to existing traffic_ctl autest machinery is made to make it simpler to extend and add more tests.

A clean up on features that makes no more sense to handle using `traffic_ctl`(like restart, etc) is also done in this PR. This was implemented when we had traffic_manager.


cmcfarlen (approved):
Awesome, thanks!

cmcfarlen:

PR #12187, opened 2025-04-14
Make chunk size parsing more strict

yknoya:
# Problem
There are cases where chunked requests that do not comply with RFC 9112 are forwarded to the origin server even when `proxy.config.http.strict_chunk_parsing` is set to 1.
For example, when sending an invalid chunked body like the one below, the request, including the chunked body, was forwarded to the origin server.

```
POST /post HTTP/1.1\r\n
Host: example.com\r\n
User-Agent: test\r\n
Transfer-Encoding: chunked\r\n
\r\n
4\n
test\r\n
0\r\n
\r\n
```

# Cause
The bug is caused by `ChunkedHandler::read_size()`. When the aforementioned chunked request is sent, `ChunkedHandler::read_size()` should be able to determine that the chunked body is invalid when it encounters the '\n' following '4'. However, `ChunkedHandler::read_size()` does not recognize it as invalid and continues processing.
Therefore, @maskit and I have fixed it to detect an invalid chunked body when it encounters the '\n' following '4'.

### DIFF
--- a/src/proxy/http/HttpTunnel.cc
+++ b/src/proxy/http/HttpTunnel.cc
@@ -183,9 +183,10 @@ ChunkedHandler::read_size()
           }
         } else {
           // We are done parsing size
-          if ((num_digits == 0 || running_sum < 0) ||       /* Bogus chunk size */
-              (!ParseRules::is_wslfcr(*tmp) && *tmp != ';') /* Unexpected character */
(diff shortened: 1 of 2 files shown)


bneradt (approved):
lgtm

cmcfarlen:

PR #12190, opened 2025-04-15
doc: ssl.client.alpn_protocol h22o update

bneradt:
ssl.client.alpn_protocol was added before HTTP/2 to origin was supported and the "h2" examples stated the lack of support. But now that we support HTTP/2 to origin since 10.x, the docs should be updated to remove the statement that HTTP/2 to origin is not supported.


maskit (approved):

cmcfarlen:

PR #12192, opened 2025-04-15
Return a 400 on chunk parse errors

### LINKED ISSUE #12171: No "400 Bad Request" response when strict_chunk_parsing is set to 1
According to the documentation for proxy.config.http.strict_chunk_parsing in https://github.com/apache/trafficserver/pull/12150, it states:
```
If enabled (``1``), |TS| returns 400 Bad Request if chunked message body is not compliant with RFC 9112.
```

However, when I set strict_chunk_parsing to 1 and send a chunked message body that is not compliant with RFC 9112, no response is returned, and the connection is dropped instead.

Here is the request content sent during the test (note that the character after '4' is '\n'):
```
POST /post HTTP/1.1\r\n
Host: example.com\r\n
User-Agent: test\r\n
Transfer-Encoding: chunked\r\n
\r\n
4\n
test\r\n
0\r\n
\r\n
```

maskit:
This fixes #12171.


bneradt (commented):
Looks promising. Hopefully an autest for this would be not too much work. In fact, I would assume there are some tests now that should be able to be updated to expect a 400 response?

bneradt (approved):

cmcfarlen:

PR #12194, opened 2025-04-16
Remove accept_fn from NetAccept

moonchen:
This function pointer always points to net_accept.  Remove the extra indirection to make the code easier to read.


cmcfarlen:

PR #12198, opened 2025-04-17
proxy.config.http.per_client.connection.allow_list.filename

bneradt:
This implements
proxy.config.http.per_client.connection.allow_list.filename, a configuration for the user to be able to provide a set of IP addresses that are not counted against
proxy.config.net.per_client.max_connections_in.


zwoop:
`allow_` sounds a little strange, `exempt_` would be better.

bneradt:
[approve ci autest 1]

bneradt:
[approve ci autest 1]

bneradt:
> `allow_` sounds a little strange, `exempt_` would be better.

Agreed. Changed to exempt terminology.

maskit:
I'm wondering how this kind of lists should be on ATS config.

`proxy_protocol_allowlist` is just a comma-separated string (it's probably because the previous config format could not have a real list structure). And `per_client.connection.exempt_list` is a string too but a filename.

I do see the benefit of having the list in a separate file, but I also think inconsistency should be avoided. Most settings that have filenames are ones that had different formats, and now we are trying to make all of them YAML. We can have real lists in records.yaml. It may be a time to think about what should be separated from records.yaml.

bneradt:
> I'm wondering how this kind of lists should be on ATS config.
> 
> `proxy_protocol_allowlist` is just a comma-separated string (it's probably because the previous config format could not have a real list structure). And `per_client.connection.exempt_list` is a string too but a filename.
> 
> I do see the benefit of having the list in a separate file, but I also think inconsistency should be avoided. Most settings that have filenames are ones that had different formats, and now we are trying to make all of them YAML. We can have real lists in records.yaml. It may be a time to think about what should be separated from records.yaml.

My internal implementation initially had this as a comma separated list, but ops asked me to make it separate file. And, as you point out, that's a pretty reasonable ask. It makes deployment of this easier. Would a rename of this configuration help things? Maybe `exempt_list_filename`?

maskit:
> Would a rename of this configuration help things? Maybe exempt_list_filename?

I'm not sure if we want to rename it. Depends on what we are going to do in the future.

The bigger issue I tried to raise was that one could ask for separating out any part of ATS config and we don't have a policy for it. `proxy_protocol_allowlist` currently cannot be in a separate file. Should we introduce `proxy_protocol_allowlist_filename` as well for consistency? Somebody might want to have entire `http` config in another file. Would we support it?

It could be a discussion topic for ATS summit or a hackathon.

bneradt:
[approve ci rocky autest 4]

PR #12200, opened 2025-04-17
Fix FreeBSD Raw Cache Disk Support

moonchen:
When we use a raw cache disk, we must obtain the disk's geometry using ioctl.  At some point it became necessary to include <sys/disk.h> in order to have these ioctl numbers in FreeBSD.


cmcfarlen:

Issue #12204, opened 2025-04-21
`openDirEntry` leak on following redirect


masaori335:
We tested https://github.com/apache/trafficserver/pull/12182 on production, the active connection leak is fixed as expected, but it introduced new leak of `openDirEntry`. ATS was killed by OOM killer in 4 days because of this leak.

For open_read retry and open_write retry after following redirect, counters should to be reset theoretically, but it triggered the leak by somehow.

PR #12205, opened 2025-04-21
Fix openDirEntry leak

masaori335:
Workaround fix for #12204. 

I haven't found how resetting counters makes openDirEntry leak. However, this leak is serious on production so we need to revert the suspicious change.

- Leave the resetting `captive_action` for the active connection leak,
- Get rid of the resetting counters for the openDirEntry leak.

We tested this patch over a week, we haven't seen any leak so far.


bryancall (approved):

cmcfarlen:

PR #12208, opened 2025-04-22
Clean up NetAccept

moonchen:
* Give sensible names to the three different accept code paths (dedicated, per-thread, periodic)
* Put common code into functions - vc initialization, send/recv buf settings
* Make QUIC UDP code path separate from TCP


moonchen:
During the course of this cleanup, I found out that the different code paths have small differences in how they set up socket and UnixNetVConnection options.  I don't think those are intentional differences, so they have been unified in this PR.

PR #12218, opened 2025-05-01
Refactor HRW matchers, add SETS

zwoop:
This also does some significant cleanup of all the classes and duplication.


zwoop:
[approve ci autest 0]

bneradt (changes requested):
Looks good.

It would be good to have more of an autest to verify the set functionality though. I know you added a set to make sure that `(GET, PUSH)` doesn't break a `GET` test, but a positive test of the set functionality would be helpful.

zwoop:
> Looks good.
> 
> It would be good to have more of an autest to verify the set functionality though. I know you added a set to make sure that `(GET, PUSH)` doesn't break a `GET` test, but a positive test of the set functionality would be helpful.

Done. I had to cleanup some other broken HRW tests too.

bneradt (approved):
Looks great.

PR #12220, opened 2025-05-02
x-ja3-via and x-ja4-via

bneradt:
Add via strings for the JA* fingerprints. This can be helpful in situations where multiple layers of proxies can add these and it may not be obvious what host added what fingerprint.


bneradt:
> Looks good. My one concern is that the `append_JA4_header` function no longer does what its name says. Consider renaming to `append_JA4_headers` (plural).

Good point. Updated. Thank you for the review.

JosiahWI (changes requested):
`string_view` should be included in the JA4 plugin.cc as well.

bneradt:
> `string_view` should be included in the JA4 plugin.cc as well.

Great catch. Included now.

JosiahWI (commented):
Looks great. Thanks for adding Autest cases for the new header! I'll let a non-Yahoo approve.

cmcfarlen (approved):

PR #12222, opened 2025-05-05
Fix compilation problem with wasm plugin + Update the use of proxy-wasm library

shukitchan:
- Wasm plugin depends on proxy-wasm libraries and there are recent updates to support WAMR-2.1.1 and wasmtime-24.0.0. This PR updates correspondingly with that.
- Since the switch to cmake from auto tools, the compilation is not working fine. This PR fixes that by probably identifying the requirements of wasm engines and adding the corresponding missing directives
- Also fedora has an older version of wamr installed and we thus turned off support of wamr in some CI jobs


plugins/experimental/wasm/CMakeLists.txt
```diff
@@ -18,5 +18,28 @@
...
+set(WASM_RUNTIME wasmlib)
+if(wamr_FOUND)
+  list(APPEND WASM_RUNTIME wamr::wamr)
+endif()
+if(wasmtime_FOUND)
+  list(APPEND WASM_RUNTIME wasmtime::wasmtime)
+endif()
+if(wasmedge_FOUND)
+  list(APPEND WASM_RUNTIME wasmedge::wasmedge)
+endif()
```
cmcfarlen, line 31: If having multiple runtime included might be an issue, you could use else if to ensure only one. You'll have to pick some priorities to the libraries, but users could still select their preference with cmake directives.

cmcfarlen (approved):
Might use if else when setting up libraries but if that isn't a problem then this looks great!  Thanks!

plugins/experimental/wasm/CMakeLists.txt
shukitchan: yes. you are right. wasmtime and wamr cannot be used together.

Issue #12140, opened 2025-03-28
10.1.x: Crash in SSL_read_early_data


bneradt:
When testing 10.1.x on a production box, I'm seeing the following crash with some frequency:

```
(gdb) bt
#0  0x00007f1504a95b1f in ssl3_ctrl (s=0x7f0abe756800, cmd=<optimized out>, larg=0, parg=0x0) at ssl/s3_lib.c:3736
#1  0x00000000007e193a in ssl_callback_info (ssl=0x7f0abe756800, where=<optimized out>, ret=<optimized out>) at /home/bneradt/build/_scm/trafficserver10.1/src/iocore/net/SSLUtils.cc:1092
#2  0x00007f1504b11cb4 in tls_finish_handshake (s=0x7f0abe756800, wst=<optimized out>, clearbufs=<optimized out>, stop=1) at ssl/statem/statem_lib.c:1519
#3  0x00007f1504b056ec in write_state_machine (s=0x7f0abe756800) at ssl/statem/statem.c:860
#4  state_machine (s=0x7f0abe756800, server=1) at ssl/statem/statem.c:488
#5  0x00007f1504aa5aa2 in SSL_read_early_data (s=0x7f0abe756800, buf=0x7f096e1ff000, num=16384, readbytes=0x7f14f83fcf48) at ssl/ssl_lib.c:2377
#6  0x00000000007c59f1 in SSLNetVConnection::_ssl_accept (this=0x7f07e7408000) at /home/bneradt/build/_scm/trafficserver10.1/include/tscore/Ptr.h:116
#7  0x00000000007ca782 in SSLNetVConnection::sslServerHandShakeEvent (this=this@entry=0x7f07e7408000, err=@0x7f14f83fd464: 0) at /home/bneradt/build/_scm/trafficserver10.1/src/iocore/net/SSLNetVConnection.cc:1249
#8  0x00000000007cb1c9 in SSLNetVConnection::sslStartHandShake (this=0x7f07e7408000, event=<optimized out>, err=@0x7f14f83fd464: 0) at /home/bneradt/build/_scm/trafficserver10.1/src/iocore/net/SSLNetVConnection.cc:1050
#9  0x00000000007c7a49 in SSLNetVConnection::net_read_io (this=0x7f07e7408000, nh=0x7f14fd8ce690) at /home/bneradt/build/_scm/trafficserver10.1/src/iocore/net/SSLNetVConnection.cc:507
[...]

PR #12224, opened 2025-05-06
Call SSL_get_negotiated_group not during read early data

bneradt:
Recent versions of OpenSSL define SSL_CB_HANDSHAKE_DONE as follows:

> Callback has been called because a handshake is finished. It also
> occurs if the handshake is paused to allow the exchange of early data.

Calling SSL_get_negotiated_group duing early data situations was causing a crash. This moves the calling of SSL_get_negotiated_group next to _record_tls_handshake_end_time where it will reliably not be called in the context of processing early data.

Fixes: #12140


maskit:
We should call the new function from `QUICNetVConnection::_switch_to_established_state()` as well.

bneradt:
[approve ci autest 0]

maskit (approved):

cmcfarlen:

PR #12229, opened 2025-05-08
Cleanup: Just use vector for seen in RamCacheLRU

hnakamur:


JosiahWI:
Why would someone do this in the first place? Some kind of oversight, maybe from a previous refactor?

The most interesting reason I've thought of so far to use `*std::vector<bool>` is to reduce the stack memory footprint of the containing structure. I don't think `RamCacheLRU` needs to have a low stack size though, right?

hnakamur:
The previous refactor was [Improve seen list for the RAM cache by zwoop · Pull Request #10662 · apache/trafficserver](https://github.com/apache/trafficserver/pull/10662).

Should I change the type of `seen` to `std::unique_ptr<std::vector<bool>>` as suggested at https://github.com/apache/trafficserver/pull/10662#discussion_r1406592391

src/iocore/cache/RamCacheLRU.cc
```diff
@@ -54,7 +54,7 @@ struct RamCacheLRU : public RamCache {
   void init(int64_t max_bytes, StripeSM *stripe) override;
 
   // private
-  std::vector<bool> *seen = nullptr;
+  std::vector<bool> seen{};
```
JosiahWI, line 57: The `{}` is actually redundant here, but doesn't hurt.

JosiahWI (approved):
Talked to @zwoop off-platform; the pointer was left in to minimize the scope of his changes. Thank you for the improvement, @hnakamur!

src/iocore/cache/RamCacheLRU.cc
hnakamur: I removed `{}` at [0b4de59](https://github.com/apache/trafficserver/pull/12229/commits/0b4de5905caefe754f433476fdb3c39159057eb1).

JosiahWI (approved):

Issue #12214, opened 2025-04-28
10.1: No origin side connection reuse


bneradt:
We put 10.1 on a production box. I noticed that the following stats were showing pretty much no origin side connection reuse:

- `proxy.process.http.outgoing_requests`
- `proxy.process.http.total_server_connections`

I'm not sure whether it is relevant, but we use global pool sharing (`proxy.config.http.server_session_sharing.pool`).

I observed that the following two reverts resolved the issue:

```
* commit dbef3627f0929f35c5b35c73f2cab6609c07b367 (HEAD -> origin_connection_issue, origin/origin_connection_issue)
| Author: bneradt <bneradt@yahooinc.com>
| Date:   Mon Apr 28 11:29:14 2025 -0500
| 
|     Revert "Cleanup SSLSNIConfig (#11027)"
|     
|     This reverts commit 94c5b42057f87118b467527066cecab406b7ca3f.
| 
* commit 9b399d5cc397bb4e2a22ae4121440b04d2948728
| Author: bneradt <bneradt@yahooinc.com>
| Date:   Mon Apr 28 11:13:48 2025 -0500
| 
|     Revert "Cleanup: Remove server name stuff from NetVConnection (#11745)"
|     
|     This reverts commit 2d34cd33e0b77b0ef59a0fcf675bd713687a4d4b.
| 
```

I'm not sure whether both are necessary, but the revert was cleaner with both.

cmcfarlen:
#11027 and #11745

maskit:
It seems like to be caused by #117458. My theory is that the connection pool cannot find a connection for server name X, because none of the connections in the pool doesn't say they are for server name X.

Before the change, we used `NetVConnectin::get_sni_servername()` that internally calls `SSL_get_servername()`, and that OpenSSL function works for both incoming and outoging connections. But now we use `TLSSNISupport::get_sni_server_name` and that only works for incoming connections because `_sni_server_name` seems to be only set on incoming connections.

We probably need to let TLSSNISupport know the server name when we call `SSL_set_tlsext_host_name` for an outgoing connection.

maskit:
Maybe `TLSSNISupport` should provide a function that internally calls `SSL_set_tlsext_host_name`.

bneradt:
> It seems like to be caused by [#11745](https://github.com/apache/trafficserver/pull/11745).

You are correct, @maskit. I confirmed that reverting only #11745 resolves the issue.

PR #12231, opened 2025-05-09
Update SNI for origin side connections

bneradt:
Origin side connection pooling was not able to access the appropriate SNI of those connections, resulting in those connections not being able to be reused appropriately. This addresses that issue by setting the internal SNI buffer for those connections to the appropriate SNI so that the origin conection pooling feature has access to the SNI value.

This also populates the SNI buffer when an SSLNetVConnection is bound to an SSL object.

Fixes: #12214

---

### For Review

Essentially, the problem with #12214 was that the `_sni_server_name` was not set in certain origin side connection situations. This address that by doing two things:

* Add `TLSSNISupport::set_sni_server_name` for callers instead of calling `SSL_set_tlsext_host_name` directly. This function both calls `SSL_set_tlsext_host_name` and updates `_sni_server_name`.
* On `TLSSNISupport::bind` (which is used when an SSL instead is being reused for a new SSLNetVConnection), update `_sni_server_name` from the SNI of the SSL instance.

### DIFF
--- a/include/iocore/net/TLSSNISupport.h
+++ b/include/iocore/net/TLSSNISupport.h
@@ -58,8 +58,30 @@ public:
 
   int perform_sni_action(SSL &ssl);
   // Callback functions for OpenSSL libraries
+
+  /** Process a CLIENT_HELLO from a client.
+   *
+   * This is for client-side connections.
+   */
   void on_client_hello(ClientHello &client_hello);
+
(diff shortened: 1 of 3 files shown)


bneradt:
[approve ci autest 1]

maskit (approved):
Although this fixes the bug, it's unclear why we have to call `SSL_get_servername` for some cases.

cmcfarlen:

PR #12233, opened 2025-05-09
revert #10996 move session pool metrics outside critical section

traeak:
With ATS10 we found that the session pool metrics aren't properly decrementing.  Removing this PR fixes the issue and makes the code simpler and really shouldn't have any impact on performance.

Consider for back port.


cmcfarlen (approved):

cmcfarlen:

PR #12234, opened 2025-05-09
HRW: Allow using a request header for IP lookups

zwoop:
Using this, you can optionally request for the GEO conditions to get the client IP from a client request header, e.g.

```
cond %{GEO:COUNTRY:X-Client-IP} =US [NOT]
    set-status 403
```

### DIFF
--- a/doc/admin-guide/plugins/header_rewrite.en.rst
+++ b/doc/admin-guide/plugins/header_rewrite.en.rst
@@ -343,6 +343,16 @@ setting headers. For example::
         set-header ATS-Geo-ASN %{GEO:ASN}
         set-header ATS-Geo-ASN-NAME %{GEO:ASN-NAME}
 
+Additionally, you can specify a client request header to use to extract an IP address
+value from. This is specified as a second argument, separated with a second colon.
+For example::
+
+    cond %{GEO:COUNTRY:X-Client-IP} =US
+        set-status 403
+
+If there is no request header matching the condition, the condition will always return
+an empty string, as if the client IP was not found in the GEO-IP database.
+
 GROUP
 ~~~~~
 ::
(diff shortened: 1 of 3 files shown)


zwoop:
[approve ci autest 2]

doc/admin-guide/plugins/header_rewrite.en.rst
```diff
@@ -323,6 +323,16 @@ setting headers. For example::
         set-header ATS-Geo-ASN %{GEO:ASN}
         set-header ATS-Geo-ASN-NAME %{GEO:ASN-NAME}
 
+Additionally, you can specify a client request header to use to extract an IP address
+value from. This is specified as a second argument, separated with a second comma.
+For example::
+
+    cond %{GEO:COUNTRY:X-Client-IP} =US
+        set-status 403
```
bneradt: > This is specified as a second argument, separated with a second comma.

Should this say, separated by a second `colon` instead of `comma` ?

zwoop: Fixed.

bneradt (dismissed):

maskit:
I proposed a universal way to use an IP address from an HTTP header field (or somewhere else). 
https://github.com/apache/trafficserver/pull/12359

On the PR, I added a new plugin, real-ip, that reads an IP address from an HTTP header field and provides it as client's real IP address. header_rewrite uses the  IP address if INBOUND_IP_SOURCE is set to PLUGIN.

If you want to control things within hrw rules, you can add a hrw operator to call the API and set client's IP address.


I think `%{GEO:COUNTRY:X-Client-IP}` is going to be painful if you have a lot of hrw rules because every place that uses IP address needs to be modified.

PR #12236, opened 2025-05-09
grpc: tunnel_handler_trailer: check for valid consumer VC

bneradt:
Verify that the consumer has a valid VConnection before setting up a grpc tunnel for it. Otherwise, avoid the tunnel. Without this change, the tunnel can crash when it tries to access a non-existent VConnection.


bryancall:
@bneradt can you please resolve the conflict

bneradt:
Converting to draft as I don't trust this patch yet. More resources got spent on #12325 because that was a bigger crasher for us. Hopefully I can come back around to investigating this.

PR #12238, opened 2025-05-10
Cleanup: Use std::unique_ptr where easily applicable

hnakamur:


src/tscore/ConsistentHash.cc
```diff
@@ -47,7 +47,7 @@ ATSConsistentHash::insert(ATSConsistentHashNode *node, float weight, ATSHash64 *
   if (h) {
     thash = h;
   } else if (hash) {
-    thash = hash;
+    thash = &*hash;
```
cmcfarlen, line 50: There are a couple of these uses of `&*` vs `.get()`.  Why not use `.get()` here as well? Its more clear imo than `&*`.

cmcfarlen (changes requested):
Consider using `.get()` instead of `&*`.  Also there are conflicts.

hnakamur:
Thank you. I fixed it to use `.get()` instead of `&*` at https://github.com/apache/trafficserver/pull/12238/commits/2470141455ba3a9a21455407fba665299404d011 and resolved conflicts at https://github.com/apache/trafficserver/pull/12238/commits/029e6c48a377902b2b28be79507ba230cf3bc732

bryancall:
@hnakamur Can you please take care of the conflict?  It would be great to get this merged in.

hnakamur:
Thank you. I merged master and resolved conflicts at e9d009d

cmcfarlen (approved):

PR #12241, opened 2025-05-13
Serve stale contents until max-age + max_stale_age

masaori335:
This clarifies behavior of `proxy.config.http.cache.max_stale_age` config. 

For now, `HttpTransact::is_stale_cache_response_returnable` simply compares `current_age` and `max_stale_age` config. The problem is if `max_stale_age` is shorter than `max-age`, ATS can't serve stale contents as expected.

Screenshot 2025-05-13 at 11 27 46

It's discussable this is a bug fix or incompatible config change.

/cc @mlibbey

### DIFF
--- a/tests/gold_tests/cache/replay/negative-revalidating-enabled.replay.yaml
+++ b/tests/gold_tests/cache/replay/negative-revalidating-enabled.replay.yaml
@@ -78,7 +78,7 @@ sessions:
       status: 200
 
   # Verify that with negative_revalidating enabled, we serve the 200 OK out of
-  # the cache even though it is stale (but younger than max_stale_age).
+  # the cache even though it is stale (but younger than max-age + max_stale_age).
   - client-request:
       method: "GET"
       version: "1.1"
@@ -115,10 +115,9 @@ sessions:
         - [ Host, example.com ]
         - [ uuid, 14 ]
 
-          # After this delay, the item is 8 seconds old. This makes it:
-          #  6 seconds beyond the server's max-age of 2 seconds and
-          #  2 seconds beyond ATS's max_stale_age of 6 seconds.
-      delay: 4s
+          # After this delay, the item is 10 seconds old. This is 2 seconds beyond
+          #  the server's max-age of 2 seconds + ATS's max_stale_age of 6 seconds.
+      delay: 6s
 
     server-response:
       status: 503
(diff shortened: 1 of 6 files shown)


mlibbey:
I think this is a compatible bug fix, as the current behavior covers the "riskier" cases, but not the "safer" cases -- the fix begins covering the safer case. 
Currently, the max_stale_age setting only applies to objects with shorter Cache-Control: max-age  (aka ttl). The shorter the ttl, the larger the percentage of potential stale time. For instance, the default 7 day max_stale_age would be 700% of a 1 day ttl; and obviously grows the shorter the ttl. Conversely, the default is only 25% of a 1 month ttl and 2% of a 1 year percent. It's much safer to allow an object with 1yr ttl to be served an extra 2% than to allow an 1hr ttl object to be served for so long. 
(We should really have a percentage setting instead-- for instance setting stale to be 25% would allow a 1hr ttl object to be served stale for an extra 15 min, whereas a 1 month ttl object would be allowed to be served for an extra week.)

mlibbey:
The bug part of the diagram would be:
```
<-----------------  Fresh
<----- CC: max-age ----->
config: max_stale-->
```
Does not allow serving stale

tests/gold_tests/cache/replay/negative-revalidating-enabled.replay.yaml
```diff
@@ -115,10 +115,9 @@ sessions:
         - [ Host, example.com ]
         - [ uuid, 14 ]
 
-          # After this delay, the item is 8 seconds old. This makes it:
-          #  6 seconds beyond the server's max-age of 2 seconds and
-          #  2 seconds beyond ATS's max_stale_age of 6 seconds.
-      delay: 4s
+          # After this delay, the item is 10 seconds old. This is 2 seconds beyond
```
mlibbey, line 118: Wonder if we should keep the 8sec one (or make it 7sec) to show that its greater than max_stale_age yet still successfully serves stale ... or maybe just change the first stale check to be at 7s?

masaori335: This is checking current age exceeds `max-age + max_stale_age` (8 seconds) and ATS returns 503.

> to show that its greater than max_stale_age yet still successfully serves stale

This case is covered by `proxy_serve_stale` test. 
https://github.com/apache/trafficserver/pull/12241/files#diff-40842de8789380a64ac754078299609752920b52aae3d85a201d1ceab191d650R122-R124

bryancall:
The new meaning makes sense based on the definition of max-stale in [RFC 9111](https://www.rfc-editor.org/rfc/rfc9111#name-max-stale)

cmcfarlen:

PR #12245, opened 2025-05-15
Connection tracker: Add functions to query inbound groups.

brbzull0:
This, as well as the outbound groups will be used to report its data using the jsonrpc handlers.(upcoming [PR](https://github.com/apache/trafficserver/pull/12260))
This also fixes an issue in the build of the output json stream(`outbound_to_json_string`).

Some output:

IN
```bash
[I] Peer Connection Tracking
Current | Block |                  Address |                     Hostname Hash |    Match |
  ------|-------|--------------------------|-----------------------------------|----------|
1       | 1     | 127.0.0.1:35204          | 00000000000000000000000000000000  | ip       |
  ------|-------|--------------------------|-----------------------------------|----------|
```
_Note: `hash=00000000000000000000000000000000` as the `fqdn` is empty._

OUT
```bash
[O] Peer Connection Tracking
Current | Block |                  Address |                     Hostname Hash |    Match |
  ------|-------|--------------------------|-----------------------------------|----------|
5       | 0     | 127.0.0.1:80             | f528764d624db129b32c21fbca0cb8d6  | both     |
  ------|-------|--------------------------|-----------------------------------|----------|
```

### DIFF
--- a/src/iocore/net/ConnectionTracker.cc
+++ b/src/iocore/net/ConnectionTracker.cc
@@ -318,13 +318,61 @@ ConnectionTracker::get_outbound_groups(std::vector<std::shared_ptr<Group const>>
   }
 }
 
+void
+ConnectionTracker::get_inbound_groups(std::vector<std::shared_ptr<Group const>> &groups)
+{
+  std::lock_guard<std::mutex> lock(_inbound_table._mutex); // TABLE LOCK
+  groups.resize(0);
+  groups.reserve(_inbound_table._table.size());
+  for (auto &&[key, group] : _inbound_table._table) {
+    groups.push_back(group);
+  }
+}
+std::string
+ConnectionTracker::inbound_to_json_string()
+{
+  std::string                    text;
+  size_t                         extent = 0;
+  static const swoc::bwf::Format header_fmt{R"({{"count": {}, "list": [  )"};
+  static const swoc::bwf::Format item_fmt{
+    R"(  {{"type": "{}", "ip": "{}", "fqdn": "{}", "current": {}, "max": {}, "blocked": {}, "alert": {}}},
+)"};
+  static const std::string_view trailer{" \n]}"};
+
+  static const auto printer = [](swoc::BufferWriter &w, Group const *g) -> swoc::BufferWriter & {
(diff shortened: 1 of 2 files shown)


src/iocore/net/ConnectionTracker.cc
```diff
@@ -318,13 +318,61 @@ ConnectionTracker::get_outbound_groups(std::vector<std::shared_ptr<Group const>>
...
+  static const auto printer = [](swoc::BufferWriter &w, Group const *g) -> swoc::BufferWriter & {
+    w.print(item_fmt, g->_match_type, g->_addr, g->_fqdn, g->_count.load(), g->_count_max.load(), g->_blocked.load(),
+            g->get_last_alert_epoch_time());
+    return w;
+  };
+
+  swoc::FixedBufferWriter                   null_bw{nullptr}; // Empty buffer for sizing work.
+  std::vector<std::shared_ptr<Group const>> groups;
+
+  self_type::get_inbound_groups(groups);
```
bneradt, line 351: Looks like this line is the only difference between this and `outbound_to_json_string`. Maybe consolidate these somehow? Maybe keep the interface as is with public `inbound_to_json_string` and `outbound_to_json_string` with a private `_to_json_string_helper` that takes a functor it can call?

src/iocore/net/ConnectionTracker.cc
```diff
@@ -360,14 +408,43 @@ ConnectionTracker::outbound_to_json_string()
...
+  dump_outbound(f);
+  dump_inbound(f);
+}
+
+void
+ConnectionTracker::dump_inbound(FILE *f)
+{
+  std::vector<std::shared_ptr<Group const>> groups;
+
+  self_type::get_inbound_groups(groups);
```
bneradt, line 444: Similar statement as above.

bneradt (changes requested):
Looks great.  This will be really helpful to expose visibility to this data.

src/iocore/net/ConnectionTracker.cc
brbzull0: yup, agreed.
Done!

serrislew (approved):

PR #12246, opened 2025-05-15
header_rewrite: Add INBOUND_IP_SOURCE controller

maskit:
The new controller enables users to choose which IP address to use for operators such as `IP`, `INBOUND`, `CIDR`, and `GEO`.

### DIFF
--- a/doc/admin-guide/plugins/header_rewrite.en.rst
+++ b/doc/admin-guide/plugins/header_rewrite.en.rst
@@ -1153,6 +1153,10 @@ Controller         Operators/Conditions  Description
 TIMEZONE           ``NOW``               If ``GMT`` is passed, the operators and conditions use GMT regardles of the timezone setting
                                          on your system. The default value is ``LOCAL``.
 ================== ===================== =============================================================================================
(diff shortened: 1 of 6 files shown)


bryancall:
```
/home/jenkins/workspace/Github_Builds/docs/src/doc/admin-guide/plugins/header_rewrite.en.rst:1124: WARNING: Blank line required after table. [docutils]
/home/jenkins/workspace/Github_Builds/docs/src/doc/admin-guide/plugins/header_rewrite.en.rst:1127: WARNING: Definition list ends without a blank line; unexpected unindent. [docutils]
/home/jenkins/workspace/Github_Builds/docs/src/doc/admin-guide/plugins/header_rewrite.en.rst:1127: ERROR: Malformed table.
Bottom/header table border does not match top border.
```

maskit:
This is an alternative approach to use an IP address that is not from the socket.

PR #12234 allows getting an IP address from an arbitrary header, but that is unsafe in general. The validity of IP address is unverified. Users would have to have either another HRW rule to verify the validity (but how?), or a plugin to validate the header value.

bneradt (approved):

cmcfarlen:
Added to milestone 10.1.0 via #12246

PR #12247, opened 2025-05-15
hrw4u is a new header_rewrite configuration DSL

zwoop:
Eventually, if this seems useful, we should make a native C++ parser / emitter for the plugin itself, in C++.


zwoop:
[approve ci autest 1]

zwoop:
[approve ci docs
]

zwoop:
[approve ci docs]

zwoop:
[hrw4u-1.0.0-py3-none-any.whl.gz](https://github.com/user-attachments/files/20615777/hrw4u-1.0.0-py3-none-any.whl.gz)

doc/admin-guide/configuration/index.en.rst
```diff
@@ -33,3 +33,4 @@ Proxy Cache Configuration
    transparent-forward-proxying.en
    hierarchical-caching.en
    proxy-protocol.en
+   hrw4u.en
```
bneradt, line 36: This seems out of place here since this describes the DSL for the header rewrite plugin. The other stuff here concerns core proxy configuration.

Honestly, I suggest we simply replace the current header rewrite docs with what you have above and link old docs from there in case people want to reference the old syntax. (In the transition period, people will need docs for the old language to help understand their scripts.)

doc/admin-guide/configuration/hrw4u.en.rst
```diff
@@ -0,0 +1,600 @@
...
+
+HRW4U replaces the free-form text parsing of `header_rewrite` with a formally defined
+grammar using ANTLR. This makes HRW4U easier to parse, validate, and extend.
+
+Rather than repeating `header_rewrite` documentation, please refer to:
+  - :ref:`admin-plugins-header-rewrite` for feature behavior and semantics
+  - This page focuses on syntax and behavior *differences* in HRW4U
+
+Why HRW4U?
+----------
```
bneradt, line 38: Judging from my observation from the summit, I think we bought the coolaide. I suggest making these _the_ header_rewrite docs. In my opinion we don't need to say "hrw4u" - this is just how you configure header_rewrite now. The old way will be supported through at least 10, and likely 11 too? So we can keep a reference to the old docs and people can use that.

bneradt (approved):
Looks great to me. Good test coverage.

I have some doc suggestions, but I don't want to delay things on them so I'm approving this. Let's make changes as a separate PR. I'd like to land this and get it into Yahoo so we can start trying it out and provide feedback that way.

PR #12248, opened 2025-05-16
Updated testing and fixed bug in SSLSNIConfig

bryancall:
If the `sni.yaml` configuration file contained multiple wildcard entries, an incorrect wildcard entry could be matched instead of the exact match listed first in the file.

This issue occurred because the `rank` for wildcard entries was set to 0 (the default value) and was not updated in the move constructor when the list was resized to accommodate a new entry.

### DIFF
--- a/src/iocore/net/SSLSNIConfig.cc
+++ b/src/iocore/net/SSLSNIConfig.cc
@@ -77,6 +77,7 @@ NamedElement::operator=(NamedElement &&other)
   if (this != &other) {
     match               = std::move(other.match);
     inbound_port_ranges = std::move(other.inbound_port_ranges);
+    rank                = other.rank;
   }
   return *this;
 }
--- a/src/iocore/net/unit_tests/sni_conf_test.yaml
+++ b/src/iocore/net/unit_tests/sni_conf_test.yaml
@@ -36,3 +36,13 @@ sni:
   http2_buffer_water_mark: 256
 - fqdn: foo.bar.com
   http2: false
+
+# test with mixed-case
+- fqdn: "MiXeDcAsE.foo.com"
+  http2: true
+  http2_buffer_water_mark: 256
+  inbound_port_ranges: 31337
+
+# # test with mixed-case glob
+- fqdn: "*.MiXeDcAsE.com"
+  http2: false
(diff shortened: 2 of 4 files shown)


bryancall:
Looking at the updated docs globs in the middle are no longer supported.  It adds the entry with the asterisk in the middle to the exact match list with no warning.

src/iocore/net/SSLSNIConfig.cc
```diff
@@ -77,6 +77,7 @@ NamedElement::operator=(NamedElement &&other)
   if (this != &other) {
     match               = std::move(other.match);
     inbound_port_ranges = std::move(other.inbound_port_ranges);
+    rank                = other.rank;
```
bryancall, line 80: This is the bug fix. ⬆️

The rest is just more tests and an updated config.

bryancall:
Hey @masaori335, you added in the performance enhancements and ranking to begin with, so you should know more about this code than anyone else.  Can you please review the bug fix and the added tests?

If you can think of more tests I can add them too.  I have another PR to change the PCRE regexes to use the `Regex` class (PCRE2 wrapper) and I want to make sure the tests will catch any new bugs.

I have to check to see if we are covering case were we verify the capture group.

bryancall:
@masaori335 Can you take a look at this?  I would like to get it merged in, I have another PR that relies on this one.

src/iocore/net/SSLSNIConfig.cc
masaori335: This makes sense

src/iocore/net/unit_tests/sni_conf_test.yaml
```diff
@@ -36,3 +36,16 @@ sni:
...
+  http2: true
+  http2_buffer_water_mark: 256
+  inbound_port_ranges: 31337
+
+# test with mixed-case glob
+- fqdn: "*.MiXeDcAsE.com"
+  http2: false
+
+# test glob in the middle, this will be an exact match
+- fqdn: "cat.*.com"
```
masaori335, line 51: Discussion around the #9736, we concluded don't allow this. This should be error.

> 1. Allow single left-most `*` in the fqdn field

https://docs.trafficserver.apache.org/en/latest/admin-guide/files/sni.yaml.en.html

masaori335:
The main change seems reasonable, but the wildcard in middle is we don't support. If you rush, we can merge this and deal with the issue later.

masaori335 (approved):

cmcfarlen:

PR #12249, opened 2025-05-18
Cleanup: Use enum class

hnakamur:
This pull request converts non-class enums to enum classes.
There are still many non-class enums not addressed in this pull request.
However, since the changes have grown quite large, I'm submitting it at this point.


bneradt (approved):
lgtm. Thank you for the improvement.

PR #12250, opened 2025-05-20
Add per_server.connection metrics

serrislew:
Add following metrics for origin server connections. Follows `connection.match` rules and appends it as a suffix in the metric name
- `total_connections`: count
- `active_connections`: gauge, replaces _count when enabled to keep blocking and metrics consistent
- `blocked_connections`: count to measure the blocked connections

configured with `metric_enabled` and `metric_prefix`

Also adds `port` to `proxy.config.http.per_server.connection.match` as it is in matching

### DIFF
--- a/include/iocore/net/ConnectionTracker.h
+++ b/include/iocore/net/ConnectionTracker.h
@@ -43,6 +43,7 @@
 #include "swoc/TextView.h"
 #include <tscore/MgmtDefs.h>
 #include "iocore/net/SessionSharingAPIEnums.h"
+#include "tsutil/Metrics.h"
 
 /**
  * Singleton class to keep track of the number of inbound and outbound connections.
@@ -83,6 +84,8 @@ public:
   struct GlobalConfig {
     std::chrono::seconds client_alert_delay{60}; ///< Alert delay in seconds.
     std::chrono::seconds server_alert_delay{60}; ///< Alert delay in seconds.
+    bool                 metric_enabled{false};  ///< Enabling per server metrics.
+    std::string          metric_prefix;          ///< Per server metric prefix.
   };
 
   // The names of the configuration values.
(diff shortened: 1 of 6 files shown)


include/iocore/net/ConnectionTracker.h
```diff
@@ -322,6 +333,31 @@ ConnectionTracker::Group::hash(const Key &key)
...
+    metric_name += ats_ip_nptop(&key._addr.sa, buf, sizeof(buf));
+    break;
+  case MATCH_HOST:
+    metric_name += std::string(fqdn);
+    break;
+  case MATCH_BOTH:
+    metric_name += std::string(fqdn) + "." + ats_ip_nptop(&key._addr.sa, buf, sizeof(buf));
+    break;
+  default:
+    metric_name += "";
```
bneradt, line 356: It might be better to actually remove the default so that the compiler will warn if a new value is added and the switch statement isn't updated to accomodate it.

bneradt (changes requested):
Looks interesting.

src/records/RecordsConfig.cc
```diff
@@ -367,12 +367,16 @@ static const RecordElement RecordsConfig[] =
   ,
   {RECT_CONFIG, "proxy.config.http.per_server.connection.max", RECD_INT, "0", RECU_DYNAMIC, RR_NULL, RECC_STR, "^[0-9]+$", RECA_NULL}
   ,
-  {RECT_CONFIG, "proxy.config.http.per_server.connection.match", RECD_STRING, "both", RECU_DYNAMIC, RR_NULL, RECC_STR, "^(?:ip|host|both|none)$", RECA_NULL}
```
serrislew, line 370: The match type shows "port" but not "none". Was this intentional?

bneradt: I think you're right that this must have been missed. Clearly port is supported, looking at the implementation's set of supported tags:
https://github.com/apache/trafficserver/blob/3fc0ae45f47d83c1cccc40934746c2238928c6af/src/iocore/net/ConnectionTracker.cc#L72-L74

cmcfarlen:
This PR was added to the 10.1.x branch by #12269

PR #12257, opened 2025-05-27
HTTP/2: Wait END_STREAM flag on half-closed(local) stream state

masaori335:
# Problem
Recently, we found that some clients send `HEADERS` frame followed by an empty `DATA` frame with the `END_STREAM` flag even for GET requests. 

- `HEADERS` frame with `END_HEADERS` flag
- zero-length `DATA` frame with `END_STREAM` flag

Sometimes ATS closes the stream immediately after it sends DATA frame to the client, right after handling the `HEADERS` frame. In this case, ATS treats the zero-length DATA frame as a stream error because the stream is already closed.

However, this appears to be a bug. The spec[*1] says that an endpoint should keep the stream in the `half-closed (local)` state after sending `END_STREAM` flag and wait to receive `END_STREAM` flag from the peer before closing it.

With this change, ATS waits for the `END_STREAM` flag until the `http.transaction_no_activity_timeout_in` expires.

[*1] https://datatracker.ietf.org/doc/html/rfc9113#name-stream-states

### DIFF
--- a/src/proxy/http2/Http2Stream.cc
+++ b/src/proxy/http2/Http2Stream.cc
@@ -31,6 +31,7 @@
 #include "tscore/Diags.h"
 #include "tscore/HTTPVersion.h"
 #include "tscore/ink_assert.h"
+#include "tsutil/DbgCtl.h"
 
 #include <numeric>
 
@@ -590,7 +591,7 @@ Http2Stream::transaction_done()
   SCOPED_MUTEX_LOCK(lock, this->mutex, this_ethread());
   super::transaction_done();
 
-  if (!closed) {
+  if (!closed && _state == Http2StreamState::HTTP2_STREAM_STATE_CLOSED) {
     do_io_close(); // Make sure we've been closed.  If we didn't close the _proxy_ssn session better still be open
   }
   Http2ConnectionState &state = this->get_connection_state();
@@ -1081,11 +1082,20 @@ Http2Stream::clear_io_events()
   }
 }
 
-//  release and do_io_close are the same for the HTTP/2 protocol
+/**
+  Callback from HttpSM
+
+  release and do_io_close are the same for the HTTP/2 protocol
+ */
 void
 Http2Stream::release()
 {
-  this->do_io_close();
+  if (_state == Http2StreamState::HTTP2_STREAM_STATE_CLOSED) {
+    this->do_io_close();
+    return;
+  }
+
+  Http2StreamDebug("release is called, but this stream is not ready");
 }
 
 void
--- a/src/proxy/http2/Http2ConnectionState.cc
+++ b/src/proxy/http2/Http2ConnectionState.cc
@@ -175,9 +175,15 @@ Http2ConnectionState::rcv_data_frame(const Http2Frame &frame)
 
     // Pure END_STREAM
     if (payload_length == 0) {
+      if (stream->get_state() == Http2StreamState::HTTP2_STREAM_STATE_CLOSED) {
+        stream->initiating_close();
+        return Http2Error(Http2ErrorClass::HTTP2_ERROR_CLASS_NONE);
+      }
+
       if (stream->is_read_enabled()) {
         stream->signal_read_event(VC_EVENT_READ_COMPLETE);
       }
+
       return Http2Error(Http2ErrorClass::HTTP2_ERROR_CLASS_NONE);
     }
   } else {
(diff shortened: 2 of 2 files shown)


maskit:
> The spec[*1] says that an endpoint should keep the stream in the half-closed (local) state after sending END_STREAM flag and wait to receive END_STREAM flag from the peer before closing it.

That's not true. Server can actively close the stream without waiting for END_STREAM if the server knows that receiving the rest of request is unnecessary.
https://datatracker.ietf.org/doc/html/rfc9113#section-8.1-11

There is a way to close unnecessary streams cleanly.
https://datatracker.ietf.org/doc/html/rfc9113#section-5.1-7.14.5
https://datatracker.ietf.org/doc/html/rfc9113#section-5.1-7.14.6

We should not retain resources to be used unnecessarily.

maskit:
It doesn't seem like the fundamental issue is fixed.

I have `transaction_active_timeout_in: 1` and sending a POST request that continues over 1 second causes stream errors with error code `STREAM_CLOSED`.

src/proxy/http2/Http2Stream.cc
```diff
@@ -1081,11 +1085,20 @@ Http2Stream::clear_io_events()
...
 void
 Http2Stream::release()
 {
-  this->do_io_close();
+  if (_state == Http2StreamState::HTTP2_STREAM_STATE_CLOSED) {
+    this->do_io_close();
+    return;
+  }
+
+  Http2StreamDebug("release is called, but this stream is not ready");
```
bneradt: Maybe:

```
Delaying do_io_close() until stream is in the closed state.
```

tests/gold_tests/h2/clients/h2empty_data_frame.py
```diff
@@ -0,0 +1,128 @@
...
+    ctx.verify_mode = ssl.CERT_NONE
+    ctx.set_alpn_protocols(['h2'])
+
+    # open a socket to the server and initiate TLS/SSL
+    tls_socket = socket.create_connection((SERVER_NAME, SERVER_PORT))
+    tls_socket = ctx.wrap_socket(tls_socket, server_hostname=SERVER_NAME)
+    return tls_socket
+
+
+def makerequest(port: int, path: str, n: int) -> None:
```
bneradt: nit: `make_request`

tests/gold_tests/h2/clients/h2empty_data_frame.py
```diff
@@ -0,0 +1,128 @@
...
+        tls_socket.sendall(h2_connection.data_to_send())
+
+    # tell the server we are closing the h2 connection
+    h2_connection.close_connection()
+    tls_socket.sendall(h2_connection.data_to_send())
+
+    # close the socket
+    tls_socket.close()
+
+    if not error_code == 0:
```
bneradt:
```python
if error_code != 0:
```

tests/gold_tests/h2/http2_empty_data_frame.test.py
```diff
@@ -0,0 +1,71 @@
...
+            {
+                'proxy.config.diags.debug.enabled': 1,
+                'proxy.config.diags.debug.tags': 'http2',
+                'proxy.config.ssl.server.cert.path': '{0}'.format(self._ts.Variables.SSLDir),
+                'proxy.config.ssl.server.private_key.path': '{0}'.format(self._ts.Variables.SSLDir),
+                'proxy.config.http.insert_response_via_str': 2,
+                'proxy.config.http2.active_timeout_in': 3,
+                'proxy.config.http2.stream_error_rate_threshold': 0.1  # default
+            })
+        self._ts.Disk.remap_config.AddLine('map / http://127.0.0.1:{0}/'.format(self._server.Variables.Port))
```
bneradt: `f'...'` rather than `.format()` (for new code anyway).

tests/gold_tests/h2/http2_empty_data_frame.test.py
```diff
@@ -0,0 +1,71 @@
...
+        self._ts.Disk.ssl_multicert_config.AddLine('dest_ip=* ssl_cert_name=server.pem ssl_key_name=server.key')
+
+    def __setupClient(self):
+        self._ts.Setup.CopyAs("clients/h2empty_data_frame.py", Test.RunDirectory)
+
+    def run(self):
+        tr = Test.AddTestRun()
+
+        tr.Processes.Default.StartBefore(self._ts)
+        tr.Processes.Default.StartBefore(self._server, ready=When.PortOpen(self._server.Variables.Port))
```
bneradt: Is this `ready=` condition not the default for `MakeHttpBinServer`?

tests/gold_tests/h2/http2_empty_data_frame.test.py
```diff
@@ -0,0 +1,71 @@
...
+                'proxy.config.http2.stream_error_rate_threshold': 0.1  # default
+            })
+        self._ts.Disk.remap_config.AddLine('map / http://127.0.0.1:{0}/'.format(self._server.Variables.Port))
+        self._ts.Disk.ssl_multicert_config.AddLine('dest_ip=* ssl_cert_name=server.pem ssl_key_name=server.key')
+
+    def __setupClient(self):
+        self._ts.Setup.CopyAs("clients/h2empty_data_frame.py", Test.RunDirectory)
+
+    def run(self):
+        tr = Test.AddTestRun()
```
bneradt: Adding descriptions to `AddTestRun()` is helpful when reading failed tests.

tests/gold_tests/h2/http2_empty_data_frame.test.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def __setupTS(self):
+        self._ts = Test.MakeATSProcess(f"ts", enable_tls=True, enable_cache=True)
+        self._ts.addDefaultSSLFiles()
+        self._ts.Disk.records_config.update(
+            {
+                'proxy.config.diags.debug.enabled': 1,
+                'proxy.config.diags.debug.tags': 'http2',
+                'proxy.config.ssl.server.cert.path': '{0}'.format(self._ts.Variables.SSLDir),
+                'proxy.config.ssl.server.private_key.path': '{0}'.format(self._ts.Variables.SSLDir),
```
bneradt:
```
                'proxy.config.ssl.server.cert.path': 'self._ts.Variables.SSLDir,
                'proxy.config.ssl.server.private_key.path': self._ts.Variables.SSLDir,
```

bneradt (changes requested):
Looks great to me. I just have a few minor comments. Thank you for addressing this issue.

tests/gold_tests/h2/http2_empty_data_frame.test.py
masaori335: https://github.com/apache/trafficserver/blob/132e01c4aea486c209fa5ea8e39cb34627442304/tests/gold_tests/autest-site/httpbin.test.ext#L40

Seems so! Thank you.

bneradt (approved):
Looks great.

cmcfarlen:

PR #12261, opened 2025-05-28
Add the cqssg log field for TLS group name logging

bneradt:
TLS key exchange terminology has shifted from curves to groups since future cryptographic mechanisms may not be eliptic curve (such as post quantum KEM). cqssg is a new log field being added to print the TLS group name for client side connections. The documentation is also updated to encourage uses to, long term, phase out the use of cqssu in favor of cqssg.


jasmine-nahrain (approved):

masaori335 (approved):

cmcfarlen:

PR #12264, opened 2025-05-31
Fix narrowing conversion of connection ID

bneradt:
ProxySession stores the connection ID as an int64_t, but in the HttpSM and HttpUserAgent, it was stored and transferred as an int type (typically 32 bits). This resulted in unnecessary narrowing. This updates the state machine to use an 64 bit type to alleviate the narrowing.


JosiahWI:
[approve ci autest 2]

cmcfarlen:

PR #12265, opened 2025-05-31
SnowflakeID for organizationally unique identifiers

bneradt:
UUID values are 128 bit values designed to be univerally unique identifiers across all practical uses and environments. Snowflake ID values are 64 bit values that are designed with a more limited goal: to be unique across all values in a single organization. By limiting their scope, they limit their size which is helpful for performance and storage needs.

This updates the connection id to use SnowflakeID values so that they are unique across reboots and typically unique across ATS instances in a cdn.

This also adds psfid, a logging field similar to puuid, but that logs the process's snowflake ID instead of its UUID.

Depends upon: #12264


cmcfarlen:
IMO, we shouldn't use `ink_` filenames for new code, but just some `PascalCase.cc` file like the other newer files in `tscore`.

bneradt:
[approve ci]

bneradt:
> IMO, we shouldn't use `ink_` filenames for new code, but just some `PascalCase.cc` file like the other newer files in `tscore`.

Yeah, that makes sense. I followed the precedent set by the uuid logic because this seemed similar. But I agree...I'll change this to SnowflakeID.{cc,h}.

src/tscore/SnowflakeID.cc
```diff
@@ -0,0 +1,138 @@
...
+SnowflakeIDUtils::get_string() const
+{
+  if (m_id_string.empty()) {
+    // Base64 encode the snowflake ID as m_id_string.
+    size_t const max_encoded_size = ats_base64_encode_dstlen(sizeof(m_snowflake_value));
+    auto         encoded_buffer   = std::make_unique<char[]>(max_encoded_size);
+    size_t       encoded_length   = 0;
+    auto        *snowflake_char_p = reinterpret_cast<char const *>(&m_snowflake_value);
+    if (ats_base64_encode(snowflake_char_p, sizeof(m_snowflake_value), encoded_buffer.get(), max_encoded_size, &encoded_length)) {
+      m_id_string = std::string(encoded_buffer.get(), encoded_length);
```
cmcfarlen: This function isn't called from any const functions.  So, you could probably remove `const` on this function and remove `mutable` from `m_id_string`.  Or, make the callers `const`

src/tscore/SnowflakeID.cc
```diff
@@ -0,0 +1,138 @@
...
+  global_machine_id = machine_id;
+}
+
+std::string_view
+SnowflakeIDUtils::get_string() const
+{
+  if (m_id_string.empty()) {
+    // Base64 encode the snowflake ID as m_id_string.
+    size_t const max_encoded_size = ats_base64_encode_dstlen(sizeof(m_snowflake_value));
+    auto         encoded_buffer   = std::make_unique<char[]>(max_encoded_size);
```
cmcfarlen: I recommend making `encoded_buffer` a `std::array<char, max_encoded_size>`. This is fine since `ats_base64_encode_dstlen` is constexpr so the size can be known at compile time.  This will avoid an allocation.

src/tscore/SnowflakeID.cc
```diff
@@ -0,0 +1,138 @@
...
+    }
+  }
+  return m_id_string;
+}
+
+uint64_t
+SnowflakeID::generate_next_snowflake_value()
+{
+  ink_release_assert(SnowflakeIDUtils::get_machine_id() != 0);
+  snowflake_t new_snowflake{.value = 0};
```
cmcfarlen: you can default initialize this `new_snowflake{}` and get the same.  You could also put a default value on `.value` in the type to be explicit about it.

src/tscore/SnowflakeID.cc
```diff
@@ -0,0 +1,138 @@
...
+#include <cstdint>
+#include <cinttypes>
+#include <memory>
+#include <mutex>
+#include <string_view>
+
+std::atomic<uint64_t> SnowflakeIDUtils::global_machine_id{0};
+
+uint64_t   SnowflakeID::m_last_timestamp{0};
+uint64_t   SnowflakeID::m_last_sequence{0};
```
cmcfarlen: Instead of having some global mutable state here, consider creating a SnowflakeID generator class that manages this.  Maybe that instance would still be global mutable state, but idk if it would have to be.  It would still be cleaner than these globals.

cmcfarlen (commented):
Still working through this, but here are some comments.

src/tscore/SnowflakeID.cc
bneradt: Can you explain your reasoning here more? const seems conceptually correct from an API standpoint: `m_id_string` is the cached value of the lazilly computed string, of which there is only one correct value for the id. Isn't this one of the proper uses of mutable? If this is used in other places eventually (snowflake id is a pretty generic concept), it would be frustrating for this to be non-const if someone is trying to call this in a const function.

src/tscore/SnowflakeID.cc
bneradt: Good idea.

src/tscore/SnowflakeID.cc
bneradt: Just to make sure I'm tracking, you prefer a global singleton to hold these?

src/tscore/SnowflakeID.cc
cmcfarlen: Neither `SnowflakeID::get_string` nor `SnowflakeIdNoSequence::get_string` are const which call this.  The easiest fix would be to make those const.

src/tscore/SnowflakeID.cc
cmcfarlen: Ideally we don't add any global mutable state, and the snowflakeid generator is owned by the object the generates the ids.  Short of that, separating the id generation into its own class and having a global one that is used by default is preferable and would allow other implementations of ID generation in the future.

src/tscore/SnowflakeID.cc
bneradt: I see. Yes, it is good to make those const as well. Thanks for catching this.

src/tscore/SnowflakeID.cc
bneradt: Sounds good.

cmcfarlen (approved):
Cool. Thanks!

PR #12267, opened 2025-06-02
Include records.yaml line and column in the logs when error is found.

brbzull0:
For instance, with the following config:

```yaml
records:
  http:
    per_server:
      connection:
        match: both*ioajs
        new:
          field: 11:
```

We will get a warning log with the line and the column from the records.yaml file exactly where the error was detected.

In the above example, match and new.field have different sort of errors but both now log the line+column.

```
WARNING: We have found the following issues when reading the records node:
 Warn:   Warn: proxy.config.http.per_server.connection.match - Validity Check error at line=37, col=9. Pattern '^(?:ip|host|both|none)$' failed against 'both*ioajs'. Default value will be used
  Warn: Ignoring field 'field' [proxy.config.http.per_server.connection.new.field] at line=39, col=11. Not registered and Unknown tag type '?
```

### DIFF
--- a/include/records/RecYAMLDefs.h
+++ b/include/records/RecYAMLDefs.h
@@ -83,6 +83,16 @@ struct CfgNode {
     }
   }
 
+  std::string
+  mark_as_view(swoc::TextView fmt = "Line: {}, Column: {}") const
+  {
+    swoc::LocalBufferWriter<128> lbw;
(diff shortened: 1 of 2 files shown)


brbzull0:
[approve ci autest 1]

masaori335 (approved):
Thank you!

cmcfarlen:

PR #12268, opened 2025-06-02
Improve accuracy of alloc metrics with freelists

cmcfarlen:
Allocator metrics are super handy for understanding memory pressure on ATS, but if proxy allocators are in use, then the metrics were not accurate since the proxy allocators would not return it to the base allocator.  This PR corrects the accuracy of the allocator metrics even if ProxyAllocators are in use.


bneradt (approved):

cmcfarlen:

PR #12271, opened 2025-06-03
Try both old(9.2) and new cache key generation

cmcfarlen:
PR #11519 was a security related PR that removed support for URL param, but also introduced a change that altered the cache key produced from a URL, even if the URL did not have a param component.  While this doesn't affect the cache disk layout or cause the cache to reset, it does cause any objects stored by prior versions of ATS to not be found.  This has the same effect as resetting the cache when upgrading to ATS10.

This PR mitigates this by introducing a cache key lookup compatibility option.  If this option is set, then when ATS encounters a cache miss, it will retry the lookup using the prior versions URL cachekey algorithm.  If it then gets a hit, it increments a metric and then proceeds as if the first attempt was found.  If it still gets a miss, the transaction will continue on as a miss.  In this mode, cache writes will always use the new algorithm and so over time, the need for having this mode enabled goes away.

This route was chosen over reverting the PR because of the security nature of the change and it could be a strategy for future releases to be able to make similar changes to cache and offer a backward compatibility mode for lookups that can be used until the cache wraps.

### DIFF
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -2390,6 +2390,19 @@ Cache Control
    Establishes a guaranteed maximum lifetime boundary for object freshness.
    Setting this to ``0`` disables the feature.
 
+.. ts:cv:: CONFIG proxy.config.http.cache.try_compat_key_read INT 0
+   :reloadable:
+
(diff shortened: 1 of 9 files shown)


ezelkow1 (approved):
tested here on 10.0.4 and looked good

cmcfarlen:
This was added to milestone 10.1.0 via #12283

PR #12223, opened 2025-05-06
Add server_group_list option to sni

jasmine-nahrain:
Add ability to specify TLS key exchange group in sni.yaml

PR #12273, opened 2025-06-03
Cherrypick #12223 & #12110

jasmine-nahrain:


maskit:
We probably need to cherry-pick those separately to generate changelog.
https://github.com/apache/trafficserver/blob/master/tools/git/changelog.pl

maskit:
Maybe not for changelog... but commits should be separated so that we can revert them separately.

cmcfarlen:
I picked these individually.  Thanks!

PR #12274, opened 2025-06-04
healthchecks: fix event matching for directory events

bneradt:
For directory changes, we compared the event's name (which indicates which file in the directory changed) against the config's basename but limited the check to basename_len via strncmp. Unwittingly, though, if event->name was more bytes than basename_len, we could incorrectly match on the substring.

For instance, with a config like this:

```
/acme /tmp/sb/healthchecks/acme text/plain 200 404
/acme-ssl /tmp/sb/healthchecks/acme-ssl text/plain 200 404
```

An event for acme-ssl changes would accidentally match for acme, because acme's basename_len is 4, while event->name is acme-ssl, thus strncmp on the event would incorrectly indicate a match because the first four bytes were the same for each.

This extracts event/config file matching to a function so it is easier to understand and ensures that event->name and basename are the same length to avoid the accidental match.

This also adds an autest for healthchecks which reproduces the above issue and verifies basic file event processing functionality.

### DIFF
--- a/plugins/healthchecks/healthchecks.cc
+++ b/plugins/healthchecks/healthchecks.cc
@@ -166,25 +166,54 @@ setup_watchers(int fd)
 /* Separate thread to monitor status files for reload */
 #define INOTIFY_BUFLEN (1024 * sizeof(struct inotify_event))
 
+/** Determine whether @a event applies to the @a finfo configured file. */
+static bool
+event_matches_config(struct inotify_event *event, HCFileInfo *finfo)
+{
+  if (!finfo || !event) {
+    return false;
+  }
+  if (event->wd == finfo->wd) {
+    // Easy: the event is for this configured file we are watching.
+    return true;
+  }
+  if (event->wd != finfo->dir->wd) {
+    // The event is not for this file, nor for the parent directory. No match.
+    return false;
+  }
+
+  // The event applies to a change in a directory that contains files we are
+  // configured to watch. Does the directory event apply to this file?
+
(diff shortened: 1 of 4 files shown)


plugins/healthchecks/healthchecks.cc
```diff
@@ -166,25 +166,54 @@ setup_watchers(int fd)
...
+  // The event applies to a change in a directory that contains files we are
+  // configured to watch. Does the directory event apply to this file?
+
+  if (!finfo->basename || finfo->basename_len <= 0) {
+    // This configured finfo is not for a specific file.
+    return false;
+  }
+  if (strnlen(event->name, NAME_MAX) != finfo->basename_len) {
+    return false;
+  }
```
bneradt, line 194: The addition of this length check is the functional change. This is the fix for the incorrect matching bug described in the description.

plugins/healthchecks/healthchecks.cc
```diff
@@ -220,8 +249,7 @@ hc_thread(void *data ATS_UNUSED)
         struct inotify_event *event = (struct inotify_event *)&buffer[i];
         HCFileInfo           *finfo = g_config;
 
-        while (finfo && !((event->wd == finfo->wd) ||
-                          ((event->wd == finfo->dir->wd) && !strncmp(event->name, finfo->basename, finfo->basename_len)))) {
```
bneradt, line 224: This loop's logic became `event_matches_config`

zwoop:
@cmcfarlen this should go into 10.1.0

cmcfarlen:

PR #12278, opened 2025-06-10
Update url_m_loc variable to the newly generated one in set-destination

jasmine-nahrain:
When `set-destination URL` is called, it creates a new url_m_loc with the new url and assigns that to the buffer. The requestUrl is never updated therefore any subsequent set-destination call will use the incorrect url_m_loc. This change tracks if the url has been changed, if so it will get the url_m_loc from the http hdr which has the updated loc.
This issue is also present in 9.2

### DIFF
--- a/plugins/header_rewrite/operators.cc
+++ b/plugins/header_rewrite/operators.cc
@@ -344,7 +344,7 @@ OperatorSetDestination::exec(const Resources &res) const
         if (TSUrlCreate(bufp, &new_url_loc) == TS_SUCCESS && TSUrlParse(bufp, new_url_loc, &start, end) == TS_PARSE_DONE &&
             TSHttpHdrUrlSet(bufp, res.hdr_loc, new_url_loc) == TS_SUCCESS) {
           // This case creates a new_url_loc which should be used for any other set-destination case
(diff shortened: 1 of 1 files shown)


plugins/header_rewrite/operators.cc
```diff
@@ -343,6 +343,8 @@ OperatorSetDestination::exec(const Resources &res) const
         TSMLoc      new_url_loc;
         if (TSUrlCreate(bufp, &new_url_loc) == TS_SUCCESS && TSUrlParse(bufp, new_url_loc, &start, end) == TS_PARSE_DONE &&
             TSHttpHdrUrlSet(bufp, res.hdr_loc, new_url_loc) == TS_SUCCESS) {
+          // This case creates a new_url_loc which should be used for any other set-destination case
+          res._rri->requestUrl = reinterpret_cast<TSMLoc>(new_url_loc);
```
masaori335: It seems like we don't need this cast, because the `new_url_loc` is `TSMLoc`.

cmcfarlen:

zwoop:
Ugh, I messed up the Milestone on this, sorry

Issue #12262, opened 2025-05-29
traffic_ctl prints confusing error messages even though the command seems to work.


cmcfarlen:
```
traffic_ctl config reload
Server Error found:
[9] Error during execution
- [0] Errors while reloading configurations.
- [0] Errors while reloading configurations.
- [0] Unknown file change .
```

We've been seeing this for a while now.

brbzull0:
Interesting.

All this errors comes from the `FileManager` inside `traffic_server`,  if you enable debug with tag `filemanager`, more details should be posted on the `traffic.out`.

`Unknown file change .` seems to be printing an empty config name(asociated with the reload file).


~~@cmcfarlen  which ATS version are you using here?~~ I was able to reproduce it with master.

I'll dig into this locally.

PR #12282, opened 2025-06-11
Fix FileManager to handle file updates properly.

brbzull0:
The issue was related to the files that aren't associated with a record name. The code wasn't taking into consideration that there are files that
are bound to a parent file and not to a record. This fixes the issue and avoid reporting misleading errors to the rpc caller.

Also:
- Changes a function to take a `const char*` instead of a `char*`, this is done to avoid a `const_cast`. Also makes more sense. 

- `getConfigObj()` is removed, unused function.

Fixes issue #12262


cmcfarlen (approved):
Cool.  Thanks for fixing this!

cmcfarlen:
Added to milestone 10.1.0 via #12326

PR #12285, opened 2025-06-12
Fix for issue 12279 - verify_config - Fix error checking when parsing records.yaml

### LINKED ISSUE #12279: traffic_server -C verify_config fails on 10.0.x branch
Hello!

10.0.x branch.

Command `traffic_server -C verify_config` fails (return code is not 0) when there are env. variables that overrides config parameters.

For example in our case we have following message:
```
[Jun 10 10:19:39.270] traffic_server WARNING: We have found the following issues when reading the records node:
 Debug:   Debug: 'proxy.config.cache.ram_cache.size' was override with '17138964234' using an env variable
  Debug: 'proxy.config.diags.debug.tags' was override with '' using an env variable
```

brbzull0:
Fixes #12279

### DIFF
--- a/src/traffic_server/traffic_server.cc
+++ b/src/traffic_server/traffic_server.cc
@@ -964,7 +964,7 @@ cmd_verify(char * /* cmd ATS_UNUSED */)
     fprintf(stderr, "INFO: Successfully loaded %s\n\n", ts::filename::REMAP);
   }
 
-  if (auto ret = RecReadYamlConfigFile(); !ret.empty()) {
+  if (auto ret = RecReadYamlConfigFile(); !ret) {
     exitStatus |= (1 << 1);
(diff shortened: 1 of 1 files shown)


src/traffic_server/traffic_server.cc
```diff
@@ -964,7 +964,7 @@ cmd_verify(char * /* cmd ATS_UNUSED */)
     fprintf(stderr, "INFO: Successfully loaded %s\n\n", ts::filename::REMAP);
   }
 
-  if (auto ret = RecReadYamlConfigFile(); !ret.empty()) {
```
brbzull0, line 967: errata may contain some notes and warnings which shouldn't be considered as errors. Checking for `empty()` wasn't the right check. Fixed now.

randall (approved):

PR #12286, opened 2025-06-12
Add some more hrw4u help

mlibbey:
Added a missing python requirement, and a few more steps to get hrw4u to build for the first time.

Also included a few more doc hints on usage.


doc/admin-guide/configuration/hrw4u.en.rst
```diff
@@ -220,6 +225,13 @@ set-debug         set-debug()                  Enables ATS txn debug
 skip-remap        skip-remap()                 Skip remap processing (open proxy)
 ================= ============================ ================================
 
+String concatenations
+---------------------
+
+You can concatenate values using strings, condition values and variable expansions on the same line in
+operators using. For instance, `outbound.req.CustomHeader “Hello from {inbound.ip}:{inbound.port}”`. As
+a result, the set-redirect's `[QSA]` flag would be implemented as `set-redirect(302, "https://...{inbound.url.query}")``
```
zwoop: Maybe make sure the ? is here? I'm not sure if ATS's query value includes the ? or not... I want to say not but I can test tomorrow.

tools/hrw4u/bootstrap.sh
```diff
@@ -20,12 +20,22 @@ set -e
 
 VENV_NAME="hrw4u"
 
+if ! which antlr; then
```
zwoop, line 23: Does this work? Don't you have to put that into backticks or $() at least ?

zwoop (commented):
Thanks, a few comments inline.

tools/hrw4u/bootstrap.sh
mlibbey: I got the
if ! command; from
https://www.shellcheck.net/ -- the PR passes shellcheck without issues. When I replace `which antlr` with `which sldkjcdslkfjsdlfkj` I get the echos.

doc/admin-guide/configuration/hrw4u.en.rst
mlibbey: Updated.

Issue #11758, opened 2024-09-09
Blind tunneling breaks with Google Chrome and TLS Kyber cipher


fsecure-kilkanen:
# Executive summary

When TLS Client Hello is split into two TCP segments and segments arrive to TrafficServer in different parts, blind tunneling flow breaks.

# Details

## Use case

Trafficserver runs as a transparent forward proxy for TLS traffic. A custom plugin uses SSL_CERT_HOOK to determine destination hostname of the TLS connection.

If the destination is accepted, connection is converted into a blind tunnel by calling `TSVConnTunnel(sslvc)` and `TSVConnReenable(sslvc)` in the hook callback function.

## Affected versions

Reproduced with TrafficServer 8.x branch and TrafficServer 9.2.4. Issue should also exist in 10.x branch because affected code paths have not changed.

## The issue

When a connection is started with Kyber cipher, the TLS Client Hello packet contains more key material than previous ciphers. With Kyber cipher, the TLS Client Hello packet is 1700-2100 bytes in size.

Since the network MTU is usually 1500, it means that the TLS Client Hello is split into two TCP segments. In some conditions, clients' network connectivity is slow such that the gap between the first and second segment of Client Hello is big enough to trigger the bug.

https://github.com/apache/trafficserver/blob/master/src/iocore/net/SSLNetVConnection.cc#L423 is the place where TLS Client Hello is read. This is a call to read data from socket that is for the incoming TCP connection. There is no guarantee that this call will return both segments of the Client Hello. If data for complete Client Hello is received here, the blind tunnel operation is succesful.

However, if only first segment of Client Hello is received, then the following sequence of events happen:

https://github.com/apache/trafficserver/blob/master/src/iocore/net/SSLNetVConnection.cc#L1365 OpenSSL `ssl_accept()` is called. It returns `SSL_ERROR_WANT_READ`, which is returned as `SSL_HANDSHAKE_WANT_READ`.

https://github.com/apache/trafficserver/blob/master/src/iocore/net/SSLNetVConnection.cc#L661 is the branch taken in this case. Processing goes to `update_rbio` at https://github.com/apache/trafficserver/blob/master/src/iocore/net/SSLNetVConnection.cc#L529 with `move_to_socket` set to `true`.

In `update_rbio`, OpenSSL BIO is set to the file descriptor of the connection, and existing handshake buffers are deallocated.

[...]

PR #12290, opened 2025-06-13
Support CLIENT_HELLO split across multiple packets

bneradt:
Our TLS CLIENT_HELLO processing logic assumed all CLIENT_HELLO bytes came in a single TCP packet. However, with more recent cryptographic ciphers, the CLIENT_HELLO is often greater than the standard 1,500 byte MTU, so the CLIENT_HELLO is being delivered in multiple packets. This updates our logic to properly buffer and parse data across multiple socket reads.

Fixes: #11758

### DIFF
--- a/src/iocore/net/SSLNetVConnection.cc
+++ b/src/iocore/net/SSLNetVConnection.cc
@@ -397,13 +397,13 @@ proxy_protocol_bypass:
   if (r > 0) {
     this->handShakeBuffer->fill(r);
 
-    char *start              = this->handShakeReader->start();
-    char *end                = this->handShakeReader->end();
-    this->handShakeBioStored = end - start;
+    auto const total_chain_size = this->handShakeReader->read_avail();
+    this->handShakeBioStored    = total_chain_size;
+    char *buffer_for_bio        = this->_getCoalescedHandShakeBuffer(total_chain_size);
 
     // Sets up the buffer as a read only bio target
     // Must be reset on each read
-    BIO *rbio = BIO_new_mem_buf(start, this->handShakeBioStored);
+    BIO *rbio = BIO_new_mem_buf(buffer_for_bio, this->handShakeBioStored);
     BIO_set_mem_eof_return(rbio, -1);
     SSL_set0_rbio(this->ssl, rbio);
   } else {
@@ -436,15 +436,14 @@ SSLNetVConnection::update_rbio(bool move_to_socket)
     this->handShakeBioStored = 0;
     // Load up the next block if present
     if (this->handShakeReader->is_read_avail_more_than(0)) {
-      // Setup the next iobuffer block to drain
-      char *start              = this->handShakeReader->start();
-      char *end                = this->handShakeReader->end();
-      this->handShakeBioStored = end - start;
+      auto const total_chain_size = this->handShakeReader->read_avail();
+      this->handShakeBioStored    = total_chain_size;
+      char *buffer_for_bio        = this->_getCoalescedHandShakeBuffer(total_chain_size);
       Dbg(dbg_ctl_ssl, "Adding %d bytes to the ssl rbio", this->handShakeBioStored);
 
       // Sets up the buffer as a read only bio target
       // Must be reset on each read
-      BIO *rbio = BIO_new_mem_buf(start, this->handShakeBioStored);
+      BIO *rbio = BIO_new_mem_buf(buffer_for_bio, this->handShakeBioStored);
       BIO_set_mem_eof_return(rbio, -1);
       SSL_set0_rbio(this->ssl, rbio);
       retval = true;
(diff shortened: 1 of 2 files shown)


tests/gold_tests/tls/tls_tunnel.test.py
```diff
@@ -284,6 +301,35 @@
...
+tr = Test.AddTestRun("Single large CLIENT_HELLO blind tunnel test")
+tr.TimeOut = 5
+tr.Setup.Copy('split_client_hello.py')
+tr.Setup.Copy('receive_split_client_hello.py')
+p = tr.Processes.Default
+p.StartBefore(split_client_hello_server, ready=When.PortOpen(server_port))
+p.Command = f'{sys.executable} split_client_hello.py 127.0.0.1 {ts.Variables.ssl_port} -s 0'
+p.ReturnCode = 0
+tr.StillRunningAfter = ts
+tr.StillRunningAfter = dns
```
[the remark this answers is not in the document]
bneradt: Sorry Copilot, you just don't understand autest's declarative nature.

tests/gold_tests/tls/receive_split_client_hello.py
```diff
@@ -0,0 +1,129 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import argparse
+import socket
+import sys
+
+# See split_client_hello.py for the original CLIENT_HELLO packet.
+EXPECTED_CLIENT_HELLO_LENGTH = 1993
```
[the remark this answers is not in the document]
bneradt: True. I'll add a comment.

src/iocore/net/SSLNetVConnection.cc
```diff
@@ -568,7 +578,13 @@ SSLNetVConnection::net_read_io(NetHandler *nh)
...
       // move over to the socket if we haven't already
       if (this->handShakeBuffer != nullptr) {
-        read.triggered = update_rbio(true);
+        bool const in_client_hello =
+          this->get_handshake_hook_state() == TLSEventSupport::SSLHandshakeHookState::HANDSHAKE_HOOKS_CLIENT_HELLO;
+        // Only transfer buffers to the socket once the CLIENT_HELLO is
+        // finished. We need to keep our buffers updated until then in case we
+        // enter tunnel mode.
+        Dbg(dbg_ctl_ssl, "Updating our buffers, in CLIENT_HELLO: %s", in_client_hello ? "true" : "false");
+        read.triggered = update_rbio(!in_client_hello);
```
bneradt, line 586: Only transition to socket mode from our manually fed buffer if we're not still processing the CLIENT_HELLO. While we process the CLIENT_HELLO, we need to keep buffering it in case we want to replay it in the blind tunnel case.

src/iocore/net/SSLNetVConnection.cc
```diff
@@ -1211,39 +1227,39 @@ SSLNetVConnection::sslServerHandShakeEvent(int &err)
...
+  bool const in_client_hello =
+    this->get_handshake_hook_state() == TLSEventSupport::SSLHandshakeHookState::HANDSHAKE_HOOKS_CLIENT_HELLO;
+  // We only feed CLIENT_HELLO bytes into our temporary buffers. If we are past
+  // the CLIENT_HELLO, then no need to buffer.
+  if (in_client_hello && this->handShakeReader) {
     if (BIO_eof(SSL_get_rbio(this->ssl))) { // No more data in the buffer
-      // Is this the first read?
-      if (!this->handShakeReader->is_read_avail_more_than(0) && !this->handShakeHolder->is_read_avail_more_than(0)) {
+                                            // Is this the first read?
 #if TS_USE_TLS_ASYNC
```
bneradt, line 1236: If the CLIENT_HELLO is split on multiple packets, then we need to call `this->read_raw_data` for each packet (see 1246). Before this change, we assumed we would only ever need to call it once, thus we only called it if handShakeReader/Holder was empty. This change makes us do the raw read for each packet that comes in, so we can populate our buffered CLIENT_HELLO and feed it to the SSL socket.

src/iocore/net/SSLNetVConnection.cc
```diff
@@ -1211,39 +1227,39 @@ SSLNetVConnection::sslServerHandShakeEvent(int &err)
...
+          SSLVCDebug(this, "SSL handshake error: read_retval=%d", retval);
           return EVENT_ERROR;
         }
-      } else {
-        update_rbio(false);
+      } else if (retval == 0) {
+        // EOF, go away, we stopped in the handshake
+        SSLVCDebug(this, "SSL handshake error: EOF");
+        return EVENT_ERROR;
       }
```
bneradt, line 1245: the read_faw_data logic above does what update_rbio would have been doing (read from the socket and populate our buffers), so no need to call it here.

tomuxi:
I guess no hope in getting this fix backported to older major versions?

bneradt:
> I guess no hope in getting this fix backported to older major versions?

The cherry-pick for 9.x is non-trivial, but I can look into it.

src/iocore/net/SSLNetVConnection.cc
```diff
@@ -433,6 +440,7 @@ SSLNetVConnection::update_rbio(bool move_to_socket)
       char *start              = this->handShakeReader->start();
       char *end                = this->handShakeReader->end();
```
moonchen: This will give the end of the current block, so the client hello must fit within the first IO buffer block.  The block size defaults to 32k bytes.  We should document this limitation and instruct the administrator to increase `proxy.config.ssl.misc.io.max_buffer_index` when they want to accommodate larger client hello messages.

bneradt: Good point. I added a commit that will handle the situation if the data spans multiple blocks.

bneradt:
> > I guess no hope in getting this fix backported to older major versions?
> 
> The cherry-pick for 9.x is non-trivial, but I can look into it.



> > I guess no hope in getting this fix backported to older major versions?
> 
> The cherry-pick for 9.x is non-trivial, but I can look into it.

I made a 9.2.x backport:
https://github.com/apache/trafficserver/pull/12319

moonchen (approved):

cmcfarlen:

PR #12297, opened 2025-06-16
Add max inclusion depth to esi plugin for 10.1.x

shukitchan:


cmcfarlen (approved):

shukitchan:
this is probably not needed.

PR #12302, opened 2025-06-17
stats_over_http: Add Prometheus format option

bneradt:
This adds the Prometheus format as an optional output to the stats_over_http plugin. This also adds /json, /csv, and /prometheus path suffix support in case using the Accept header is problematic.


bneradt:
[approve ci]

ezelkow1 (approved):
lgtm. Tested with both the accept header and new path options, could get the right types each time

cmcfarlen:

PR #12306, opened 2025-06-18
hrw4u: add inbound.cookie

mlibbey:
This allows an configuration to either write conditionals with cookie (which stems from original header_rewrite), or a new inbound.cookie -- which might be more expected given the other 'contextual' objects.


zwoop (changes requested):
One thought, one "fix" required :)

1. Is || good in the documentation here? Makes it look like you have to do both in an "or" clause maybe ?
2. You also have to update the AST file, since you change the input file. The way to do that is to run the below command.

```
./scripts/testcase.py --update
```

This will update all test cases if they have changed, so make sure you don't accidentally update something where it shouldn't update (won't be a problem here).

bryancall (approved):

mlibbey:
Raw cookie seemed confusing. So, instead have (in|out)bound.cookie ... made docs easier to explain too...
Testcases updated :)

zwoop (approved):

bryancall (approved):

PR #12307, opened 2025-06-19
Fix potential mutex destroy while locked in origin_server_auth plugin

cmcfarlen:
It's possible depending on refcounts of underlying mutex objects that when tearing down an instance of the origin_server_auth plugin that we might call `pthread_mutex_destroy` while holding a lock on the same mutex.  This causes ATS to abort.

The fix involves deferring the continuation destroy until after the mutex is unlocked.  At that time, it doesn't matter what the refcount is and it is safe to destroy the mutex.

### DIFF
--- a/plugins/origin_server_auth/origin_server_auth.cc
+++ b/plugins/origin_server_auth/origin_server_auth.cc
@@ -48,6 +48,7 @@
 #include <ts/remap.h>
 #include <ts/remap_version.h>
 #include <tsutil/TsSharedMutex.h>
+#include <utility>
 #include "ts/apidefs.h"
 #include "tscore/ink_config.h"
 #include "swoc/TextView.h"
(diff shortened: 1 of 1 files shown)


plugins/origin_server_auth/origin_server_auth.cc
```diff
@@ -1269,11 +1276,18 @@ TSRemapNewInstance(int argc, char *argv[], void **ih, char * /* errbuf ATS_UNUSE
 void
 TSRemapDeleteInstance(void *ih)
 {
-  S3Config *s3           = static_cast<S3Config *>(ih);
-  TSMutex   s3_mutex_ptr = s3->config_reloader_mutex();
-  TSMutexLock(s3_mutex_ptr);
-  delete s3;
-  TSMutexUnlock(s3_mutex_ptr);
+  S3Config *s3 = static_cast<S3Config *>(ih);
```
bryancall: Couldn't this all be done in the destructor for s3?

cmcfarlen: Maybe, but the code to avoid the race condition has to modify s3 and then fix the order of the destruction.

cmcfarlen:
[approve ci autest 1]

cmcfarlen:

PR #12308, opened 2025-06-20
traffic_ctl - Set the appropriate error exit code.

### LINKED ISSUE #12210: traffic_ctl error handling
I ran `traffic_ctl config set` command without enough privilege accidentally, however  `traffic_ctl` was just quiet. Also, return code was `0`. It'd be very helpful show error message or something if an error happens on running command.

brbzull0:
Beside errors in the logic which was already covered this also marks the return code to error when the RPC sends back a response with some sort of failure.
With this change traffic_ctl callers can now validate the exit code also considering the RPC errors/failures.

Fixes #12210


masaori335 (approved):
Great! This is so helpful!

cmcfarlen:
Added to milestone 10.1.0 via #12327

PR #12309, opened 2025-06-20
hrw4u -- accept text from stdin

mlibbey:
Sometimes it it nice to have hrw4u accept input from stdin instead of a file


zwoop:
[approve ci autest 1]

PR #12316, opened 2025-06-25
Add ability for stats over http to run as a remap plugin

ezelkow1:
Tested as global and remap, with path as param as well as config file

### DIFF
--- a/doc/admin-guide/plugins/stats_over_http.en.rst
+++ b/doc/admin-guide/plugins/stats_over_http.en.rst
@@ -41,7 +41,11 @@ default URL::
 
 where host and port is the hostname/IP and port number of the server.
 
+This plugin can also run as a remap plugin with an optional stats over http config file as a parameter::
 
+    map / http://target @plugin=stats_over_http.so @pparam=config.file
+
+Much like the global version this pparam can also be the path to be used on the remap in order to get the stats
 
 Plugin Options
 ==============
--- a/plugins/stats_over_http/stats_over_http.cc
+++ b/plugins/stats_over_http/stats_over_http.cc
@@ -43,6 +43,7 @@
 #include <zlib.h>
 
 #include <ts/remap.h>
+#include <ts/remap_version.h>
 #include "swoc/TextView.h"
 #include "tscore/ink_config.h"
 #include <tsutil/ts_ip.h>
@@ -653,7 +654,7 @@ stats_dostuff(TSCont contp, TSEvent event, void *edata)
 }
 
 static int
-stats_origin(TSCont contp, TSEvent /* event ATS_UNUSED */, void *edata)
+stats_origin(config_t *this_config, TSEvent /* event ATS_UNUSED */, void *edata, bool is_remap)
 {
   TSCont          icontp;
   stats_state    *my_state;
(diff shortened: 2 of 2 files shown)


ezelkow1:
I suppose this should also have a docs change, which I can add tomorrow, but any feedback appreciated since I whipped this up pretty quick from what I vaguely remember doing before for this

bryancall:
[approve ci autest]

doc/admin-guide/plugins/stats_over_http.en.rst
```diff
@@ -41,7 +41,11 @@ default URL::
 
 where host and port is the hostname/IP and port number of the server.
 
+This plugin can also run as a remap plugin with an optional stats over http config file as a parameter::
 
+    map / http://target @plugin=stats_over_http.so @pparam=config.file
```
zwoop, line 46: Nitpicky, but this seems like a dangerous example :). Shouldn't we use a more specific path here, like

```
map /_stats_over_http ...
```

or maybe even more specific (since the above captures the path for all hosts)?

Also, maybe mention that the "target" doesn't imply / mean anything, and can be whatever ?

plugins/stats_over_http/stats_over_http.cc
```diff
@@ -43,6 +43,7 @@
 #include <zlib.h>
 
 #include <ts/remap.h>
+#include <ts/remap_version.h>
```
zwoop, line 46: Separate issue, but seeing this, makes me thing we ought to add the define from this file into ts/remap.h.

plugins/stats_over_http/stats_over_http.cc
```diff
@@ -800,10 +802,54 @@ stats_origin(TSCont contp, TSEvent /* event ATS_UNUSED */, void *edata)
...
+      (config_holder->config_path == nullptr)) {
+    config_holder->config->stats_path = argv[2] + ('/' == argv[2][0] ? 1 : 0);
+  } else if ((config_holder->config != nullptr) && (config_holder->config->stats_path.empty())) {
+    config_holder->config->stats_path = DEFAULT_URL_PATH;
+  }
+  return TS_SUCCESS;
+}
+
+TSRemapStatus
+TSRemapDoRemap(void *id, TSHttpTxn rh, TSRemapRequestInfo * /* ATS_UNUSED */)
```
zwoop, line 845: Why `id` ? Typically we call this `ih`

plugins/stats_over_http/stats_over_http.cc
```diff
@@ -800,10 +802,54 @@ stats_origin(TSCont contp, TSEvent /* event ATS_UNUSED */, void *edata)
...
+  config_holder = new_config_holder(argc > 1 ? argv[2] : nullptr);
+  *instance     = static_cast<void *>(config_holder);
+  /* Path was not set during load, so the param was not a config file, we also
+    have an argument so it must be the path, set it here.  Otherwise if no argument
+    then use the default _stats path */
+  if ((config_holder->config != nullptr) && (config_holder->config->stats_path.empty()) && (argc > 1) &&
+      (config_holder->config_path == nullptr)) {
+    config_holder->config->stats_path = argv[2] + ('/' == argv[2][0] ? 1 : 0);
+  } else if ((config_holder->config != nullptr) && (config_holder->config->stats_path.empty())) {
+    config_holder->config->stats_path = DEFAULT_URL_PATH;
```
zwoop, line 839: Maybe I'm not reading this right, but shouldn't the default path here be the "from_url" path in the RRI (the remap.config "from URL") ?

It kinda feels weird that you can override the path  with the pparam or config file. What would happen if I do

```
map http://localhost/_stats_over_http http://default @plugin=stats_over_http.so @pparam=/bar
```

How could that ever match on the path now ?

PR #12317, opened 2025-06-25
Add metric when parent marked down from inactive timeout

serrislew:
Similar to `parent_marked_down_count`, add metric for when parent is marked down because of inactive timeouts. Provides more visibility


bneradt (approved):

cmcfarlen:

PR #12318, opened 2025-06-25
Increment parent mark down if config allows

serrislew:
Current metric is inaccurate if config disables parent mark down. Should only increment if ATS is actually marking a parent down


bryancall (approved):

cmcfarlen:

PR #12322, opened 2025-06-30
gdb-helpers.py: Add tunnel events to the history parser

bneradt:
This adds the TS_EVENT_HTTP_TUNNEL_EVENT_* events to the history parser in gdb-helpers.py

### DIFF
--- a/tools/gdb-helpers.py
+++ b/tools/gdb-helpers.py
@@ -170,6 +170,17 @@ class HistoryEntryPrinter:
         elif event_num == 2002:
             return 'TS_EVENT_SSL_SESSION_REMOVE'
 
+        elif event_num == 2301:
+            return 'HTTP_TUNNEL_EVENT_DONE'
+        elif event_num == 2302:
+            return 'HTTP_TUNNEL_EVENT_PRECOMPLETE'
+        elif event_num == 2303:
+            return 'HTTP_TUNNEL_EVENT_CONSUMER_DETACH'
+        elif event_num == 2304:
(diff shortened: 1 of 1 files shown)


tools/gdb-helpers.py
```diff
@@ -170,6 +170,15 @@ def event_value_to_description(event_num: int) -> str:
...
 
+        elif event_num == 2301:
+            return 'TS_EVENT_HTTP_TUNNEL_EVENT_DONE'
+        elif event_num == 2302:
+            return 'TS_EVENT_HTTP_TUNNEL_EVENT_PRECOMPLETE'
+        elif event_num == 2303:
+            return 'TS_EVENT_HTTP_TUNNEL_EVENT_CONSUMER_DETACH'
+        elif event_num == 2304:
+            return 'TS_EVENT_HTTP_TUNNEL_EVENT_ACTIVITY_CHECK'
+
```
masaori335, line 183: We want `PARSE_ERROR` for `2305` too?
https://github.com/apache/trafficserver/blob/632c8f09afdebc907db35f287632b0e260df5780/include/proxy/http/HttpTunnel.h#L47-L51

bneradt: Good catch. 

Also, since these also aren't in the TS_EVENT lists in apidefs.h.in, I'm also going to keep them their accurate names as seen in the code (`HTTP_TUNNEL_EVENT.*`).

bneradt:
[approve ci autest rocky osx freebsd clang-analyzer]

zwoop (approved):

zwoop:
Looks like the changes were made, so this looks good.

PR #12323, opened 2025-06-30
Expose JSON client timeout to CLI

cmcfarlen:
This allows the user to change the timeout when reading a response from the JSONRPC server.

This exposes the attempts and timeout (in ms) separately, but I'm open to changing this to a simpler single config if that makes more sense.


brbzull0:
[approve ci autest 1]

cmcfarlen:

PR #12331, opened 2025-07-03
ja3_fingerprint: fix leak of duplicate header fields

bneradt:
When looping through duplicate header fields with TSMimeHdrFieldNextDup, each iteration must free the TSMLoc. ja3_fingerprint had a loop where it did not do this, thus it leaked these. This adds the required TSHandleMLocRelease.


masaori335 (approved):
Looks good.

cmcfarlen:

PR #12333, opened 2025-07-03
stats_over_http: ensure BrotliEncoderDestroyInstance

bneradt:
stats_over_http calls BrotliEncoderCreateInstance, but we weren't always calling BrotliEncoderDestroyInstance. This adds it to the b_stream destructor to ensure it gets called.

### DIFF
--- a/plugins/stats_over_http/stats_over_http.cc
+++ b/plugins/stats_over_http/stats_over_http.cc
@@ -247,7 +247,7 @@ stats_cleanup(TSCont contp, stats_state *my_state)
   }
 
   TSVConnClose(my_state->net_vc);
-  TSfree(my_state);
+  delete my_state;
   TSContDestroy(contp);
 }
 
@@ -932,7 +932,7 @@ static config_t *
 new_config(std::fstream &fh)
 {
   config_t *config    = nullptr;
-  config              = new config_t();
+  config              = new config_t;
   config->recordTypes = DEFAULT_RECORD_TYPES;
   config->stats_path  = "";
   std::string cur_line;
@@ -984,7 +984,7 @@ static void
 delete_config(config_t *config)
 {
   Dbg(dbg_ctl, "Freeing config");
-  TSfree(config);
+  delete config;
 }
 
 // standard api below...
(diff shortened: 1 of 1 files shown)


plugins/stats_over_http/stats_over_http.cc
```diff
@@ -120,33 +120,49 @@ static bool             is_ipmap_allowed(const config_t *config, const struct so
...
+  size_t              avail_out = 0;
+  size_t              total_in  = 0;
+  size_t              total_out = 0;
+
+  ~b_stream()
+  {
+    if (br) {
+      BrotliEncoderDestroyInstance(br);
+      br = nullptr;
+    }
```
bneradt, line 136: This is the real fix. 

However, with a destructor, b_stream is no longer a POD, so the memset in `stats_origin` no longer works and rightfully causes a compiler warning. So the rest of this patch is so that new of `stats_stat` 0-initializes the content.

bneradt:
[approve ci autest]

plugins/stats_over_http/stats_over_http.cc
```diff
@@ -726,9 +743,8 @@ stats_origin(TSCont contp, TSEvent /* event ATS_UNUSED */, void *edata)
   /* This is us -- register our intercept */
   Dbg(dbg_ctl, "Intercepting request");
 
-  my_state = (stats_state *)TSmalloc(sizeof(*my_state));
-  memset(my_state, 0, sizeof(*my_state));
-  icontp = TSContCreate(stats_dostuff, TSMutexCreate());
+  my_state = new stats_state;
```
cmcfarlen, line 746: You switched this to use `new`, but the cleanup of this still uses `TSfree` instead of `delete` (in `stats_cleanup`)

cmcfarlen (changes requested):
Please fix the `new`/`free` mismatch

plugins/stats_over_http/stats_over_http.cc
bneradt: oh yeah, good catch.

bneradt: I found another one too while fixing this. :)

cmcfarlen:

PR #12336, opened 2025-07-08
ProxyProtocol: add missing proxy protocol on clear http/1.1 private session

m-matth:
Add missing proxy protocol packet on clear http/1.1 origin connection whis Authorization/Www-Authenticate headers


bryancall:
[approve ci]

maskit:
Thank you for the report, and this PR. The code change you made on this PR helped me find a suspicious change.

I haven't checked the actual behavior yet, but current code does seem to have an issue. `do_outbound_proxy_protocol` used to be called from `HttpSM::state_http_server_open` on `NET_EVENT_OPEN` event before  [this change](https://github.com/apache/trafficserver/commit/7d5e673309), but now it's called on `CONNECT_EVENT_TXN` event indirectly by `write_outbound_proxy_protocol`. I feel like `write_outbound_proxy_protocol` should be called on `CONNECT_EVENT_DIRECT` event as well.

> i created a small PR for quick fix purposes only

Would you update this PR if we provide you feedback?

m-matth:
> Thank you for the report, and this PR. The code change you made on this PR helped me find a suspicious change.
> 
> I haven't checked the actual behavior yet, but current code does seem to have an issue. `do_outbound_proxy_protocol` used to be called from `HttpSM::state_http_server_open` on `NET_EVENT_OPEN` event before [this change](https://github.com/apache/trafficserver/commit/7d5e673309), but now it's called on `CONNECT_EVENT_TXN` event indirectly by `write_outbound_proxy_protocol`. I feel like `write_outbound_proxy_protocol` should be called on `CONNECT_EVENT_DIRECT` event as well.
> 
> > i created a small PR for quick fix purposes only
> 
> Would you update this PR if we provide you feedback?

Sure

PR #12338, opened 2025-07-08
Add support for http+unix scheme to remap config

maskit:
This is to have remap rules that only match if UDS is used.

The rules would be like:
```
map_with_recv_port http+unix://service.example/ http://origin.example/
```

### DIFF
--- a/doc/admin-guide/files/remap.config.en.rst
+++ b/doc/admin-guide/files/remap.config.en.rst
@@ -323,6 +323,35 @@ Examples
     regex_map http://x([0-9]+).z.com/ http://real-x$1.z.com/
     regex_redirect http://old.(.*).z.com http://new.$1.z.com
 
+.. _map_with_recv_port:
+
+map_with_recv_port
+==================
+
+Format::
+
+    map_with_recv_port client-URL origin-server-URL
+
+``map_with_recv_port`` supports two special URL schemes, ``http+unix`` and ``https+unix``.
+These are useful if you want to have different mapping rules or differnt plugin configuration for requests recevied via Unix Domain Socket.
+
+map_with_recv_port Examples
+---------------------------
+
+::
+
+   map_with_recv_port http://foo.example.com:8000/ http://x.example.com/
+   map_with_recv_port http://foo.example.com:8888/ http://y.example.com/
+
+Explanation: Requests received on port 8000 and 8888 are forwarded to different servers.
+
+::
+
+   map                     http://foo.example.com/ http://x.example.com/ @plugin=plugin1.so
+   map_with_recv_port http+unix://foo.example.com/ http://x.example.com/
+
+Explanation: All requests are forwarded to the same server, but plugin1 does not run for requests received via Unix Domain Socket.
+
 .. _map_with_referer:
 
 map_with_referer


src/proxy/http/remap/RemapConfig.cc
```diff
@@ -1255,7 +1255,8 @@ remap_parse_config_bti(const char *path, BUILD_TABLE_INFO *bti)
     // includes support for FILE scheme
     if ((fromScheme != std::string_view{URL_SCHEME_HTTP} && fromScheme != std::string_view{URL_SCHEME_HTTPS} &&
          fromScheme != std::string_view{URL_SCHEME_FILE} && fromScheme != std::string_view{URL_SCHEME_TUNNEL} &&
-         fromScheme != std::string_view{URL_SCHEME_WS} && fromScheme != std::string_view{URL_SCHEME_WSS}) ||
+         fromScheme != std::string_view{URL_SCHEME_WS} && fromScheme != std::string_view{URL_SCHEME_WSS} &&
+         fromScheme != std::string_view{URL_SCHEME_HTTP_UDS} && fromScheme != std::string_view{URL_SCHEME_HTTPS_UDS}) ||
         (toScheme != std::string_view{URL_SCHEME_HTTP} && toScheme != std::string_view{URL_SCHEME_HTTPS} &&
          toScheme != std::string_view{URL_SCHEME_TUNNEL} && toScheme != std::string_view{URL_SCHEME_WS} &&
          toScheme != std::string_view{URL_SCHEME_WSS})) {
```
bryancall, line 1262: Shouldn't the error message also change?

bryancall (changes requested):
Seems odd to create a new scheme for uds and convert the existing scheme to it.  I would use port 0 to designate UDS.  I took a look to see what Nginx does and hey use UDS file name as the port value.   Nice thing about this approach is you can have multiple UDS "ports" for different services.

maskit:
> Seems odd to create a new scheme for uds and convert the existing scheme to it. I would use port 0 to designate UDS. I took a look to see what Nginx does and hey use UDS file name as the port value. Nice thing about this approach is you can have multiple UDS "ports" for different services.

I already considered those options. Those were my first idea and I tried to implement those, but I concluded that those are not really feasible options unless we make a significant change.

Having a UDS filename as a port value would make the URL malformed. The port portion cannot have a string. This is not  only about the parser, but also the data structure. We would have to change URLImpl structure, and I suppose that affects cache. On the other hand, use of "+" in the scheme is allowed by the spec. The idea is from "git+ssh", application + transport, although the format is not standardized.

Using port number 0 is also problematic, because we have code to auto complete a port number based on the scheme (see `URL::port_get`). Port 0 can only mean that the port information is omitted. This is also the reason why requests for UDS have to use `map_with_recv_port`. If `map` is used, the port information in HTTP request needs to be used by the definition, and most clients do not have the default port number (80 or 443) for a scheme. So ATS does the auto completion, and use 80 or 443, then such requests never match with remap rules that have port number 0 even if we could have such rules (again, omitting a port number, or using 0 let ATS auto complete a port number, so we cannot have such rule). The special scheme indicate that the port number is unavailable, not omitted.

Like you pointed out, a limitation of this implementation is that it cannot distinguish multiple UDS file paths. Some might want to specify UDS file path for a remap rule. I'm not sure how the file path should be given but I wouldn't reject the need. However, I think changes for that should be on top of this change. In the future, we might have support for named pipe as well. Then we'd need a way to distinguish UDS and named pipe. Maybe "http+pipe://" ?

cmcfarlen:
I think we should use `http+unix` (I don't think `uds` is often used for domain sockets, Vijay may have made that up) to mean "received from any unix socket".  I would be fine with also supporting `:0` for unix sockets, but it does seem less specific.  Ideally we could support the `[:/path/to/socket]` but I don't think the URL class supports this? If we want to add this feature, definitely do it in another PR.

jeredfloyd:
Copying my message to ts-dev here:

I think something like the ACL Filters functionality is the right way to do what you want -- there is already "@in_ip" to select mappings based on the IP address on which the request was received.  Based on that, I would propose adding a parameter "@in_proto" that can take values "unix", "http", "http2", etc.  You can then create different mapping rules that are selected based on incoming protocol without modification to the target schema.

Digging in a bit more to my thought process that led to this, I think there are two separate matters to consider here.

The first is HTTP over sockets.  This isn't really HTTP(/[12]) if it's not over TCP, nor HTTP/3 if it's not over QUIC, is it?  The addressing schema for unix sockets is different, because in the general sense of trying to address an origin over a socket we're referring to a socket path rather than a hostname or IP address.  From ATS' perspective (as a reverse proxy, at least) the only things that matter are the Host header and the request target, but it's worth considering how a theoretical client would be addressing HTTP-over-socket.  It looks like WHATWG had some inconclusive discussion on this back in 2021, and it's worth reviewing that issue first: https://github.com/whatwg/url/issues/577

More importantly, though, modifying the target schema for translating the map target based on how it arrived to ATS feels like an abstraction violation.  There isn't how we currently remap differently based on received interface, or IP address ("@in_ip" ACL Filters), and we don't support this based on protocol (HTTP/1 vs 2 vs 3).  (The closest is map_with_recv_port, which is minimally documented and I'm not entirely certain what it does.)  This says to me that mis-using the schema part of the map target is not the right way to solve your problem.

Therefore, I think an "@in_proto" extension for the ACL Filter mechanism is most parallel to the existing map language.

maskit:
@jeredfloyd Thank you for your comment. 

If we were introducing remap.yaml and we could do completely different things, I'd like the idea of a separate matching parameter. However, remap ACL or `@something` on remap rules is not really an option for this, because `@something` is not a part of a matching pattern. Those are more like actions used after matching. Remap rules work like "If a request matches with this pattern, apply this ACL, and run these plugins". If the ACL deny the access, it results in a 403 response and that's it. That is not what I want at a minimum.  In my use case, both entry points (ip and unix socket) are valid, and I want to have different plugin configurations on those remap rules.

I read the long discussion on WHATWG, and the scheme idea made most sense to me. I think their discussion was much more complicated than this discussion because they were thinking of standardizing it. Although we should try to stick with standards as much as possible, what I'm proposing is just an application specific format for a config file, and I suppose many libraries can still parse or generate the URL-ish matching pattern string. Also, client applications are not required to do anything special. Requests made by curl command with --unix-path option matches with the rules.

For your last point, as you already found, `map_with_recv_port` does the work. I used it on the example in the description. That keyword is the way to remap based on received interface/port. With `remap_with_recv_port`, the remap processor takes received interface/port in consideration, and users need to tell on their remap rules which interface/port to compare with.

jeredfloyd:
@maskit I generally agree with your analysis, but I think parallel functionality already exists with `@in_ip` which is why I propose `@in_proto` as a different implementation. It's possible that I'm misunderstanding what the current functionality does and these filters do not allow for overloading of targets?

From https://docs.trafficserver.apache.org/en/latest/admin-guide/files/remap.config.en.html examples:
```
map http://foo.example.com/  http://foo.example.com/ @action=allow @src_ip=10.5.2.1 @in_ip=72.209.23.4
```

So I propose:
```
map http://foo.example.com/  http://origin-unix.example.com/ @action=allow @in_proto=unix
map http://foo.example.com/  http://origin-http.example.com/ @action=allow @in_proto=http
```

If that doesn't actually work the way I think I'll stop arguing, but I do think the above is a much cleaner abstraction as it does not create non-standard URIs.

maskit:
@jeredfloyd Unfortunately, `@in_ip` doesn't work as you expect. It's just an additional ACL that will be checked AFTER remapping.

If I have this remap rule below on my remap.config and I accessed the ATS like `curl http://127.0.0.1:8080/`, it returns `403 Access Denied`. Not `404 Not Found on Accelerator`.

```
map / http://origin:8888/file1 @action=allow @in_ip=192.168.0.1
```

That means the remap rule matched where the interface/address used is 127.0.0.1.

And having these two causes an error and ATS doesn't start up, because the mapping rules are identical from remap perspective.
```
map / http://origin:8888/file1 @action=allow @in_ip=192.168.0.1
map / http://origin:8888/file2 @action=allow @in_ip=127.0.0.1
```

So `@in_proto` could be added as an ACL condition, but it can't be used when ATS lookups a remap rule that matches.

jeredfloyd:
@maskit Aw, darn.  I still think allowing overloading with `@in_proto` rules is semantically cleaner, but I also don't have time to try and implement that :-)  I've offered my opinion, so I'll step back now.

maskit:
@jeredfloyd I agree with you on the point that specifying an interface or a protocol by something separated from URL would be cleaner. I wish I could do that. Sadly it's not a feasible option at this moment unless somebody makes the big change for the remap processor so the matcher does not rely on just URL strings. I'll definitely take your idea in consideration if we make the big change. Thank you for your feedback.

bryancall (approved):

maskit:
Thank you all for your feedback.

Although this PR was approved, I'll make these changes on this PR:
- Change the scheme name to `http+unix`
- Add tests
- Add documentation

src/proxy/http/remap/RemapConfig.cc
maskit: Updated the error message.

Issue #12340, opened 2025-07-09
Crash in HostDBContinuation::remove_and_trigger_pending_dns


masaori335:
```
#0  Mutex_lock (m=0x0, t=0x7fbd23202e80) at /include/iocore/eventsystem/Lock.h:320
#1  Mutex_lock (m=..., t=0x7fbd23202e80) at /include/iocore/eventsystem/Lock.h:351
#2  MutexLock::MutexLock (this=0x7fbd1a8fc438, am=..., t=0x7fbd23202e80) at /include/iocore/eventsystem/Lock.h:457
#3  HostDBContinuation::remove_and_trigger_pending_dns (this=<optimized out>) at /src/iocore/hostdb/HostDB.cc:1309
#4  0x0000556e430724e5 in HostDBContinuation::dnsEvent (this=0x7f963789e000, event=<optimized out>, e=<optimized out>) at /src/iocore/hostdb/HostDB.cc:1129
#5  0x0000556e43059cfe in Continuation::handleEvent (this=<optimized out>, event=600, data=0x7eda1cd80680) at /include/iocore/eventsystem/Continuation.h:236
#6  DNSEntry::postOneEvent (this=0x7f96b648c000) at /src/iocore/dns/DNS.cc:1512
#7  0x0000556e431d3425 in Continuation::handleEvent (this=<optimized out>, event=1, data=0x7fbd0e001200) at /include/iocore/eventsystem/Continuation.h:236
#8  EThread::process_event (this=0x7fbd23202e80, e=0x7fbd0e001200, calling_code=1) at /src/iocore/eventsystem/UnixEThread.cc:162
#9  0x0000556e431d36d7 in EThread::process_queue (this=0x7fbd23202e80, NegativeQueue=0x7fbd1a8fc6e0, ev_count=0x7fbd1a8fc6fc, nq_count=0x7fbd1a8fc6f8) at /src/iocore/eventsystem/UnixEThread.cc:197
#10 0x0000556e431d3d49 in EThread::execute_regular (this=0x7fbd23202e80) at /src/iocore/eventsystem/UnixEThread.cc:255
#11 0x0000556e431d4778 in EThread::execute (this=0x7fbd23202e80) at /src/iocore/eventsystem/UnixEThread.cc:348
#12 0x0000556e431d2207 in spawn_thread_internal (a=0x7fbd26347dc0) at /src/iocore/eventsystem/Thread.cc:75
[...]

PR #12341, opened 2025-07-09
Add action & mutex check before triggering pending dns

masaori335:
Address #12340.

### DIFF
--- a/src/iocore/hostdb/HostDB.cc
+++ b/src/iocore/hostdb/HostDB.cc
@@ -1294,6 +1294,10 @@ HostDBContinuation::remove_and_trigger_pending_dns()
   }
   EThread *thread = this_ethread();
   while ((c = qq.dequeue())) {
+    if (c->action.cancelled || c->mutex == nullptr) {
+      continue;
+    }
+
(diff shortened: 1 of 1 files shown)


src/iocore/hostdb/HostDB.cc
```diff
@@ -1294,6 +1294,10 @@ HostDBContinuation::remove_and_trigger_pending_dns()
   }
   EThread *thread = this_ethread();
   while ((c = qq.dequeue())) {
+    if (c->action.cancelled || c->mutex == nullptr) {
```
masaori335, line 1297: @bneradt your original fix of #9742 seems right :)

Do you know where/how the `HostDBContinuation` is the queue is handled (or cancelled) in other places?

masaori335: It looks like `hostdb_cont_free` is called when this HostDBContinuation is waiting in the local queue (`qq`).

masaori335: This change could be workaround, but to fix this completely, it looks like we need to make this `HostDBContinuation` refcount object.

bneradt (approved):

cmcfarlen:

PR #12342, opened 2025-07-09
traffic_ctl - Add ECONNREFUSED to the retry list when connection to the rpc.

brbzull0:
Better naming and make sure it retries when refused, without this change,  `connect(attempts=1/5)` seems that it didn't even try. I may be adding more as necesary.

```
RPC Node Error: connect(attempts=1/5): Couldn't open connection with /opt/ats/var/trafficserver/jsonrpc20.sock. Last error: Connection refused(111)
```


brbzull0:
[approve ci autest 1]

masaori335 (approved):
Retrying on `ECONNREFUSED` seems useful in some cases.

PR #12344, opened 2025-07-09
header_rewrite: Add optional switches to change plugin behavior on load

maskit:
This adds these switches that can be used on plugin.config or remap.config:
- --timezone
- --inbound-ip-source


maskit:
[approve ci autest 1]

bryancall (approved):

PR #12347, opened 2025-07-09
Add config file support for cache_fill

jasmine-nahrain:
This is just copying over the existing code from `background_fetch` to enable config file support for `cache_fill`

### DIFF
--- a/plugins/experimental/cache_fill/configs.h
+++ b/plugins/experimental/cache_fill/configs.h
@@ -1,0 +1,89 @@
+/** @file
+
+    Plugin to perform background fetches of certain content that would
+    otherwise not be cached. For example, Range: requests / responses.
+
+    @section license License
+
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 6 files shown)


plugins/experimental/cache_fill/configs.h
```diff
@@ -0,0 +1,89 @@
+/** @file
+
+    Plugin to perform background fetches of certain content that would
+    otherwise not be cached. For example, Range: requests / responses.
```
brbzull0, line 4: I know this comes from the `background_fetch`, wonder if is this fully accurate for this plugin?

brbzull0 (approved):
Looks good, just  one nit.

Also, would be nice also to  update the docs to include some notes about the supported config.

Thanks.

plugins/experimental/cache_fill/configs.h
jasmine-nahrain: The docs actually already mention it, it was not supported though. I want to leave the implementation the exact same for now.  I'll make some more changes later

cmcfarlen:

PR #12348, opened 2025-07-10
Cleanup HostDBContinuation

masaori335:
It looks like some `HostDBContinuation` code are unused. To be honest, I'm surprised. Am I missing some compile option?


cmcfarlen:
Was this code used for the persistence feature that was removed?

masaori335:
Seems like the iterate functions are introduced by this commit https://github.com/apache/trafficserver/commit/94f4083d02c545275177718a9a6c239de78ce978 for HTTP UI. We got rid of the UI, but when we expand `traffic_ctl host` command to HostDB, we can use this code?

bryancall (approved):

PR #12352, opened 2025-07-11
traffic_ctl: get all host statuses by empty arg

masaori335:
`server_get_status` returns all host statuses if given param is empty, but `traffic_ctl` requires one argument. This relaxes the restriction.
https://github.com/apache/trafficserver/blob/636772a5c5ad89d19d4fcdc9a1a797228c243237/src/proxy/HostStatus.cc#L458-L468


cmcfarlen:

PR #12355, opened 2025-07-11
10.1.x: stats_over_http: ensure BrotliEncoderDestroyInstance (#12333)

bneradt:
This is a backport of #12333 to 10.1.x.  We won't need this if we decide to backport the prometheus update in #12302 because then there won't be conflicts. But this was easy enough to creat that I went ahead and made it in case we don't want the Prometheus update for stats_over_http. If we wind up #12302 , then this PR can be closed without merging because the patch in #12333 would backport cleanly.

---

stats_over_http calls BrotliEncoderCreateInstance, but we weren't always calling BrotliEncoderDestroyInstance. This adds it to the b_stream destructor to ensure it gets called.

(cherry picked from commit 636772a5c5ad89d19d4fcdc9a1a797228c243237)

### DIFF
--- a/plugins/stats_over_http/stats_over_http.cc
+++ b/plugins/stats_over_http/stats_over_http.cc
@@ -105,7 +105,7 @@ struct config_holder_t {
   config_t       *config;
 };
 
-enum output_format { JSON_OUTPUT, CSV_OUTPUT };
+enum output_format_t { JSON_OUTPUT, CSV_OUTPUT };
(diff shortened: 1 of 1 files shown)


maskit:
> traffic_server FATAL: Could not bind or listen to port 42847, mptcp enabled: 0 (error: 98) Address already in use -1 : 37

maskit:
[approve ci autest 1]

cmcfarlen:
Closing this as we picked #12302 instead.

Issue #12321, opened 2025-06-27
Investigate optimizing Huffman Decoding for http2 header compression


pixitha:
When we were troubleshooting some TLS/encryption performance issues I noticed that one of the top cpu users on perf top for traffic server was `huffman_decode`. I went poking around in the code and noticed that our implementation is somewhat old, and does bit by bit down the tree. Doing a bit of digging around the web I found some newer implementations that we might be able to use, that perform better.

https://blog.litespeedtech.com/2019/09/16/fast-huffman-decoder/ Has a really good overview of the code they are using: "The idea is to process compressed data several bits — instead of one bit — at a time"

I'm mostly only suggesting that we look at swapping our logic in the actual lookup process, not swapping the whole h2 library.

https://github.com/litespeedtech/ls-hpack (MIT License)

### Stats from perf top: ATS9

```bash
Samples: 1M of event 'cycles', 4000 Hz, Event count (approx.): 170770658686 lost: 0/0 drop: 0/0
Overhead  Shared Object                                                   Symbol
   7.12%  libcrypto.so.1.1                                                [.] __ecp_nistz256_mul_montq                   ◆
   5.58%  libpcre.so.1.2.13                                               [.] match                                      ▒
   4.42%  libcrypto.so.1.1                                                [.] __ecp_nistz256_sqr_montq                   ▒
   4.32%  libluajit-5.1.so.2.1.0                                          [.] match.part.0                               ▒
   4.20%  [kernel]                                                        [k] nmi                                        ▒
   2.17%  traffic_server                                                  [.] huffman_decode
```

or from another view:

```bash
Samples: 3M of event 'cycles', 4000 Hz, Event count (approx.): 151569433376 lost: 0/0 drop: 0/0

+   26.91%     libcrypto.so.1.1                                                                                          ◆
+   14.41%     [kernel]                                                                                                  ▒
-   12.00%     traffic_server                                                                                            ▒
        2.06%     [.] huffman_decode                                                                                     ▒
        0.76%     [.] _mime_hdr_field_list_search_by_string                                                              ▒
        0.37%     [.] (anonymous namespace)::mime_mem_print_                                                             ▒
        0.37%     [.] (anonymous namespace)::escapify_url_common                                                         ▒
        0.22%     [.] hdrtoken_tokenize                                                                                  ▒
        0.16%     [.] mime_parser_parse                                                                                  ▒
        0.14%     [.] mime_hdr_field_find                                                                                ▒
[...]

PR #12357, opened 2025-07-11
Implement HuffmanCodec with LiteSpeed

bneradt:
Performance tooling highlights hpack decoding as an expensive operation for us. This improves hpack performance by using the MIT licensed LiteSpeed implementation.

When testing this in production, it seems to address the huffman_decode performance hot spot. It no longer appears near the top in `perf top`.

Fixes: #12321


bneradt:
[approve ci autest]

maskit (changes requested):
Looks like it's a simple replacement. I don't have any concerns on the changes.

Could you note the version number or commit id of the borrowed code somewhere? Since their code is still updated sometimes, we probably need the information to see if we have important changes on our copy.

bneradt:
> Looks like it's a simple replacement. I don't have any concerns on the changes.
> 
> Could you note the version number or commit id of the borrowed code somewhere? Since their code is still updated sometimes, we probably need the information to see if we have important changes on our copy.

Good idea. I added a README.md to the lib/ls-hpack directory that gives some history of our use of LiteSpeed and references the version we've borrowed from.

maskit:
[approval ci autest 2]

maskit:
[approve ci autest 2]

PR #12359, opened 2025-07-13
TS API: Add TSHttpTxnVerifiedAddrSet/Get and real-ip plugin

maskit:


doc/admin-guide/plugins/realip.en.rst
```diff
@@ -0,0 +1,86 @@
...
+   .. include:: ../../common.defs
+
+.. _admin-plugins-realip:
+
+Real IP Plugin
+*******************
+
+Description
+===========
+The `realip` plugin reads an IP address from a specified data source such as HTTP header field, and make it available for ATS and
```
bneradt: `make it`  -> `makes it`

doc/admin-guide/plugins/realip.en.rst
```diff
@@ -0,0 +1,86 @@
...
+its plugins as the verified client address.
+
+To use the verified client IP address for ACL, :ts:cv:`proxy.config.acl.subjects` needs to be set to `PLUGIN`.
+Similary, if you want to use the IP address on header_rewrite plugin, its `INBOUND_IP_SOURCE` needs to be set to `PLUGIN`.
```
bneradt: Let's link `header_rewrite` to the header rewrite plugin docs.

doc/admin-guide/plugins/realip.en.rst
```diff
@@ -0,0 +1,86 @@
...
+
+
+Configuration
+=============
+
+To enable the `realip` plugin, insert the following line in :file:`plugin.config`:
```
bneradt: I think you'll need two `::` at the end to make this a code block.

doc/admin-guide/plugins/realip.en.rst
```diff
@@ -0,0 +1,86 @@
...
+
+    realip.so <config.yaml>
+
+The plugin supports 2 mode:
```
bneradt: `mode` -> `modes`

doc/admin-guide/plugins/realip.en.rst
```diff
@@ -0,0 +1,86 @@
...
+
+simple
+------
+
+The plugin reads an IP address from a specified HTTP header field. These settings bewlow are available for this mode:
```
bneradt: `bewlow`

doc/developer-guide/api/functions/TSHttpTxnVerifiedAddrSet.en.rst
```diff
@@ -0,0 +1,64 @@
...
+
+Description
+===========
+
+This function enables plugins to provide a reliable client IP address for |TS| and other plugins.
+This is useful if there is a proxy in front of |TS| and it forwards client's IP address by HTTP header field, PROXY protocol, etc.
+Plugins that call this function are expected to check the validity of the IP address.
+
+The provided address will be used for remap ACL if :ts:cv:`proxy.config.acl.subjects` is set to `PLUGIN`.
+Plugins can get the provided address by calling the getter function bewlow if those need client's real IP address.
```
bneradt: `bewlow`

doc/developer-guide/api/functions/TSHttpTxnVerifiedAddrSet.en.rst
```diff
@@ -0,0 +1,64 @@
...
+
+    #include <ts/ts.h>
+
+.. function:: TSReturnCode TSHttpTxnVerifiedAddrGet(TSHttpTxn txnp, const struct sockaddr **addr)
+
+Description
+===========
+
+This is the getter version for the above setter. This returns 1 if a verified address is available.
+Please note that a port number is not always available even if the function returns 1.
```
bneradt: It's probably good to explicitly say that the memory for the `addr` output parameter is managed by ATS core and should not be freed by the caller.

doc/developer-guide/api/functions/TSHttpTxnVerifiedAddrSet.en.rst
```diff
@@ -0,0 +1,64 @@
...
+.. function:: TSReturnCode TSHttpTxnVerifiedAddrSet(TSHttpTxn txnp, const struct sockaddr *addr)
+
+Description
+===========
+
+This function enables plugins to provide a reliable client IP address for |TS| and other plugins.
+This is useful if there is a proxy in front of |TS| and it forwards client's IP address by HTTP header field, PROXY protocol, etc.
+Plugins that call this function are expected to check the validity of the IP address.
+
+The provided address will be used for remap ACL if :ts:cv:`proxy.config.acl.subjects` is set to `PLUGIN`.
```
bneradt: Probably good to explicitly say that ATS core makes and maintains its own copy of the memory from `addr`.

bneradt (changes requested):
Looks great. Maybe add an autest for it? It would be good to get coverage on at least the simple case just to verify the setter/getter api is working correctly. We could have a test plugin that simply reads an IP from something like a `client-ip` request header, sets it, gets the value, and adds a `client-ip` response header that then can be verified by the autest.

maskit:
> Maybe add an autest for it? It would be good to get coverage on at least the simple case just to verify the setter/getter api is working correctly. We could have a test plugin that simply reads an IP from something like a client-ip request header, sets it, gets the value, and adds a client-ip response header that then can be verified by the autest.

Yes, I will add it. I'm going to write a test for the experimental plugin (the simple mode), and use header_rewrite as well to test the getter API.

maskit:
I updated the documentation, and added tests. Please take another look.

tests/gold_tests/pluginTest/tsapi/test_TSHttpTxnVerifiedAddr.test.py
```diff
@@ -0,0 +1,70 @@
...
+    })
+
+# ----
+# Test Cases
+# ----
+
+tr = Test.AddTestRun()
+tr.MakeCurlCommand(f'-v http://127.0.0.1:{ts.Variables.port}/httpbin/get', ts=ts)
+tr.Processes.Default.ReturnCode = 0
+tr.Processes.Default.StartBefore(httpbin, ready=When.PortOpen(httpbin.Variables.Port))
```
bneradt, line 65: I think that the httpbin.test.ext should take care of this ready already.

tests/gold_tests/pluginTest/tsapi/test_TSHttpTxnVerifiedAddr.test.py
```diff
@@ -0,0 +1,70 @@
...
+        'proxy.config.diags.debug.tags': 'http|test_TSHttpTxnVerifiedAddr',
+        'proxy.config.ssl.server.cert.path': f'{ts.Variables.SSLDir}',
+        'proxy.config.ssl.server.private_key.path': f'{ts.Variables.SSLDir}'
+    })
+
+# ----
+# Test Cases
+# ----
+
+tr = Test.AddTestRun()
```
bneradt, line 62: AddTestRun descriptions can be helpful.

tests/gold_tests/pluginTest/tsapi/test_TSHttpTxnVerifiedAddr.test.py
```diff
@@ -0,0 +1,70 @@
...
+
+# ----
+# Setup ATS
+# ----
+ts = Test.MakeATSProcess("ts", enable_tls=True)
+
+# add ssl materials like key, certificates for the server
+ts.addDefaultSSLFiles()
+
+ts.Disk.remap_config.AddLines([f'map /httpbin/ http://127.0.0.1:{httpbin.Variables.Port}/'])
```
bneradt, line 40: AddLine since only one line.

tests/gold_tests/pluginTest/tsapi/test_TSHttpTxnVerifiedAddr.test.py
```diff
@@ -0,0 +1,70 @@
...
+# ----
+httpbin = Test.MakeHttpBinServer("httpbin")
+
+# ----
+# Setup ATS
+# ----
+ts = Test.MakeATSProcess("ts", enable_tls=True)
+
+# add ssl materials like key, certificates for the server
+ts.addDefaultSSLFiles()
```
bneradt, line 38: Do we use TLS in this test? If not, we can simplify by removing the https config.

maskit:
Thanks for the feedback. I'll make another PR to update the test because this PR is blocking #12501.

tests/gold_tests/pluginTest/tsapi/test_TSHttpTxnVerifiedAddr.test.py
maskit: Looks like you're right, but I see a couple of tests that do the same. I'll update all those tests.

PR #12360, opened 2025-07-14
[10.1.x] Run autests with curl UDS default

serrislew:
10.1.x rebase of #12168 
Will need to include #12350 changes after merged (to fix autests)


tests/gold_tests/autest-site/curl.test.ext
```diff
@@ -30,11 +30,11 @@ Tools to help with TestRun commands
 #
 
 
-def spawn_curl_commands(self, cmdstr, count, retcode=0, use_default=True):
+def spawn_curl_commands(self, cmdstr, count, retcode=0, use_default=True, ts):
```
bneradt: Let's add the fix for this positional parameter being after ones that take a default.

serrislew: I was planning on including #12350 changes (which has the fix and other fixes) after that one gets merged

serrislew:
Putting to draft until #12350 gets merged as this backport depends on those fixes to be included

bneradt (approved):

PR #12363, opened 2025-07-15
Adds a %{CERT} and %{CLIENT-CERT} cond

zwoop:
NOTE (yes you Brian):

I'm working on refactoring some attests, starting with header_rewrite tests, and I will add tests for the certificates as part of that refactoring.


plugins/header_rewrite/resources.cc
```diff
@@ -126,6 +132,32 @@ Resources::gather(const ResourceIDs ids, TSHttpHookID hook)
...
+  }
+
+  if (ids & RSRC_MTLS_CERTIFICATE) {
+    Dbg(pi_dbg_ctl, "\tAdding Cripts Certs::Client");
+    mtls_cert = new cripts::Certs::Client(*client_conn);
+  }
+
+  if (ids & RSRC_SERVER_CERTIFICATE) {
+    Dbg(pi_dbg_ctl, "\tAdding Cripts Certs::Server");
+    server_cert = new cripts::Certs::Server(*client_conn);
```
serrislew, line 157: shouldn't this be `server_conn`?

plugins/header_rewrite/value.cc
```diff
@@ -55,6 +55,7 @@ Value::set_value(const std::string &val)
 
           if (parser.parse_line(_value)) {
             tcond_val->initialize(parser);
+            require_resources(tcond_val->get_resource_ids());
```
serrislew, line 58: why does Value need the resource ids? doesn't this get defined when initialize() is called

plugins/header_rewrite/resources.cc
zwoop: > shouldn't this be `server_conn`?

No, it is confusing, but this says "get the server certificate used for the TLS handshake on the clients connection into ATS".

zwoop: What gets complicated is that there are 4 possible certificates, that Cripts support. On the inbound connection it can get the client certificate and the server certificate (that's on client_conn), and on the outbound connection it can get same (at which point the server certificate is the certificate on the upstream TLS connection / server).

plugins/header_rewrite/value.cc
zwoop: > why does Value need the resource ids? doesn't this get defined when initialize() is called

So this is actually a bug in the original header_rewrite code. When you use a condition as a value, it didn't use to update the required resources properly. This mostly worked ok for regular cases, but with new resource values, it would fail. And yes, the resources are set in the initializations, but was never propagated from the "value" back to the operator (e.g. set-header):

The use case here are things like

```
   set-header X-Cert "%{CERT:SUBBJECT"}
```

zwoop:
[approve ci autest 2]

plugins/header_rewrite/value.cc
zwoop: Ok, as discussed on slack, I've made the code cleanup to make this less confusing. There's no difference in behavior, just more clear as to why we have to update the owners resource requirements.

doc/admin-guide/plugins/header_rewrite.en.rst
```diff
@@ -496,6 +496,42 @@ As a special matcher, the inbound IP addresses can be matched against a list of
...
+    cond %{INBOUND:TLS} ="" [NOT]
+      set-header X-Client-Cert "%{INBOUND:CLIENT-CERT:PEM}"
+      set-header X-Client-Cert-Subject "%{INBOUND:CLIENT-CERT:SUBJECT}"
+      set-header X-Client-Cert-Issuer "%{INBOUND:CLIENT-CERT:ISSUER}"
+      set-header X-Server-Cert-Subject "%{INBOUND:SERVER-CERT:SUBJECT}""
+
+The ``SAN:`` fields will return a semicolon-separated list of the respective
+values, there can be zero, one or many of each SAN type. Example:
+
+    cond %{CLIENT-CERT:SAN:DNS} /example\.com/
```
serrislew, line 551: `%{INBOUND:CLIENT-CERT:SAN:DNS}`

maskit:
I wonder what would be a way to read [additional certificates](https://www.ietf.org/archive/id/draft-ietf-httpbis-secondary-server-certs-01.html). `%{OUTBOUND:SERVER-CERT[1]}` ? It's not supported by even ATS core but it may be worth considering to avoid ugly syntax.

doc/admin-guide/plugins/header_rewrite.en.rst
```diff
@@ -496,6 +496,42 @@ As a special matcher, the inbound IP addresses can be matched against a list of
     and the configuration parser will error out. The format here is very specific, in particular no
     white spaces are allowed between the ranges.
 
+If |ATS| is built with :ref:`Cripts <developer-guide-cripts>` support, a number of additional
+qualifiers are available exclusively on TLS sessions, for X509 certificate introspection.
+The client certificate (for mutual TLS) is accessed qith qualifier prefix of ``CLIENT-CERT:``
```
bneradt: `qith`

doc/admin-guide/plugins/header_rewrite.en.rst
```diff
@@ -496,6 +496,42 @@ As a special matcher, the inbound IP addresses can be matched against a list of
...
+
+    cond %{SEND_RESPONSE_HDR_HOOK} [AND]
+    cond %{INBOUND:TLS} ="" [NOT]
+      set-header X-Client-Cert "%{INBOUND:CLIENT-CERT:PEM}"
+      set-header X-Client-Cert-Subject "%{INBOUND:CLIENT-CERT:SUBJECT}"
+      set-header X-Client-Cert-Issuer "%{INBOUND:CLIENT-CERT:ISSUER}"
+      set-header X-Server-Cert-Subject "%{INBOUND:SERVER-CERT:SUBJECT}""
+
+The ``SAN:`` fields will return a semicolon-separated list of the respective
+values, there can be zero, one or many of each SAN type. Example:
```
bneradt: Two colons needed at the end, `Example::`

bneradt (changes requested):
Looks helpful.

zwoop:
> I wonder what would be a way to read [additional certificates](https://www.ietf.org/archive/id/draft-ietf-httpbis-secondary-server-certs-01.html). `%{OUTBOUND:SERVER-CERT[1]}` ? It's not supported by even ATS core but it may be worth considering to avoid ugly syntax.

I think the "syntax" for this would likely have to be nested :'s, if / when we want to support this. E.g.

```
%{OUTBOUND-SERVER-CERT:1:CN}
```

with the omission of the index be an implicit "0".

doc/admin-guide/plugins/header_rewrite.en.rst
zwoop: Fixed

doc/admin-guide/plugins/header_rewrite.en.rst
zwoop: Fixed

maskit:
> `%{OUTBOUND-SERVER-CERT:1:CN}`
with the omission of the index be an implicit "0".

So the second field may or may not be an index number. I can live with it but it sounds like the parser is going to be a bit complicated.

zwoop:
[approve ci autest]

zwoop:
I missed the reference to the secondary certs RFC... Maybe this should be another qualifier, like

```
%{INBOUND:SECONDARY-CERT:#: ...}
```

with the # being required. We'd also need to have some sort of "counter" for this I think (as a new condition qualifier)

zwoop:
@maskit  What should we do here? Land it as is, and fix secondary certs later?  Or try to fix it now (but openssl / ats doesn't support secondary certs I think?)

maskit:
> ```
> %{INBOUND:SECONDARY-CERT:#: ...}
> ```

This looks fine, and I don't think it needs to be implemented now. Since it's going to be another qualifier we should be able to add it anytime later. I just wanted to make sure that we won't need to make incompatible changes or introduce a hacky syntax.

maskit (approved):

bneradt (approved):

bneradt:
10.1.x cherry-pick in: #12738

PR #12371, opened 2025-07-18
Hrw autest refactor

zwoop:


tests/gold_tests/pluginTest/header_rewrite/gold/cond-elif.gold
```diff
@@ -1,5 +1,5 @@
 ``
-> GET http://www.example.com/from_path/hrw-sets.png``
+> GET http://www.example.com/from_1/hrw-sets.png``
```
maskit, line 2: Unique URL is helpful, but let's not use sequential numbers or letters. Inserting tests would be painful.

zwoop: How would that be painful? I guess I don't really care what we name them, but they have to be named something :). I'd imagined that if you are adding a new test to an existing autest, you append to the current list of remap rules if you need a new one.  Sequential numbering. There'd never be a reason to add a rule in the middle of remap rules I don't think. If there is, at some point, just name it /from_x_y ? E.g. /from_4_1.

We could automate the numbering, but when I looked at that, it makes making the subsequent tests more difficult (you need some sort of identifier).

zwoop:
[approve ci]

tests/gold_tests/pluginTest/header_rewrite/gold/cond-elif.gold
zwoop: Also, note that it's not the request that's unique, it's the remap rule. One remap rule can handle multiple requests (test cases).

maskit: > There'd never be a reason to add a rule in the middle of remap rules I don't think. If there is, at some point, just name it /from_x_y ? E.g. /from_4_1.

I think similar test cases should be close each other. We could do "4_1", but that doesn't completely resolve the issue. Let's say we have "4_1", "4_2", and "4_3". If I wanted to insert a test between "4_1" and "4_2", I'd have to make "4_1_1" or renumber "4_2" and "4_3".

Also even for the case that I append tests, somebody else could be appending another tests on the same file. Then the number would conflicts. Either of those have to update the test number.

I don't think whether tests or remap rules matters in this discussion.

zwoop: But the order doesn't matter. It's just an identifier, which I thought was easier to make numbered, but we can call them bob and alice if you want ...

zwoop:
[approve ci docs]

zwoop:
[approve ci]

zwoop:
[approve ci autest-2]

PR #12372, opened 2025-07-20
Move code for SSL from UnixNetVC to SSLNetVC

maskit:
This just delegates the details to a subclass.


bneradt (approved):
Looks like a straightforward port.

Issue #12369, opened 2025-07-18
TrafficServer 9.2.x fails to cache 200 responses after 302 redirects despite valid caching headers


shongkaikai:
Description:
When following 302 redirects, TrafficServer fails to cache the subsequent 200 response even when all caching conditions are met. This occurs with the documented configuration for redirect handling and caching.

Expected Behavior:
The 200 response from the final destination (http://172.16.0.89:8013/a) should be cached according to its Cache-Control: max-age=63 and Expires headers, similar to direct 200 responses.

Observed Behavior:
All requests trigger full round-trips to origin servers within the cache lifetime window (63s), indicating cache misses. Normal 200 responses (non-redirected) are cached properly.

Reproduction Steps:

Configure with record.config:
```
CONFIG proxy.config.http.redirect.actions STRING default:follow
CONFIG proxy.config.http.number_of_redirections INT 1
CONFIG proxy.config.http.cache.http INT 1
```

Make rapid sequential requests (<2s interval):

bash
curl -v 'http://172.16.0.82:2500/302/a'
Observe origin server logs showing repeated requests

first origin server response is 302：
```
HTTP/1.1 302 Moved Temporarily
Server: openresty/1.27.1.2
Date: Fri, 18 Jul 2025 07:59:54 GMT
Content-Type: text/html
Content-Length: 151
Connection: keep-alive
Location: http://172.16.0.89:8013/a
```

Final 200 response contains valid caching directives:
```
HTTP/1.1 200 OK
Cache-Control: max-age=63
Expires: Fri, 18 Jul 2025 08:01:58 GMT
ETag: "6867a133-9"
```

JosiahWI:
Could this be a duplicate of #9275? The fix for that issue unveiled a different issue that caused a crash, so I don't think it was ever backported to 9.x.

masaori335:
Seems like we faced the same issue with 10.0.x. It looks like combination of #11542 and #12215 fixed this. They're going to be released by next release 10.1.0. Please try it out.

bryancall:
This is a duplicate of: #10955

PR #12373, opened 2025-07-21
Fix: [ATS 200 response after 302 not cached] (#12369)

shongkaikai:
Fix: [ATS 9.2.x: 200 response after 302 not cached]

With backport markers:
- Project: 9.2.x
- Project: 10.x

### DIFF
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -2062,7 +2062,7 @@ HttpSM::state_read_server_response_header(int event, void *data)
     t_state.api_next_action       = HttpTransact::StateMachineAction_t::API_READ_RESPONSE_HDR;
 
     // if exceeded limit deallocate postdata buffers and disable redirection
-    if (!(enable_redirection && (redirection_tries < t_state.txn_conf->number_of_redirections))) {
+    if (!(enable_redirection && (redirection_tries <= t_state.txn_conf->number_of_redirections))) {
       this->disable_redirect();
     }
 


zwoop:
Two thoughts:

1. This doesn't seem complete, assuming it's right, since these parameters are used in several places of the HttpSM.
2. This was change ~4 years ago, and as far as I can tell, the PR here reverses the changes made then.

```
commit 1c3a1a41bbf90581a137505f55e06d98fe6aabe1
Author:     Damian Meden <damian.meden@gmail.com>
AuthorDate: Thu May 27 21:06:39 2021 +0100
Commit:     GitHub <noreply@github.com>
CommitDate: Thu May 27 13:06:39 2021 -0700

    Redirect - Make TS to honour the number_of_redirections configuration value (#7867)
    
    This fix includes a unit test to validate the expected behavior.
```

We'll have to have Damien look at this I think.

JosiahWI:
The code change does not seem related to the issue in the title (which I will link). Why does allowing an extra redirection cycle affect whether we cache the response to any of the other redirected requests? I don't think the number of redirections should affect caching whatsoever if the concerns are properly isolated.

How have you confirmed whether or not this patch resolves your issue?

shongkaikai:
> The code change does not seem related to the issue in the title (which I will link). Why does allowing an extra redirection cycle affect whether we cache the response to any of the other redirected requests? I don't think the number of redirections should affect caching whatsoever if the concerns are properly isolated.
> 
> How have you confirmed whether or not this patch resolves your issue?

In the state_read_server_response_header(), When redirection_tries == t_state.txn_conf->number_of_redirections, enable_redirection is set to false, but it should be true at this point.

enable_redirection == false will cause s->cache_info.action = CACHE_DO_NO_ACTION in HttpTransact::handle_forward_server_connection_open, which in turn leads to failure when adding a cache consumer in perform_cache_write_action.

brbzull0:
Hi. I think there is something missing here, this change alone may not fix all things, as this will break some other stuffs. Would it be possible to add an autest for this case?

brbzull0:
[approve ci autest]

bneradt:
[approve ci autest 1]

masaori335:
From my local test, the master and 10.1.x branch doesn't have the issue. It looks like the solution is #11542 + #12215. 
@shongkaikai would you check cherry-picking #11542 and #12215 to the 9.2.x branch?

bryancall:
This is only an issue on the 9.2.x branch and this PR is for master where the issue doesn't exist.

bryancall:
This issue is also related: #10955.  We are going to keep this issue open to track the bug and close others as duplicates.

PR #12374, opened 2025-07-21
cache_promote: Bugfix, don't repurpose policies without labels

### LINKED ISSUE #12368: cache_promote.so adding --stats-enable-with-id requires restart
The `--stats-enable-with-id` argument on cache_promote requires a full ATS restart to function correctly. This isn't documented, and is painful if you want to only enable cache promote statistics for a short time to do some analysis.

To reproduce:

Create a simple remap rule:

```
map http://localhost:8080/batch1 http://localhost:8081/batch1 \
    @plugin=cache_promote.so @pparam=--policy=lru @pparam=--hits=2
```

Start ATS, then edit the remap rule to add `--stats-enable-with-id` and `traffic_ctl config reload`

```
map http://localhost:8080/batch1 http://localhost:8081/batch1 \
    @plugin=cache_promote.so @pparam=--policy=lru @pparam=--hits=2 @pparam=--stats-enable-with-id=remap1
```

Then send some requests and you'll observe the stats remain at 0

```
$ curl -v http://localhost:8080/batch1?max-age=5 
$ traffic_ctl metric match cache_promote
...
plugin.cache_promote.remap1.total_requests 0
...
```

Restart ATS and send another request and you'll see the metric be correctly incremented

```
$ curl -v http://localhost:8080/batch1?max-age=5 
$ traffic_ctl metric match cache_promote
...
plugin.cache_promote.remap1.total_requests 1
...
```

zwoop:
Leaving it up to the RMs for back ports, this has been a bug since v8.1... :/ It's generally not an issue unless you have two identical LRUs / settings, at which point you can get an accidental share.

This fixes #12368.

(there are actually two bugs here, but the stats-id one is less serious than the label issue).

### DIFF
--- a/plugins/cache_promote/cache_promote.cc
+++ b/plugins/cache_promote/cache_promote.cc
@@ -77,13 +77,13 @@ cont_handle_policy(TSCont contp, TSEvent event, void *edata)
           // Do nothing, just let it handle the lookup.
           DBG("cache-status is %d (hit), nothing to do", obj_status);
 
-          if (config->getPolicy()->_stats_enabled) {
+          if (!config->getPolicy()->_stats_id.empty()) {
             TSStatIntIncrement(config->getPolicy()->_cache_hits_id, 1);
           }
           break;
         }
       }
-      if (config->getPolicy()->_stats_enabled) {
+      if (!config->getPolicy()->_stats_id.empty()) {
         TSStatIntIncrement(config->getPolicy()->_total_requests_id, 1);
       }
     } else {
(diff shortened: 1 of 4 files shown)


zwoop:
This fixes #12368

bneradt (approved):

bryancall (approved):

mattyw:
👍 Thanks @zwoop . This fixes the issue as reported. The label issue I was going to report separately but happy to see it was fixed as part of this

cmcfarlen:

PR #12377, opened 2025-07-22
Cripts: Refactor the cache key / URL APIs

zwoop:
- cleans up the notion around cached URLs and headers, and cache keys.
- cleans up the Now / time functionality, allowing for UTC etc.
- adds APIs to set the lookup status as well

Note: This was split out of the Cache Groups PR, since this is universally useful / need in Cripts.

### DIFF
--- a/include/cripts/Headers.hpp
+++ b/include/cripts/Headers.hpp
@@ -142,42 +142,6 @@ public:
 
   }; // End class cripts::Header::Method
 
-  class CacheStatus
-  {
-    using self_type = CacheStatus;
-
-  public:
-    CacheStatus() = delete;
-    CacheStatus(Header *owner) : _owner(owner) {}
-
-    cripts::string_view GetSV();
-
-    operator cripts::string_view() { return GetSV(); }
-
-    cripts::string_view::const_pointer
-    Data()
-    {
-      return GetSV().data();
-    }
-
-    cripts::string_view::size_type
-    Size()
-    {
-      return GetSV().size();
-    }
-
-    cripts::string_view::size_type
-    Length()
-    {
-      return GetSV().size();
-    }
-
-  private:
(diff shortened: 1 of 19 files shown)


zwoop:
[approve ci autest 2]

bneradt:
[approve ci autest 2]

include/cripts/Headers.hpp
```diff
@@ -199,21 +163,17 @@ class Header
...
-    operator=(const char (&str)[N])
-    {
-      return operator=(cripts::string_view(str, str[N - 1] ? N : N - 1));
-    }
+    // template <size_t N>
+    // self_type &
+    // operator=(const char (&str)[N])
+    //{
+    //  return operator=(cripts::string_view(str, str[N - 1] ? N : N - 1));
+    //}
```
bneradt: Maybe just remove this commented out block?

include/cripts/Matcher.hpp
```diff
@@ -20,8 +20,10 @@
 // Setup for PCRE2
 #define PCRE2_CODE_UNIT_WIDTH 8
 #include <pcre2.h>
+#include <algorithm>
 #include <vector>
 #include <tuple>
+#include <algorithm>
```
bneradt: Two `algorithm` imports.

bneradt (changes requested):

include/cripts/Headers.hpp
zwoop: So, I left it here, because at some point it was needed and useful, but it breaks some dependencies in this new code. Let me add a code here, for clarity.

include/cripts/Matcher.hpp
zwoop: Fixed

bneradt (approved):

PR #12379, opened 2025-07-22
Fix nightly OS build issue - null stringview

ezelkow1:
At some point this line changed from && to ||, so now both are evaluated and this is undefined when hostname is null.


ezelkow1:
@hnakamur Would appreciate any input since I think your PR changed this, if there's a reason we shouldnt change this back

ezelkow1:
[approve ci autest 1]

hnakamur:
I confirmed I made a mistake at the commit .086847590ce708ccba0346fbdf06148a08db5b28.
@ezelkow1 Thank you for your fix!

PR #12383, opened 2025-07-23
Cripts: Add Geo APIs to the cripts::IP object

zwoop:
This is step 1 towards a clean / simple abstraction where the effective IP (either from socket, or PROXY protocol, or a header) can be used to do Geo ACLs / lookups.


doc/developer-guide/cripts/cripts-connections.en.rst
```diff
@@ -154,8 +180,40 @@ accessed via the ``Log()`` method. See the ``tcpinfo`` plugin in ATS for details
 Geo-IP
 ======
 
-If ATS has been built with Geo-IP support, the connection object will provide access to the Geo-IP
-data for the connection. The following methods will then be available:
+If ATS has been built with Geo-IP support, both connection objects and IP objects will provide
```
bneradt, line 183: `ATS` -> `|TS|`

bneradt (approved):
I think we should use `|TS|` maybe? Probably not a big deal. I'll approve as is.

PR #12389, opened 2025-07-24
hrw4u: Check operators, add http header validation, url validation

zwoop:
This makes sure we validate all operator qualifiers, and allow for @-headers even though @ isn't allowed by the RFC standards.

### DIFF
--- a/tools/hrw4u/src/symbols.py
+++ b/tools/hrw4u/src/symbols.py
@@ -29,19 +29,17 @@ class SymbolResolver:
         "http.status.reason": ("set-status-reason", Validator.quoted_or_simple(), False, None),
         "http.status": ("set-status", Validator.range(0, 999), False, None),
         "inbound.conn.dscp": ("set-conn-dscp", Validator.nbit_int(6), False, None),
-        "inbound.cookie.": (["rm-cookie", "set-cookie"], Validator.quoted_or_simple(), False, None),
-        "inbound.req.": (["rm-header", "set-header"], Validator.quoted_or_simple(), False, None),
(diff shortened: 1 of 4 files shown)


tools/hrw4u/src/symbols.py
```diff
@@ -29,19 +29,17 @@ class SymbolResolver:
...
+        "outbound.cookie.": (["rm-cookie", "set-cookie"], Validator.http_token(), False, None),
+        "outbound.req.": (["rm-header", "set-header"], Validator.http_token(), False, {"PRE_REMAP", "REMAP", "READ_REQUEST"}),
         "outbound.resp.":
-            (
-                ["rm-header",
-                 "set-header"], Validator.quoted_or_simple(), False, {"PRE_REMAP", "REMAP", "READ_REQUEST", "SEND_REQUEST"}),
+            (["rm-header", "set-header"], Validator.http_token(), False, {"PRE_REMAP", "REMAP", "READ_REQUEST", "SEND_REQUEST"}),
         "outbound.status.reason": ("set-status-reason", Validator.range(0, 999), False, {"PRE_REMAP", "REMAP", "READ_REQUEST"}),
         "outbound.status": ("set-status", Validator.range(0, 999), False, {"PRE_REMAP", "REMAP", "READ_REQUEST"}),
     }
```
bneradt, line 45: The change looks fine, but is this Validator for `set-status-reason` correct? Should that be alpha-numeric? I assume `set-status` takes something like a `200`, while the reason is `OK`?

bneradt (approved):

PR #12391, opened 2025-07-25
Add range request support to cache fill + global plugin support

jasmine-nahrain:
Currently, cache_fill will do a background fetch for a range request but only for that range. These changes remove the range header from the request before making the background fetch so we fetch the whole body and store that into cache.
The config variable --allow-304 has been removed as we do not check the status of the transaction. The background fetch will be triggered without considering the status.
Added two new configs
--range-req-only -> only trigger a background fetch for range requests
--cache-range-req -> turn on/off background fetch for range requests

### DIFF
--- a/plugins/experimental/cache_fill/background_fetch.cc
+++ b/plugins/experimental/cache_fill/background_fetch.cc
@@ -45,17 +45,6 @@ namespace cache_fill_ns
 DbgCtl dbg_ctl{PLUGIN_NAME};
 }
 
-// This is the list of all headers that must be removed when we make the actual background
-// fetch request for range requests.
-static const std::array<const std::string_view, 6> FILTER_HEADERS{
-  {{TS_MIME_FIELD_RANGE, static_cast<size_t>(TS_MIME_LEN_RANGE)},
-   {TS_MIME_FIELD_IF_MATCH, static_cast<size_t>(TS_MIME_LEN_IF_MATCH)},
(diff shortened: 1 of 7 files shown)


jasmine-nahrain:
[approve ci autest 2]

plugins/experimental/cache_fill/background_fetch.cc
```diff
@@ -176,6 +208,12 @@ BgFetchData::initialize(TSMBuffer request, TSMLoc req_hdr, TSHttpTxn txnp)
             if (set_header(mbuf, hdr_loc, TS_MIME_FIELD_HOST, TS_MIME_LEN_HOST, hostp, len)) {
               Dbg(dbg_ctl, "Set header Host: %.*s", len, hostp);
             }
+            // Next, remove the Range headers and IMS (conditional) headers from the request
```
masaori335, line 200: Which hook this will run? Do we really need to remove conditional headers?

jasmine-nahrain: Without removing them, only the range requested will be cached. If we remove them, then the whole body will be cached. This is for the background_fetch request.

jasmine-nahrain:
[approve ci autest 2]

ezelkow1 (approved):
Tested here and it worked for me. Noted one oddity where if the first req issued is a range request you will get back a 200 instead of the full content. However this is also how background fetch is currently implemented so no change in behavior, just something to be aware of

cmcfarlen:

PR #12392, opened 2025-07-25
Add support for storing and retrieving metric type

cmcfarlen:
There is a need to be able to programmatically know the type of a metric for prometheus metrics.  This PR adds support for that by using from bits from the Metric's ID.


bneradt:
[approve ci autest 2]

bneradt:
[approve ci autest 2]

PR #12393, opened 2025-07-25
TLS group name: handle SSL resumption

bneradt:
This is an increment on top of: #12261

This properly handles retrieval of the TLS group name in the cases of TLS resumption. TLS group naming is used in the cqssg log field.


bneradt:
[approve ci autest 2]

maskit:
It doesn't seem like we need to add a new variable to record a group name. Although both OpenSSL and BoringSSL have different APIs for curve and group, the getters are actually just aliases unlike the setters for context objects. We should be able to use the same variable for both curve and group.

`TLSBasicSupport::get_tls_group` which was added by https://github.com/apache/trafficserver/pull/12261 is probably unnecessary. I don't know why the previous code doesn't work for group, but the function to get curve/group id is identical.
```
$ git grep SSL_get_shared_curve include/
include/openssl/ssl.h.in:# define SSL_get_shared_curve          SSL_get_shared_group
```

Also, `SSL_get0_group_name` is only available on OpenSSL 3.2 or later. I see the `ifdef` and `get_tls_group` probably returns `nullptr` on OpenSSL 3.0, but I'd be surprised if there's no way to achieve it on 3.0.

bneradt:
> Also, `SSL_get0_group_name` is only available on OpenSSL 3.2 or later. I see the `ifdef` and `get_tls_group` probably returns `nullptr` on OpenSSL 3.0, but I'd be surprised if there's no way to achieve it on 3.0.

This is a great point. I updated the patch for openssl 3.0 and 3.1 via SSL_get_negotiated_group. Since 3.2, the docs say to prefer the new SSL_get0_group_name, but for 3.0 and 3.1 we can use the former. See:

```cpp
#elif HAVE_SSL_GET_NEGOTIATED_GROUP                    // OpenSSL 3.0/3.1
  int group_nid = SSL_get_negotiated_group(ssl);
  if (group_nid != NID_undef) {
    char const *group_name = OBJ_nid2sn(group_nid);
    return group_name != nullptr ? std::string_view(group_name) : "";
  }
  return "";
```

bneradt:
[approve ci autest 2]

PR #12394, opened 2025-07-26
Examine C++ module support for IOUring

JosiahWI:
This is an experimental patch to showcase C++20 modules. I have chosen the inkuring module because it is one of the smallest components of ATS and the change only impacts a handful of files in other components. EDIT: I have also converted ink_hrtime, because only the OS X pipeline builds inkuring.

I learned a great deal making this change and discovered a couple ways that GCC 15.1 will experience an internal error. One is to `#include` a symbol which is also imported from a module. I think another is to do a circular import; GCC detects simple cycles, but I suspect I found one it missed. That is the reason there is one symbol left declared in a .h file.

I had to fix some compile errors due to missing includes, confirming for me that C++ modules are inherently quite a bit more robust than header files for general usecases. I am not sure what to expect for compile times, but that is something to investigate.

Conclusions from CI pipeline:
- (OSX) CMake doesn't support compiling modules with Unix Makefiles. Unix Makefiles is the default generator on UNIX-like systems, therefore losing the ability to build with it would be a pretty hard hit to take, I think.
- (Format,RAT) The minimum CMake version to build this is somewhere at or above CMake 3.28. I built with CMake 3.31.


bneradt:
This is really interesting @JosiahWI. I'm pleased with how "easy" it is to set up builds for this - it looks like most of the heavy lifting is actually done by cmake. It sounds like we'll probably need to wait until the 11 release when we can update the minimum cmake version, but we should definitely keep this in mind for that.

PR #12396, opened 2025-07-28
Format SDK_Overridable_Configs

masaori335:
When we add a new config, this list has big changes because each line has 2 configs. To avoid it, making one line has one config.


src/api/InkAPITest.cc
```diff
@@ -8653,77 +8653,145 @@ EXCLUSIVE_REGRESSION_TEST(SDK_API_TSHttpConnectServerIntercept)(RegressionTest *
 ////////////////////////////////////////////////
 
 // The order of these should be the same as TSOverridableConfigKey
+// clang-format off
```
masaori335, line 8656: I haven't found good clang-format config to achieve this.

shukitchan (approved):
I think this looks fine

bneradt (approved):
Good idea.

src/api/InkAPITest.cc
JosiahWI: What do you think about adding a comment that we did this to reduce diff sizes?

masaori335: I guess, it's obvious :) (I don't object to add comments, though)

PR #12399, opened 2025-07-28
HRW: Fix parser when conditions are values

zwoop:
The improvements to stricter parsing of int types in #12218 introduces a subtle bug when conditions are used in string values (like "Random is %{RANDOM:100}".


bneradt (approved):

bneradt:
10.1.x cherry-pick in: #12738

PR #12400, opened 2025-07-28
conf_remap: Fix build to add appropriate lib deps

zwoop:


plugins/conf_remap/CMakeLists.txt
```diff
@@ -17,3 +17,5 @@
 
 add_atsplugin(conf_remap conf_remap.cc)
 verify_remap_plugin(conf_remap)
+
+target_link_libraries(conf_remap PRIVATE libswoc::libswoc yaml-cpp::yaml-cpp)
```
JosiahWI, line 21: I see that conf_remap uses libswoc, but where is the yaml-cpp dependency coming from?

JosiahWI (changes requested):

plugins/conf_remap/CMakeLists.txt
zwoop: It comes from this include

```
#include "tsutil/YamlCfg.h"
```

JosiahWI: Would be better not to depend on tsutil, but that's out of scope.

JosiahWI (approved):

bneradt (approved):

PR #12405, opened 2025-07-30
Fix user_agent_session_miss for ATS SSL session cache

bneradt:
proxy.config.ssl.session_cache.value allows the user to configure whether OpenSSL internally manages the TLS session cache for resumption or whether ATS does via callbacks. For our user_agent_session_* metrics, we use the SSL_CTX_sess_* OpenSSL API to populate them. This works for hit, timeout, etc, for both when the session cache is managed by OpenSSL and when it is managed by ATS, with the exception of SSL_CTX_sess_misses. That metric only counts internal cache misses:

> SSL_CTX_sess_misses() returns the number of sessions proposed by
> clients that were not found in the internal session cache in server
> mode.

The result is that when ATS manages the session cache, which is the default configuration, proxy.process.ssl.user_agent_session_miss is always zero. This patch uses our ssl_session_cache_miss metric to augment the stat to be accurate for ATS-managed caching.

### DIFF
--- a/doc/admin-guide/monitoring/statistics/core/ssl.en.rst
+++ b/doc/admin-guide/monitoring/statistics/core/ssl.en.rst
@@ -100,7 +100,13 @@ SSL/TLS
 .. ts:stat:: global proxy.process.ssl.ssl_session_cache_miss integer
    :type: counter
 
+.. ts:stat:: global proxy.process.ssl.ssl_session_cache_timeout integer
+   :type: counter
+
 .. ts:stat:: global proxy.process.ssl.ssl_origin_session_cache_miss integer
+   :type: counter
+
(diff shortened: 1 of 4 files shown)


bneradt:
> Can we do the same for hits as well?
> 
> These functions always return 0 on BoringSSL, so this change would only show the misses mysteriously if BoringSSL is used. I don't really care about the other two, but hits and misses should be a pair.
> 
> * SSL_CTX_sess_accept_good
> * SSL_CTX_sess_hits
> * SSL_CTX_sess_misses
> * SSL_CTX_sess_timeouts

Sounds good. Thank you for representing the BoringSSL perspective. I added a commit to address hits for BoringSSL. We don't to do this for OpenSSL because we would double count then, since SSL_CTX_sess_hits does report the external cache value for OpenSSL.

bneradt:
[approve ci]

PR #12407, opened 2025-07-31
Fix potential stderr closure bug in LogFile::check_fd()

JakeChampion:
The check_fd() function had a critical bug where it would unconditionally close and reopen log files to verify existence. This was dangerous for special files like "stderr" and "stdout" which map to standard file descriptors - closing stderr (fd 2) could break error reporting for the entire process.

Fixed by:
- Skip existence checks entirely for "stderr" and "stdout" files
- Use fstat() on open file descriptors to detect unlinked files instead of access() + close/reopen pattern
- Check st_nlink == 0 to detect when files have been externally deleted
- Fall back to path-based existence check only when fstat() fails


JakeChampion:
> Looks good. You might be fixing an issue commented in the autests too. Can you please try running with this applied to your patch:
> 
> ```diff
> diff --git a/tests/gold_tests/logging/log-filenames.test.py b/tests/gold_tests/logging/log-filenames.test.py
> index a4fdbc73b..ec6e8928a 100644
> --- a/tests/gold_tests/logging/log-filenames.test.py
> +++ b/tests/gold_tests/logging/log-filenames.test.py
> @@ -257,11 +257,4 @@ class stderrTest(LogFilenamesTest):
>  DefaultNamedTest()
>  CustomNamedTest()
>  stdoutTest()
> -
> -# The following stderr test can be run successfully by hand using the replay
> -# files from the sandbox. All the expected output goes to stderr. However, for
> -# some reason during the AuTest run, the stderr output stops emitting after the
> -# logging.yaml file is parsed. This is left here for now because it is valuable
> -# for use during development, but it is left commented out so that it doesn't
> -# produce the false failure in CI and developer test runs.
> -# stderrTest()
> +stderrTest()
> ```
> 
> That patch works fine for me locally with or without your patch, but I wonder whether it's related to what you are fixing.

I pushed that patch up, the tests passed with it ☺️

bneradt (approved):
Thank you for contributing the fix!

PR #12408, opened 2025-07-31
slice plugin: simplify self healing by switching to ident header

traeak:
Previous PR (#12092) added an "ident" header which allowed the CRR plugin to mark a "stale" slice as "fresh" if the the identifiers matched the reference slice.

This PR switches from the slice plugin itself constructing an IMS header and instead rely on the CRR plugin to mark a successful fresh cache lookup as stale if the passed in Ident header doesn't match the cached ident header.

Support for this special IMS header for slice and cache_range_requests is removed in this PR.

Logic in the slice plugin is now simplified as interior range request heal attempt is automatically handled by the CRR plugin.  The slice plugin now only needs to handle the case where a post healed interior slice after refetch by the CRR plugin returns an identifier different from the reference slice. This triggers a reference slice refetch (to be marked stale by the CRR plugin).  Also works to remove a reference slice from cache that 404's at the origin.


plugins/cache_range_requests/cache_range_requests.cc
```diff
@@ -134,25 +133,16 @@ create_pluginconfig(int argc, char *const argv[])
...
 
   for (;;) {
-    int const opt = getopt_long(argc, argv, "i:", longopts, nullptr);
+    int const opt = getopt_long(argc, argv, "j:", longopts, nullptr);
     if (-1 == opt) {
       break;
     }
 
     switch (opt) {
-    case 'c': {
```
serrislew, line 143: the above options should be removed as well line 121 and 123

doc/admin-guide/plugins/cache_range_requests.en.rst
```diff
@@ -158,13 +127,25 @@ to this header.
 
 .. code::
 
-    X-Crr-Ident: Etag: "foo"
-    X-Crr-Ident: Last-Modified: Tue, 19 Nov 2019 13:26:45 GMT
+    X-Crr-Ident: Etag "foo"
```
serrislew, line 130: Is it obvious that this is the Ident order of which they are checked? If not, maybe it would be helpful to state that so even if Stale is explicitly used, it won't be used if Etag is used

plugins/slice/server.cc
```diff
@@ -201,9 +203,11 @@ handleFirstServerHeader(Data *const data, TSCont const contp)
     return HeaderState::Fail;
   }
 
-  // save data header string
-  data->m_datelen = sizeof(data->m_date);
-  header.valueForKey(TS_MIME_FIELD_DATE, TS_MIME_LEN_DATE, data->m_date, &data->m_datelen);
+  /*
```
serrislew: is this meant to be deleted?

plugins/slice/server.cc
```diff
@@ -507,17 +489,6 @@ handleNextServerHeader(Data *const data)
       }
 
       return true;
-
-    } break;
-      // Refetch the reference slice
-    case BlockState::ActiveRef: {
```
serrislew, line 513: why is this state not needed anymore?

plugins/slice/server.cc
```diff
@@ -445,60 +457,30 @@ handleNextServerHeader(Data *const data)
...
+      data->m_lastmodifiedlen = lastmodifiedlen;
+      if (0 < lastmodifiedlen) {
+        strncpy(data->m_lastmodified, lastmodified, lastmodifiedlen);
       }
 
+      // potentially new content length
+      data->m_contentlen = blockcr.m_length;
+
       // Reset for first block
       if (Config::RefType::First == data->m_config->m_reftype) {
```
serrislew, line 479: so now when we see a mismatch with the ref, we set the tags to CRR and refetch ref slice. What is the difference between Active and ActiveInt before? should we be resetting for the first block on the Active state? isn't that processing the current block? does this stop all subsequent requests?

tests/gold_tests/pluginTest/cache_range_requests/cache_range_requests_ident.test.py
```diff
@@ -160,36 +182,63 @@
...
+
+# 9 Test - Ensure "fresh" is in cache
+tr = Test.AddTestRun("0- fresh range cache check")
+ps = tr.Processes.Default
+tr.MakeCurlCommand(curl_and_args + ' http://ident/pathfresh -r 0-', ts=ts)
+ps.ReturnCode = 0
+ps.Streams.stdout.Content = Testers.ContainsExpression("X-Cache: hit-fresh", "expected cache fresh")
+tr.StillRunningAfter = ts
+
+# 10 request with different etag and ensure it goes stale
```
serrislew: should there be a `X-Crr-Ident: Stale` header test too?

plugins/slice/server.cc
traeak: I believe ActiveRef mean active slice refetch.  This is no longer needed as non first slice refetches are entirely handled by the CRR plugin.  Comment must be wrong?

The process is as follows:
- slice plugin requests first slice.  This establishes the identifier (etag or lm).
- slice plugin requests subsequent slices, passing along the first slice identifer.
- CRR plugin during cache lookup compares the identifier to cached (if fresh).  Identifier mismatch causes CRR to mark it stale and pass along to parent.
- slice plugin: if the refetched slice ident mismatches the first slice, the transaction is cancelled. This new slice is established as reference and the old reference slice is refetched through the CRR plugin.

I'll look at doing a better job updating the docs.  With this change there's no need for the slice plugin itself to deal with refetching anything other than the first (reference) slice anymore.

doc/admin-guide/plugins/cache_range_requests.en.rst
traeak: refactored the autest

tests/gold_tests/pluginTest/cache_range_requests/cache_range_requests_ident.test.py
traeak: added

plugins/cache_range_requests/cache_range_requests.cc
traeak: removed

plugins/slice/server.cc
traeak: the new mismatch slice indientifiers including the content length become the new reference.

doc/admin-guide/plugins/cache_range_requests.en.rst
```diff
@@ -112,37 +112,6 @@ which includes information about the partial content range.  In this mode,
 all requests (include partial content) will use consistent hashing method
 for parent selection.
 
-X-Crr-Ims header support
```
serrislew, line 115: are there any guidelines on removing parameters or configs? (i.e. configs are being deprecated notification) or when users upgrade their ATS version and their configs are not longer valid, they are expected to check ATS docs of their replacement?

plugins/cache_range_requests/cache_range_requests.cc
```diff
@@ -261,8 +261,16 @@ range_header_check(TSHttpTxn txnp, pluginconfig *const pc)
   TSMLoc    hdr_loc = TS_NULL_MLOC;
 
   if (TS_SUCCESS == TSHttpTxnClientReqGet(txnp, &hdr_buf, &hdr_loc)) {
+    defer
```
serrislew, line 264: curious if this will be a new standard or just crr specific?

traeak: This is just specific to CRR, but seems to be pretty useful.

doc/admin-guide/plugins/cache_range_requests.en.rst
traeak: I'll put some thought into that.  I had originally tagged this for ats11 but notice it was bumped back to 10.2.

serrislew (approved):

serrislew:
Not sure about guidelines on removing features for plugins but other than that, LGTM

PR #12409, opened 2025-07-31
Change non-boolean enabled configs to mode

bneradt:
Most "enabled" configurations are booleans that toggle on or off a feature. But we have a few that are named "enabled" but actually take an enumeration of values to control behavior. This is confusing terminology. This patch keeps support for the enabled configuration as currently named, adds a new "mode" configuration for these, and deprecates the old "enabled" config in the docs.

proxy.config.ssl.session_cache.enabled was somewhat complicated: with the 10.0.0 release, we implemented it as "value" to try to address the issue of this patch just for that parameter, but kept the "enabled" config description in the docs. So the docs were out of sync with the code. This adds support for "enabled" and immediately deprecates that, since users may have tried to use that, silently keeps support for "value" in case that was used, and adds the now desired "mode" configuration.

If a user passes both the deprecated "enabled" config as well as the new "mode" config, then they must be the same value, otherwise an Emergency message is logged because it is impossible for ATS to discern which value should be used.

Aside from this Emergency log on mismatch just mentioned, this patch should be backwards compatible without introducing any changes in behavior for current configs.


tests/gold_tests/records/legacy_config/full_records.config
```diff
@@ -310,6 +311,8 @@ CONFIG proxy.config.http.connect_ports STRING 443
 CONFIG proxy.config.config_update_interval_ms INT 3000
 CONFIG proxy.config.raw_stat_sync_interval_ms INT 5000
 CONFIG proxy.config.remote_sync_interval_ms INT 5000
+CONFIG proxy.config.log.logging.mode INT 3
+CONFIG proxy.config.log.logging.mode INT 3
```
maskit: Is this intentional duplicated line for testing?

maskit (changes requested):
The description makes sense. Just one possible error in the change.

tests/gold_tests/records/legacy_config/full_records.config
bneradt: Good catch.

brbzull0:
It seems some of the autest got caught up into the new `Emergency`  mismatch.

Linking [this](https://github.com/apache/trafficserver/pull/11902) as it had some discussion related to this.

brbzull0:
100% onboard with this change. 
Now I think some may still do things like this, and using the `debug.enabled` as example as it's probably the most manually  record set.
```
traffic_ctl config set proxy.config.diags.debug.enabled 1
```
The output would be:
```
[9] Error during execution
- [2000] Record not found. [2000]
```

I think would be a good idea, having some mapping for this so if you hit one of the variables in that mapping you could get an error, like:

```
$ traffic_ctl config set proxy.config.diags.debug.enabled 1
Warning: proxy.config.diags.debug.enabled is deprecated, use proxy.config.diags.debug.mode instead.
```


We could even have this inside ATS, but not sure how far we wanna go.

bneradt:
After crawling in the weeds for a while, I think this will be better for the user as a simple rename in 11.x, rather than trying to support both in 10.x:
#12412

PR #12413, opened 2025-08-01
RecordsConfig.cc: Ensure RechCheckT/regex consistency

bneradt:
It is easy to get lost in the meaning of the various RecordElement
specifications when adding a new configuration element. We have several
configurations that have regexes specified to verify sane user input but
do not have a RecCheckT provided. If the RechCheckT is RECC_NULL, the
regex is not used (see RecordValidityCheck). Thus this is always a
mistake.

This patch adds a compile-time check that RecCheckT is non-RECC_NULL if
regex is provided and fixes those that violate this validation. It also
adds as a compile time check that is done as a runtime check in
initialize_record() that if RechCheckT is non-RECC_NULL, regex is
non-nullptr.

### DIFF
--- a/include/records/RecDefs.h
+++ b/include/records/RecDefs.h
@@ -99,16 +99,16 @@ namespace detail
 #define REC_PERSISTENCE_TYPE(P) rec::detail::is_valid_persistence<P>::value
 
 enum RecUpdateT {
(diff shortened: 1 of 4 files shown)


bneradt:
[approve ci fedora]

masaori335 (approved):
Great job!

Issue #12412, opened 2025-08-01
11.x: Rename some enabled configs to mode


bneradt:
Per the discussion here:
https://github.com/apache/trafficserver/pull/11902

It would be clearer for end users if the configs that are named `enabled` were booleans. Currently that is the case for many of them, but the following take an enumeration of values that control behavior:

```
proxy.config.output.logfile.rolling_enabled
proxy.config.mlock_enabled
proxy.config.http.response_server_enabled
proxy.config.log.logging_enabled
proxy.config.log.rolling_enabled
proxy.config.diags.debug.enabled
proxy.config.diags.logfile.rolling_enabled
proxy.config.ssl.session_cache.enabled
```

Each of these would be better off being renamed to `.mode` configs.

I briefly attempted to add support for both the `.mode` versions as well as the `enabled` versions (see #12409), but this becomes very tricky. What if both are in the config and have different values? Emit an emergency, I thought, because how can ATS decide between them? OK, but be sure to handle the case where one is the implicit default. Oh, and be sure to handle plugin overrides. And be careful with those that are overridable configs. Etc. In the end I thought the rules for supporting both would be more confusing for the user and hard to not implement without bugs.

bneradt:
To avoid conflicts, I suggest delaying doing this until we are closer to release. The patch is pretty straightforward to just rename them, but will make patches not align when we do. (Don't forget to update our variable names when the configs are updated.)

PR #12414, opened 2025-08-01
proxy.config.ssl.session_cache.mode

### LINKED ISSUE #11901: ssl.session_cache.enabled unknown

bneradt:
In the 10.0.0 release, we renamed proxy.config.ssl.session_cache.enabled to proxy.config.ssl.session_cache.value because enabled seemed incorrect for a configuration that took a non-boolean enumeration of values. We accidentially left the documentation describing
proxy.config.ssl.session_cache.enabled, however. In discussion about this, we decided to rename such configurations to "mode" instead of value. Another issue (#12412) records making these renames for 11.0.0. This PR just addresses the current issues with
proxy.config.ssl.session_cache.enabled by:

* Adding implementation support for proxy.config.ssl.session_cache.enabled since we documented that already but document it as deprecated.
* Keep implementation support for proxy.config.ssl.session_cache.value since users might be using that.
* Add proxy.config.ssl.session_cache.mode as the preferred new mode.

The user can set values for any of these via:

* implicit default config values,
* plugin default values via TSMgmtIntCreate,
* explicit records.yaml values, or
* override via environment variable.

It will probably be a very rare user that implements any of these across values, but in case they do, this patch implements a selection of the finally chosen config value based on a priority base of the above, with "mode" valued over "value" valued over "enabled".

Fixes: #11901


maskit (approved):

ezelkow1 (approved):

cmcfarlen:
Milestone 10.1.0 via #12419

PR #12416, opened 2025-08-04
server.honor_cipher_order: Clarify documentation

### LINKED ISSUE #12382: Server cipher order setting affects more than cipher suite list.
We noticed when trying to set the TLS Groups config for the server, that it was not honoring client TLS Group/Curves and was instead forcing the client to use the server order.

This comes back to openssl API, `SSL_OP_CIPHER_SERVER_PREFERENCE` which is what is set when you configure (Default) the [proxy.config.ssl.server.honor_cipher_order](https://docs.trafficserver.apache.org/en/latest/admin-guide/files/records.yaml.en.html#proxy-config-ssl-server-honor-cipher-order).

I suggest that we update the docs here to reflect that it actually sets the server order for cipher suites **AND** tls groups/ec curves list **AND** signature algorithms

Just a few weeks ago they changed the code and the docs for the upcoming OpenSSL 3.6 release:

> Introduce SSL_OP_SERVER_PREFERENCE superceding misleadingly named SSL_OP_CIPHER_SERVER_PREFERENCE.
> 
> Michael Baentsch
https://github.com/openssl/openssl/commit/51ce5499f9bd1f12cf08f511faaf163b0c4448bb
https://github.com/openssl/openssl/pull/27981

This appears to be to help clear up confusion around this exact issue, since openssl only supports the server order list for ALL settings, not just the cipher suites.

For ATSv11, it might be worth changing the actual name of the config option in ATS records to reflect that, but that is a larger change.

bneradt:
Clarify in records.yaml.en.rst that server.honor_cipher_order controls TLS server preference for TLS groups and signature algorithms in addition to ciphers.

This also makes use of the SSL_OP_SERVER_PREFERENCE instead of the misleading SSL_OP_CIPHER_SERVER_PREFERENCE when available.

Fixes: #12382

---

# For review

The updated doc can be seen, albeit temporarilly until jenkins cleans up the archive, here:
https://ci.trafficserver.apache.org/view/Github/job/Github_Builds/job/docs/7246/artifact/output/12416/docbuild/html/admin-guide/files/records.yaml.en.html#proxy-config-ssl-server-honor-cipher-order

### DIFF
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -3721,8 +3721,10 @@ SSL Termination
 
 .. ts:cv:: CONFIG proxy.config.ssl.server.honor_cipher_order INT 1
 
-   By default (``1``) |TS| will use the server's cipher suites preferences instead of the client preferences.
-   By disabling it (``0``) |TS| will use client's cipher suites preferences.
+   By default (``1``) |TS| will use the server's preferences for cipher suites, supported groups, and
+   signature algorithms instead of the client preferences. By disabling it (``0``) |TS| will use the
+   client's preferences. Note that despite the configuration name mentioning "cipher_order", this
+   setting controls server preference for multiple aspects of TLS negotiation, not just cipher suites.
 
 .. ts:cv:: CONFIG proxy.config.ssl.server.prioritize_chacha INT 0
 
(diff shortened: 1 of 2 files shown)


maskit (approved):

shukitchan (approved):

cmcfarlen:
Milestone 10.1.0 via #12420

PR #12419, opened 2025-08-04
10.1.x: proxy.config.ssl.session_cache.mode (#12414)

bneradt:
In the 10.0.0 release, we renamed proxy.config.ssl.session_cache.enabled to proxy.config.ssl.session_cache.value because enabled seemed incorrect for a configuration that took a non-boolean enumeration of values. We accidentially left the documentation describing
proxy.config.ssl.session_cache.enabled, however. In discussion about this, we decided to rename such configurations to "mode" instead of value. Another issue (#12412) records making these renames for 11.0.0. This PR just addresses the current issues with
proxy.config.ssl.session_cache.enabled by:

* Adding implementation support for proxy.config.ssl.session_cache.enabled since we documented that already but document it as deprecated.
* Keep implementation support for proxy.config.ssl.session_cache.value since users might be using that.
* Add proxy.config.ssl.session_cache.mode as the preferred new mode.

The user can set values for any of these via:

* implicit default config values,
* plugin default values via TSMgmtIntCreate,
* explicit records.yaml values, or
* override via environment variable.

It will probably be a very rare user that implements any of these across values, but in case they do, this patch implements a selection of the finally chosen config value based on a priority base of the above, with "mode" valued over "value" valued over "enabled".

Fixes: #11901
(cherry picked from commit 87bb60259b25734e651e5019b410a9ccf8cbd371)


cmcfarlen:
[approve ci]

PR #12425, opened 2025-08-05
AuTest: cache after following redirect

masaori335:
Add AuTest to cover the https://github.com/apache/trafficserver/issues/12369 case, cache after following redirect.

The master branch and 10.1.x branch should pass without any changes.


masaori335:
I observed failure against the 10.0.x branch. This test says below as expected.
```
[INFO]: Equals Violation: Different. Key: "/a/path/resource", Field Name: "x-cache", Correct Value: "hit-fresh", Actual Value: "miss" : 18
```

cmcfarlen:

PR #12429, opened 2025-08-05
Add log messages about QUIC cert loading for consistency

maskit:
There are "loading" message with quic/ssl label, but there's only "finished" message for ssl.

The added code is almost just a copy and paste from the one for SSL.


maskit:
[approve ci autest 2]

bneradt (approved):
Like you say, it like you copy and pasted fine to me. :)

PR #12433, opened 2025-08-06
header_rewrit: Move template functions into a header file

maskit:
Autest header_rewrite_url failed on my dev env. Apparently the specialized template function is not used, and the generic one does not consider EXT condition modifier.


zwoop:
I have a rewrite for all autest of header rewrite. I'd prefer not messing with those until I've landed my mega patch.

zwoop (changes requested):
The point of having this in the cc file is to avoid unnecessary code duplications. I'd like to understand what the issue is here before we consider this change.

PR #12437, opened 2025-08-07
Error messages when server closes connection

moonchen:
A server can close a connection while we're expecting a response.  For example, the server may close the connection due to a keep-alive timeout.  This results in a 502 response to the client.  Log an error in this case, so that the administrator has a chance to adjust the keep-alive timeout on the ATS side to avoid this situation.

### DIFF
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -1033,7 +1033,6 @@ HttpSM::state_read_push_response_header(int event, void *data)
 
   switch (event) {
   case VC_EVENT_EOS:
-    log_server_close_with_origin(t_state, "Server closed connection while reading PUSH response header.");
     _ua.get_entry()->eos = true;
     // Fall through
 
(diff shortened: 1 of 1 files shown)


bneradt (changes requested):
Looks like we're OK with this from our PR/issue meeting. But let's include the origin URL.

src/proxy/http/HttpSM.cc
bneradt, line 1051: The new message says "Server closed...", but the code around these events seem to indicate client events while waiting for the response. Maybe "Client closed..." is appropriate? Or maybe I'm misunderstanding.

src/proxy/http/HttpSM.cc
```diff
@@ -1945,6 +1971,7 @@ HttpSM::state_read_server_response_header(int event, void *data)
 
   switch (event) {
   case VC_EVENT_EOS:
+    log_server_close_with_origin(t_state, "Server closed connection while reading response header");
```
bneradt: Minor: maybe also end this in a period to be consistent with the other log.

bneradt (changes requested):
Looks good. Thank you for adding the URL.

src/proxy/http/HttpSM.cc
moonchen: You're right.  This isn't a server close as I thought.  I'm going to remove this message and only keep the one where the server closes.

src/proxy/http/HttpSM.cc
moonchen: Done

PR #12438, opened 2025-08-08
escalate plugin: --get-method-only

bneradt:
Implement --get-method-only for the escalate plugin, limiting the failover feature to GET method requests only. All other requests of different methods are not failed over.


shukitchan (approved):
This looks good. We had been doing using lua instead. But i think this is better

bneradt:
Replaced by #12448

PR #12439, opened 2025-08-10
This cleans up some of the hrw4u example tests

zwoop:
This also helps the upcoming tests for u4rhw to reuse / run against the same output and input and validate properly.


zwoop:
[approve ci rocky]

PR #12446, opened 2025-08-12
Retry connect if server closes before receiving response

cmcfarlen:
This moves the check for retry or not into the `VC_EVENT_EOS` case and no longer does a fall-through.  Looking at the comment, it looks like the intent was to stop retrying when bytes were received, but neglected to handle the EOS case so the fall-through would unconditionally end retries.  This is not the behavior from 9.2, which would retry here.

This is certainly a funny edge-case in the protocol, and since we have sent the request to the server at this point, it is possible that the server has changed state and retries might not be appropriate.  It seems more likely that we wrote the request to an already closed connection, so a retry should be fine.

Since it's open to interpretation or situational, it might be something to add a config option for.  For now though, I'd like to just restore the 9.2 behavior except in the case when we have received some bytes and the connection has closed.

### DIFF
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -1933,13 +1933,15 @@ HttpSM::state_read_server_response_header(int event, void *data)
   switch (event) {
   case VC_EVENT_EOS:
     server_entry->eos = true;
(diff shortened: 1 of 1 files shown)


cmcfarlen:

PR #12447, opened 2025-08-12
escalate: Add x-escalate-redirect header (#1092)

bneradt:
If the escalate.so plugin escalates a request, add the x-escalate-redirect header as an indicator that it did so. Adding the header can be disabled via: @pparam=--no-redirect-header


shukitchan (dismissed):

ezelkow1:
Looks good, though my only question would be do we want to enable this new behavior by default? Since it would be changing how this acts for current users, rather than just making it disabled by default

ezelkow1 (dismissed):

cmcfarlen (approved):

PR #12448, opened 2025-08-12
escalate.so: --escalate-non-get-methods

bneradt:
Implement --escalate-non-get-methods for the escalate plugin. By default, only GET methods are escalated because typical failover servers are caches of the original and cannot handle non-idempotent requests. However, for situations where non-GET is supported, this method allows the admin to escalate those requests as well.

---

This replaces #12438: the concept is similar, but with this PR the default behavior is to avoid redirecting non-GET methods with the option provided to escalate them if wanted.


bneradt:
[approve ci centos]

Issue #12432, opened 2025-08-06
Autest for stats_over_http plugin fails if ENABLE_ALLOCATOR_METRICS is true


maskit:
Apparently `proxy.process.allocator.inuse.ioBufAllocator[0]`, the brackets are not valid characters for prometheus metric name.

PR #12450, opened 2025-08-13
stats_over_http: respect Prometheus character constraints

bneradt:
With certain configurations, metrics can have '[' and ']' characters. Prometheus does not like these. This patch enforces that the metrics get exported with Prometheus's advertised character set limitations:

[a-zA-Z_:][a-zA-Z0-9_:]*

Fixes: #12432


maskit (approved):

cmcfarlen:

PR #12451, opened 2025-08-13
Add the inverse hrw4u tool, u4wrh

zwoop:
This also cleans up some other things in hrw4u in general.

Idea from Mo Chen.


tools/hrw4u/requirements.txt
```diff
@@ -1,4 +1,4 @@
-#
+requirements.txt#
```
bneradt: Is this a typo?

bneradt (changes requested):
Good idea.

bneradt:
[approve ci rocky]

PR #12452, opened 2025-08-13
Print note when listen is delayed

masaori335:
I faced odd situation that ATS started but not listening ports. If I enable debug log, this message appears. 
```
[Aug 13 06:49:40.513] traffic_server DIAG: <traffic_server.cc:2304 (main)> (http_listen) Delaying listen, waiting for cache initialization
```
It looks like ATS delayed listen to wait for cache initialization. Probably, the root cause is a bad disk, but this message should in the `Note` log to give us what's going on.

### DIFF
--- a/src/traffic_server/traffic_server.cc
+++ b/src/traffic_server/traffic_server.cc
@@ -829,7 +829,7 @@ CB_After_Cache_Init()
     // The delay_listen_for_cache value was 1, therefore the main function
     // delayed the call to start_HttpProxyServer until we got here. We must
     // call accept on the ports now that the cache is initialized.
-    Dbg(dbg_ctl_http_listen, "Delayed listen enable, cache initialization finished");
+    Note("Delayed listen enable, cache initialization finished");
     start_HttpProxyServer();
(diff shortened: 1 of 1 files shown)


src/traffic_server/traffic_server.cc
```diff
@@ -829,7 +829,7 @@ CB_After_Cache_Init()
     // The delay_listen_for_cache value was 1, therefore the main function
     // delayed the call to start_HttpProxyServer until we got here. We must
     // call accept on the ports now that the cache is initialized.
-    Dbg(dbg_ctl_http_listen, "Delayed listen enable, cache initialization finished");
+    Note("Delayed listen enable, cache initialization finished");
```
bneradt: I know this isn't your wording, but now that this is becoming a NOTE, can you please reword to "Enabling listen, cache ..." so that it corresponds with the "Delaying listen, ..."

masaori335:
@bneradt Reworded. Please take another look.

bneradt (approved):

cmcfarlen:

PR #12453, opened 2025-08-13
Catch2: update to v3.9.1

bneradt:
This updates our Catch2 version from v2.13.8 to the latest v3.9.1. It also transitions us from the deprecated monolithic header file to the new library model using CMake via FetchContent.

This has several advantages:

* This uses the library model which the authors of Catch2 advocate for. The monolithic header we used before this patch is not "the primarily supported option".
* Since not all of catch.hpp is built for every UT, compilation times are quicker.
* Since the new v3 version provides a main function definition via Catch2::Catch2WithMain, we no longer need the empty unit_test_main.cc files we had before.
* The updated Catch v3 is purported to execute its checks more efficiently (according to the changelog).

As a part of this transition, this removes
src/records/unit_tests/unit_test_main_on_eventsystem.cc since that was added with #7704 for DynamicStats which has since been removed. In v2, this simply didn't run any tests. In v3, this failed because no tests were run.

---

Note:
* This is a unit test only change. Aside from adding const to `operator swoc::IPAddr()` in ink_inet.h
* Everyone of our tests have at least their includes updated, but the vast majority of the file changes is simply the addition of Catch2 itself into `lib/`.


lib/CMakeLists.txt
```diff
@@ -19,6 +19,26 @@
...
 
+if(BUILD_TESTING)
+  include(FetchContent)
+  fetchcontent_declare(
+    Catch2
+    GIT_REPOSITORY https://github.com/catchorg/Catch2.git
+    GIT_TAG v3.9.1
+  )
+  fetchcontent_makeavailable(Catch2)
+
```
bneradt: This is one of the ways the Catch docs recommend retrieving the source. Are there any concerns about using FetchContent for ATS? I'm guessing it's fine, but we don't use it elsewhere yet and wanted to know whether @JosiahWI or @cmcfarlen (or others of course) had any concerns.

bneradt: Thanks for the feedback @JosiahWI. To avoid fetching during builds I created a clone in lib/Catch2 like we have for lib/swoc, etc.

lib/Catch2/.conan/build.py
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env python
```
cmcfarlen: Could probably remove files from other build systems, bezel, Conan.  Also the other dot files and directories. not a big deal.

cmcfarlen (approved):
Thanks for updating our catch2 library!

lib/Catch2/.conan/build.py
bneradt: Good point. I added these for removal and updated the PR:
```
 .bazelrc .conan/ .github/
```

PR #12457, opened 2025-08-15
[cache_promote] Add --disable-on-redirect option

masaori335:
We faced an issue with combination of `cache_promote` and `cachekey` plugin on following redirect. When the `cachekey` plugin tweaks cache key and it'll be the same before and after following redirect, `cache_promote` plugin's LRU counter doesn't work expected and response from origin server is not cached.

To resolve this situation, a new option `--disable-on-redirect` is introduced. That makes `cache_promote` plugin runs only once.

### DIFF
--- a/tests/gold_tests/pluginTest/cache_promote/replay/cache_promote.replay.yaml.tmpl
+++ b/tests/gold_tests/pluginTest/cache_promote/replay/cache_promote.replay.yaml.tmpl
@@ -1,0 +1,166 @@
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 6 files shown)


tests/gold_tests/pluginTest/cache_promote/replay/cache_promote.replay.yaml.tmpl
```diff
@@ -0,0 +1,166 @@
...
+      headers:
+        fields:
+          - [ uuid, 1-2 ]
+          - [ x-debug, x-cache ]
+
+    proxy-response:
+      status: 200
+      headers:
+        fields:
+        - [ x-cache, {{value: "hit-fresh", as: equal }} ]
```
masaori335, line 166: Note: without the `--disable-on-following` option, this case fails.

bneradt:
[approve ci]

bneradt:
[approve ci autest 1]

cmcfarlen:

PR #12459, opened 2025-08-15
ESI: --allowed-response-codes

bneradt:
Adds --allowed-response-codes to the ESI plugin for response code filtering for which ESI transformations will take place. Some origins may add X-Esi headers to every response, even 5xx responses or the like. Setting those up for transformation can be wasteful.


bneradt:
[approve ci]

shukitchan (approved):
It looks good. Thanks.

PR #12460, opened 2025-08-18
Implement probe-full-json feature for xdebug plugin

bneradt:
The probe-full-json directive emits the same data as probe but in a
complete JSON object containing nodes for client/server request headers,
response body, and client/server response headers.

### DIFF
--- a/plugins/xdebug/xdebug_headers.cc
+++ b/plugins/xdebug/xdebug_headers.cc
@@ -41,6 +41,8 @@ namespace
 class EscapeCharForJson
 {
 public:
+  EscapeCharForJson(bool full_json) : _full_json(full_json) {}
+
   std::string_view
   operator()(char const &c)
   {
@@ -49,19 +51,23 @@ public:
     }
     if ((IN_NAME == _state) && (':' == c)) {
       _state = BEFORE_VALUE;
-      return {"' : '"};
+      if (_full_json) {
+        return {R"(":")"};
+      } else {
+        return {"' : '"};
+      }
     }
     if ('\r' == c) {
       return {""};
     }
     if ('\n' == c) {
-      std::string_view result{_after_value()};
+      std::string_view result{_after_value(_full_json)};
 
       if (BEFORE_NAME == _state) {
         return {""};
       } else if (BEFORE_VALUE == _state) {
         // Failsafe -- missing value -- this should never happen.
-        result = _missing_value();
+        result = _missing_value(_full_json);
       }
       _state = BEFORE_NAME;
       return result;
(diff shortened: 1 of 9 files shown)


mlibbey:
wonder if I can sneak in a feature request here -- it would be nice to have the next-hop IP that ats is talking to in the probe. eg, it includes the Host: header, but not the actual IP that it talked to.

bneradt:
> wonder if I can sneak in a feature request here -- it would be nice to have the next-hop IP that ats is talking to in the probe. eg, it includes the Host: header, but not the actual IP that it talked to.

Sounds interesting. Let's do that as a separate PR. Can you please file an issue @mlibbey? Maybe we can include that in both probe and probe-full-json.

plugins/xdebug/xdebug_headers.cc
```diff
@@ -191,14 +222,70 @@ print_response_headers(TSHttpTxn txn, std::stringstream &output)
...
+{
+  TSMBuffer buf_c, buf_s;
+  TSMLoc    hdr_loc;
+
+  bool has_client = false;
+
+  Dbg(dbg_ctl_hdrs, "Printing client request headers for full JSON");
+  if (TSHttpTxnClientReqGet(txn, &buf_c, &hdr_loc) == TS_SUCCESS) {
+    output << "{\"client-request\":{";
+    print_headers(buf_c, hdr_loc, output, !JSON_COMPLIANT);
```
cmcfarlen: What does the `_json` version of print request headers pass `false` for the `json_compliant` flag?

plugins/xdebug/xdebug_headers.h
```diff
@@ -0,0 +1,83 @@
...
+#include <sstream>
+#include "ts/ts.h"
+
+namespace xdebug
+{
+
+/**
+ * Whether to include whitespace in the output.
+ */
+static constexpr bool JSON_COMPLIANT = true;
```
cmcfarlen: This constant and having `json_compliant` passed around is confusing.  Why do we need this? What determines when it should be true? Why is sometimes a variable passed and sometimes hardcoded to this?  Some functions have a flag and some have a `_json` suffix.

cmcfarlen (commented):
I'm not sure if I should request a change or I just don't understand what's going on but this PR is difficult to follow.  I don't understand why there are so many checks on `json_compliant` variables and what that means.  Shouldn't there just be be two "printers", one that does the old thing and one does compliant json?

plugins/xdebug/xdebug_headers.cc
bneradt: You're right. I had renamed this variable to try to make it less confusing, but inverted the symantic without realizing it. I'll update this. In fact, I'll dispense of the generic "json_compliant" verbage and just use "full_json" to make it more obvious that it ties to the xprobe_full_json user option.

plugins/xdebug/xdebug_headers.h
bneradt: Agreed. I updated the naming to all be "full_json" to corresond directly with the user option. Can you please re-review and let me know what you think.

Thanks for the feedback.

PR #12462, opened 2025-08-20
Cripts: Fix standalone query URI component

zwoop:
This fixes a regression introduced with the Convenience APIs, and is a rare edge cases where URI components can be created outside of an owning URL.


zwoop:
I *think* this needs to go into 10.1 as well, assuming the convenience APIs are in 10.1.

cmcfarlen:

PR #12463, opened 2025-08-20
xdebug probe-full-json: encoding the origin body

bneradt:
For aything beyond very simple text bodies, JSON parsers for probe-full-json were breaking upon the content of the body (such as html bodies, for example). This patch adds escaping and hex encoding of body content to preserve the JSON format. It also adds a way to not include the body if the user so desires.


bneradt:
[approve ci]

bneradt:
[approve ci]

mlibbey:
Super useful! Both having real json, and the nobody option!

plugins/xdebug/xdebug_transforms.cc
```diff
@@ -82,11 +84,107 @@ getPostBodyFullJson(TSHttpTxn txn)
...
+      if (c < 0x20) {
+        char esc[6];
+        esc[0]                  = '\\';
+        esc[1]                  = 'u';
+        esc[2]                  = '0';
+        esc[3]                  = '0';
+        static const char hex[] = "0123456789abcdef";
+        esc[4]                  = hex[(c >> 4) & 0x0F];
+        esc[5]                  = hex[c & 0x0F];
+        TSIOBufferWrite(output_buffer, esc, 6);
```
cmcfarlen: Seems like you could call `write_hex` here and reuse that code.

plugins/xdebug/xdebug_transforms.cc
```diff
@@ -82,11 +84,107 @@ getPostBodyFullJson(TSHttpTxn txn)
...
+        written += 6;
+      } else {
+        TSIOBufferWrite(output_buffer, reinterpret_cast<const char *>(&data[i]), 1);
```
cmcfarlen: This can just be `&c` rather than recast?

plugins/xdebug/xdebug_transforms.cc
```diff
@@ -82,11 +84,107 @@ getPostBodyFullJson(TSHttpTxn txn)
...
+{
+  // Convert each byte to two hex characters
+  static const char hex_chars[] = "0123456789abcdef";
+
+  // Process in chunks to keep stack usage reasonable
+  const int64_t CHUNK = 1024; // 1KB of raw -> 2KB hex
+  int64_t       idx   = 0;
+
+  while (idx < len) {
+    int64_t           take = (len - idx > CHUNK) ? CHUNK : (len - idx);
```
cmcfarlen: imo, `std::min(len-idx, CHUNK)` reads better.

plugins/xdebug/xdebug_transforms.cc
```diff
@@ -82,11 +84,107 @@ getPostBodyFullJson(TSHttpTxn txn)
...
+    std::vector<char> hex_output(static_cast<size_t>(take * 2));
```
cmcfarlen: Use `std::array<char, CHUNK*2>`.  vector will allocate every time through the loop.

cmcfarlen (changes requested):
I started reviewing this yesterday but got pulled away. I'll look more today.

plugins/xdebug/xdebug_transforms.cc
bneradt: Good point.

plugins/xdebug/xdebug_transforms.cc
bneradt: Agreed.

plugins/xdebug/xdebug_transforms.cc
bneradt: Good point. I'll have to make sure the `written` value is updated appropriately though...

cmcfarlen (approved):
Thanks! This is neat.

PR #12465, opened 2025-08-22
Fallback to malloc for larger regex buffers

cmcfarlen:
Some configurations of PCRE lean harder on the context allocator.  In this case, the small buffer in RegexMatches is not adequate so we start returning nullptr.  This PR improves the allocator by falling back to malloc for larger buffers, and then freeing them properly.

### DIFF
--- a/src/tsutil/Regex.cc
+++ b/src/tsutil/Regex.cc
@@ -129,11 +129,7 @@ RegexMatches::RegexMatches(uint32_t size)
   pcre2_general_context *ctx = pcre2_general_context_create(&RegexMatches::malloc, &RegexMatches::free, static_cast<void *>(this));
 
   pcre2_match_data *match_data = pcre2_match_data_create(size, ctx);
-  if (match_data == nullptr) {
-    // buffer was too small, allocate from heap
-    debug_assert_message(false, "RegexMatches data buffer too small, increase the buffer size in Regex.h");
-    match_data = pcre2_match_data_create(size, RegexContext::get_instance()->get_general_context());
-  }
+  debug_assert_message(match_data, "Failed to allocate pcre2 match data from custom context");
 
   _MatchData::set(_match_data, match_data);
 }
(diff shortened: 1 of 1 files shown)


cmcfarlen:
[approve ci]

src/tsutil/Regex.cc
```diff
@@ -152,8 +151,18 @@ RegexMatches::malloc(size_t size, void *caller)
     return ptr;
   }
 
-  // return nullptr if buffer is too small
-  return nullptr;
+  return ::malloc(size);
```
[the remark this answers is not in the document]
bneradt: We don't generally check for null on malloc - things will be so obviously bad that it's not worth cluttering the code with it.

src/tsutil/Regex.cc
```diff
@@ -126,8 +126,7 @@ struct RegexMatches::_MatchData {
 //----------------------------------------------------------------------------
 RegexMatches::RegexMatches(uint32_t size)
 {
-  pcre2_general_context *ctx =
-    pcre2_general_context_create(&RegexMatches::malloc, [](void *, void *) -> void {}, static_cast<void *>(this));
+  pcre2_general_context *ctx = pcre2_general_context_create(&RegexMatches::malloc, &RegexMatches::free, static_cast<void *>(this));
 
   pcre2_match_data *match_data = pcre2_match_data_create(size, ctx);
```
bryancall, line 131: With the original code `RegexMatches::malloc()` would return  nullptr if there is no space in the buffer and it will fall back to malloc and free in the `pcre2_match_data_create()` call below.   I didn't call `free()` if the buffer had enough space already, the no-op free above.

There is a unit test that covers the case where we run out of buffer space.

If you would rather do your approach with checking the pointer to see if it needs to be freed then you should also remove the code below that checks to see if there is a nullptr and falling back to `malloc()` and `free()`.

Maybe just adding better comments to the original code so people know how the fallback to malloc and free works.

bryancall: The file for the unit tests is: `test_Regex.cc`

cmcfarlen:

PR #12466, opened 2025-08-22
Comment out the matcher header_rewrite test

cmcfarlen:
The test I added got much more complicated to link if cripts is in the picture, so I commented it out.  If we want to test that the matcher can be case insensitive, we could do it another way (autest perhaps?).


zwoop (approved):

bneradt (approved):

cmcfarlen:
[approve ci autest 0]

PR #12470, opened 2025-08-26
Skip SO_REUSEPORT for AF_UNIX sockets

cmcfarlen:
unix sockets don't support reuse port, so this call isn't needed and on some kernel versions actually returns an error.  So just skip this call for domain sockets.


maskit:
https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net.git/commit/?id=5b0af621c3f6

bneradt (approved):

cmcfarlen:
Added to milestone 10.1.1 via #12474

PR #12472, opened 2025-08-27
Fix validation regex for proxy.config.http2.default_buffer_water_mark

masaori335:
Default value is `-1`, but the validation doesn't allow `-`.


cmcfarlen:

PR #12476, opened 2025-08-29
Add a per client connection max exempt list cript

bneradt:
Note that this builds on top of #12198, so feel free to ignore that commit when reviewing this. I kept the units of work separate to make this specific cripts work easier to review.

---

This updates our build system to be able to make pre-compiled cripts.
Thus, adding a cript is now as easy as adding a plugin. Simply use
add_cript instead of add_atsplugin in CMakeLists.txt, and it will build
your cript for you.

As a part of this, this adds connection_exempt_list.cript, our first
cript plugin that sets the per client exempt list.

### DIFF
--- a/src/api/InkAPI.cc
+++ b/src/api/InkAPI.cc
@@ -8947,12 +8947,15 @@ TSReturnCode
 TSConnectionLimitExemptListAdd(std::string_view ip_ranges)
 {
   swoc::TextView ip_ranges_tv{ip_ranges};
+
   while (auto ip_range_tv = ip_ranges_tv.take_prefix_at(',')) {
     swoc::IPRange ip_range;
+
     if (!ip_range.load(ip_range_tv)) {
       return TS_ERROR;
     }
     bool success = ConnectionTracker::add_client_exempt_range(ip_range);
+
     if (!success) {
       return TS_ERROR;
     }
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -170,7 +170,6 @@ rc/trafficserver.service
 .libs/
 
 .svn/
-.vscode/
```
zwoop, line 173: I'd prefer not removing this, I have quite a few files in my .vscode :)

plugins/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@
 #######################
 
 include(add_atsplugin)
+include(add_cript)
```
zwoop, line 19: I think, but not certain (i'm not a cmake expert) that this is redundant.

include/cripts/Epilogue.hpp
```diff
@@ -752,10 +752,12 @@ TSPluginInit(int argc, const char *argv[])
     inst->NeedCallback(enabled_txn_hooks);
     TSContDataSet(contp, context);
     TSHttpHookAdd(TS_HTTP_TXN_START_HOOK, contp); // This acts similarly to the DoRemap callback
+  } else if (needs_glb_init) {
```
zwoop: I don't remember much of this code, but would this not leak the context and inst now ?

zwoop (changes requested):

.gitignore
bneradt: Makes sense. Per your suggestion offline, this seems to work:

```patch
-.vscode/
+.vscode/*
+!.vscode/settings.json```

plugins/CMakeLists.txt
bneradt: Yes, you're correct.

I think we want to keep it here, but remove it from `plugins/experimental/connection_exempt_list/CMakeLists.txt`. I did so and pushed.

include/cripts/Epilogue.hpp
bneradt: I think you're right. I added the cleanup in there now too.

zwoop (approved):

PR #12481, opened 2025-09-05
test: make test_layout independent of TS_ROOT

raboof:
So running the tests with TS_ROOT set works as well.


maskit:
Thank you for the fix. I'm wondering if we need the `unsetenv` on the other tests in test_layout as well. If you run all the tests then the one you added may work for other tests as well, but if you run just one test like `test_tscore [update_sysconfdir]` you may see errors.

cmcfarlen:

PR #12483, opened 2025-09-05
RAM cache stats and logs updates

moonchen:
* Add counters for hits to the other two kinds of memory caches that we have.
* Add a counter for reads that miss all memory caches.
* Change hits to the write aggregation buffer to HIT_RAM, since the aggregation buffer is RAM.

The goals of this change are:

1. Capture all RAM cache hits in stats and logs in some way.
2. Maintain backwards compatibility with existing log processing and stats collection.


ezelkow1 (approved):
Tested here, after updates could see hits being logged in the agg buffer

ezelkow1:
[approve ci autest 1]

PR #12484, opened 2025-09-06
prscs: proxy response status code setter

bneradt:
In complex setups where there are multiple plugins that can be setting
an error response, it can be non-trival to determine which component
intervened for any given transaction. This adds the prscs log field,
proxy response status code setter, which provides an identifying string
for the component (a plugin name, or ip_allow, for instance) that set
the transaction's response status.


tests/gold_tests/pluginTest/header_rewrite/header_rewrite.test.py
```diff
@@ -27,25 +29,46 @@
...
+    "body": ""
+}
 server.addResponse("sessionfile.log", request_header, response_header)
+
 ts.Disk.records_config.update(
     {
         'proxy.config.diags.debug.enabled': 1,
         'proxy.config.diags.show_location': 0,
-        'proxy.config.diags.debug.tags': 'header.*',
+        'proxy.config.diags.debug.tags': 'header|http',
```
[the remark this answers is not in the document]
bneradt: Nah, you don't need `.*` for debug tag logging.

src/api/InkAPI.cc
```diff
@@ -2948,9 +2948,21 @@ TSHttpHdrStatusSet(TSMBuffer bufp, TSMLoc obj, TSHttpStatus status)
   SET_HTTP_HDR(h, bufp, obj);
   ink_assert(static_cast<HdrHeapObjType>(h.m_http->m_type) == HdrHeapObjType::HTTP_HEADER);
   h.status_set(static_cast<HTTPStatus>(status));
+
+  if (txnp != nullptr && !setter.empty()) {
+    sdk_assert(sdk_sanity_check_txn(txnp) == TS_SUCCESS);
+    HttpSM *sm                               = reinterpret_cast<HttpSM *>(txnp);
+    sm->t_state.http_return_code_setter_name = setter;
+  }
```
cmcfarlen, line 2955: This seems out of place here.  This function was just operating on its arguments, but the change to also update the transaction is only situationally relevant.  Wouldn't setting the return code setter be a separate function?

cmcfarlen: `TSHttpTxnStatusSet` also sets `sm->t_state.http_return_code`, but that doesn't happen here which seems like the two fields could get out of sync?

bneradt:
Thanks for the review. A quick answer to this question:

> Also, wasn't there some discussion about removing proxy from the log field name?

The concern was not with "proxy" but with my previous "plugin" terminology. "plugin" is wrong because aspects of ATS core can set the response, not just plugins. This now uses "proxy". I think that's better. "proxy" in prscs fits well with the other p* log fields which are also proxy related.

src/api/InkAPI.cc
bneradt: Sounds good. Yes, it is ugly to have to pass txnp for TSHttpHdrStatusSet.

I added a commit that avoids this by associating the txn with the TSMLoc when the Req/Rsp is retrieved. That way, later, we can look it up in the associated map. How does this look?

cmcfarlen (approved):

PR #12485, opened 2025-09-08
Make sure we don't create a tunnel when no request body is present.

brbzull0:
This patch makes sure that we do not create the tunnel either with CL==0
 or no request body present.

I found out that in some cases the tunnel were created anyway when no body was present and some requests ended up in 408.
The main reason was the tunnel was created with CL = -1, so the tunnel was timing out waiting for data to be read on an “invalid” tunnel.


src/proxy/http/HttpSM.cc
```diff
@@ -6308,9 +6308,11 @@ HttpSM::do_drain_request_body(HTTPHdr &response)
 void
 HttpSM::do_setup_client_request_body_tunnel(HttpVC_t to_vc_type)
 {
-  if (t_state.hdr_info.request_content_length == 0) {
-    // No tunnel is needed to transfer 0 bytes. Simply return without setting up
-    // a tunnel nor any of the other related logic around request bodies.
+  if (t_state.hdr_info.request_content_length == 0 ||
```
maskit: `t_state.hdr_info.request_content_length == 0` should be unnecessary because `has_request_body` checks `t_state.hdr_info.request_content_length > 0`.

maskit (approved):

cmcfarlen:

PR #12486, opened 2025-09-08
Add tests for compress plugin handling of chunked responses

JakeChampion:
add a new test suite to verify that the cache-write process does not crash when using the compress plugin with chunked origin responses


bryancall:
@JakeChampion Should this still be marked draft, or do you want someone to review it?

JakeChampion:
@bryancall this is still a draft, not yet ready for a full review but I am happy if people want to glance over and leave any comments if there are things which should be changed/improved - I am still trying to get the test to fail, we see the below assert fail for our production traffic:
>Fatal: /trafficserver/src/proxy/http/HttpTunnel.cc:1005: failed assertion `c->skip_bytes <= c->buffer_reader->read_avail()`

I'm trying to write a test which will also fail this assert, then hopefully I can also write a patch that will fix the issue 🤞

JakeChampion:
hmm this isn't right, i will close this and look further into it

PR #12490, opened 2025-09-09
Fix pqus marshaling

bneradt:
The pqus scheme marshaling logic used "strlen" which accidentally picked up LogAccess::strlen rather than the intended std strlen. This resulted in padded-counting rather than actual strlen counting, which threw off our buffer counting logic and led to corruption and/or crashing.

The updated ip_allow.test.py reproduced a crash without the LogAccess.cc patch included in this commit. With the LogAccess patch, the test now reproduces the expected scheme.

### DIFF
--- a/src/proxy/logging/LogAccess.cc
+++ b/src/proxy/logging/LogAccess.cc
@@ -161,7 +161,7 @@ LogAccess::marshal_proxy_host_ip(char *buf)
 int
 LogAccess::marshal_process_uuid(char *buf)
 {
-  int len = round_strlen(TS_UUID_STRING_LEN + 1);
+  int len = padded_length(TS_UUID_STRING_LEN + 1);
 
   if (buf) {
     const char *str = const_cast<char *>(Machine::instance()->process_uuid.getString());
@@ -211,7 +211,7 @@ LogAccess::marshal_config_str_var(char *config_var, char *buf)
 {
   auto str{RecGetRecordStringAlloc(config_var)};
   auto c_str{ats_as_c_str(str)};
-  int  len = LogAccess::strlen(c_str);
+  int  len = LogAccess::padded_strlen(c_str);
   if (buf) {
     marshal_str(buf, c_str, len);
   }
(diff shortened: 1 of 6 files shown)


src/proxy/logging/LogAccess.cc
```diff
@@ -1940,17 +1940,9 @@ LogAccess::marshal_client_req_url_scheme(char *buf)
...
-  // calculate the padded length only if the actual length
-  // is not zero. We don't want the padded length to be zero
-  // because marshal_mem should write the DEFAULT_STR to the
-  // buffer if str is nil, and we need room for this.
-  //
-  if (alen) {
-    plen = round_strlen(alen + 1); // +1 for trailing 0
+    alen = ::strlen(str);
   }
+  plen = round_strlen(alen + 1); // +1 for trailing 0
```
bneradt: * The `::` prefix before strlen is the only real production change.
* I simplified the plen calculation because the preceding logic makes it so that alen is always non-empty. This current plen logic was copied from elsewhere (the method marshaling) and is not applicable here.

src/proxy/logging/LogAccess.cc
```diff
@@ -1940,17 +1940,9 @@ LogAccess::marshal_client_req_url_scheme(char *buf)
...
-  }
-
-  // calculate the padded length only if the actual length
-  // is not zero. We don't want the padded length to be zero
-  // because marshal_mem should write the DEFAULT_STR to the
-  // buffer if str is nil, and we need room for this.
-  //
-  if (alen) {
-    plen = round_strlen(alen + 1); // +1 for trailing 0
+    alen = ::strlen(str);
```
bneradt, line 1943: This is the key change: use `::strlen` instead of what was `LogAccess:strlen`.

cmcfarlen (approved):
Thanks! Changing the name to something that doesn't shadow global symbols will prevent this trap from catching others out!

cmcfarlen:
Added to milestone 10.1.1 via #12519

PR #12492, opened 2025-09-09
Docs: Correct the mismatch between the documentation and actual behavior

maskit:
PR #9383 added the fallback mechanism for ports that are configured to accept PROXY protocol, and the existence of PROXY header is no longer mandatory.


cmcfarlen:

Issue #12443, opened 2025-08-11
Failure to build ATS 10.1.0-rc1 on Debian 13


ksqrtr:
```
[236/1195] Building CXX object src/proxy/logging/CMakeFiles/logging.dir/LogFilter.cc.o
FAILED: src/proxy/logging/CMakeFiles/logging.dir/LogFilter.cc.o 
/usr/bin/c++ -DOPENSSL_API_COMPAT=10002 -DOPENSSL_IS_OPENSSL3 -DPACKAGE_NAME="\"Apache Traffic Server\"" -DPACKAGE_VERSION=\"10.1.0\" -Dlinux -I/home/vagrant/work/trafficserver/include -I/home/vagrant/work/trafficserver/build/include -I/home/vagrant/work/trafficserver/lib/swoc/include -I/home/vagrant/work/trafficserver/lib/yamlcpp/include -isystem /usr/include/jemalloc -isystem /home/vagrant/work/trafficserver/lib/systemtap -isystem /opt/include -pthread -O3 -g -DNDEBUG -std=c++20 -fdiagnostics-color=always -Wno-invalid-offsetof -pipe -Wall -Wextra -Wno-noexcept-type -Wsuggest-override -Wno-vla-extension -fno-strict-aliasing -Wno-format-truncation -mcx16 -Wno-deprecated-declarations -Werror -MD -MT src/proxy/logging/CMakeFiles/logging.dir/LogFilter.cc.o -MF src/proxy/logging/CMakeFiles/logging.dir/LogFilter.cc.o.d -o src/proxy/logging/CMakeFiles/logging.dir/LogFilter.cc.o -c /home/vagrant/work/trafficserver/src/proxy/logging/LogFilter.cc
In file included from /home/vagrant/work/trafficserver/lib/swoc/include/swoc/swoc_ip.h:21,
                 from /home/vagrant/work/trafficserver/lib/swoc/include/swoc/bwf_ip.h:15,
                 from /home/vagrant/work/trafficserver/src/proxy/logging/LogFilter.cc:34:
In copy constructor ‘swoc::_1_5_13::IPRangeView::storage_type::storage_type(const swoc::_1_5_13::IPRangeView::storage_type&)’,
    inlined from ‘swoc::_1_5_13::IPRangeView::IPRangeView(const swoc::_1_5_13::IPRangeView&)’ at /home/vagrant/work/trafficserver/lib/swoc/include/swoc/IPRange.h:604:7,
[...]

PR #12493, opened 2025-09-09
Remove maybe problematic constructors on storage_type

cmcfarlen:
This might fix the Debian tripe build with gcc 14.2.0.

It complains about using maybe uninialized values, but commenting these and using the
defaults seems to work.


cmcfarlen:
Fixes #12443

cmcfarlen:
[approve ci autest 2]

PR #12494, opened 2025-09-09
esi: Add URL to many error messages

bneradt:
It was difficult to track down some issues with ESI because ops didn't easily know which URLs some error messages belonged to. This adds identifying URLs to many of the error messages.

### DIFF
--- a/plugins/esi/combo_handler.cc
+++ b/plugins/esi/combo_handler.cc
@@ -45,6 +45,18 @@
 using namespace std;
 using namespace EsiLib;
 using namespace swoc::literals;
+#include "http_utils.h"
+
+/** Extracts a URL string from buffer and location objects.
+ *
+ * @param[in] bufp The buffer containing the URL data.
+ * @param[in] url_loc The memory location pointing to the URL within the
+ *   buffer.
+ *
+ * @return The URL. Returns UNKNOWN_URL_STRING if the URL cannot be retrieved due to
+ *   API failures or invalid parameters.
+ */
+// Use shared helper from url_utils.h
 
 #define DEBUG_TAG "combo_handler"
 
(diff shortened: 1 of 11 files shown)


shukitchan:
I will take a look at it this week

shukitchan:
[approve ci osx]

plugins/esi/combo_handler.cc
```diff
@@ -585,18 +597,21 @@ getClientRequest(TSHttpTxn txnp, TSMBuffer bufp, TSMLoc hdr_loc, TSMLoc url_loc,
   int         query_len;
   const char *query = TSUrlHttpQueryGet(bufp, url_loc, &query_len);
 
+  // Get URL string for error reporting.
+  std::string url_string = getUrlString(bufp, url_loc);
```
cmcfarlen: In the happy path, this is unused.  Could move this call so it only happens when necessary.

plugins/esi/http_utils.cc
```diff
@@ -0,0 +1,67 @@
...
+  TSMLoc      url_loc;
+
+  if (TSHttpTxnClientReqGet(txnp, &req_bufp, &req_hdr_loc) == TS_SUCCESS) {
+    if (TSHttpTxnPristineUrlGet(txnp, &req_bufp, &url_loc) == TS_SUCCESS) {
+      int   length     = 0;
+      char *url_string = TSUrlStringGet(req_bufp, url_loc, &length);
+      if (url_string) {
+        request_url.assign(url_string, length);
+        TSfree(url_string);
+      }
```
cmcfarlen: Seems like this could just call the new `getUrlString`.  Consider returning `std::optional<std::string>` instead of a magic value if that seems appropriate.

cmcfarlen (changes requested):

plugins/esi/combo_handler.cc
bneradt: Good point.

plugins/esi/http_utils.cc
bneradt: > Seems like this could just call the new getUrlString.

Yes, good find.

> Consider returning std::optional<std::string> instead of a magic value if that seems appropriate.

Let's stick with `UNKNOWN_URL_STRING` since that gets used by the caller for the error message when the URL could not be received. Otherwise we'll have a bunch of `url_string ? url_string.value() : "(unknown)"` calls.

shukitchan (approved):
It looks fine. Thanks.

cmcfarlen (approved):

PR #12501, opened 2025-09-10
Logging: Add chiv field from real-ip plugin

serrislew:
Adds `chiv` logging field that is verified by real-ip plugin. If no IP is available, `chi` value is used

Dependent on #12359 changes to expose verified client IP address
- Update: merged

### DIFF
--- a/doc/admin-guide/logging/formatting.en.rst
+++ b/doc/admin-guide/logging/formatting.en.rst
@@ -480,6 +480,7 @@ Network Addresses, Ports, and Interfaces
 
 .. _chi:
 .. _chih:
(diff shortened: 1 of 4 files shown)


bneradt (approved):

bneradt (commented):
The implementation looks fine.

Are we sure we don't want to fallback to `-` though if there is no verified IP? How will an ops person distinguish between a verified IP and just chi if it falls back to chi?

bneradt (approved):
Let's move forward with what you have. If we later want a log field that is just the explicitly verified IP and `-` if it isn't verified, we can add that later.

PR #12502, opened 2025-09-11
Fix background fill on TLS connections

### LINKED ISSUE #12473: Background fill does not work on HTTPS connections
There is ATS built-in background fill feature that is to continue fetching documents from origins even after client connections are aborted. The name is confusing but this is different from background_fetch plugin.

The background *fill* feature is enabled by default (`proxy.config.http.background_fill_active_timeout=0`, and `proxy.config.http.background_fill_completed_threshold=0`), and it works fine if the client connection is cleartext HTTP, but I found that it does not work if the client connection is HTTPS (TLS + TCP).

I checked the behavior like below:
- Run ATS (port 8080)
- Run origin server (`$ cat | nc -l 8888`)
- Send a request (`$ curl -v http://127.0.0.1:8080/`)
- Close the client connection (`Ctrl+C` on the curl command)
- Send a response (Type a response on the cat command)
- Shutdown the origin server (`Ctrl+D` on the cat command)
- Send another request (`$ curl -v http://127.0.0.1:8080/`)

Same steps on HTTPS doesn't return the response from cache.

masaori335:
Fix #12473 

Prior to this change, EOS event handling in `HttpSM::state_watch_for_client_abort` was completely different from underlaying protocols. If it's TLS and client aborts early, ATS kills HttpTunnel and terminates HttpSM immediately. Which means ATS doesn't wait response from origin server and the background fill feature can't kick in. This change the behavior to almost the same as plaintext HTTP.

TLS 1.3 has half-open support, so we should implement it, but it's out of scope from this PR.

Co-authored-by: Masakazu Kitajo <maskit@apache.org>

### DIFF
--- a/tests/gold_tests/h2/http2_rst_stream.test.py
+++ b/tests/gold_tests/h2/http2_rst_stream.test.py
@@ -58,8 +58,7 @@ tr.Processes.Default.Streams.All += Testers.ContainsExpression(
 tr.Processes.Default.Streams.All += Testers.ContainsExpression(
     'Submitted RST_STREAM frame for key 1 on stream 1.', 'Send RST_STREAM frame.')
 
-# NOTE: Background Fill fix changed this behavior
-# server.Streams.All += Testers.ExcludesExpression('RST_STREAM', 'Server is not affected.')
+server.Streams.All += Testers.ContainsExpression('RST_STREAM', 'Origin Server received RST_STREAM frame.')
 
 ts.Disk.traffic_out.Content += Testers.ContainsExpression('Received HEADERS frame', 'Received HEADERS frame.')
 


src/iocore/net/UnixNetVConnection.cc
```diff
@@ -625,6 +625,11 @@ UnixNetVConnection::net_write_io(NetHandler *nh)
   NetState     *s = &this->write;
   Continuation *c = this->write.vio.cont;
 
+  if (c == nullptr) {
+    // If Continuatin to callback is nullptr, we can do nothing
```
bneradt: Continuatin -> Continuation

src/iocore/net/SSLNetVConnection.cc
```diff
@@ -897,6 +899,49 @@ SSLNetVConnection::do_io_close(int lerrno)
...
+{
+  if (get_tunnel_type() == SNIRoutingType::BLIND) {
+    // we don't have TLS layer control of blind tunnel
+    UnixNetVConnection::do_io_shutdown(howto);
+    return;
+  }
+
+  switch (howto) {
+  case IO_SHUTDOWN_READ:
+    // No need to call SSL API
```
bneradt, line 913: For the comment: probably good to explain why the API should not be called, rather than just say it doesn't need to be called.

tests/gold_tests/h2/http2_rst_stream.test.py
```diff
@@ -58,7 +58,8 @@
 tr.Processes.Default.Streams.All += Testers.ContainsExpression(
     'Submitted RST_STREAM frame for key 1 on stream 1.', 'Send RST_STREAM frame.')
 
-server.Streams.All += Testers.ExcludesExpression('RST_STREAM', 'Server is not affected.')
+# NOTE: Background Fill fix changed this behavior
+# server.Streams.All += Testers.ExcludesExpression('RST_STREAM', 'Server is not affected.')
```
bneradt: Should this be a ContainsExpression then? Otherwise, maybe best just to remove these lines.

bneradt (changes requested):
Looks good.

tests/gold_tests/h2/http2_rst_stream.test.py
masaori335: Thank you for taking a look, I was wondering what this check should do. Changed to `ContainsExpression` to make sure origin server also received the `RST_STREAM` frame.

bneradt (approved):
Looks good to me.

cmcfarlen:

PR #12505, opened 2025-09-11
Adds content type parameter handling to compress plugin

JakeChampion:
Introduces a new `content_type_ignore_parameters` configuration option that allows Content-Type pattern matching to ignore parameter substrings like charset specifications.

When enabled, patterns like `"application/json"` will match responses with `"application/json; charset=utf-8"`, improving compression coverage for responses that include Content-Type parameters, without having to have `*` on the end of patterns anymore.

Defaults to false to maintain backward compatibility with existing configurations.

### DIFF
--- a/plugins/compress/configuration.cc
+++ b/plugins/compress/configuration.cc
@@ -200,7 +200,7 @@ strip_params(std::string_view v)
 {
   swoc::TextView tv{v};
   tv = tv.take_prefix_at(';');
-  tv.rtrim_if([](int ch) { return std::isspace(ch); });
+  tv.rtrim_if(&::isspace);
   return tv;
 }
 


plugins/compress/configuration.cc
```diff
@@ -190,6 +191,18 @@ HostConfiguration::is_status_code_compressible(const TSHttpStatus status_code) c
...
+static std::string_view
+strip_params(std::string_view v)
+{
+  if (auto pos = v.find(';'); pos != std::string_view::npos) {
+    v = v.substr(0, pos);
+  }
+  // trim trailing spaces
+  while (!v.empty() && isspace(static_cast<unsigned char>(v.back())))
+    v.remove_suffix(1);
+  return v;
```
bneradt: Let's refactor this to use swoc::TextView. I think take_prefix_at and rtrim should be useful here.

plugins/compress/configuration.h
```diff
@@ -58,10 +58,22 @@ class HostConfiguration : private atscppapi::noncopyable
...
+      minimum_content_length_(1024),
+      content_type_ignore_parameters_(false)
   {
   }
 
+  bool
+  content_type_ignore_parameters()
+  {
+    return content_type_ignore_parameters_;
+  }
```
bneradt, line 70: Looks like this can be a `const` function.

tests/gold_tests/pluginTest/compress/compress-content-type-params.test.py
```diff
@@ -0,0 +1,72 @@
...
+            {
+                f"""
+map /ignore-params-false/ http://127.0.0.1:{self.server.Variables.http_port}/ \
+    @plugin=compress.so \
+    @pparam={Test.RunDirectory}/ignore-params-false.config
+map /ignore-params-true/ http://127.0.0.1:{self.server.Variables.http_port}/ \
+    @plugin=compress.so \
+    @pparam={Test.RunDirectory}/ignore-params-true.config
+"""
+            })
```
bneradt: Add lines more naturally takes a list of lines. Here's an example from another test:

```python
tm.Disk.remap_config.AddLines(
    [
        f"map http://alpha.ex http://alpha.ex:{pv_port}",
        f"map http://bravo.ex http://bravo.ex:{pv_port}",
        f"map http://charlie.ex http://charlie.ex:{pv_port}",
        f"map http://delta.ex http://delta.ex:{pv_port}",
    ])
```

bneradt (changes requested):
Looks great to me. Just a few minor comments. @shukitchan would also be good to review.

I appreciate the well-organized class based test using proxy verifier.

plugins/compress/configuration.cc
bneradt: I think this will work:

```cpp
std::string_view
strip_params(std::string_view v)
{
  swoc::TextView tv{v};
  tv = tv.take_prefix_at(';');
  tv.rtrim_if(&isspace);
  return tv;
}
```

bneradt: For reference, I host libswoc docs here along with the rendered doxygen:
https://libswoc.brianneradt.com/

tests/gold_tests/pluginTest/compress/compress-content-type-params.test.py
JakeChampion: Thank you, I have updated to use a list now 🙇 [`1c94e05` (#12505)](https://github.com/apache/trafficserver/pull/12505/commits/1c94e053ed46ed59a9a73f42367761f87109616c)

plugins/compress/configuration.h
JakeChampion: oh yes! Applied in [`1af6c59` (#12505)](https://github.com/apache/trafficserver/pull/12505/commits/1af6c5937408666575a9e0cfce9f32628f275e55)

plugins/compress/configuration.cc
JakeChampion: swoc::TextView looks good, I could not get `tv.rtrim_if(&isspace)` to compile, i went with this instead

```c++
std::string_view
strip_params(std::string_view v)
{
  swoc::TextView tv{v};
  tv = tv.take_prefix_at(';');
  tv.rtrim(swoc::CharSet(" \t\n\r\f\v"));
  return tv;
}
```

https://github.com/apache/trafficserver/pull/12505/commits/f07156d0c0dc921a8d96bbe2ce6969f951f15395

bneradt: We should be able to get isspace to work. It would be nice to avoid the list of characters. We use it in other plugins, like here:

https://github.com/apache/trafficserver/blob/2128398aa01cc624da4acbe1a3153cc5da35cfb8/plugins/header_rewrite/matcher.h#L210

And here:
https://github.com/apache/trafficserver/blob/2128398aa01cc624da4acbe1a3153cc5da35cfb8/plugins/origin_server_auth/origin_server_auth.cc#L623

Trying to be helpful: did you include `#include <cctype>` ?

shukitchan (approved):

plugins/compress/configuration.cc
JakeChampion: The only way I can get `isspace` to work is to cast it like, otherwise the compiler can't resolve which version to use - i did this in [`247e887` (#12505)](https://github.com/apache/trafficserver/pull/12505/commits/247e88757432e09738b5d6046fc50b8ef0000eba)

```
tv.rtrim_if(static_cast<int (*)(int)>(&isspace));
```

shukitchan:
I try the copilot. I also wonder if the static cast is needed at all. The other code suggestion is also interesting.

plugins/compress/configuration.cc
```diff
@@ -190,6 +195,15 @@ HostConfiguration::is_status_code_compressible(const TSHttpStatus status_code) c
   return it != compressible_status_codes_.end();
 }
 
+std::string_view
+strip_params(std::string_view v)
+{
+  swoc::TextView tv{v};
+  tv = tv.take_prefix_at(';');
+  tv.rtrim_if([](int ch) { return std::isspace(ch); });
```
shukitchan: Actually, I wonder if this will work?

`tv.rtrim_if(&isspace);`

shukitchan: and we can combine the two lines as well?

`tv = tv.take_prefix_at(';').rtrim_if(&isspace);`

JakeChampion: > Actually, I wonder if this will work?
> 
> `tv.rtrim_if(&isspace);`

This does not compile unfortunately

plugins/compress/configuration.cc
bneradt: Oh, the issue is `using namespace std;` at the top of the file. That interferes with template deduction for `isspace` because there is both `std::isspace` and `::isspace`. Can you try this instead (effectively, `&::isspace`):

```patch
diff --git a/plugins/compress/configuration.cc b/plugins/compress/configuration.cc
index 1f1ebf960..b93dee8b3 100644
--- a/plugins/compress/configuration.cc
+++ b/plugins/compress/configuration.cc
@@ -200,7 +200,7 @@ strip_params(std::string_view v)
 {
   swoc::TextView tv{v};
   tv = tv.take_prefix_at(';');
-  tv.rtrim_if([](int ch) { return std::isspace(ch); });
+  tv.rtrim_if(&::isspace);
   return tv;
 }
 ```

bneradt (approved):
Looks good to me. Thank you for persevering with the `isspace` use.

PR #12504, opened 2025-09-11
background_fetch: add Unix domain support.

vuori:
The `background_fetch` plugin would crash when the original request arrived over a Unix domain socket. This adds support for `sockaddr_un` client addresses, and also a missing `return false` to prevent an assert from failing later after the error branch falls through.

This has only been tested on the 10.1.x branch.

Closes #12488

PR #12510, opened 2025-09-15
lua: minimal support for Unix socket incoming connections

vuori:
This adds minimal support to the Lua plugin for cases where the incoming connection is through a Unix socket. Crashing issues are unlikely, but there are two type confusions where an `AF_UNIX` sockaddr will get treated as `AF_INET6`.

Basically same deal as #12504


shukitchan:
I will take a look at it this week

doc/admin-guide/plugins/lua.en.rst
```diff
@@ -1838,6 +1838,7 @@ Socket address family
 
     TS_LUA_AF_INET (2)
     TS_LUA_AF_INET6 (10)
+    TS_LUA_AF_UNIX (1)
```
shukitchan, line 1841: I think these variables are really used when we are making changes to the outgoing address. So there is no need to support AF_UNIX here

plugins/lua/ts_lua_server_request.cc
```diff
@@ -149,6 +149,9 @@ ts_lua_inject_server_request_server_addr_api(lua_State *L)
 
   lua_pushinteger(L, AF_INET6);
   lua_setglobal(L, "TS_LUA_AF_INET6");
+
+  lua_pushinteger(L, AF_UNIX);
+  lua_setglobal(L, "TS_LUA_AF_UNIX");
```
shukitchan, line 154: I think these variables are really used when we are making changes to the outgoing address. So there is no need to support AF_UNIX here

shukitchan (changes requested):
I also think that we need to make changes in ts_lua_vconn.cc
ts_lua_vconn_get_remote_addr() is also not taking care of the situation correctly. 

Thanks.

vuori:
`ts_lua_vconn_get_remote_addr` support was added.

`(TS_LUA_)AF_UNIX` is now returned as the third value from `ts_lua_client_request_client_addr_get_addr` `ts_lua_vconn_get_remote_addr` so the new constant is needed for comparisons in Lua code.

shukitchan:
Thanks. 
I am still not seeing your changes on `ts_lua_vconn_get_remote_addr`
Perhaps there is some error with the branch?

vuori:
Sorry, git error.

shukitchan (approved):

shukitchan:
On second thought, perhaps we should just take care of the rest consistently. 

may be we should fix up `ts_lua_server_request.cc` as well. 
- There are places where we are retrieving the ip address or port of the next destination. Even though they are not going to be AF_UNIX, for completeness sake, we should just take care of handling only for AF_INET/AF_INET6 and use the default for anything else.
- There are also places where we can `set` address in this file. It makes no sense for anything other than INET/INET6. So we can just error out gracefully in those places as well if invalid values are passed. 

Thanks so much for your patience and help.

shukitchan (changes requested):
Please also consider my latest notes and perhaps to make further changes to ts_lua_server_request.cc

Thanks.

vuori:
I'll try to look at this during the next week.

shukitchan:
@vuori are you still working on this? Thanks.

vuori:
Hopefully yes sometime soonish, but the last weeks have been non-stop firefighting on other fronts and there are still a few blazes to put out this week.

PR #12511, opened 2025-09-15
Add a freebsd ifdef check for tcpinfo retrans

ezelkow1:
Tested building cripts on freebsd and hit this issue with the tcpinfo struct differences. I dont think retrans is implemented on freebsd so this probably wont give valid data, but at least it will build now.


cmcfarlen:

PR #12512, opened 2025-09-15
API for Strategy rebind during transaction

traeak:
This adds to the ts api to allows the selected parent strategy to be changed during a transaction.

APIs added:

- void TSHttpTxnNextHopStrategySet(TSHttpTxn txnp, void *const strategy);
- TSReturnCode TSHttpTxnNextHopStrategyGet(TSHttpTxn txnp, void **strategy);
- TSReturnCode TSHttpTxnNamedNextHopStrategyGet(char const *name, void **strategy);

### DIFF
--- a/include/proxy/http/HttpTransact.h
+++ b/include/proxy/http/HttpTransact.h
@@ -724,11 +724,11 @@ public:
     //  able to defer some work in building the request
     TransactFunc_t pending_work = nullptr;
 
-    HttpRequestData                           request_data;
-    ParentConfigParams                       *parent_params     = nullptr;
-    std::shared_ptr<NextHopSelectionStrategy> next_hop_strategy = nullptr;
-    ParentResult                              parent_result;
-    CacheControlResult                        cache_control;
+    HttpRequestData           request_data;
+    ParentConfigParams       *parent_params     = nullptr;
+    NextHopSelectionStrategy *next_hop_strategy = nullptr;
+    ParentResult              parent_result;
+    CacheControlResult        cache_control;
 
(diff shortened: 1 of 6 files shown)


zwoop:
Did you email the mailing list with the new API additions ?

traeak:
I think more tests need to be added for corner cases.  Also on request am working on setting up a stress test case to ensure that converting the strategy shared_ptr to plain pointers doesn't result in faults.

bryancall:
@jrushford Hey, do you want to take a look at this PR for us?

jrushford:
@bryancall Yes, I've been talking with Bryan about it.  I'll take a look but, give me a day as I'm currently traveling.

jrushford:
> I think more tests need to be added for corner cases. Also on request am working on setting up a stress test case to ensure that converting the strategy shared_ptr to plain pointers doesn't result in faults.

+1 on more stress testing.  I was worried that changing from shared_ptr to plain pointers might result in memory leaks or crashes when remap.config and strategies.yaml is reloaded.  So far though it looks good to me.  I'll spin up some VM's and test it.

traeak:
I stress tested using real time traffic replay (with strategies enabled), constantly forcing reloads every few seconds for several hours.  I didn't observe any hiccups or issues.  (A shame tsan is messy at the moment).  Chris' presentation during the summit seemed to agree with my assessment that UrlRewrite which holds the StrategyFactory lives as long as any active transaction are still around, ensuring the strategies live as long as necessary.

jrushford (approved):
I've reviewed it and it looks fine to me and I discussed the stress testing with Brian and am satisfied that this is ok.

PR #12514, opened 2025-09-15
test: make test_layout independent of TS_ROOT part 2

maskit:
Apply the fix from #12481 to the other tests as well.


cmcfarlen:

PR #12521, opened 2025-09-16
header_rewrite: Add set-effective-address operator

maskit:
This adds a new header_rewrite operator to set client's effective address. It internally calls `TSHttpTxnVerifiedAddrSet`, and the address will be available for ATS core and other plugins.

The name is not `set-verified-address` because the operator also changes the behavior of header_rewrite plugin by changing `INBOUND_IP_SOURCE` to `PLUGIN` so the address is available for other conditions and operators.

I could add `set-verified-address` instead but it would probably be called always with `set-plugin-cntl INBOUND_IP_SOURCE PLUGIN` if a verified address is set by header rewrite rules. So I made the convenient operator that does the both.


tests/gold_tests/pluginTest/header_rewrite/header_rewrite_effective_address.test.py
```diff
@@ -0,0 +1,56 @@
...
+ts.Disk.plugin_config.AddLine(f'header_rewrite.so {Test.RunDirectory}/{conf_name}')
+ts.Disk.remap_config.AddLine(f'map http://www.example.com http://127.0.0.1:{server.Variables.Port}')
+
+# Test that the IP address in Real-IP request header is returned in Effective-IP response header.
+expected_output = "gold/header_rewrite_effective_address.gold"
+tr = Test.AddTestRun()
+tr.MakeCurlCommand(
+    f'--http1.1 -H "Host: www.example.com" -H "Real-IP: 1.2.3.4" --verbose "http://127.0.0.1:{ts.Variables.port}"', ts=ts)
+tr.Processes.Default.ReturnCode = 0
+tr.Processes.Default.StartBefore(server, ready=When.PortOpen(server.Variables.Port))
```
bneradt: microserver actually has a pretty robust ready condition with debug logging and the like.

maskit: Removed the custom ready condition.

PR #12522, opened 2025-09-18
redirect cleanup: cleanup plugins and tunnels

bneradt:
This adds more cleanup logic to redirect handling logic, cleaning up the tunnel and PluginVCs on redirect. This addresses a timeout crash we received on dispatch to a defunct plugin upon following a redirect from the escalate plugin.


bneradt:
Making this a draft because I want to bake this in Yahoo prod for longer to make sure this patch is stable.

bneradt:
Closing as this seemed to be the more relevant fix: #12561

PR #12524, opened 2025-09-19
watchdog: alert on ET_NET thread stalls beyond threshold

moonchen:
Adds a watchdog thread that warns when a net thread remains in the work phase longer than a configurable duration.

**Config:**
  proxy.config.thread_watchdog.timeout_ms (default: 1000)

**Why:**
  Net threads should not stall; doing so adds latency to all transactions multiplexed on that thread. Stalls may indicate a misbehaving plugin, overload, or a Traffic Server bug.

On trigger, a warning is logged with the offending thread number and elapsed time to aid targeted diagnostics.

### DIFF
--- a/src/traffic_server/traffic_server.cc
+++ b/src/traffic_server/traffic_server.cc
@@ -2138,10 +2138,10 @@ main(int /* argc ATS_UNUSED */, const char **argv)
   init_memory_tracker(nullptr, RECD_NULL, RecData(), nullptr);
 
   // Start the watchdog
-  int                  watchdog_timeout_ms = RecGetRecordInt("proxy.config.thread_watchdog.timeout_ms").value_or(1000);
-  std::span<EThread *> net_threads{eventProcessor.thread_group[ET_NET]._thread,
-                                   static_cast<size_t>(eventProcessor.thread_group[ET_NET]._count)};
-  watchdog = std::make_unique<Watchdog::Monitor>(net_threads, std::chrono::milliseconds{watchdog_timeout_ms});
+  int watchdog_timeout_ms = RecGetRecordInt("proxy.config.thread_watchdog.timeout_ms").value_or(1000);
+  watchdog                = std::make_unique<Watchdog::Monitor>(eventProcessor.thread_group[ET_NET]._thread,
+                                                                static_cast<size_t>(eventProcessor.thread_group[ET_NET]._count),
+                                                                std::chrono::milliseconds{watchdog_timeout_ms});
 
   {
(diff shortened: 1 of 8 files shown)


moonchen:
This is a PR that would be nicer if we had `std::jthread` and `stop_token` support across all platforms.  Ubuntu 20, FreeBSD 13, and macOS 14 are too old.

src/traffic_server/traffic_server.cc
```diff
@@ -2131,6 +2137,12 @@ main(int /* argc ATS_UNUSED */, const char **argv)
   RecRegisterConfigUpdateCb("proxy.config.dump_mem_info_frequency", init_memory_tracker, nullptr);
   init_memory_tracker(nullptr, RECD_NULL, RecData(), nullptr);
 
+  // Start the watchdog
+  int watchdog_timeout_ms = RecGetRecordInt("proxy.config.thread_watchdog.timeout_ms").value_or(1000);
+  watchdog                = std::make_unique<Watchdog::Monitor>(eventProcessor.thread_group[ET_NET]._thread,
```
zwoop: Should we make this optional, as in, if proxy.config.thread_watchdog == 0, we don't setup the watchdog ?

moonchen: Done.

moonchen:
[approve ci autest]

src/traffic_server/traffic_server.cc
```diff
@@ -2131,6 +2137,14 @@ main(int /* argc ATS_UNUSED */, const char **argv)
   RecRegisterConfigUpdateCb("proxy.config.dump_mem_info_frequency", init_memory_tracker, nullptr);
   init_memory_tracker(nullptr, RECD_NULL, RecData(), nullptr);
 
+  // Start the watchdog
+  int watchdog_timeout_ms = RecGetRecordInt("proxy.config.exec_thread.watchdog.timeout_ms").value_or(1000);
+  if (watchdog_timeout_ms > 0) {
+    watchdog = std::make_unique<Watchdog::Monitor>(eventProcessor.thread_group[ET_NET]._thread,
+                                                   static_cast<size_t>(eventProcessor.thread_group[ET_NET]._count),
+                                                   std::chrono::milliseconds{watchdog_timeout_ms});
```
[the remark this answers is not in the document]
moonchen: The sentinel values are chosen this way to tolerate delays in the startup of ET_NET threads.  See this code in watchdog.cc:

```
if (last_sleep == std::chrono::steady_clock::time_point::min()) {
        // initial value sentinel - event loop hasn't started
        continue;
      }
```

zwoop (approved):

PR #12528, opened 2025-09-23
HRW: Support \r, \n and \t as ctrl chars in values

zwoop:
This allows to have \r\n when calling e.g. set-body. I'm considering this a bug in header_rewrite.


bneradt (approved):
Looks reasonable.

cmcfarlen:
Added to milestone 10.1.1 via #12547

PR #12529, opened 2025-09-24
Fix connection leak on client abort

masaori335:
We faced two issues with https://github.com/apache/trafficserver/pull/12502 deployment.

A). client side connection is not closed when the client aborts during cache read
B). HTTP/2 connection is closed if a stream is reset even if other streams are alive

Approach for A). restore code prior to #12502 and narrow down condition to keep HttpSM running
Approach for B). call `ProxyTransaction::do_io_shutdown()` instead of `NetVConnection::do_io_shutdown()`

### DIFF
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -869,15 +869,24 @@ HttpSM::state_watch_for_client_abort(int event, void *data)
    * client.
    */
   case VC_EVENT_EOS: {
-    // We got an early EOS. To trigger background fill, do NOT kill HttpSM.
+    // We got an early EOS.
     if (!terminate_sm) { // Not done already
       NetVConnection *netvc = _ua.get_txn()->get_netvc();
-      if (netvc) {
-        if (_ua.get_txn()->allow_half_open()) {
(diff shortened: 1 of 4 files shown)


src/proxy/http/HttpSM.cc
```diff
@@ -869,15 +869,24 @@ HttpSM::state_watch_for_client_abort(int event, void *data)
...
-      if (netvc) {
-        if (_ua.get_txn()->allow_half_open()) {
+      if (_ua.get_txn()->allow_half_open() || tunnel.has_consumer_besides_client()) {
+        if (netvc) {
           netvc->do_io_shutdown(IO_SHUTDOWN_READ);
-        } else {
-          netvc->do_io_shutdown(IO_SHUTDOWN_READWRITE);
         }
+      } else if (t_state.txn_conf->cache_http &&
+                 (server_entry != nullptr && server_entry->vc_read_handler == &HttpSM::state_read_server_response_header)) {
```
masaori335, line 880: If we compare with 244d82a (the parent commit of #12502), adding this else if block is the main change.

cmcfarlen:

PR #12530, opened 2025-09-25
header_rewrite: Fix a bug that parameters on remap rules are overwritten by subsequent rules

maskit:
If the parameters added by https://github.com/apache/trafficserver/pull/12344 are used on remap rules, the values are overwritten by subsequent remap rules.


maskit:
The intpr package stopped the crashes we saw on usgrx1-edge-bx-011.

PR #12531, opened 2025-09-30
Fixes sni.yaml-configured initial window sizing

randall:
Without this fix, the sni.yaml configured http2_initial_window size is never sent to the client in the non-dynamic window scaling mode.

### DIFF
--- a/src/proxy/http2/Http2ConnectionState.cc
+++ b/src/proxy/http2/Http2ConnectionState.cc
@@ -1369,10 +1369,7 @@ Http2ConnectionState::send_connection_preface()
   configured_settings.set(HTTP2_SETTINGS_MAX_CONCURRENT_STREAMS, _adjust_concurrent_stream());
 
   uint32_t const configured_initial_window_size = this->_get_configured_receive_session_window_size();
-  if (this->_has_dynamic_stream_window()) {
-    // Since this is the beginning of the connection and there are no streams
-    // yet, we can just set the stream window size to fill the entire session
-    // window size.
+  if (configured_initial_window_size > HTTP2_INITIAL_WINDOW_SIZE) {
     configured_settings.set(HTTP2_SETTINGS_INITIAL_WINDOW_SIZE, configured_initial_window_size);
   }
 


src/proxy/http2/Http2ConnectionState.cc
```diff
@@ -1368,14 +1368,18 @@ Http2ConnectionState::send_connection_preface()
 
   configured_settings.set(HTTP2_SETTINGS_MAX_CONCURRENT_STREAMS, _adjust_concurrent_stream());
 
-  uint32_t const configured_initial_window_size = this->_get_configured_receive_session_window_size();
+  uint32_t configured_initial_window_size;
   if (this->_has_dynamic_stream_window()) {
```
maskit: Can't we just remove this condition, `_has_dynamic_stream_window`, instead? It looks like `_get_configured_receive_session_window_size` is supposed to return an appropriate value based on the policy. I wonder why we wanted to modify `configured_settings` only if the policy is dynamic.

https://github.com/apache/trafficserver/blob/534475f5b4f6278bc8981ff7965ee382b0ccf644/src/proxy/http2/Http2ConnectionState.cc#L2894-L2921

maskit: @bneradt Do you remember why?

bneradt: @randall : your patch looks fine in its intention.

I also think that @maskit's tweak on top of your patch is a correct and good improvement. It's been 3 years, but I believe I was simply trying to avoid unecessary use of `HTTP2_SETTINGS_INITIAL_WINDOW_SIZE` and for some reason thought it was only needed for dynamic streams. But clearly that's not the case. Let's just send the `HTTP2_SETTINGS_INITIAL_WINDOW_SIZE` regardless of dynamic or not.

However, we maybe should not send the setting if the configured value is already the default `HTTP2_INITIAL_WINDOW_SIZE`.

randall: Fair enough, I'll update the PR to do all of this.

src/proxy/http2/Http2ConnectionState.cc
```diff
@@ -1368,11 +1368,12 @@ Http2ConnectionState::send_connection_preface()
 
   configured_settings.set(HTTP2_SETTINGS_MAX_CONCURRENT_STREAMS, _adjust_concurrent_stream());
 
-  uint32_t const configured_initial_window_size = this->_get_configured_receive_session_window_size();
-  if (this->_has_dynamic_stream_window()) {
-    // Since this is the beginning of the connection and there are no streams
-    // yet, we can just set the stream window size to fill the entire session
-    // window size.
+  uint32_t const configured_initial_window_size = this->_get_configured_initial_window_size();
```
maskit: I think we should use `_get_configured_receive_session_window_size`.

randall: that makes more sense now. I've updated the PR

maskit (commented):
Looks good to me, but no approval from me because of the different org policy.

maskit:
I added this to 10.1.x project because this is a regression.

src/proxy/http2/Http2ConnectionState.cc
```diff
@@ -1369,10 +1369,11 @@ Http2ConnectionState::send_connection_preface()
...
 
   uint32_t const configured_initial_window_size = this->_get_configured_receive_session_window_size();
-  if (this->_has_dynamic_stream_window()) {
-    // Since this is the beginning of the connection and there are no streams
-    // yet, we can just set the stream window size to fill the entire session
-    // window size.
+
+  // Since this is the beginning of the connection and there are no streams
+  // yet, we can just set the stream window size to fill the entire session
+  // window size.
```
bneradt: @randall , you've been very patient with us. Thank you for fixing this. I have one more nit: for context, this comment really only applied to the dynamic stream policy in which the stream size shifts dynamically as a function of the session window and number of streams. The idea was that since there were currently no streams at the beginning of the connection, we could make the next stream use the entire session (connection) window if it wanted. This comment doesn't really apply therefore to the static cases in which we configure the stream and session windows separately.

tl;dr: Let's just remove the comment.

The code change looks good to me.

bneradt (changes requested):

src/proxy/http2/Http2ConnectionState.cc
randall: it's ok! I'll fix up. better to do it now.

bneradt (approved):
Thank you @randall !

cmcfarlen:

PR #12532, opened 2025-09-30
Check valid ts in curl command

serrislew:
Add ts argument to new autest: `x_probe_full_json` and add additional conditional in curl command


serrislew:
[approve ci autest]

bneradt (approved):

PR #12537, opened 2025-10-02
Fix traffic_ctl server AuTest

masaori335:
Disable `proxy.config.exec_thread.autoconfig.enabled` to pin exec_thread number to 4.


cmcfarlen:

PR #12539, opened 2025-10-02
TLS: Add per-curve handshake time metrics

moonchen:
Track handshake duration for each TLS curve/group to enable performance analysis across different elliptic curves. Extends existing per-group connection counters with corresponding handshake_time metrics.


bneradt (approved):
Looks great to me.

PR #12540, opened 2025-10-03
remap_acl speedup via config reload

bneradt:
This changes the remap_acl.test.py test to use config reloads instead of
spinning up new ATS insteads for every TestRun. Now that we've reduced
our reload scanning interval via #12541, these reloads in test are
pretty quick.

Before:
```
./autest.sh --sandbox /tmp/sb --clean=none -f remap_acl  171.62s user 57.13s system 52% cpu 7:12.57 total
```

After:
```
./autest.sh --sandbox /tmp/sb --clean=none -f remap_acl  14.39s user 10.78s system 19% cpu 2:09.23 total
```

So the time to run this test goes from about 7 minutes to about 2 minutes.

### DIFF
--- a/tests/gold_tests/remap/remap_acl.test.py
+++ b/tests/gold_tests/remap/remap_acl.test.py
@@ -24,15 +24,36 @@ from yaml import load, dump
 from yaml import CLoader as Loader
 from typing import List, Tuple
 
+from ports import get_port
+
 Test.Summary = '''
 Verify remap.config acl behavior.
 '''
 
 
+def update_config_file(path1: str, content1: str, path2: str, content2: str) -> None:
+    """Update two config files.
+
+    This is used for some of the updates to the config files between test runs.
+
+    :param path1: The path to the first config file.
+    :param content1: The content to write to the first config file.
+    :param path2: The path to the second config file.
+    :param content2: The content to write to the second config file.
+    """
+    with open(path1, 'w') as f:
+        f.write(content1 + '\n')
+    with open(path2, 'w') as f:
+        f.write(content2 + '\n')
+
+
 class Test_remap_acl:
     """Configure a test to verify remap.config acl behavior."""
 
-    _ts_counter: int = 0
+    _ts: 'TestProcess' = None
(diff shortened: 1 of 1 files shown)


bneradt:
Adding @brbzull0 for the config update logic. I ran into bugs when I updated the `proxy.config.config_update_interval_ms` - the wait_for_cache and some other configs didn't get set appropriately. This fixes the update logic so that both are set now. I kept this as a separate commit.

bneradt:
> Adding @brbzull0 for the config update logic. I ran into bugs when I updated the `proxy.config.config_update_interval_ms` - the wait_for_cache and some other configs didn't get set appropriately. This fixes the update logic so that both are set now. I kept this as a separate commit.

This is getting more complicated because some tests were broken due to the longer reload interval. I will hold off on this remap_reload update until I fix the other issues in a separate PR:
#12541

In the meantime, I'll make this as a draft and rebase once that lands.

tests/gold_tests/remap/remap_acl.test.py
```diff
@@ -104,9 +124,64 @@ def _configure_traffic_server(self, tr: 'TestRun', proxy_protocol: bool) -> None
...
+
+            p.Env = ts.Env
+            tr.StillRunningAfter = ts
+
+            remap_cfg_path = os.path.join(ts.Variables.CONFIGDIR, 'remap.config')
+            ip_allow_path = os.path.join(ts.Variables.CONFIGDIR, 'ip_allow.yaml')
+            p.Setup.Lambda(
+                lambda: update_config_file(
+                    remap_cfg_path, '\n'.join(remap_config_lines), ip_allow_path, '\n'.join(self._ip_allow_lines)))
+
```
bneradt, line 147: This is the heart of this patch:
* For the first test run, start ATS via MakeATSProcess (see the `else:` below this).
* For subsequent test runs, update the config and do a config reload instead of starting a new test process. Doing a config reload is faster than starting a brand new ATS process for every test run.

masaori335 (approved):
Cool! I like this reloading idea!

PR #12541, opened 2025-10-03
autest: tighten config_update_interval_ms

bneradt:
This sets the default proxy.config.config_update_interval_ms to 20ms so that config reloads are done more quickly. This required fixing a couple things:

* the config update logic didn't work with this new config setting when the other config updates were run later in the trafficserver.test.ext file.
* the ssl_key_dialog which was accidentally working because the final TestRun was running before the config reload completed. This fixes the test to systematically wait for the config reload to fail and has the curl command use the correct passphrases now.

PR #12542, opened 2025-10-06
autest: tighten config_update_interval_ms

bneradt:
Same as #12541, but applied to all tests

This sets the default proxy.config.config_update_interval_ms to 10ms so that config reloads are done more quickly. This required fixing a couple things:

* the config update logic didn't work with this new config setting when the other config updates were run later in the trafficserver.test.ext file.
* the ssl_key_dialog which was accidentally working because the final TestRun was running before the config reload completed. This fixes the test to systematically wait for the config reload to fail and has the curl command use the correct passphrases now.


bneradt:
[approve ci autest 0]

bneradt:
[approve ci autest]

bneradt:
Closing this because this was used to experiment with CI. Continuing on this work with #12541

PR #12546, opened 2025-10-06
Add flags option to Regex::exec

bryancall:
Flags for Regex::exec are needed to convert plugins that use the PCRE_NOTEMPTY feature.


bryancall:
[approve ci autest 0]

PR #12550, opened 2025-10-08
ControlMatcher: add filename to warnings

bneradt:
This adds the filename to the warning strings associated with ControlMatcher so the user knows where to look for issues.


bneradt:
Replaced with the more comprehensive: #12550

PR #12551, opened 2025-10-08
ControlMatcher: error versus NOMATCH checking

bneradt:
This updates the ControlMatcher logic to distinguish between error conditions versus NOMATCH conditions. Without this patch, ControlMatcher was emitting incorrect regex matching warnings for NOMATCH conditions.

### DIFF
--- a/src/proxy/ControlMatcher.cc
+++ b/src/proxy/ControlMatcher.cc
@@ -476,13 +476,15 @@ RegexMatcher<Data, MatchResult>::Match(RequestData *rdata, MatchResult *result) 
   // The function unescapifyStr() is already called in
   // HttpRequestData::get_string(); therefore, no need to call again here.
   for (int i = 0; i < num_el; i++) {
-    if (regex_array[i].exec(url_str) == true) {
+    RegexMatches matches;
+    int          r = regex_array[i].exec(url_str, matches);
+    if (r >= 0) {
       Dbg(dbg_ctl_matcher, "%s Matched %s with regex at line %d", matcher_name, url_str, data_array[i].line_num);
       data_array[i].UpdateMatch(result, rdata);
-    } else {
+    } else if (r != PCRE2_ERROR_NOMATCH) {
       // An error has occurred
-      Warning("Error matching regex at line %d.", data_array[i].line_num);
-    } // else it's -1 which means no match was found.
+      Warning("Error matching regex for url: %s:%d (%d)", file_name ? file_name : "unknown", data_array[i].line_num, r);
+    } // else: PCRE2_ERROR_NOMATCH
   }
   ats_free(url_str);
 }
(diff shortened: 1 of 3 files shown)


src/proxy/ControlMatcher.cc
```diff
@@ -476,13 +476,15 @@ RegexMatcher<Data, MatchResult>::Match(RequestData *rdata, MatchResult *result)
...
   // HttpRequestData::get_string(); therefore, no need to call again here.
   for (int i = 0; i < num_el; i++) {
-    if (regex_array[i].exec(url_str) == true) {
+    RegexMatches matches;
+    int          r = regex_array[i].exec(url_str, matches);
+    if (r > 0) {
       Dbg(dbg_ctl_matcher, "%s Matched %s with regex at line %d", matcher_name, url_str, data_array[i].line_num);
       data_array[i].UpdateMatch(result, rdata);
-    } else {
+    } else if (r < 0 && r != PCRE2_ERROR_NOMATCH) {
```
bryancall: You can do `} else if (r < -1) {`.   That was in the code before the Regex convertion.  I should have listen to my gut when converting this over to use the `bool` interface.  I thought there might be an issue, but I didn't want to create a matches object.

src/proxy/ControlMatcher.cc
```diff
@@ -476,13 +476,15 @@ RegexMatcher<Data, MatchResult>::Match(RequestData *rdata, MatchResult *result)
   // The function unescapifyStr() is already called in
   // HttpRequestData::get_string(); therefore, no need to call again here.
   for (int i = 0; i < num_el; i++) {
-    if (regex_array[i].exec(url_str) == true) {
+    RegexMatches matches;
+    int          r = regex_array[i].exec(url_str, matches);
+    if (r > 0) {
```
bryancall: This was `if (r > -1) {` before

src/proxy/ControlMatcher.cc
```diff
@@ -521,14 +523,16 @@ HostRegexMatcher<Data, MatchResult>::Match(RequestData *rdata, MatchResult *resu
     url_str = "";
   }
   for (int i = 0; i < num_el; i++) {
-    if (this->regex_array[i].exec(url_str) == true) {
+    RegexMatches matches;
+    int          r = this->regex_array[i].exec(url_str, matches);
+    if (r > 0) {
```
bryancall: This logic was a little funky before it was:
```
    r = pcre_exec(this->re_array[i], nullptr, url_str, strlen(url_str), 0, 0, nullptr, 0);
    if (r != -1) {
      Dbg(dbg_ctl_matcher, "%s Matched %s with regex at line %d", const_cast<char *>(this->matcher_name), url_str,
          this->data_array[i].line_num);
      this->data_array[i].UpdateMatch(result, rdata);
    } else {
      // An error has occurred
      Warning("error matching regex at line %d", this->data_array[i].line_num);
    }
```

src/proxy/ControlMatcher.cc
bneradt: Sounds good.

src/proxy/ControlMatcher.cc
bryancall: This should be `r >= 0 or r > -1`

PR #12552, opened 2025-10-08
libswoc 1.5.15: version update

bneradt:
This updates our lib/swoc version to 1.5.15 to match the apache/trafficserver-libswoc version with which it is identical. This is a version metadata only update to document the parity. There is no functional change with this PR.


masaori335:
What we should do for 10.1.x branch?

bneradt:
> What we should do for 10.1.x branch?

I say we leave it for now. The #12198 needs IPRangeSet::erase from the update, but we don't plan to backport that to 10.1.0.

PR #12554, opened 2025-10-09
realip: Add null check for netvc

maskit:
Apparently, client netvc is not always available on `READ_REQUEST_HDR_HOOK`.


bryancall:
I would like to see the implementation being safer by doing null check on the input arguments and return TS_ERROR.

maskit:
> Looks fine, approving as is. But would a debug log for the else case be helpful potentially?

There's a debug log on the caller side, and that works for all address sources.

> I would like to see the implementation being safer by doing null check on the input arguments and return TS_ERROR.

I do not think that is a good idea. If we took that way, null check would be done again and again by all functions that receive vconn on every call. So redundant.

PR #12557, opened 2025-10-10
HRW: Fixes a regression bug due to #12536

zwoop:
I will adjust the autest for this once this land, and #12371 lands.


cmcfarlen:
This fixes an issue where hrw rules that used the implicit hook would not function correctly.  We verified this fix in production.

PR #12562, opened 2025-10-14
Hrw: Supports nested If

zwoop:
This refactors the parsing / evaluation, to put most of the exec / evaluation logic into
a new OperatorIf. This is not a real operator, that configurations would use, but allows
for the conditionals to be nested. The syntax in HRW is not awesome, but best we can
do within the constraints of backwards compatibility.

As such, this PR also includes changes and additions to hrw4u, where nested if's are
now supported and added to the existing if() / else semantics. E.g.

```
SEND_RESPONSE {
    if inbound.status == 200 {
        inbound.resp.X-When-200-Before = "Yes";
        if inbound.req.X-Foo == "foo" {
            inbound.resp.X-Foo = "Yes";
            if inbound.req.X-Bar == "bar" with NOCASE {
                inbound.resp.X-Foo-And-Bar = "Yes";
            } elif inbound.req.X-Fie == "fie" with NOCASE {
                inbound.resp.X-Foo-And-Fie = "Yes";
            }
        } elif inbound.req.X-Foo == "maybe" {
.
.
.

```


zwoop:
Ignored the nitpick comments, the current code is inline with other uses within the code base. Consistency is better.

cmcfarlen (approved):
LGTM.  Idk if I understand why If has to be an Operator.  I guess sometimes it is called via the virtual exec function?  The nested ones?

zwoop:
> LGTM. Idk if I understand why If has to be an Operator. I guess sometimes it is called via the virtual exec function? The nested ones?

Does it have to be? No. But it simplified evaluations significantly, since the the operator list can now contain other OperatorIf's, giving us the nesting. This is similar to how we made groups ("nested conditions"), where every condition is a GROUP condition, which then contains other conditions and operators. Magic.

PR #12564, opened 2025-10-15
url_sig: use Regex instead of pcre

bneradt:
This removes the use of the deprecated pcre library and replaces it with the ATS core tsutil Regex class.


bryancall (approved):

plugins/experimental/url_sig/url_sig.cc
```diff
@@ -51,8 +47,7 @@ struct config {
   TSHttpStatus err_status;
   char        *err_url;
   char         keys[MAX_KEY_NUM][MAX_KEY_LEN];
-  pcre        *regex;
-  pcre_extra  *regex_extra;
+  Regex       *excl_regex;
```
cmcfarlen, line 50: Can this be a value vs pointer? or `std::unique_ptr`?

bneradt: Yes, I thought that too. I started making it a unique_ptr, then abandoned it for this commit because this file is all clearly c-style. Stuff is malloc'd and memset to zero and the like, so not constructor friendly. I'm up for changing that, using a constructor, but was thinking if we do that it would be good to do that as a separate PR.

cmcfarlen (approved):

plugins/experimental/url_sig/url_sig.cc
bneradt: https://github.com/apache/trafficserver/pull/12569

PR #12566, opened 2025-10-15
HRW: Better txn slot management

zwoop:
This also fixes a real bug, where the txn slot for the state variables could overlap / conflict with the control mechanism and its txn slot. They need to have different names.


zwoop:
Some or all of this needs to go to 10.1.x. Particularly the name of the private txn slot is critically, that will definitely break any configuration using both slots.

maskit (approved):

cmcfarlen:
Added to milestone 10.1.1 via #12589

PR #12567, opened 2025-10-15
Configurable Hash Algorithm and Parameters for Consistent Hash Parent Selection

PSUdaemon:
## Overview

  Makes consistent hash parent selection fully configurable by adding support for:
  - Selectable hash algorithms (SipHash-2-4 and SipHash-1-3)
  - Configurable hash seeds for cryptographic key customization
  - Configurable replica count for hash ring distribution tuning

  Configuration is available at three levels: global defaults (records.yaml), per-rule (parent.config), and per-strategy (strategies.yaml).

  ## Motivation

  The current implementation hard-codes SipHash-2-4 with zero seeds and 1024 replicas. This change provides flexibility for operators to:
  1. Choose faster hash algorithms based on performance requirements
  2. Customize hash seeds for security or cache isolation
  3. Tune replica counts for optimal distribution vs memory usage
  4. Configure these parameters globally or per-rule/strategy

  ## Configuration

  ### Parameters

  | Parameter | Type | Default | Description |
  |-----------|------|---------|-------------|
  | `hash_algorithm` | string | siphash24 | Hash algorithm: siphash24 or siphash13 |
  | `hash_seed0` | uint64_t | 0 | First 64 bits of hash seed (decimal integer) |
  | `hash_seed1` | uint64_t | 0 | Second 64 bits of hash seed (decimal integer) |
  | `hash_replicas` | int | 1024 | Virtual nodes per host on hash ring |

  ### Global Defaults (records.yaml)

  ```yaml
  http:
    parent_proxy:
      consistent_hash_algorithm: siphash24
      consistent_hash_seed0: 0
      consistent_hash_seed1: 0
      consistent_hash_replicas: 1024
```

  Per-Rule Configuration (parent.config)

```
  dest_domain=example.com parent=p1:80,p2:80 round_robin=consistent_hash hash_algorithm=siphash13 hash_seed0=12345678901234567 hash_replicas=2048
```

  Per-Strategy Configuration (strategies.yaml)

  ```yaml
  strategies:
    - strategy: 'my-strategy'
      policy: consistent_hash
      hash_algorithm: siphash13
      hash_seed0: 12345678901234567
      hash_seed1: 9876543210987654321
      hash_replicas: 2048
      groups:
        - *primary_group
```

  Hash Algorithms

  - SipHash-2-4 (siphash24, default): Cryptographically strong, DoS-resistant
  - SipHash-1-3 (siphash13): ~50% faster, still DoS-resistant

  Hash Seeds:
  - SipHash uses both hash_seed0 and hash_seed1 as a 128-bit key
  - Future 64-bit algorithms (XXH3, RapidHash) will use only hash_seed0

  Migration Considerations

  Warning: Changing hash_algorithm, hash_seed0, or hash_seed1 will redistribute requests across parent proxies, causing:
  - Cache churn
  - Increased origin load during transition
  - Temporary cache hit rate reduction

  Plan changes during low-traffic periods and monitor origin load.

### DIFF
--- a/src/proxy/http/remap/NextHopConsistentHash.cc
+++ b/src/proxy/http/remap/NextHopConsistentHash.cc
@@ -27,6 +27,7 @@
 #include "iocore/utils/Machine.h"
 #include "tsutil/YamlCfg.h"
 #include "proxy/http/remap/NextHopConsistentHash.h"
+#include "proxy/ParentSelection.h"
 
 // hash_key strings.
 constexpr std::string_view hash_key_url           = "url";
@@ -57,16 +58,10 @@ std::shared_ptr<HostRecord>
 NextHopConsistentHash::chashLookup(const std::shared_ptr<ATSConsistentHash> &ring, uint32_t cur_ring, ParentResult &result,
                                    HttpRequestData &request_info, bool *wrapped, uint64_t sm_id)
 {
-  uint64_t hash_key = 0;
-  // Create hash instance based on configured algorithm and seeds
-  std::unique_ptr<ATSHash64> hash;
-  if (hash_algorithm == "siphash13") {
-    hash = std::make_unique<ATSHash64Sip13>(hash_seed0, hash_seed1);
-  } else {
-    hash = std::make_unique<ATSHash64Sip24>(hash_seed0, hash_seed1);
-  }
-  HostRecord            *host_rec = nullptr;
-  ATSConsistentHashIter *iter     = &result.chashIter[cur_ring];
+  uint64_t                   hash_key = 0;
+  std::unique_ptr<ATSHash64> hash     = createHashInstance(parseHashAlgorithm(hash_algorithm), hash_seed0, hash_seed1);
+  HostRecord                *host_rec = nullptr;
+  ATSConsistentHashIter     *iter     = &result.chashIter[cur_ring];
 
   if (result.chash_init[cur_ring] == false) {
     hash_key                    = getHashKey(sm_id, request_info, hash.get());
@@ -196,13 +191,7 @@ NextHopConsistentHash::NextHopConsistentHash(const std::string_view name, const 
                                 "', this strategy will be ignored.");
   }
 
-  // Create the hash instance based on configured algorithm and seeds
-  if (hash_algorithm == "siphash13") {
-    hash = std::make_unique<ATSHash64Sip13>(hash_seed0, hash_seed1);
-  } else {
-    // Default to siphash24
-    hash = std::make_unique<ATSHash64Sip24>(hash_seed0, hash_seed1);
-  }
+  hash = createHashInstance(parseHashAlgorithm(hash_algorithm), hash_seed0, hash_seed1);
 
   // load up the hash rings.
   for (uint32_t i = 0; i < groups; i++) {
--- a/include/proxy/ParentConsistentHash.h
+++ b/include/proxy/ParentConsistentHash.h
@@ -48,8 +48,6 @@ class ParentConsistentHash : public ParentSelectionStrategy
   uint64_t                           hash_seed0;
   uint64_t                           hash_seed1;
 
-  std::unique_ptr<ATSHash64> createHashInstance(ParentHashAlgorithm algo, uint64_t seed0, uint64_t seed1);
-
 public:
   static const int PRIMARY   = 0;
   static const int SECONDARY = 1;
--- a/include/proxy/ParentSelection.h
+++ b/include/proxy/ParentSelection.h
@@ -35,11 +35,13 @@
 #include "proxy/ControlMatcher.h"
 #include "records/RecProcess.h"
 #include "tscore/ConsistentHash.h"
+#include "tscore/Hash.h"
 #include "tscore/Tokenizer.h"
 #include "tscore/ink_apidefs.h"
 #include "proxy/HostStatus.h"
 
 #include <algorithm>
+#include <memory>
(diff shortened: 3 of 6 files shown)


jrushford:
@PSUdaemon this is relevant to strategies also as it uses the SipHash-2-4 for its consistent hash used for parent selection.  You might consider a PR for strategies.yaml.  I believe that strategies will deprecate parent selection in the future.

Regards
Jrushford

PSUdaemon:
> @PSUdaemon this is relevant to strategies also as it uses the SipHash-2-4 for its consistent hash used for parent selection. You might consider a PR for strategies.yaml. I believe that strategies will deprecate parent selection in the future. Regards Jrushford

@jrushford, I looked into this and seems like a logical next PR. If the community likes this one and merges it, I'll do that next.

maskit:
It seems like the description was generated by AI, which is fine and sorry if it wasn't. Was the code generated by AI? If so, can you make sure that the code is compliant with the guideline from ASF?
https://www.apache.org/legal/generative-tooling.html

PSUdaemon:
> It seems like the description was generated by AI, which is fine and sorry if it wasn't. Was the code generated by AI? If so, can you make sure that the code is compliant with the guideline from ASF? https://www.apache.org/legal/generative-tooling.html

I'm not sure if I totally follow this, but it appears to be more about copyright. I removed WyHash to remove questions about the source of code. We can add that or other hash functions later and debate that issue there. This commit also now has more infrastructure that Craig and @jrushford requested.

I can attest that the other code is safe to include. The SipHash implementation is just a refactor of the existing version and the rest is adding configs and uses the local code style.

src/proxy/http/remap/NextHopConsistentHash.cc
```diff
@@ -57,17 +57,23 @@ std::shared_ptr<HostRecord>
 NextHopConsistentHash::chashLookup(const std::shared_ptr<ATSConsistentHash> &ring, uint32_t cur_ring, ParentResult &result,
                                    HttpRequestData &request_info, bool *wrapped, uint64_t sm_id)
 {
-  uint64_t               hash_key = 0;
-  ATSHash64Sip24         hash;
+  uint64_t hash_key = 0;
+  // Create hash instance based on configured algorithm and seeds
+  std::unique_ptr<ATSHash64> hash;
+  if (hash_algorithm == "siphash13") {
```
maskit: Looks like it's almost a duplicate of `ParentConsistentHash::createHashInstance`. Can we have just one place to do this?

maskit:
> I'm not sure if I totally follow this, but it appears to be more about copyright. I removed WyHash to remove questions about the source of code. We can add that or other hash functions later and debate that issue there. This commit also now has more infrastructure that Craig and @jrushford requested.

Thank you for the update.

This is just a suggestion, but it would be more interesting if we could add the support for WyHash or other algorithms without changing ATS core. Apparently, WyHash has a successor (Rapidhash). Adding the support for WyHash might only increase the size of code we have to maintain. I'd suggest introducing `TSHashAlgoRegister(const char *name, HashAlgoMethod *method)` or some such so that users can add preferred hash algorithms by using/writing TS plugins. It'd also allow us to write a "hash" algorithm that is more like mapping and that could be useful for testing or experiments.

PSUdaemon:
> > I'm not sure if I totally follow this, but it appears to be more about copyright. I removed WyHash to remove questions about the source of code. We can add that or other hash functions later and debate that issue there. This commit also now has more infrastructure that Craig and @jrushford requested.
> 
> Thank you for the update.
> 
> This is just a suggestion, but it would be more interesting if we could add the support for WyHash or other algorithms without changing ATS core. Apparently, WyHash has a successor (Rapidhash). Adding the support for WyHash might only increase the size of code we have to maintain. I'd suggest introducing `TSHashAlgoRegister(const char *name, HashAlgoMethod *method)` or some such so that users can add preferred hash algorithms by using/writing TS plugins. It'd also allow us to write a "hash" algorithm that is more like mapping and that could be useful for testing or experiments.

I'll definitely consider this.

PSUdaemon:
[approve ci autest 1]

zwoop:
Seeing Rushford's comment, makes you wonder why we have two implementations of this underlying consistent hashing? Shouldn't the old parent.config hashing, and the new strategies.yaml hashing use the same core hash functions / code ?

PSUdaemon:
> Seeing Rushford's comment, makes you wonder why we have two implementations of this underlying consistent hashing? Shouldn't the old parent.config hashing, and the new strategies.yaml hashing use the same core hash functions / code ?

The hash functions and consistent hash ring implementation are the same between them. The differences are `ParentConsistentHash` and `NextHopConsistentHash`. `NextHopConsistentHash` is a superset of `ParentConsistentHash`. It's mostly about the flexibility of the classes. For example, `ParentConsistentHash` has a primary and a secondary ring, but `NextHopConsistentHash` has a `Vec` of rings. So you have have an arbitrary amount. `NextHopConsistentHash` has more options for hash keys as well. They could be unified, however I think what @jrushford was getting at was that we would deprecate `parent.config` in favor of `strategies.yaml`.

maskit (commented):
My suggestion was for the future work that would add new hash algorithms, and I'm not waiting for the change on this PR. No approval from me because of the review policy (same org) but looks good to me.

traeak:
The branch should not have been made directly in apache/trafficserver, it must come from your own personal clone of the trafficserver repo.  Just make sure you do that next time.

PR #12568, opened 2025-10-16
Use ts::Metrics for socks metrics

cmcfarlen:


cmcfarlen:
[approve ci autest 1]

PR #12570, opened 2025-10-16
Convert slice to use Regex

cmcfarlen:


cmcfarlen:
[approve ci autest 1]

PR #12572, opened 2025-10-16
access_control: use Regex instead of pcre

bneradt:
Updates the access_control plugin to use tsutil Regex instead of pcre.

### DIFF
--- a/plugins/experimental/access_control/pattern.cc
+++ b/plugins/experimental/access_control/pattern.cc
@@ -24,6 +24,8 @@
 
 #include "pattern.h"
 
+#include <tsutil/Regex.h>
+
 static void
 replaceString(String &str, const String &from, const String &to)
 {
@@ -40,6 +42,8 @@ replaceString(String &str, const String &from, const String &to)
 
 Pattern::Pattern() : _pattern(""), _replacement("") {}
 
+Pattern::~Pattern() = default;
+
 /**
  * @brief Initializes PCRE pattern by providing the subject and replacement strings.
  * @param pattern PCRE pattern, a string containing PCRE patterns, capturing groups.
@@ -49,8 +53,6 @@ Pattern::Pattern() : _pattern(""), _replacement("") {}
 bool
 Pattern::init(const String &pattern, const String &replacement, bool replace)
 {
-  pcreFree();
-
   _pattern.assign(pattern);
   _replacement.assign(replacement);
   _replace = replace;
@@ -59,7 +61,7 @@ Pattern::init(const String &pattern, const String &replacement, bool replace)
 
   if (!compile()) {
     AccessControlDebug("failed to initialize pattern:'%s', replacement:'%s'", pattern.c_str(), replacement.c_str());
-    pcreFree();
+    _re.reset();
     return false;
   }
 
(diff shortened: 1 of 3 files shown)


plugins/experimental/access_control/pattern.cc
```diff
@@ -214,22 +195,13 @@ Pattern::process(const String &subject, StringVector &result)
...
-  if (matchCount < 0) {
-    if (matchCount != PCRE_ERROR_NOMATCH) {
-      AccessControlError("matching error %d", matchCount);
-    }
+  if (!_re || _re->empty()) {
     return false;
   }
 
-  return true;
+  return _re->exec(subject);
```
bryancall: NOTEMPTY was removed.   Looks like it needs to be added to the flags.

plugins/experimental/access_control/pattern.cc
```diff
@@ -214,22 +195,13 @@ Pattern::process(const String &subject, StringVector &result)
...
-  int matchCount;
   AccessControlDebug("matching '%s' to '%s'", _pattern.c_str(), subject.c_str());
 
-  if (!_re) {
-    return false;
-  }
-
-  matchCount = pcre_exec(_re, _extra, subject.c_str(), subject.length(), 0, PCRE_NOTEMPTY, nullptr, 0);
-  if (matchCount < 0) {
-    if (matchCount != PCRE_ERROR_NOMATCH) {
```
bryancall, line 226: This check and error was also removed.  Looks like it should be added back in.

plugins/experimental/access_control/pattern.cc
```diff
@@ -280,21 +245,16 @@ Pattern::capture(const String &subject, StringVector &result)
...
 
-  if (!_re || !_replace) {
+  if (!_re || _re->empty() || !_replace) {
     AccessControlError("regular expression not initialized or not configured to replace");
     return false;
   }
 
-  matchCount = pcre_exec(_re, nullptr, subject.c_str(), subject.length(), 0, PCRE_NOTEMPTY, ovector, OVECOUNT);
+  RegexMatches matches(TOKENCOUNT);
+  int          matchCount = _re->exec(subject, matches);
```
bryancall: NOTEMPTY was not passed here

bryancall (changes requested):

plugins/experimental/access_control/pattern.cc
bneradt: Isn't NOTEMPTY the default?

```cpp
enum REFlags {
  RE_CASE_INSENSITIVE = 0x00000008u, ///< Ignore case (default: case sensitive).
  RE_UNANCHORED       = 0x00000400u, ///< Unanchored (DFA defaults to anchored).
  RE_ANCHORED         = 0x80000000u, ///< Anchored (Regex defaults to unanchored).
  RE_NOTEMPTY         = 0x00000004u  ///< Not empty (default: may match empty string).
};
```

bneradt: Oh, I see. That "default" doesn't mean that NOTEMPTY is default, but is explaining that without the NOTEMPTY flag, empty strings will be matched.

Thanks for catching this.

bryancall (approved):
✅ **APPROVED**

Excellent work! This PR successfully migrates the access_control plugin from raw PCRE to tsutil::Regex with proper attention to detail.

## Key Strengths:
1. ✅ **RE_NOTEMPTY flag preserved** in all 3 exec() calls (lines 205, 232, 268)
2. ✅ **Proper error handling** with new RE_ERROR_NOMATCH and RE_ERROR_NULL constants
3. ✅ **Memory safety** with std::unique_ptr and RAII
4. ✅ **Comprehensive testing** including new unit tests for error codes

## Verification:
- All RE_NOTEMPTY flags present and correct
- Error constants properly mapped and tested
- No linter errors
- Follows same pattern as PR #12576

## Risk: LOW 🟢
- Behavioral compatibility maintained
- Good test coverage
- Clean migration strategy

Ready for merge!

PR #12573, opened 2025-10-16
Update header_rewrite to use Regex

cmcfarlen:


cmcfarlen:
[approve ci autest 1]

plugins/header_rewrite/regex_helper.cc
```diff
@@ -17,39 +17,29 @@
...
-    return false;
-  }
-  regexExtra = pcre_study(regex, 0, &errorStudy);
-  if ((regexExtra == nullptr) && (errorStudy != nullptr)) {
-    return false;
-  }
-  if (pcre_fullinfo(regex, regexExtra, PCRE_INFO_CAPTURECOUNT, &regexCcount) != 0) {
+  bool ok = regex.compile(regexString, error, errorOffset, nocase ? PCRE_CASELESS : 0);
+
+  if (!ok) {
```
bneradt: Maybe just `if (!regex.compile...`

plugins/header_rewrite/regex_helper.cc
```diff
@@ -17,39 +17,29 @@
 */
 #include "regex_helper.h"
 #include "lulu.h"
+#include "ts/ts.h"
+#include "tsutil/Regex.h"
+#include <pcre.h>
```
bneradt: We shouldn't need `pcre.h` anymore, right?

zwoop:
[approve ci autest]

cmcfarlen:
> Looks good.
> 
> We have to remember to remove `PCRE::PCRE` from our `CMakeLists.txt` files. I almost missed that in my url_sig PR.

I figured we would take pass after these are done to remove the PCRE dep and cleanup cmake.  I can go ahead and remove it though.

plugins/header_rewrite/matcher.h
```diff
@@ -149,7 +150,6 @@ template <class T> class Matchers : public Matcher
         auto &re = std::get<regexHelper>(_data);
 
         if (!re.setRegexMatch(s, has_modifier(mods, CondModifiers::MOD_NOCASE))) {
-          TSError("[%s] Invalid regex: failed to precompile: %s", PLUGIN_NAME, s.c_str());
           Dbg(pi_dbg_ctl, "Invalid regex: failed to precompile: %s", s.c_str());
           throw std::runtime_error("Malformed regex");
         }
```
bneradt, line 155: I assume we're removing the error since this is matching not pre-compiling. Is the Dbg log message wrong too then?

bneradt (changes requested):

plugins/header_rewrite/matcher.h
cmcfarlen: I moved the error into the helper class so it could also print the regex error.  Currently you only get the regex that failed, now you'll get why and what character too.

cmcfarlen:
[approve ci autest 3]

plugins/header_rewrite/matcher.h
bneradt: Gotcha. Should we move the Dbg log there too though with the extra information.

PR #12576, opened 2025-10-16
prefetch: use Regex instead of pcre

moonchen:
This is part of the effort to remove pcre dependency.

### DIFF
--- a/include/tsutil/Regex.h
+++ b/include/tsutil/Regex.h
@@ -47,10 +47,6 @@ enum REErrors {
   RE_ERROR_NULL    = -51 ///< NULL code or subject was passed.
 };
 
-enum REErrorC {
-  RE_ERROR_NOMATCH = -1, ///< No match.
-};
-
 /// @brief Wrapper for PCRE2 match data.
 class RegexMatches
 {


plugins/prefetch/pattern.cc
```diff
@@ -129,33 +126,13 @@ Pattern::init(const String &config)
...
 }
 
 /**
  * @brief Destructor, frees PCRE related resources.
  */
-Pattern::~Pattern()
-{
-  pcreFree();
-}
+Pattern::~Pattern() {}
```
bneradt: Should probably just remove this and use `= default;` in the class declaration.

plugins/prefetch/pattern.h
```diff
@@ -23,22 +23,16 @@
...
  */
 class Pattern
 {
 public:
-  static const int TOKENCOUNT = 10;             /**< @brief Capturing groups $0..$9 */
-  static const int OVECOUNT   = TOKENCOUNT * 3; /**< @brief pcre_exec() array count, handle 10 capture groups */
+  static const int TOKENCOUNT = 10; /**< @brief Capturing groups $0..$9 */
 
   Pattern();
   virtual ~Pattern();
```
bneradt: Is inheritance actually used here? If not, let's remove the `virtual`.

plugins/prefetch/pattern.cc
moonchen: Removed

plugins/prefetch/pattern.h
moonchen: Removed the custom destructor.

bryancall (changes requested):
The migration looks good overall, but there's a functional regression: the original code used `PCRE_NOTEMPTY` flag in all three `pcre_exec()` calls, but the new code doesn't pass `RE_NOTEMPTY` to `Regex::exec()`.

**Issue**: `PCRE_NOTEMPTY` prevents matching empty strings, which is important for patterns like `.*` that could otherwise match at the beginning of an empty string.

**Required changes**:

1. **Line 185** in `match()`:
```cpp
return _regex.exec(subject, RE_NOTEMPTY);
```

2. **Line 203** in `capture()`:
```cpp
int matchCount = _regex.exec(subject, matches, RE_NOTEMPTY);
```

3. **Line 234** in `replace()`:
```cpp
int matchCount = _regex.exec(subject, matches, RE_NOTEMPTY);
```

This maintains the original PCRE behavior and prevents subtle matching bugs.

moonchen:
Added the flag back to `exec()`.

moonchen:
[approve ci autest]

moonchen:
[approve ci autest]

moonchen:
[approve ci autest]

plugins/prefetch/pattern.cc
```diff
@@ -204,22 +177,13 @@ Pattern::process(const String &subject, StringVector &result)
...
-  matchCount = pcre_exec(_re, _extra, subject.c_str(), subject.length(), 0, PCRE_NOTEMPTY, nullptr, 0);
-  if (matchCount < 0) {
-    if (matchCount != PCRE_ERROR_NOMATCH) {
-      PrefetchError("matching error %d", matchCount);
-    }
+  if (_regex.empty()) {
     return false;
   }
 
-  return true;
```
bneradt, line 222: There's a few of these PrefetchError's that are lost in the shuffle. I think we should be able to preserve these using a different Regex interface.

bneradt (commented):
Looks good. Just an observation on logging.

bneradt (approved):
Looks great to me. Thanks for helping us get off of pcre. :)

PR #12585, opened 2025-10-20
traffic_cache: use single regex in url_matcher port

moonchen:
This match is a single regex match.  DFA is meant for multiple regex matching.

### DIFF
--- a/src/traffic_cache_tool/CacheDefs.h
+++ b/src/traffic_cache_tool/CacheDefs.h
@@ -320,7 +320,7 @@ struct url_matcher {
       std::cout << "Check your regular expression" << std::endl;
       return;
     }
-    if (!port.compile(R"([0-9]+$)")) {
+    if (!port.compile(R"(^[0-9]+$)")) {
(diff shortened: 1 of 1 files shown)


bryancall:
[approve ci]

src/traffic_cache_tool/CacheDefs.h
```diff
@@ -307,7 +307,7 @@ struct url_matcher {
         std::cout << "Check your regular expression" << std::endl;
       }
 
-      if (!port.compile(R"([0-9]+$)")) {
+      if (!port.compile(R"(^[0-9]+$)")) {
         std::cout << "Check your regular expression" << std::endl;
         return;
       }
```
bryancall, line 313: Does the port compile below need to change too?

bryancall (changes requested):

src/traffic_cache_tool/CacheDefs.h
moonchen: Good catch!  Adding it now.

PR #12588, opened 2025-10-20
hrw4u: Add support for += / add-header

zwoop:
This also cleans up the maps.py Maps definitions, including moving most of the parameter values into a type class. This makes for more readable code, even though it adds a bit additional code.


zwoop:
I resolved the nitpick comments from CoPilot.

PR #12591, opened 2025-10-21
Add ts::Metrics::StaticString and remove RecRegisterStatString

cmcfarlen:


cmcfarlen:
[approve ci autest 3]

include/tsutil/Metrics.h
```diff
@@ -535,6 +537,41 @@ class Metrics
 
   }; // class Counter
 
+  class StaticString
+  {
+  public:
+    using StringStorage = std::map<std::string, std::string>;
```
bneradt: Seems like ordering isn't a value-add for this use case. Might as well use `std::unordered_map`?

PR #12592, opened 2025-10-21
Speed up day/month parsing

moonchen:
The "DFA" class is quite slow for parsing dates in headers because it uses regex matching.  I instead pack characters into an integer and do a simple linear search.

I benchmarked this approach against other methods like Trie, DFA, SIMD, tree search.  I found this method is among the fastest for this task, and remains portable, and readable.  It's at least 10x faster than the previous way on x64/arm CPUs I tested.

### DIFF
--- a/src/proxy/hdrs/MIME.cc
+++ b/src/proxy/hdrs/MIME.cc
@@ -54,16 +54,18 @@ using swoc::TextView;
  *                                                                     *
  ***********************************************************************/
 
-static constexpr std::array<std::string_view, 7> day_names = {
+namespace
+{
+constexpr std::array<std::string_view, 7> day_names = {
   "Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat",
 };
 
-static constexpr std::array<std::string_view, 12> month_names = {
+constexpr std::array<std::string_view, 12> month_names = {
   "Jan", "Feb", "Mar", "Apr", "May", "Jun", "Jul", "Aug", "Sep", "Oct", "Nov", "Dec",
 };
 
 template <size_t count>
-static consteval std::array<uint32_t, count>
+consteval std::array<uint32_t, count>
 make_packed(const std::array<std::string_view, count> &names)
 {
   std::array<uint32_t, count> packed{};
@@ -80,14 +82,14 @@ make_packed(const std::array<std::string_view, count> &names)
   return packed;
 }
 
-static constexpr std::array<uint32_t, day_names.size()>   day_names_packed   = make_packed(day_names);
-static constexpr std::array<uint32_t, month_names.size()> month_names_packed = make_packed(month_names);
+constexpr std::array<uint32_t, day_names.size()>   day_names_packed   = make_packed(day_names);
+constexpr std::array<uint32_t, month_names.size()> month_names_packed = make_packed(month_names);
 
 // Case-insensitive match of first 3 characters of input string against array of names
 // Longer strings will match if their first 3 characters match - this is intentional for
 // matching non-standard day/month names like "Thursday" or "September".
 template <size_t count>
-__attribute__((always_inline)) static constexpr int
+__attribute__((always_inline)) constexpr int
 match_3char_ci(const std::string_view s, const std::array<uint32_t, count> &names_packed)
 {
   if (s.size() < 3) {
@@ -104,6 +106,7 @@ match_3char_ci(const std::string_view s, const std::array<uint32_t, count> &name
   }
   return -1;
 }
+} // namespace
 
 struct MDY {
   uint8_t  m;


src/proxy/hdrs/MIME.cc
```diff
@@ -51,17 +53,58 @@ using swoc::TextView;
...
+match_3char_ci(const std::string_view s, const std::array<uint32_t, count> &names_packed)
+{
+  if (s.size() < 3) {
+    return -1;
+  }
+
+  auto           tl     = [](char c) -> char { return (c >= 'A' && c <= 'Z') ? (c + 32) : c; };
+  const uint32_t packed = (tl(s[0]) << 16) | (tl(s[1]) << 8) | tl(s[2]);
+
+  for (size_t i = 0; i < count; i++) {
```
masaori335, line 102: It's pretty interesting this liner search is the fastest. I thought switching this packed string is faster than for loop. Probably, target vector is short enough.

cmcfarlen: Since the packed names are constexpr, the compiler knows the length and contents at compile time.  So it can just hardcode the comparisons.  No loops, just cmp and jmp.

https://godbolt.org/z/Pzb7WT6Mr

cmcfarlen (changes requested):
Definitely worth adding a unit test for this.  Very cool though!

src/proxy/hdrs/MIME.cc
```diff
@@ -51,17 +53,58 @@ using swoc::TextView;
  *                          C O N S T A N T S                          *
  *                                                                     *
  ***********************************************************************/
-static DFA *day_names_dfa   = nullptr;
-static DFA *month_names_dfa = nullptr;
 
-static constexpr const char *day_names[] = {
+static constexpr std::array<std::string_view, 7> day_names = {
```
cmcfarlen: consider anonymous namespace instead of static for files scoped things.

moonchen: Done.

moonchen:
> Definitely worth adding a unit test for this. Very cool though!

I used the existing unit tests here: https://github.com/apache/trafficserver/blob/e319f593b51d55f983c970822ad45aa04975a7ea/src/proxy/hdrs/unit_tests/test_Hdrs.cc#L627

src/proxy/hdrs/MIME.cc
moonchen: It's definitely benefitting from the fact that there are so few days in a week or months in a year.

moonchen:
[approve ci autest]

moonchen:
Benchmark results from my laptop

```------------------------------------------------------------------------
Benchmark                              Time             CPU   Iterations
------------------------------------------------------------------------
BM_DFA_Day_Match                    68.4 ns         68.4 ns     10407374
BM_PCRE2Mark_Day_Match              20.1 ns         20.1 ns     35627398
BM_RE2Set_Day_Match                 65.3 ns         65.3 ns     10563008
BM_PlainTrie_Day_Match              9.62 ns         9.61 ns     71774260
BM_AhoCorasick_Day_Match            14.7 ns         14.7 ns     48626645
BM_AutoVec_Day_Match                4.62 ns         4.62 ns    162166921
BM_Switch_Day_Match                 4.12 ns         4.12 ns    152373655
BM_Neon_Day_Match                   2.34 ns         2.34 ns    293739981
BM_Scalar_Day_Match                 1.88 ns         1.88 ns    345227233
BM_NeonOpt_Day_Match                2.13 ns         2.13 ns    317109412
BM_Packed_Day_Match                 1.96 ns         1.96 ns    358410093
BM_DFA_Month_Match                  96.4 ns         96.4 ns      7190993
BM_PCRE2Mark_Month_Match            22.4 ns         22.3 ns     30183993
BM_RE2Set_Month_Match               66.0 ns         66.0 ns     10484692
BM_PlainTrie_Month_Match            9.33 ns         9.33 ns     73313015
BM_AhoCorasick_Month_Match          13.8 ns         13.8 ns     48758062
BM_AutoVec_Month_Match              4.84 ns         4.84 ns    137373420
BM_Switch_Month_Match               4.88 ns         4.88 ns    136962179
BM_Neon_Month_Match                 2.53 ns         2.53 ns    263606819
BM_Scalar_Month_Match               1.98 ns         1.98 ns    326133547
BM_NeonOpt_Month_Match              2.53 ns         2.52 ns    279464546
BM_Packed_Month_Match               2.19 ns         2.18 ns    326716545
BM_DFA_Combined_Match                121 ns          121 ns      5850055
BM_PCRE2Mark_Combined_Match         21.7 ns         21.7 ns     32689964
BM_RE2Set_Combined_Match            62.9 ns         62.9 ns     10471204
BM_PlainTrie_Combined_Match         10.3 ns         10.3 ns     63429928
BM_AhoCorasick_Combined_Match       14.6 ns         14.6 ns     45255012

PR #12593, opened 2025-10-21
Add to API to get a strategy during TSRemapNewInstance

traeak:
This gives the API access to the RemapConfig `url_mapping` during TSRemapNewInstance calls.  This contains both the assigned stragies pointer and a pointer to the loaded strategy factory.

This modifies the strategies ts API from the previous PR.

API is:

valid during TSRemapNewInstance:
```
TSStrategy TSRemapNextHopStrategyFind(const char *name);
TSStrategy TSRemapNextHopStrategyGet();
void TSRemapNextHopStrategySet(TSStrategy strategy);
```

valid during transaction (the Find call is changed):
```
TSStrategy TSHttpTxnNextHopStrategyFind(TSHttpTxn txnp, const char *name);
TSStrategy TSHttpTxnNextHopStrategyGet(TSHttpTxn txnp);
void TSHttpTxnNextHopStrategySet(TSHttpTxn txnp, TSStrategy strategy);
```

with utility function to get the null terminated strategy name (returns "null" if nullptr strategy):
```
char const *TSNextHopStrategyNameGet(TSStrategy strategy);
```

The header_rewrite and regex_remap plugins are modified to look up named strategies during TSRemapNewInstance instead of performing strategy factory lookups during each transaction remap hook.

### DIFF
--- a/doc/developer-guide/api/functions/TSRemapNextHopStrategySet.en.rst
+++ b/doc/developer-guide/api/functions/TSRemapNextHopStrategySet.en.rst
@@ -18,8 +18,8 @@
 
 .. default-domain:: cpp
 
-TSHttpTxnNextHopNamedStrategyGet
-********************************
+TSRemapNextHopNameGet
+***********************
 
 Synopsis
 ========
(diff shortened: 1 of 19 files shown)


doc/developer-guide/api/functions/TSRemapNextHopStrategySet.en.rst
```diff
@@ -0,0 +1,56 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+   implied.  See the License for the specific language governing
+   permissions and limitations under the License.
+
+.. include:: ../../../common.defs
+
+.. default-domain:: cpp
+
+TSRemapNextHopNameGet
```
zwoop: Should this be TSRemapNextHopNameSet() ?

zwoop (changes requested):

doc/developer-guide/api/functions/TSRemapNextHopStrategySet.en.rst
traeak: TSRemapNextHopStrategySet actually. I guess copilot didn't catch that one.

PR #12594, opened 2025-10-21
Maxmind pcre

ezelkow1:
Convert to regex from pcre

### DIFF
--- a/plugins/experimental/maxmind_acl/mmdb.cc
+++ b/plugins/experimental/maxmind_acl/mmdb.cc
@@ -411,8 +411,6 @@ Acl::parseregex(const YAML::Node &regex, bool allow)
             Dbg(dbg_ctl, "Adding regex: %s, for country: %s", temp._regex_s.c_str(), i[y].as<std::string>().c_str());
             if (allow) {
               allow_regex[i[y].as<std::string>()].push_back(temp);
-
-              // allow_regex[i[y].as<std::string>()].data().push_back(temp);
             } else {
               deny_regex[i[y].as<std::string>()].push_back(temp);
             }
@@ -524,7 +522,6 @@ Acl::eval(TSRemapRequestInfo * /* rri ATS_UNUSED */, TSHttpTxn txnp)
 
   MMDB_entry_data_list_s *entry_data_list = nullptr;
   if (result.found_entry) {
-    Dbg(dbg_ctl, "Found entry for this IP");
     int status = MMDB_get_entry_data_list(&result.entry, &entry_data_list);
     if (MMDB_SUCCESS != status) {
       Dbg(dbg_ctl, "Error looking up entry data: %s", MMDB_strerror(status));
--- a/plugins/experimental/maxmind_acl/mmdb.h
+++ b/plugins/experimental/maxmind_acl/mmdb.h
@@ -51,9 +51,7 @@ using namespace maxmind_acl_ns;
 
 struct plugin_regex {
   std::string _regex_s;
-  // pcre       *_rex;
-  // pcre_extra *_extra;
-  Regex *_rex;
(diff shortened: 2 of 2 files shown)


plugins/experimental/maxmind_acl/mmdb.cc
```diff
@@ -399,19 +399,11 @@ Acl::parseregex(const YAML::Node &regex, bool allow)
...
-              TSError("[%s] Failed to study regular expression in %s:%s", PLUGIN_NAME, temp._regex_s.c_str(), error);
-              return;
-            }
-          } else {
-            TSError("[%s] Failed to compile regular expression in %s: %s", PLUGIN_NAME, temp._regex_s.c_str(), error);
+          std::string error;
+          int         erroroffset = 0;
+          temp._rex               = new Regex();
+          if (!temp._rex->compile(temp._regex_s, error, erroroffset, 0)) {
+            TSError("[%s] Failed to compile regular expression in %s", PLUGIN_NAME, temp._regex_s.c_str());
```
cmcfarlen, line 406: Should delete `temp._rex` here since this returns without saving it.

plugins/experimental/maxmind_acl/mmdb.h
```diff
@@ -71,6 +66,20 @@ class Acl
...
+        }
+      }
+    }
+    for (auto &i : deny_regex) {
+      for (auto &y : i.second) {
+        if (y._rex != nullptr) {
+          delete y._rex;
+        }
+      }
+    }
```
cmcfarlen, line 82: Its seems like the regex could be added to multiple or these vectors.  Is it safe to delete them repeatedly if the pointers are copied?  Seems strange.

cmcfarlen (changes requested):
definitely a potential leak here, but also the cleanup looks strange since the temp plugin_regex could be pushed to multiple vectors.

plugins/experimental/maxmind_acl/mmdb.h
ezelkow1: Took me a while to actually remember what was going on in the parsing regex portion. It actually has undocumented capabilities I dont remember doing. Each rule is really supposed to be map(country-code)(regex-vector), however it will work with entries of [Country, Country, Regex] and add them properly, allowing this sharing of regexes.

Contemplating if I should just fix the feature and remove that capability of doing multiple countries to single regex since its undocumented as is and then would get rid of this issue

cmcfarlen: Could use shared_ptrs to refcount the instances, but I'm not sure how thread safe that would be.  @bryancall should Regex have a copy constructor?  Seems like somewhere else I saw needed to copy a Regex and ended up manually creating new and recompiling.

ezelkow1: I am also fine with just removing that loop in the regex parsing, so it would basically just take the first entry as the country code and the last entry as the regex, and skip any extra country codes that might exist. This is how its documented to work, being able to do this sharing shouldnt even technically be there anyway

bryancall: Seems odd to copy a compiled regex instead of sharing one.  My initial thought is we should be sharing the compiled regex instead.

If we want to implement a copy constructor I would use pcre2_code_copy() that can copy the pcre2 compiled, since we don't have the pattern and I wouldn't want to deal with reference counting and shallow copies.  If we want this as a feature open an issue and assign it to me.  It wouldn't be hard to add it.

bryancall: I created a PR for it #12612.  If we want it then we can merge it in.

ezelkow1:
moving to use the copy constructor

PR #12560, opened 2025-10-13
[header_rewrite] Fix set-body

moonchen:
1. set-body with an empty string argument causes heap corruption.  The cause is incorrect handling of zero-length strings in MIOBuffer::append_xmalloced.
2. set-body with variable expansion fails to expand the variable. The cause is a typo in the OperatorSetBody::exec() method.

Add autest for these bugs.

PR #12595, opened 2025-10-21
HRW: Adopt new hrw tests to new infrastructure

zwoop:
This adopts the new HRW autest from #12560 to the new plumbing put in place.


tests/gold_tests/pluginTest/header_rewrite/gold/set_body_status.gold
```diff
@@ -0,0 +1,16 @@
...
+``
+< HTTP/1.1 200 OK
+< Date: ``
+< Proxy-Connection: keep-alive
+< Server: ATS/``
+< Cache-Control: no-store
+< Content-Type: text/html
+< Content-Language: en
+< Content-Length: 3
+``
```
moonchen, line 16: I think we're also expecting a body here with the string "200".

moonchen (changes requested):
Thanks for incorporating the new autest into the existing framework.  I just noticed one small thing that was missed in the move.

tests/gold_tests/pluginTest/header_rewrite/gold/set_body_status.gold
zwoop: Ok. I was going to add that but it needed an extra file in the way things worked, and I figure the length would be sufficient :) I'll take a look again to see if there's an easy way to capture the stdout in this new infrastructure.

bneradt: Maybe update the tests to be able to use stdall instead of stdout?

bneradt (changes requested):

tests/gold_tests/pluginTest/header_rewrite/gold/set_body_status.gold
bneradt: Oh, I see. You verified both stdout and stderr to address @moonchen 's comment. Sounds good.

PR #12596, opened 2025-10-21
Another attempt to keep plugin DSOs in place

moonchen:
Even after the previous attempt to keep these plugin DSO files on the filesystem, they were still being deleted when we reload remap.config.


moonchen:
[approve ci fedora]

cmcfarlen:

PR #12599, opened 2025-10-22
regex_remap plugin fix for no specified strategy

traeak:
This expands the autest and fixes a bug where strategy ends up being cleared if there's no @strategies option.


cmcfarlen (approved):
LGTM, I suppose this is only an issue now because you can set the strategy in a plugin?

traeak:
correct i introduced this bug with the strategies in plugins code.

PR #12602, opened 2025-10-23
Reduce the calls to ink_get_hrtime in the event loop

cmcfarlen:
We have noticed that under high RPS loads ATS spends measurable amount of time just getting the current time.  In #10163, after some benchmarks, we changed from using a thread-local variable to cache the time to just calling `ink_get_hrtime` which fixed the time issue we had as well as increased RPS for some benchmarks.  That PR replaced all references to the time cache variable with calls to `ink_get_hrtime`.  That meant, however, that each iteration through the event loop calls `ink_get_hrtime` multiple times, including one time for each event processed.

This PR attempts to relax that a bit, but also provides a knob to control how often the event loop time is updated (as a percentage chance to be updated after an event is processed).  More often can lead to more accurate timers, but also increases the number of calls to get the time.

### DIFF
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -383,7 +383,8 @@ Thread Variables
    will create its own domain socket with a ``-<thread id>`` suffix added to the
    end of the path.
 
-.. ts:cv:: CONFIG proxy.config.exec_thread.event_time_update_rate INT 10
+.. ts:cv:: CONFIG proxy.config.exec_thread.loop_time_update_probability INT 10
+   :reloadable:
 
    This dynamically loadable setting controls the rate that exec thread loop timestamps are
    updated after processing an event given as a percentage from 0 to 100. 0
--- a/src/iocore/eventsystem/EventSystem.cc
+++ b/src/iocore/eventsystem/EventSystem.cc
@@ -44,8 +44,8 @@ ink_event_system_init(ts::ModuleVersion v)
 
   RecEstablishStaticConfigInt32(thread_freelist_low_watermark, "proxy.config.allocator.thread_freelist_low_watermark");
 
-  extern int event_time_update_rate;
-  RecEstablishStaticConfigInt32(event_time_update_rate, "proxy.config.exec_thread.event_time_update_rate");
+  extern int loop_time_update_probability;
+  RecEstablishStaticConfigInt32(loop_time_update_probability, "proxy.config.exec_thread.loop_time_update_probability");
 
   int chunk_sizes[DEFAULT_BUFFER_SIZES] = {0};
   {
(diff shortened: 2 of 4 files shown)


src/iocore/eventsystem/UnixEThread.cc
cmcfarlen, line 322: note the time is updated after waiting here unconditionally.

doc/admin-guide/files/records.yaml.en.rst
```diff
@@ -383,6 +383,18 @@ Thread Variables
    will create its own domain socket with a ``-<thread id>`` suffix added to the
    end of the path.
 
+.. ts:cv:: CONFIG proxy.config.exec_thread.event_time_update_rate INT 10
```
zwoop: Instead of spelling out that this is reloadable etc., you should add

```
   :reloadable:
```

src/iocore/eventsystem/EventSystem.cc
```diff
@@ -44,6 +44,9 @@ ink_event_system_init(ts::ModuleVersion v)
 
   RecEstablishStaticConfigInt32(thread_freelist_low_watermark, "proxy.config.allocator.thread_freelist_low_watermark");
 
+  extern int event_time_update_rate;
+  RecEstablishStaticConfigInt32(event_time_update_rate, "proxy.config.exec_thread.event_time_update_rate");
```
zwoop: Not sure I like the name of this config :)

Feels like it should be clear that it's for event metrics, and not the event system itself ?

bryancall:
Screenshot 2025-11-03 at 9 07 33 AM

doc/admin-guide/files/records.yaml.en.rst
```diff
@@ -383,6 +383,19 @@ Thread Variables
    will create its own domain socket with a ``-<thread id>`` suffix added to the
    end of the path.
 
+.. ts:cv:: CONFIG proxy.config.exec_thread.loop_time_update_probability INT 10
+   :reloadable:
+
+   This dynamically loadable setting controls the rate that exec thread loop timestamps are
```
zwoop, line 389: Should this say "this dynamic reloadable" ? Or just "this reloadable" ?

src/iocore/eventsystem/UnixEThread.cc
```diff
@@ -53,6 +54,7 @@ char const *const EThread::Metrics::Slice::STAT_NAME[] = {
 };
 
 int thread_max_heartbeat_mseconds = THREAD_MAX_HEARTBEAT_MSECONDS;
+int loop_time_update_probability  = 10;
```
zwoop, line 57: Mostly a question: Would it make sense if we were to start making these types of config driven globals std::atomic ?

PR #12603, opened 2025-10-23
Add traffic_ctl hostdb status

masaori335:
Dump all `HostDBRecords` to see current health state of each origin servers in the HostDB.

### DIFF
--- a/src/mgmt/rpc/handlers/hostdb/HostDB.cc
+++ b/src/mgmt/rpc/handlers/hostdb/HostDB.cc
@@ -48,6 +48,8 @@ str(HostDBType type)
   case HostDBType::UNSPEC:
     return "UNSPEC";
   }
+
+  return "";
 }
 
 constexpr std::string_view


src/traffic_ctl/traffic_ctl.cc
```diff
@@ -88,6 +88,7 @@ main([[maybe_unused]] int argc, const char **argv)
   auto &storage_command    = parser.add_command("storage", "Manipulate cache storage").require_commands();
   auto &plugin_command     = parser.add_command("plugin", "Interact with plugins").require_commands();
   auto &host_command       = parser.add_command("host", "Interact with host status").require_commands();
+  auto &hostdb_command     = parser.add_command("hostdb", "Interact with host status").require_commands();
```
brbzull0: should this be `Interact with HostDB status`? Just to differentiate from the one above.

masaori335: Ah, right!

src/mgmt/rpc/handlers/hostdb/HostDB.cc
```diff
@@ -0,0 +1,178 @@
...
+  switch (type) {
+  case HostDBType::ADDR:
+    return "ADDR";
+  case HostDBType::SRV:
+    return "SRV";
+  case HostDBType::HOST:
+    return "HOST";
+  case HostDBType::UNSPEC:
+    [[fallthrough]];
+  default:
```
bneradt: Let's leave `default` out so we get a compiler error if another HostDBType is added.

masaori335: Makes sense! Done.

brbzull0 (approved):

masaori335:
@bneradt Updated to follow your comment. PTAL!

bneradt:
> @bneradt Updated to follow your comment. PTAL!

Looks good, but we'll have to make the compiler happy by returning `""` at the end now:

```
../src/mgmt/rpc/handlers/hostdb/HostDB.cc:51:1: error: control reaches end of non-void function [-Werror=return-type]
   51 | }
      | ^
```

PR #12608, opened 2025-10-24
Exclude hop-by-hop headers from AWS v4 signature in origin_server_auth plugin

mlibbey:
The plugin was incorrectly including hop-by-hop headers like 'TE' in the AWS v4 signature calculation, causing origin servers to return 400 errors.

Per AWS documentation,
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html hop-by-hop headers that are frequently altered during transit should not be included in signatures. This includes:
- connection
- x-amzn-trace-id
- user-agent
- keep-alive
- transfer-encoding
- te
- trailer
- upgrade
- proxy-authorization
- proxy-authenticate

These headers are now excluded from the default exclude list to prevent signature validation failures at the origin server.


jasmine-nahrain:
Looks like the test failed because user-agent is no longer in the output.  You can just remove user-agent from the gold file and it should pass

mlibbey:
Thanks @jasmine-nahrain -- updated and the checks are now passing :)

bneradt (approved):
Looks reasonable.

cmcfarlen:
This was marked for 10.1.x but these files don't exist in that branch.

jasmine-nahrain:
It must be the UDS tests that dont exist

PR #12609, opened 2025-10-24
Doc: storage.yaml

masaori335:
Revive #11000 for the config project. Diffs from #11000 are

1. This deprecates `storage.config` and `volume.config`, but keeps them for compatibility, so we can merge to master branch.
2. Rename `spans.id` to `spans.name` for clarity 
3. Add `volumes.avg_obj_size` and `volumes.fragment_size` support

If this looks good, I'll work on implementation.

### DIFF
--- a/doc/admin-guide/files/storage.yaml.en.rst
+++ b/doc/admin-guide/files/storage.yaml.en.rst
@@ -43,7 +43,7 @@ The format of the :file:`storage.yaml` file is a series of lines of the form
          hash_seed: # optional, used to isolate lookup from path changes
      volumes:       # optional
        - id:        # identifier [1-255]
-         size:      # optional, size in percentage
+         size:      # optional, size in bytes or percentage
          scheme:    # optional, default to "http"
          ram_cache: # optional, default to "true"
          spans:     # optional
@@ -86,7 +86,7 @@ For :code:`volumes` the keys are
 |               | string      | use.                                                                                                    |
 +---------------+-------------+---------------------------------------------------------------------------------------------------------+
 | ram_cache     | boolean     | Control of ram caching for this volume. Default is ``true``. This may be desirable if you are using     |
-|               |             | something like ramdisks, to avoid wasting RAM and cpu time on double caching objects.                   |
+|               |             | something like ramdisks, to avoid wasting RAM and CPU time on double caching objects.                   |
 +---------------+-------------+---------------------------------------------------------------------------------------------------------+
 | avg_obj_size  | integer     | Overrides the global :ts:cv:`proxy.config.cache.min_average_object_size` configuration for this volume. |
 |               |             | This is useful if you have a volume that is dedicated for say very small objects, and you need a lot of |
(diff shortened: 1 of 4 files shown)


bryancall:
**Feature comparison: storage.config + volume.config vs storage.yaml**

I compared the documented features in the existing config files against the new `storage.yaml` docs and found a few gaps:

| # | Missing from storage.yaml docs | Notes |
|---|-------------------------------|-------|
| 1 | **Human-readable size suffixes** (K, M, G, T) | `storage.config` documents these. The new YAML examples only show raw byte values (e.g., `134217728` instead of `128M`). Should document whether YAML sizes support the same suffixes. |
| 2 | **Fragment size maximum** (4MB) | `volume.config` documents "this setting has a maximum value of 4MB". Not mentioned in `storage.yaml` `fragment_size` description. |
| 3 | **Cache invalidation warning** | Both `storage.config` and `volume.config` warn that changes effectively invalidate the cache. The new `storage.yaml` docs do not include this warning. |
| 4 | **Minimum span/volume sizes** | `storage.config` says "a formatted or raw disk must be at least 128 MB". `volume.config` says "128 MB is the smallest value" and sizes must be multiples of 128 MB. Not mentioned in `storage.yaml`. |
| 5 | **Striping behavior** | `volume.config` explains how volumes stripe across disks for parallel I/O. This is useful context missing from the new docs. |

Items 1-3 are the most important -- users will hit these in practice. 4 and 5 are nice-to-have.

Otherwise the docs look thorough. The examples are clear, the backwards compatibility section is helpful, and the `avg_obj_size` / `fragment_size` per-volume overrides are well documented.

doc/admin-guide/files/storage.yaml.en.rst
```diff
@@ -0,0 +1,497 @@
...
+
+   cache:
+     spans:
+       - name: disk.2
+         path: /dev/disk2
+     volumes:
+       - id: 1
+         spans:
+           - use: disk.2
+             size: 100%
```
bneradt: Would it be better in this example translation to use `id: 3` since the original had `volume=3`? I think the old value of 3 might be referenced in something like the `hosting.config` right?

bneradt (changes requested):

bryancall (changes requested):
Please see my comment on the PR about the changes in the documentation

doc/admin-guide/files/storage.yaml.en.rst
masaori335: Updated. I have no idea why "volume=3" is selected.

masaori335:
@bryancall 
> 1. Human-readable size suffixes (K, M, G, T)

Add an example with 64G

> 2. Fragment size maximum (4MB)

Updated

> 3. Cache invalidation warning

There is a warning after the key descriptions

> 4. Minimum span/volume sizes

It's already mentioned exactly same as storage.config ( it was just copied )

> 5. Striping behavior

Copied the sentence from volume.config

masaori335:
[approve ci centos]

PR #12612, opened 2025-10-24
Add copy constructor and assignment operator to Regex class

bryancall:
Implements deep copy using pcre2_code_copy() to enable safe copying of Regex objects and storage in STL containers.

## Changes

- Added copy constructor and assignment operator
- Uses pcre2_code_copy() for independent copies
- Preserves all pattern properties (flags, JIT, etc.)
- Copy assignment uses copy-and-swap idiom to avoid code duplication
- Safe self-assignment handling
- Added comprehensive unit tests (77 assertions)

## Benefits

- Enables natural C++ value semantics for Regex objects
- Allows storage in STL containers like std::vector
- Each copy is fully independent (no shared state)
- Exception-safe copy assignment

## Testing

All existing tests pass, plus 3 new test cases:
- Copy constructor (8 sections, 30+ assertions)
- Copy assignment (6 sections, 30+ assertions)  
- Copy with flags (1 section, 6+ assertions)

Total: 212 assertions passing

## Motivation

This solves memory management issues in plugins that need to share compiled regex patterns. For example, PR #12594 (maxmind_acl) currently has unsafe pointer sharing that can be replaced with safe value copying using this feature.

### DIFF
--- a/src/tsutil/Regex.cc
+++ b/src/tsutil/Regex.cc
@@ -155,6 +155,7 @@ RegexMatches::malloc(size_t size, void *caller)
   return ::malloc(size);
 }
 
+//----------------------------------------------------------------------------
 void
 RegexMatches::free(void *p, void *caller)
 {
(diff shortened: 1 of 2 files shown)


ezelkow1 (approved):
lgtm, will try with maxmind once this is in

PR #12613, opened 2025-10-24
Fix stringop-truncation warning in ats_unix_append_id

bryancall:
Fixes Fedora 42 gcc release build failure.

Replace `strncat` with `memcpy` to satisfy GCC's `-Werror=stringop-truncation` warning. The original code was safe due to the bounds check, but GCC couldn't determine this. Using `memcpy` makes the intent clearer and eliminates the warning.

**Affected Builds:**
- Fedora 41/42 gcc release

### DIFF
--- a/include/tscore/ink_inet.h
+++ b/include/tscore/ink_inet.h
@@ -323,8 +323,15 @@ inline void
 ats_unix_append_id(sockaddr_un *s, int id)
 {
   char tmp[16];
-  int  cnt     = snprintf(tmp, sizeof(tmp), "-%d", id);
-  int  old_len = ats_unix_path_len(s);
+  int  cnt = snprintf(tmp, sizeof(tmp), "-%d", id);
+
+  // Defensive check: snprintf can return negative on error or >= sizeof(tmp) if truncated
+  if (cnt < 0 || cnt >= static_cast<int>(sizeof(tmp))) {
+    ink_assert(!"snprintf failed or truncated in ats_unix_append_id");
+    return;
+  }
+
+  int old_len = ats_unix_path_len(s);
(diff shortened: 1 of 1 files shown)


include/tscore/ink_inet.h
```diff
@@ -323,9 +323,10 @@ inline void
 ats_unix_append_id(sockaddr_un *s, int id)
 {
   char tmp[16];
-  int  cnt = snprintf(tmp, sizeof(tmp), "-%d", id);
-  if (static_cast<size_t>(ats_unix_path_len(s) + cnt) < TS_UNIX_SIZE) {
-    strncat(s->sun_path, tmp, cnt);
+  int  cnt     = snprintf(tmp, sizeof(tmp), "-%d", id);
```
moonchen: snprintf can return a negative error or a value >= sizeof(tmp) when truncated. I recommend early return for cnt < 0 and either clamp cnt to sizeof(tmp)-1 before copying from tmp or write directly into sun_path with the remaining capacity; otherwise memcpy may read past tmp and cause UB.

bryancall: The size of tmp is over the size needed to convert an int to a string.   The format string and size of int are known, so I don't see how there would be an error.

bryancall: From Claude:
```
Buffer overflow: Impossible (snprintf guarantees this)
Negative return: Impossible (no encoding for integers)
Return > 16: Impossible (max is 12 + null = 13)
The check if (cnt < 0 || cnt >= static_cast<int>(sizeof(tmp))) is defensive programming but will never trigger in practice. It's not harmful, but it's also not strictly necessary for this specific case.
```

PR #12614, opened 2025-10-24
Fix nonnull warning in uri_signing unit tests

bryancall:
Fixes Fedora 41 gcc release build failure.

Add `nullptr` check before `strcmp` to satisfy GCC's `-Werror=nonnull` warning. The `expected_strip` parameter can be `nullptr` in some test cases (lines 325, 330, 362, 371, 376), so we need to check for null before passing to `strcmp`.

**Error:**
```
error: argument 2 null where non-null expected [-Werror=nonnull]
  228 |     if (strcmp(uri_strip, expected_strip) != 0) {
```

**Affected Builds:**
- Fedora 41 gcc release

### DIFF
--- a/plugins/experimental/uri_signing/unit_tests/uri_signing_test.cc
+++ b/plugins/experimental/uri_signing/unit_tests/uri_signing_test.cc
@@ -227,13 +227,11 @@ jws_parsing_helper(const char *uri, const char *paramName, const char *expected_
     resp = true;
     if (expected_strip != nullptr) {
       if (strcmp(uri_strip, expected_strip) != 0) {
-        cjose_jws_release(jws);
         resp = false;
       }
     } else {
       // expected_strip == nullptr means we expect uri_strip to be empty
       if (uri_strip[0] != '\0') {
-        cjose_jws_release(jws);
         resp = false;
       }
     }


plugins/experimental/uri_signing/unit_tests/uri_signing_test.cc
bneradt, line 243: Now that you're posting this fix, it makes me wonder whether nullptr is being handled properly at all by this helper function. If I understand correctly, if expected_strip is nullptr, the caller expects that paramName value to be empty. It doesn't look like this helper function verifies that but it should. Your fix doesn't make this worse, but while we're at it, it should be easy to fix that.

bneradt (changes requested):

bneradt (dismissed):
lgtm

plugins/experimental/uri_signing/unit_tests/uri_signing_test.cc
```diff
@@ -225,9 +225,17 @@ jws_parsing_helper(const char *uri, const char *paramName, const char *expected_
...
+      if (strcmp(uri_strip, expected_strip) != 0) {
+        cjose_jws_release(jws);
+        resp = false;
+      }
+    } else {
+      // expected_strip == nullptr means we expect uri_strip to be empty
+      if (uri_strip[0] != '\0') {
+        cjose_jws_release(jws);
+        resp = false;
+      }
```
JosiahWI, line 236: What's the reason for adding this new behavior?

bneradt: I requested it here:
https://github.com/apache/trafficserver/pull/12614#discussion_r2461797249

The callers used nullptr, but we didn't verify anything with it. Bryan is now adding verification when nullptr is passed in.

JosiahWI (dismissed):

JosiahWI (approved):

PR #12615, opened 2025-10-24
fedora:43 autest updates

bneradt:
This updates two autests that were failing on fedora:43:

* Python 3.14: The http2_write_threshold trickle python scripts did not work on the newer Python 3.14, which is bit more strict about possibly uninitialized variables. fedora:43 uses Python 3.14 instead of the Python 3.13 of fedora:42. This tweaks the scripts to make Python 3.14 happy.
* openssl 3.5 s_client output: The tls_check_dual_cert_selection_plugin expected different output from s_client. fedora:42 used openssl 3.2, fedora:43 has openssl 3.5. This updates the regex expectations for cert information to match for both versions.

### DIFF
--- a/tests/gold_tests/h2/trickle_client.py
+++ b/tests/gold_tests/h2/trickle_client.py
@@ -17,7 +17,6 @@
 #  limitations under the License.
 
 import argparse
-from email.message import EmailMessage as HttpHeaders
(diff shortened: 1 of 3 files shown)


cmcfarlen:

PR #12616, opened 2025-10-24
Fix benchmarks, add benchmark_Random

cmcfarlen:
```
benchmark name                       samples       iterations    est run time
IncRand                                        100             1     1.07609 s
                                        10.7421 ms    10.7171 ms    10.7663 ms
                                        124.984 us    111.625 us    141.798 us

ts::Random                                     100             1     7.69725 s
                                         76.151 ms    76.0106 ms    76.3351 ms
                                        813.041 us    657.296 us    1.22737 ms

std::mt19937_64                                100             1     1.43045 s
                                        14.2323 ms    14.1915 ms    14.2728 ms
                                         207.64 us    185.487 us    234.206 us

std::ranlux48_base                             100             1     1.34634 s
                                        13.5751 ms    13.5245 ms    13.6269 ms
                                        260.993 us    228.319 us    304.472 us

std::ranlux24_base                             100             1     1.47979 s
                                        14.9566 ms    14.9128 ms    14.9949 ms
                                        209.487 us    180.369 us    252.217 us

std::uniform_int_distribution                  100             1     5.52595 s
                                        55.4003 ms    55.3131 ms     55.484 ms
                                        436.427 us    388.289 us    499.327 us

```

### DIFF
--- a/tools/benchmark/benchmark_Random.cc
+++ b/tools/benchmark/benchmark_Random.cc
@@ -36,65 +36,40 @@ TEST_CASE("BenchRandom", "[bench][random]")
 {
   InkRand gen(42);
   ts::Random::seed(13);
-  int iterations = 1000000;
 
   BENCHMARK("IncRand")
   {
-    uint64_t sum = 0;
-    for (int i = 0; i < iterations; i++) {
-      sum += gen.random();
-    }
-    return sum;
+    return gen.random();
   };
 
   BENCHMARK("ts::Random")
   {
-    uint64_t sum = 0;
-    for (int i = 0; i < iterations; i++) {
-      sum += ts::Random::random();
-    }
-    return sum;
+    return ts::Random::random();
   };
 
   std::mt19937_64 mt;
   BENCHMARK("std::mt19937_64")
   {
-    uint64_t sum = 0;
-    for (int i = 0; i < iterations; i++) {
-      sum += mt();
-    }
-    return sum;
+    return mt();
   };
 
   std::ranlux48_base rb;
   BENCHMARK("std::ranlux48_base")
   {
-    uint64_t sum = 0;
-    for (int i = 0; i < iterations; i++) {
-      sum += rb();
-    }
-    return sum;
+    return rb();
   };
 
   std::ranlux24_base rb24;
   BENCHMARK("std::ranlux24_base")
   {
-    uint64_t sum = 0;
-    for (int i = 0; i < iterations; i++) {
-      sum += rb24();
-    }
-    return sum;
+    return rb24();
   };
 
   std::uniform_int_distribution<uint64_t> mtdist{0, UINT64_MAX};
 
   BENCHMARK("std::uniform_int_distribution")
   {
-    uint64_t sum = 0;
-    for (int i = 0; i < iterations; i++) {
-      sum += mtdist(mt);
-    }
-    return sum;
+    return mtdist(mt);
   };
 }
 


bneradt:
Looks good. Can you explain the fix part of this patch a bit? What was the test doing before your fix?

tools/benchmark/benchmark_Random.cc
```diff
@@ -0,0 +1,126 @@
...
+  int iterations = 1000000;
+
+  BENCHMARK("IncRand")
+  {
+    uint64_t sum = 0;
+    for (int i = 0; i < iterations; i++) {
+      sum += gen.random();
+    }
+    return sum;
+  };
```
JosiahWI, line 43: ~~Have you compared results with a signed integer? I hope that the performance cost of the overflow check necessitated by the unsigned arithmetic is negligible in the results, but it would ease my mind to see a confirmation.~~

cmcfarlen:
The fix was just catch2 changes required to compile.  The only change is adding the random gen benchmark.

tools/benchmark/benchmark_Random.cc
JosiahWI: Catch2 will already perform multiple iterations of the benchmark. What is the reason for performing a fixed number of iterations in the benchmarked routine? Does Catch2 underestimate the number of runs to perform?

tools/benchmark/benchmark_Random.cc
```diff
@@ -0,0 +1,126 @@
...
+  {
+    uint64_t sum = 0;
+    for (int i = 0; i < iterations; i++) {
+      sum += mtdist(mt);
+    }
+    return sum;
+  };
+}
+
+TEST_CASE("RandomDistribution", "[random][distribution]")
```
JosiahWI, line 76: What's the rationale for including this test case in this PR (and this file)?

tools/benchmark/benchmark_Random.cc
cmcfarlen: Oh, that's a good idea.  I can add that from the std side easily.  From what I understand, the random engines are uniform integer generators, but they can have bias around some ranges.  I wanted to measure that bias to see if we actually care about that enough to accept the overhead the uniform distribution has.  I don't think we do, but I'm still investigating.

JosiahWI: My first comment here is incorrect. In this particular case there is no difference in the signed vs unsigned integer arithmetic on x86_64.

tools/benchmark/benchmark_Random.cc
cmcfarlen: I don't have one other than this work was about InkRandom and I didn't want to slow down the normal unit tests.  I could move this one to the relevant unit test since I don't think it takes too long to run as it is.

bneradt:
> The fix was just catch2 changes required to compile. The only change is adding the random gen benchmark.

I see. These tests weren't building in CI before because it requires ENABLE_BENCHMARKS. Thanks for the clarification.

bryancall (approved):
LGTM. Nice addition of the Random benchmark - the performance comparison between InkRand and ts::Random is particularly useful. The Catch2 API fixes are straightforward and necessary for building with current Catch2 versions.

JosiahWI:
I don't have any issue with the code that is worth blocking a merge with, so I'm not requesting changes, but I still find it very odd that the code being benchmarked does a fixed number of iterations instead of relying on Catch2 to choose the number of iterations; that also has the effect of making the benchmarks less concise.

(If I understand the comment about bias correctly, the number of iterations is fixed to ensure that parts of the random sequence that might be biased get covered by the benchmark, but it wasn't clear to me, so I'd like to check that understanding.

cmcfarlen:
It doesn't change the relative perf of this set of benchmarks so I removed iterations.

cmcfarlen:
```
benchmark name                       samples       iterations    est run time
IncRand                                        100          1058     1.7986 ms
                                        17.5023 ns    17.2608 ns    17.7926 ns
                                         1.3458 ns    1.12171 ns    1.64001 ns

ts::Random                                     100           219     1.8615 ms
                                        86.8963 ns    85.3129 ns    88.7114 ns
                                         8.6841 ns    7.43794 ns    10.0167 ns

std::mt19937_64                                100           805     1.8515 ms
                                        22.7784 ns    22.4652 ns    23.5745 ns
                                        2.38374 ns    1.15658 ns    4.95821 ns

std::ranlux48_base                             100           873     1.8333 ms
                                        20.5588 ns    20.4266 ns    20.7493 ns
                                       0.795424 ns   0.521651 ns    1.38877 ns

std::ranlux24_base                             100           903      1.806 ms
                                        20.6602 ns    20.2624 ns    21.1987 ns
                                        2.35113 ns    1.83909 ns    3.01295 ns

std::uniform_int_distribution                  100           284      1.846 ms
                                         65.698 ns     64.881 ns    66.6444 ns
                                        4.46073 ns    3.68056 ns     5.3558 ns

```

PR #12617, opened 2025-10-27
Change polite_hook_wait Au test to demonstrate a useful scenario.

wkaras:
See https://lists.apache.org/thread/t6j6kk7454tow5j3tqh9nfsz32rlvw5w .


wkaras:
I don't think my change (to an Au test dedicated plugin) could actually break AIO unit testing.

wkaras:
When I first wrote this Au test, I wasn't aware that TSHttpTxnReenable() had logic in it that allowed it to be called in a spawned thread.

wkaras:
[approve ci fedora]

wkaras:
Hey Chris!  Can you merge it?  I'm using my non-Yahoo account that doesn't have write access to TS.

PR #12618, opened 2025-10-27
Respect Vary rules when ignore_accept_encoding_mismatch is set to `2` (default value)

### LINKED ISSUE #7458: cache vary problem with normalized_ae ( always prefer gzip ) ？
I had set `proxy.config.http.normalize_ae to 2` and enabled brotli ( br ) compression，

when ats(v9.0.0) had cached both gzip and brotli compressed version of the object,  
and **if gzip object was generated after brotli one**， ats will always return gzip **when client supports both gzip and brotli.**

steps to reproduce the problem:

`1、Accept-Encoding: gzip,br`
    ats will return br encoding as expected

`2、Accept-Encoding: gzip`
    ats will return gzip encoding as expected

`3、Accept-Encoding: gzip,br`
    **** _**ats will always return gzip encoding , which should be br encoding.**_ ****

`4、Accept-Encoding: br`
    ats will return br encoding as expected.

JakeChampion:
I believe the previous implementation was a bug because it was treating any non-zero `proxy.config.http.cache.ignore_accept_encoding_mismatch` value as an indication to "ignore the mismatch", but the the documented default value of 2 is meant to honor `Vary: Accept-Encoding`

> .. ts:cv:: CONFIG proxy.config.http.cache.ignore_accept_encoding_mismatch INT 2
>    :reloadable:
>    :overridable:
> 
>    When enabled with a value of ``1``, |TS| serves documents from cache with a
>    ``Content-Encoding:`` header even if it does not match the ``Accept-Encoding:``
>    header of the request. If set to ``2`` (default), this logic only happens in the absence of a
>    ``Vary`` header in the cached response (which is the recommended and safe use).
> 
> .. note::
> 
>    This option should only be enabled with ``1`` if you're having
>    problems with caching *and* you origin server doesn't set the ``Vary``
>    header. Alternatively, if the origin is incorrectly setting
>    ``Vary: Accept-Encoding`` or doesn't respond with ``406 (Not Acceptable)``
>    you can also enable this configuration with a ``1``.

previously, the default value `2` which should still respect `Vary: Accept-Encoding` when the origin supplies it, ended up collapsing distinct encoding variants into a single cache alternate, meaning requests with `Accept-Encoding: br` could get a response cached for `gzip`, which the tests confirmed

changing the check to only be for when the value is set to `1` looks to be the behaviour we do want, and the original test assertions now pass

Fixes #7458

### DIFF
--- a/src/iocore/cache/HttpTransactCache.cc
+++ b/src/iocore/cache/HttpTransactCache.cc
@@ -1244,7 +1244,11 @@ HttpTransactCache::CalcVariability(const HttpConfigAccessor *http_config_params,
 
         // Disable Vary mismatch checking for Accept-Encoding.  This is only safe to
         // set if you are promising to fix any Accept-Encoding/Content-Encoding mismatches.
-        if (http_config_params->get_ignore_accept_encoding_mismatch() &&
+        // Only suppress variability checks when the operator explicitly set
+        // proxy.config.http.cache.ignore_accept_encoding_mismatch to 1. The
+        // documented default value of 2 should continue to enforce Vary header
+        // semantics whenever the origin sends one.
+        if ((http_config_params->get_ignore_accept_encoding_mismatch() == 1) &&
             !strcasecmp(const_cast<char *>(field->str), "Accept-Encoding")) {
           continue;
         }
--- a/tests/gold_tests/headers/normalized_ae_match_vary_cache.test.py
+++ b/tests/gold_tests/headers/normalized_ae_match_vary_cache.test.py
@@ -31,6 +31,37 @@ testName = "NORMALIZE_AE_MATCH_VARY"
 replay_file = "replays/normalized_ae_varied_transactions.replay.yaml"
 server = Test.MakeVerifierServerProcess("server", replay_file)
 
+# Verify that cache hit requests never reach the server
+# Case 2 (normalize_ae:1) cache hits
+server.Streams.stdout += Testers.ExcludesExpression(
+    "uuid: 12", "Verify empty Accept-Encoding (uuid 12) is a cache hit and doesn't reach the server.")
+server.Streams.stdout += Testers.ExcludesExpression(
+    "uuid: 13", "Verify deflate request (uuid 13) is a cache hit and doesn't reach the server.")
+server.Streams.stdout += Testers.ExcludesExpression(
+    "uuid: 14", "Verify br,compress request (uuid 14) is a cache hit and doesn't reach the server.")
+server.Streams.stdout += Testers.ExcludesExpression(
+    "uuid: 16", "Verify br,compress,gzip request (uuid 16) is a cache hit and doesn't reach the server.")
+# Case 3 (normalize_ae:2) cache hits
+server.Streams.stdout += Testers.ExcludesExpression(
+    "uuid: 22", "Verify empty Accept-Encoding (uuid 22) is a cache hit and doesn't reach the server.")
+server.Streams.stdout += Testers.ExcludesExpression(
+    "uuid: 23", "Verify deflate request (uuid 23) is a cache hit and doesn't reach the server.")
+server.Streams.stdout += Testers.ExcludesExpression(
+    "uuid: 26", "Verify br,compress,gzip request (uuid 26) is a cache hit and doesn't reach the server.")
+server.Streams.stdout += Testers.ExcludesExpression(
+    "uuid: 27", "Verify compress,gzip request (uuid 27) is a cache hit and doesn't reach the server.")
+# Case 4 (normalize_ae:3) cache hits
+server.Streams.stdout += Testers.ExcludesExpression(
+    "uuid: 32", "Verify empty Accept-Encoding (uuid 32) is a cache hit and doesn't reach the server.")
+server.Streams.stdout += Testers.ExcludesExpression(
+    "uuid: 33", "Verify deflate request (uuid 33) is a cache hit and doesn't reach the server.")
+server.Streams.stdout += Testers.ExcludesExpression(
+    "uuid: 37", "Verify compress,gzip request (uuid 37) is a cache hit and doesn't reach the server.")
+server.Streams.stdout += Testers.ExcludesExpression(
+    "uuid: 38", "Verify br;q=1.1 request (uuid 38) is a cache hit and doesn't reach the server.")
+server.Streams.stdout += Testers.ExcludesExpression(
+    "uuid: 39", "Verify br,gzip;q=0.8 request (uuid 39) is a cache hit and doesn't reach the server.")
+
 ts = Test.MakeATSProcess("ts", enable_cache=True)
 ts.Disk.remap_config.AddLine(
     f"map http://www.ae-0.com http://127.0.0.1:{server.Variables.http_port}" +
--- a/tests/gold_tests/headers/replays/normalized_ae_varied_transactions.replay.yaml
+++ b/tests/gold_tests/headers/replays/normalized_ae_varied_transactions.replay.yaml
@@ -96,7 +96,7 @@ sessions:
         - [ X-Response-Identifier, { value: Empty-Accept-Encoding, as: equal } ]
         - [ X-Cache, { value: miss, as: equal } ]
 
-  # Accept-Encoding header deflate would match the alternate of empty Accept-Encoding header
+  # request deflate Accept-Encoding when origin lacks that variant
   - client-request:
       method: "GET"
       version: "1.1"
@@ -113,13 +113,12 @@ sessions:
       <<: *404_response
 
     proxy-response:
-      status: 200
+      status: 404
       headers:
         fields:
-        - [ X-Response-Identifier, { value: Empty-Accept-Encoding, as: equal } ]
-        - [ X-Cache, { value: hit-fresh, as: equal } ]
+        - [ X-Cache, { value: miss, as: equal } ]
 
-  # Accept-Encoding header br, compress would match the alternate of empty Accept-Encoding header
+  # load an alternate of br Accept-Encoding header
   - client-request:
       method: "GET"
       version: "1.1"
@@ -133,14 +132,23 @@ sessions:
       delay: 100ms
 
     server-response:
-      <<: *404_response
+      status: 200
+      reason: OK
+      headers:
+        fields:
+        - [ Transfer-Encoding, chunked ]
+        - [ Cache-Control, max-age=300 ]
+        - [ Content-Encoding, br ]
+        - [ Vary, Accept-Encoding ]
+        - [ Connection, close ]
+        - [ X-Response-Identifier, Br-Accept-Encoding ]
 
     proxy-response:
       status: 200
       headers:
         fields:
-        - [ X-Response-Identifier, { value: Empty-Accept-Encoding, as: equal } ]
-        - [ X-Cache, { value: hit-fresh, as: equal } ]
+        - [ X-Response-Identifier, { value: Br-Accept-Encoding, as: equal } ]
+        - [ X-Cache, { value: miss, as: equal } ]
 
   # load an alternate of gzip Accept-Encoding header
   - client-request:
@@ -176,7 +184,7 @@ sessions:
         - [ X-Response-Identifier, { value: Gzip-Accept-Encoding, as: equal } ]
         - [ X-Cache, { value: miss, as: equal } ]
 
-  # Accept-Encoding header br, compress, gzip would match the alternate of gzip Accept-Encoding header
+  # load an alternate of br Accept-Encoding header
   - client-request:
       method: "GET"
       version: "1.1"
@@ -190,14 +198,23 @@ sessions:
       delay: 100ms
 
     server-response:
-      <<: *404_response
+      status: 200
+      reason: OK
+      headers:
+        fields:
+        - [ Transfer-Encoding, chunked ]
+        - [ Cache-Control, max-age=300 ]
+        - [ Content-Encoding, br ]
+        - [ Vary, Accept-Encoding ]
+        - [ Connection, close ]
+        - [ X-Response-Identifier, Br-Accept-Encoding ]
 
     proxy-response:
       status: 200
       headers:
         fields:
-        - [ X-Response-Identifier, { value: Gzip-Accept-Encoding, as: equal } ]
-        - [ X-Cache, { value: hit-fresh, as: equal } ]
+        - [ X-Response-Identifier, { value: Br-Accept-Encoding, as: equal } ]
+        - [ X-Cache, { value: miss, as: equal } ]
 
 
   # Case 2 proxy.config.http.normalize_ae:1
@@ -699,10 +716,6 @@ sessions:
         - [ X-Response-Identifier, { value: Gzip-Accept-Encoding, as: equal } ]
         - [ X-Cache, { value: miss, as: equal } ]
 
-  # NOTICE: This case should load an alternate of br, gzip Accept-Encoding header.
-  #         However, due to the implementation of calculate_quality_of_match(),
-  #         ATS matches the alternate of gzip Accept-Encoding header.
-  #         The result is DIFFERENT from the description of proxy.config.http.normalize_ae: 3
   - client-request:
       method: "GET"
       version: "1.1"
@@ -733,10 +746,8 @@ sessions:
       status: 200
       headers:
         fields:
-        # - [ X-Response-Identifier, { value: Br-Gzip-Accept-Encoding, as: equal } ]
-        # - [ X-Cache, { value: miss, as: equal } ]
-        - [ X-Response-Identifier, { value: Gzip-Accept-Encoding, as: equal } ]
-        - [ X-Cache, { value: hit-fresh, as: equal } ]
+        - [ X-Response-Identifier, { value: Br-Gzip-Accept-Encoding, as: equal } ]
+        - [ X-Cache, { value: miss, as: equal } ]
 
   # Accept-Encoding header compress, gzip would match the alternate of gzip Accept-Encoding header
   - client-request:
@@ -784,11 +795,6 @@ sessions:
         - [ X-Response-Identifier, { value: Br-Accept-Encoding, as: equal } ]
         - [ X-Cache, { value: hit-fresh, as: equal } ]
 
-  # NOTICE: This case should make Accept-Encoding header br, gzip;q=0.8 match
-  #         the alternate of br, gzip Accept-Encoding header.
-  #         However, due to the implementation of calculate_quality_of_match(),
-  #         ATS matches the alternate of gzip Accept-Encoding header.
-  #         The result is DIFFERENT from the description of proxy.config.http.normalize_ae: 3
   - client-request:
       method: "GET"
       version: "1.1"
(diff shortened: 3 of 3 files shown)


src/iocore/cache/HttpTransactCache.cc
```diff
@@ -979,60 +979,38 @@ HttpTransactCache::calculate_quality_of_accept_encoding_match(MIMEField *accept_
...
+    bool  found_match = false;
+
     for (c_value = c_values_list.head; c_value; c_value = c_value->next) {
       float this_q = -1.0;
       if (!match_accept_content_encoding(c_value->str, accept_field, &wildcard_present, &wildcard_q, &this_q)) {
         goto encoding_wildcard;
       }
-      combined_q *= this_q;
+      if (this_q >= 0.0) {
+        found_match = false;
```
JosiahWI:
```suggestion
        found_match = true;
```

?

JakeChampion:
> Looks like the file needs a newline added (at the end). I did not notice whether that was the case before you applied my suggestion. Did the restored AuTests pass prior to applying my suggestion?

yes, they all passed before the suugestion was applied

JosiahWI:
```
   Run: 0-tr: Failed
     Test : Checking that True == _isRunningAfter - Passed
        Reason: Returned Value: True == _isRunningAfter
     Starting TestRun 0-tr : No Issues found - Passed
        Reason: Started!
     Process: Default: Failed
       Setting up : Copying '/home/jenkins/workspace/Github_Builds/autest/src/build/proxy-verifier-v2.12.0/linux-amd64' to '/tmp/sandbox/normalized_ae_match_vary_cache/client/bin' - Passed
       Setting up : Copying 'replays/normalized_ae_varied_transactions.replay.yaml' to '/tmp/sandbox/normalized_ae_match_vary_cache/client/normalized_ae_varied_transactions.replay.yaml' - Passed
       Setting up : Copying '/home/jenkins/workspace/Github_Builds/autest/src/tests/gold_tests/autest-site/../../tools/proxy-verifier/ssl/client.pem' to '/tmp/sandbox/normalized_ae_match_vary_cache/client/client.pem' - Passed
       Setting up : Copying '/home/jenkins/workspace/Github_Builds/autest/src/tests/gold_tests/autest-site/../../tools/proxy-verifier/ssl/ca.pem' to '/tmp/sandbox/normalized_ae_match_vary_cache/client/ca.pem' - Passed
          Reason: Returned Value 1 != 0
          Reason: Returned value: 2.515380620956421 < 600.0
       file /tmp/sandbox/normalized_ae_match_vary_cache/_output/0-tr-Default/stream.stdout.txt : There should be no Proxy Verifier violation errors. - Failed
          Reason: Contents of /tmp/sandbox/normalized_ae_match_vary_cache/_output/0-tr-Default/stream.stdout.txt contains expression: "Violation|Invalid status"
             Details:
               [INFO]: Equals Violation: Different. Key: "06", Field Name: "x-cache", Correct Value: "miss", Actual Value: "hit-fresh" : 27
               [INFO]: Equals Violation: Different. Key: "06", Field Name: "x-response-identifier", Correct Value: "Br-Accept-Encoding", Actual Value: "Gzip-Accept-Encoding" : 28
               [INFO]: Equals Violation: Different. Key: "36", Field Name: "x-cache", Correct Value: "miss", Actual Value: "hit-fresh" : 65
               [INFO]: Equals Violation: Different. Key: "36", Field Name: "x-response-identifier", Correct Value: "Br-Gzip-Accept-Encoding", Actual Value: "Gzip-Accept-Encoding" : 66
               [INFO]: Equals Violation: Different. Key: "39", Field Name: "x-response-identifier", Correct Value: "Br-Gzip-Accept-Encoding", Actual Value: "Gzip-Accept-Encoding" : 72
```

Looks like the tests no longer pass after the suggestion was applied.

src/iocore/cache/HttpTransactCache.cc
```diff
@@ -979,60 +979,38 @@ HttpTransactCache::calculate_quality_of_accept_encoding_match(MIMEField *accept_
...
+    bool  found_match = false;
+
     for (c_value = c_values_list.head; c_value; c_value = c_value->next) {
       float this_q = -1.0;
       if (!match_accept_content_encoding(c_value->str, accept_field, &wildcard_present, &wildcard_q, &this_q)) {
         goto encoding_wildcard;
       }
-      combined_q *= this_q;
+      if (this_q >= 0.0) {
+        found_match = true;
```
JakeChampion:
```suggestion
        found_match = false;
```

JakeChampion:
@JosiahWI I think the actual issue ended up being a bug in how we were deciding when to disable the vary mis-match for accept-encoding

tests/gold_tests/headers/replays/normalized_ae_varied_transactions.replay.yaml
```diff
@@ -110,16 +110,25 @@ sessions:
...
+        - [ X-Response-Identifier, Deflate-Accept-Encoding ]
 
     proxy-response:
       status: 200
       headers:
         fields:
-        - [ X-Response-Identifier, { value: Empty-Accept-Encoding, as: equal } ]
-        - [ X-Cache, { value: hit-fresh, as: equal } ]
+        - [ X-Response-Identifier, { value: Deflate-Accept-Encoding, as: equal } ]
+        - [ X-Cache, { value: miss, as: equal } ]
```
JosiahWI: The implication of the test setup in the old revision is that the server does not have a content representation with the `deflate` encoding. The server response here should be a version of the resource with no content encoding, or a 406 response. The 404 response used in the old revision seems ok for the purpose of the test as well. In any case, the origin should not return a deflate encoded resource here.

The interesting question here is whether ATS is conforming to Vary semantics by replying out of cache with the empty response. Section 5.3.4 of RFC 4231 explicitly says that the **origin server** should respond with no encoding if it can't supply a requested encoding. Together with the rules for caching responses with a Vary header described in section 4.1 of RFC 7234, I believe the old revision of this test may be nonconforming.

In summary, I think the correct expectation for the test, to minimize unnecessary changes, should be:

- client requests deflate encoding
- server replies with 404 Not Found
- proxy forwards 404 Not Found (cache miss)

or alternatively:

- client requests deflate encoding
- server replies with 200 OK and no content encoding
- proxy forwards the 200 OK (cache miss)

tests/gold_tests/headers/replays/normalized_ae_varied_transactions.replay.yaml
```diff
@@ -133,14 +142,23 @@ sessions:
...
+        - [ X-Response-Identifier, Br-Accept-Encoding ]
 
     proxy-response:
       status: 200
       headers:
         fields:
-        - [ X-Response-Identifier, { value: Empty-Accept-Encoding, as: equal } ]
-        - [ X-Cache, { value: hit-fresh, as: equal } ]
+        - [ X-Response-Identifier, { value: Br-Accept-Encoding, as: equal } ]
+        - [ X-Cache, { value: miss, as: equal } ]
```
JosiahWI, line 151: Similarly to the previous case, the server should return 404 Not Found here, and the proxy should forward it.

tests/gold_tests/headers/replays/normalized_ae_varied_transactions.replay.yaml
```diff
@@ -162,6 +180,7 @@ sessions:
         fields:
         - [ Transfer-Encoding, chunked ]
         - [ Cache-Control, max-age=300 ]
+        - [ Content-Encoding, gzip ]
```
JosiahWI: This is not strictly necessary, but I suppose it doesn't hurt.

tests/gold_tests/headers/replays/normalized_ae_varied_transactions.replay.yaml
```diff
@@ -190,14 +209,23 @@ sessions:
...
+        - [ X-Response-Identifier, Br-Accept-Encoding ]
 
     proxy-response:
       status: 200
       headers:
         fields:
-        - [ X-Response-Identifier, { value: Gzip-Accept-Encoding, as: equal } ]
-        - [ X-Cache, { value: hit-fresh, as: equal } ]
+        - [ X-Response-Identifier, { value: Br-Accept-Encoding, as: equal } ]
+        - [ X-Cache, { value: miss, as: equal } ]
```
JosiahWI, line 217: Why shouldn't the client reply with the gzip encoded alternate from the previous response as in the old revision?

tests/gold_tests/headers/replays/normalized_ae_varied_transactions.replay.yaml
```diff
@@ -722,6 +751,7 @@ sessions:
         fields:
         - [ Transfer-Encoding, chunked ]
         - [ Cache-Control, max-age=300 ]
+        - [ Content-Encoding, "br, gzip" ]
```
JosiahWI: [nitpick] This does not make sense. Just choose one or the other or leave it off. it doesn't matter for the purpose of the test.

src/iocore/cache/HttpTransactCache.cc
```diff
@@ -1244,7 +1244,11 @@ HttpTransactCache::CalcVariability(const HttpConfigAccessor *http_config_params,
 
         // Disable Vary mismatch checking for Accept-Encoding.  This is only safe to
         // set if you are promising to fix any Accept-Encoding/Content-Encoding mismatches.
-        if (http_config_params->get_ignore_accept_encoding_mismatch() &&
+        // Only suppress variability checks when the operator explicitly set
+        // proxy.config.http.cache.ignore_accept_encoding_mismatch to 1. The
+        // documented default value of 2 should continue to enforce Vary header
+        // semantics whenever the origin sends one.
+        if ((http_config_params->get_ignore_accept_encoding_mismatch() == 1) &&
```
JosiahWI, line 1251: This looks correct. Good fix. ~~Sadly, I think this may not be covered by any AuTests. (The AuTest adjusted in this PR has `ignore_accept_encoding_mismatch = 2`).~~ This is covered by the AuTest in this PR because the old behavior was to ignore an encoding mismatch even when the option required ATS to enforce Vary header semantics.

tests/gold_tests/headers/replays/normalized_ae_varied_transactions.replay.yaml
JakeChampion: I believe this is because this origin’s is configured with normalize_ae set to 0 which keeps the Accept-Encoding header verbatim, no normalisation - and the accept-encoding header value is different to the previously seen values, which is why this is a cache miss

JosiahWI: That does seem to be the case. Good catch.

JosiahWI:
The Rocky CI ran into this again (I've seen it before):

```
146/146 Test  #26: test_cache_CacheAggregateWriteBuffer ...***Timeout 1500.11 sec
Randomness seeded to: 57145569
===============================================================================
All tests passed (2 assertions in 2 test cases)

=================================================================
==7089==ERROR: AddressSanitizer: heap-use-after-free on address 0x6220000039c8 at pc 0x000000c81eeb bp 0x7fbfc60e1aa0 sp 0x7fbfc60e1a90
READ of size 8 at 0x6220000039c8 thread T1 ([ET_NET 0])
```

JosiahWI:
[approve ci rocky]

JakeChampion:
> I am suspicious of all the tests in this AuTest (including ones you have not touched) that reply with the empty-encoding alternate in response to requests for a deflate encoded resource. I think that is nonconforming to the RFCs and needs to be fixed. I would appreciate if @maskit or someone could double check my work on that.
> 
> Other than that general concern, I have some comments pertaining directly to the changes in this PR. Please take a look.

I believe those tests are correct because of https://www.rfc-editor.org/rfc/rfc9110.html#section-12.5.3-10.2 which states
>If the representation has no content coding, then it is acceptable by default unless specifically excluded by the Accept-Encoding header field stating either "identity;q=0" or "*;q=0" without a more specific entry for "identity".

we are not sending either "identity;q=0" or "*;q=0" - so ATS can send back the identity response it has

JosiahWI:
> > I am suspicious of all the tests in this AuTest (including ones you have not touched) that reply with the empty-encoding alternate in response to requests for a deflate encoded resource. I think that is nonconforming to the RFCs and needs to be fixed. I would appreciate if @maskit or someone could double check my work on that.
> > Other than that general concern, I have some comments pertaining directly to the changes in this PR. Please take a look.
> 
> I believe those tests are correct because of https://www.rfc-editor.org/rfc/rfc9110.html#section-12.5.3-10.2 which states
> 
> > If the representation has no content coding, then it is acceptable by default unless specifically excluded by the Accept-Encoding header field stating either "identity;q=0" or "*;q=0" without a more specific entry for "identity".
> 
> we are not sending either "identity;q=0" or "*;q=0" - so ATS can send back the identity response it has

Whoops, I was looking at the obsoleted RFCs, thanks you for the correction! I think it is still non-conforming for a proxy to reply out-of-cache in this scenario, however: https://www.rfc-editor.org/rfc/rfc9111#section-4.1-1.

We could argue about wording, but I think it will be more profitable to talk about the overall meaning of the RFCs. First let me go over my understanding of the background, just so I know we're on the same page about the setup. Suppose that ATS receives a request to `ae-0.com/path` with `Accept-Encoding: ` as in the AuTests here, and that no alternate for that resource has been cached. ATS will request the resource from the origin server, get a response with `Vary: Accept-Encoding` and cache it. Now, suppose that ATS receives another request for `ae-0.com/path` with `Accept-Encoding: deflate`. Should ATS reply out of cache? The user agent must be capable of processing a response with no content-encoding, so functionally everything should be fine if we reply out of cache. Here's the issue: _why_ did the user agent send the `Accept-Encoding: deflate` header? The only reason to send that header is to let us know we can compress the content to save bandwidth, which is desirable. [Section 12.5.3](https://www.rfc-editor.org/rfc/rfc9110.html#section-12.5.3-16) of RFC 9110 provides evidence for this understanding, by suggesting that this optimization is so valuable for transferring large content that a server might even want to inform a client that it ought to consider accepting compressed content. ATS, by replying out of cache with uncompressed content, has ignored the possibility that the origin **does have** deflate compression. Up to this point, there is no problem with current ATS behavior, because we see that in fact ATS will _not immediately_ reply out of cache, but will go to the origin server instead to see if it has an alternate with deflate encoding.

The real issue I see with the current behavior is that ATS, having received a response (404 Not Found in the case of the AuTest) from the origin for `Accept-Encoding: deflate`, decides to instead serve the 200 OK response. `Accept-Encoding: deflate` does not match `Accept-Encoding: ` by the rules in RFC 9111 (which encompasses the rules ATS uses for normalizing the header). Is it acceptable for a proxy to reply from cache with a response that is different than the origin's response to the request, which it also knows about? The server is not constrained by `Vary`, so it may choose which encoding to send. The proxy is bound to obey the `Vary` header and therefore does not have the flexibility that the origin server has.

This has been fixed in the case 1 deflate test in this PR. I am not sure why the similar tests for cases 2, 3, and 4 still pass.

JakeChampion:
@JosiahWI I believe cases 2,3, and 4 are also correct because we are normalising the request's accept-encoding header, and deflate is not one that we keep during that process. perhaps we need to update the autests to assert that no server-response is to be expected at all because it is a cache hit?

JosiahWI:
This comment from #7458 seems to suggest that the normalize_ae option is not intended to affect the selection of an alternate from cache: https://github.com/apache/trafficserver/issues/7458#issuecomment-1009516940. I wonder if there are any other rules for normalization that do affect alternate selection from cache...

JosiahWI:
@JakeChampion You are right, I had missed that ATS cleared the `Accept-Encoding: ` field in some cases with `normalize_ae` set to a nonzero value. In those cases, it is correct to match the previous `Accept-Encoding: ` response, as you said.

Assuming that `normalize_ae` does only affect the request to origin and not which alternate is served from cache, is it correct to serve the cached document after getting the 404 from the origin? My intuition says that the 404 should replace the existing 200 in cache, but I haven't checked the RFC, so I may be wrong about that.

Ideally, I think it would be good to serve directly from cache and skip the request to origin (unless the response is stale), as you suggested. (In other words, the `Accept-Encoding` header _should_ be normalized before looking for alternates, not after).

tests/gold_tests/headers/normalized_ae_match_vary_cache.test.py
```diff
@@ -31,7 +31,43 @@
...
+    "uuid: 38", "Verify br;q=1.1 request (uuid 38) is a cache hit and doesn't reach the server.")
+server.Streams.stdout += Testers.ExcludesExpression(
+    "uuid: 39", "Verify br,gzip;q=0.8 request (uuid 39) is a cache hit and doesn't reach the server.")
+
+ts = Test.MakeATSProcess("ts", enable_cache=True, select_ports=False)
+# Hardcode port to avoid port binding issues
+ts.Variables.port = 8765
+ts.Disk.records_config.update({
+    'proxy.config.http.server_ports': f'{ts.Variables.port}',
+})
```
JosiahWI: I'm curious why this was necessary.

JosiahWI (approved):
I don't know if we needed quite that many assertions to be convinced, but it's certainly very convincing. :)

This looks good. Maybe @bneradt could double check the port thing since he is more familiar with AuTests, but I'm comfortable putting this in master.

JakeChampion:
Apologies @JosiahWI - I did a rebase to tidy up the commits and it looks to have lost your approval

tests/gold_tests/headers/normalized_ae_match_vary_cache.test.py
JakeChampion: for some reason this autest for me would exit before the server started up - it looked to be due to a port conflict, which i solved with this hard-coding, this wasn't meant to be pushed up though, it was meant to be local only for me, apologies

JosiahWI:
I'll wait for CI to pass (it looked like there were problems that may have been related to port selection in some AuTests. 🤔) and approve again. We do squash on merge which will tidy up the commit history.

tests/gold_tests/headers/normalized_ae_match_vary_cache.test.py
JosiahWI: Bummer. These local AuTest issues can greatly impede ease of development but are not always easy to debug.

JakeChampion:
> I'll wait for CI to pass (it looked like there were problems that may have been related to port selection in some AuTests. 🤔) and approve again. We do squash on merge which will tidy up the commit history.

Looks like it all passed on ci now, yay. I think one of the next things I will look into is the flakey port selection in the autests because I do see that very often when running them locally as well.

JosiahWI (approved):

maskit:
Looks like you guys figured it out.

There is cryptic code that was added when the setting value 2 was introduced (70815db3b237d5e572ff9bc0a4df81d64157ba92). I bet the line fixed on this PR was supposed to be checked in a similar way on the commit, but it's too cryptic IMO. I like the explicit comparison better 👍 
https://github.com/apache/trafficserver/blob/577cbed448295a263e7911c7baadabbed6f927de/src/iocore/cache/HttpTransactCache.cc#L313-L332

Issue #12605, opened 2025-10-23
Support  the creation of  a mutually exclusive groups for `ArgParser`


brbzull0:
With the change, `ArgParser` will make sure that only one of the arguments in the mutually exclusive group was present on the command line.

Right now you can have two command options that should be exclusively and handle them internally by checking one after another, the idea would be to have something like we have in [python](https://docs.python.org/3/library/argparse.html#mutual-exclusion).

A clear example of this would be:

```
server_command.add_command("drain", "Drain the requests", [&]() { command->execute(); })
  .add_example_usage("traffic_ctl server drain [OPTIONS]")
  .add_option("--no-new-connection", "-N", "Wait for new connections...")
  .add_option("--undo", "-U", "Recover server from the drain mode");
```

They should be either one or the other, but there seems no way to express that atm.

PR #12621, opened 2025-10-28
ArgParser: Add mutually exclusive option groups

brbzull0:
Implement support for mutually exclusive option groups in ArgParser, allowing options that cannot be used together to be properly validated and documented.

This feature is useful for conflicting options like `--verbose/--quiet` or `--enable/--disable`, where only one option from a group should be specified at a time.

fixes https://github.com/apache/trafficserver/issues/12605

### DIFF
--- a/include/tscore/ArgParser.h
+++ b/include/tscore/ArgParser.h
@@ -316,7 +316,7 @@ protected:
   static void set_test_mode(bool test = true);
 
   // When true, do_exit() throws instead of calling exit()
-  static thread_local bool _test_mode;
+  static bool _test_mode;
 
   // Converted from 'const char **argv' for the use of parsing and help
   AP_StrVec _argv;
@@ -327,7 +327,6 @@ protected:
 
   friend class Command;
   friend class Arguments;
-  friend class TestArgParser;
 };
 
 } // namespace ts
--- a/src/tscore/ArgParser.cc
+++ b/src/tscore/ArgParser.cc
@@ -42,7 +42,7 @@ int usage_return_code = EX_USAGE;
 
 namespace ts
 {
-thread_local bool ArgParser::_test_mode = false;
+bool ArgParser::_test_mode = false;
 
 ArgParser::ArgParser() {}
 
@@ -336,13 +336,8 @@ ArgParser::Command::add_option_to_group(std::string const &group_name, std::stri
 
   auto it_mutex_group = _mutex_groups.find(group_name);
   if (it_mutex_group == _mutex_groups.end()) {
-    // Auto-create group if it doesn't exist. No need to call add_mutex_group first.
-    auto ret = _mutex_groups.emplace(group_name, MutexGroup(group_name, false /** required */));
-    if (!ret.second) {
-      std::cerr << "Error: Failed to auto-create mutex group '" << group_name << "'" << std::endl;
-      exit(1);
-    }
-    it_mutex_group = ret.first;
+    std::cerr << "Error: Mutex group '" << group_name << "' not found" << std::endl;
+    ArgParser::do_exit(1);
   }
 
   // Add the option normally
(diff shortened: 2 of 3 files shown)


doc/developer-guide/internal-libraries/ArgParser.en.rst
```diff
@@ -122,6 +122,38 @@ which is equivalent to
 
 In this case, `subinit` is the subcommand of `init` and `--initoption` is a switch of command `remove`.
 
+Mutually Exclusive Groups
```
brbzull0, line 125: AI generated.

brbzull0:
[approve autest 1]

brbzull0:
[approve ci autest 1]

bneradt:
[approve ci autest]

src/tscore/ArgParser.cc
```diff
@@ -287,6 +306,55 @@ ArgParser::Command::add_option(std::string const &long_option, std::string const
...
+    ArgParser::do_exit(1);
+  }
+
+  auto it_mutex_group = _mutex_groups.find(group_name);
+  if (it_mutex_group == _mutex_groups.end()) {
+    // Auto-create group if it doesn't exist. No need to call add_mutex_group first.
+    auto ret = _mutex_groups.emplace(group_name, MutexGroup(group_name, false /** required */));
+    if (!ret.second) {
+      std::cerr << "Error: Failed to auto-create mutex group '" << group_name << "'" << std::endl;
+      exit(1);
```
cmcfarlen: do_exit

include/tscore/ArgParser.h
```diff
@@ -262,6 +310,14 @@ class ArgParser
   void add_description(std::string const &descr);
 
 protected:
+  // Exit handler - throws in test mode, calls exit() otherwise
+  static void do_exit(int code);
+  // Enable test mode - makes do_exit() throw instead of calling exit() for unit testing
+  static void set_test_mode(bool test = true);
+
+  // When true, do_exit() throws instead of calling exit()
+  static thread_local bool _test_mode;
```
cmcfarlen: Does this need to be thread_local? Could it not just be a member variable on this instance?

include/tscore/ArgParser.h
```diff
@@ -271,6 +327,7 @@ class ArgParser
 
   friend class Command;
   friend class Arguments;
+  friend class TestArgParser;
```
cmcfarlen: I don't think this needs to be a friend to work

src/tscore/ArgParser.cc
```diff
@@ -287,6 +306,55 @@ ArgParser::Command::add_option(std::string const &long_option, std::string const
...
+                                        unsigned arg_num, std::string const &default_value, std::string const &key)
+{
+  if (group_name.empty()) {
+    std::cerr << "Error: Mutex group name cannot be empty" << std::endl;
+    ArgParser::do_exit(1);
+  }
+
+  auto it_mutex_group = _mutex_groups.find(group_name);
+  if (it_mutex_group == _mutex_groups.end()) {
+    // Auto-create group if it doesn't exist. No need to call add_mutex_group first.
```
cmcfarlen: Is this a good idea? I'd rather just fail and have the programmer create the group.

include/tscore/ArgParser.h
brbzull0: you are right!

src/tscore/ArgParser.cc
brbzull0: yeah, I wasn't very sure about having an error here, but seems not right to hide this from the proprammer I guess,  let's have an error instead.

include/tscore/ArgParser.h
brbzull0: I was under the impression that catch2 can run tests in parallel, was trying to prevent that, but looking around seems it's not(in this case at least). I'll remove the thread_local.
Also, I thought about as member of `ArgParser`, but then I would need to pass an instance(ptr) of `ArgParser` to the `Command` class as most(if not all) of the `do_exit` calls are made from that(`Command`) class.

cmcfarlen (approved):
Nice one.  Thanks!

PR #12626, opened 2025-10-30
Cleanup: Remove unused handlers of HttpSM

masaori335:
Two HostDB related handers are not used.
```
int state_pre_resolve(int event, void *data);
int state_mark_os_down(int event, void *data);
```

This handler is declared but doesn't have definition.
```
int state_auth_callback(int event, void *data);
```


JosiahWI (approved):
Straightforward cleanup.

Issue #12543, opened 2025-10-06
AUTEST - Improve JSONRPC  output tests


brbzull0:
Find a way to improve how we validate expected output from the jsonrpc. 
Having a static output check may force devs to update unrelated tests to what they are actually working on.

PR #12627, opened 2025-10-30
Test: `traffic_ctl_output_test*`: Add validate functions for specific JSON fields validation

brbzull0:
Add validate_json_contains() to validate specific JSON fields without exact output matches. Updated traffic_ctl_server_output.test.py to use the new method for server status checks, making tests more maintainable.

Also re-order some functions.

fixes https://github.com/apache/trafficserver/issues/12543

With the valuable help of Claude

Error logs would look something like this:

```
# stream.all.txt
FAIL: is_ssl_handshaking_stopped = false (expected true)

```


masaori335 (approved):
Seems very useful!

PR #12629, opened 2025-10-31
records: Accept `const char*` in `RecSetRecordString` to avoid const_cast.

brbzull0:
Didn't like how the cast was used in the caller, so now the cast is  hidden in one place (implementation detail).

API changed but it's just internal .


cmcfarlen (approved):
I've run into this annoyance as well.  Thanks for cleaning it up!

PR #12634, opened 2025-11-03
stats_over_http: fix incorrect use of C string functions.

vuori:
Handling of the `Accept-Encoding` field in the `stats_over_http` plugin used `strstr()` on the result of `TSMimeHdrFieldValueStringGet` leading to read of uninitialized memory. As a workaround, make a null-terminated copy of the header string.

As a matter of opinion, I think the compression support in `stats_over_http` should be completely removed as it significantly complicates the code for questionable benefit. Ideally users who care about this could stack the `compress` plugin, though I'm not clear on whether this works with intercept?

### DIFF
--- a/plugins/stats_over_http/stats_over_http.cc
+++ b/plugins/stats_over_http/stats_over_http.cc
@@ -804,7 +804,17 @@ stats_origin(TSCont contp, TSEvent /* event ATS_UNUSED */, void *edata)
   my_state->encoding    = encoding_format_t::NONE;
   if (accept_encoding_field != TS_NULL_MLOC) {
     int         len = -1;
-    const char *str = TSMimeHdrFieldValueStringGet(reqp, hdr_loc, accept_encoding_field, -1, &len);
+    const char *mime_str = TSMimeHdrFieldValueStringGet(reqp, hdr_loc, accept_encoding_field, -1, &len);
+    char       *str = nullptr;
+
+    // Ensure str is null-terminated for strstr()
+    if (len > 0) {
+      str = TSstrndup(mime_str, len);
+      if (str == nullptr) {
+        len = -1;
+      }
+    }
+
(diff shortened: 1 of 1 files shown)


plugins/stats_over_http/stats_over_http.cc
```diff
@@ -804,7 +804,17 @@ stats_origin(TSCont contp, TSEvent /* event ATS_UNUSED */, void *edata)
   my_state->encoding    = encoding_format_t::NONE;
   if (accept_encoding_field != TS_NULL_MLOC) {
     int         len = -1;
-    const char *str = TSMimeHdrFieldValueStringGet(reqp, hdr_loc, accept_encoding_field, -1, &len);
+    const char *mime_str = TSMimeHdrFieldValueStringGet(reqp, hdr_loc, accept_encoding_field, -1, &len);
+    char       *str = nullptr;
+
+    // Ensure str is null-terminated for strstr()
+    if (len > 0) {
+      str = TSstrndup(mime_str, len);
```
cmcfarlen, line 812: Instead of this, could you create a std::string_view from `{mime_str, Len}` and use `std::string_view::find` to search for the string?

vuori: ugghhh I'll think about it. Not very enthusiastic at all about all this code that just duplicates `compress`, the only reason I'm touching this is to make a thing that insists on sending `Accept-Encoding: identity` work reliably.

ezelkow1:
[approve ci]

plugins/stats_over_http/stats_over_http.cc
bneradt: I like the std::string_view suggestion too. It would be nice to avoid the cost of the allocation and the need to free.

Thank you for working on this.

bneradt: Thank you for the patch, @vuori . I'm picking up the work here:
https://github.com/apache/trafficserver/pull/12722

PR #12636, opened 2025-11-04
records:  Replace record registration macros with non macro solution.

brbzull0:
This has been in my radar for quite some time as while doing some work I came across this macros, hard to follow and easy to make a mistake, so I replaced with a more type safe/eary to read solution. 

- Replaces `REC_REGISTER_STAT_XXX` and `REC_REGISTER_CONFIG_XXX` macros with template implementations using tag dispatch to properly distinguish RecInt from RecCounter (both int64_t aliases).
- Remove unused unit test file.
- Added basic API unit test in a new file.


moonchen (approved):
Looks great.  Thanks for untangling those macros! The new unit coverage is nice.

PR #12638, opened 2025-11-04
Fix DNS error logging to use ERR_DNS_FAIL

bneradt:
Fixed a pointer comparison bug where the error_body_type string literal was compared by pointer address instead of value. This now passes the Dns_error_body constant directly so the pointer comparison succeeds and DNS errors are correctly logged as ERR_DNS_FAIL.

---

# For Review

* Dns_error_body declared here: https://github.com/apache/trafficserver/blob/0d4f77c0d250ddbadf39a63cfe434b13cf164644/src/proxy/http/HttpTransact.cc#L59
* Used here: https://github.com/apache/trafficserver/blob/0d4f77c0d250ddbadf39a63cfe434b13cf164644/src/proxy/http/HttpTransact.cc#L8058

Note that in the comparison referenced above, it was a pointer comparison rather than a string comparison. Thus it would always fail since the two always pointed to two different string literals. This change makes the pointer comparison work as expected.


masaori335 (approved):
Nice catch!

bneradt:
For 10.1.x:
https://github.com/apache/trafficserver/pull/12656

PR #12639, opened 2025-11-04
Parallel dir entry sync options

zwoop:
The idea here is to allow a configurable amount of directory synchronization to disk, where -1 means "as much parallelism as there are drives". This does have the downside of consuming additional memory during the sync phase.

The default is no parallelization still (same behavior as before). This does however move the sync continuation to the ET_TASK threads, even though most of their work is done on the AIO threads (but feels more correct to move these away from ET_NET).

### DIFF
--- a/src/iocore/cache/P_CacheDir.h
+++ b/src/iocore/cache/P_CacheDir.h
@@ -28,6 +28,7 @@
 #include "iocore/eventsystem/Continuation.h"
 #include "iocore/aio/AIO.h"
 #include "tscore/Version.h"
+#include "tscore/hugepages.h"
 
 #include <cstdint>
 #include <ctime>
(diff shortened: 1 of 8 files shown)


zwoop:
Behavior with the setting set to "1":

```
[Nov  4 17:38:31.792] [ET_NET 1] DIAG: <CacheDir.cc:888 (dir_sync_init)> (dir_sync) Disk /tmp/CACHE2/cache.db: 4 stripe(s) assigned to task 0
[Nov  4 17:38:31.792] [ET_NET 1] DIAG: <CacheDir.cc:888 (dir_sync_init)> (dir_sync) Disk /tmp/CACHE/cache.db: 4 stripe(s) assigned to task 0
```

Behavior when set to -1:

```
[Nov  4 17:39:56.977] [ET_NET 1] DIAG: <CacheDir.cc:888 (dir_sync_init)> (dir_sync) Disk /tmp/CACHE/cache.db: 4 stripe(s) assigned to task 0
[Nov  4 17:39:56.977] [ET_NET 1] DIAG: <CacheDir.cc:888 (dir_sync_init)> (dir_sync) Disk /tmp/CACHE2/cache.db: 4 stripe(s) assigned to task 1
```

src/iocore/cache/P_CacheDir.h
```diff
@@ -39,6 +40,7 @@ struct CacheVC;
 class CacheEvacuateDocVC;
 
 // #define LOOP_CHECK_MODE 1
```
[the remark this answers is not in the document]
zwoop: This is inline with existing code semantics.

src/records/RecordsConfig.cc
```diff
@@ -853,6 +853,8 @@ static constexpr RecordElement RecordsConfig[] =
   ,
   {RECT_CONFIG, "proxy.config.cache.dir.sync_max_write", RECD_INT, "2097152", RECU_DYNAMIC, RR_NULL, RECC_NULL, nullptr, RECA_NULL}
   ,
+  {RECT_CONFIG, "proxy.config.cache.dir.sync_parallel_tasks", RECD_INT, "1", RECU_DYNAMIC, RR_NULL, RECC_NULL, nullptr, RECA_NULL}
```
bneradt: `RECU_DYNAMIC` means reloadable. Is that somehow the case?

zwoop: No it's not. Fixed.

bneradt (approved):

zwoop:
[approve ci]

PR #12640, opened 2025-11-05
remap_reload.test.py: fix override warnings

bneradt:
remap_reload.test.py updates the remap.config file and issues a `traffic_ctl config reload` to dynamically integrate those changes into the running ATS process. Before this patch, this was done via creating a new tr.Disk.File() object and writing to it. But this resulted in warnings for overriding the already existing remap.config Disk.File from the setup of the ATS process in the first place. This transitions the test to modify the file via a Setup function instead. This achieves the same result without alarming the autest framework about the conflicting Disk.File object.

This patch cleans up these warnings:

```
╰─➤  ./autest.sh --sandbox /tmp/sb --clean=none -f remap_reload
Running Test remap_reload:Warning: Overriding file object /tmp/sb/remap_reload/ts/config/remap.config
Warning: Overriding file object /tmp/sb/remap_reload/ts/config/remap.config
........ Passed
``

### DIFF
--- a/tests/gold_tests/remap/remap_reload.test.py
+++ b/tests/gold_tests/remap/remap_reload.test.py
@@ -15,6 +15,20 @@
 #  limitations under the License.
 import os
 
+
+def update_remap_config(path: str, lines: list) -> None:
+    """Update the remap.config file.
+
+    This is used to update the config file between test runs without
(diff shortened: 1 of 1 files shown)


cmcfarlen:

PR #12642, opened 2025-11-05
vscode: associate .test.ext files with Python

bneradt:
This adds a .vscode/settings.json file to make vscode aware that .test.ext files are Python files.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+    "files.associations": {
+        "*.test.ext": "python"
+    }
+}
+


.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "files.associations": {
+        "*.test.ext": "python"
```
masaori335, line 3: This is helpful. However, why we can't use `.py` for these python files?

bneradt: Good question. `test.ext` is currently what the autest framework expects for these extensions. Maybe we can update it to also accept `.ext.py` or something?

masaori335: I'm +1 to changing them to `.ext.py`

cmcfarlen:

PR #12649, opened 2025-11-06
autest: --thread-limit 1 by default for Proxy Verifier

bneradt:
For production traffic simulation, Proxy Verifier by default runs
through the specified connections in parallel. But for our end to end
tests, that hardly ever makes sense. Rather we want the tests to
deterministically run through each specified connection sequentially.
This patch does so by passing --thread-limit 1 as a default option to
the Proxy Verifier client in the client extension. Many tests were
specifying this already anyway, so this saves some burden on client
configuration for each of the test descriptions.


cmcfarlen:

PR #12653, opened 2025-11-10
Add percentage weight to server_groups_list

jasmine-nahrain:
This adds the option to choose how often a server group is used in sni.

### DIFF
--- a/src/iocore/net/SNIActionPerformer.cc
+++ b/src/iocore/net/SNIActionPerformer.cc
@@ -521,6 +521,10 @@ ServerGroupsList::SNIAction(SSL &ssl, const Context & /* ctx ATS_UNUSED */) cons
   for (auto const &g : server_groups_list) {
     total += g.weight;
   }
+  if (total < 1) {
+    Warning("server_groups_list has invalid weights (sum <= 0)");
+    return SSL_TLSEXT_ERR_ALERT_WARNING;
+  }
 
   int         r         = random() % total;
   int         culmative = 0;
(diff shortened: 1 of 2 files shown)


doc/admin-guide/files/sni.yaml.en.rst
```diff
@@ -171,6 +171,37 @@ server_groups_list                       Inbound   Specifies an override to the
                                                    :file:`records.yaml` configuration. See the
                                                    `OpenSSL SSL_CTX_set_groups_list <https://docs.openssl.org/3.5/man3/SSL_CTX_set1_curves/>`_
                                                    documentation.
+                                                   
+                                                   Each item contains a group key identifying the server group name and optionally, 
+                                                   a percentage key that determines how frequently that group is selected when multiple 
+                                                   entries are present.
+                                                   
+                                                   Percentages are relative; they do not need to sum to 100. Any omitted percentage defaults to 100.
```
maskit: Maybe it's a ratio then?

maskit: Or maybe a weight as you say on the subject.

maskit:
I'm not too sure if this is a good approach. groups_list is probably not the only setting that users would want to configure like this. And some of those may not be controllable on sni.yaml.

I feel like making A/B test plugin is a better approach. I can't think of any reason to have the mechanism on ATS core.

maskit:
This is what I'm thinking. We'd need to introduce a new TS API to set group_list, but the idea is basically enabling plugins to do something on an SNI action. The difference from a global plugin that uses `TS_SSL_SERVERNAME_HOOK` would be that the plugin doesn't need to lookup server names. The plugin (`TSSNIDoAction`) will be called only if ATS core finds a plugin-action for the server name, like remap plugins are called only if there's a match.
https://github.com/apache/trafficserver/pull/12658

jasmine-nahrain:
[approve ci autest 1]

jasmine-nahrain:
The approach recommended in #12658 should be explored if picking this back up

PR #12658, opened 2025-11-11
Add support for SNI plugin

maskit:
This enables writing SNI actions as plugins.

The implementation is experimental. We'd have to introduce new TS APIs to do various things on the plugins, but that would be next step.

To use SNI plugins, you need to use `plugins` keyword on sni.yaml. For example:
```yaml
sni:
- fqdn: example.com
  http2: on
  plugins:
    - path: my_sni_action.so
      parameters: test
```

This is a draft. It does not have code for config reload and cleanup.


bryancall:
Can we use the SSL plugin interface instead of adding a new type of plugin?

maskit:
`TS_SSL_SERVERNAME_HOOK` could be used on global plugins, but it would require all those plugins to have own config files like sni.conf and also server name lookup which could potentially be done in a different manner (e.g. wildcard vs regex). In other words, it would be like writing a remap plugin as a global plugin and do all the remap work on the plugin. I don't think that makes sense.

maskit:
This PR was just to show an alternative approach, and I don't have plans to make this progress. Please feel free to take the unfinished code and finish it if you like the idea.

PR #12664, opened 2025-11-13
Store proxy-verifier in .git directory and fix AUTEST_OPTIONS escaping

bneradt:
- Store proxy-verifier in .git directory to share across build directories  
- Fix AUTEST_OPTIONS space escaping issue in cmake targets


cmcfarlen (commented):
I can understand downloading to the .git directory to save from multiple builds downloading the same file, but I'm not sure I understand why we wouldn't want to just download it at configure time if the AUTEST build option is specified.  This adds quite a bit of complexity to the autest parts of the build (arguably the most complicated part already).

bneradt:
> I can understand downloading to the .git directory to save from multiple builds downloading the same file, but I'm not sure I understand why we wouldn't want to just download it at configure time if the AUTEST build option is specified. This adds quite a bit of complexity to the autest parts of the build (arguably the most complicated part already).

Sounds good. I reverted that part so that proxy verifier is now fetched via the cmake config step again.

Thanks for the review!

cmcfarlen (approved):

cmcfarlen:

PR #12665, opened 2025-11-15
Fix negative_caching_lifetime being overridden by ttl-in-cache

bneradt:
When both negative_caching_lifetime and cache.config ttl-in-cache are configured, ttl-in-cache was incorrectly taking precedence for negative responses (404, 403, 500, etc.). This caused negative responses to be cached for the ttl-in-cache duration instead of the intended shorter negative_caching_lifetime. The fix modifies what_is_document_freshness() to skip the ttl-in-cache check for negatively cached responses, allowing them to use the Expires header set by negative_caching_lifetime instead.

---

# For Review

Configuring github to ignore whitespace is helpful in viewing this patch.


masaori335 (approved):
Looks good to me.

Issue #6261, opened 2019-12-13
Cannot log Transfer-Encoding:chunked from origin


ema:
I have tried logging the value of `Transfer-Encoding` as returned by an origin server. The log format string looks like this:

`'test: %<chi> - %<caun> [%<cqtn>] "%<cqtx>" %<pssc> %<pscl> BerespHeader:Transfer-Encoding:%<{Transfer-Encoding}ssh>'`

When the origin returns `Transfer-Encoding: gzip` things look fine:

`test: 127.0.0.1 - - [13/Dec/2019:10:58:05 +0100] "GET http://127.0.0.1:6533/te HTTP/1.1" 200 3 BerespHeader:Transfer-Encoding:gzip`

If instead `Transfer-Encoding` is `chunked`, the value mysteriously disappears:
`test: 127.0.0.1 - - [13/Dec/2019:10:58:25 +0100] "GET http://127.0.0.1:6533/te HTTP/1.1" 200 0 BerespHeader:Transfer-Encoding:-`

I have confirmed with the xdebug plugin that the origin does indeed send TE:chunked.
8.0.5 is affected.

PR #12667, opened 2025-11-17
Transfer-Encoding:chunked log field fix

bneradt:
When an origin server returns Transfer-Encoding:chunked, ATS removes
this header during response processing to handle dechunking. This
prevented the original header value from being logged via
%<{Transfer-Encoding}ssh>. This fix stores the original
Transfer-Encoding value before deletion and uses the stored value if
requested via the ssh field.

Fixes: #6261

### DIFF
--- a/include/proxy/http/HttpTransact.h
+++ b/include/proxy/http/HttpTransact.h
@@ -635,6 +635,7 @@ public:
     bool            trust_response_cl          = false;
     ResponseError_t response_error             = ResponseError_t::NO_RESPONSE_HEADER_ERROR;
     bool            extension_method           = false;
+    std::string     server_response_transfer_encoding; ///< Storage for logging.
 
     _HeaderInfo() {}
   };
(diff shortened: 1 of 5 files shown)


include/proxy/http/HttpTransact.h
```diff
@@ -635,6 +635,7 @@ class HttpTransact
     bool            trust_response_cl          = false;
     ResponseError_t response_error             = ResponseError_t::NO_RESPONSE_HEADER_ERROR;
     bool            extension_method           = false;
+    std::string     server_response_transfer_encoding; ///< Storage for logging.
```
masaori335, line 638: IIUC, `current.server->transfer_encoding` stores the type of Transfer-Encoding. De we really need copy of the header here?

bneradt: Thanks for the review. 

current.server->transfer_encoding is an enum over these values:

```
  enum class TransferEncoding_t {
    NONE = 0,
    CHUNKED,
    DEFLATE,
  };
```

That will not allow us to fully replicate the original string, which might contain "deflate", "chunked", or the like and any combination of them. The ops users of this field want to know the full string sent from the origin, but we delete the field entirely. Thus it is always currently empty. This PR stores the origin resonse's string for logging.

masaori335: Makes sense. Thank you for explanation.

PR #12669, opened 2025-11-18
Virtualhost config with remap rules

serrislew:
Initial POC to support virtualhost config schema with remap rules. 
- Resolves to a single virtualhost entry
   - Remap rule logic remains the same. Order ranking still applies (i.e. rules defined first have priority) 
   - Virtualhost remap rules have priority. If no remap rule matches or virtualhost is found, existing remap.config is then considered
- Virtualhost domains support exact and wildcard (only single left-most "*") matching
- Supports traffic_ctl config reload of single virtualhost entry
```
$ traffic_ctl config reload --virtualhost foo
┌ Virtualhost: foo
└┬ Reload status: ok
 ├ Message: Virtualhost successfully reloaded
```

### DIFF
--- a/src/mgmt/rpc/handlers/config/Configuration.cc
+++ b/src/mgmt/rpc/handlers/config/Configuration.cc
@@ -225,7 +225,12 @@ reload_virtualhost_config(std::string_view const & /* id ATS_UNUSED */, YAML::No
   }
 
   VirtualHost::scoped_config vhost_config;
-  auto                       vhost_entry = vhost_config->find_by_id(name);
+  if (!vhost_config) {
+    resp.note("Virtualhost not initialized");
+    return resp;
+  }
+
+  auto vhost_entry = vhost_config->find_by_id(name);
   if (vhost_entry) {
     resp.result()["virtualhost"] = name;
     resp.result()["status"]      = "ok";
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -4432,7 +4432,7 @@ HttpSM::set_virtualhost_entry(std::string_view domain)
     return;
   }
 
-  auto vhost_entry = vhost_config->find_by_domain(std::string{domain});
+  auto vhost_entry = vhost_config->find_by_domain(domain);
   if (vhost_entry) {
     SMDbg(dbg_ctl_url_rewrite, "Found virtualhost: %s", vhost_entry->get_id().c_str());
     // Explicitly acquire() since HttpSM holds raw pointer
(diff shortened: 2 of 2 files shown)


bneradt:
[approve ci centos]

bneradt:
[approve ci centos]

bneradt:
[approve ci centos]

zwoop:
First, I really like the concepts of virtual host configurations, and eliminating the (IMO weird) ordering requirements that remap.config has today. That's great!

I have some concerns about the path forward here, and how this essentially encapsulates the existing "map" directives as strings, rather than YAML'fying them properly. What I have had in mind is something more inline with the below (just an example, not a final proposal), where the top-level selector is a FQDN, and under there we have sub-selectors. Initially, only PATH is such a sub-selector, but one could in the future easily see adding sub selectors for e.g. header values, or request method (GET vs POST) etc. 

Key here is to avoid putting us in a situation where the YAML has to dramatically change in future version. Essentially, future proof the virtual host YAML, such that adding features later (such as method sub-selectors) does not break existing configurations.

Here's an example of how a currently supported remap.config could be moved to a new YAML format, putting all the path matches into one single virtual host YAML config. (Note that the format here allows multiple top level host selectors, I just have one here).

```
remap:
  - maps:
      scheme: http
      host: example.com
    paths:
      - path: /foo
        origin:
          scheme: http
          host: foo.example.com
          path: /

      - path: /bar
        origin
          scheme: http
          host: bar.example.com
          path: /some_special_path
        plugins:
          - name: header_rewrite.so
            parameters:
              - "/etc/trafficserver/foo.hrw"
          - name: cachekey.so
            parameters:
              - "/etc/trafficserver/foo.cachekey

```

Again, this is not a final proposal / schema, just thinking out loud.

serrislew:
@zwoop I see your point. Since this is an extension of our existing configs, we have the flexibility to change the schema as we want. Though we want to keep improving and modernizing our configs, we should start with a good foundation that allows us to build features more smoothly. I still think our v0 should not add any new features, we can build on those with future remap versions. 

I also think that the domains node might still be good to have so it's clear exactly which domains are resolving to the virtualhost config. As we introduce other configs (i.e. sni.yaml domains), it might not be as clear to the user that it has to resolve across all the host selectors of each config node. We can add warnings if host selectors don't match the initial domain section. We can discuss this more later. 

Nonetheless, I think your concern is valid. I will look into YAML'fying the remap node but keeping the logic changes as minimal as possible, since the virtualhost schema already introduces a lot of new changes. Thank you for your feedback!

PR #12670, opened 2025-11-18
Ignore malformed Cache-Control directives per RFC 7234

### LINKED ISSUE #12029: Stale negative responses returned under unknown conditions
I have been seeing stale negative getting stuck in cache on ATS 10.0.2 and 10.0.3, but have been unable to reproduce this under controlled conditions. The problem has only been seen under relatively high loads, but it _might_ be that the issue occurs more often if the response is cached due to an internal request (from a Lua fetch call). Whatever the cause, the cached negative response is returned for all requests. Usually the cached response disappears after restarting ATS.

Sample response:

```
< HTTP/1.1 400 Bad Request
< Cache-Control: public; max-age=30
< Date: Mon, 10 Feb 2025 11:40:36 GMT
< Content-Length: 0
< Age: 866
```

This is a relatively short-lived instance, sometimes the times go up to tens of hours for such response.

The negative caching here is prompted only by the `Cache-Control` header. There are no `cache.config` pins for the origin so this shouldn't be https://github.com/apache/trafficserver/issues/11854 and `negative_caching_enabled` is set to 0. The above response was from a forward-proxied request with no remap entry, but the problem also occurs for remapped reverse-proxy requests.

I have unfortunately not been able to gather debug logs for this, since it has only been seen on busy servers and seems to depend on in-memory state.

bneradt:
Detect and ignore Cache-Control directives that are malformed (e.g.,
using semicolons instead of commas). When the parser cannot fully
consume a Cache-Control directive value, leaving unparsed non-whitespace
content, the directive is treated as unrecognized and ignored per RFC
7234 Section 5.2.

For example, a malformed header such as 'public; max-age=30' (note the
incorrect semicolon separator) would previously set the 'public' bit but
ignore the unparsed 'max-age=30', causing the response to be cached with
incorrect heuristic lifetimes. Not only is this incorrect per the RFC,
but the intended max-age was, from the user's perspective, mysteriously
ignored. Now the entire malformed directive is ignored, allowing default
caching rules to apply correctly.

Fixes: #12029

### DIFF
--- a/src/proxy/hdrs/MIME.cc
+++ b/src/proxy/hdrs/MIME.cc
@@ -3752,6 +3752,9 @@ MIMEHdrImpl::recompute_cooked_stuff(MIMEField *changing_field_or_null)
 
         for (s = csv_iter.get_first(field, &len); s != nullptr; s = csv_iter.get_next(&len)) {
           e = s + len;
+          // Store set mask bits from this CSV value so we can clear them if needed.
+          uint32_t csv_value_mask = 0;
+
           for (c = s; (c < e) && (ParseRules::is_token(*c)); c++) {
             ;
           }
@@ -3766,6 +3769,7 @@ MIMEHdrImpl::recompute_cooked_stuff(MIMEField *changing_field_or_null)
             HdrTokenHeapPrefix *p                  = hdrtoken_wks_to_prefix(token_wks);
             mask                                   = p->wks_type_specific.u.cache_control.cc_mask;
             m_cooked_stuff.m_cache_control.m_mask |= mask;
+            csv_value_mask                        |= mask;
 
 #if TRACK_COOKING
             Dbg(dbg_ctl_http, "                        set mask 0x%0X", mask);
@@ -3795,6 +3799,26 @@ MIMEHdrImpl::recompute_cooked_stuff(MIMEField *changing_field_or_null)
                 if (token_wks == MIME_VALUE_MAX_STALE.c_str()) {
                   m_cooked_stuff.m_cache_control.m_secs_max_stale = INT_MAX;
                 }
+              }
+            }
+
+            // Detect whether there is any more non-whitespace content after the
+            // directive. This indicates an unrecognized or malformed directive.
+            // This can happen, for instance, if the host uses semicolons
+            // instead of commas as separators which is against RFC 7234 (see
+            // issue #12029). Regardless of the cause, this means we need to
+            // ignore the directive and clear any mask bits we set from it.
+            while (c < e && ParseRules::is_ws(*c)) {
+              ++c;
+            }
+            if (c < e) {
+              // There's non-whitespace content that wasn't parsed. This means
+              // that we cannot really understand what this directive is.
+              // Per RFC 7234 Section 5.2: "A cache MUST ignore unrecognized cache
+              // directives."
+              if (csv_value_mask != 0) {
+                // Reverse the mask that we set above.
+                m_cooked_stuff.m_cache_control.m_mask &= ~csv_value_mask;
               }
             }
           }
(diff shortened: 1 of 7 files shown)


mlibbey:
Not to prevent this one ... but is it worth doing a quick look at 
https://cache-tests.fyi/#cc-parse
 to see if there are other quick wins nearby?
For instance, I think the current web page is saying ATS isn't handling
- Does HTTP cache ignore max-age with space before the =?
- Does HTTP cache ignore max-age with space after the =?
- HTTP cache must not reuse a response with a single-quoted Cache-Control: max-age

bneradt:
> Not to prevent this one ... but is it worth doing a quick look at https://cache-tests.fyi/#cc-parse to see if there are other quick wins nearby? For instance, I think the current web page is saying ATS isn't handling
> 
> * Does HTTP cache ignore max-age with space before the =?
> * Does HTTP cache ignore max-age with space after the =?
> * HTTP cache must not reuse a response with a single-quoted Cache-Control: max-age

Thank you @mlibbey . I updated the code for these conditions as well and beefed up the tests for it too.

zwoop (approved):
Looks good. Once landed, please do a cache-test comparison before / after this PR (compare master to 10.1 I guess?) and see that we only have fixes and no new defects.

src/proxy/hdrs/MIME.cc
```diff
@@ -3774,27 +3778,72 @@ MIMEHdrImpl::recompute_cooked_stuff(MIMEField *changing_field_or_null)
...
+              } else {
+                // No '=' found, or whitespace before '='. This is malformed.
+                // For directives that require values, this is an error.
+                // Clear the mask for this directive.
+                csv_value_mask                         = 0;
+                m_cooked_stuff.m_cache_control.m_mask &= ~mask;
+              }
+            }
+
+            // Detect whether there is any more non-whitespace content after the
```
zwoop, line 3830: This feels little heavy on the comments, but oh well.

bneradt:
> Looks good. Once landed, please do a cache-test comparison before / after this PR (compare master to 10.1 I guess?) and see that we only have fixes and no new defects.

Good point. I'm working on this before merging instead of after to avoid surprises. This is what I found:

## Test Fixed (Expected Improvements)

| Test | Description | Before | After |
|------|-------------|--------|-------|
| `freshness-max-age-single-quoted` | `max-age='100'` (single-quoted) | FAIL - incorrectly cached | PASS |

This was a cache-test failure which was red and is now green. As such, we now align more closely with the RFC and have another passing cache-test result.

Now that we are more strict on the RFC compliance, there are other changes though:

## Other Test Changes from Increased Strictness

| Test | Description | Before | After | Notes |
|------|-------------|--------|-------|-------|
| `freshness-max-age-space-after-equals` | `max-age= 100` (space after `=`) | cached | not cached |
| `freshness-max-age-space-before-equals` | `max-age =100` (space before `=`) | cached | not cached |
| `freshness-max-age-100a` | `max-age=100a` | cached |not cached | Trailing non-digit now causes directive to be ignored |
| `freshness-max-age-a100` | `max-age=a100` | cached | not cached | Leading non-digit now causes directive to be ignored |
| `freshness-max-age-decimal-five` | `max-age=100.5` | cached | not cached | Decimal values now treated as malformed |
| `freshness-max-age-decimal-zero` | `max-age=100.0` |cached | not cached | Decimal values now treated as malformed |
| `freshness-max-age-quoted` | `max-age="100"` (double-quoted) |cached | not cached | Double-quoted values now treated as malformed |

Observe that each of these are considered "grey" areas in the table:
https://cache-tests.fyi/#cc-parse

For reference, max-age should be `delta-seconds` which is just digits and explicitly should not be quoted. So these examples all violate the RFC:

https://www.rfc-editor.org/rfc/rfc9111.html#name-max-age

https://www.rfc-editor.org/rfc/rfc9111.html#section-1.2.2:

> delta-seconds  = 1*DIGIT

## Conclusion

This patch fixes a cache-test, but does make our max-age more strict in other regards too. Therefore, some things that were cached before, such as `max-age=a100`, will no longer be cached. I believe that is fine, and the cache-tests themselves seem to imply this by making the results grey. As a point of comparison, this patch makes our behavior more like Chrome's for each of these cases, aside from the "space-after-equals" test which we reject and they cache.

bneradt:
After talking in our PR/issue meeting, I think the safest thing to do is to push this out to 11-Dev. The fix for this isn't urgent for anyone, so targeting 11-Dev seems like the prudent approach since it will potentially change cache behavior by making the parsing more strict.

PR #12672, opened 2025-11-18
cookie_remap: set_sendto_headers

bneradt:
This adds a new set_sendto_headers configuration option that allows setting arbitrary HTTP headers (including the Host header) on requests to sendto destinations when cookie rules match. Header values support dynamic substitution of regex capture groups ($1, $2, etc.) and path variables ($path, $cr_req_url, etc.). When the Host header is set via this option, pristine_host_hdr is automatically disabled for that transaction.

### DIFF
--- a/doc/admin-guide/plugins/cookie_remap.en.rst
+++ b/doc/admin-guide/plugins/cookie_remap.en.rst
@@ -42,6 +42,7 @@ Cookie Based Routing Inside TrafficServer Using cookie_remap
     * :ref:`else: url [optional] <else-url-optional>`
     * :ref:`connector: and <connector-and>`
     * :ref:`disable_pristine_host_hdr: true|false [optional] <disable-pristine-host-hdr>`
+    * :ref:`set_sendto_headers: [optional] <set-sendto-headers>`
 
   * :ref:`Reserved path expressions <reserved-path-expressions>`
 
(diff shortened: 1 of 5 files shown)


bryancall (changes requested):
Thanks for this feature - it's a useful addition for cookie-based routing scenarios!

Besides what Copilot stated:

**Request for refactoring:**

The regex capture group substitution logic (`$1`, `$2`, etc.) is now duplicated in two places:

1. `op::process()` for the sendto URL (lines ~745-765)
2. `TSRemapDoRemap()` for header values (lines ~1361-1385)

These two implementations have slightly different behavior (as Copilot noted, the header version has a bug with unmatched `$` characters, while the sendto version handles them correctly).

**Suggestion:** Extract this into a shared helper function, something like:

```cpp
void substitute_regex_captures(std::string &str, 
                               std::vector<std::string> const &matches, 
                               int ccount)
{
  std::string::size_type pos  = str.find('$');
  std::string::size_type ppos = 0;
  std::string result;
  
  while (pos != std::string::npos) {
    if (pos + 1 < str.length() && isdigit(str[pos + 1])) {
      int ix = str[pos + 1] - '0';
      if (ix <= ccount && ix < static_cast<int>(matches.size())) {
        result += str.substr(ppos, pos - ppos);
        result += matches[ix];
        ppos = pos + 2;
      }
    }
    pos = str.find('$', pos + 1);
  }
  result += str.substr(ppos);
  str = std::move(result);
}
```

This would:
- Fix the bug in header substitution (by using the same logic as sendto URL)
- Reduce code duplication  
- Make future maintenance easier
- Ensure consistent behavior between URL and header substitution

Once the shared function exists, both call sites can use it.

Also, I'd suggest adding a test case that includes a literal `$` in a header value (like `X-Price: $100`) to verify the fix works correctly.

Issue #9179, opened 2022-11-04
Canceled streams are not cleaned up in time


mrdg:
We run a proxy server in front of ATS and noticed this problem when the number of streams on a connections gets close to the maximum, and some streams are canceled by the proxy (generally because the original request was canceled).

I posted some test code and logs of a reproduction case in [this gist](https://gist.github.com/mrdg/240bb61a0df2b17b8cc9148c4d540ed9). I tested this on master with `proxy.config.http2.max_concurrent_streams_in` set to 5 to make it easier to parse the logs, but we see the issue with much higher limits as well. The test client sends 5 requests, then cancels those, and then sends 5 more, which should be accepted, but it appears that ATS is refusing them. The client code logs this:

```
2022/11/04 12:38:26 http2: Transport received RST_STREAM stream=11 len=4 ErrCode=REFUSED_STREAM
2022/11/04 12:38:26 http2: Transport received RST_STREAM stream=13 len=4 ErrCode=REFUSED_STREAM
2022/11/04 12:38:26 http2: Transport received RST_STREAM stream=15 len=4 ErrCode=REFUSED_STREAM
2022/11/04 12:38:26 http2: Transport received RST_STREAM stream=17 len=4 ErrCode=REFUSED_STREAM
2022/11/04 12:38:26 http2: Transport received RST_STREAM stream=19 len=4 ErrCode=REFUSED_STREAM
```

Looking at the ATS logs, I see that the previously canceled streams are deleted after the second batch of requests is rejected:

```
[Nov  4 12:38:26.285] [ET_NET 1] DEBUG: <Http2CommonSession.cc:250 (do_start_frame_read)> (http2_cs) [1] receiving frame header
[Nov  4 12:38:26.285] [ET_NET 1] DEBUG: <Http2CommonSession.cc:267 (do_start_frame_read)> (http2_cs) [1] frame header length=10, type=1, flags=0x5, streamid=11
[Nov  4 12:38:26.285] [ET_NET 1] DEBUG: <Http2ConnectionState.cc:232 (rcv_headers_frame)> (http2_con) [1] [11] Received HEADERS frame
[Nov  4 12:38:26.286] [ET_NET 1] DEBUG: <Http2ConnectionState.cc:1936 (send_rst_stream_frame)> (http2_con) [1] [11] Send RST_STREAM frame
...
[...]

PR #12673, opened 2025-11-19
Fix HTTP/2 canceled stream cleanup timing

bneradt:
When streams are canceled via RST_STREAM, the stream deletion now happens immediately before the next frame is processed. This prevents the race condition where new HEADERS frames are incorrectly refused because the stream count hasn't been decremented yet. Additionally, when streams are refused due to exceeding max_concurrent_streams, the header blocks are now properly decoded to maintain HPACK dynamic table synchronization per RFC 9113, and RST_STREAM with REFUSED_STREAM is sent instead of terminating the connection with COMPRESSION_ERROR.

Fixes: #9179

---

I have run this on a machine taking production traffic and it ran fine.

### DIFF
--- a/src/proxy/http2/Http2ConnectionState.cc
+++ b/src/proxy/http2/Http2ConnectionState.cc
@@ -345,14 +345,23 @@ Http2ConnectionState::rcv_headers_frame(const Http2Frame &frame)
       stream     = this->create_stream(stream_id, error);
       new_stream = true;
       if (!stream) {
-        // Terminate the connection with COMPRESSION_ERROR because we don't decompress the field block in this HEADERS frame.
-        // TODO: try to decompress to keep HPACK Dynamic Table in sync.
+        // Per RFC 9113, we must decode the header block even when refusing the stream to keep
+        // the HPACK dynamic table in sync. Create a temporary unregistered stream for decoding.
         if (error.cls == Http2ErrorClass::HTTP2_ERROR_CLASS_STREAM) {
-          return Http2Error(Http2ErrorClass::HTTP2_ERROR_CLASS_CONNECTION, Http2ErrorCode::HTTP2_ERROR_COMPRESSION_ERROR,
-                            error.msg);
+          uint32_t const initial_local_stream_window = this->acknowledged_local_settings.get(HTTP2_SETTINGS_INITIAL_WINDOW_SIZE);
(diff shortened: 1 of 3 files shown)


src/proxy/http2/Http2ConnectionState.cc
```diff
@@ -692,6 +704,11 @@ Http2ConnectionState::rcv_rst_stream_frame(const Http2Frame &frame)
     Http2StreamDebug(this->session, stream_id, "Parsed RST_STREAM frame: Error Code: %u", rst_stream.error_code);
     stream->set_rx_error_code({ProxyErrorClass::TXN, static_cast<uint32_t>(rst_stream.error_code)});
     stream->initiating_close();
+    // Delete the stream immediately to free up the stream count before the next frame is processed.
+    // This prevents the race condition where HEADERS frames are incorrectly refused due to the
+    // stream count not being decremented yet. The destructor will handle the case where the stream
+    // has already been removed from the stream list.
+    this->delete_stream(stream);
```
maskit, line 711: Shouldn't we call `THREAD_FREE(stream, http2StreamAllocator, this_ethread());` instead of calling these two, `initiating_close` and `delete_stream`?

src/proxy/http2/Http2ConnectionState.cc
```diff
@@ -345,14 +345,23 @@ Http2ConnectionState::rcv_headers_frame(const Http2Frame &frame)
...
+          stream = THREAD_ALLOC_INIT(http2StreamAllocator, this_ethread(), this->session->get_proxy_session(), stream_id,
+                                     this->peer_settings.get(HTTP2_SETTINGS_INITIAL_WINDOW_SIZE), initial_local_stream_window,
+                                     !STREAM_IS_REGISTERED);
+          if (!stream) {
+            // Failed to create even a temporary stream, this is a serious error.
+            return Http2Error(Http2ErrorClass::HTTP2_ERROR_CLASS_CONNECTION, Http2ErrorCode::HTTP2_ERROR_INTERNAL_ERROR,
+                              "failed to create temporary stream for header decoding");
+          }
+          free_stream_after_decoding  = true;
+          reset_header_after_decoding = true;
```
maskit, line 361: Why don't you delete the stream here and return `error` that probably contains `HTTP2_ERROR_REFUSED_STREAM`?

PR #12676, opened 2025-11-19
Require and call destructors on ClassAllocated objects

cmcfarlen:
This removes the flag to call the destructor on ClassAllocator objects.  ClassAllocator now requires that objects have a noexcept destructor that it can call on all objects.  This is much safer than the explicit flag and should not add any overhead for trivially destructed types.  I've added unit tests for this and also ran ASAN to make sure that the tests don't report memory leaks.

The unit tests were written by Claude, other code is my own.

### DIFF
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -251,11 +251,8 @@ HttpSM::get_server_connect_timeout()
 
 HttpSM::HttpSM() : Continuation(nullptr), vc_table(this) {}
 
-void
-HttpSM::cleanup()
+HttpSM::~HttpSM()
 {
-  t_state.destroy();
-  api_hooks.clear();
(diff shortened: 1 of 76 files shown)


cmcfarlen:
[approve ci ubuntu rocky]

src/proxy/http/HttpSM.cc
```diff
@@ -280,7 +277,6 @@ HttpSM::cleanup()
 void
 HttpSM::destroy()
 {
-  cleanup();
```
bneradt, line 283: Nice cleanup.

bneradt (approved):
So the big win here is that `criptContextAllocator` is now deallocated by default. That addresses the leak.

cmcfarlen:
Added to milestone 10.1.1 via #12720

Issue #9113, opened 2022-09-29
"Targeted HTTP Cache Control“ specs supports


scw00:
https://httpwg.org/specs/rfc9213.html

shukitchan:
Before implementing this into core, i think we can already do this with header rewrite or lua plugin

scw00:
Great.

pmbanugo:
Hi, I'm new here and exploring how I can use traffic server. is this something that won't get implemented in traffic server? 

I'm looking for a means to support that header since I'm not sure traffic server supports s-maxage

bneradt:
> Hi, I'm new here and exploring how I can use traffic server. is this something that won't get implemented in traffic server?
> 
> I'm looking for a means to support that header since I'm not sure traffic server supports s-maxage

Hi @pmbanugo. We do not support the Targeted Cache Control spec, but ATS does handle `s-maxage`. In order to verify this, I added some tests for the s-maxage feature and it passes those tests:

https://github.com/apache/trafficserver/pull/11601

PR #12679, opened 2025-11-20
Implement RFC 9213 Targeted HTTP Cache Control

bneradt:
This adds support for targeted Cache-Control headers (like CDN-Cache-Control) that allow cache directives to be targeted at specific caches. The implementation includes a configurable, priority-ordered list of targeted headers via proxy.config.http.cache.targeted_cache_control_headers, which is overridable per-remap rule. When a targeted header is present, it takes precedence over the standard Cache-Control header for caching decisions. Targeted headers are passed through downstream to allow proper cache hierarchy behavior.

Fixes: #9113

### DIFF
--- a/src/api/InkAPI.cc
+++ b/src/api/InkAPI.cc
@@ -7568,7 +7568,15 @@ TSHttpTxnConfigStringSet(TSHttpTxn txnp, TSOverridableConfigKey conf, const char
     break;
   case TS_CONFIG_HTTP_CACHE_TARGETED_CACHE_CONTROL_HEADERS:
     if (value && length > 0) {
-      auto &parsed                                            = ParsedConfigCache::lookup(conf, std::string_view(value, length));
+      auto &parsed = ParsedConfigCache::lookup(conf, std::string_view(value, length));
+      // This is intentionally a non-owning copy of the parsed representation.
+      // ParsedConfigCache::ParsedValue owns the backing string in conf_value_storage,
+      // and TargetedCacheControlHeaders stores string_view entries into that stable
+      // storage. The per-transaction override struct in HttpTransact::State is raw
+      // storage (not an owning/destructed HttpConfigParams object), so this path
+      // does not free conf_value. Reusing the cached parsed object avoids reparsing
+      // and avoids allocating/duplicating a second backing string on every txn
+      // override update, while preserving valid lifetimes for all string_view data.
       s->t_state.my_txn_conf().targeted_cache_control_headers = std::get<TargetedCacheControlHeaders>(parsed.parsed);
     } else {
       s->t_state.my_txn_conf().targeted_cache_control_headers = TargetedCacheControlHeaders{};


src/proxy/hdrs/MIME.cc
```diff
@@ -3723,13 +3724,34 @@ MIMEHdrImpl::recompute_cooked_stuff(MIMEField *changing_field_or_null)
...
+            break;
+          }
+        }
+      }
+    }
+
+    // If no targeted header was found, fall back to standard Cache-Control.
+    if (!field) {
+      field = mime_hdr_field_find(this, static_cast<std::string_view>(MIME_FIELD_CACHE_CONTROL));
+    }
```
bneradt, line 3749: This is the actualy interesting production change (part 1).

src/proxy/http/HttpSM.cc
```diff
@@ -2059,14 +2059,20 @@ HttpSM::state_read_server_response_header(int event, void *data)
...
       break;
     }
 
+    // Recompute cooked cache control with targeted headers (pass nullptr if not configured).
+    const char *targeted_headers =
+      (t_state.txn_conf->targeted_cache_control_headers && t_state.txn_conf->targeted_cache_control_headers[0] != '\0') ?
+        t_state.txn_conf->targeted_cache_control_headers :
+        nullptr;
+    t_state.hdr_info.server_response.m_mime->recompute_cooked_stuff(nullptr, targeted_headers);
+
```
bneradt: This is the other interesting production change (part 2).

This is "recooked" here to avoid intermingling the targeted cache control configuration into the MIME/parser interface. The recompute should be cheap, so I think this is OK. But if wanted, I can look into trying to get the parse to happen initially at parse_resp.

bneradt:
[approve ci autest]

JakeChampion:
@bneradt if you wanted to - you could run this patchset/branch against the tests in [cache-tests](https://github.com/http-tests/cache-tests) <https://cache-tests.fyi/#cdn-cache-control> and see how many ATS is now passing for ☺️

bneradt:
> I know we suck in the area of overridable string configurations. But if I understand this, it will essentially “parse” this config string on every request, to find the headers that may contain the CC value(s) for this request ?
> 
> I know it’s hard, but could we pre-parse this (on load and reload) such that we build WKS for the common headers at least? And for the rest, have an array of TextView’s into the “raw” config string ?

These are good observations. I think I addressed these both - but let me know if I misunderstood something. I added CDN-Cache-Control to the WKS list and configured custom parsing to the possible targeted header list so that the parse is done once and cached from there as a string_view list.

bneradt:
Hi @JakeChampion

> @bneradt if you wanted to - you could run this patchset/branch against the tests in [cache-tests](https://github.com/http-tests/cache-tests) https://cache-tests.fyi/#cdn-cache-control and see how many ATS is now passing for ☺️

I ran the tests and no tests change. Note that this patch adds targeted cache control header support, which I don't think the cache-tests have tests for.

But it's good to verify that no new tests fail at least. :)

JakeChampion:
> Hi @JakeChampion
> 
> > @bneradt if you wanted to - you could run this patchset/branch against the tests in [cache-tests](https://github.com/http-tests/cache-tests) https://cache-tests.fyi/#cdn-cache-control and see how many ATS is now passing for ☺️
> 
> I ran the tests and no tests change. Note that this patch adds targeted cache control header support, which I don't think the cache-tests have tests for.
> 
> But it's good to verify that no new tests fail at least. :)

The section I linked too includes cdn-cache-control which I think is what you are also adding in this patch set, did those tests begin to pass?

src/proxy/hdrs/MIME.cc
```diff
@@ -3725,13 +3726,26 @@ MIMEHdrImpl::recompute_cooked_stuff(MIMEField *changing_field_or_null)
...
+  /////////////////////////////////////////////////////////////////////////////
+  // (1) cook the Cache-Control header (or targeted variant) if present     //
+  /////////////////////////////////////////////////////////////////////////////
 
   // to be safe, recompute unless you know this call is for other cooked field
   if ((changing_field_or_null == nullptr) || (changing_field_or_null->m_wks_idx != MIME_WKSIDX_PRAGMA)) {
-    field = mime_hdr_field_find(this, static_cast<std::string_view>(MIME_FIELD_CACHE_CONTROL));
+    field = nullptr;
+
+    // Check for targeted cache control headers first (in priority order).
```
bneradt, line 3738: assert that targeted_headers is non-nullptr.

src/proxy/http/HttpConfig.cc
```diff
@@ -694,6 +694,40 @@ const MgmtConverter HttpStatusCodeList::Conv{
...
+//
+/////////////////////////////////////////////////////////////
+
+void
+TargetedCacheControlHeaders::parse(std::string_view src)
+{
+  count = 0;
+  swoc::TextView config_view{src};
+
+  while (config_view && count < MAX_HEADERS) {
```
zwoop: What happens if the config has more than MAX_HEADERS? Silent ignored it seems? Should we throw a warning at least (seems like that would indicate a configuration error).

include/proxy/http/HttpConfig.h
```diff
@@ -92,6 +92,44 @@ class HttpStatusCodeList
...
+ *
+ * Instead of parsing a comma-separated string on each request, this class
+ * stores the header names as an array of string_views into a stable backing
+ * string. The Converter ensures the string is parsed once at config load time
+ * and whenever the per-transaction override is set.
+ */
+class TargetedCacheControlHeaders
+{
+public:
+  static const MgmtConverter Conv;
```
zwoop, line 106: I know we don't do this consistently, but if the copy constructor / assignment constructor shouldn't be used (seems iffy with the raw pointer), should we = delete; them here ?

src/proxy/http/HttpSM.cc
```diff
@@ -2100,14 +2100,18 @@ HttpSM::state_read_server_response_header(int event, void *data)
...
       t_state.http_return_code = HTTPStatus::BAD_GATEWAY;
       call_transact_and_set_next_state(HttpTransact::BadRequest);
       break;
     }
 
+    // Recompute cooked cache control with targeted headers.
+    t_state.hdr_info.server_response.m_mime->recompute_cooked_stuff(nullptr,
+                                                                    t_state.txn_conf->targeted_cache_control_headers.get_headers(),
+                                                                    t_state.txn_conf->targeted_cache_control_headers.get_count());
+
```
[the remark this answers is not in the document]
bneradt: Fixed.

src/proxy/http/HttpConfig.cc
```diff
@@ -694,6 +694,40 @@ const MgmtConverter HttpStatusCodeList::Conv{
...
+const MgmtConverter TargetedCacheControlHeaders::Conv{
+  [](const void *data) -> std::string_view {
+    const TargetedCacheControlHeaders *hdrs = static_cast<const TargetedCacheControlHeaders *>(data);
+    return hdrs->conf_value ? hdrs->conf_value : "";
+  },
+  [](void *data, std::string_view src) -> void {
+    TargetedCacheControlHeaders *hdrs = static_cast<TargetedCacheControlHeaders *>(data);
+    // The string_views in headers[] point into conf_value, so conf_value must
+    // remain stable for the lifetime of the parsed headers.
+    hdrs->parse(src);
```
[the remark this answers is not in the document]
bneradt: Fixed.

include/proxy/http/HttpConfig.h
bneradt: Because this is a part of the ParsedConfig logic, this actually has to be copyable/assignable. A comment was added to explain this.

bneradt:
[approve ci centos autest 1]

bneradt:
> > Hi @JakeChampion
> > > @bneradt if you wanted to - you could run this patchset/branch against the tests in [cache-tests](https://github.com/http-tests/cache-tests) https://cache-tests.fyi/#cdn-cache-control and see how many ATS is now passing for ☺️
> > 
> > 
> > I ran the tests and no tests change. Note that this patch adds targeted cache control header support, which I don't think the cache-tests have tests for.
> > But it's good to verify that no new tests fail at least. :)
> 
> The section I linked too includes cdn-cache-control which I think is what you are also adding in this patch set, did those tests begin to pass?

Oh, I see. Thank you for being patient with me and clarifying this @JakeChampion . 

This is a good observation/question. No, this patch as is doesn't change those because this patch doesn't change default behavior. If we made cdn-cache-control treated as a targeted cache control header by default, then this should flip those to green. I'm guessing we don't want to make such a modifying change in a minor release (10.2.0) though.

But this is a good thing to discuss. Maybe we can make that change in 10.2? Or maybe add this feature with the default to not honor any targeted cache control headers in 10.2 but enable it by default in 11? @zwoop : I'll defer to your wisdom on this: should we treat cdn-cache-control as a targeted cache control header for ATS by default in both 10.2 and 11, or just in 11, or in neither?

zwoop:
I think it's a good idea to enable that in the CI test setup (since it's an RFC from several years ago). But I think we should defer enabling it by default until ATS v11 (which is "this year", so not too long of a wait).

bneradt:
> I think it's a good idea to enable that in the CI test setup (since it's an RFC from several years ago). But I think we should defer enabling it by default until ATS v11 (which is "this year", so not too long of a wait).

Sounds good. Once this lands I'll make a separte PR for 11-Dev to make it on by default.

bneradt:
OK, @JakeChampion and @zwoop : I re-ran the cache tests with this branch, enabling cdn-cache-control as a targeted cache-control header. The results look good:

```
Newly passing CDN cases include:

cdn-date-update-exceed
cdn-expires-update-exceed
cdn-fresh-cc-nostore
cdn-max-age
cdn-max-age-0
cdn-max-age-age
cdn-max-age-case-insensitive
cdn-max-age-cc-max-age-invalid-expires
cdn-max-age-expires
cdn-max-age-extension
cdn-max-age-max
cdn-max-age-max-plus
cdn-max-age-short-cc-max-age
cdn-remove-age-exceed
cdn-remove-header
```

When we change the default to enable cdn-cache-control as a targeted header by default in 11, these should start showing up as passing there too.

JakeChampion:
> OK, @JakeChampion and @zwoop : I re-ran the cache tests with this branch, enabling cdn-cache-control as a targeted cache-control header. The results look good:
> 
> ```
> Newly passing CDN cases include:
> 
> cdn-date-update-exceed
> cdn-expires-update-exceed
> cdn-fresh-cc-nostore
> cdn-max-age
> cdn-max-age-0
> cdn-max-age-age
> cdn-max-age-case-insensitive
> cdn-max-age-cc-max-age-invalid-expires
> cdn-max-age-expires
> cdn-max-age-extension
> cdn-max-age-max
> cdn-max-age-max-plus
> cdn-max-age-short-cc-max-age
> cdn-remove-age-exceed
> cdn-remove-header
> ```
> 
> When we change the default to enable cdn-cache-control as a targeted header by default in 11, these should start showing up as passing there too.

This is the best news! Nice work!

cmcfarlen:
Cherry-picked to 10.2.x

PR #12680, opened 2025-11-20
cleanup ProxyProtocol and prevent memory leak

cmcfarlen:
This adds the rule of 5 functions to ProxyProtocol to allow safe copy construction and assignment.
It also avoids a memory leak of the use of ProxyProtocol in HttpTransaction::State where it wasn't getting destructed (HttpSM has a State and is class allocated without the flag to run the destructor).

#12676 will fix the memory leak in a different way, but the rule of 5 is still needed where ProxyProtocol is copied.

### DIFF
--- a/include/iocore/net/ProxyProtocol.h
+++ b/include/iocore/net/ProxyProtocol.h
@@ -25,14 +25,11 @@
 
 #pragma once
 
-#include "tscore/ink_memory.h"
-#include <filesystem>
 #include <tscore/ink_inet.h>
 #include <swoc/TextView.h>
 #include <unordered_map>
 #include <cstdlib>
 #include <optional>
-#include <utility>
 
 enum class ProxyProtocolVersion {
   UNDEFINED,


src/iocore/net/ProxyProtocol.cc
```diff
@@ -552,14 +552,9 @@ ProxyProtocol::set_additional_data(std::string_view data)
 {
   uint16_t len = data.length();
   Dbg(dbg_ctl_proxyprotocol_v2, "Parsing %d byte additional data", len);
-  additional_data = static_cast<char *>(ats_malloc(len));
-  if (additional_data == nullptr) {
-    Dbg(dbg_ctl_proxyprotocol_v2, "Memory allocation failed");
-    return -1;
-  }
-  data.copy(additional_data, len);
+  additional_data.assign(data);
```
maskit, line 555: Don't we need to pass `len`?

maskit: Ah, nvm. I thought `data` is `char[]`.

include/iocore/net/ProxyProtocol.h
```diff
@@ -25,11 +25,14 @@
 
 #pragma once
 
+#include "tscore/ink_memory.h"
+#include <filesystem>
```
maskit: What needs this?

cmcfarlen: idk why my editor things filesystem is needed in this file...  I'll clean that up

src/iocore/net/unit_tests/test_ProxyProtocol.cc
```diff
@@ -625,3 +625,293 @@ TEST_CASE("ProxyProtocol v2 Builder", "[ProxyProtocol][ProxyProtocolv2]")
...
+
+      // Set properly formatted TLV data: type=0x01, length=0x0004, value="test"
+      std::string_view tlv_data("\x01\x00\x04test", 7);
+      pp.set_additional_data(tlv_data);
+
+      auto tlv = pp.get_tlv(0x01);
+      REQUIRE(tlv.has_value());
+      CHECK(tlv.value() == "test");
+    }
+    // Object destroyed here - ASAN will catch any memory leaks
```
bneradt, line 885: Neat

cmcfarlen:

PR #12681, opened 2025-11-20
Remove some unnecessary stub code

cmcfarlen:
These are needed anymore and can cause duplicate symbol errors on some configs/linkers.


cmcfarlen:
heh, maybe not always unecessary!

PR #12682, opened 2025-11-20
Remove Extendible and AcidPtr (unused code)

cmcfarlen:
This code is not used anywhere in the codebase.

Does anyone use it in external plugins?


bneradt:
> This code is not used anywhere in the codebase.
> 
> Does anyone use it in external plugins?

FWIW: I checked our Yahoo repos and don't see these used anywhere.

PR #12686, opened 2025-11-21
Fix proxy-verifier storage for git worktrees

bneradt:
PR #12664 changed proxy-verifier storage to be in
${CMAKE_SOURCE_DIR}/.git. Sadly, this fails in git worktrees where .git is a file pointing to the actual git directory. Use git rev-parse --git-common-dir to dynamically detect the correct git directory, which works for both regular repositories and worktrees. This ensures proxy-verifier binaries are stored in a shared location accessible to all worktrees.

### DIFF
--- a/cmake/proxy-verifier.cmake
+++ b/cmake/proxy-verifier.cmake
@@ -15,7 +15,7 @@
 #
 #######################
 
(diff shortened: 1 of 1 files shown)


ezelkow1 (approved):

cmcfarlen:

PR #12688, opened 2025-11-21
Add USDT for cache directory insertion and deletion

moonchen:
Add cache directory insert and delete USDT.  These will emit the cache directory stripe, segment, segment offset (implying bucket and depth), disk offset, and approx size.

Pass the subclass StripeSM instead of superclass Stripe into helper functions in order to access stripe->fd.

Add an autest for these.


src/iocore/cache/CacheDir.cc
```diff
@@ -724,9 +734,12 @@ Directory::remove(const CacheKey *key, StripeSM *stripe, Dir *del)
           return 0;
       }
 #endif
-      if (dir_compare_tag(e, key) && dir_offset(e) == dir_offset(del)) {
+      int64_t offset = dir_offset(e);
+      if (dir_compare_tag(e, key) && offset == dir_offset(del)) {
         ts::Metrics::Gauge::decrement(cache_rsb.direntries_used);
         ts::Metrics::Gauge::decrement(stripe->cache_vol->vol_rsb.direntries_used);
+        ATS_PROBE7(cache_dir_remove, stripe ? stripe->fd : -1, s, dir_to_offset(e, seg), offset, dir_approx_size(e),
+                   key->slice64(0), key->slice64(1));
```
bneradt: Minor: probably just remove the null check on stripe. You don't do that elsewhere in this patch, and it was just dereferenced on the previous line. So if there's a problem with it being null, it won't be your fault. :) 

Maybe add `ink_assert(stripe != nullptr);` to the top of the function stating the invariant that it whould be non-null? Taking a `&` is a possibility too, but then this patch is increasing beyond scope. So the assert should be fine.

bneradt (changes requested):
Looks reasonable.

Can we put the test in tests/gold_tests/ats_probe though instead of making a new directory? They seem related enough they can share the same organizing directory.

src/iocore/cache/CacheDir.cc
moonchen: Thanks.  This was a vestige I forgot to remove.

moonchen: Removed with a new commit.

PR #12690, opened 2025-11-21
Fix false positive -Wrestrict warning in test_ja4.cc with GCC 12

bryancall:
Suppress false positive compiler warning when assigning single character string literal to std::string. This is a known GCC 12 bug that triggers when the optimizer inlines string operations with 1-character strings.

The warning shows nonsensical memory access values (exabytes at negative offsets) which confirms it's a compiler analysis bug, not a real issue.

Fixes build failure on Debian 12 with GCC 12 in Release mode.

See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105329

### DIFF
--- a/plugins/experimental/ja4_fingerprint/test_ja4.cc
+++ b/plugins/experimental/ja4_fingerprint/test_ja4.cc
@@ -230,16 +230,7 @@ TEST_CASE("JA4")
           "when we create a JA4 fingerprint, "
           "then indices [8,9] thereof should contain \"aa\".")
   {
-    // Suppress false positive -Wrestrict warning in GCC 12 when optimizing
-    // See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105329
-#if defined(__GNUC__) && !defined(__clang__)
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wrestrict"
-#endif
-    TLS_summary.ALPN = "a";
-#if defined(__GNUC__) && !defined(__clang__)
-#pragma GCC diagnostic pop
-#endif
+    TLS_summary.ALPN = 'a';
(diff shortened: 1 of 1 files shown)


plugins/experimental/ja4_fingerprint/test_ja4.cc
```diff
@@ -230,7 +230,16 @@ TEST_CASE("JA4")
           "when we create a JA4 fingerprint, "
           "then indices [8,9] thereof should contain \"aa\".")
   {
+    // Suppress false positive -Wrestrict warning in GCC 12 when optimizing
+    // See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105329
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wrestrict"
+#endif
     TLS_summary.ALPN = "a";
```
JosiahWI: Would assigning `'a'` instead of `"a"` silence the warning? That would be cleaner; I did not know the string interface supported that until now.

bryancall:
Good suggestion! Using `std::string(1, 'a')` is indeed cleaner than the pragma workaround. I've pushed a new commit (f63dee7db) that implements this approach - it constructs the string explicitly from a char which avoids the GCC 12 false positive entirely.

This is a better solution as it:
- Avoids compiler-specific pragmas
- Makes the intent clearer
- Eliminates the warning at the source rather than suppressing it

bryancall:
Updated to use direct char assignment (`TLS_summary.ALPN = 'a';`) which is even cleaner - `std::string` has an assignment operator that takes a single char.

JosiahWI:
Needs a non-Yahoo approval, but this looks good.

bryancall:
@JosiahWI for simple changes we agreed that we don't need cross company review.

PR #12696, opened 2025-11-22
cmake format target: add whitespace correction

bneradt:
Our format target does formatting of C++ and Python files via clang-format and yapf. Annoyingly, it hasn't done some extra whitespace checks that CI does. So a PR can be made that passes `format` locally but then CI will reject it for trailing whitespace, or DOS line endinds. This adds a new tools/whitespace-format.sh script that automatically fixes trailing whitespace and DOS carriage returns, matching the checks performed by CI. This script is then integrated into the `format` cmake target so it gets run for that.

I verified that the script works on both Linux and Mac.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -806,10 +806,18 @@ add_custom_target(
   VERBATIM
 )
 
+add_custom_target(
+  whitespace-format
+  ${CMAKE_SOURCE_DIR}/tools/whitespace-format.sh
+  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+  COMMENT "fixing whitespace issues (trailing whitespace, DOS line endings)"
+  VERBATIM
+)
+
 # Add a format target that runs all the formatters.
 add_custom_target(
   format
-  DEPENDS clang-format yapf cmake-format
+  DEPENDS clang-format yapf cmake-format whitespace-format
   COMMENT "formatting all files"
 )
 
--- a/tools/whitespace-format.sh
+++ b/tools/whitespace-format.sh
@@ -1,0 +1,68 @@
+#!/usr/bin/env bash
+#
+#  Script to fix whitespace issues: trailing whitespace and DOS carriage returns.
+#
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+set -e
+
+REPO_ROOT=$(cd $(dirname $0) && git rev-parse --show-toplevel)
+echo "Fixing whitespace issues in ${REPO_ROOT}"
+
+# macOS requires an empty string argument for in-place sed edits.
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    SED_INPLACE=(sed -i '')
+else
+    SED_INPLACE=(sed -i)
+fi
+
+FILES_MODIFIED=0
+
+echo "Checking for trailing whitespace..."
+TRAILING_WS_FILES=$(git grep -IE ' +$' | eval "grep -F -v 'lib/yamlcpp' | grep -F -v 'lib/Catch2' | grep -F -v 'lib/systemtap' | grep -F -v '.gold:' | grep -F -v '.test_input'" | cut -d: -f1 | sort -u || true)
+
+if [ -n "$TRAILING_WS_FILES" ]; then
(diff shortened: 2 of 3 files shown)


bneradt:
clang-format removes trailing white space, but people leave trailing white space in non c++ files. That causes issues with PRs as our CI system enforces no trailing white spaces anywhere (save in excepted files, which whitespace-format.sh replicates.) We might as well take care of this for devs.

Why not do this? It runs in under half a second. It saves some annoyance.

tools/whitespace-format.sh
```diff
@@ -0,0 +1,95 @@
...
+# Get the repository root
+REPO_ROOT=$(cd $(dirname $0) && git rev-parse --show-toplevel)
+
+EXCLUDE_PATTERNS=(
+    "lib/yamlcpp"
+    "lib/Catch2"
+    "lib/systemtap"
+    ".gold:"
+    ".test_input"
+)
```
cmcfarlen: This is defined here, but not used below for selecting files for WS and DOS?

tools/whitespace-format.sh
```diff
@@ -0,0 +1,95 @@
...
+
+GREP_EXCLUDE_ARGS=""
+for pattern in "${EXCLUDE_PATTERNS[@]}"; do
+    GREP_EXCLUDE_ARGS="${GREP_EXCLUDE_ARGS} | grep -F -v '${pattern}'"
```
cmcfarlen: This variable seems unused.  I thought maybe grep will use it, but couldn't find info on that.  Also it's not exported.  Maybe just remove this?

tools/whitespace-format.sh
```diff
@@ -0,0 +1,95 @@
...
+
+# Fix trailing whitespace.
+echo "Checking for trailing whitespace..."
+TRAILING_WS_FILES=$(git grep -IE ' +$' | eval "grep -F -v 'lib/yamlcpp' | grep -F -v 'lib/Catch2' | grep -F -v 'lib/systemtap' | grep -F -v '.gold:' | grep -F -v '.test_input'" | cut -d: -f1 | sort -u || true)
+
+if [ -n "$TRAILING_WS_FILES" ]; then
+    while IFS= read -r file; do
+        if [ -f "$file" ]; then
+            echo "  Fixing trailing whitespace in: $file"
+            "${SED_INPLACE[@]}" 's/[[:space:]]*$//' "$file"
```
cmcfarlen, line 42: I wonder if this always applicable? One case I can think of where it might not be correct is if there was some C++ raw string that needed trailing whitespace. I know clangd has special code to handle some corner cases.

CMakeLists.txt
```diff
@@ -806,10 +806,18 @@ add_custom_target(
...
+  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+  COMMENT "fixing whitespace issues (trailing whitespace, DOS line endings)"
+  VERBATIM
+)
+
 # Add a format target that runs all the formatters.
 add_custom_target(
   format
-  DEPENDS clang-format yapf cmake-format
+  DEPENDS clang-format yapf cmake-format whitespace-format
```
cmcfarlen, line 820: I rarely run this target.  Often yapf doesn't work, so I just run the `cmake-format` when that check complains.

tools/whitespace-format.sh
```diff
@@ -0,0 +1,95 @@
...
+
+# Fix DOS carriage returns.
+echo "Checking for DOS carriage returns..."
+DOS_FILES=$(git grep -IE $'\r$' | eval "grep -F -v 'lib/yamlcpp' | grep -F -v 'lib/Catch2' | grep -F -v 'lib/systemtap' | grep -F -v '.test_input'" | cut -d: -f1 | sort -u || true)
+
+if [ -n "$DOS_FILES" ]; then
+    while IFS= read -r file; do
+        if [ -f "$file" ]; then
+            echo "  Removing DOS carriage returns from: $file"
+            "${SED_INPLACE[@]}" $'s/\r$//' "$file"
```
cmcfarlen, line 55: I think there can be some traps here as well.  If we end up with binary files in the repo, this might destroy them.

bneradt: The `-I` in grep ignores binary files.

tools/whitespace-format.sh
bneradt: Maybe in theory. Keep in mind this will fail CI today if that ever is the case because that check is done in CI now. My intention with this PR is to pull the CI check to a local change so that things fail locally rather than later in CI.

For reference, here's the CI check that this file borrows logic from:
https://github.com/apache/trafficserver-ci/blob/2286af7d567d1f36a78aeb138a97cf9f3dc70b45/jenkins/github/format.pipeline#L62-L74

Even better: with this change, I can take these special white space checks out of CI and just use this format target alone. Then if such a string is added, this script can be updated for it along with the PR that introduces such problems.

bneradt:
> since this doesn't check for issues in the pre-commit hook, it would only reliably help if users ran the format target before each PR.

Good point. I updated the pre-commit hook.

CMakeLists.txt
bneradt: This is interesting, albeit not something this patch will address. I never have issues yapf, but I always develop in a container from CI that has the tools (namely, python, pip, pipenv) that makes this work reliably.

I assume you have python issues? I don't know of a way to address python formatting more reliably. It would be nice if we had a compiled binary we can ship like we do with clang-format, but I think that's not available.

cmcfarlen:
Removing from 10.1 project as there are conflicts and we shouldn't be adding anything but cherry-picks to that branch anyway.

PR #12699, opened 2025-11-24
Fix proxy-verifier path for non-worktree repositories

bneradt:
The git rev-parse --git-common-dir command added in #12686 returns an absolute path to the relevant .git direcotry for worktrees. Unfortunately, it does a relative path for "regular" non-worktree clones. Without converting to an absolute path, PROXY_VERIFIER_PATH becomes relative and fails to resolve correctly when passed to autest. This adds get_filename_component to ensure GIT_COMMON_DIR is always absolute, fixing the issue for both repository types.


cmcfarlen:

PR #12700, opened 2025-11-24
Revert "cache-*.test.py tests to use ATSReplayTest"

bneradt:
Reverts apache/trafficserver#12694

We have to circle back around on these. It turns out that CI was not running the autests due to needing #12699.


bneradt:
[approve ci autest 1]

PR #12701, opened 2025-11-24
cache-*.test.py tests to use ATSReplayTest (#12694)

bneradt:
Convert various cache-*.test.py tests to use the new ATSReplayTest
extension. These were generally pretty easy to convert because many
of them were Proxy Verifier replay file based tests already anyway.

To support this, this also adds diags.log and traffic.out log
verification support to ATSReplayTest.

---

This is the same as #12694, but has dns fixes. See the second commit.


traeak (approved):
i see no issue with the 'dns fixes' commit.

Issue #6900, opened 2020-06-16
request buffer with post_copy_size= 0 returns 403.


zizhong:
A better option is to disable the request buffer feature if post_copy_size is 0.

zizhong:
It's been a long time but from the ticket name, it could be that when a POST request with empty post body, ATS returns 403 with request buffer enabled.
cc @bneradt @SolidWallOfCode

PR #12702, opened 2025-11-25
Fix request buffering with post_copy_size=0

bneradt:
When proxy.config.http.request_buffer_enabled is set to 1 with proxy.config.http.post_copy_size set to 0, POST requests were failing. The fix disables request buffering when post_copy_size is 0, treating it as an indication to not buffer POST data.

Fixes: #6900

### DIFF
--- a/src/proxy/http/HttpConfig.cc
+++ b/src/proxy/http/HttpConfig.cc
@@ -1345,9 +1345,13 @@ HttpConfig::reconfigure()
   params->redirection_host_no_port          = INT_TO_BOOL(m_master.redirection_host_no_port);
   params->oride.number_of_redirections      = m_master.oride.number_of_redirections;
   params->post_copy_size                    = m_master.post_copy_size;
-  params->redirect_actions_string           = ats_strdup(m_master.redirect_actions_string);
-  params->redirect_actions_map = parse_redirect_actions(params->redirect_actions_string, params->redirect_actions_self_action);
-  params->http_host_sni_policy = m_master.http_host_sni_policy;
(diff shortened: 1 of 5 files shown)


masaori335:
This looks good. I'd also add a WARNING log with this combination to notify admin a mistake.

bneradt:
> This looks good. I'd also add a WARNING log with this combination to notify admin a mistake.

Good idea. Warning added.

masaori335 (approved):
Looks good!

maskit:
Sounds like `request_buffer_enabled` could be deprecated.

bneradt:
> Sounds like `request_buffer_enabled` could be deprecated.

That is possible. There's a couple points to consider here:

* This feature was around before I was, but it could be that when this was initially implemented, it was assumed it was easier to set post_copy_size once to a global value (it is not currently overridable), then toggle the buffering via the request_buffer_enabled boolean via override. That's a line of reasoning I could see someone making for keeping them separate.
* Maybe the whole feature should be deprecated in favor of transforms? With a transform, any plugin can do their own buffering rather than core doing it. Although, maybe providing the buffering as a service is helpful since manually buffering from a transform isn't trivial.

I'm in favor of your suggestion: making `post_copy_size` overridable and deprecating `request_buffer_enabled`.

maskit:
Good point. I didn't think about overridability, and I think it explains why we have the two.

> I'm in favor of your suggestion: making post_copy_size overridable and deprecating request_buffer_enabled.

Sounds like a plan.

PR #12703, opened 2025-11-25
use catch_discover_test to define tests

cmcfarlen:
This changes the ctest test definitions to be each TEST_CASE or SCENARIO macro instead of each test executable.

It requires some changes to tests to work since it will run more tests in parallel, so it might not be a good fit for us.  It does tell you which test failed, rather than which executable, and it enables some tooling to run individual tests.


cmcfarlen:
[approve ci]

ezelkow1:
[approve ci]

Issue #12691, opened 2025-11-21
Odd regex_remap target match string when "host" option is used - contains partial scheme string


djcarlin:
Using ATS 10.1

I am trying to write a blanket http to https redirect remap rule.  I did the following using both the `host` and `pristine` options of regex_remap plugin:

```
# RFC2606 says this tld should never resolve - destination of remap rule is unused in this configration
regex_map http://(.*)   https://unused.invalid @plugin=regex_remap.so @pparam=/opt/regex_remap/plaintext-redirect.config   @pparam=host    @pparam=pristine
```

And the contents of /opt/regex_remap/plaintext-redirect.config regex_remap plugin config is the following:

`^(.*)$   https://$1   @status=307
`

Whats interesting is that the target match string contains the `//` from the scheme of the pristine URL which I thought was odd:

```
[Nov 21 20:45:28.692] [ET_NET 10] DIAG: <regex_remap.cc:948 (TSRemapDoRemap)> (regex_remap) Target match string is `//thisisatest.yahoo.com/?foo=bar&bar=baz'
[Nov 21 20:45:28.692] [ET_NET 10] DIAG: <regex_remap.cc:1014 (TSRemapDoRemap)> (regex_remap) New URL is estimated to be 51 bytes long, or less
[Nov 21 20:45:28.692] [ET_NET 10] DIAG: <regex_remap.cc:1015 (TSRemapDoRemap)> (regex_remap) New URL is https:////thisisatest.yahoo.com/?foo=bar&bar=baz (length 48)
[Nov 21 20:45:28.692] [ET_NET 10] DIAG: <regex_remap.cc:1016 (TSRemapDoRemap)> (regex_remap)     matched rule 1 [^(.*)$]
[Nov 21 20:45:28.692] [ET_NET 10] DIAG: <regex_remap.cc:1028 (TSRemapDoRemap)> (regex_remap) Redirecting URL, status=307

```

So the working config needs to remove the `//` from the destination scheme to work:

`^(.*)$   https:$1   @status=307
`

Only with this 2nd config do I get the proper redirect.

bneradt:
I did notice that the docs do actually hint at this behavior:

https://docs.trafficserver.apache.org/en/latest/admin-guide/plugins/regex_remap.en.html

> Finally, to match on the host as well, use the option ‘host’, e.g.
>
> ... @pparam=maps.reg @pparam=host
> With this enabled, the string that you will need to match will look like
>
> //host/path?query=bar 

[...]

PR #12706, opened 2025-11-25
11.x: regex_remap: Remove // prefix from host match string

bneradt:
When the host option was enabled, regex_remap prepended '//' before the host in the match string, resulting in '//host/path' instead of 'host/path'. This was confusing because users associated the '//' with the 'http://' scheme, leading to double-slash issues when writing HTTP-to-HTTPS redirect rules. The match string now uses 'host/path' format. Documentation and tests have been updated accordingly, including new examples for host and method options.

Fixes: #12691

There was a dev mailing list email about this and no one complained about the change. Applying lazy consensus, I'm marking this PR as ready for review.

### DIFF
--- a/doc/admin-guide/plugins/regex_remap.en.rst
+++ b/doc/admin-guide/plugins/regex_remap.en.rst
@@ -97,10 +97,10 @@ Finally, to match on the host as well, use the option 'host', e.g. ::
 
 With this enabled, the string that you will need to match will look like ::
 
-    //host/path?query=bar
+    host/path?query=bar
 
 
-A typical regex would look like ::
+A typical regex, without the above mentioned method or host options, would look like ::
 
     ^/(ogre.*)/more     http://www.ogre.com/$h/$0/$1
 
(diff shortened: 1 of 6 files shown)


bneradt:
@zwoop had the following comment to the mailing list thread:

```
I don’t have a strong opinion either way, but I have a vague recollection
that this came from the fact that someone used to have hostnames in the
path. E.g. something like (maybe…)

    https://cdn.example.com/origin1.example.com/some/path/
    https://cdn.example.com/origin2.example.com/some/path/

etc. And I *think* that's the reason for keeping the // with the authority,
making it easy, safe (and maybe even possible sometimes) to distinguish it
from the path. I.e. a regex like

   /\/\/cdn.*.com\/(.*\.example\.com)/

So, I’m -0 on this change, and I don’t know if it’s a use case anyone have
now. As such, I’m ok landing this, but given a choice I probably wouldn’t.

Cheers,

— Leif
```

This is a good point. That being the case, I'm going to close this PR and replace it with a documentation update PR making sure this behavior is well explained: #12730

PR #12707, opened 2025-11-26
Fix for fetchsm chunked response handling with dechunking turned off

traeak:
We ran into this bug with OCSP stapling turned on which by default doesn't have dechunking turned on but does have skip remap turned on.

At the time one of the cloud* cdn edges was sporadically returning chunked 404 responses to stapling requests which caused ATS to crash.

I was able to recreate the crash and verify the fix by adding a fetch option to the lua plugin itself.


bneradt (approved):

cmcfarlen:

PR #12712, opened 2025-12-01
Fix cache test reenable loop in VC_EVENT_READ_READY handler

bryancall:
## Problem

The `test_cache_Populated_Cache` unit test was failing with "Subprocess aborted" showing thousands of repeated reenable calls on the same CacheVC object. The Jenkins CI logs showed an infinite loop of reenable messages:

https://ci.trafficserver.apache.org/job/master/job/os_build/47228/console

## Root Cause

In `src/iocore/cache/unit_tests/main.cc`, the `CacheReadTest::read_event` handler was calling `process_event()` **inside the while loop** that consumes data blocks:

```cpp
case VC_EVENT_READ_READY: {
  while (this->_reader->block_read_avail()) {
    // ... consume block ...
    this->process_event(event);  // ❌ Called for EVERY block
  }
  break;
}
```

This meant `reenable()` was being called for every block of data, which is incorrect. When reading from a populated cache with many blocks, this resulted in excessive reenables and triggered the test abort.

## The Fix

Move the `process_event()` call **outside the while loop** so `reenable()` is only called **once per event** after all available data has been consumed:

```cpp
case VC_EVENT_READ_READY: {
  while (this->_reader->block_read_avail()) {
    // ... consume block ...
  }
  this->process_event(event);  // ✅ Called once after consuming all data
  break;
}
```

This matches the pattern used in `CacheTestSM` (the production regression test code) and follows standard async I/O patterns: consume all available data, then signal readiness for more.

## Testing

- ✅ `test_cache_Cache` - Passed (fresh cache)
- ✅ `test_cache_Populated_Cache` - **Passed** (previously failing)

The fix reduces reenable calls from thousands to just a handful per test run.

### DIFF
--- a/src/iocore/cache/unit_tests/main.cc
+++ b/src/iocore/cache/unit_tests/main.cc
@@ -319,19 +319,23 @@ CacheReadTest::read_event(int event, void *e)
     this->process_event(event);
     break;
   case VC_EVENT_READ_READY: {
+    // Consume all available data blocks and validate against expected data
     while (this->_reader->block_read_avail()) {
       auto str = this->_reader->block_read_view();
       if (memcmp(str.data(), this->_cursor, str.size()) == 0) {
         this->_reader->consume(str.size());
         this->_cursor += str.size();
-        this->process_event(event);
       } else {
+        // Data corruption detected - fail the test immediately
         CHECK(false);
         this->close();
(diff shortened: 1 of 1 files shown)


bryancall:
[approve ci rocky]

PR #12663, opened 2025-11-13
doc: fix regex_remap overridable config syntax description

bneradt:
The documentation incorrectly suggested using
@overridable-config=<value> syntax for overriding configuration
variables. No such overridable-config option exists. The correct syntax
is to use the config name directly with @<config-name>=<value> (e.g.,
@proxy.config.url_remap.pristine_host_hdr=0). This commit removes the
incorrect reference and adds an example.

---

## For Review

The rendered doc can be seen here:
https://ci.trafficserver.apache.org/job/Github_Builds/job/docs/7939/artifact/output/12663/docbuild/html/admin-guide/plugins/regex_remap.en.html

PR #12716, opened 2025-12-02
10.1: Fix regex_remap doc for overridable config syntax (#12663)

bneradt:
This is a backport of #12663 to 10.1.x

---

The documentation incorrectly suggested using
@overridable-config=<value> syntax for overriding configuration variables. No such overridable-config option exists. The correct syntax is to use the config name directly with @<config-name>=<value> (e.g., @proxy.config.url_remap.pristine_host_hdr=0). This commit removes the incorrect reference and adds an example.

(cherry picked from commit 69d1c79c02dd4ac01e31e68129c372abd119710b)


bneradt:
[approve ci rocky]

PR #12717, opened 2025-12-02
Cache volumes: RAM cache settings and remap option

zwoop:
This adds new configuration options for volume.config, as well as
adding an `@volume= `setting to remap.config (avoiding the hosting.config
stepping stone).

### DIFF
--- a/src/iocore/cache/CacheHosting.cc
+++ b/src/iocore/cache/CacheHosting.cc
@@ -852,8 +852,10 @@ ConfigVolumes::BuildListFromString(char *config_file_path, char *file_buf)
       } else {
         ink_release_assert(!"Unexpected non-HTTP cache volume");
       }
-      Dbg(dbg_ctl_cache_hosting, "added volume=%d, scheme=%d, size=%d percent=%d, ramcache enabled=%d", volume_number,
-          static_cast<int>(scheme), size, in_percent, ramcache_enabled);
+      Dbg(dbg_ctl_cache_hosting,
+          "added volume=%d, scheme=%d, size=%d percent=%d, ramcache enabled=%d, "
+          "ram_cache_size=%" PRId64 ", ram_cache_cutoff=%" PRId64,
+          volume_number, static_cast<int>(scheme), size, in_percent, ramcache_enabled, ram_cache_size, ram_cache_cutoff);
     }
 
     tmp = bufTok.iterNext(&i_state);
--- a/tests/gold_tests/cache/cache_volume_features.replay.yaml
+++ b/tests/gold_tests/cache/cache_volume_features.replay.yaml
@@ -17,7 +17,16 @@
 meta:
   version: "1.0"
 
-# Configuration section for autest integration
+# This test proves that multi-parameter volume.config lines are parsed
+# correctly.  The generated volume.config lines are:
+#
+#   volume=1 scheme=http size=50% ram_cache_size=32M ram_cache_cutoff=8K
+#   volume=2 scheme=http size=50%
+#
+# All 5 key=value pairs on line 1 must be parsed from the single line.
+# If the inner loop's "tmp = line_end" advancement were broken (as claimed
+# in a PR review), only "volume=1" would be parsed, scheme and size would
+# be missing, and the volume would be rejected -- disabling the cache.
 autest:
   description: "Test cache volume features: per-volume RAM cache and @volume= directive"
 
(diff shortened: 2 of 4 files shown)


doc/admin-guide/files/volume.config.en.rst
```diff
@@ -73,20 +73,65 @@ Optional directory entry sizing
...
+
+For example, if the global RAM cache size is 4GB and you allocate 1GB to volume 1
+and 512MB to volume 2, the remaining 2.5GB will be distributed among other volumes
+using the normal proportional allocation based on disk space.
+
+**Important notes:**
+
+* If the sum of all ``ram_cache_size`` allocations exceeds the global RAM cache size,
+  a warning is logged and the private allocations are disabled, falling back to
+  the standard shared allocation.
```
cmcfarlen: Should this be a fatal error on startup? It might be hard to notice a warning and otherwise traffic_server would run with unexpected config.

cmcfarlen: Should it also be an error if `ramcache=false` and `ram_cache_size` is specified? Or just make sure not to subtract from `ram_cache.size` if the size is specified and `ramcache=false`?

zwoop: Fixed.

src/iocore/cache/Cache.cc
```diff
@@ -100,6 +100,7 @@ ClassAllocator<EvacuationBlock, false>    evacuationBlockAllocator("evacuationBl
 ClassAllocator<CacheRemoveCont, false>    cacheRemoveContAllocator("cacheRemoveCont");
 ClassAllocator<EvacuationKey, false>      evacuationKeyAllocator("evacuationKey");
 std::unordered_set<std::string>           known_bad_disks;
```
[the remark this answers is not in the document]
zwoop: Added a comment.

src/iocore/cache/CacheHosting.cc
```diff
@@ -765,11 +784,42 @@ ConfigVolumes::BuildListFromString(char *config_file_path, char *file_buf)
...
+        }
+        while (*tmp && (ParseRules::is_digit(*tmp) || strchr("kmgtKMGT", *tmp))) {
+          tmp++;
+        }
       }
 
       // ends here
       if (end < line_end) {
-        tmp++;
+        tmp = line_end;
```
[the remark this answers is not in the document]
zwoop: Fixed.

zwoop:
[approve ci autest 1]

zwoop:
I've addressed CoPilot's reviews again.

src/iocore/cache/CacheHosting.cc
```diff
@@ -765,11 +784,42 @@ ConfigVolumes::BuildListFromString(char *config_file_path, char *file_buf)
...
+        if (!ParseRules::is_digit(*tmp)) {
+          err = "Invalid ram_cache_size value (must start with a number, e.g., 10G)";
+          break;
+        }
+        ram_cache_size = ink_atoi64(tmp);
+
+        if (ram_cache_size < 0) {
+          err = "Invalid ram_cache_size value (must be >= 0)";
+          break;
+        }
```
bryancall, line 798: `ink_atoi64` only recognizes uppercase suffixes (`K`, `M`, `G`, `T`), but this skip loop accepts both cases via `strchr("kmgtKMGT", *tmp)`. If someone writes `ram_cache_size=2g`, `ink_atoi64` returns `2` (no multiplier applied), the skip loop eats the `g`, and the volume silently gets a 2-byte RAM cache allocation.

Either restrict the skip to uppercase only:
```cpp
while (*tmp && (ParseRules::is_digit(*tmp) || strchr("KMGT", *tmp))) {
```
Or better, reject lowercase with an error. Same issue applies to `ram_cache_cutoff`, `avg_obj_size`, and `fragment_size` parsing.

src/proxy/http/remap/RemapConfig.cc
```diff
@@ -1197,12 +1209,55 @@ remap_parse_config_bti(const char *path, BUILD_TABLE_INFO *bti)
...
+          goto MAP_ERROR;
+        }
+
+        for (const char *p = volume_str; *p; p++) {
+          if (*p != ',' && (*p < '0' || *p > '9')) {
+            snprintf(errStrBuf, sizeof(errStrBuf), "Invalid character '%c' in @volume=%s at line %d", *p, volume_str, cln + 1);
+            errStr = errStrBuf;
+            goto MAP_ERROR;
+          }
+        }
```
bryancall, line 1256: This character-level validation allows degenerate inputs: `@volume=,,`, `@volume=0`, `@volume=,1,`, `@volume=999`. For the eager path (cache ready), `createCacheHostRecord()` catches bad volume numbers. But for the deferred path (initial startup), the string is stored as-is and errors only surface later as a non-fatal `Error()` log — the remap rule silently runs without a volume override.

Consider adding lightweight range validation here (split on commas, verify each number is 1–255, reject empty segments).

src/proxy/ReverseProxy.cc
```diff
@@ -168,6 +167,65 @@ reloadUrlRewrite()
...
+  init_store_volume_host_records(table->reverse_mappings);
+  init_store_volume_host_records(table->permanent_redirects);
+  init_store_volume_host_records(table->temporary_redirects);
+  init_store_volume_host_records(table->forward_mappings_with_recv_port);
+
+  table->release();
+}
+
 int
 url_rewrite_CB(const char * /* name ATS_UNUSED */, RecDataT /* data_type ATS_UNUSED */, RecData data, void *cookie)
```
bryancall, line 230: Minor: `UrlRewrite *table = rewrite_table;` is a non-atomic read of a pointer that `reloadUrlRewrite()` modifies via `ink_atomic_swap`. The startup-only constraint makes this safe in practice, but an atomic load would make the intent explicit:
```cpp
UrlRewrite *table = ink_atomic_load(&rewrite_table);
```

tests/gold_tests/cache/cache_volume_features.replay.yaml
```diff
@@ -0,0 +1,254 @@
...
+
+      # Invalid volume selection (should fallback)
+      - from: "http://invalid.example.com/"
+        to: "http://backend.ex:{SERVER_HTTP_PORT}/"
+        options:
+          - "@volume=99"
+
+# Proxy verifier test sessions
+sessions:
+  # Test 1: Volume selection via @volume=1
```
bryancall: Does this test actually exercise the intended behavior? If cache is ready when remap loads, `createCacheHostRecord("99", ...)` will fail (volume 99 isn't in `volume.config`) and the remap rule fails to load entirely. If cache isn't ready (deferred), the string is stored and `initVolumeHostRec` fails later with a non-fatal error — the rule works but without volume override.

If the intent is "invalid volume gracefully degrades to default selection," it might be worth making that explicit in the test description.

src/proxy/ReverseProxy.cc
```diff
@@ -168,6 +167,65 @@ reloadUrlRewrite()
...
+    }
+  }
+}
+
+// This is called after the cache is initialized, since we may need the volume_host_records.
+// Must only be called during startup before any remap reload can occur.
+void
+init_remap_volume_host_records()
+{
+  UrlRewrite *table = rewrite_table;
```
bryancall: Minor: `UrlRewrite *table = rewrite_table;` is a non-atomic read of a pointer that `reloadUrlRewrite()` modifies via `ink_atomic_swap`. The startup-only constraint makes this safe in practice, but an atomic load would make the intent explicit:
```cpp
UrlRewrite *table = ink_atomic_load(&rewrite_table);
```

tests/gold_tests/cache/cache_volume_features.replay.yaml
bryancall: Does this test actually exercise the intended behavior? If cache is ready when remap loads, `createCacheHostRecord("99", ...)` will fail (volume 99 isn't in `volume.config`) and the remap rule fails to load entirely. If cache isn't ready (deferred), the string is stored and `initVolumeHostRec` fails later with a non-fatal error — the rule works but without volume override.

If the intent is "invalid volume gracefully degrades to default selection," it might be worth making that explicit in the test description.

zwoop:
I added some more autests to show that the current logic is correct. We've going back and forth on this issue a couple of times, and I'm 93% certain that the parser is correct (and the tests does show that it works at least).

zwoop:
[approve CI]

zwoop:
[approve ci rat]

bryancall:
[approve ci]

cmcfarlen:
Cherry-picked to 10.2.x

PR #12718, opened 2025-12-02
Fix DOS pre-commit hook check.

bneradt:
It turns out \r isn't recognized by -E (extended regex), using -P (perl) does recognize it.


tools/git/pre-commit
```diff
@@ -149,6 +150,4 @@ else
     echo
 fi
 
-# Cleanup before exit
-rm -f "$clang_patch_file" "$yapf_patch_file"
```
bneradt, line 153: This didn't include all three files and is redundant with the trap command near the top of the file. So I'll clean this up by just removing this.

PR #12719, opened 2025-12-02
Simplify overridable config code

bneradt:
This commit simplifies the process of adding overridable configurations by using a new OverridableConfigDefs.h file as the largely single source of truth.

OverridableConfigDefs.h contains an X-macro that helps autogenerate the following which used to need to be manually updated:

- Lua plugin enum and var array (ts_lua_http_config.cc)
- String-to-enum mapping (overridable_txn_vars.cc)
- _conf_to_memberp switch statement (InkAPI.cc)
- SDK_Overridable_Configs test array (InkAPITest.cc)

Files that still require manual updates when adding a new config:

- include/ts/apidefs.h.in: enum value for ABI stability.
- include/proxy/http/OverridableConfigDefs.h: X-macro entry in enum order.
- include/proxy/http/HttpConfig.h: struct member.
- docs: Documentation files.

### DIFF
--- a/src/api/InkAPI.cc
+++ b/src/api/InkAPI.cc
@@ -7203,28 +7203,105 @@ _memberp_to_generic(MgmtFloat *ptr, MgmtConverter const *&conv) -> typename std:
   return ptr;
 }
 
-// Converter dispatch macros for the X-macro expansion.
+/**
+ * X-Macro Dispatch for _conf_to_memberp()
+ *
+ * Don't be intimidated by the macros below - they're simpler than they look!
+ * The end result is just a switch statement with one case per config, like:
+ *
+ *   switch (conf) {
+ *     case TS_CONFIG_HTTP_CHUNKING_ENABLED:
+ *       ret = _memberp_to_generic(&overridableHttpConfig->chunking_enabled, conv);
+ *       break;
+ *     case TS_CONFIG_HTTP_DOWN_SERVER_CACHE_TIME:
+ *       conv = &HttpDownServerCacheTimeConv;
+ *       ret = &overridableHttpConfig->down_server_timeout;
+ *       break;
+ *     // ... ~130 more cases, one per overridable config ...
+ *   }
+ *
+ * The macros just auto-generate these cases from OverridableConfigDefs.h so we
+ * don't have to maintain them by hand. Here's how an entry becomes a case:
+ *
+ * Example 1 - Standard config (GENERIC converter):
+ *   Entry in OverridableConfigDefs.h:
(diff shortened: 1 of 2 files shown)


src/api/InkAPI.cc
```diff
@@ -7202,6 +7203,27 @@ _memberp_to_generic(MgmtFloat *ptr, MgmtConverter const *&conv) -> typename std:
...
+  case TS_CONFIG_##KEY: ret = &overridableHttpConfig->MEMBER; conv = &HttpStatusCodeList::Conv; break;
+#define _CONF_CASE_ConnectionTracker_MIN_SERVER_CONV(KEY, MEMBER)               \
+  case TS_CONFIG_##KEY: ret = &overridableHttpConfig->MEMBER; conv = &ConnectionTracker::MIN_SERVER_CONV; break;
+#define _CONF_CASE_ConnectionTracker_MAX_SERVER_CONV(KEY, MEMBER)               \
+  case TS_CONFIG_##KEY: ret = &overridableHttpConfig->MEMBER; conv = &ConnectionTracker::MAX_SERVER_CONV; break;
+#define _CONF_CASE_ConnectionTracker_SERVER_MATCH_CONV(KEY, MEMBER)             \
+  case TS_CONFIG_##KEY: ret = &overridableHttpConfig->MEMBER; conv = &ConnectionTracker::SERVER_MATCH_CONV; break;
+#define _CONF_CASE_HttpTransact_HOST_RES_CONV(KEY, MEMBER)                      \
+  case TS_CONFIG_##KEY: ret = &overridableHttpConfig->MEMBER; conv = &HttpTransact::HOST_RES_CONV; break;
+#define _CONF_CASE_DISPATCH(KEY, MEMBER, RECORD_NAME, DATA_TYPE, CONV) _CONF_CASE_##CONV(KEY, MEMBER)
```
masaori335, line 7293: This macro seems really complicated. Can we generate a unordered_map of key and converter? Then lookup converter and dispatch it?

What I'm worrying is when we need another special converter, we need to add another macro here to generate the switch case.

bneradt: I agree this is complicated. :)

I took a crack at using a map, but I think I wound up making it more complicated.
https://github.com/bneradt/trafficserver/tree/streamline_overridable_map

How about this: I beefed up the documentation quite a bit on this. Does that help? Note that the great majority of people won't have to deal with custom converters and this macro will just take care of things for them automatically.

masaori335: Thank you for trying the idea. Comments for custom converter is helpful!

PR #12722, opened 2025-12-03
stats_over_http: fix incorrect use of C string functions

bneradt:
The strstr() calls were being passed strings returned from TSMimeHdrFieldValueStringGet, which returns a non-null-terminated string. This could lead to reading uninitialized memory.

Fix this by using std::string_view with its find() method to search within the non-null-terminated string without requiring a copy.

This takes over the work started by @vuori in #12634.


cmcfarlen:

PR #12725, opened 2025-12-03
Cripts: tighten context cleanup

moonchen:
There were some missing reset() calls for Cripts::Context that cause state not to be cleaned up between hooks on a single transaction.


zwoop (approved):

cmcfarlen:

Issue #12611, opened 2025-10-24
retry fails if api_server_addr_set is true


Coder-Xu777:
In trafficserver 9.2.x, I have some doubts about the logic of the following code snippet.
https://github.com/apache/trafficserver/blob/9.2.x/proxy/http/HttpTransact.cc :3864
```
      if (s->api_server_addr_set) {
        // If the plugin set a server address, back up to the OS_DNS hook
        // to let it try another one. Force OS_ADDR_USE_CLIENT so that
        // in OSDNSLoopkup, we back up to how_to_open_connections which
        // will tell HttpSM to connect the origin server.

        s->dns_info.os_addr_style = DNSLookupInfo::OS_Addr::OS_ADDR_USE_CLIENT;
        TRANSACT_RETURN(SM_ACTION_API_OS_DNS, OSDNSLookup);
      }
```
The comments said we can try another one in  OS_DNS hook, but there is no way to change the value of dns_info.lookup_success unless we add new api and it seems that none dns lookup operation could be performed during this process. Consequently, when OSDNSLookup is invoked, dns_info.lookup_success just keep false causing a failure ultimately.

PR #12727, opened 2025-12-03
Add test for TSHttpTxnServerAddrSet retry behavior (issue #12611)

bryancall:
This PR adds a test that demonstrates issue #12611: TSHttpTxnServerAddrSet() retry fails because the OS_DNS hook is not called again when connection fails.

## Test Description

The test plugin (`test_TSHttpTxnServerAddrSet_retry`):
- Hooks into `TS_HTTP_OS_DNS_HOOK`
- On first call: Sets a non-routable address (192.0.2.1:80) that will fail to connect
- On subsequent calls: Sets localhost (127.0.0.1:8080) that should work
- Tracks and logs how many times the OS_DNS hook was called

## Expected Results

**On master (bug exists):**
- OS_DNS hook is called only ONCE
- Connection fails with the bad address
- Plugin never gets a chance to provide an alternative address
- Test logs: `BUG CONFIRMED: OS_DNS hook was only called ONCE`

**After fix is applied:**
- OS_DNS hook should be called multiple times on retry
- Plugin can provide a different address
- Test logs: `SUCCESS: OS_DNS hook was called X times`

## Test Output on Master (Confirms Bug)

```
[TSHttpTxnServerAddrSet_retry] OS_DNS hook called, count=1
[TSHttpTxnServerAddrSet_retry] Attempt 1: Setting BAD address 192.0.2.1:80 (will fail)
[TSHttpTxnServerAddrSet_retry] Transaction closing. OS_DNS was called 1 time(s)
[TSHttpTxnServerAddrSet_retry] *** BUG CONFIRMED: OS_DNS hook was only called ONCE. ***
```

This test is a precursor to the fix PR for #12611.

### DIFF
--- a/tests/gold_tests/pluginTest/tsapi/test_TSHttpTxnServerAddrSet_retry.test.py
+++ b/tests/gold_tests/pluginTest/tsapi/test_TSHttpTxnServerAddrSet_retry.test.py
@@ -75,9 +75,9 @@ class TestIssue12611:
 
         tr.StillRunningAfter = self._ts
 
-        # Check the diags.log for our diagnostic messages
-        # After fix: OS_DNS should be called multiple times on retry
-        self._ts.Disk.diags_log.Content += Testers.ContainsExpression("OS_DNS hook called, count=1", "First OS_DNS call logged")
+        # Override the default diags.log error check - we use TSError() for test output
+        # Using '=' instead of '+=' replaces the default "no errors" check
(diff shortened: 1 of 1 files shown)


bneradt:
[approve ci debian fedora rocky ubuntu]

bneradt:
[approve ci debian fedora rocky ubuntu]

bryancall:
[approve ci]

bryancall:
[approve ci]

bryancall:
This PR was just for testing and seeing if the autest would fail correctly in CI.  It was helpful in finding an issue with CI and not correctly failing.  The new PR that includes the fix and this test is: #12733

PR #12729, opened 2025-12-03
Fix issue where origins could be unintentionally marked as down

yknoya:
# Problem
https://github.com/apache/trafficserver/pull/9181 introduced an issue where an origin server was marked as down even though a connection had been successfully established.

This issue occurs under the following conditions:
1. `proxy.config.http.server_session_sharing.match` is set to a value other than `none` (i.e., server session reuse is enabled).
2. A server session is reused when connecting to the origin.
3. The connection is closed after sending a request to the origin.
4. Condition 3 occurs repeatedly until it reaches the threshold defined by `proxy.config.http.connect_attempts_rr_retries`.

The issue has been confirmed in the following branches/versions (other versions not tested):
- master (90dbc21)
- 10.1.0
- 9.2.11

# Cause
When ATS begins processing an origin connection, it executes `t_state.set_connect_fail(EIO)` to tentatively set `connect_result` to `EIO`:
https://github.com/apache/trafficserver/blob/90dbc21a541986db6b223649cb4f798e190a550f/src/proxy/http/HttpSM.cc#L8054
https://github.com/apache/trafficserver/blob/90dbc21a541986db6b223649cb4f798e190a550f/include/proxy/http/HttpTransact.h#L932

If server session reuse is not possible, `connect_result` is cleared once the connection is established:
https://github.com/apache/trafficserver/blob/90dbc21a541986db6b223649cb4f798e190a550f/src/proxy/http/HttpSM.cc#L1860

However, when a server session is reused, `connect_result` is not cleared and remains set to `EIO`.
This regression was triggered by the change introduced in  https://github.com/apache/trafficserver/pull/9181 .

Before the PR was merged, `t_state.set_connect_fail(EIO)` was not executed when a server session was reused.
After the PR, it is executed regardless of whether a server session is reused or not.

With `connect_result` incorrectly left as `EIO`, if the connection is closed after sending a request to the origin, the following call chain leads to execution of `HttpSM::mark_host_failure`, causing the `fail_count` to be incremented:
1. https://github.com/apache/trafficserver/blob/90dbc21a541986db6b223649cb4f798e190a550f/src/proxy/http/HttpTransact.cc#L3466
2. https://github.com/apache/trafficserver/blob/90dbc21a541986db6b223649cb4f798e190a550f/src/proxy/http/HttpTransact.cc#L3786
3. https://github.com/apache/trafficserver/blob/90dbc21a541986db6b223649cb4f798e190a550f/src/proxy/http/HttpTransact.cc#L3884
4. https://github.com/apache/trafficserver/blob/90dbc21a541986db6b223649cb4f798e190a550f/src/proxy/http/HttpSM.cc#L4630
5. https://github.com/apache/trafficserver/blob/90dbc21a541986db6b223649cb4f798e190a550f/src/proxy/http/HttpSM.cc#L5876

If this happens repeatedly and reaches the threshold defined by `proxy.config.http.connect_attempts_rr_retries`, the origin server is incorrectly marked as down:
https://github.com/apache/trafficserver/blob/90dbc21a541986db6b223649cb4f798e190a550f/src/proxy/http/HttpSM.cc#L5876-L5885

Since the connection to the origin is actually successful, marking it as down is incorrect.

# Fix
Update the logic so that `t_state.set_connect_fail(EIO)` is executed only when establishing a new connection to the origin (i.e., when a server session is not reused), and ensure that `connect_result` is cleared once the connection succeeds.

Additionally, when `multiplexed_origin` is true, `connect_result` was also not being cleared after a successful connection.
In this case, although `t_state.set_connect_fail(EIO)` is executed (see below), the lack of a corresponding clear operation results in `connect_result` remaining `EIO`:
https://github.com/apache/trafficserver/blob/90dbc21a541986db6b223649cb4f798e190a550f/src/proxy/http/HttpSM.cc#L5706-L5723

This patch ensures that `connect_result` is cleared whenever the connection succeeds, regardless of whether `multiplexed_origin` is enabled.

### DIFF
--- a/src/proxy/http/ConnectingEntry.cc
+++ b/src/proxy/http/ConnectingEntry.cc
@@ -139,7 +139,7 @@ ConnectingEntry::state_http_server_open(int event, void *data)
     while (!connect_sms.empty()) {
       auto entry = connect_sms.begin();
       SCOPED_MUTEX_LOCK(lock, (*entry)->mutex, this_ethread());
-      (*entry)->t_state.set_fail(lerrno);
+      (*entry)->t_state.set_connect_fail(lerrno);
       (*entry)->server_connection_provided_cert = vc_provided_cert;
       (*entry)->handleEvent(event, data);
       connect_sms.erase(entry);
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -1112,7 +1112,7 @@ HttpSM::state_raw_http_server_open(int event, void *data)
 
     netvc->set_inactivity_timeout(get_server_inactivity_timeout());
     netvc->set_active_timeout(get_server_active_timeout());
-    t_state.set_success();
+    t_state.current.server->clear_connect_fail();
 
     if (get_tunnel_type() != SNIRoutingType::NONE) {
       tunnel.mark_tls_tunnel_active();
@@ -1125,9 +1125,9 @@ HttpSM::state_raw_http_server_open(int event, void *data)
   case NET_EVENT_OPEN_FAILED:
     if (t_state.cause_of_death_errno == -UNKNOWN_INTERNAL_ERROR) {
       if (event == VC_EVENT_EOS) {
-        t_state.set_fail(EPIPE);
+        t_state.set_connect_fail(EPIPE);
       } else {
-        t_state.set_fail(EIO);
+        t_state.set_connect_fail(EIO);
       }
     }
     t_state.current.state = HttpTransact::OPEN_RAW_ERROR;
@@ -1822,6 +1822,10 @@ HttpSM::state_http_server_open(int event, void *data)
       // The buffer we create will be handed over to the eventually created server session
       _netvc->do_io_write(this, 1, _netvc_reader);
       _netvc->set_inactivity_timeout(this->get_server_connect_timeout());
+
+      // Pre-emptively set a server connect failure that will be cleared once a WRITE_READY is received from origin or
+      // bytes are received back
+      t_state.set_connect_fail(EIO);
     } else { // in the case of an intercept plugin don't to the connect timeout change
       SMDbg(dbg_ctl_http_connect, "not setting handler for connection handshake");
       this->create_server_txn(this->create_server_session(*_netvc, _netvc_read_buffer, _netvc_reader));
@@ -1840,6 +1844,7 @@ HttpSM::state_http_server_open(int event, void *data)
   case CONNECT_EVENT_TXN:
     SMDbg(dbg_ctl_http, "Connection handshake complete via CONNECT_EVENT_TXN");
     if (this->create_server_txn(static_cast<PoolableSession *>(data))) {
+      t_state.current.server->clear_connect_fail();
       handle_http_server_open();
     } else { // Failed to create transaction.  Maybe too many active transactions already
       // Try again (probably need a bounding counter here)
@@ -1852,11 +1857,12 @@ HttpSM::state_http_server_open(int event, void *data)
     // Update the time out to the regular connection timeout.
     SMDbg(dbg_ctl_http_ss, "Connection handshake complete");
     this->create_server_txn(this->create_server_session(*_netvc, _netvc_read_buffer, _netvc_reader));
+    t_state.current.server->clear_connect_fail();
     handle_http_server_open();
     return 0;
   case VC_EVENT_INACTIVITY_TIMEOUT:
   case VC_EVENT_ACTIVE_TIMEOUT:
-    t_state.set_fail(ETIMEDOUT);
+    t_state.set_connect_fail(ETIMEDOUT);
   /* fallthrough */
   case VC_EVENT_ERROR:
   case VC_EVENT_EOS:
@@ -1865,7 +1871,7 @@ HttpSM::state_http_server_open(int event, void *data)
     t_state.outbound_conn_track_state.clear();
     if (_netvc != nullptr) {
       if (event == VC_EVENT_ERROR || event == NET_EVENT_OPEN_FAILED) {
-        t_state.set_fail(_netvc->lerrno);
+        t_state.set_connect_fail(_netvc->lerrno);
       }
       this->server_connection_provided_cert = _netvc->provided_cert();
       _netvc->do_io_write(nullptr, 0, nullptr);
@@ -2037,7 +2043,7 @@ HttpSM::state_read_server_response_header(int event, void *data)
     if (allow_error == false) {
       SMDbg(dbg_ctl_http_seq, "Error parsing server response header");
       t_state.current.state = HttpTransact::PARSE_ERROR;
-      t_state.set_fail(EBADMSG);
+      t_state.set_connect_fail(EBADMSG);
 
       // If the server closed prematurely on us, use the
       //   server setup error routine since it will forward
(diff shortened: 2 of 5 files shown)


src/proxy/http/HttpSM.cc
```diff
@@ -1823,6 +1823,9 @@ HttpSM::state_http_server_open(int event, void *data)
       _netvc->do_io_write(this, 1, _netvc_reader);
       _netvc->set_inactivity_timeout(this->get_server_connect_timeout());
 
+      // Pre-emptively set a server connect failure that will be cleared once a WRITE_READY is received from origin or
+      // bytes are received back
+      t_state.set_connect_fail(EIO);
```
masaori335: I guess you're just moving existing code. However, can we set `EIO` only if we observe errors? IMO, we easily make a mistake with current approach.

yknoya: Thanks for the review.  
I also thought it would be better to set `EIO` only when an actual error is detected, but due to time constraints I applied the current workaround-like approach.  
I'll look into whether that approach is feasible, so please give me some more time.

masaori335: Let me know if you find it's tough. This approach is not ideal but fixes the bug, so I don't want to block for long time. We can clean it up later.

yknoya: Sorry for the delay.  
I investigated the points you raised and implemented some changes.  
Below is a summary of the investigation performed during the modification.

First, I found that the master branch already contains logic that invokes the `set_connect_fail` method when a connection attempt fails:  
https://github.com/apache/trafficserver/blob/90dbc21a541986db6b223649cb4f798e190a550f/src/proxy/http/ConnectingEntry.cc#L142  
https://github.com/apache/trafficserver/blob/90dbc21a541986db6b223649cb4f798e190a550f/src/proxy/http/HttpSM.cc#L1874  
https://github.com/apache/trafficserver/blob/90dbc21a541986db6b223649cb4f798e190a550f/src/proxy/http/HttpSM.cc#L1130-L1136  

Since `set_connect_fail` is executed when a connection actually fails, it seemed unnecessary to pre-set `EIO` before the connection is made, so I looked into the reasoning.  
My assumption is that the pre-set `EIO` exists to avoid updating `connect_result` when `set_connect_fail` is invoked **after** the connection has already succeeded.  
There are several locations where `set_connect_fail` may run after the connection is established, but due to the following logic, only `cause_of_death_errno` is updated while `connect_result` is left unchanged.  
This is because `connect_result` is set to 0 after the connection succeeds.  
https://github.com/apache/trafficserver/blob/90dbc21a541986db6b223649cb4f798e190a550f/include/proxy/http/HttpTransact.h#L931-L936  

Based on these findings, I identified the following issues:
1. The `set_connect_fail` method is used not only for connection-related failures but also for other types of errors (its name no longer matches its actual behavior).  
2. The logic determines whether the error occurred during connection by relying on a pre-set `EIO`, which is not ideal.

I was not able to resolve these issues perfectly cleanly, but I believe improvements are possible, and I applied the following changes:
- Renamed `set_connect_fail` to `set_fail`.  
- Changed the initial value of `connect_result` to `-UNKNOWN_INTERNAL_ERROR` (instead of `EIO`), matching the initial value of `cause_of_death_errno` and making the intent clearer.  
- Updated `set_fail` so that it determines whether the failure occurred during connection by checking `next_action` and the value of `connect_result`.  
- Removed the logic that pre-sets `EIO` before a connection attempt.

Additionally, although slightly outside the main issue, I made the following related improvements:
- Added a `set_success` method to reset both `connect_result` and `cause_of_death_errno`, since callers of `clear_connect_fail` were not resetting `cause_of_death_errno`.  
- Renamed `clear_connect_fail` to `set_connect_success`.  
- Ensured that `set_success` is always invoked upon successful connection by calling it within `handle_http_server_open`.

src/proxy/http/ConnectingEntry.cc
```diff
@@ -139,7 +139,9 @@ ConnectingEntry::state_http_server_open(int event, void *data)
     while (!connect_sms.empty()) {
       auto entry = connect_sms.begin();
       SCOPED_MUTEX_LOCK(lock, (*entry)->mutex, this_ethread());
-      (*entry)->t_state.set_connect_fail(lerrno);
+      if (lerrno != -UNKNOWN_INTERNAL_ERROR) {
+        (*entry)->t_state.set_fail(lerrno);
+      }
```
yknoya: In cases where a connection timeout occurs, the code path temporarily assigned `EIO` (5) before ultimately setting `ETIMEDOUT` (110), as shown in the logs below.
Assigning an incorrect intermediate errno is undesirable, so this change ensures that an inappropriate errno is not set during the transition.
```
[Dec 16 15:23:26.800] [ET_NET 0] DIAG: <ConnectingEntry.cc:48 (state_http_server_open)> (http_connect) entered inside ConnectingEntry::state_http_server_open
[Dec 16 15:23:26.800] [ET_NET 0] DIAG: <ConnectingEntry.cc:130 (state_http_server_open)> (http_connect) Stop 1 state machines waiting for failed origin
[Dec 16 15:23:26.800] [ET_NET 0] DIAG: <TLSEventSupport.cc:153 (callHooks)> (ssl) sslHandshakeHookState=TS_SSL_HOOK_PRE_CONNECT eventID=110
[Dec 16 15:23:26.800] [ET_NET 0] DIAG: <TLSEventSupport.cc:271 (callHooks)> (ssl) iterated to curHook=(nil)
[Dec 16 15:23:26.800] [ET_NET 0] DIAG: <HttpTransact.h:933 (set_fail)> (http) Setting upstream connection failure -19999 to 5
[Dec 16 15:23:26.800] [ET_NET 0] DIAG: <HttpSM.cc:2645 (main_handler)> (http) [0] VC_EVENT_INACTIVITY_TIMEOUT/TS_EVENT_VCONN_INACTIVITY_TIMEOUT, 105
[Dec 16 15:23:26.800] [ET_NET 0] DIAG: <HttpSM.cc:1796 (state_http_server_open)> (http_track) [0] entered inside state_http_server_open: VC_EVENT_INACTIVITY_TIMEOUT/TS_EVENT_VCONN_INACTIVITY_TIMEOUT
[Dec 16 15:23:26.800] [ET_NET 0] DIAG: <HttpSM.cc:1797 (state_http_server_open)> (http) [0] [&HttpSM::state_http_server_open, VC_EVENT_INACTIVITY_TIMEOUT/TS_EVENT_VCONN_INACTIVITY_TIMEOUT]
[Dec 16 15:23:26.800] [ET_NET 0] DIAG: <HttpTransact.h:933 (set_fail)> (http) Setting upstream connection failure 5 to 110
[Dec 16 15:23:26.800] [ET_NET 0] DIAG: <HttpTransact.cc:3432 (HandleResponse)> (http_trans) [0] Entering HttpTransact::HandleResponse
```

masaori335:
Thank you for cleaning up code. However, it's getting big than expected. Let's make another PR for cleanup and focus on bug fix on this PR. 

IMO, touching `connect_result ` and `cause_of_death_errno ` in the `set_connect_fail()` function is making things complicated. Anyway, let's talk this more in another PR.

yknoya:
We decided to split this PR into two parts: a bug fix and a cleanup.
The following commits will be reverted and moved to a separate PR:
- https://github.com/apache/trafficserver/pull/12729/commits/bd1c1920e3862d0dd4fa0e1f7771b3fa8b0572fc
- https://github.com/apache/trafficserver/pull/12729/commits/76cd57e8b574a7d0ff0d66fc784a9d0e97c460f5

bneradt:
[approve ci]

PR #12730, opened 2025-12-03
regex_remap: Document host option // prefix behavior

bneradt:
The host option prepends // to the match string, which can be unintuitive when regex_remap rules. This adds documentation explaining why the // prefix exists (to distinguish hostnames from hostname-like path components) and provides practical examples showing how to write redirect rules correctly using https:$1 instead of https://$1.

Fixes: #12691

### DIFF
--- a/doc/admin-guide/plugins/regex_remap.en.rst
+++ b/doc/admin-guide/plugins/regex_remap.en.rst
@@ -217,7 +217,9 @@ a malformed URL with four slashes::
     # redirect.conf:
     ^(.*)$    https:$1    @status=307
 
-This will redirect ``http://example.com/path`` to ``https://example.com/path``.
+This will redirect any plaintext HTTP (``http://``) request for any host to the
+``https://`` equivalent. For instance, a request to ``http://example.com/path``
+would be redirected to ``https://example.com/path``.
 
 **Example: Method-based routing**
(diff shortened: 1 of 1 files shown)


doc/admin-guide/plugins/regex_remap.en.rst
```diff
@@ -171,3 +182,54 @@ as a redirect (Location:).
...
+the ``//``, the substitution uses ``https:`` without ``//`` to avoid producing
+a malformed URL with four slashes::
+
+    # remap.config:
+    regex_map http://(.*) https://unused.invalid @plugin=regex_remap.so @pparam=redirect.conf @pparam=host @pparam=pristine
+
+    # redirect.conf:
+    ^(.*)$    https:$1    @status=307
+
+This will redirect ``http://example.com/path`` to ``https://example.com/path``.
```
zwoop: Minor detail, but should we say that this matches any fqdn/hostname, and rewrites it to https://? I mean it kinda reads like it's only doing it for example.com.

bneradt: Good point. I updated and pushed a commit on top of this rewording the paragraph to make this clearer.

cmcfarlen:

PR #12731, opened 2025-12-04
Catch2: Use CAPTURE and GENERATE

bneradt:
Add CAPTURE statements to parameterized tests to improve failure diagnostics. Convert many for-loop based parameterized tests to use GENERATE which provides automatic sectioning, continues running after failures, and produces clearer test output.

### DIFF
--- a/src/proxy/hdrs/unit_tests/test_Hdrs.cc
+++ b/src/proxy/hdrs/unit_tests/test_Hdrs.cc
@@ -21,6 +21,7 @@
 
 #include <string>
 #include <string_view>
+#include <vector>
 #include <cstring>
 #include <cctype>
 #include <bitset>
@@ -38,6 +39,8 @@ using namespace std::literals;
 #include "tsutil/PostScript.h"
 
 #include <catch2/catch_test_macros.hpp>
+#include <catch2/generators/catch_generators.hpp>
+#include <catch2/generators/catch_generators_range.hpp>
 
 #include "proxy/hdrs/HTTP.h"
 #include "proxy/hdrs/HttpCompat.h"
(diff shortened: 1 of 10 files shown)


src/proxy/hdrs/unit_tests/test_Hdrs.cc
```diff
@@ -50,58 +53,55 @@ TEST_CASE("HdrTestHttpParse", "[proxy][hdrtest]")
...
-  for (auto const &test : tests) {
-    HTTPHdr  req_hdr;
-    HdrHeap *heap = new_HdrHeap(HdrHeap::DEFAULT_SIZE + 64); // extra to prevent proxy allocation.
-
-    req_hdr.create(HTTPType::REQUEST, HTTP_1_1, heap);
+  HTTPHdr  req_hdr;
+  HdrHeap *heap = new_HdrHeap(HdrHeap::DEFAULT_SIZE + 64); // extra to prevent proxy allocation.
 
-    http_parser_clear(&parser);
+  req_hdr.create(HTTPType::REQUEST, HTTP_1_1, heap);
```
bneradt, line 95: Note that this http_parser_clear and code like this isn't needed now that GENERATE creates its own section for each which starts state over between iterations.

src/proxy/hdrs/unit_tests/test_URL.cc
```diff
@@ -76,16 +80,9 @@ TEST_CASE("Validate Scheme", "[proxy][validscheme]")
...
-  for (auto i : scheme_test_cases) {
-    // it's pretty hard to debug with
-    //     CHECK(validate_scheme(i.text) == i.valid);
-
-    std::string_view text = i.text;
-    if (validate_scheme(text) != i.valid) {
-      std::printf("Validation of scheme: \"%s\", expected %s, but not\n", text.data(), (i.valid ? "true" : "false"));
-      CHECK(false);
-    }
-  }
```
bneradt, line 88: Note that the workarounds like this aren't necessary anymore with the CAPTURE.

cmcfarlen (approved):
Very cool.  Thanks!

PR #12732, opened 2025-12-04
Fix autest.sh.in to preserve autest exit code

bryancall:
The trailing if-blocks for restoring UDS test directories were overwriting autest's exit code with 0, causing test failures to be silently ignored in CI.

Now we capture the exit code immediately after autest runs and exit with it at the end of the script.

This was discovered while investigating why PR #12727 was passing CI despite having a failing test.


cmcfarlen:

Issue #12292, opened 2025-06-16
MgmtConverter runs on remapping


masaori335:
Some overridable configs are running converter on remapping. It'd be great if they're converted on loading configs.

- e.g. `HttpTransact::HOST_RES_CONV` for `proxy.config.hostdb.ip_resolve`
```
Process 2881 stopped
* thread #12, name = '[ET_NET 9]', stop reason = breakpoint 1.1
    frame #0: 0x000000010013e0e0 traffic_server`HttpTransact::$_1::operator()(this=0x0000000105291917, data=0x00000001053c83a8, src="ipv4") const at HttpTransact.cc:92:81
   89  	                                                },
   90  	                                                [](void *data, std::string_view src) -> void {
   91  	                                                  HostResData *res_data = static_cast<HostResData *>(data);
-> 92  	                                                  parse_host_res_preference(src.data(), res_data->order);
   93  	                                                }};
   94  	
   95  	static char range_type[] = "multipart/byteranges; boundary=RANGE_SEPARATOR";
Target 0: (traffic_server) stopped.
(lldb) bt
* thread #12, name = '[ET_NET 9]', stop reason = breakpoint 1.1
  * frame #0: 0x000000010013e0e0 traffic_server`HttpTransact::$_1::operator()(this=0x0000000105291917, data=0x00000001053c83a8, src="ipv4") const at HttpTransact.cc:92:81
    frame #1: 0x000000010013e0a8 traffic_server`HttpTransact::$_1::__invoke(data=0x00000001053c83a8, src="ipv4") at HttpTransact.cc:90:49
    frame #2: 0x000000010154ee80 libtsapi.dylib`TSHttpTxnConfigStringSet(txnp=0x00000001053c7000, conf=TS_CONFIG_HTTP_HOST_RESOLUTION_PREFERENCE, value="ipv4", length=4) at InkAPI.cc:7668:9
    frame #3: 0x000000010b146a44 conf_remap.so`TSRemapDoRemap(ih=0x000000010ada0000, rh=0x00000001053c7000, (null)=0x0000000105291fc8) at conf_remap.cc:346:9
    frame #4: 0x00000001001ab778 traffic_server`RemapPluginInfo::doRemap(this=0x0000000101f46920, ih=0x000000010ada0000, rh=0x00000001053c7000, rri=0x0000000105291fc8) at RemapPluginInfo.cc:229:14
[...]

PR #12734, opened 2025-12-05
Add pre-conversion API for overridable string configs

bneradt:
Introduce TSConfigStringPreconvert() and
TSHttpTxnConfigPreconvertedValueSet() to allow plugins to parse string configuration values once at load time rather than on every transaction. This eliminates repeated parsing overhead for configs like proxy.config.hostdb.ip_resolve and
proxy.config.http.negative_caching_list when applied via conf_remap. The API supports nullptr to represent NULL/unset values distinct from empty strings.

Fixes: #12292

---

Marking as a draft while I send an email to the dev mailing list about the new API.


bneradt:
I prefer this caching solution over this:
#12735

Rather than adding an interface that the plugin writer has to use, the latter patch caches the converted values and uses those when previously computed.

PR #12735, opened 2025-12-05
Automatic caching of parsed STRING config values

bneradt:
This introduces ParsedConfigCache in HttpConfig to automatically cache parsed results for STRING configs that require expensive parsing. When TSHttpTxnConfigStringSet() is called for configs like negative_caching_list, insert_forwarded, server_session_sharing.match, negative_revalidating_list, or host_res_data (ip_resolve), the parsing now happens once per unique value and is cached for subsequent calls. This optimization is transparent to plugins - they call TSHttpTxnConfigStringSet() as usual and automatically benefit from the caching.

Fixes: #12292

### DIFF
--- a/include/proxy/http/HttpConfig.h
+++ b/include/proxy/http/HttpConfig.h
@@ -41,8 +41,8 @@
 #include <cctype>
 #include <string_view>
 #include <chrono>
-#include <shared_mutex>
 #include <functional>
+#include <variant>
 
 #include "iocore/eventsystem/IOBuffer.h"
 #include "swoc/swoc_ip.h"
@@ -57,6 +57,7 @@
 #include "iocore/net/ConnectionTracker.h"
 #include "iocore/net/SessionSharingAPIEnums.h"
 #include "records/RecProcess.h"
+#include "tsutil/Bravo.h"
 #include "tsutil/ts_ip.h"
 #include "tsutil/Metrics.h"
 #include "ts/apidefs.h"
@@ -875,13 +876,15 @@ public:
 class ParsedConfigCache
 {
 public:
-  /** Pre-parsed representations for configs that need special parsing. */
+  /** Pre-parsed representations for configs that need special parsing.
+   *
+   * Uses std::variant since each cache entry only stores one type of parsed
+   * result. The conf_value_storage string is always present as it owns the
+   * string data that the parsed structures may reference.
+   */
   struct ParsedValue {
-    HostResData                 host_res_data{};
-    HttpStatusCodeList          status_code_list{};
-    HttpForwarded::OptionBitSet forwarded_bitset{};
-    MgmtByte                    server_session_sharing_match{0};
-    std::string                 conf_value_storage{}; // Owns the string data.
+    std::string conf_value_storage{}; // Owns the string data.
+    std::variant<std::monostate, HostResData, HttpStatusCodeList, HttpForwarded::OptionBitSet, MgmtByte> parsed{};
   };
 
   /** Return the parsed value for the configuration.
@@ -919,7 +922,7 @@ private:
   };
 
   std::unordered_map<std::pair<TSOverridableConfigKey, std::string>, ParsedValue, CacheKeyHash> _cache;
-  mutable std::shared_mutex                                                                     _mutex;
+  ts::bravo::shared_mutex                                                                       _mutex;
 };
 
 /////////////////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


src/api/InkAPI.cc
```diff
@@ -7527,23 +7536,22 @@ TSHttpTxnConfigStringSet(TSHttpTxn txnp, TSOverridableConfigKey conf, const char
     break;
   case TS_CONFIG_HTTP_NEGATIVE_CACHING_LIST:
     if (value && length > 0) {
-      OverridableHttpConfigParams *target      = &s->t_state.my_txn_conf();
-      target->negative_caching_list.conf_value = const_cast<char *>(value);
-      return _eval_conv(target, conf, value, length);
+      auto &parsed                                   = ParsedConfigCache::lookup(conf, std::string_view(value, length));
+      s->t_state.my_txn_conf().negative_caching_list = parsed.status_code_list;
```
bneradt: These `ParsedConfigCache::lookup` are the high level view of where the value of this patch happens. Instead of calling the conversion logic for each of these "complex" strings (like the negative caching status list, in this example) every transaction, this parsed value retrieval is called through an interface that caches the results so that the compution is only done once.

bneradt:
[approve ci fedora]

include/proxy/http/HttpConfig.h
```diff
@@ -852,6 +856,72 @@ struct HttpConfigParams : public ConfigInfo {
...
+  struct CacheKeyHash {
+    std::size_t
+    operator()(const std::pair<TSOverridableConfigKey, std::string> &k) const
+    {
+      return std::hash<int>()(static_cast<int>(k.first)) ^ (std::hash<std::string>()(k.second) << 1);
+    }
+  };
+
+  std::unordered_map<std::pair<TSOverridableConfigKey, std::string>, ParsedValue, CacheKeyHash> _cache;
+  mutable std::shared_mutex                                                                     _mutex;
```
masaori335: `ts::bravo::shared_mutex` might be faster? Because this map is read heavy.

include/proxy/http/HttpConfig.h
```diff
@@ -852,6 +856,72 @@ struct HttpConfigParams : public ConfigInfo {
...
+ * value don't re-parse.
+ *
+ * The static lookup() method handles everything: check cache, parse if needed,
+ * store in cache, and return the result.
+ */
+class ParsedConfigCache
+{
+public:
+  /** Pre-parsed representations for configs that need special parsing. */
+  struct ParsedValue {
```
masaori335, line 885: Kind of nit-picking, `union` or `std::variant` is better? Because we only use one of these members.

include/proxy/http/HttpConfig.h
```diff
@@ -852,6 +856,72 @@ struct HttpConfigParams : public ConfigInfo {
...
+
+  const ParsedValue &lookup_impl(TSOverridableConfigKey key, std::string_view value);
+  ParsedValue        parse(TSOverridableConfigKey key, std::string_view value);
+
+  // Custom hash for the cache key.
+  struct CacheKeyHash {
+    std::size_t
+    operator()(const std::pair<TSOverridableConfigKey, std::string> &k) const
+    {
+      return std::hash<int>()(static_cast<int>(k.first)) ^ (std::hash<std::string>()(k.second) << 1);
```
masaori335, line 920: I guess this 1 bit left-shift is trying to avoid collision. Because If hash of `k.first` and `k.second` is the same value, it'll be 0, right?

~I'm a bit wondering if we should shift more than 1 bit. The length of return value of `std::hash` is fixed?~

include/proxy/http/HttpConfig.h
bneradt: Good idea. I forgot about bravo.

include/proxy/http/HttpConfig.h
bneradt: True. Sematically, std::variant is nice because it communicates explicitly that only one is used at a time.

include/proxy/http/HttpConfig.h
bneradt: Absolutely right.

The 1-bit shift breaks the symmetry of XOR to avoid collisions when both hashes happen to be equal (which would XOR to 0) and to reduce collision patterns from the commutative property of XOR. The shift amount doesn't need to be larger than 1 — any non-zero shift achieves the asymmetry goal.

Issue #12728, opened 2025-12-03
Regression tests with -R 3 crashes


bneradt:
After @bryancall fixed a hang in the regression tests (#12721) we enabled `-R 3` in CI but faced crashes in ATS with debian, fedora, rocky, and ubuntu:

```
+ /tmp/ats/bin/traffic_server -K -R 3
[Dec  3 21:19:31.553] traffic_server NOTE: records parsing completed.
[Dec  3 21:19:31.553] traffic_server NOTE: /tmp/ats/etc/trafficserver/records.yaml finished loading
[Dec  3 21:19:31.558] traffic_server NOTE: Traffic Server is running unprivileged, not switching to user 'nobody'
[Dec  3 21:19:31.646] [ET_NET 0] NOTE: hosting.config loading ...
[Dec  3 21:19:31.647] [ET_NET 0] NOTE: hosting.config finished loading
[Dec  3 21:19:31.647] [ET_NET 0] NOTE: cache enabled
[Dec  3 21:19:31.647] [ET_NET 0] NOTE: Traffic Server is fully initialized.
REGRESSION_TEST initialization begun
REGRESSION TEST SSLReverseHostname started
    REGRESSION_RESULT SSLReverseHostname:                       PASSED
REGRESSION TEST SSLWildcardMatch started
    REGRESSION_RESULT SSLWildcardMatch:                         PASSED
REGRESSION TEST RegressionSM started
    REGRESSION_RESULT RegressionSM:                             PASSED
REGRESSION TEST ram_cache started
RPRINT ram_cache: RamCache LRU Fixed Size Hit Rate 0.610229
RPRINT ram_cache: RamCache LRU Variable Size Hit Rate 0.556396
RPRINT ram_cache: RamCache LRU Nominal Size 1048576 Size 1046544
RPRINT ram_cache: RamCache LRU Test Done
RPRINT ram_cache: RamCache CLFUS Fixed Size Hit Rate 0.604370
RPRINT ram_cache: RamCache CLFUS Variable Size Hit Rate 0.638306
RPRINT ram_cache: RamCache CLFUS Nominal Size 1048576 Size 1040032
RPRINT ram_cache: RamCache CLFUS Test Done
RPRINT ram_cache: RamCache LRU Fixed Size Hit Rate 0.807304
RPRINT ram_cache: RamCache LRU Variable Size Hit Rate 0.795143
RPRINT ram_cache: RamCache LRU Nominal Size 16777216 Size 16758960
[...]

PR #12737, opened 2025-12-06
regression tests: reduce ram_cache test scope for CI stability

bneradt:
The ram_cache test was failing in CI when run with -R 3 due to the 256MB cache size test running over 4 million iterations. This caused timeouts or OOM kills in slower CI environments. Reduced the maximum cache size from 256MB to 16MB and lowered the minimum regression level from EXTENDED to NIGHTLY.

Fixes: #12728


cmcfarlen:

PR #12741, opened 2025-12-09
Fix HTTP caching compliance by adding Vary headers for compressible content

mlibbey:
The compress plugin was only adding Vary: Accept-Encoding headers when content was actually going to be compressed, not when content could be compressed. This can cause downstream caches to never get the compressed version in cache.

Now adds Vary: Accept-Encoding headers for all compressible content regardless of whether compression is applied, ensuring proper HTTP cache behavior.

### DIFF
--- a/plugins/compress/compress.cc
+++ b/plugins/compress/compress.cc
@@ -330,7 +330,7 @@ compress_transform_init(TSCont contp, Data *data)
   }
 
   if (content_encoding_header(bufp, hdr_loc, data->compression_type, data->compression_algorithms) == TS_SUCCESS &&
-      vary_header(bufp, hdr_loc) == TS_SUCCESS && etag_header(bufp, hdr_loc) == TS_SUCCESS) {
+      etag_header(bufp, hdr_loc) == TS_SUCCESS) {
     downstream_conn         = TSTransformOutputVConnGet(contp);
     data->downstream_buffer = TSIOBufferCreate();
     data->downstream_reader = TSIOBufferReaderAlloc(data->downstream_buffer);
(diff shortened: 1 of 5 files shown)


mlibbey:
@JakeChampion because of big recent changes to compress.

bneradt:
In case it's helpful:
https://ci.trafficserver.apache.org/job/Github_Builds/job/centos/5791/console

```
../../../plugins/compress/compress.cc: In function 'void Compress::{anonymous}::handle_range_request(TSMBuffer, TSMLoc, const Compress::HostConfiguration*)':
../../../plugins/compress/compress.cc:100:35: error: passing 'const Compress::HostConfiguration' as 'this' argument discards qualifiers [-fpermissive]
  100 |     switch (hc->range_request_ctl()) {
      |                                   ^
In file included from ../../../plugins/compress/compress_common.h:41,
                 from ../../../plugins/compress/compress.cc:36:
../../../plugins/compress/configuration.h:94:3: note:   in call to 'Compress::RangeRequestCtrl Compress::HostConfiguration::range_request_ctl()'
   94 |   range_request_ctl()
      |   ^~~~~~~~~~~~~~~~~
```

mlibbey:
Sorry for the initial build failures -- it was building locally :( (The Rocky crash looks to be in a cache subsystem regression test).
Think it really ready for review.

cmcfarlen (approved):
Seems good.  I looked for, but did didn't notice any, missing API handle cleanup.

cmcfarlen:
Added to 10.1.1 milestone via #12764

PR #12742, opened 2025-12-09
xdebug: probe-full-json fix single quote escapes

bneradt:
The xdebug plugin was incorrectly escaping single quotes as \' in probe-full-json output, which is not a valid JSON escape sequence and broke JSON parsing for headers like Content-Security-Policy that contain CSP directives with single quotes (e.g., 'self', 'unsafe-inline'). This fix modifies the EscapeCharForJson class to only escape single quotes in legacy probe format (which uses single-quoted strings) while leaving them unescaped in full JSON mode, producing RFC 8259-compliant JSON output that can be piped directly to tools like jq.


plugins/xdebug/xdebug_escape.cc
```diff
@@ -0,0 +1,102 @@
...
+  case '\b':
+    return {"\\b"};
+  case '\f':
+    return {"\\f"};
+  case '\t':
+    return {"\\t"};
+  default:
+    return {&c, 1};
+  }
+}
```
bneradt, line 74: This got moved to it's own file for unit testability reasons. But the only functional change in this patch is that `\'` is no longer escaped.

plugins/xdebug/xdebug_headers.cc
```diff
@@ -38,109 +38,6 @@ namespace
...
-      return result;
-    }
-    if (BEFORE_NAME == _state) {
-      _state = IN_NAME;
-    } else if (BEFORE_VALUE == _state) {
-      _state = IN_VALUE;
-    }
-    switch (c) {
-    case '\'':
-      return {"\\\'"};
```
bneradt, line 82: This is not in the new switch (see the xdebug_escape.cc version).

masaori335 (approved):
Looks good.

PR #12744, opened 2025-12-10
Skip lua_proxy_protocol test if curl command is old

maskit:
`--haproxy-clientip` flag for curl command is only available version 8.2.0 or above.

We have the check on test_TSVConnPPInfo.
https://github.com/apache/trafficserver/blob/163056e768a147822d0e2810b48fc1d15ccf0097/tests/gold_tests/pluginTest/tsapi/test_TSVConnPPInfo.test.py#L25-L28


shukitchan (dismissed):

maskit:
I didn't know we can check the options as well. I updated the tests with `HasCurlOption".

PR #12748, opened 2025-12-10
Fix http2 content-length header check (#12747)

### LINKED ISSUE #12747: HTTP2 HEAD request fails when Content-Length header is non-zero
When doing a HEAD request to an HTTP2 origin, if the origin response contains "Content-Length" header with a non-zero value, it fails with 502 error.
This seems to be because of a check inside http2 handling which rejects request when "Content-Length" header and actual payload length don't match.

However, it is specified in [rfc7540#section-8.1.2.6](https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.6) that `"A response that is defined to have no payload, as described in [RFC7230], Section 3.3.2, can have a non-zero content-length header field, even though no content is included in DATA frames."`

The issue seems to be in [includes/proxy/http2/Http2Stream.h in payload_length_is_valid_fuction](https://github.com/apache/trafficserver/blob/8bdb73ec298ce360e1a6dd74572a9897d445c58d/include/proxy/http2/Http2Stream.h#L390) which does not take this case into account
```C++
inline bool
Http2Stream::payload_length_is_valid() const
{
  uint32_t content_length = _receive_header.get_content_length();
  if (content_length != 0 && content_length != data_length) {
    Warning("Bad payload length content_length=%d data_legnth=%d session_id=%" PRId64, content_length,
            static_cast<int>(data_length), _proxy_ssn->connection_id());
  }
  return content_length == 0 || content_length == data_length;
}
```

###### Working GET query

```
floom@floom-pc:~$ curl -v https://my-test-delivery-service.ats.my-domain.com
...
< HTTP/2 200 
< server: nginx/1.24.0 (Ubuntu)
< date: Tue, 09 Dec 2025 17:29:37 GMT
< content-type: text/html
< content-length: 615
< last-modified: Fri, 14 Feb 2025 12:00:53 GMT
< etag: "67af3075-267"
< accept-ranges: bytes
< age: 0
< x-cache: skipped
< 
* TLSv1.2 (IN), TLS header, Supplemental data (23):
<!DOCTYPE html>
<html>
<head>
<title>Welcome to nginx!</title>
<style>
html { color-scheme: light dark; }
body { width: 35em; margin: 0 auto;
font-family: Tahoma, Verdana, Arial, sans-serif; }
</style>
</head>
<body>
<h1>Welcome to Inst1!</h1>
<p>If you see this page, the nginx web server is successfully installed and
working. Further configuration is required.</p>

<p>For online documentation and support please refer to
<a href="http://nginx.org/">nginx.org</a>.<br/>
Commercial support is available at
<a href="http://nginx.com/">nginx.com</a>.</p>

<p><em>Thank you for using nginx.</em></p>
</body>
</html>
```

###### Failing HEAD query
```
floom@floom-pc:~$ curl -I https://my-test-delivery-service.ats.my-domain.com
HTTP/2 502 
date: Tue, 09 Dec 2025 17:21:51 GMT
cache-control: no-store
content-type: text/html
content-language: en
x-cache: skipped
content-length: 294
```

###### Same HEAD query directly to origin
```
floom@floom-pc:~$ curl -I http://192.168.1.150
HTTP/2 200 OK
server: nginx/1.24.0 (Ubuntu)
date: Tue, 09 Dec 2025 17:28:16 GMT
content-type: text/html
content-length: 615
last-modified: Fri, 14 Feb 2025 12:00:53 GMT
etag: "67af3075-267"
accept-ranges: bytes
```

###### Traffic server logs
```
[Dec  9 18:06:38.417] [ET_NET 15] WARNING: Bad payload length content_length=615 data_legnth=0 session_id=554
[Dec  9 18:06:38.417] [ET_NET 15] ERROR: HTTP/2 stream error code=0x01 client_ip=192.168.1.150 session_id=554 stream_id=3 recv data bad payload length
```

floom4:
HEAD method with "Content-Length" header was returning 502 code because header differs from payload length as there is no payload in this case.

See https://datatracker.ietf.org/doc/html/rfc7540#section-8.1.2.6

> A response that is defined to have no payload, as described in [RFC7230], Section 3.3.2, can have a non-zero content-length header field, even though no content is included in DATA frames.

Fixes #12747 

The change fixes the behavior by having `payload_length_is_valid()` return true if `data_length` is 0 meaning that the response has no payload, allowing for non-zero 'Content-Length' in this case as specified in the rfc.

### DIFF
--- a/include/proxy/http2/Http2Stream.h
+++ b/include/proxy/http2/Http2Stream.h
@@ -390,11 +390,21 @@ inline bool
 Http2Stream::payload_length_is_valid() const
 {
   uint32_t content_length = _receive_header.get_content_length();
-  if (content_length != 0 && content_length != data_length) {
+  uint64_t mask           = (MIME_PRESENCE_IF_UNMODIFIED_SINCE | MIME_PRESENCE_IF_MODIFIED_SINCE | MIME_PRESENCE_IF_RANGE |
+                   MIME_PRESENCE_IF_MATCH | MIME_PRESENCE_IF_NONE_MATCH);
+
+  // Skip Content-Length check on [RFC 7230] 3.3.2 conditions
+  bool is_payload_precluded =
+    this->is_outbound_connection() && (_send_header.method_get_wksidx() == HTTP_WKSIDX_HEAD ||
+                                       (_send_header.method_get_wksidx() == HTTP_WKSIDX_GET && _send_header.presence(mask) &&
+                                        _receive_header.status_get() == HTTPStatus::NOT_MODIFIED));
+
+  if (content_length != 0 && !is_payload_precluded && content_length != data_length) {
     Warning("Bad payload length content_length=%d data_legnth=%d session_id=%" PRId64, content_length,
             static_cast<int>(data_length), _proxy_ssn->connection_id());
+    return false;
   }
-  return content_length == 0 || content_length == data_length;
+  return true;
 }
 
 inline bool


maskit:
Good catch. However, as the RFC says, "A response that is defined to have no payload" is the condition. It looks like your change allows arbitrarily response that have no payload.

floom4:
My understanding of the RFC is that the condition defining a valid response without payload is the stream ending at header or continuation frame.
Considering that data_length is only incremented inside *rcv_data_frame*, I assumed that having data_length == 0 means that there was no data frame.

However, if you prefer we could skip this check in *rcv_header_frame* and *rcv_continuation_frame* if and only if they are not trailing headers (since trailing headers implies that data frames where received)

Here is an alternative condition that could be used:
```C++
if (stream->receive_end_stream && stream->trailing_header_is_possible() && !stream->payload_length_is_valid()) {
```
In both:
- rcv_header_frame https://github.com/apache/trafficserver/blob/0e4287020548e7316156a4debfed49c0fedc7312/src/proxy/http2/Http2ConnectionState.cc#L505
- rcv_continuation_frame https://github.com/apache/trafficserver/blob/0e4287020548e7316156a4debfed49c0fedc7312/src/proxy/http2/Http2ConnectionState.cc#L1117

maskit:
I don't think what frame carries END_STREAM flag matters. A DATA frame that carries just END_STREAM flag (no payload) is valid.

If a request is GET and the response has `content-length: 123`, `data_length` must be 123, right? Because a response for a GET request is not defined to have no payload. A response for a HEAD request is.

floom4:
You're right I misunderstood the RFC, I went back to it and in my understanding, the only cases where a non-zero 'Content-Lenght' header without payload is allowed are:
- HEAD response
- 304 response to conditional GET

From [Section 3.3.2 of [RFC7230]](https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2)
```
A server MAY send a Content-Length header field in a response to a
   HEAD request (Section 4.3.2 of [RFC7231]); ...

A server MAY send a Content-Length header field in a 304 (Not
   Modified) response to a conditional GET request (Section 4.1 of
   [RFC7232]); ...
```
It indeed means that, while other types of request and response may have an empty payload (or data_length == 0), they must not have a `Content-Length` header with a value different from 0.

Here is a proposal to check that the current frame is part of an HEAD response that we could use in rcv_headers_frame and rcv_continuation_frame:
```
stream->is_outbound_connection() && stream->get_sm()->t_state.hdr_info.server_request.method_get_wksidx() == HTTP_WKSIDX_HEAD
```
My understanding is that *is_outbound_connection* represents communication with origin which in case of receiving frame means response from origin
I assumed that *server_request* field would be set since the previous condition ensure that we are at the response step and consequently, the request was created and sent.

I think the same fields can be used to know if the frame is part of a response to a conditional GET request, however to find the http status code seems trickier since the response headers are not parsed yet as we just received the frame and are still validating it.

Maybe we could only fix HEAD request for the moment, does the above conditional statement work for you ?

maskit:
> My understanding is that is_outbound_connection represents communication with origin which in case of receiving frame means response from origin
I assumed that server_request field would be set since the previous condition ensure that we are at the response step and consequently, the request was created and sent.

Correct.

> I think the same fields can be used to know if the frame is part of a response to a conditional GET request, however to find the http status code seems trickier since the response headers are not parsed yet as we just received the frame and are still validating it.

`stream->decode_header_blocks` is called before `!stream->payload_length_is_valid()` in `rcv_headers_frame` and rcv_continuation_frame. The status code should be available. But if I'm missing something and it's tricky, I'm fine with fixing just the HEAD request case.


The proposed condition looks fine. Since those are accessible from `stream`, I think you can use the condition in `payload_length_is_valid`. And then it should work for rcv_data_frame as well.

include/proxy/http2/Http2Stream.h
```diff
@@ -390,11 +391,22 @@ inline bool
 Http2Stream::payload_length_is_valid() const
 {
   uint32_t content_length = _receive_header.get_content_length();
-  if (content_length != 0 && content_length != data_length) {
+  HTTPHdr *server_request = &this->get_sm()->t_state.hdr_info.server_request;
```
maskit: What was the reason you didn't use `_send_header` ?

floom4: Good catch, it seems that I overlooked this attribute when going through the code, this is definitely better.
I updated the patch.

cmcfarlen:

PR #12749, opened 2025-12-10
HRW: Produce errors when bad mods are used

zwoop:
This produces errors in the log output, but they are not considered load / reload failures. I.e. ATS will still startup as before. We should fix this in ATS v11.

Before this, configs like this would silently succeed, with the bad modifier ignored:

```
cond %{CLIENT-HEADER:X-Foo} ="foo" [BAD]
```

### DIFF
--- a/plugins/header_rewrite/parser.h
+++ b/plugins/header_rewrite/parser.h
@@ -195,11 +195,21 @@ public:
     return _val;
   }
 
+  // Check if the modifier exists and consume it from the list.
   bool
-  mod_exist(const std::string &m) const
+  consume_mod(const std::string &m)
   {
-    return std::find(_mods.begin(), _mods.end(), m) != _mods.end();
+    auto it = std::find(_mods.begin(), _mods.end(), m);
+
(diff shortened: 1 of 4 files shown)


zwoop:
[approve ci]

plugins/header_rewrite/parser.h
```diff
@@ -195,12 +195,22 @@ class Parser
...
   {
-    return std::find(_mods.begin(), _mods.end(), m) != _mods.end();
+    auto it = std::find(_mods.begin(), _mods.end(), m);
+
+    if (it != _mods.end()) {
+      _mods.erase(it);
+      return true;
+    }
+    return false;
   }
```
bneradt, line 209: Maybe rename this since `mod_exists` sounds like a read-only function. Maybe `consume_mod`?

zwoop: I've renamed this function as suggested.

bneradt (approved):
lgtm

PR #12750, opened 2025-12-10
ASAN: Fix leak with the Stripe raw-dirs allocation

zwoop:
Since this can be allocated in two ways, we have to also keep track how it was allocated for when the destructor is called.


zwoop:
@cmcfarlen This may be a candidate for ATS 10.1.

bneradt:
Thanks for fixing.

cmcfarlen:

PR #12755, opened 2025-12-11
ssl_multicert.config -> ssl_multicert.yaml

bneradt:
Replace the ssl_multicert.config format with YAML format, following the pattern established by sni.yaml. The new ssl_multicert.yaml uses a top-level 'ssl_multicert' key containing a sequence of certificate entries.

This also supports config conversion via:

```
traffic_ctl config ssl_multicert <old_config> <new_config>
```

---

# Community Decisions

- [x] Do we want these conversion scripts to be python or c++ files? This implements the conversion via `traffic_ctl config convert` per @cmcfarlen 's suggestion.
- [ ] Do we want to support both the old and new formats at the same time? At 10.x, we transitioned to only supporting records.yaml, so we have that as maybe something of a precedent. This PR supports both the old and new formats, but we have to decide on our policy for the old formats for 11.

At least for ssl_multicert, either of these options are possible from an implementation standpoint.

@cmcfarlen had an interesting suggestion to bake the config conversion logic into traffic_ctl. Maybe have something like:

```
traffic_ctrl config convert ssl_multicert|plugins|... <old_path> <new_path>
```


src/config/ssl_multicert.cc
```diff
@@ -0,0 +1,512 @@
...
+    write_int_field(KEY_SSL_TICKET_ENABLED, entry.ssl_ticket_enabled);
+    write_int_field(KEY_SSL_TICKET_NUMBER, entry.ssl_ticket_number);
+  }
+
+  return out.str();
+}
+
+std::string
+SSLMultiCertMarshaller::to_json(SSLMultiCertConfig const &config)
+{
```
brbzull0, line 360: Why not just use the yamlcpp to emit json? Although there is not a direct way, you can [emulate](https://stackoverflow.com/a/43903922) it:

```
YAML:Emitter Emitter;
emitter << YAML:: DoubleQuoted << YAML::Flow << /* rest of code */;
```

traffic_ctl is using [this](https://github.com/apache/trafficserver/blob/0d52043048bf31fee0dbd36958de12d9792caa77/include/shared/rpc/yaml_codecs.h#L201-L217).

There are some limitation though.

src/traffic_ctl/traffic_ctl.cc
```diff
@@ -137,6 +139,20 @@ main([[maybe_unused]] int argc, const char **argv)
 
   config_command.add_command("registry", "Show configuration file registry", Command_Execute)
     .add_example_usage("traffic_ctl config registry");
+
+  // ssl-multicert subcommand
+  auto &ssl_multicert_command =
+    config_command.add_command("ssl-multicert", "Manage ssl_multicert configuration").require_commands();
+  ssl_multicert_command.add_command("show", "Show the ssl_multicert configuration in JSON format", Command_Execute)
```
brbzull0: By default is JSON??  does it make sense to drop YAML too? should be straight forward as the YAML node is what we read from the file.?

if so, I'd suggest  using [mutex groups](https://docs.trafficserver.apache.org/en/latest/developer-guide/internal-libraries/ArgParser.en.html#mutually-exclusive-groups) for the options.

src/config/ssl_multicert.cc
bneradt: Good suggestion. I'm used to manually rolling this for efficiency reasons - that's what we do in traffic_dump for instance. But that happens on a per transaction basis and can be pretty "hot" code. But this shouldn't be.

I'll look into the emitter option. If it simplifies the code and that should be worth it here.

src/traffic_ctl/traffic_ctl.cc
bneradt: Sounds good.

bneradt: Oh, and in retrospect, let's make the default YAML. That's a good point - a typical user will prefer yaml over json. But we'll keep the --json for tools which may prefer that.

src/config/ssl_multicert.cc
```diff
@@ -0,0 +1,457 @@
...
+
+/// Check if a string ends with a suffix.
+bool
+ends_with(std::string const &str, std::string_view suffix)
+{
+  if (suffix.size() > str.size()) {
+    return false;
+  }
+  return str.compare(str.size() - suffix.size(), suffix.size(), suffix) == 0;
+}
```
cmcfarlen: Perhaps these are AI generated, but we have existing utility for these helpers. If we want to keep these, maybe move them to a `ts::config` namespace so other config parsers don't also have to generate them.  Probably best to just use swoc:TextView

src/iocore/net/YamlSSLMultiCertConfig.cc
```diff
@@ -0,0 +1,137 @@
...
+constexpr char YAML_SSL_CERT_NAME[]      = "ssl_cert_name";
+constexpr char YAML_DEST_IP[]            = "dest_ip";
+constexpr char YAML_SSL_KEY_NAME[]       = "ssl_key_name";
+constexpr char YAML_SSL_CA_NAME[]        = "ssl_ca_name";
+constexpr char YAML_SSL_OCSP_NAME[]      = "ssl_ocsp_name";
+constexpr char YAML_SSL_KEY_DIALOG[]     = "ssl_key_dialog";
+constexpr char YAML_DEST_FQDN[]          = "dest_fqdn";
+constexpr char YAML_SSL_TICKET_ENABLED[] = "ssl_ticket_enabled";
+constexpr char YAML_SSL_TICKET_NUMBER[]  = "ssl_ticket_number";
+constexpr char YAML_ACTION[]             = "action";
```
cmcfarlen: These are duplicated in src/config/ssl_multicert.cc.  A lot of this parser code below is also duplicated there. This code should just use the parser code in the config library so it doesn't have to change in multiple places later.

cmcfarlen (changes requested):
I really like the concept of having parsing code and conversions in one place.  It's really clean to do the conversion natively with traffic_ctl.  I do think the runtime code should also use this code to load config as well so there isn't code duplication.

src/config/ssl_multicert.cc
bneradt: Yeah, TextView is way better.

src/iocore/net/YamlSSLMultiCertConfig.cc
bneradt: Yes, good catch.

cmcfarlen (approved):
Looks good! Thanks for working on this!

PR #12760, opened 2025-12-15
Fix xdebug transform calling WRITE_READY when no data consumed

bryancall:
The xdebug probe-full-json transform was calling `TSContCall(WRITE_READY)` even when no data was available to consume (`towrite == 0`). This caused unnecessary callbacks to the transform, and under certain conditions could lead to a tight loop that would starve other transactions on the same ET_NET thread.

## The Bug

When using `X-Debug: probe-full-json=nobody` (or other probe-full-json variants), if the origin server was slow to send body data, the transform would:
1. Be called with `WRITE_READY` event
2. See no data available (`towrite == 0`)
3. Still call `TSVIOReenable` and `TSContCall(WRITE_READY)` back to upstream
4. Get called again immediately (step 1)

This creates a tight loop that can cause:
- High CPU usage on the affected ET_NET thread
- Starvation of other transactions on the same thread
- Requests appearing to "hang"

## The Fix

Add a check to only call `TSVIOReenable` and `TSContCall` when `towrite > 0`, meaning we actually consumed data. If no data was available, we simply return and wait for the upstream to call us again when more data arrives.

## Testing

Added a new autest (`x_probe_full_json_slow_origin`) that uses a slow origin server with chunked encoding to verify the transform handles delayed body data correctly.

### DIFF
--- a/plugins/xdebug/xdebug_transforms.cc
+++ b/plugins/xdebug/xdebug_transforms.cc
@@ -297,9 +297,9 @@ body_transform(TSCont contp, TSEvent event, void * /* edata ATS_UNUSED */)
 
     if (TSVIONTodoGet(src_vio) > 0) {
       if (towrite > 0) {
-        // Only reenable and notify upstream if we consumed data.
-        // If no data was available, just return and wait for the
-        // upstream to call us again when more data arrives.
+        // Only reenable when we consumed data. If we reenable and call
+        // WRITE_READY when towrite is 0 (no data available yet), we create
+        // a tight loop that starves other transactions and causes high CPU.
         TSVIOReenable(data->output_vio);
         TSContCall(TSVIOContGet(src_vio), TS_EVENT_VCONN_WRITE_READY, src_vio);
       }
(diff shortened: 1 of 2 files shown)


jasmine-nahrain (dismissed):

plugins/xdebug/xdebug_transforms.cc
```diff
@@ -296,8 +296,13 @@ body_transform(TSCont contp, TSEvent event, void * /* edata ATS_UNUSED */)
     }
 
     if (TSVIONTodoGet(src_vio) > 0) {
-      TSVIOReenable(data->output_vio);
-      TSContCall(TSVIOContGet(src_vio), TS_EVENT_VCONN_WRITE_READY, src_vio);
+      if (towrite > 0) {
+        // Only reenable and notify upstream if we consumed data.
+        // If no data was available, just return and wait for the
+        // upstream to call us again when more data arrives.
```
bneradt: Can we replace this "what" comment with a "why" comment? This comment says what the code is doing (not calling reenable unless we have consumed data), but that is rather obvious. Can we reword this to explain that the reason we do this is to address the transform loop described in the PR comment?

jasmine-nahrain (approved):

bneradt (approved):

PR #12761, opened 2025-12-16
Use ProxyProtocol SRC IP as Client IP

cmcfarlen:
~~Restore old feature to copy PP src IP to remote ip of NetVConn, but this time the copy is configurable.  Expand autest to cover new setup.~~

Edit: This doesn't restore the overwriting the remote_addr aspect of the old behavior, but does effectively replace the client IP with the proxy protocol address. /endedit

When the proxy protocol feature was added in #3958, the behavior was that if the `server_port` was configured for proxy protocol, and there was valid information available, the PP SRC IP address was alway copied to the NetVConnection's remote IP address.  This meant that for proxy protocol ports, the client's IP (as seen by the API and logs) was not the direct peer host, but the peer of the prior HOPs connection (presumably the original requesting client IP).  This might be desirable if ATS was used with some load balancer ingress to provide the PP and the ATS config was written without that hop in mind.  As far as I can tell, this feature was introduced for ATS 9 and back ported to later versions of 8.

Later, #8544 was filed to request that the prior hop's IP address be made available to the log and an `rchi` be added to that end. PR #8893 was created and that changed the behavior of the proxy protocol feature to no longer copy the PP SRC IP to the remote ip address.  This fixed the chi log to always report next hop, but removed the utility of having the "true" client IP for use in plugins, acls and logging.

This PR intends to revert to the mean and add the ability to restore the PP IP usage, but put that behavior behind a new server_port config.  This new flag should allow the change without breaking compatability with 10, but also allow restoring the behavior that was in ATS 9.

Draft for now to finish docs and discussion.

### DIFF
--- a/include/iocore/net/NetVConnection.h
+++ b/include/iocore/net/NetVConnection.h
@@ -524,8 +524,7 @@ protected:
     N_SERVICES,
   };
 
-  IpEndpoint local_addr;
-  // This can be the remote peer address or the proxy protocol SRC IP address
+  IpEndpoint    local_addr;
   IpEndpoint    remote_addr;
   ProxyProtocol pp_info;
 
--- a/tests/gold_tests/autest-site/trafficserver.test.ext
+++ b/tests/gold_tests/autest-site/trafficserver.test.ext
@@ -437,7 +437,7 @@ def MakeATSProcess(
             ppcfg = "pp"
             if enable_proxy_protocol_cp_src:
                 ppcfg += ":pp-clnt"
-            port_str += f" {p.Variables.proxy_protocol_port}:{ppcfg} {p.Variables.proxy_protocol_portv6}:pp:ipv6"
+            port_str += f" {p.Variables.proxy_protocol_port}:{ppcfg} {p.Variables.proxy_protocol_portv6}:{ppcfg}:ipv6"
             if enable_tls:
                 port_str += f" {p.Variables.proxy_protocol_ssl_port}:{ppcfg}:ssl {p.Variables.proxy_protocol_ssl_portv6}:{ppcfg}:ssl:ipv6"
             if enable_uds:


cmcfarlen:
After sleeping on this, I decided that I don't like that the meaning of remote address in NetVConnection is overloaded for proxy protocol.  So, I reverted that and instead introduced `get_client_addr` and friends to return either the proxy protocol IP or the remote IP based on the config.  This also provides the ability to implement the `rchi` log fields that was requested in #8544.  Now the TSHttpTxnClientAddrGet calls the new `get_client_addr` and the remote address can still be acquired by `TSNetVConnRemoteAddrGet` without ambiguity.

cmcfarlen:
autest failed: proxy_serve_stale_dns_fail

cmcfarlen:
[approve ci autest 1]

src/iocore/net/SNIActionPerformer.cc
```diff
@@ -417,7 +417,7 @@ SNI_IpAllow::SNIAction(SSL &ssl, ActionItem::Context const & /* ctx ATS_UNUSED *
   const sockaddr *client_ip = nullptr;
   for (int i = 0; i < IpAllow::Subject::MAX_SUBJECTS; ++i) {
     if (IpAllow::Subject::PEER == IpAllow::subjects[i]) {
-      client_ip = ssl_vc->get_remote_addr();
+      client_ip = ssl_vc->get_client_addr();
```
maskit: Hmm, I don't think we should do this. If the setting value were CLIENT then this would make sense, but "PEER" was picked as a term that means the direct peer to distinguish it from the ambiguous term "client".

Also, this is unrelated to the change, I found that this if-else lacks the support for PLUGIN (verified address). I'll add it.

src/iocore/net/SSLClientUtils.cc
```diff
@@ -81,7 +81,7 @@ verify_callback(int signature_ok, X509_STORE_CTX *ctx)
       Dbg(dbg_ctl_ssl_verify, "verification error:num=%d:%s:depth=%d", err, X509_verify_cert_error_string(err), depth);
       const char *sni_name;
       char        buff[INET6_ADDRSTRLEN];
-      ats_ip_ntop(netvc->get_remote_addr(), buff, INET6_ADDRSTRLEN);
+      ats_ip_ntop(netvc->get_client_addr(), buff, INET6_ADDRSTRLEN);
```
maskit: Looks like this function is for origin connections. If so, remote_addr means origin's address, and client_addr means ATS's address.

maskit: Actually, it may not. But it's confusing because the client is ATS on origin connections.

src/iocore/net/SSLDiags.cc
```diff
@@ -164,7 +164,7 @@ SSLDiagnostic(const SourceLocation &loc, bool debug, SSLNetVConnection *vc, cons
   ip_text_buffer ip_buf = {'\0'};
 
   if (vc) {
-    ats_ip_ntop(vc->get_remote_addr(), ip_buf, sizeof(ip_buf));
+    ats_ip_ntop(vc->get_client_addr(), ip_buf, sizeof(ip_buf));
```
maskit: Same as above. The name doesn't make sense on origin connections.

src/iocore/net/SNIActionPerformer.cc
cmcfarlen: The point of this PR and the new pp-clnt flag (if configured) is to use the actual/true client address (from proxy protocol) for PEER or Client instead of the literal peer host's address.  So, in all places where PEER is used, we need to check if the PP SRC IP should be used instead.  This is what `get_client_addr` does.  I'm open to better names for this function.

src/iocore/net/SSLClientUtils.cc
cmcfarlen: What is a better name? `effective_remote_addr`? I agree it's odd if the connection is not the incoming request one, but it looked like this function is generic.  The only important function call change in this last commit is for ACLs.  That is client side and does require `get_client_addr`.

tests/gold_tests/autest-site/trafficserver.test.ext
```diff
@@ -431,11 +432,14 @@ def MakeATSProcess(
             port_str += " {ssl_port}:quic {ssl_portv6}:quic:ipv6".format(
                 ssl_port=p.Variables.ssl_port, ssl_portv6=p.Variables.ssl_portv6)
         if enable_proxy_protocol:
-            port_str += f" {p.Variables.proxy_protocol_port}:pp {p.Variables.proxy_protocol_portv6}:pp:ipv6"
+            ppcfg = "pp"
+            if enable_proxy_protocol_cp_src:
+                ppcfg += ":pp-clnt"
+            port_str += f" {p.Variables.proxy_protocol_port}:{ppcfg} {p.Variables.proxy_protocol_portv6}:pp:ipv6"
```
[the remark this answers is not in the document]
cmcfarlen: nice one copilot!

include/iocore/net/NetVConnection.h
```diff
@@ -510,7 +524,8 @@ class NetVConnection : public VConnection, public PluginUserArgs<TS_USER_ARGS_VC
     N_SERVICES,
   };
 
-  IpEndpoint    local_addr;
+  IpEndpoint local_addr;
+  // This can be the remote peer address or the proxy protocol SRC IP address
```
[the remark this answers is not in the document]
cmcfarlen: copilot is correct, but I'm just going to remove this comment.

cmcfarlen:
[approve ci freebsd]

PR #12763, opened 2025-12-16
Adds metrics and log fields for tracking TLS handshake bytes

zwoop:


include/iocore/net/NetVConnection.h
```diff
@@ -322,6 +322,24 @@ class NetVConnection : public VConnection, public PluginUserArgs<TS_USER_ARGS_VC
...
+   *
+   * For TLS connections, captures the bytes read/written during the TLS handshake.
+   * If already captured, returns the stored values.
+   *
+   * @param[out] bytes_in  Bytes received during handshake
+   * @param[out] bytes_out Bytes transmitted during handshake
+   * @return true if values were captured on this call, false if already captured or not applicable
+   */
+  virtual bool
+  capture_tls_handshake_bytes(uint64_t &bytes_in, uint64_t &bytes_out)
```
maskit: I don't like to see TLS stuff in NetVConnection. `TLSBasicSupport` is probably a better place.

src/proxy/logging/LogAccess.cc
```diff
@@ -2096,6 +2096,27 @@ LogAccess::marshal_client_req_squid_len(char *buf)
...
+  Client request squid length plus TLS handshake bytes received for TLS connections.
+  -------------------------------------------------------------------------*/
+int
+LogAccess::marshal_client_req_squid_len_tls(char *buf)
+{
+  if (buf) {
+    int64_t val = 0;
+
+    if (m_client_request) {
+      val = m_client_request->length_get() + m_http_sm->client_request_body_bytes;
```
maskit, line 2222: What if the handshake has early data? I guess the size of the request gets double counted.

include/iocore/net/NetVConnection.h
zwoop: Hmmm, I changed the name of this already, it should have been capture_tls_handshake_bytes()

src/proxy/logging/LogAccess.cc
zwoop: Yeah, most likely, at least on the "rx" side.

src/iocore/net/TLSBasicSupport.cc
```diff
@@ -72,6 +73,45 @@ TLSBasicSupport::clear()
...
+
+  uint64_t bio_in  = rbio ? BIO_number_read(rbio) : 0;
+  uint64_t bio_out = wbio ? BIO_number_written(wbio) : 0;
+
+  // Subtract early data (TLS 1.3 0-RTT) from incoming bytes
+  if (auto *early_data = TLSEarlyDataSupport::getInstance(ssl); early_data) {
+    size_t early_data_len = early_data->get_early_data_len();
+
+    if (early_data_len > 0 && bio_in > early_data_len) {
+      bio_in -= early_data_len;
```
maskit: I don't think we should subtract the length of early data here. It's inaccurate as the size of TLS handshake bytes, and we can do the math on the caller side. The subtraction only makes sense where you want to avoid double counting the size of HTTP requests.

Also I don't want the dependency for TLSEarlyDataSuport here.

zwoop:
I think I've addressed @maskit's concerns in the latest update.

doc/admin-guide/logging/formatting.en.rst
```diff
@@ -451,6 +453,12 @@ cqcl  Client Request         Client request content length, in bytes.
 cqhl  Client Request         Client request header length, in bytes.
 cqql  Client Request         Client request header and content length combined,
                              in bytes.
+cqqtl Client Request         Same as cqql_, but for the first transaction on a
```
maskit, line 456: We may want a note (maybe as a limitation) about the accuracy. In some cases, the logged value wouldn't really reflect the amount of data transferred on wire to make the first transaction.

doc/admin-guide/logging/formatting.en.rst
```diff
@@ -451,6 +453,10 @@ cqcl  Client Request         Client request content length, in bytes.
...
+                             TLS connection, also includes TLS handshake bytes
+                             received from the client. Note that this metrics
```
bneradt: `this metrics` -> `this metric`

src/iocore/net/TLSBasicSupport.cc
```diff
@@ -72,6 +72,36 @@ TLSBasicSupport::clear()
...
+  BIO *rbio = SSL_get_rbio(ssl);
+  BIO *wbio = SSL_get_wbio(ssl);
+
+  uint64_t bio_in  = rbio ? BIO_number_read(rbio) : 0;
+  uint64_t bio_out = wbio ? BIO_number_written(wbio) : 0;
+
+  bytes_in = _tls_handshake_bytes_in = bio_in;
+  bytes_out = _tls_handshake_bytes_out = bio_out;
+
+  return true;
```
bneradt: Shouldn't that first return return true? It set values.

Maybe this is intentional? We only return true the first time the BIO is read? Regardless, let's add doxygen comments explaining the semantics of the returned boolean.

src/iocore/net/SSLStats.cc
```diff
@@ -280,6 +280,8 @@ SSLInitializeStatistics()
   ssl_rsb.user_agent_version_too_high       = Metrics::Counter::createPtr("proxy.process.ssl.user_agent_version_too_high");
   ssl_rsb.user_agent_version_too_low        = Metrics::Counter::createPtr("proxy.process.ssl.user_agent_version_too_low");
   ssl_rsb.user_agent_wrong_version          = Metrics::Counter::createPtr("proxy.process.ssl.user_agent_wrong_version");
+  ssl_rsb.tls_handshake_bytes_in_total      = Metrics::Counter::createPtr("proxy.process.ssl.total_handshake_bytes_read_in");
+  ssl_rsb.tls_handshake_bytes_out_total     = Metrics::Counter::createPtr("proxy.process.ssl.total_handshake_bytes_write_in");
```
bneradt: Seems to be a mismatch in variable name and the metric used. Should the variable be `in_total` or the metric pulled from `write_out`?

bneradt (changes requested):

include/iocore/net/TLSBasicSupport.h
```diff
@@ -103,4 +105,6 @@ class TLSBasicSupport
 
   ink_hrtime _tls_handshake_begin_time = 0;
   ink_hrtime _tls_handshake_end_time   = 0;
+  uint64_t   _tls_handshake_bytes_in   = 0;
+  uint64_t   _tls_handshake_bytes_out  = 0;
```
bneradt: These are cache variables that are lazy initialized, right? If that's right, let's make them mutable so that get_tls_handshake_bytes can be const. As a getter, that would make more sense to the caller.

bneradt (dismissed):

src/iocore/net/SSLStats.cc
zwoop: Fixed

include/iocore/net/TLSBasicSupport.h
zwoop: Fixed

src/iocore/net/TLSBasicSupport.cc
zwoop: I made a comment, but not doxygen. We are split-brain on this, and I don't think we use Doxygen in here normally.

zwoop:
[approve ci autest 0]

zwoop:
[approve ci centos]

bneradt (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #12764, opened 2025-12-16
Fix HTTP caching compliance by adding Vary headers for compressible c…

cmcfarlen:
…ontent (#12741)

The compress plugin was only adding Vary: Accept-Encoding headers when content was actually going to be compressed, not when content could be compressed. This can cause downstream caches to never get the compressed version in cache.

Now adds Vary: Accept-Encoding headers for all compressible content regardless of whether compression is applied, ensuring proper HTTP cache behavior.

(cherry picked from commit 0d52043048bf31fee0dbd36958de12d9792caa77)


cmcfarlen:
[approve ci autest 0]

PR #12766, opened 2025-12-19
Cleanup connect result handling

yknoya:
# Background
This PR is derived from https://github.com/apache/trafficserver/pull/12729.  
Since the changes in that PR mixed a bug fix and cleanup work, we decided to split them into two separate PRs.  
This PR extracts and focuses on the cleanup portion.

The motivation for this cleanup is that pre-setting `EIO` in `connect_result` before a connection attempt is error-prone.  
Instead, we want to set an error in `connect_result` only when a connection error actually occurs.  
This work was triggered by the following review comment:  
https://github.com/apache/trafficserver/pull/12729#discussion_r2601216919

> [!NOTE]
> For the background on why `EIO` started being pre-set in `connect_result` before connection attempts, see https://github.com/apache/trafficserver/pull/7580

# Cleanup details
See the following comment for details:  
https://github.com/apache/trafficserver/pull/12729#discussion_r2618628728

### DIFF
--- a/src/proxy/http/HttpTransact.cc
+++ b/src/proxy/http/HttpTransact.cc
@@ -3559,7 +3559,7 @@ HttpTransact::handle_response_from_parent(State *s)
   switch (s->current.state) {
   case CONNECTION_ALIVE:
     TxnDbg(dbg_ctl_http_trans, "[hrfp] connection alive");
-    s->current.server->connect_result = 0;
+    s->set_success();
     SET_VIA_STRING(VIA_DETAIL_PP_CONNECT, VIA_DETAIL_PP_SUCCESS);
     if (s->parent_result.retry) {
       markParentUp(s);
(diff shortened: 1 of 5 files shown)


src/proxy/http/HttpTransact.cc
bneradt, line 3598: Should this use set_fail?

bneradt (changes requested):
I think this looks fine as far as I can tell. I'm a bit nervous, though, since modifications in this logic tend to result in subtle bugs. I'll build a dev rpm and try this in a production box and report back.

bneradt (changes requested):
When I cherry-picked this locally and put this in production, it was unstable. I got the following crash:

```
traffic_server: received signal 11 (Segmentation fault)
traffic_server - STACK TRACE:
/lib64/libpthread.so.0(+0x12990)[0x7f575fd65990]
/opt/edge/2.0/trafficserver/10.0/bin/traffic_server(HdrHeap::unmarshal(int, int, HdrHeapObjImpl**, RefCountObj*)+0x7)[0x6b3057]
/opt/edge/2.0/trafficserver/10.0/bin/traffic_server(HTTPInfo::unmarshal(char*, int, RefCountObj*)+0xeb)[0x6b0d8b]
/opt/edge/2.0/trafficserver/10.0/bin/traffic_server[0x7200c9]
/opt/edge/2.0/trafficserver/10.0/bin/traffic_server(CacheVC::handleReadDone(int, Event*)+0xe58)[0x7248f8]
/opt/edge/2.0/trafficserver/10.0/bin/traffic_server(AIOCallback::io_complete(int, void*)+0x251)[0x745861]
/opt/edge/2.0/trafficserver/10.0/bin/traffic_server(EThread::process_event(Event*, int)+0x3dc)[0x87b35c]
/opt/edge/2.0/trafficserver/10.0/bin/traffic_server(EThread::process_queue(Queue<Event, Event::Link_link>*, int*, int*)+0x276)[0x87bb56]
/opt/edge/2.0/trafficserver/10.0/bin/traffic_server(EThread::execute_regular()+0x147)[0x87bfd7]
/opt/edge/2.0/trafficserver/10.0/bin/traffic_server(EThread::execute()+0x1c2)[0x87cc12]
/opt/edge/2.0/trafficserver/10.0/bin/traffic_server[0x879702]
/lib64/libpthread.so.0(+0x81ca)[0x7f575fd5b1ca]
/lib64/libc.so.6(clone+0x43)[0x7f575f9b68d3]
```

Sadly, it's not obvious how that relates to the patch. But that's the problem with this code. We do use `global` origin side connection share pooling, so we make use of the ConnectingEntry logic.

yknoya:
@bneradt 
Thanks for the review and for testing this in production. I'll investigate the cause of the crash, so please give me some time.

PR #12767, opened 2025-12-19
Use malloc allocated buffers for h2 headers if enabled

cmcfarlen:
Internal testing shows this change increased performance on high RPS setups enough to warrant a code change.  I've conditioned this change on the existing `ENABLE_MALLOC_ALLOCATOR` build flag so that it's opt-in.

### DIFF
--- a/include/proxy/http2/Http2Stream.h
+++ b/include/proxy/http2/Http2Stream.h
@@ -209,10 +209,14 @@ private:
   Http2StreamState _state      = Http2StreamState::HTTP2_STREAM_STATE_IDLE;
   int64_t          _http_sm_id = -1;
 
-  HTTPHdr   _receive_header;
+  HTTPHdr _receive_header;
+#if TS_USE_MALLOC_ALLOCATOR
+  MIOBuffer _receive_buffer{BUFFER_SIZE_INDEX_FOR_XMALLOC_SIZE(4096)};
+#else
   MIOBuffer _receive_buffer{BUFFER_SIZE_INDEX_4K};
-  VIO       read_vio;
-  VIO       write_vio;
+#endif
+  VIO read_vio;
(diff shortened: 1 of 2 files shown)


cmcfarlen:
[approve ci autest 3]

include/proxy/http2/Http2Stream.h
```diff
@@ -209,10 +209,14 @@ class Http2Stream : public ProxyTransaction
...
 
-  HTTPHdr   _receive_header;
+  HTTPHdr _receive_header;
+#if TS_USE_MALLOC_ALLOCATOR
+  MIOBuffer _receive_buffer{BUFFER_SIZE_INDEX_FOR_XMALLOC_SIZE(4096)};
+#else
   MIOBuffer _receive_buffer{BUFFER_SIZE_INDEX_4K};
-  VIO       read_vio;
-  VIO       write_vio;
+#endif
```
bneradt, line 217: Can `HTTP2_HEADER_BUFFER_SIZE_INDEX` be constexpr so that `_receive_buffer` is allocated with that value? Or do compilers not like that?

bneradt (changes requested):

include/proxy/http2/Http2Stream.h
cmcfarlen: The issue here is the visibility of `HTTP2_HEADER_BUFFER_SIZE_INDEX`.  It's not included from this file.  I didn't want to change the dependencies of the files.  I presume that was why it didn't previously use the buffer size index.

I couldn't find any good common header to move the definition to either.

bneradt: Sounds good. Thanks for interacting with the idea.

cmcfarlen:

PR #12771, opened 2025-12-22
Decouple recursive var in CacheVC

serrislew:
https://github.com/apache/trafficserver/commit/06a99bb1e6140bf4c2aae091102b0406529144a4 causes a leak of CacheVConnection objects.

`CacheVC::recursive` already used for user callback do_io_read/do_io_write/do_io_close/calluser/callcont and die() is only called if value is 0

Seems like using same value for `CacheVC::openReadStartEarliest` recursive checks can increase above value and skips the die() call. Decoupling for different use cases


bneradt (approved):
Looks good.

Just to check: I don't see `read_recursive` being initialized. I assume `read_recursive` is initialized by some memset(0) call somewhere along with the other members?

serrislew:
memset is called CacheVC() initialization in [CacheVC.cc](https://github.com/apache/trafficserver/blob/master/src/iocore/cache/CacheVC.cc#L161-L166)

cmcfarlen:

PR #12772, opened 2025-12-23
Do not merge: Fix potential null pointer dereference in virtualhost remap logic

bneradt:
# Do Not Merge

I'm just creating this patch to test whether this fixes the centos CI failure on #12669.

---

Add null checks before calling m_remap->release() in do_remap_request().
This prevents a segfault when switching between virtualhost and global
remap tables if m_remap is unexpectedly null.


bneradt:
Still crashes with CentOS.

PR #12773, opened 2025-12-23
DO NOT MERGE: Add demangled backtrace to crash logs

bneradt:
# DO NOT MERGE

Adds bt to crashlog. Testing this with Serris's patch (#12669) since that currently crashes in the CentOS build.

---

Previously, backtraces only appeared in crash logs when libunwind
(TS_USE_REMOTE_UNWINDING) was available. Without it, the crash log
would say 'Unable to retrieve backtrace' while the backtrace was
only written to stderr (traffic.out).

This change adds an in-process backtrace mechanism that:
- Generates a demangled backtrace from the crashing thread
- Sends it over the existing pipe to traffic_crashlog
- Uses it as a fallback when ServerBacktrace() fails

Changes:
- Add ink_stack_trace_get() to generate demangled backtraces using
  abi::__cxa_demangle()
- Modify crash_logger_invoke() to capture and send the backtrace
- Update traffic_crashlog to read and include the in-process backtrace
- Use std::string for cleaner memory management

### DIFF
--- a/include/tscore/ink_stack_trace.h
+++ b/include/tscore/ink_stack_trace.h
@@ -23,6 +23,8 @@
 
 #pragma once
 
+#include <string>
+
 // The max number of levels in the stack trace
 #define INK_STACK_TRACE_MAX_LEVELS 100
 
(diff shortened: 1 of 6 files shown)


bneradt:
[approve ci centos]

bneradt:
[approve ci centos]

bneradt:
[approve ci centos]

bneradt:
It's hopeless. CI dockers images do not allow the permissions required for traffic_crashlog to get the backtrace information of the ATS process.

PR #12774, opened 2025-12-23
Add .git-blame-ignore-revs

cmcfarlen:
This file can contain commit hashes that should be ignored when doing git blame.  The main purpose of this is to ignore commits that only change formatting to better see actual changes.  I only added the one that was annoying me.

You might need to add some config to use this:

```
git config blame.ignoreRevsFile .git-blame-ignore-revs
```


cmcfarlen:
[approve ci]

PR #12775, opened 2025-12-24
Add backtrace to crash logs

bneradt:
Send crash information from the signal handler to traffic_crashlog and
record it in crash logs.

This also:

* Adds a 10-second timeout around backtrace collection to prevent the
  crashlog helper from hanging indefinitely if ptrace/waitpid blocks.
  This addresses a TODO comment that has been in the code for a while.
* Removes the "Memory Regions" section from the crash logs. These are
  very rarely useful and otherwise add significant noise to the crash
  log. A post mordem gdb section on a core should be able to get this
  information if it is necessary in a particular instance.


bneradt (commented):
@cmcfarlen suggests looking into just doing the bt of the crashing thread instead of all.

Issue #12776, opened 2025-12-31
CentOS 7 CI Crashes Due to DbgCtl Use-After-Free


bneradt:
## Problem                                                                                                                                                                           
                                                                                                                                                                                     
CentOS 7 ASF CI runs are failing with segmentation faults during regression test execution. The crashes occur frequently when running `traffic_server -K -R 3`.                      
                                                                                                                                                                                     
## Investigation                                                                                                                                                                     
                                                                                                                                                                                     
Using diagnostic logging, it was identified that the crash is caused by a use-after-free bug in the `DbgCtl` reference counting system. Despite careful design to handle static initialization/destruction order issues, the reference counting mechanism seems to not prevent a use after free of the std::map - at least in the older compiled version of ATS from CentOS 7.

**The rest of the below is from Cursor. I include it in case it is helpful.** I'm not entirely convinced its root cause concerning one thread deleting the tag registry impacting the other threads is correct or not. It is clear, however, that simply not freeing the std::map registry does make the test stable.

## Root Cause

### The Reference Counting Issue

The `DbgCtl` class uses a shared registry with reference counting:
- Each `DbgCtl` object increments `registry_reference_count` on construction.
- Each `DbgCtl` destructor calls `_rm_reference()`, which decrements the count.
- When `ref_count` reaches 0, the registry (including the `std::map` of all tags) is deleted.

**The fatal flaw:** The reference count tracks how many `DbgCtl` objects currently exist, but it cannot account for whether those objects will be accessed before they're destroyed.

### How the Crash Occurs

1. **Multiple Threads/Contexts**: The application has DbgCtl objects across multiple threads. Some are static/global, some are thread-local, and some may be in function scopes.

2. **Thread Exit During Execution**: When a thread exits mid-execution (not just at program exit), all its DbgCtl objects destruct:
   ```
   Thread A exits → its DbgCtl objects destruct → _rm_reference() called repeatedly
   ref_count: 17 → 16 → 15 → ... → 1 → 0
   ```
                                                                                                                                                                                     
3. **Registry Deleted Prematurely**: When the last DbgCtl from Thread A destructs, `ref_count` hits 0, and the registry (with its `std::map` containing all 229 tag entries) is deleted.

5. **Dangling Pointers in Other Threads**: Thread B is still running and has DbgCtl objects with `_ptr` pointing into the now-deleted registry:

```cpp
   bool DbgCtl::on() const {
     // ...
[...]

PR #12777, opened 2025-12-31
Fix DbgCtl use-after-free shutdown crash via leaky singleton

bneradt:
The DbgCtl registry was experiencing use-after-free crashes during program shutdown, particularly visible in CI regression tests. The root cause was the undefined destruction order of static objects in C++.

Problem:
- DbgCtl objects can be static or thread-local with lifetimes spanning program execution
- When one compilation unit's DbgCtl destructs and triggers registry cleanup, other compilation units may still have DbgCtl objects with pointers into that registry
- Thread exit order is also unpredictable, causing similar issues when thread-local DbgCtl objects destruct

This implements the "leaky singleton" pattern where the registry is:
- Created on first use
- Never destroyed (destructor is now = default)
- Memory (~20KB) reclaimed by OS at process exit

Fixes: #12776

### DIFF
--- a/ci/asan_leak_suppression/regression.txt
+++ b/ci/asan_leak_suppression/regression.txt
@@ -19,5 +19,8 @@ leak:RegressionTest_HostDBProcessor
 leak:RegressionTest_DNS
 leak:RegressionTest_UDPNet_echo
 leak:HttpConfig::startup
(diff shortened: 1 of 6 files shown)


bneradt:
[approve ci fedora]

serrislew (approved):

PR #12784, opened 2026-01-06
Fix convert2yaml.py to handle files without trailing newline

bryancall:
The script was using `value[:-1]` to strip the trailing newline, which incorrectly truncates the last character if the file doesn't end with a newline. Changed to `value.rstrip('\n')` which properly handles both cases.

Added test case for config files without trailing newlines.


cmcfarlen:

PR #12791, opened 2026-01-08
Add state transition diagram of StripeSM

masaori335:
No code change. Add a plantuml diagram as a comment to describe state transition of StripeSM. 

Screenshot 2026-01-08 at 11 01 37

### DIFF
--- a/src/iocore/cache/StripeSM.h
+++ b/src/iocore/cache/StripeSM.h
@@ -60,6 +60,56 @@ struct StripeInitInfo;
 class CacheEvacuateDocVC;
 class RamCache;
 
+/**
+  @class StripeSM
+
+  @startuml
+  hide empty description
+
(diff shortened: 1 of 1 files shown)


src/iocore/cache/StripeSM.h
```diff
@@ -60,6 +60,56 @@ struct StripeInitInfo;
...
+    2. Directly from CacheVC::handleWrite
+    3. Directly from CacheSync::mainEvent
+    4. Directly from aggWriteDone
+  end note
+
+  aggWrite --> aggWriteDone
+
+  note bottom of aggWriteDone
+    calls aggWrite() directly
+    without changing state
```
masaori335, line 83: This is a pitfall. Change the state back to `aggWrite` before calling it seems better to me.

bryancall (approved):

PR #12797, opened 2026-01-09
Restore DbgCtl::_rm_reference() for ABI compatibility

bneradt:
PR #12777 introduced a leaky singleton pattern for DbgCtl to fix use-after-free crashes during shutdown. However, removing the _rm_reference() method broke ABI compatibility with existing plugins that were compiled against the old header, where the destructor called this method. This commit restores _rm_reference() as a no-op stub, allowing old plugins to load successfully while maintaining the leaky singleton behavior.

### DIFF
--- a/include/tsutil/DbgCtl.h
+++ b/include/tsutil/DbgCtl.h
@@ -151,6 +151,11 @@ private:
 
   static const _TagData *_new_reference(char const *tag);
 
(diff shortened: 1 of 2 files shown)


bneradt:
> Mind putting the word `deprecated` in the comment? This might make it easier to find when we do cleanup later.

Good idea.

PR #12798, opened 2026-01-09
Fix uninitialized LogConfig member variable

bneradt:
Valgrind reported a conditional jump depending on an uninitialized value in Log::periodic_tasks(). The root cause was that roll_log_files_now was not initialized in LogConfig. While we're at it, I'm initializing m_partition_space_left in the member declaration as well.

Fixed by adding default member initializers in the class definition, consistent with the other members.

Valgrind report that reported this:

```
  Thread 30 [LOG_FLUSH]:
  Conditional jump or move depends on uninitialised value(s)
     at 0x68F310: Log::periodic_tasks(long)
     by 0x6901FF: Log::flush_thread_main(void*)
     by 0x695FEC: LoggingFlushContinuation::mainEvent(int, void*)
     by 0x883EA2: EThread::execute()
     by 0x87FD61: spawn_thread_internal(void*)
   Uninitialised value was created by a heap allocation
     at 0x4C2A593: operator new(unsigned long)
     by 0x695440: Log::init(int)
```

### DIFF
--- a/include/proxy/logging/LogConfig.h
+++ b/include/proxy/logging/LogConfig.h
@@ -163,8 +163,8 @@ public:
   bool    reconfiguration_needed  = false;
   bool    logging_space_exhausted = false;
   int64_t m_space_used            = 0;
-  int64_t m_partition_space_left;
-  bool    roll_log_files_now; // signal that files must be rolled
+  int64_t m_partition_space_left  = static_cast<int64_t>(UINT_MAX);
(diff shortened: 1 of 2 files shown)


bneradt:
Thanks for the review.

> I wonder if coverity found this?

Maybe coverity could have found it? I found it via the valgrind run I mentioned in the description while trying to track down a different issue.

PR #12799, opened 2026-01-12
fix(rpc): improve error reporting for socket path issues.

brbzull0:
Sometimes, when you use autest, you could end up with a large socket pathname.
You only discover this once you try to access the jsonrpc node, but even when
you get an error, it's hard to identify the root cause.

With this change, you'll now find a clearer error in the logs that points you
in the right direction.

Replace generic `ENAMETOOLONG` errno with an error that indicates the RPC
socket issue. Also improve startup error message clarity.

Note: _Using `fprintf` as it seems the log machinery may not be fully initialized at this point._

### DIFF
--- a/include/mgmt/rpc/server/CommBase.h
+++ b/include/mgmt/rpc/server/CommBase.h
@@ -36,7 +36,7 @@ struct BaseCommInterface {
   virtual std::string const &name() const                        = 0;
 };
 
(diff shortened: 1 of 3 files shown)


brbzull0:
[approve ci]

ezelkow1:
[approve ci]

PR #12802, opened 2026-01-12
Fix origin_server_auth URL encoding for mixed-encoding URLs

bryancall:
## Summary
Fixed the `isUriEncoded()` and `canonicalEncode()` functions in the origin_server_auth plugin to properly handle URLs with mixed encoding (some characters encoded, some not).

## Bug Description
When a URL has mixed encoding (e.g., `/app/(channel)/%5B%5Bparts%5D%5D/page.js` where parentheses are NOT encoded but brackets ARE encoded), the AWS v4 signature calculation was incorrect:

- `isUriEncoded()` found `%5B` and returned `true`, incorrectly assuming the entire string was fully encoded
- `canonicalEncode()` returned the string as-is
- Signature was calculated for the partially-encoded path
- S3/GCP expected signature for the fully-encoded canonical path
- Result: 403 SignatureDoesNotMatch

## AWS SigV4 Canonical URI Encoding

Per the [AWS SigV4 spec](https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html):

> "URI-encode every byte except the unreserved characters: 'A'-'Z', 'a'-'z', '0'-'9', '-', '.', '_', and '~'."

This means characters like `(`, `)`, `!`, `*`, `'` **must be percent-encoded** in the canonical URI for signature calculation, even though they are listed as "safe" for [S3 object key names](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html).

| Character | S3 "Safe" for key names | SigV4 Unreserved | Must encode for signature |
|-----------|--------------------------|------------------|---------------------------|
| `(` | ✅ Yes | ❌ No | ✅ Yes (`%28`) |
| `)` | ✅ Yes | ❌ No | ✅ Yes (`%29`) |
| `!` | ✅ Yes | ❌ No | ✅ Yes (`%21`) |
| `*` | ✅ Yes | ❌ No | ✅ Yes (`%2A`) |
| `'` | ✅ Yes | ❌ No | ✅ Yes (`%27`) |

## Fix
1. **`isUriEncoded()`**: Now checks the ENTIRE string and returns `false` if ANY character that should be encoded is found unencoded
2. **`canonicalEncode()`**: For partially-encoded strings, decodes first then re-encodes to ensure consistent canonical output
3. Added `uriDecode()` helper function

## Testing
- Added unit tests for mixed encoding scenarios
- Added tests for all S3 "safe" characters that need SigV4 encoding (`!`, `*`, `'`, `(`, `)`)
- Added autest for end-to-end verification
- All 60 test cases pass (184 assertions)

## References
- AWS SigV4 spec: https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html
- S3 object key naming: https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html

### DIFF
--- a/plugins/origin_server_auth/aws_auth_v4.cc
+++ b/plugins/origin_server_auth/aws_auth_v4.cc
@@ -137,26 +137,22 @@ uriEncode(const String &in, bool isObjectName)
 }
 
 /**
- * @brief Check if a string is FULLY URI-encoded per AWS SigV4 canonical encoding rules.
+ * @brief Check if a string is already in AWS SigV4 canonical form.
+ *
+ * A string is canonical if it either:
+ *   1. Contains only unreserved characters (A-Z, a-z, 0-9, '-', '.', '_', '~')
+ *      and optionally '/' for object names - no encoding needed
+ *   2. Is properly percent-encoded with UPPERCASE hex digits
  *
  * @see AWS spec: http://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html
  * @see https://docs.aws.amazon.com/IAM/latest/UserGuide/create-signed-request.html
  *
- * @note According to RFC 3986, if a string contains '%' it must be followed by 2 hexadecimal
- *       symbols, otherwise '%' should be encoded as %25:
- *          https://tools.ietf.org/html/rfc3986#section-2.1
- *
- * @note This function checks if ALL characters that need encoding ARE encoded, AND that
- *       all percent-encoded sequences use UPPERCASE hex digits (per AWS SigV4 requirement).
- *       A string with mixed encoding (some chars encoded, some not) or lowercase hex digits
- *       returns false, which triggers canonicalEncode() to normalize via decode/re-encode.
- *
- * @param in string to be URI checked
+ * @param in string to check
  * @param isObjectName if true, '/' is allowed unencoded (object name context).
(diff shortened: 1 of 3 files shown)


maskit:
> canonicalEncode(): For partially-encoded strings, decodes first then re-encodes to ensure consistent canonical output

How do you know if it's partial/mixed encoding? The input might be unencoded and the expectation could be encoding `%5B` as `%255B`.

bryancall:
Good question about partially-encoded URLs like `/app/(channel)/%5B%5Bparts%5D%5D/page.js`.

To generate a matching signature, we need to canonicalize the URI the same way AWS does. Our approach decodes then re-encodes to produce a consistent canonical form:

```
Input (any of these):
  /app/(channel)/%5B%5Bparts%5D%5D/page.js     (partial)
  /app/%28channel%29/%5B%5Bparts%5D%5D/page.js (full)
  /app/(channel)/%5b%5bparts%5d%5d/page.js     (lowercase hex)

Canonical form (always):
  /app/%28channel%29/%5B%5Bparts%5D%5D/page.js
```

This ensures our signature matches AWS regardless of how the client encoded the request.

maskit:
I'm not sure if IT understood my question correctly. I'm saying the input it thinks partially-encoded could be actually fully-unencoded.

PR #12803, opened 2026-01-12
Fix NetAcceptAction::cancel() use-after-free race condition

bneradt:
Fix a race condition between NetAcceptAction::cancel() and NetAccept::acceptEvent() where the server pointer could be dereferenced after the NetAccept object was deleted.

The race occurred as follows:
1. Thread T4 calls NetAcceptAction::cancel(), sets cancelled=true
2. Thread T3 running acceptEvent() sees cancelled==true
3. Thread T3 deletes the NetAccept (including embedded Server)
4. Thread T4 tries to call server->close() on freed memory

The fix uses std::atomic<Server*> with atomic exchange to ensure only one thread can successfully obtain and use the server pointer. Both cancel() and the cleanup paths before delete this atomically exchange the pointer with nullptr - whichever succeeds first closes the server, the other becomes a no-op.

This addresses the TODO comment that was in the code: "// TODO fix race between cancel accept and call back"

ASAN report this fixes (seen intermittently on rocky CI builds):

```
==8850==ERROR: AddressSanitizer: heap-use-after-free on address 0x616000028cb4 at pc 0x000001346739 bp 0x7fa40fd2f580 sp 0x7fa40fd2f570 WRITE of size 4 at 0x616000028cb4 thread T4 ([ET_NET 2])
    #0 0x1346738 in UnixSocket::close() ../src/iocore/eventsystem/UnixSocket.cc:138
    #1 0x12b44ed in Server::close() ../src/iocore/net/Server.cc:88
    #2 0x121fb95 in NetAcceptAction::cancel(Continuation*) ../src/iocore/net/P_NetAccept.h:71
    #3 0x7fa41686d082 in TSActionCancel(tsapi_action*) ../src/api/InkAPI.cc:5828
    ...

0x616000028cb4 is located 308 bytes inside of 576-byte region [0x616000028b80,0x616000028dc0) freed by thread T3 ([ET_NET 1]) here:
    #0 0x7fa416d2036f in operator delete(void*, unsigned long)
    #1 0x12593c4 in NetAccept::~NetAccept() ../src/iocore/net/P_NetAccept.h:128
    #2 0x12bebf0 in NetAccept::acceptEvent(int, void*) ../src/iocore/net/UnixNetAccept.cc:484
    ...
```

### DIFF
--- a/src/iocore/net/UnixNetAccept.cc
+++ b/src/iocore/net/UnixNetAccept.cc
@@ -479,6 +479,7 @@ NetAccept::acceptEvent(int event, void *ep)
   MUTEX_TRY_LOCK(lock, m, e->ethread);
   if (lock.is_locked()) {
     if (action_->cancelled) {
+      // Server was already closed by whoever called cancel().
       e->cancel();
       Metrics::Gauge::decrement(net_rsb.accepts_currently_open);
       delete this;
@@ -487,6 +488,7 @@ NetAccept::acceptEvent(int event, void *ep)
 
     int res;
     if ((res = net_accept(this, e, false)) < 0) {
+      action_->cancel();
       Metrics::Gauge::decrement(net_rsb.accepts_currently_open);
       /* INKqa11179 */
       Warning("Accept on port %d failed with error no %d", ats_ip_port_host_order(&server.addr), res);
@@ -637,7 +639,7 @@ Ldone:
   return EVENT_CONT;
 
 Lerror:
-  server.close();
+  action_->cancel();
   e->cancel();
   Metrics::Gauge::decrement(net_rsb.accepts_currently_open);
   delete this;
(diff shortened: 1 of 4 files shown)


src/iocore/net/UnixNetAccept.cc
```diff
@@ -480,13 +480,21 @@ NetAccept::acceptEvent(int event, void *ep)
   if (lock.is_locked()) {
     if (action_->cancelled) {
       e->cancel();
+      Server *s = action_->server.exchange(nullptr, std::memory_order_acq_rel);
+      if (s != nullptr) {
+        s->close();
+      }
```
cmcfarlen: This is repetitive and could be an issue if the repeated code needs to change.  I recommend adding a function to clear and close, or perhaps calling cancel on the action is fine.

cmcfarlen (approved):
Nice!

PR #12806, opened 2026-01-13
Fix Valgrind uninitialized memory warnings

bneradt:
Valgrind reported some uses of uninitialized memory when running the
regression tests under it. This patch addresses the following valgrind
reported issues:

1. LogBuffer flush:
   Syscall param write(buf) points to uninitialised byte(s)
      at Log::flush_thread_main(void*)
      by LogBuffer::LogBuffer(LogConfig const*, LogObject*, ...)

2. Cache test pwrite:
   Syscall param pwrite64(buf) points to uninitialised byte(s)
      at AIOThreadInfo::aio_thread_main(AIOThreadInfo*)
      by Stripe::Stripe(CacheDisk*, long, long, int, int)


masaori335 (approved):
Nice catch. Both memories are just allocated and not initialized.

PR #12809, opened 2026-01-14
header_rewrite fix storage lifetime for LAST-CAPTURE

cmcfarlen:
There is a bug in LAST-CAPTURE where the stored matches would point into a temporary and cause issues with using the last captured value.

This PR move the storage for the regex subject to be in the Resources object so it can live through the entire hook.


bneradt (approved):

cmcfarlen:
Added to milestone 10.1.1 via #12819

PR #12811, opened 2026-01-14
ja*: Fix preserve logic to check for any header in fingerprint group

bneradt:
Problem:
When --preserve was enabled and a request passed through multiple proxies, each header was checked individually. This could result in mismatched fingerprint data - for example, x-ja3-raw being added by a downstream proxy while x-ja3-sig was preserved from an upstream proxy.

Solution:
The JA3 and JA4 fingerprint plugins now check if ANY header in a fingerprint group exists before adding headers. If any header in the group exists, ALL headers in that group are skipped.

Changes:
- ja3_fingerprint: Added group-level checks for JA3 headers
- ja4_fingerprint: Added --preserve option and group-level check for JA4 headers
- Updated tests to verify group-level preserve behavior


plugins/experimental/ja4_fingerprint/plugin.cc
```diff
@@ -334,12 +369,36 @@ handle_read_request_hdr(TSCont cont, TSEvent event, void *edata)
...
+  return false;
+}
+
 void
 append_JA4_headers(TSCont /* cont ATS_UNUSED */, TSHttpTxn txnp, std::string const *fingerprint)
 {
   TSMBuffer bufp;
   TSMLoc    hdr_loc;
-  if (TS_SUCCESS == TSHttpTxnClientReqGet(txnp, &bufp, &hdr_loc)) {
+  if (TS_SUCCESS != TSHttpTxnClientReqGet(txnp, &bufp, &hdr_loc)) {
```
jasmine-nahrain, line 389: Would it make more sense to say TS_ERROR == ? I think it might be easier to read the code that way

bneradt: Yeah, I see your perspective. I think I'll stick with this though for the sake of consistency since that's what the other checks for error do in this plugin.

jasmine-nahrain (approved):

PR #12813, opened 2026-01-15
Support remap.yaml

serrislew:
Support YAML format for remap rules. Not replacing remap.config but providing additional remap option to define configs in yaml. If no `remap.yaml` is found, defaults back to `remap.config`

"Prerequisite" for #12669, to YAML'fy remap configs before moving into virtualhost.yaml


src/proxy/http/remap/RemapYamlConfig.cc
```diff
@@ -0,0 +1,1170 @@
...
+    return swoc::Errata("Can't load toURL from URL class");
+  }
+  pargv[parc++] = ats_strdup(err);
+  ats_free(err);
+
+  // Add plugin parameters
+  if (node["params"] && node["params"].IsSequence()) {
+    for (const auto &param : node["params"]) {
+      std::string pparam_str = param.as<std::string>();
+      pargv[parc++]          = ats_strdup(pparam_str.c_str());
```
bneradt: Are these ats_strdup freed? Indices 0 and 1 are below, but it looks like these from the loop need to be freed.

src/proxy/http/remap/RemapYamlConfig.cc
```diff
@@ -0,0 +1,1170 @@
...
+    uint32_t elevate_access = 0;
+    elevate_access          = RecGetRecordInt("proxy.config.plugin.load_elevated").value_or(0);
+    ElevateAccess access(elevate_access ? ElevateAccess::FILE_PRIVILEGE : 0);
+
+    pi =
+      bti->rewrite->pluginFactory.getRemapPlugin(swoc::file::path(plugin_name), parc, pargv, error, isPluginDynamicReloadEnabled());
+  } // done elevating access
+
+  if (nullptr == pi) {
+    return swoc::Errata("failed to instantiate plugin ({}) to remap rule: {}", plugin_name.c_str(), error.c_str());
```
bneradt, line 477: This early return will leak the pargv strdups.

Can we use std::vector<std::string> instead to manage this memory?

src/proxy/http/remap/RemapYamlConfig.cc
```diff
@@ -0,0 +1,1170 @@
...
+  // Check for activate_filters directive
+  if (node["activate_filter"]) {
+    std::string filter_name = node["activate_filter"].as<std::string>();
+
+    // Check if for ip_allow filter
+    if (strcmp(filter_name.c_str(), "ip_allow") == 0) {
+      bti->ip_allow_check_enabled_p = true;
+      return {};
+      ;
+    }
```
bneradt, line 497: Extra `;`

src/proxy/http/remap/RemapYamlConfig.cc
```diff
@@ -0,0 +1,1170 @@
...
+  if (node["src_ip_invert"]) {
+    if (node["src_ip_invert"].IsSequence()) {
+      for (const auto &src_ip_invert : node["src_ip_invert"]) {
+        auto errata = parse_src_ip(src_ip_invert.as<std::string>(), true);
+        if (!errata.is_ok()) {
+          if (new_rule_flg) {
+            delete rule;
+            *rule_pp = nullptr;
+          }
+          return errata;
```
bneradt: This is repeated frequently. Maybe something `PostScript` can help with?

src/proxy/http/remap/RemapYamlConfig.cc
```diff
@@ -0,0 +1,1170 @@
...
+        if (new_rule_flg) {
+          delete rule;
+          *rule_pp = nullptr;
+        }
+        return errata;
+      }
+    }
+  }
+
+  // Parse src_ip_category (and src_ip_category)
```
bneradt: Probably: "(and src_ip_category_invert)"

bneradt (changes requested):
Looks good. Thanks for helping us make progress on this.

src/proxy/http/remap/unit-tests/test_RemapRulesYaml.cc
```diff
@@ -0,0 +1,256 @@
...
+
+  void
+  testRunStarting(Catch::TestRunInfo const & /* testRunInfo ATS_UNUSED */) override
+  {
+    Thread *main_thread = new EThread();
+    main_thread->set_specific();
+
+    DiagsPtr::set(new Diags("test_RemapRulesYaml", "*", "", new BaseLogFile("stderr")));
+    // diags()->activate_taglist(".*", DiagsTagType_Debug);
+    // diags()->config.enabled(DiagsTagType_Debug, 1);
```
bneradt: Should these commented out lines maybe be removed? Or is this a convenience for future debugging? If the latter, maybe add a comment saying something "Uncomment these for extra debug output during the test run."

src/proxy/http/remap/RemapYamlConfig.cc
```diff
@@ -0,0 +1,1128 @@
...
+parse_yaml_filter_directive(const YAML::Node &node, BUILD_TABLE_INFO *bti)
+{
+  acl_filter_rule *rp;
+
+  // Check for activate_filters directive
+  if (node["activate_filter"]) {
+    std::string filter_name = node["activate_filter"].as<std::string>();
+
+    // Check if for ip_allow filter
+    if (strcmp(filter_name.c_str(), "ip_allow") == 0) {
```
bneradt: I think `filter_name == "ip_allow"` should work

src/proxy/http/remap/RemapYamlConfig.cc
```diff
@@ -0,0 +1,1128 @@
...
+    }
+  }
+  return {};
+}
+
+swoc::Errata
+parse_yaml_plugins(const YAML::Node &node, url_mapping *url_mapping, BUILD_TABLE_INFO *bti)
+{
+  char *err;
+  char *pargv[1024];
```
bneradt: Can `pargv` be a `std::vector<std::string>`?

src/proxy/http/remap/RemapYamlConfig.cc
```diff
@@ -0,0 +1,1128 @@
...
+  std::string_view fromHost{}, toHost{};
+  std::string_view fromUrl{}, toUrl{};
+  std::string_view fromPath;
+  char            *fromHost_lower     = nullptr;
+  char            *fromHost_lower_ptr = nullptr;
+  char             fromHost_lower_buf[1024];
+  mapping_type     maptype;
+  url_mapping     *new_mapping = nullptr;
+
+  UrlRewrite::RegexMapping *reg_map = nullptr;
```
bneradt: Maybe `std::unique_ptr` for these two?

src/proxy/http/remap/RemapYamlConfig.cc
```diff
@@ -0,0 +1,1128 @@
...
+    return swoc::Errata("Invalid argument(s)");
+  }
+
+  if (dbg_ctl_url_rewrite.on()) {
+    printf("validate_filter_args: ");
+    for (const auto &rule : node) {
+      printf("\"%s\" ", rule.first.as<std::string>().c_str());
+    }
+    printf("\n");
+  }
```
bneradt: I believe these should use DbgPrint instead of printf.

src/proxy/http/remap/RemapYamlConfig.cc
```diff
@@ -0,0 +1,1128 @@
...
+
+/** will process the regex mapping configuration and create objects in
+    output argument reg_map. It assumes existing data in reg_map is
+    inconsequential and will be perfunctorily null-ed;
+*/
+static bool
+process_regex_mapping_config(const char *from_host_lower, url_mapping *new_mapping, UrlRewrite::RegexMapping *reg_map)
+{
+  std::string_view to_host{};
+  int              to_host_len;
```
bneradt: This looks like it has code duplicated with a function of the same name in `RemapConfig.cc`. Maybe we can extract out logic to a shared function to deduplicate?

src/proxy/http/remap/unit-tests/test_RemapRulesYaml.cc
```diff
@@ -0,0 +1,256 @@
...
+    Layout::create();
+    RecProcessInit(diags());
+    LibRecordsConfigInit();
+  }
+};
+
+CATCH_REGISTER_LISTENER(TestListener);
+
+swoc::file::path
+write_test_remap(const std::string &config, const std::string &tag)
```
bneradt, line 73: Are these temporary files ever cleaned up? I don't think they are removed.

src/proxy/http/remap/RemapYamlConfig.cc
```diff
@@ -0,0 +1,1128 @@
...
+  Dbg(dbg_ctl_url_rewrite, "[%s] including remap configuration from %s", __func__, path);
+  success = remap_parse_yaml_bti(path, &nbti);
+
+  // The sub-parse might have updated the rules list, so push it up to the parent parse.
+  bti->rules_list = nbti.rules_list;
+
+  if (success) {
+    // register the included file with the management subsystem so that we can correctly
+    // reload them when they change
+    load_remap_file_cb(ts::filename::REMAP, path);
```
bneradt: Should this be `REMAP_YAML` ?

bneradt (changes requested):

src/proxy/http/remap/unit-tests/test_RemapRulesYaml.cc
serrislew: Hmm not sure, it was copied from `test_RemapRules.cc`. I will remove it

src/proxy/http/remap/RemapYamlConfig.cc
serrislew: Yeah this was also copied from `RemapConfig.cc`. I will update, thanks!

src/proxy/http/remap/RemapYamlConfig.cc
```diff
@@ -0,0 +1,1046 @@
...
+remap_parse_yaml_bti(const char *path, BUILD_TABLE_INFO *bti)
+{
+  try {
+    Dbg(dbg_ctl_remap_yaml, "Parsing YAML config file: %s", path);
+
+    YAML::Node config = YAML::LoadFile(path);
+
+    if (config.IsNull()) {
+      Dbg(dbg_ctl_remap_yaml, "Empty YAML config file");
+      return true; // a missing file is ok - treat as empty, no rules.
```
brbzull0, line 974: the comment `a missing file is ok`  could be a bit misleading for the reader, if the file is missing, I think it will never get this far, IIRC it will throw `YAML::BadFile`.

serrislew: but since we currently support both remap.yaml and remap.config, then missing file would just assume legacy config. 

And remap.config also supports missing file [here](https://github.com/serrislew/trafficserver/blob/b01d0d8e401623b2ad63024d4962cf9ea882e31e/src/proxy/http/remap/RemapConfig.cc#L1095-L1098)

src/proxy/http/remap/RemapYamlConfig.cc
```diff
@@ -0,0 +1,1046 @@
...
+    for (const auto &param : node["params"]) {
+      param_storage.push_back(param.as<std::string>());
+      Dbg(dbg_ctl_remap_yaml, "  Plugin param: %s", param_storage.back().c_str());
+    }
+  }
+
+  std::vector<char *> pargv;
+  for (auto &param : param_storage) {
+    pargv.push_back(const_cast<char *>(param.c_str()));
+  }
```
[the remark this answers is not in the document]
bneradt: Codex highlighted this too. On the face of it, looking at the patch, this makes sense. Is the `url_mapping->getPluginParamsStorage` shared across the plugins in a remap node?

bneradt (changes requested):

bneradt (approved):

serrislew:
Note: updates were made
- `mmdb` plugin: checks both remap configs to tie the active file during reload
- `FileManager.h`: allows deletion to be recognized for config reload so if remap.yaml gets deleted, remap.config can be used after the reload
- uses new reload ConfigRegistry

bneradt (approved):

src/mgmt/config/FileManager.cc
```diff
@@ -381,6 +381,16 @@ FileManager::ConfigManager::checkForUserUpdate(FileManager::RollBackCheckType ho
   ink_mutex_acquire(&fileAccessLock);
 
   if (this->statFile(&fileInfo) < 0) {
+    // File doesn't exist. If it previously existed (fileLastModified > 0),
+    // treat the deletion as a change so the reload handler can fall back
+    // to an alternative config file (e.g. remap.yaml -> remap.config).
+    if (fileLastModified > 0) {
+      if (how == FileManager::ROLLBACK_CHECK_AND_UPDATE) {
+        fileLastModified = 0;
```
brbzull0, line 389: I think I can add this feature inside config_registry at some point in the future.

brbzull0 (approved):

serrislew:
Merged with #12997

PR #12818, opened 2026-01-16
Keep a copy of the effective client ip for logging

cmcfarlen:
Fixes a bug introduced in #12761:

If the transaction client vc went away before logging, then the update marshal client ip function didn't actually marshal anything and it messed up parsing the log entry later.

This fix is to just keep a copy of the effective IP for logging purposes, per tradition.


cmcfarlen:

PR #12820, opened 2026-01-16
hrw4u: Fix section validations and operators

zwoop:


tools/hrw4u/tests/test_lsp.py
```diff
@@ -360,10 +360,11 @@ def shared_lsp_client():
...
 
 @pytest.mark.parametrize(
-    "section,prefix,should_allow", [
+    "section,prefix,should_allow",
+    [
         ("SEND_REQUEST", "outbound.req.", True),
-        ("REMAP", "outbound.req.", False),
-        ("SEND_REQUEST", "outbound.resp.", False),
+        ("REMAP", "outbound.req.", True),  # outbound.req. is available in REMAP
+        ("SEND_REQUEST", "outbound.resp.", True),  # outbound.resp. is available in SEND_REQUEST
```
bneradt: Is this correct? `outbound.resp` is available in `SEND_REQUEST` hook? Maybe there's a reason for this, but SEND_REQUEST hook seems to early to interact with `outbound.resp`.

bneradt (changes requested):

tools/hrw4u/tests/test_lsp.py
zwoop: I think I fixed these.

zwoop:
[approve ci autest]

PR #12821, opened 2026-01-16
Coverity Fixes

PSUdaemon:
Fixes 19 Coverity warnings. Also found and fixed a real bug - inverted condition in Plugin.cc that would crash on counter record lookups.

Most fixes are straightforward: uninitialized variables, unchecked return values, missing null checks, resource leaks in tests. A few are false positive suppressions with comments explaining why.

Destructor exception safety fixes for `HttpSM` and `HttpTransact::State`.

One commit per issue, CID in each commit message.

### DIFF
--- a/src/proxy/Plugin.cc
+++ b/src/proxy/Plugin.cc
@@ -254,7 +254,7 @@ plugin_expand(char *arg)
   }
   case RECD_COUNTER: {
     auto count_val{RecGetRecordCounter(arg)};
-    if (count_val) {
+    if (!count_val) {
       goto not_found;
     }
     str = static_cast<char *>(ats_malloc(128));


PSUdaemon:
[approve ci OSX]

PSUdaemon:
[approve ci osx]

zwoop:
The elimination of lock guards for string metrics (with the finalize thing) seems weird. @cmcfarlen Can't we create string metrics after startup / initialization? Maybe we can't?

On the try / catch, I'm not convinced we should do that, but rather, understand why Coverity don't like those destructors / release mechanism, and/or why they can raise exceptions.

src/traffic_crashlog/traffic_crashlog.cc
```diff
@@ -148,6 +148,8 @@ crashlog_exec_pgm(FILE *fp, pid_t pid)
 int
 main(int /* argc ATS_UNUSED */, const char **argv)
 {
+  // coverity[fun_call_w_exception] - ats_as_c_str checks optional before .value(),
+  // and callers guard with if(opt) before calling ats_as_c_str(opt)
   FILE           *fp;
   char           *logname;
```
bneradt, line 202: Can you explain how this suppression is working? I would have guessed tha the coverity exemption comment would have to be before the function call itself rather than before the FILE declaration. I might be missing something though.

bneradt (changes requested):
Looks great. Thanks for the fixes.

I walked through all the commits (thanks for keeping them separate) and they all seem make sense to me, but I have a question about the crashlog one.

src/traffic_crashlog/traffic_crashlog.cc
PSUdaemon: My understanding is that it can be at function scope as well, but that's probably too broad for this instance and I have moved it. Thanks for pointing this out.

PSUdaemon:
> The elimination of lock guards for string metrics (with the finalize thing) seems weird. @cmcfarlen Can't we create string metrics after startup / initialization? Maybe we can't?

It appears that plugin metric creation goes through `Metrics::Storage::create()`, not `StaticString::createString()`. That has its own mutex.

PSUdaemon:
[approve ci osx]

PSUdaemon:
[approve ci autest]
[approve ci freebsd]

cmcfarlen:
> The elimination of lock guards for string metrics (with the finalize thing) seems weird. @cmcfarlen Can't we create string metrics after startup / initialization? Maybe we can't?
> 
> On the try / catch, I'm not convinced we should do that, but rather, understand why Coverity don't like those destructors / release mechanism, and/or why they can raise exceptions.

Yeah, it's possible that some external plugin (or someone's custom code that isn't open sourced) makes string metrics after startup.  I don't think we should restrict it to static initializers.

PSUdaemon:
[approve ci debian]

PR #12823, opened 2026-01-20
Count proxy.process.http.incoming_requests at transaction start

bneradt:
The proxy.process.http.incoming_requests metric was only incremented in HttpTransact::HandleRequest(), which is called late in the request processing pipeline. Requests that received error or redirect responses before reaching HandleRequest were never counted, causing significant underreporting during high-error-rate scenarios.

This fix moves the increment to HttpSM::attach_client_session(), which is called at the start of every transaction. This ensures all requests are counted, including those that fail header parsing, receive remap redirects, or hit early error conditions. This aligns with the documented behavior that incoming_requests should include errors and redirects (see doc/appendices/command-line/traffic_top.en.rst):

    "Total number of client requests serviced by Traffic Server. This
    includes both successful content-bearing responses as well as
    errors, redirects, and not-modified IMS responses."

### DIFF
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -451,6 +451,11 @@ HttpSM::attach_client_session(ProxyTransaction *txn)
   t_state.client_info.is_transparent = netvc->get_is_transparent();
   t_state.client_info.port_attribute = static_cast<HttpProxyPort::TransportType>(netvc->attributes);
 
+  Metrics::Counter::increment(http_rsb.incoming_requests);
+  if (t_state.client_info.port_attribute == HttpProxyPort::TRANSPORT_SSL) {
+    Metrics::Counter::increment(http_rsb.https_incoming_requests);
+  }
+
   // Record api hook set state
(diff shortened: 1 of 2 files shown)


maskit:
The new location looks better, but I'm wondering if this is the earliest possible point to increment the metric. I think H2 requests that have decoding errors (e.g. unrecognized pseudo headers, compression errors, etc) are still not counted.

bneradt:
> The new location looks better, but I'm wondering if this is the earliest possible point to increment the metric. I think H2 requests that have decoding errors (e.g. unrecognized pseudo headers, compression errors, etc) are still not counted.

You're right that H2 HPACK decode errors occur before new_transaction() and won't be counted. This fix addresses HTTP-level early failures (remap redirects, header parse errors, early error responses) which was the primary concern — these are now counted correctly.

H2/H3 frame-level decode failures are a different category. Counting those would require incrementing in the frame processing code before decode, which is more involved.

I suggest doing this simple move change which relatively easily makes the current counting more accurate.

PR #12824, opened 2026-01-23
Fix Cache Result Code of negative revalidation cases

masaori335:
We found an issue of Cache Result Code (`%<crc>`) logging field. When ATS served stale contents by negative revalidation feature, `TCP_REFRESH_FAIL_HIT` is expected. However, we saw `TCP_MISS` and `TCP_IMS_MISS`.

This PR includes extension of access log verification of `ats_replay`. (the first commit)


bneradt (approved):
lgtm

PR #12834, opened 2026-01-27
Make 9.2.x buildable on macOS 26

maskit:
Adding `-Wno-cast-function-type-mismatch` isn't great, but I don't think we should make changes that this check may find, on 9.2.x.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -853,7 +853,7 @@ AS_IF([test "x${has_optimizer_flags}" = "xno"],
         ]
 )
 
-AS_IF([test "x$ac_cv_c_compiler_version" = "x17"], [NO_WARN_CAST_FUNCTION_TYPE_MISMATCH="-Wno-cast-function-type-mismatch"], [NO_WARN_CAST_FUNCTION_TYPE_MISMATCH="$ac_cv_c_compiler_version"])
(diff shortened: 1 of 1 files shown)


include/tscore/ink_endian.h
```diff
@@ -26,8 +26,10 @@
 #include "tscore/ink_config.h"
 
 #ifdef HAVE_SYS_ENDIAN_H
+#if !defined(darwin)
```
maskit, line 29: Old versions of macOS didn't have be64toh and the other functions, but now those are defined on endian.h, and those cause errors due to duplication.

I could add a check for the availability of those functions, but this small change is enough. We don't have this issue on newer codebase.

configure.ac
```diff
@@ -658,6 +658,11 @@ AM_PROG_AS
 
 AX_COMPILER_VENDOR
 
+AC_MSG_CHECKING([for C compiler version])
+# Capture the output of the compiler version command
+ac_cv_c_compiler_version=`${CC} --version 2>&1 | grep "Apple clang version" | head -n 1 | awk '{print $4}' | cut -d'.' -f1`
```
maskit, line 663: This is hacky and only works for Apple Clang 17, but I don't think we will need to maintain it for a long time.

maskit:
@ezelkow1 I'll leave this open for you not to mess up the ongoing release process. I believe the Milestone needs to be updated to 9.2.13. Please push the merge button when you are ready.

PR #12841, opened 2026-01-29
Revert "Fix NetAcceptAction::cancel() use-after-free race condition"

bneradt:
Reverts apache/trafficserver#12803

I suspect this is introducing different crashes in NetAccept::cancel.

### DIFF
--- a/src/iocore/net/P_NetAccept.h
+++ b/src/iocore/net/P_NetAccept.h
@@ -43,7 +43,6 @@
 #include "iocore/net/NetAcceptEventIO.h"
 #include "Server.h"
 
-#include <atomic>
 #include <vector>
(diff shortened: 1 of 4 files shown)


maskit:
Should we remove Milestone from both this and https://github.com/apache/trafficserver/pull/12803 ? I don't think these two need to be on 10.2 release note since nothing changed effectively.

bneradt:
> Should we remove Milestone from both this and #12803 ? I don't think these two need to be on 10.2 release note since nothing changed effectively.

Thanks. That's a good idea.

zwoop:
Yeah, definitely clear both the Milestones, unless the first one has been in a release.

PR #10980, opened 2024-01-09
TSFile - remove from background_fetch plugin

ezelkow1:
TSFile and related plugin API are being deprecated. This removes those from the background_fetch plugin, along with some other clean up.

This replaces #10909, attribution should go to @SolidWallOfCode who is the main contributor of this code (and included in the contributors list)

Co-authored-by: Alan M. Carroll <amc@apache.org>

PR #12842, opened 2026-01-29
Fix Coverity UNINIT issues

bryancall:
## Summary

Fix four Coverity UNINIT (uninitialized variable) issues:

- **CID 1497238**: Value-initialize `DiagsConfigState` in `reconfigure_diags()`
- **CID 1644237**: Value-initialize `ParsedValue` in `ParsedConfigCache::parse()`
- **CID 1644228**: Value-initialize `TLSClientHelloSummary` in `test_ja4.cc`
- **CID 1533658**: Fix inverted condition for Client-IP parsing in background_fetch and cache_fill plugins

The last fix (CID 1533658) is an actual logic bug introduced in #10980 - the condition `cfg_value.front() == '*'` should be `cfg_value.front() != '*'` to correctly skip loading when the value is exactly `"*"` (which signals "match any address" via an empty IPRange).

### DIFF
--- a/plugins/background_fetch/configs.cc
+++ b/plugins/background_fetch/configs.cc
@@ -136,7 +136,7 @@ BgFetchConfig::readConfig(const char *config_file)
         if ("Client-IP"_tv == cfg_name) {
           swoc::IPRange r;
           // '*' is special - match any address. Signalled by empty range.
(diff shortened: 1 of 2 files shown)


bryancall:
[approve ci]

PR #12843, opened 2026-01-30
Coverity Fixes

PSUdaemon:
Fix one new Coverity issue and attempt to fix 3 of the previous suppressions that were not recognized by the last Coverity run.

### DIFF
--- a/src/iocore/cache/Stripe.cc
+++ b/src/iocore/cache/Stripe.cc
@@ -170,10 +170,10 @@ Stripe::_init_directory(std::size_t directory_size, int header_size, int footer_
   this->directory.footer = reinterpret_cast<StripeHeaderFooter *>(this->directory.raw_dir + footer_offset);
 }
 
+// coverity[exn_spec_violation] - ink_assert aborts (doesn't throw), Dbg is exception-safe
 Stripe::~Stripe()
 {
   if (this->directory.raw_dir != nullptr) {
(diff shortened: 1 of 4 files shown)


ezelkow1 (approved):

src/iocore/cache/Stripe.cc
```diff
@@ -170,10 +170,10 @@ Stripe::_init_directory(std::size_t directory_size, int header_size, int footer_
...
 }
 
+// coverity[exn_spec_violation] - ink_assert aborts (doesn't throw), Dbg is exception-safe
 Stripe::~Stripe()
 {
   if (this->directory.raw_dir != nullptr) {
-    // coverity[fun_call_w_exception] - ink_assert aborts (doesn't throw), Dbg is exception-safe
     // Debug logging to track cleanup - helps correlate with crash location
     Dbg(dbg_ctl_cache_free, "Stripe %s: freeing raw_dir=%p size=%zu huge=%s", hash_text.get() ? hash_text.get() : "(null)",
         this->directory.raw_dir, this->directory.raw_dir_size, this->directory.raw_dir_huge ? "true" : "false");
```
bneradt, line 179: Although, I think in our experience, Dbg statements in destructors can be problematic due to unfortunate ordering of destructed global static objects.

But we can address that separately.

bneradt (approved):

PR #12845, opened 2026-01-31
Add fixes around fail_action 5

ezelkow1:
This adds some fixes around fail_action 5. First it prevents a possible looping scenario, previously after exhausting waiting for a read lock after having failed in a write lock it would loop back around to check for write again. Now it will just go to origin. While this allows an extra request through it can prevent some odd scenarios, including one where it just writes back to the cache again anyway which would already be useless. Also if the original rww request is very slow then this allows waiters to bypass it.

Second there could be an issue around redirects, normally a redirect is fine since it would have a different cachekey but if we enable use_orig then we could end up in the same looping contention in that scenario as well.

Also adds some autests, though we should keep an eye on these since they could be timing dependent. So if we start getting failures we can turn them off.


tests/gold_tests/cache/cache-read-retry-mode.test.py
```diff
@@ -0,0 +1,33 @@
...
+1. Basic read-while-writer behavior with fail_action=5
+2. READ_RETRY mode configuration is accepted and functional
+3. System does not crash under write lock contention
+4. Requests are served correctly when read retries are exhausted
+'''
+
+Test.ContinueOnFail = True
+
+# Tests using Proxy Verifier replays (deterministic, CI-friendly)
+# These validate basic READ_RETRY behavior without timing-sensitive parallel requests
```
bneradt: Given the pretty thorough `Test.Summary` above, let's just remove these comments.

tests/gold_tests/cache/replay/cache-read-retry-basic.replay.yaml
```diff
@@ -0,0 +1,180 @@
...
+      proxy.config.diags.debug.enabled: 1
+      proxy.config.diags.debug.tags: 'http|cache|http_cache'
+      # Enable READ_RETRY mode
+      proxy.config.http.cache.open_write_fail_action: 5
+      # Configure retry parameters
+      proxy.config.http.cache.max_open_write_retries: 1
+      proxy.config.http.cache.max_open_write_retry_timeout: 0
+      proxy.config.http.cache.max_open_read_retries: 20
+      proxy.config.http.cache.open_read_retry_time: 100
+      # Enable read-while-writer
```
bneradt: Let's remove this comment too.

tests/gold_tests/cache/replay/cache-read-retry-basic.replay.yaml
```diff
@@ -0,0 +1,180 @@
...
+# - Only ONE origin connection should be made (request collapsing)
+#
+
+meta:
+  version: "1.0"
+
+autest:
+  dns:
+    name: 'dns-read-retry-basic'
+  description: 'Test cache_open_write_fail_action = 5 (READ_RETRY) with successful read-while-writer'
```
bneradt: Let's switch the order of these - description should be at the top of the autest section.

tests/gold_tests/cache/replay/cache-read-retry-basic.replay.yaml
```diff
@@ -0,0 +1,180 @@
...
+              - [X-Request, first]
+
+        proxy-request:
+          headers:
+            fields:
+              - [X-Request, {value: 'first', as: equal}]
+
+        server-response:
+          # Slow response (3 seconds) to allow second request to retry reads
+          delay: 3s
```
bneradt: 3 seconds is more than we need. I know it only saves a second, but let's do 2s here and 3 for the later third-request-cache-hit transaction.

Remember to update the "Scenario" comments at the top of this file with the updated timing.

tests/gold_tests/cache/replay/cache-read-retry-exhausted.replay.yaml
```diff
@@ -0,0 +1,235 @@
...
+# 4. Content is cached properly
+#
+
+meta:
+  version: "1.0"
+
+autest:
+  dns:
+    name: 'dns-read-retry-exhausted'
+  description: 'Test cache_open_write_fail_action = 5 (READ_RETRY) with slow origin and concurrent requests to verify request collapsing and stability'
```
bneradt: Again, let's switch the description to be at the top. That's more readable for humans.

ezelkow1:
[approve ci autest]

tests/gold_tests/cache/replay/cache-read-retry-exhausted.replay.yaml
```diff
@@ -0,0 +1,235 @@
...
+      proxy.config.diags.debug.tags: 'http|cache|http_cache|http_trans'
+      # Enable READ_RETRY mode
+      proxy.config.http.cache.open_write_fail_action: 5
+      # Configure retry parameters to exhaust quickly
+      proxy.config.http.cache.max_open_write_retries: 1
+      proxy.config.http.cache.max_open_write_retry_timeout: 0
+      # Only 5 read retries @ 100ms = 500ms total (not enough for 10s origin)
+      proxy.config.http.cache.max_open_read_retries: 5
+      proxy.config.http.cache.open_read_retry_time: 100
+      # Enable read-while-writer
```
bneradt: We can remove this comment.

tests/gold_tests/cache/replay/cache-read-retry-basic.replay.yaml
```diff
@@ -0,0 +1,180 @@
...
+          url: /bigfile
+          headers:
+            fields:
+              - [uuid, third-request-cache-hit]
+              - [Host, example.com]
+              - [X-Request, third]
+
+        # Server should NOT receive this request (should be cache hit)
+        server-response:
+          status: 200
```
bneradt: Let's make this a 400 response so we are doubly sure we get the cached 200. That adds another check that we get the cached, not origin, response.

tests/gold_tests/cache/replay/cache-read-retry-exhausted.replay.yaml
```diff
@@ -0,0 +1,235 @@
...
+
+        proxy-request:
+          headers:
+            fields:
+              - [X-Request, {value: 'second-exhausted', as: equal}]
+
+        # This request should NOT reach the server if read-while-writer works
+        # It should be served from Request 1's response
+        server-response:
+          status: 200
```
bneradt: Again, let's make this a 400 so we are sure we get the cached 200.

tests/gold_tests/cache/replay/cache-read-retry-exhausted.replay.yaml
```diff
@@ -0,0 +1,235 @@
...
+              - [X-Response, {value: 'first-origin', as: equal}]
+
+  #############################################################################
+  # Second session: Concurrent request that exhausts READ_RETRY attempts
+  # This runs in parallel and triggers the bug path
+  #############################################################################
+  - transactions:
+      - client-request:
+          # Small delay to ensure first request gets write lock first
+          delay: 50ms
```
bneradt: I suggest 100ms for this one, and 200ms for the second one.

tests/gold_tests/cache/replay/cache-read-retry-exhausted.replay.yaml
```diff
@@ -0,0 +1,235 @@
...
+              # Should get first-origin via read-while-writer (request collapsing works)
+              - [X-Response, {value: 'first-origin', as: equal}]
+
+  #############################################################################
+  # Third session: Another concurrent request to stress test the system
+  #############################################################################
+  - transactions:
+      - client-request:
+          # Arrives after second request starts but before first completes
+          delay: 100ms
```
bneradt: 200ms per my above suggestion.

tests/gold_tests/cache/replay/cache-read-retry-exhausted.replay.yaml
```diff
@@ -0,0 +1,235 @@
...
+              - [X-Request, third-concurrent]
+
+        proxy-request:
+          headers:
+            fields:
+              - [X-Request, {value: 'third-concurrent', as: equal}]
+
+        # This request will also be served via read-while-writer
+        server-response:
+          status: 200
```
bneradt: 400 instead of 200, please. Again, so we ensure we get the cached, not origin, response.

tests/gold_tests/cache/replay/cache-read-retry-exhausted.replay.yaml
```diff
@@ -0,0 +1,235 @@
...
+          url: /slowfile
+          headers:
+            fields:
+              - [uuid, fourth-request-verify]
+              - [Host, example.com]
+              - [X-Request, fourth-verify]
+
+        # Server should NOT receive this request (should be cache hit)
+        server-response:
+          status: 200
```
bneradt: 400 instead of 200 to ensure we get the cached response, not the origin one.

bneradt (changes requested):

tests/gold_tests/cache/replay/cache-read-retry-exhausted.replay.yaml
```diff
@@ -0,0 +1,235 @@
...
+
+    records_config:
+      proxy.config.diags.debug.enabled: 1
+      proxy.config.diags.debug.tags: 'http|cache|http_cache|http_trans'
+      # Enable READ_RETRY mode
+      proxy.config.http.cache.open_write_fail_action: 5
+      # Configure retry parameters to exhaust quickly
+      proxy.config.http.cache.max_open_write_retries: 1
+      proxy.config.http.cache.max_open_write_retry_timeout: 0
+      # Only 5 read retries @ 100ms = 500ms total (not enough for 10s origin)
```
bneradt: 10s origin? I think it's a 3s origin, right?

tests/gold_tests/cache/cache-read-retry-mode.test.py
```diff
@@ -0,0 +1,33 @@
...
+Test.ATSReplayTest(replay_file="replay/cache-read-retry-basic.replay.yaml")
+Test.ATSReplayTest(replay_file="replay/cache-read-retry-exhausted.replay.yaml")
```
bneradt: I'm not sure that these two tests test something fundamentally different. Nor tha the second is doing anything related to "exhausted". Maybe just use the second one since it is a bit more complex than the first and rename it to simply `cache-read-retry.replay.yaml`?

bneradt (changes requested):

src/proxy/http/HttpTransact.cc
```diff
@@ -3342,6 +3347,15 @@ HttpTransact::set_cache_prepare_write_action_for_new_request(State *s)
...
     ink_release_assert(s->redirect_info.redirect_in_process);
     s->cache_info.action = CacheAction_t::WRITE;
+  } else if (s->cache_info.write_lock_state == CacheWriteLock_t::READ_RETRY &&
+             (!s->redirect_info.redirect_in_process || s->txn_conf->redirect_use_orig_cache_key)) {
+    // Defensive: Should not reach here if HandleCacheOpenReadMiss check is working.
+    // If we somehow get here in READ_RETRY state, bypass cache unless we're in a redirect
+    // that uses a different cache key (redirect_use_orig_cache_key == 0).
+    // When redirect_use_orig_cache_key is enabled, the redirect uses the same cache key
+    // as the original request, so we'd hit the same write lock contention.
+    Warning("set_cache_prepare_write_action_for_new_request called with READ_RETRY state");
```
bneradt: Let's make this an Error instead of Warning.

bneradt (changes requested):

bneradt (approved):

PR #12846, opened 2026-02-02
Fix HostDBInfo::is_down condition

masaori335:
The condition of `HostDBInfo::is_down` seems opposite.

This function is only used by `HostDBRecord::select_best_srv()` , so only SRV record case is affected.

https://github.com/apache/trafficserver/blob/2a30bccc413c7a4abfcc2a410386a3469273ae55/src/iocore/hostdb/HostDB.cc#L1609-L1611

### DIFF
--- a/include/iocore/hostdb/HostDBProcessor.h
+++ b/include/iocore/hostdb/HostDBProcessor.h
@@ -230,7 +230,7 @@ HostDBInfo::is_alive()
 }
 
 /**
-  Check if this HostDBInfo is currently marked down (true) or not (false). Returns true while within the `fail_window` period after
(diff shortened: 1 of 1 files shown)


bryancall (approved):
Please make comments, it is confusing

bneradt:
Approved. Maybe check with @serrislew and @cmcfarlen as well since it sounds like they looked into this previously.

masaori335:
I wrote comment to clarify. Please take another look.

masaori335:
I talked with @serrislew and @cmcfarlen, we found her patch has almost the same change in this function with other changes. And we agreed with shipping this PR first.

PR #12851, opened 2026-02-03
Cleanup test_HostFile

masaori335:
`test_HostFile.cc` has some own implementations to avoid link issues. ~However, the issue seems resolved by CMake.~ At least, I don't see any build/link issue on my mac with LLVM toolchain. Let's see on other platforms.


masaori335:
Interesting. Linux builds are failed. It looks like `lld` (or modern linker) doesn't have the issue, but we still have issue with `ld`.

PR #12854, opened 2026-02-03
Fix LoadedPlugins::remove crash during static destruction

bneradt:
When a PluginFactory is destroyed during static destruction (after main exits), the EThreads are already gone. LoadedPlugins::remove() was crashing because it tried to acquire a mutex lock using this_ethread(), which returns nullptr during static destruction.

Now check if this_ethread() returns nullptr and handle cleanup directly in that case, since we're exiting anyway and don't need the mutex protection or scheduled deletion.

This issue was discovered while testing the header_rewrite run-plugin feature, which creates a PluginFactory that persists as a static object and is destroyed during static destruction.


masaori335 (approved):
Makes sense.

zwoop (approved):

cmcfarlen:

PR #12855, opened 2026-02-03
Fix header_rewrite run-plugin relative path resolution

bneradt:
This will need the #12854 PluginFactory fix before this test passes.

---

The plugin factory was not initialized when header_rewrite was used only as a remap plugin without a geo database. This caused run-plugin with relative paths to fail because there were no search directories configured. Now initHRWLibraries() is always called in TSRemapNewInstance().

Added a test to the header_rewrite bundle to verify run-plugin works with relative paths in remap mode.


zwoop (approved):
Love it

cmcfarlen:

PR #12857, opened 2026-02-04
Fix autest compatibility with Fedora 43 / Python 3.14

bryancall:
## Summary

This PR fixes multiple autest compatibility issues that cause test failures on modern systems like Fedora 43 with Python 3.14 and OpenSSL 3.x.

### Python 3.14 Compatibility Fixes
- Fix `kwargs.iteritems()` → `kwargs.items()` in `microserver.test.ext`
- Add missing `os` import in `microserver.test.ext` and `conditions.test.ext`
- Add explicit UTF-8 encoding to socket `decode()` calls
- Replace deprecated `socket.error` with `OSError` in `ports.py`
- Add socket timeout to prevent `recv()` from hanging indefinitely

### TLS Protocol Deprecation Fixes
- Add `HasLegacyTLSSupport()` condition to detect TLSv1.0/TLSv1.1 support
- Skip `tls_client_versions` tests on systems without legacy TLS support
- Modern OpenSSL 3.x + Fedora crypto-policies disable TLSv1.0/TLSv1.1 by default
- Detection uses actual TLS connection test to handle crypto-policy restrictions

### Timing Improvements
- Increase sleep duration in `polite_hook_wait.cc` from 200ms to 500ms to account for variable scheduling latency

## Test Plan

Tested on:
- Fedora 43 (Server Edition)
- Python 3.14.2
- GCC 15.2.1
- OpenSSL 3.x with DEFAULT crypto-policy

Results:
- TLS client_versions tests now properly skip instead of failing
- Core autest functionality works correctly
- No regressions observed

### DIFF
--- a/tests/gold_tests/autest-site/conditions.test.ext
+++ b/tests/gold_tests/autest-site/conditions.test.ext
@@ -16,6 +16,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import os
 import subprocess
 import json
 import re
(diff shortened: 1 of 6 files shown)


bryancall:
## Review Feedback Response

Addressed Copilot's comments in commit 20a8bafe4b.

### Fixed

| # | Comment | Fix |
|---|---------|-----|
| 1 | HasLegacyTLSSupport connects to www.google.com:443, depends on external network | Replaced with `127.0.0.1:1` (localhost, closed port). "Connection refused" = TLSv1 available, "no protocols available" = blocked by crypto-policy. No network dependency. |
| 2 | TimeoutExpired returns True (false positive) | Changed to return False -- timeout on localhost means something is wrong, not that TLSv1 works |
| 3 | SkipUnless in tls_client_versions.test.py passes invalid reason string | Removed the string arg. SkipUnless only accepts Condition objects; the reason is already in the Condition definition. |
| 4 | SkipUnless in tls_client_versions_minmax.test.py passes invalid reason string | Same fix as above |

### Not Fixed (with rationale)

| # | Comment | Reason |
|---|---------|--------|
| 1 | Only probes TLSv1.0, not TLSv1.1 | Crypto-policies (Fedora DEFAULT, RHEL DEFAULT, etc.) always disable TLSv1.0 and TLSv1.1 together. Testing one is sufficient. Added docstring explaining this rationale. |

bneradt (approved):

cmcfarlen:
Cherry-picked to 10.2.x

cmcfarlen:

PR #12860, opened 2026-02-05
Add: Server-side TLS handshake milestones and timing fixes

bryancall:
### Problem

Only client-side TLS handshake timing was available via `TS_MILESTONE_TLS_HANDSHAKE_START/END`. There was no way to measure origin-side TLS handshake overhead in the slow log or via the milestone API. Additionally, `SSLNetVConnection::sslClientHandShakeEvent()` was recording the TLS handshake start time for outbound connections but never calling `_record_tls_handshake_end_time()`, so the outbound TLS end time was always 0. The slow log's `difference_msec()` could also produce garbage values when only `ms_start` was unset (0) but `ms_end` was set.

### Changes

* **Add `TS_MILESTONE_SERVER_TLS_HANDSHAKE_START/END` milestones** -- new enum values in `apidefs.h.in` (ABI compatible, appended before `TS_MILESTONE_LAST_ENTRY`). Capture server TLS timing in `HttpSM::state_http_server_open()` for both the H2 multiplexed (`CONNECT_EVENT_TXN`) and H1 direct (`VC_EVENT_WRITE_COMPLETE`) connection paths. Update Lua plugin constants, `LogField.cc` milestone mappings, and API documentation.
* **Fix missing outbound TLS handshake end time** -- `sslClientHandShakeEvent()` (which handles the outbound/origin side despite the name) was recording start time but never calling `_record_tls_handshake_end_time()`. Add the missing call so `get_tls_handshake_end_time()` returns a valid value for HttpSM milestone capture.
* **Fix `difference_msec()` to check both start and end** -- previously only checked `ms_end == 0`. If `ms_start` was unset but `ms_end` was set, the subtraction produced a garbage value. Now checks both for 0 before computing the difference.
* **Rename slow log field `tls_handshake` to `ua_tls_handshake`** -- clarifies this is the client-side TLS handshake. Add `server_tls_handshake` field as an offset from `SM_START` (consistent with other server-side slow log fields). `slow_log_report.pl` accepts both old and new field names for backward compatibility.
* **Use pre-remap URL in slow log** -- changed from `client_request` (post-remap/origin-facing) to `unmapped_url` (pre-remap) so the log shows the URL the client actually requested. Falls back to `client_request` if `unmapped_url` is not available.

### Testing

- [x] Build passes (including ASAN build)
- [x] Manual testing with client-side and server-side TLS verified both milestone pairs populate correctly
- [x] `slow_log_report.pl` correctly parses both `ua_tls_handshake` and `server_tls_handshake`
- [x] Verified milestones remain 0 for non-TLS and connection-reuse scenarios

### DIFF
--- a/src/iocore/net/SSLNetVConnection.cc
+++ b/src/iocore/net/SSLNetVConnection.cc
@@ -1580,7 +1580,13 @@ SSLNetVConnection::sslClientHandShakeEvent(int &err)
 
     sslHandshakeStatus = SSLHandshakeStatus::SSL_HANDSHAKE_DONE;
 
-    // Record TLS handshake end time for outbound connections
+    // Record TLS handshake end time for outbound (origin) connections.
+    // Despite the name, sslClientHandShakeEvent() handles the *outbound* side
+    // where ATS acts as the TLS client connecting to the origin server.
+    // (The inbound/client-facing side is handled by sslServerHandShakeEvent(),
+    // where ATS acts as the TLS server.)
+    // The begin time is set when the handshake starts; we record the end time
+    // here so HttpSM can later copy both into milestones for logging.
     if (this->get_tls_handshake_begin_time()) {
       this->_record_tls_handshake_end_time();
     }
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -1882,12 +1882,15 @@ HttpSM::state_http_server_open(int event, void *data)
     do_http_server_open();
     break;
   case CONNECT_EVENT_TXN: {
-    // Multiplexed connection path (e.g. HTTP/2): ConnectingEntry created the session
-    // and dispatched this event. Get the netvc from the PoolableSession.
+    // Multiplexed connection path (H2/H3): ConnectingEntry owns the connection and
+    // dispatches this event with the PoolableSession once the handshake completes.
+    // The HttpSM has no access to the netvc during the handshake, so we copy the
+    // timestamps that were recorded in real-time by sslClientHandShakeEvent().
+    // See also VC_EVENT_WRITE_COMPLETE below for the direct (H1) path -- these
+    // two paths are mutually exclusive.
     SMDbg(dbg_ctl_http, "Connection handshake complete via CONNECT_EVENT_TXN");
     PoolableSession *session = static_cast<PoolableSession *>(data);
     ink_assert(session != nullptr);
-    // Capture server TLS handshake timing from the session's netvc
     if (auto *netvc = session->get_netvc()) {
       if (auto tbs = netvc->get_service<TLSBasicSupport>()) {
         milestones[TS_MILESTONE_SERVER_TLS_HANDSHAKE_START] = tbs->get_tls_handshake_begin_time();
@@ -1906,10 +1909,11 @@ HttpSM::state_http_server_open(int event, void *data)
   case VC_EVENT_READ_COMPLETE:
   case VC_EVENT_WRITE_READY:
   case VC_EVENT_WRITE_COMPLETE:
-    // Direct connection path (e.g. HTTP/1): HttpSM owns the netvc directly
-    // and creates the session itself.
+    // Direct connection path (H1): HttpSM owns the netvc directly.
+    // Same milestone copy as CONNECT_EVENT_TXN above -- these two paths are
+    // mutually exclusive. Timestamps were recorded in real-time by
+    // sslClientHandShakeEvent(); we copy them into milestones here.
     SMDbg(dbg_ctl_http_ss, "Connection handshake complete");
-    // Capture server TLS handshake timing if this is a TLS connection
     if (auto tbs = _netvc->get_service<TLSBasicSupport>()) {
       milestones[TS_MILESTONE_SERVER_TLS_HANDSHAKE_START] = tbs->get_tls_handshake_begin_time();
       milestones[TS_MILESTONE_SERVER_TLS_HANDSHAKE_END]   = tbs->get_tls_handshake_end_time();


src/proxy/http/HttpSM.cc
```diff
@@ -7834,6 +7853,7 @@ HttpSM::update_stats()
           milestones.difference_sec(TS_MILESTONE_SM_START, TS_MILESTONE_DNS_LOOKUP_END),
           milestones.difference_sec(TS_MILESTONE_SM_START, TS_MILESTONE_SERVER_CONNECT),
           milestones.difference_sec(TS_MILESTONE_SM_START, TS_MILESTONE_SERVER_CONNECT_END),
+          milestones.difference_sec(TS_MILESTONE_SM_START, TS_MILESTONE_SERVER_TLS_HANDSHAKE_END),
```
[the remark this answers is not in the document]
bryancall: Yes, I thought about that and the only reason the ua tls handshake is recorded that way is because it happens before the sm start.

The other thought I had was to use the ua tls handshake start as the "start" if it is set instead of using sm start.

src/iocore/net/SSLNetVConnection.cc
```diff
@@ -1579,6 +1579,12 @@ SSLNetVConnection::sslClientHandShakeEvent(int &err)
     Metrics::Counter::increment(ssl_rsb.total_success_handshake_count_out);
 
     sslHandshakeStatus = SSLHandshakeStatus::SSL_HANDSHAKE_DONE;
+
+    // Record TLS handshake end time for outbound connections
+    if (this->get_tls_handshake_begin_time()) {
+      this->_record_tls_handshake_end_time();
+    }
```
cmcfarlen, line 1592: Why was this added? This function seems to deal with the client side, but the comment mentions outbound connections.  And, if this were the outbound side, are we expecting the metrics this function updates to include both sides TLS?

src/proxy/http/HttpSM.cc
```diff
@@ -1881,21 +1881,39 @@ HttpSM::state_http_server_open(int event, void *data)
...
+    // and dispatched this event. Get the netvc from the PoolableSession.
     SMDbg(dbg_ctl_http, "Connection handshake complete via CONNECT_EVENT_TXN");
-    if (this->create_server_txn(static_cast<PoolableSession *>(data))) {
+    PoolableSession *session = static_cast<PoolableSession *>(data);
+    ink_assert(session != nullptr);
+    // Capture server TLS handshake timing from the session's netvc
+    if (auto *netvc = session->get_netvc()) {
+      if (auto tbs = netvc->get_service<TLSBasicSupport>()) {
+        milestones[TS_MILESTONE_SERVER_TLS_HANDSHAKE_START] = tbs->get_tls_handshake_begin_time();
+        milestones[TS_MILESTONE_SERVER_TLS_HANDSHAKE_END]   = tbs->get_tls_handshake_end_time();
```
cmcfarlen, line 1897: This sets both milestones after the end time is known.  Should we set the start milestone when it starts vs here? As it is, if this code isn't reached, the start milestone won't be set.

src/proxy/http/HttpSM.cc
```diff
@@ -1881,21 +1881,39 @@ HttpSM::state_http_server_open(int event, void *data)
...
   case VC_EVENT_WRITE_COMPLETE:
-    // Update the time out to the regular connection timeout.
+    // Direct connection path (e.g. HTTP/1): HttpSM owns the netvc directly
+    // and creates the session itself.
     SMDbg(dbg_ctl_http_ss, "Connection handshake complete");
+    // Capture server TLS handshake timing if this is a TLS connection
+    if (auto tbs = _netvc->get_service<TLSBasicSupport>()) {
+      milestones[TS_MILESTONE_SERVER_TLS_HANDSHAKE_START] = tbs->get_tls_handshake_begin_time();
+      milestones[TS_MILESTONE_SERVER_TLS_HANDSHAKE_END]   = tbs->get_tls_handshake_end_time();
+    }
```
cmcfarlen, line 1920: Why are these set again?

src/iocore/net/SSLNetVConnection.cc
bryancall: The naming in this file is from the TLS role perspective, not the ATS role. `sslClientHandShakeEvent()` handles the **outbound** side where ATS acts as the TLS *client* connecting to the origin. `sslServerHandShakeEvent()` handles the **inbound** side where ATS acts as the TLS *server* accepting client connections.

The inbound side already records the end time (line 1383), but the outbound side was missing it — this is a bug fix. Without this, `get_tls_handshake_end_time()` returns zero when HttpSM tries to copy the timestamps into milestones.

I've updated the comment to clarify the naming convention and the two-stage pattern.

src/proxy/http/HttpSM.cc
bryancall: The HttpSM doesn't have access to the netvc during the handshake in the multiplexed path — ConnectingEntry owns the connection and dispatches `CONNECT_EVENT_TXN` after the handshake completes. So this is the earliest point the SM can grab the timestamps.

The actual start/end times are captured in real-time at the netvc layer (`sslClientHandShakeEvent()` records the end time, and the begin time is set when the handshake starts). We're just copying them into milestones here. If this code isn't reached (e.g. connection failure), the milestones stay at zero, which is correct — `difference_msec()` checks for zero before computing the difference.

Updated the comments to make this clearer.

src/proxy/http/HttpSM.cc
bryancall: These aren't duplicates — they're two mutually exclusive connection paths. `CONNECT_EVENT_TXN` fires for multiplexed connections (H2/H3) where ConnectingEntry manages the connection. `VC_EVENT_WRITE_COMPLETE` fires for direct connections (H1) where the SM owns the netvc. A given connection goes through one path or the other, never both.

Updated both comments to cross-reference each other and explicitly state they're mutually exclusive.

bryancall:
[approve ci autest 1]

cmcfarlen:
Cherry-picked to 10.2.x

PR #12861, opened 2026-02-05
proxy.process.http.000_responses metric

bneradt:
Add a new HTTP stat to track responses where no valid HTTP status code was sent to the client. This typically occurs when the client aborts the connection before a response is sent (ERR_CLIENT_ABORT).


cmcfarlen:
Cherry-picked to 10.2.x

PR #12863, opened 2026-02-06
Fix ppa log field

maskit:
`ppa` is a field for PP2_TYPE_AUTHORITY.

- Copy a whole pp_info for logging
- Calculate the length with alignment in consideration
- Put null terminator at the end of the log field value


cmcfarlen:

PR #12866, opened 2026-02-06
Fix failed assertion in build_response for 1xx races

bneradt:
This failed assertion was seen sometimes when running the cache-tests.

When a 1xx informational response (such as 103 Early Hints) is forwarded to the client, its headers are copied into client_response and sent via an asynchronous tunnel. The tunnel completion handler (tunnel_handler_100_continue) destroys client_response when the tunnel finishes. However, if the final server response arrives and build_response is called before the tunnel completes, client_response is still valid, triggering the assertion `!new_hdr->valid()` in copy_header_fields.

This patch addresses this by defensively destroying outgoing_response in build_response if it is already valid before calling copy_header_fields. The header data from the 1xx has already been serialized into the tunnel buffer and is no longer needed, so this should be safe.


cmcfarlen:
Cherry-picked to 10.2.x

PR #12868, opened 2026-02-09
add host_override to parent.config and other sni name fixes

traeak:
Fixes to sni name handling when using another CDN as parent.

This adds "host_override" to parent.config and has some fixes for the strategies handling of same option.

### DIFF
--- a/tests/gold_tests/tls/tls_sni_parent_failover.test.py
+++ b/tests/gold_tests/tls/tls_sni_parent_failover.test.py
@@ -88,10 +88,13 @@ dns.addRecords(records={"bar.com.": ["127.0.0.1"]})
 dns.addRecords(records={"parent.": ["127.0.0.1"]})
 dns.addRecords(records={"strategy.": ["127.0.0.1"]})
 
-ts.Disk.remap_config.AddLines([
-    "map http://parent https://parent",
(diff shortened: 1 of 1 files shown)


serrislew:
Some docs on the new feature would be helpful!

doc/admin-guide/files/parent.config.en.rst
```diff
@@ -324,6 +324,14 @@ The following list shows the possible actions and their allowed values.
 
   -  ``false`` - The default.  Do not ignore the host status.
 
+.. _parent-config-format-host_override:
+
+``host_override``
+    One of the following values:
+
+    -  ``true`` - Sets the host header of the request using the selected parent, including the SNI name to be used for any upstream TLS connection.  Useful when a parent is another CDN that requires a correct SNI name.
```
serrislew, line 332: If`pristine_host_hdr` is set to 1, does this have precedence over it?

traeak: the pristine host header looks like it applies during remap and cache lookup but not after, which is where the host_override kicks in. I'll look into it more closely.

traeak:
[approve ci autest]

cmcfarlen:
Cherry-picked to 10.2.x

PR #12869, opened 2026-02-09
Doc: mention elif, nested conditionals

mlibbey:
Missed some documentation when adding elif and nested conditionals.

### DIFF
--- a/doc/admin-guide/configuration/hrw4u.en.rst
+++ b/doc/admin-guide/configuration/hrw4u.en.rst
@@ -93,11 +93,12 @@ provides an LSP for this language, named ``hrw4u-lsp``.
 Syntax Differences
 ==================
 
-The basic structure is a `section` name defining the part of the transaction to run in
-followed by conditionals and operators. It uses `if () {} else {}` conditional syntax
-with `&& , || and ==`, with conditions and operators generally following function() or
-object.style grammar. Operator lines are terminated with `;` (whitespace is still not
-significant). For instance:
+The basic structure consists of a `section` name defining the part of the transaction to run,
+followed by conditionals and operators. It uses `if ... {...} elif ... {...} else {...}` conditional syntax
+with `&& , || and ==`. Conditionals support parenthesized grouping for complex expressions (`if (...) {...}`)
+and can be nested up to a maximum depth of 10 levels. Conditions and operators generally follow
+function() or object.style grammar. Operator lines are terminated with `;` and whitespace is still
+insignificant. For instance:
 
 .. code-block:: none
 
@@ -107,10 +108,28 @@ significant). For instance:
    }
 
    REMAP {
-     if inbound.status == 403 || access("/etc/lock") {
+     if inbound.status == 403 {
        inbound.req.X-Fail = "1";
+     } elif inbound.status == 404 {
+       inbound.req.X-NotFound = "1";
      } else {
        no-op();
+     }
+   }
+
+   READ_RESPONSE {
+     if inbound.status > 399 {
+       if inbound.status < 500 {
+         # Client error - check specific cases
+         if inbound.status == 404 {
+           counter("errors.not_found");
+         } elif inbound.status == 403 {
+           counter("errors.forbidden");
+         }
+       } else {
+         # Server error
+         counter("errors.server");
+       }
      }
    }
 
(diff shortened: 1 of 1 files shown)


bneradt (dismissed):
I verified the examples from this PR by building and running the `hrw4u` parser (via `tools/hrw4u`). Two issues found:

## 1. `>=` is not a valid operator in the grammar

The new `READ_RESPONSE` nested conditional example uses `inbound.status >= 400`, but `>=` is not a supported comparison operator. The grammar (`tools/hrw4u/grammar/hrw4u.g4`) only defines `GT: '>'` and `LT: '<'` — there are no `GTE` or `LTE` tokens. The comparison rule only supports `==`, `!=`, `>`, `<`, `~`, `!~`, and `in` (which is also reflected in the doc's own "Condition operators" table).

Running the parser on the example confirms the error:

```
/tmp/test_nested.hrw4u:2:21: error: extraneous input '=' expecting {TRUE, FALSE, STRING, IPV4_LITERAL, IPV6_LITERAL, IDENT, NUMBER, '{'}
   2 |   if inbound.status >= 400 {
     |                      ^
```

This should be `inbound.status > 399`, consistent with the existing "Normalize Statuses" example already in the doc which uses `inbound.status > 399 && inbound.status < 500`.

## 2. Comparison table entry for elif is misleading

The PR changes the comparison table to:

```
Else branches        `else` with indented ops      `elif { ... }` and `else { ... }` with full blocks
```

`elif { ... }` omits the required condition, making it look like `elif` takes a bare block. Per the grammar, `elif` requires a condition (e.g., `elif <cond> { ... }` or `elif (<cond>) { ... }`). This is also inconsistent with the syntax summary earlier in the same PR which correctly shows `if () {} elif () {} else {}`. Consider something like `` `elif (...) { ... }` and `else { ... }` with full blocks ``.

---

Everything else looks good:
- The REMAP `if/elif/else` example parses cleanly and produces correct `header_rewrite` output.
- The nesting depth of 10 is accurate (`MAX_IF_NESTING_DEPTH = 10` in `plugins/header_rewrite/header_rewrite.cc`).
- The `counter()` calls in the nested example are valid.
- The structural nesting of conditionals is correct per the grammar (since `blockItem` can contain `conditional`).

doc/admin-guide/configuration/hrw4u.en.rst
```diff
@@ -107,13 +107,31 @@ significant). For instance:
...
        inbound.req.X-Fail = "1";
+     } elif inbound.status == 404 {
+       inbound.req.X-NotFound = "1";
      } else {
        no-op();
      }
    }
 
+   READ_RESPONSE {
+     if inbound.status >= 400 {
```
bneradt: If you look at the `tools/hrw4u/grammar/hrw4u.g4`, `>=` is not actually supported. Maybe we want to add this, but I don't think this will work as is.

@zwoop : thoughts?

doc/admin-guide/configuration/hrw4u.en.rst
```diff
@@ -158,7 +176,7 @@ Syntax               Free-form                     Structured `if (...) { ... }`
 Conditions           `cond %{...}`                 Implicit in `if (...)`
 Operators            Raw text (`set-header`)       Structured assignments or statements
 Grouping             `GROUP` / `GROUP:END`         Use `()` inside `if` expressions
-Else branches        `else` with indented ops      `else { ... }` with full block
+Else branches        `else` with indented ops      `elif { ... }` and `else { ... }` with full blocks
```
bneradt: Probably good to `elif () { ... }`, with the `()` as you did above to indicate a condition should be present with `elif`. `else` is fine as is since it doesn't take a condition.

doc/admin-guide/configuration/hrw4u.en.rst
mlibbey: eek and oops! Thanks for catch!

zwoop: yeah, >= is not supported, because header_rewrite doesn't support it :) You'll have to change it to a ! <

doc/admin-guide/configuration/hrw4u.en.rst
zwoop: no, please avoid unnecessary ()'s, because they will always add a %{GROUP}. hrw4u does not require any () for simple expressions.

mlibbey: I changed to `elif ... { ... }` ... hoping that the `elif ... ` implies that you need a conditional there (and that actual "()" are not required.

bneradt: Cool. Thanks @zwoop / @mlibbey

cmcfarlen:
Cherry-picked to 10.2.x

PR #12874, opened 2026-02-10
NetAcceptAction::cancel() use-after-free fix: part 2

bneradt:
This is a revised version of PR #12803 which was reverted in PR #12841 because it lacked the idempotent cancel guard, causing its own set of crashes.

There is a race between NetAcceptAction::cancel() and NetAccept::acceptEvent() where the server pointer could be dereferenced after the NetAccept object was deleted. Thread A calls cancel(), which sets cancelled=true via Action::cancel(). Thread B running acceptEvent() sees cancelled==true and deletes the NetAccept (including the embedded Server). Thread A then calls server->close() on freed memory.

This is fixed by making the server pointer in NetAcceptAction atomic and using exchange(nullptr) so only one thread can obtain and close the server. Additionally, cancel() is made idempotent by checking !cancelled before calling Action::cancel(), which prevents ink_assert(!cancelled) assertion failures when cancel is called from both external callers (TSActionCancel) and internal error paths (acceptEvent, acceptFastEvent, acceptLoopEvent).

### DIFF
--- a/src/iocore/net/P_NetAccept.h
+++ b/src/iocore/net/P_NetAccept.h
@@ -43,6 +43,7 @@
 #include "iocore/net/NetAcceptEventIO.h"
 #include "Server.h"
 
+#include <atomic>
 #include <vector>
 
 struct NetAccept;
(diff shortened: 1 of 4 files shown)


src/iocore/net/P_NetAccept.h
```diff
@@ -60,21 +61,30 @@ AcceptFunction net_accept;
...
-    return Action::operator=(acont);
+    // Use atomic exchange so only one thread closes the server, preventing
+    // use-after-free races between cancel() and acceptEvent() cleanup.
+    Server *s = server.exchange(nullptr, std::memory_order_acq_rel);
+    if (s != nullptr) {
+      s->close();
+    }
+    if (!cancelled) {
+      Action::cancel(cont);
+    }
```
bneradt, line 87: This `!canceled` check is the only change made on top of the reverted #12803 PR. That should hopefully address the double-cancel issue (in addition to addressing the use after free issue, which the rest of the patch intends to fix).

bneradt:
[approve ci autest 0 rocky]

bneradt:
[approve ci rocky]

bneradt:
[approve ci autest rocky]

cmcfarlen:
Cherry-picked to 10.2.x

PR #12877, opened 2026-02-11
parallel autests: print failure output

bneradt:
The parallel runner captured full autest output per worker but only displayed failed test names in the summary. This made CI failures hard to debug compared to the sequential runner. Extract and print the per-test failure detail block (sub-step pass/fail, reasons, file paths) so the same diagnostic information is available in parallel mode.


cmcfarlen:
Cherry-picked to 10.2.x

PR #12878, opened 2026-02-11
Add proxy.process.http.429_responses metric

bneradt:
Add a new HTTP stat to track 429 (Too Many Requests) responses sent to clients. This is useful for monitoring rate-limiting behavior, as the rate_limit plugin defaults to returning 429 when limits are exceeded. The metric follows the same pattern as existing per-status-code counters (400-416) across HttpConfig, HttpTransact, traffic_top, and traffic_logstats.

Also enhance the ATSReplayTest framework with a metric_checks feature that allows replay YAML files to declare expected metric values. After traffic completes, follow-up test runs automatically verify the metrics via traffic_ctl. This required creating the ATS process on the Test object (rather than a TestRun) when post-traffic verification is needed, so that ATS persists across test runs.


bneradt:
[approve ci autest 0 1]

bryancall (approved):

bryancall:
Found one correctness issue in the new metric verification helper:

`Testers.ContainsExpression(f'{metric_name} {expected_value}', ...)` can false-pass when the expected value is a prefix of the actual value (for example expected `1` also matches `10`).

Would you mind switching this to an anchored regex match so the assertion is exact? Something like `^<metric_name>\\s+<expected_value>$` (with `metric_name` escaped) avoids prefix matches and makes the check robust.

bneradt:
> Found one correctness issue in the new metric verification helper:
> 
> `Testers.ContainsExpression(f'{metric_name} {expected_value}', ...)` can false-pass when the expected value is a prefix of the actual value (for example expected `1` also matches `10`).
> 
> Would you mind switching this to an anchored regex match so the assertion is exact? Something like `^<metric_name>\\s+<expected_value>$` (with `metric_name` escaped) avoids prefix matches and makes the check robust.

Thanks for the feedback. Update done.

cmcfarlen:
Cherry-picked to 10.2.x

PR #12880, opened 2026-02-12
Check state of HostDBInfo

masaori335:
`HostDBRecord::select_best_http()` function is expected return `nullptr` is all `HostDBInfo` are down state. However, a `HostDBRecord` has only one `HostDBInfo` (`info.count() == 1`), state of the info is not checked.

### DIFF
--- a/src/iocore/hostdb/HostDB.cc
+++ b/src/iocore/hostdb/HostDB.cc
@@ -1324,7 +1324,9 @@ HostDBRecord::select_best_http(ts_time now, ts_seconds fail_window, sockaddr con
       }
     }
   } else {
-    best_alive = &info[0];
+    if (!info[0].is_down(now, fail_window)) {
+      best_alive = &info[0];
+    }
   }
 
   return best_alive;
--- a/src/iocore/hostdb/HostDB.cc
+++ b/src/iocore/hostdb/HostDB.cc
@@ -1324,7 +1324,7 @@ HostDBRecord::select_best_http(ts_time now, ts_seconds fail_window, sockaddr con
       }
     }
   } else {
-    if (!info[0].is_down(now, fail_window)) {
+    if (info[0].select(now, fail_window)) {
       best_alive = &info[0];
     }
   }
--- a/tests/gold_tests/dns/dns_host_down.test.py
+++ b/tests/gold_tests/dns/dns_host_down.test.py
@@ -55,24 +55,13 @@ class DownCachedOriginServerTest:
                 'proxy.config.hostdb.host_file.path': os.path.join(Test.TestDirectory, "hosts_file"),
             })
 
-    # Even when the origin server is down, SM will return a hit-fresh domain from HostDB.
-    # After request has failed, SM should mark the IP as down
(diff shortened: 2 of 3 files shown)


masaori335:
`Test: dns_host_down: Failed` seems related to this change. I'll check it in local.

masaori335:
The `dns_host_down` AuTest is adjusted.

masaori335:
[approve ci autest 2of4]

src/iocore/hostdb/HostDB.cc
```diff
@@ -1324,7 +1324,9 @@ HostDBRecord::select_best_http(ts_time now, ts_seconds fail_window, sockaddr con
       }
     }
   } else {
-    best_alive = &info[0];
+    if (!info[0].is_down(now, fail_window)) {
```
serrislew: do we even need this? if the above conditional starts at the current target and can't find a valid one, it will return the same thing?

masaori335: I wondered the same things at first. However, it comes here and call `select()` to before setting the `best_alive`. Which means if all HostDBInfo(s) are down state, this function returns `nullptr`. 

https://github.com/apache/trafficserver/blob/0fb268b07e2b2e080af335023aade14b9c99ebc8/src/iocore/hostdb/HostDB.cc#L1314-L1325

https://github.com/apache/trafficserver/blob/0fb268b07e2b2e080af335023aade14b9c99ebc8/include/iocore/hostdb/HostDBProcessor.h#L282-L291

serrislew: But I thought we want it return nullptr if `info[0]` is down (all options are down)? `select` is the same as `is_down` except it checks the down window

masaori335: This change makes this function returning `nullptr` if `info[0]` is down. 

I should use `select()` instead of `! is_down()` for consistency?

bneradt:
[approve ci autest]

serrislew (approved):

masaori335:
I just noticed https://github.com/apache/trafficserver/pull/11814 intentionally restored the ATS9 behavior for 10.0.x.

> If a HostDB record only has one IP, always select that for use/retry regardless of fail status (This restores the ATS9 behavior)

We need to figure out where this should go.

masaori335:
Talked with @cmcfarlen, it seems good to ship this by 10.2.0 as a bug fix. 

If anybody wants to try connect attempt against down servers, below changes are recommended

- increase [`proxy.config.http.connect_attempts_max_retries_down_server`](https://docs.trafficserver.apache.org/admin-guide/files/records.yaml.en.html#proxy-config-http-connect-attempts-max-retries-down-server) (Default: 1)
- shorten [`proxy.config.http.down_server.cache_time`](https://docs.trafficserver.apache.org/admin-guide/files/records.yaml.en.html#proxy-config-http-down-server-cache-time) (Default: 60)

cmcfarlen:
Cherry-picked to 10.2.x

PR #12884, opened 2026-02-13
Fix log_retention test: use correct test run variable

brbzull0:
This was failing on my box. 

Test 7 ('Get the log to rotate.') was calling MakeCurlCommandMulti on test.tr (the test run from the TestLogRetention constructor) instead of the newly created tr. This left the new test run with no process defined, causing the test to fail with 'List came back empty.'

### DIFF
--- a/tests/gold_tests/logging/log_retention.test.py
+++ b/tests/gold_tests/logging/log_retention.test.py
@@ -488,7 +488,7 @@ tr.StillRunningAfter = test.ts
 tr.StillRunningAfter = test.server
 
 tr = Test.AddTestRun("Get the log to rotate.")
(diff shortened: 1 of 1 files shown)


brbzull0:
@bneradt:  This seems to be ok now , not sure if we haven't detect the issue because of the initial skip, but anyways. Works for me now.

bneradt (approved):

bneradt:
> @bneradt: This seems to be ok now , not sure if we haven't detect the issue because of the initial skip, but anyways. Works for me now.

Yep. This test is skipped in CI because it has timing issues. 

Thanks for the fix!

PR #12886, opened 2026-02-13
Stabilize post and slow_post autests under parallel runs

bryancall:
### Problem

The `post-early-return` and `slow_post/server_abort` autests fail intermittently under parallel execution on Fedora 43. The `nc -l` based mock server only accepts a single TCP connection, which gets consumed by the readiness probe, leaving nothing for ATS to connect to.

### Changes

* **Stabilize `thread_config` startup ordering** -- gate each `check_threads` invocation on ATS port readiness using `When.PortOpen(ts.Variables.port)` to avoid races where validation runs before `traffic_server` is ready.
* **Stabilize `post` and `slow_post` autests** -- add explicit origin readiness checks in `post-early-return` and accept the current microserver exception variant in `server_abort` for Fedora 43 compatibility.
* **Revert `thread_config` PortOpen change per review** -- keep ATS startup gating on existing fully-initialized readiness behavior. Move `post-early-return` upstream readiness to server process `Ready` fields instead of `StartBefore` calls.
* **Replace `nc`-based server with `mock_origin.py`** -- `server1.sh` only accepts a single TCP connection, consumed by the `When.PortOpen()` readiness probe. `mock_origin.py` absorbs probes, sends the configured HTTP response, and drains request data before closing to prevent TCP RST that causes 502 on H2 streams.
* **Remove unused `server1.sh`** -- replaced by `mock_origin.py`.

### Testing

- [x] 5/5 parallel ASAN runs on hera2
- [x] All 15 CI platforms green


tests/gold_tests/post/post-early-return.test.py
```diff
@@ -83,23 +83,23 @@
 # The third case has an explicit multi-second sleep which ensures the early response path is exercised
 test_run = Test.AddTestRun("http1.1 Post with small body early return")
 test_run.Processes.Default.StartBefore(Test.Processes.ts)
-test_run.Processes.Default.StartBefore(server1)
+test_run.Processes.Default.StartBefore(server1, ready=When.PortOpen(Test.Variables.upstream_port1))
```
bneradt: Rather than associate the ready with StartBefore, I think it would be preferable to associate it with the server1, server2, etc., processes themselves. That way the Ready condition is more tightly tied to the process rather than a function call associated with the process. Thus:

```python
server1.Ready = When.PortOpen(Test.Variables.upstream_port1)
```

This is how things are done, for instance, for the MakeATSProcess Processes in trafficserver.test.ext.

tests/gold_tests/thread_config/thread_config.test.py
```diff
@@ -39,7 +39,7 @@
 TS_ROOT = ts.Env['TS_ROOT']
 tr.Processes.Default.Command = f'{sys.executable} check_threads.py -p {TS_ROOT} -e 1 -a 0 -t 1 -c 1'
 tr.Processes.Default.ReturnCode = 0
-tr.Processes.Default.StartBefore(ts)
+tr.Processes.Default.StartBefore(ts, ready=When.PortOpen(ts.Variables.port))
```
bneradt: This seems incorrect. ATS is not ready when the port is open...ATS opens the port and then does a bunch of work afterwards before it is ready. Waiting on PortOpen used to cause issues in the past. This PortOpen ready condition was intentionally changed to wait upon the log message which gives a more reliable marker for ATS being ready. This is why we have the ready condition tied to the log output in trafficserver.text.ext:

```python
    p.Ready = When.FileContains(p.Disk.diags_log.AbsPath, "NOTE: Traffic Server is fully initialized")
```

Am I missing something? If I am, please explain carefully why awaiting on PortOpen is better than waiting upon the "Traffic Server is fully initialized" message.

bneradt (changes requested):

tests/gold_tests/post/post-early-return.test.py
bryancall: Great suggestion, thanks. I updated this to set readiness on the upstream process objects directly (`serverN.Ready = When.PortOpen(...)`) and switched the call sites back to plain `StartBefore(serverN)`.

tests/gold_tests/thread_config/thread_config.test.py
bryancall: You were right, and thanks for calling this out. I removed the `thread_config` PortOpen-based readiness change from this PR so ATS startup gating remains tied to the existing fully-initialized readiness behavior. I also retitled/scoped this PR to only the `post`/`slow_post` autest fixes.

bneradt (dismissed):

maskit:
[approve ci autest 1]

bryancall:
The consistent AuTest 1of4 failure was caused by a fundamental incompatibility between `nc -l` and `When.PortOpen()`:

`nc -l` only accepts a **single TCP connection**. The `PortOpen` readiness probe connects to check the port is open, consuming that one connection. When ATS then tries to proxy the request to the origin, the port is dead → 502 every time. This is why it failed deterministically on all 4 CI retries.

**The fix** (pushed in `3cf53d3`) replaces `server1.sh` with a new shared `tests/tools/mock_origin.py` that:

1. Uses Python `socket.listen()` which handles multiple `accept()` calls, so readiness probes are absorbed silently
2. Sends the configured HTTP response after reading request headers
3. **Drains remaining request data** before closing — without this, closing the connection mid-POST causes a TCP RST that makes ATS return 502 on HTTP/2 streams (this was a second issue discovered during testing)

`mock_origin.py` is a reusable tool that can replace all the ad-hoc `nc -l` server scripts in the test suite (`post/server1.sh`, `chunked_encoding/server2-4.sh`, `post_slow_server/server.sh`). It supports configurable status codes, response bodies (content-length or chunked), and response delays via CLI args.

Tested **5/5 parallel ASAN runs** passing on a local dev machine. The previous `mock_origin.py` without the drain fix was failing ~2/3 parallel runs on the HTTP/2 test case.

maskit:
> Fatal: ../src/api/InkAPITest.cc:916: failed assertion `(event == TS_EVENT_NET_ACCEPT) || (event == TS_EVENT_NET_ACCEPT_FAILED)`

I see this error on my PR as well.

maskit:
Can we remove server1.sh ?

bryancall:
Yes, it can be removed. I will push a new commit to delete it.

bneradt (approved):

PR #12887, opened 2026-02-13
Check Curl support for TLS 1.0 and 1.1 for autest

bryancall:
## Summary
- add `Condition.HasCurlTLSVersionSupport(...)` to probe curl TLS version
  capability directly
- gate TLSv1/TLSv1.1 subtests in `tls_client_versions` and
  `tls_client_versions_minmax` based on curl capability
- avoid false failures when OpenSSL policy checks pass but curl cannot
  actually attempt legacy TLS handshakes

## Test plan
- [x] run style checks via pre-commit hooks during commit
- [x] run local curl probes for TLSv1/TLSv1.1 capability against localhost
- [x] verify only autest condition/test files changed (no production code)

### DIFF
--- a/tests/gold_tests/autest-site/conditions.test.ext
+++ b/tests/gold_tests/autest-site/conditions.test.ext
@@ -146,6 +146,67 @@ def HasCurlOption(self, option):
     return self.CheckOutput(['curl', '--help', 'all'], default, "Curl needs to support option: {option}".format(option=option))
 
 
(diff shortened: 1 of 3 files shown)


maskit (approved):
This addition made me wonder what the difference from `HasLegacyTLSSupport` is but I figured it out. We might want to rename it to `HasOpenSSLLegacyTLSSupport`.

cmcfarlen:

PR #12888, opened 2026-02-14
Discourge the use of log type dINT

maskit:
We have two integer log type, sINT (single int) and dINT (double int). Those might have had different sizes in 32 bit era, but it seems like everything is 64 bit integer now. I don't see any reasons to use dINT, and the existence is just confusing.

We might want a bigger integer type in the future, but I'd suggest explicit name such like INT128 instead of "double" integer.

The enum entry is kept for compatibility.


maskit:
Autest h2origin failed.

maskit:
[approve ci autest 0]

maskit:
[approve ci autest 0]

maskit:
It turned out that some fields rely on `dINT` incorrectly. Those need to be fixed first.

maskit:
It doesn't seem like we can remove `dINT` soon

Issue #12324, opened 2025-06-30
Improving `traffic_ctl config reload`


cmcfarlen:
I would like to propose a new way for ATS to handle `config reload` that will hopefully provide more feedback and information to the user.

Current issues:
- `traffic_ctl config reload` asks `traffic_server` to reload its config.  The config reload command first enumerates the files it has watches on to check for changes and then it signals subsystems that have changes that they should reconfigure itself.  The file checks happen while the traffic_ctl client is waiting and if that process is slow, could result in a timeout which gives ambiguous results.  When the timeout happens, the configuration still proceeds.
- There isn't any structured way to get status or results of a reload.  Tooling is left to parse log files if they need to determine when and if the configuration finished successfully.

New proposal:
- Initiating a reconfigure traffic_ctl config reload should return status immediately without doing more work than necessary to either start a reconfigure or fail (if a reconfig is already running).  The status result should indicate if a reconfigure was started, and return a token representing the current reconfigure operation. (i.e either return the token for the ongoing config, or the token representing the new reconfigure if one is not running currently).  All other work should be done in a background continuation (including spinning through the changed files, even records.yaml).
- ATS should keep track of the information regarding the outcome and details of the last reconfigure.  This result should be viewable via structured data (json) from a traffic_ctl command.  The information should list the files changed and the result of whatever reconfigure operation resulted from that change (the callback).  The information should also indicate the status of the reconfigure op (success, fail or ongoing) and details about records.yaml validation. Bonus points if there is some indication of progress of incomplete operation (i.e. “10 of 25 files reconfigured”).
- A client that requests a reconfigure can use the returned token and then poll `traffic_ctl config status`to positively determine when the reconfigure operation is complete.

This change in behavior could be made transparent to traffic_ctl config reload if we adjust the client side of that command to issue the start reconfigure and then poll for status.  This way the config reload command could provide progress if requested or at least be able to give better information should there be issues with the configuration.  (i.e. it could wait until the configuration is started or wait until it is completely finished before returning status)

See the sequence diagram below that shows the new proposal.  Consider a totally hypothetical mechanism where a virtual filesystem is used to access secrets(certificates) that are stored off host and accessed over a network.  This new mechanism will not require the traffic_ctl client to wait on a slow network and could also return progress information to the client.

![Image](https://github.com/user-attachments/assets/fab7e5af-7956-403f-befa-887a74637319)

brbzull0:
I ~~may be~~  am working on this as soon as I have some time.

[...]

Issue #11724, opened 2024-08-20
SSLClientCoordinator: SSL config is reloaded multiple times


cmcfarlen:
If multiple ssl related config files are changed and then `traffic_ctl config reload` is executed, the ssl config will reload multiple times.

Reproduction:

1. Run traffic_server with a sni.yaml and ssl_multicert.config file configured
2. touch sni.yaml
3. `traffic_ctl config reload`
4. Observerve the log output:
```
[Aug 20 13:59:35.145] [ET_TASK 0] NOTE: /opt/ats/etc/trafficserver/sni.yaml loading ...
[Aug 20 13:59:35.147] [ET_TASK 0] NOTE: /opt/ats/etc/trafficserver/sni.yaml finished loading
[Aug 20 13:59:35.147] [ET_TASK 0] NOTE: ssl_multicert.config loading ...
[Aug 20 13:59:35.158] [ET_TASK 0] NOTE: /opt/ats/etc/trafficserver/ssl_multicert.config finished loading
```
5. touch sni.yaml and ssl_multicert.config files
6. `traffic_ctl config reload`
7. Observer the log output:
```
[Aug 20 13:59:53.154] [ET_TASK 0] NOTE: /opt/ats/etc/trafficserver/sni.yaml loading ...
[Aug 20 13:59:53.156] [ET_TASK 0] NOTE: /opt/ats/etc/trafficserver/sni.yaml finished loading
[Aug 20 13:59:53.156] [ET_TASK 0] NOTE: ssl_multicert.config loading ...
[Aug 20 13:59:53.166] [ET_TASK 0] NOTE: /opt/ats/etc/trafficserver/ssl_multicert.config finished loading
[Aug 20 13:59:53.166] [ET_TASK 0] NOTE: /opt/ats/etc/trafficserver/sni.yaml loading ...
[Aug 20 13:59:53.167] [ET_TASK 0] NOTE: /opt/ats/etc/trafficserver/sni.yaml finished loading
[Aug 20 13:59:53.168] [ET_TASK 0] NOTE: ssl_multicert.config loading ...
[Aug 20 13:59:53.174] [ET_TASK 0] NOTE: /opt/ats/etc/trafficserver/ssl_multicert.config finished loading
```

This happens because the callbacks setup in SSLClientCoordinator are independently attached and so the reconfigure function will be called for every updated config item specified.

This happens on the master branch, but also for 9.2.x

JosiahWI:
[...]

PR #12892, opened 2026-02-18
ATS Configuration Reload with observability/tracing - Token model

brbzull0:
## New Configuration Reload Framework

**Contents:**
- [TL;DR](#tldr)
- [How Reload Works — The Token Model](#how-reload-works--the-token-model)
- [New `traffic_ctl` Commands](#new-traffic_ctl-commands)
- [New JSONRPC APIs](#new-jsonrpc-apis)
- [Background: Issues with the Previous Reload Mechanism](#background-issues-with-the-previous-reload-mechanism)
- [What the New Design Solves](#what-the-new-design-solves)
- [Basic Design](#basic-design)
- [How Handlers Work](#how-handlers-work)
- [Design Challenges](#design-challenges)
  - [1. Handlers must reach a terminal state — or the task hangs](#1-handlers-must-reach-a-terminal-state--or-the-task-hangs)
  - [2. Parent status aggregation from sub-tasks](#2-parent-status-aggregation-from-sub-tasks)
  - [3. Startup vs. reload — same handler, different context](#3-startup-vs-reload--same-handler-different-context)
  - [4. Duplicate handler execution for multi-record triggers (fixed)](#4-duplicate-handler-execution-for-multi-record-triggers-fixed)
  - [5. Plugin support](#5-plugin-support)
- [Configs Migrated to `ConfigRegistry`](#configs-migrated-to-configregistry)
- [New Configuration Records](#new-configuration-records)
- [TODO](#todo)
  - [Future Work (separate PRs)](#future-work-separate-prs)
- [Dependencies and Related Issues](#dependencies-and-related-issues)

---

### TL;DR

> __NOTE__: **Backward compatible:** The existing `traffic_ctl config reload` command works exactly as before —
> same syntax, same behavior from the operator's perspective. Internally, it now fires the new reload
> logic, which means every reload is automatically tracked, timed, and queryable.

### How Reload Works — The Token Model

When you run `traffic_ctl config reload`, the command sends a JSONRPC request to ATS and
**returns immediately** — it does not block until every config handler finishes. Instead, ATS:

1. **Assigns a token** to the reload operation. The token is either auto-generated
   (e.g., `rldtk-1739808000000`, a timestamp-based ID) or user-supplied via `-t`
   (e.g., `-t deploy-v2.1`).
2. **Schedules the actual reload** on background threads (`ET_TASK`). Each registered config
   handler runs, reports its status (`in_progress` → `success` or `fail`), and the results are
   aggregated into a task tree.
3. **Returns the token** to the caller so you can track the reload afterwards.

The **token** is the handle for everything that follows:

| What you want | Command |
|---|---|
| Monitor progress live | `traffic_ctl config reload -t <token> -m` |
| Query final status | `traffic_ctl config status -t <token>` |
| Get detailed logs | `traffic_ctl config status -t <token> -l` |

If you don't supply `-t`, ATS generates a token automatically and prints it so you can copy-paste
it into follow-up commands. If you supply `-t deploy-v2.1`, that exact string becomes the token —
useful for CI pipelines, deploy scripts, or any workflow where you want a meaningful, predictable
identifier.

**In short:** the token is a reload's unique ID. You get it when the reload starts, and you use it
to ask "what happened?" at any point afterwards.

**New `traffic_ctl` commands:**

```bash
# Basic reload — works exactly as before, but now returns a token for tracking
$ traffic_ctl config reload
✔ Reload scheduled [rldtk-1739808000000]

  Monitor : traffic_ctl config reload -t rldtk-1739808000000 -m
  Details : traffic_ctl config reload -t rldtk-1739808000000 -s -l
```

```bash
# Monitor mode with a custom token
$ traffic_ctl config reload -t deploy-v2.1 -m
✔ Reload scheduled [deploy-v2.1]
✔ [deploy-v2.1] ████████████████████ 11/11  success  (245ms)
```

> **Note:** All subtasks — both file-based and record-triggered — are registered before the
> first status poll. After `rereadConfig()` processes files, `RecFlushConfigUpdateCbs()` immediately
> fires all pending record callbacks, and each `on_record_change()` calls `reserve_subtask()` to
> pre-register a CREATED subtask on the main task. This means the total task count is stable from
> the start (e.g., `3/11 in_progress` → `11/11 success`) rather than growing over time.

```bash
# Full status report
$ traffic_ctl config status -t deploy-v2.1
✔ Reload [success] — deploy-v2.1
  Started : 2025 Feb 17 12:00:00.123
  Finished: 2025 Feb 17 12:00:00.368
  Duration: 245ms

  ✔ 11 success  ◌ 0 in-progress  ✗ 0 failed  (11 total)

  Tasks:
   ✔ logging.yaml ··························· 120ms
   ✔ ip_allow.yaml ·························· 18ms
   ✔ remap.config ··························· 42ms
   ✔ ssl_client_coordinator ················· 35ms
   ├─ ✔ sni.yaml ··························· 20ms
   └─ ✔ ssl_multicert.config ··············· 15ms
   ...
```

**Failed reload — monitor mode:**

```bash
$ traffic_ctl config reload -t hotfix-ssl-cert -m
✔ Reload scheduled [hotfix-ssl-cert]
✗ [hotfix-ssl-cert] ██████████████░░░░░░ 9/11  fail  (310ms)

  Details : traffic_ctl config status -t hotfix-ssl-cert
```

**Failed reload — status report:**

```bash
$ traffic_ctl config status -t hotfix-ssl-cert
✗ Reload [fail] — hotfix-ssl-cert
  Started : 2025 Feb 17 14:30:10.500
  Finished: 2025 Feb 17 14:30:10.810
  Duration: 310ms

  ✔ 9 success  ◌ 0 in-progress  ✗ 2 failed  (11 total)

  Tasks:
   ✔ ip_allow.yaml ·························· 18ms
   ✔ remap.config ··························· 42ms
   ✗ logging.yaml ·························· 120ms  ✗ FAIL
   ✗ ssl_client_coordinator ················· 85ms  ✗ FAIL
   ├─ ✔ sni.yaml ··························· 20ms
   └─ ✗ ssl_multicert.config ··············· 65ms  ✗ FAIL
   ...
```

**Inline YAML reload (runtime only, not persisted to disk):**

> **Note:** Inline YAML reload is currently disabled — no config handler supports `ConfigSource::FileAndRpc`
> yet. The infrastructure is in place and will be enabled as handlers are migrated. See TODO below.

```bash
$ traffic_ctl config reload -d @ip_allow_new.yaml -t update-ip-rules -m
✔ Reload scheduled [update-ip-rules]
✔ [update-ip-rules] ████████████████████ 1/1  success  (18ms)

Note: Inline configuration is NOT persisted to disk.
      Server restart will revert to file-based configuration.
```

The `-d` flag accepts `@filename` to read from a file, or `@-` to read from stdin. The YAML file
uses **registry keys** as top-level keys — the key string passed as the first argument to
`register_config()` or `register_record_config()`. The content under each key is the actual YAML
that the config file normally contains — it is passed as-is to the handler via `ctx.supplied_yaml()`.
A single file can target multiple handlers:

```yaml
# reload_rules.yaml — multiple configs in one file
# Each top-level key is a registry key (as declared in register_config()).
# The value is the config content (inner data, not wrapped in the file's top-level key).
ip_allow:
  - apply: in
    ip_addrs: 0.0.0.0/0
    action: allow
    methods: ALL
sni:
  - fqdn: "*.example.com"
    verify_client: NONE
```

```bash
# From file — reloads both ip_allow and sni handlers
$ traffic_ctl config reload -d @reload_rules.yaml -t update-rules -m

# From stdin — pipe YAML directly into ATS
$ cat reload_rules.yaml | traffic_ctl config reload -d @- -m
```

### New `traffic_ctl` Commands

| Command | Description |
|---|---|
| `traffic_ctl config reload` | Trigger a file-based reload. Shows token and next-step hints. |
| `traffic_ctl config reload -m` | Trigger and monitor with a live progress bar. |
| `traffic_ctl config reload -s -l` | Trigger and immediately show detailed report with logs. |
| `traffic_ctl config reload -t <token>` | Reload with a custom token. |
| `traffic_ctl config reload -d @file.yaml` | Inline reload from file (runtime only, not persisted). |
| `traffic_ctl config reload -d @-` | Inline reload from stdin. |
| `traffic_ctl config reload --force` | Force a new reload even if one is in progress. |
| `traffic_ctl config reload -m -r 0.5` | Monitor with a 500ms polling interval (default: 0.5s). Accepts fractional values.  0 means no wait.|
| `traffic_ctl config reload -m -w 2` | Monitor with a 2s initial wait before first poll (default: 2s). |
| `traffic_ctl config reload -m -T 30s` | Monitor with a 30s client-side timeout. Exits with `EX_TEMPFAIL` (75) if not done. Requires `-m`. |
| `traffic_ctl config status` | Show the last reload status. |
| `traffic_ctl config status -t <token>` | Show status of a specific reload. |
| `traffic_ctl config status -c all` | Show full reload history (rolling window, oldest evicted at 100). |

All commands support `--format json` to output the raw JSONRPC response instead of human-readable
text. This is useful for automation, CI pipelines, or any tool that consumes structured output
directly:

<details>
<summary><code>traffic_ctl config status -t reload1 --format json</code></summary>

```json
{
  "tasks": [
    {
      "config_token": "reload1",
      "status": "success",
      "description": "Main reload task - 2026 Feb 18 20:03:10",
      "filename": "",
      "meta": {
        "created_time_ms": "1771444990585",
        "last_updated_time_ms": "1771444991015",
        "main_task": "true"
      },
      "log": [],
      "sub_tasks": [
        {
          "config_token": "reload1",
          "status": "success",
          "description": "ip_allow",
          "filename": "/opt/ats/etc/trafficserver/ip_allow.yaml",
          "meta": {
            "created_time_ms": "1771444991013",
            "last_updated_time_ms": "1771444991015",
            "main_task": "false"
          },
          "log": [],
          "logs": [
            "Finished loading"
          ],
          "sub_tasks": []
        },
        {
          "config_token": "reload1",
          "status": "success",
          "description": "ssl_ticket_key",
          "filename": "",
          "meta": {
            "created_time_ms": "1771444991015",
            "last_updated_time_ms": "1771444991015",
            "main_task": "false"
          },
          "log": [],
          "logs": [
            "SSL ticket key reloaded"
          ],
          "sub_tasks": []
        }
      ]
    }
  ]
}
```

</details>

### New JSONRPC APIs

| Method | Description |
|---|---|
| `admin_config_reload` | Unified reload — file-based (default) or inline when `configs` param is present. Params: `token`, `force`, `configs`. |
| `get_reload_config_status` | Query reload status by `token` or get the last N reloads via `count`. |

**Inline reload RPC example:**

```yaml
jsonrpc: "2.0"
method: "admin_config_reload"
params:
  token: "update-ip-and-sni"
  configs:
    ip_allow:
      - apply: in
        ip_addrs: 0.0.0.0/0
        action: allow
        methods: ALL
    sni:
      - fqdn: "*.example.com"
        verify_client: NONE
```

---

### Background: Issues with the Previous Reload Mechanism

The previous configuration reload relied on a loose collection of independent record callbacks
(`RecRegisterConfigUpdateCb`) wired through `FileManager` and `AddConfigFilesHere.cc`. Each config
module registered its file independently, and reloads were fire-and-forget:

- **No visibility** — There was no way to know whether a reload succeeded or failed, which handlers
  ran, or how long each one took.
- **No coordination** — Each handler ran independently with no shared context. There was no concept of
  a "reload session" grouping all config updates triggered by a single request.
- **No inline content** — Configuration could only be reloaded from files on disk. There was no way
  to push YAML content at runtime through the RPC or CLI.
- **Scattered registration** — File registrations were split between `AddConfigFilesHere.cc` (for
  `FileManager`) and individual modules (for record callbacks), making it hard to reason about which
  files were tracked and which records triggered reloads.
- **No token tracking** — There was no identifier for a reload operation, so you couldn't query the
  status of a specific reload or distinguish between overlapping reloads.

### What the New Design Solves

1. **Full reload traceability** — Every reload gets a token. Each config handler reports its status
   (`in_progress`, `success`, `fail`) through a `ConfigContext`. Results are aggregated into a task
   tree with per-handler timings and logs.
2. **Centralized registration** — `ConfigRegistry` is the single source of truth for all config files,
   their filename records, trigger records, and reload handlers.
3. **Inline YAML injection** — Handlers that opt in (`ConfigSource::FileAndRpc`) can receive YAML
   content directly through the RPC, without writing to disk. This is runtime-only — the content
   lives in memory and is lost on restart.
4. **Coordinated reload sessions** — `ReloadCoordinator` manages the lifecycle of each reload:
   token generation, concurrency control (`--force` to override), timeout detection, and history.
5. **CLI observability** — `traffic_ctl config reload -m` shows a live progress bar.
   `traffic_ctl config status` provides a full post-mortem with task tree, durations, and failure
   details.

### Basic Design

```
┌─────────────┐   JSONRPC    ┌────────────────┐
│ traffic_ctl  │────────────►│  RPC Handler   │
│ config reload│             │  reload_config  │
└─────────────┘             └───────┬────────┘
                                    │
          ┌─────────────────────────┼─────────────────────────┐
          │                         ▼                         │
          │              ┌──────────────────┐                 │
          │              │ ReloadCoordinator │                 │
          │              │  - prepare_reload │                 │
          │              │  - token tracking │                 │
          │              │  - history        │                 │
          │              └────────┬─────────┘                 │
          │                       │                           │
          │           ┌───────────┴──────────┐                │
          │           ▼                      ▼                │
          │   ┌──────────────┐      ┌──────────────────┐      │
          │   │ File-based   │      │ Inline mode      │      │
          │   │ FileManager  │      │ set_passed_config │      │
          │   │ rereadConfig │      │ schedule_reload   │      │
          │   └──────┬───────┘      └────────┬─────────┘      │
          │          │                       │                │
          │          └───────────┬───────────┘                │
          │                      ▼                            │
          │            ┌──────────────────┐                   │
          │            │  ConfigRegistry  │                   │
          │            │  execute_reload  │                   │
          │            └────────┬─────────┘                   │
          │                     ▼                             │
          │           ┌──────────────────┐                    │
          │           │  ConfigContext   │                    │
          │           │  - in_progress() │                    │
          │           │  - log()         │                    │
          │           │  - complete()    │                    │
          │           │  - fail()        │                    │
          │           │  - supplied_yaml()│                   │
          │           └────────┬─────────┘                    │
          │                    ▼                              │
          │           ┌──────────────────┐                    │
          │           │  Handler         │                    │
          │           │  (IpAllow, SNI,  │                    │
          │           │   remap, etc.)   │                    │
          │           └──────────────────┘                    │
          └───────────────────────────────────────────────────┘
```

**Key components:**

| Component | Role |
|---|---|
| `ConfigRegistry` | Singleton registry mapping config keys to handlers, filenames, trigger records. Self-registers with `FileManager`. |
| `ReloadCoordinator` | Manages reload sessions: token generation, concurrency, timeout detection, history. |
| `ConfigReloadTask` | Tracks a single reload operation as a tree of sub-tasks with status, timings, and logs. |
| `ConfigContext` | Lightweight context passed to handlers. Provides `in_progress()`, `complete()`, `fail()`, `log()`, `supplied_yaml()`, and `add_dependent_ctx()`. Safe no-op at startup (no active reload task). |
| `ConfigReloadProgress` | Periodic checker that detects stuck tasks and marks them as `TIMEOUT`. |

**Thread model:**

All reload work runs on **`ET_TASK`** threads — never on the RPC thread or event-loop threads:

1. **RPC thread** — receives the JSONRPC request (`admin_config_reload`), creates the reload token
   and task via `ReloadCoordinator::prepare_reload()`, then schedules the actual work on `ET_TASK`
   and returns immediately. The RPC response (with the token) is sent back to `traffic_ctl` before
   any handler runs.
2. **`ET_TASK` — file-based reload** — `ReloadWorkContinuation` fires on `ET_TASK`. It calls
   `FileManager::rereadConfig()`, which walks every registered file and invokes
   `ConfigRegistry::execute_reload()` for each changed config. Each handler runs **synchronously**
   on this thread.
3. **`ET_TASK` — inline (RPC) reload** — `ScheduledReloadContinuation` fires on `ET_TASK`. It calls
   `ConfigRegistry::execute_reload()` directly for the targeted config key(s). Same synchronous
   execution.
4. **Deferred handlers** — Some handlers (e.g., `LogConfig`) schedule work on other threads and
   return before completion. The `ConfigContext` they hold remains valid (weak pointer + ref-counted
   YAML node), and they must call `ctx.complete()` or `ctx.fail()` from whatever thread finishes
   the work. If they don't, the timeout checker marks the task as `TIMEOUT`.
5. **Timeout checker** — `ConfigReloadProgress` is a per-reload continuation on `ET_TASK` that polls
   periodically and marks stuck tasks as `TIMEOUT`.

Handlers block `ET_TASK` while they run. A slow handler delays all subsequent handlers in the same
reload cycle. This is the same behavior as the previous mechanism — the difference is that now it's
visible through the task tree and timings.

**Stuck reload checker:**
`ConfigReloadProgress` is a periodic continuation scheduled on `ET_TASK`. It monitors active reload
tasks and marks any that exceed the configured timeout as `TIMEOUT`. This acts as a safety net for
handlers that fail to call `ctx.complete()` or `ctx.fail()` — for example, if a handler crashes,
deadlocks, or its deferred thread never executes. The checker reads `proxy.config.admin.reload.timeout`
dynamically at each interval, so the timeout can be adjusted at runtime without a restart. This is
a simple record read (`RecGetRecordString`), not an expensive operation. Setting the
timeout to `"0"` disables it (tasks will run indefinitely until completion).
The checker is not a global poller — a new instance is created per-reload and self-terminates once
the task reaches a terminal state. No idle polling when no reload is in progress.

### How Handlers Work

**Before — scattered registration (ip_allow example):**

Registration was split across multiple files with no centralized tracking:

```c++
// 1. AddConfigFilesHere.cc — register with FileManager for mtime detection
registerFile("proxy.config.cache.ip_allow.filename", ts::filename::IP_ALLOW, NOT_REQUIRED);
registerFile("proxy.config.cache.ip_categories.filename", ts::filename::IP_CATEGORIES, NOT_REQUIRED);

// 2. IPAllow.cc — attach record callback (fire-and-forget, no status tracking)
ConfigUpdateHandler<IpAllow> *ipAllowUpdate = new ConfigUpdateHandler<IpAllow>("ip_allow");
ipAllowUpdate->attach("proxy.config.cache.ip_allow.filename");

// 3. IpAllow::reconfigure() — no context, no status, no tracing
void IpAllow::reconfigure() {
    // ... load config from disk, no way to report success/failure ...
}
```

**Now — each module self-registers with full tracing:**

Each module registers itself directly with `ConfigRegistry`. No more separate `AddConfigFilesHere.cc`
entry — the registry handles `FileManager` registration, record callbacks, and status tracking
automatically:

```c++
// IPAllow.cc — one call replaces all three steps above
config::ConfigRegistry::Get_Instance().register_config(
    "ip_allow",                                           // registry key
    ts::filename::IP_ALLOW,                               // default filename
    "proxy.config.cache.ip_allow.filename",               // record holding filename
    [](ConfigContext ctx) { IpAllow::reconfigure(ctx); }, // handler with context
    config::ConfigSource::FileOnly,                       // content source
    {"proxy.config.cache.ip_allow.filename"});            // trigger records

// Auxiliary file — attach ip_categories as a dependency (changes trigger ip_allow reload)
config::ConfigRegistry::Get_Instance().add_file_dependency(
    "ip_allow",                                  // config key to attach to
    "proxy.config.cache.ip_categories.filename", // record holding the filename
    ts::filename::IP_CATEGORIES,                 // default filename
    false);                                      // not required
```

Additional triggers can be attached from any module at any time:

```c++
// From another module — attach an extra record trigger
config::ConfigRegistry::Get_Instance().attach("ip_allow", "proxy.config.some.extra.record");
```

Composite configs can declare file dependencies and dependency keys. For example, `SSLClientCoordinator`
owns `sni.yaml` and `ssl_multicert.config` as children:

```c++
// Main registration (no file of its own — it's a pure coordinator)
// From SSLClientCoordinator::startup()
config::ConfigRegistry::Get_Instance().register_record_config(
    "ssl_client_coordinator",                                          // registry key
    [](ConfigContext ctx) { SSLClientCoordinator::reconfigure(ctx); }, // reload handler
    {"proxy.config.ssl.client.cert.path",                              // trigger records
     "proxy.config.ssl.client.cert.filename", "proxy.config.ssl.client.private_key.path",
     "proxy.config.ssl.client.private_key.filename", "proxy.config.ssl.keylog_file",
     "proxy.config.ssl.server.cert.path", "proxy.config.ssl.server.private_key.path",
     "proxy.config.ssl.server.cert_chain.filename",
     "proxy.config.ssl.server.session_ticket.enable"});

// Track sni.yaml — FileManager watches for mtime changes, record wired to trigger reload
config::ConfigRegistry::Get_Instance().add_file_and_node_dependency(
    "ssl_client_coordinator", "sni",
    "proxy.config.ssl.servername.filename", ts::filename::SNI, false);

// Track ssl_multicert.config — same pattern
config::ConfigRegistry::Get_Instance().add_file_and_node_dependency(
    "ssl_client_coordinator", "ssl_multicert",
    "proxy.config.ssl.server.multicert.filename", ts::filename::SSL_MULTICERT, false);
```

**Handler interaction with `ConfigContext`:**

Each config module implements a C++ reload handler — the callback passed to `register_config()`.
The handler reports progress through the `ConfigContext`:

```c++
void IpAllow::reconfigure(ConfigContext ctx) {
    ctx.in_progress();

    // ... load config from disk ...

    ctx.complete("Loaded successfully");
    // or on error:
    // ctx.fail(errata, "Failed to load");
}
```

When a reload fires, the handler receives a `ConfigContext`:

- **File source** — `ctx.supplied_yaml()` is undefined; the handler reads from its registered file on disk.
- **RPC source** — `ctx.supplied_yaml()` contains the YAML node passed via `--data` / RPC.
  The content is **runtime-only** and is never written to disk.

Handlers report progress:

```c++
ctx.in_progress("Parsing ip_allow.yaml");
ctx.log("Loaded 42 rules");
ctx.complete("Finished loading");
// or on error:
ctx.fail(errata, "Failed to load ip_allow.yaml");
```

**Supplied YAML — inline content via `-d` / RPC:**

> **Note:** The infrastructure for RPC-supplied YAML is fully implemented, but no handler currently
> opts into `ConfigSource::FileAndRpc`. File-based handlers use `ConfigSource::FileOnly`, and
> record-only handlers use `ConfigSource::RecordOnly` (implicitly via `register_record_config()`).

When a handler opts into `ConfigSource::FileAndRpc`, it can receive YAML content directly instead
of reading from disk. The handler checks `ctx.supplied_yaml()` to determine the source:

```c++
void IpAllow::reconfigure(ConfigContext ctx) {
    ctx.in_progress();

    YAML::Node root;
    if (auto yaml = ctx.supplied_yaml()) {
        // Inline mode: YAML supplied via -d flag or JSONRPC.
        // Not persisted to disk — runtime only.
        root = yaml;
    } else {
        // File mode: read from the registered config file on disk.
        root = YAML::LoadFile(config_filename);
    }

    // ... parse and apply config ...

    ctx.complete("Loaded successfully");
}
```

For composite configs (e.g., `SSLClientCoordinator`), handlers create child contexts to track
each sub-config independently. From `SSLClientCoordinator::reconfigure()`:

```c++
SSLConfig::reconfigure(reconf_ctx.add_dependent_ctx("SSLConfig"));
SNIConfig::reconfigure(reconf_ctx.add_dependent_ctx("SNIConfig"));
SSLCertificateConfig::reconfigure(reconf_ctx.add_dependent_ctx("SSLCertificateConfig"));
reconf_ctx.complete("SSL configs reloaded");
```

The parent task automatically aggregates status from its children. In `traffic_ctl config status`,
this renders as a tree:

```
   ✔ ssl_client_coordinator ················· 35ms
   ├─ ✔ SSLConfig ·························· 10ms
   ├─ ✔ SNIConfig ·························· 12ms
   └─ ✔ SSLCertificateConfig ·············· 13ms
```

#### 1. Handlers must reach a terminal state — or the task hangs

The entire tracing model relies on handlers calling `ctx.complete()` or `ctx.fail()` before
returning. If a handler returns without reaching a terminal state, the task stays `IN_PROGRESS`
indefinitely until the timeout checker marks it as `TIMEOUT`.

After `execute_reload()` calls the handler, it checks `ctx.is_terminal()` and emits a warning
if the handler left the task in a non-terminal state:

```c++
entry_copy.handler(ctx);
if (!ctx.is_terminal()) {
    Warning("Config '%s' handler returned without reaching a terminal state. "
            "If the handler deferred work to another thread, ensure ctx.complete() or ctx.fail() "
            "is called when processing finishes; otherwise the task will remain in progress "
            "until the timeout checker marks it as TIMEOUT.",
            entry_copy.key.c_str());
}
```

**The safety net:** `ConfigReloadProgress` runs periodically on `ET_TASK` and marks stuck tasks as
`TIMEOUT` after the configured duration (`proxy.config.admin.reload.timeout`, default: `1h`).

#### 2. Parent status aggregation from sub-tasks

Parent tasks do **not** track their own status directly — they derive it from their children.
When a child calls `complete()` or `fail()`, it notifies its parent, which re-evaluates:

- **Any child failed or timed out** → parent is `FAIL`
- **Any child still in progress** → parent stays `IN_PROGRESS`
- **All children succeeded** → parent is `SUCCESS`

This aggregation is recursive: a sub-task can have its own children (e.g.,
`ssl_client_coordinator` → `sni` + `ssl_multicert`), and status bubbles up through the tree.

One subtle issue: if a handler creates child contexts but **forgets to call `complete()` or
`fail()`** on one of them, that child stays `CREATED` and the parent never reaches `SUCCESS`.
It is the handler developer's responsibility to ensure every `ConfigContext` (and its children)
reaches a terminal state (`complete()` or `fail()`). The timeout checker is the ultimate safety
net for cases where this is not properly handled.

#### 3. Startup vs. reload — same handler, different context

Handlers are called both at startup (initial config load) and during runtime reloads. At startup,
there is no active `ReloadCoordinator` task, so all `ConfigContext` operations (`in_progress()`,
`complete()`, `fail()`, `log()`) are **safe no-ops** — they check `_task.lock()` and return
immediately if the weak pointer is expired or empty.

This avoids having two separate code paths for startup vs. reload. The handler logic is identical
in both cases:

```c++
void IpAllow::reconfigure(ConfigContext ctx) {
    ctx.in_progress();  // no-op at startup, tracks progress during reload
    // ... load config ...
    ctx.complete();     // no-op at startup, marks task as SUCCESS during reload
}
```

#### 4. Duplicate handler execution for multi-record triggers (fixed)

**Known issue:** `ssl_client_coordinator` registers multiple trigger records and file dependencies,
each wiring an independent `on_record_change` callback. When several of these fire during the same
reload, the handler executes more than once, producing duplicate entries in the reload status output.
This behavior exists on `master` as well, but was not visible because the old reload path had no status tracking. The new framework's task tree makes the duplicate executions observable.
This is a pre-existing issue present on `master` — see
[#11724](https://github.com/apache/trafficserver/issues/11724).
**This issue is resolved in this PR.**

> **Note:** This fix exists only in this PR — it relies on the `ConfigReloadTask` subtask tree
> and `config_key` tracking introduced by the new framework. The old reload path on `master` has
> no equivalent structure to deduplicate against; each `RecRegisterConfigUpdateCb` callback fires
> independently with no shared state. A fix on `master` would require adding an `atomic<bool>`
> CAS flag per config key, but that approach risks lost updates when handlers reschedule work
> by modifying their own trigger records during reconfiguration.

#### 5. Plugin support

Plugins are **not** supported by `ConfigRegistry` in this PR. The legacy reload notification
mechanism (`TSMgmtUpdateRegister`) still works — plugins registered through it will continue
to be invoked via `FileManager::invokeConfigPluginCallbacks()` during every reload cycle.
A dedicated plugin API to let plugins register their own config handlers and participate in
the reload framework will be addressed in a separate PR.

#### 6. Subtask reservation — ensuring all handlers are tracked from the start

During a file-based reload, handlers are discovered in two phases:

- **File-based handlers** (e.g., `ip_allow`, `remap`) run **synchronously** inside
  `FileManager::rereadConfig()`. Their subtasks are created and completed immediately on the
  `ET_TASK` thread that drives the reload.
- **Record-triggered handlers** (e.g., `cache_control`, `ssl_ticket_key`,
  `ssl_client_coordinator`) are activated by record callbacks. When `rereadConfig()` detects
  changed files, it calls `RecSetSyncRequired()` to flag the associated records as dirty.

To avoid a gap between file-based completions and record-triggered registrations,
`RecFlushConfigUpdateCbs()` is called immediately after `rereadConfig()`. This synchronously
fires all pending record callbacks instead of waiting for the next `config_update_cont` tick
(~3s). Each `on_record_change()` callback calls `reserve_subtask()` to pre-register a CREATED
subtask on the main task, then schedules the actual handler on `ET_TASK`.

The result is that all subtasks — both file-based and record-triggered — are registered before
the reload executor returns. The task count is stable from the first status poll:

1. `rereadConfig()` finishes — file-based subtasks are created and completed.
2. `RecFlushConfigUpdateCbs()` fires — `on_record_change()` callbacks run synchronously,
   each calling `reserve_subtask()` to register a CREATED subtask.
3. Main task sees CREATED children alongside completed children — stays IN_PROGRESS.
4. Record-triggered continuations run on `ET_TASK` — activate their reserved subtasks via
   `create_config_context()`, execute the handler, and complete the subtask.
5. When the last handler completes — main task reaches SUCCESS.

As a safety net, `add_sub_task()` also calls `aggregate_status()` when the parent has already
reached SUCCESS, reverting it to IN_PROGRESS. This handles any edge case where a subtask is
registered after all known work has completed (e.g., from a plugin or deferred callback).

---

### Configs Migrated to `ConfigRegistry`

| Config | Key | File |
|---|---|---|
| IP Allow | `ip_allow` | `ip_allow.yaml` |
| IP Categories | (dependency of `ip_allow`) | `ip_categories.yaml` |
| Cache Control | `cache_control` | `cache.config` |
| Cache Hosting | `cache_hosting` | `hosting.config` |
| Parent Selection | `parent_proxy` | `parent.config` |
| Split DNS | `split_dns` | `splitdns.config` |
| Remap | `remap` | `remap.config` ¹ |
| Logging | `logging` | `logging.yaml` |
| SSL/TLS (coordinator) | `ssl_client_coordinator` | — |
| SNI | (dependency of `ssl_client_coordinator`) | `sni.yaml` |
| SSL Multicert | (dependency of `ssl_client_coordinator`) | `ssl_multicert.config` |
| SSL Ticket Key | `ssl_ticket_key` | (record-only, no file) |
| Records | `records` | `records.yaml` |
| Storage | (static, no handler) | `storage.config` |
| Socks | (static, no handler) | `socks.config` |
| Volume | (static, no handler) | `volume.config` |
| Plugin | (static, no handler) | `plugin.config` |
| JSONRPC | (static, no handler) | `jsonrpc.yaml` |

¹ Remap migration will be refactored after [#12813](https://github.com/apache/trafficserver/pull/12813) (remap.yaml)
and [#12669](https://github.com/apache/trafficserver/pull/12669) (virtual hosts) land.

### New Configuration Records

```yaml
records:
  admin:
    reload:
      # Maximum time a reload task can run before being marked as TIMEOUT.
      # Supports duration strings: "30s", "5min", "1h". Set to "0" to disable.
      # Default: 1h. Updateable at runtime (RECU_DYNAMIC).
      timeout: 1h

      # How often the progress checker polls for stuck tasks (minimum: 1s).
      # Supports duration strings: "1s", "5s", "30s".
      # Default: 2s. Updateable at runtime (RECU_DYNAMIC).
      check_interval: 2s
```

---

### TODO
- [x] **Clean up** - Run some clean up on this code.
- [x] **Documentation** — User-facing docs added: `traffic_ctl.en.rst` (commands & options), `jsonrpc-api.en.rst` (`admin_config_reload` / `get_reload_config_status`), and `config-reload-framework.en.rst` (developer guide).
- [x] **Remove legacy reload infrastructure** — `ConfigUpdateHandler`/`ConfigUpdateContinuation` removed from `ConfigProcessor.h`. Remaining `registerFile()` calls in `AddConfigFilesHere.cc` will be retired via inventory-only `ConfigRegistry` entries (see below).
- [x] **Consolidate `AddConfigFilesHere.cc` into `ConfigRegistry`** — Static files (`storage.config`, `volume.config`, `plugin.config`, etc.) registered as inventory-only entries. `records.yaml` registered with reload handler. `AddConfigFilesHere.cc` removed.
- [x] **Additional tests** — Expand autest coverage.

#### Future Work (separate PRs)
- **Improve error reporting** — All config loaders are migrated to `ConfigRegistry`. Remaining work: fully log detailed errors via `ctx.log()`, `ctx.fail()`, etc.
- **Enable inline YAML for more handlers** — Currently file-based handlers use `ConfigSource::FileOnly` and record-only handlers use `ConfigSource::RecordOnly`. Migrate file-based handlers to `FileAndRpc` so they can read YAML directly from the RPC (via `ctx.supplied_yaml()`).
- **Autest reload extension** — Implement an autest extension that checks reload success/failure via the JSONRPC status API (`traffic_ctl config status -t <token>`) instead of grepping log files.
- **Trace record-triggered reloads** — Record-based reloads (via `trigger_records` / `RecRegisterConfigUpdateCb`) are not currently tracked. Create a main task with a synthetic token so they appear in `traffic_ctl config status`.
- **Expose ConfigRegistry to plugins** — Add a plugin API so plugins can register their own config handlers and participate in the reload framework.

---

### Dependencies and Related Issues

Cherry-pick [#12934](https://github.com/apache/trafficserver/pull/12934) for an Argparser fix.

Fixes [#12324 — Improving `traffic_ctl config reload`](https://github.com/apache/trafficserver/issues/12324).

This PR will likely land after:
- [#12813](https://github.com/apache/trafficserver/pull/12813)
- [#12669](https://github.com/apache/trafficserver/pull/12669)

There should be no major conflicts with those PRs. Conversation and coordination needs to be done before merging.

### DIFF
--- a/src/iocore/net/SSLConfig.cc
+++ b/src/iocore/net/SSLConfig.cc
@@ -800,8 +800,8 @@ SSLTicketKeyConfig::startup()
 {
   config::ConfigRegistry::Get_Instance().register_record_config("ssl_ticket_key",       // key
                                                                 [](ConfigContext ctx) { // handler callback
-                                                                  // eventually ctx should passed throuough to the reconfigure fn to
-                                                                  // the loaders to it can show more details.
+                                                                  // eventually ctx should be passed through to the reconfigure fn
+                                                                  // and the loaders so it can show more details.
                                                                   if (SSLTicketKeyConfig::reconfigure(ctx)) {
                                                                     ctx.complete("SSL ticket key reloaded");
                                                                   } else {
--- a/src/mgmt/config/ConfigContext.cc
+++ b/src/mgmt/config/ConfigContext.cc
@@ -129,7 +129,7 @@ ConfigContext::add_dependent_ctx(std::string_view description)
   if (auto p = _task.lock()) {
     auto child = p->add_child(description);
     // child task will get the full content of the parent task
-    // TODO: eventyually we can have a "key" passed so child module
+    // TODO: eventually we can have a "key" passed so child module
     // only gets their node of interest.
     child._supplied_yaml = _supplied_yaml;
     return child;
--- a/src/proxy/IPAllow.cc
+++ b/src/proxy/IPAllow.cc
@@ -125,6 +125,7 @@ IpAllow::reconfigure(ConfigContext ctx)
 {
   self_type  *new_table;
   std::string text;
+  std::string errata_text;
   Note("%s loading ...", ts::filename::IP_ALLOW);
   ctx.in_progress();
 
@@ -132,17 +133,19 @@ IpAllow::reconfigure(ConfigContext ctx)
   // IP rules need categories, so load them first (if they exist).
   if (auto errata = new_table->BuildCategories(); !errata.is_ok()) {
     swoc::bwprint(text, "{} failed to load", new_table->ip_categories_config_file);
-    Error("%s\n%s", text.c_str(), swoc::bwprint(text, "{}", errata).c_str());
+    swoc::bwprint(errata_text, "{}", errata);
+    Error("%s\n%s", text.c_str(), errata_text.c_str());
     ctx.fail(errata, "{} failed to load", new_table->ip_categories_config_file);
     delete new_table;
     return;
   }
   if (auto errata = new_table->BuildTable(); !errata.is_ok()) {
     swoc::bwprint(text, "{} failed to load", ts::filename::IP_ALLOW);
+    swoc::bwprint(errata_text, "{}", errata);
     if (errata.severity() <= ERRATA_ERROR) {
-      Error("%s\n%s", text.c_str(), swoc::bwprint(text, "{}", errata).c_str());
+      Error("%s\n%s", text.c_str(), errata_text.c_str());
     } else {
-      Fatal("%s\n%s", text.c_str(), swoc::bwprint(text, "{}", errata).c_str());
+      Fatal("%s\n%s", text.c_str(), errata_text.c_str());
     }
     ctx.fail(errata, "{} failed to load", ts::filename::IP_ALLOW);
     delete new_table;
--- a/src/proxy/http/HttpConfig.cc
+++ b/src/proxy/http/HttpConfig.cc
@@ -191,7 +191,7 @@ int
 HttpConfigCont::handle_event(int /* event ATS_UNUSED */, void * /* edata ATS_UNUSED */)
 {
   if (ink_atomic_increment(&http_config_changes, -1) == 1) {
-    HttpConfig::reconfigure(); // TODO: who's calling?
+    HttpConfig::reconfigure();
   }
   return 0;
 }
--- a/src/traffic_ctl/jsonrpc/CtrlRPCRequests.h
+++ b/src/traffic_ctl/jsonrpc/CtrlRPCRequests.h
@@ -81,7 +81,7 @@ struct ConfigReloadResponse {
   };
   std::vector<Error> error; ///< Error list, if any.
 
-  // when requestiong existing tasks.
+  // when requesting existing tasks.
   std::vector<ReloadInfo> tasks;
 
   std::string              created_time;
--- a/tests/gold_tests/parent_config/parent_config_reload.test.py
+++ b/tests/gold_tests/parent_config/parent_config_reload.test.py
@@ -79,7 +79,7 @@ p.ReturnCode = Any(0, -2)
 p.Ready = When.FileContains(ts.Disk.diags_log.Name, "parent.config finished loading", 3)
 p.Timeout = 20
 tr.Processes.Default.StartBefore(p)
-## TODO: we should have an extension like When.ReloadCompleted(token, success) to validate this inetasd of parsing
+## TODO: we should have an extension like When.ReloadCompleted(token, success) to validate this instead of parsing
 ##       diags.
 tr.Processes.Default.Command = 'echo "waiting for parent.config reload after record change"'
 tr.TimeOut = 25
--- a/tests/gold_tests/remap/remap_reload.test.py
+++ b/tests/gold_tests/remap/remap_reload.test.py
@@ -116,7 +116,3 @@ tr.AddVerifierClientProcess("client_3", replay_file_3, http_ports=[tm.Variables.
 
 tr = Test.AddTestRun("post update golf")
 tr.AddVerifierClientProcess("client_4", replay_file_4, http_ports=[tm.Variables.port])
-
-tr = Test.AddTestRun("remap_config reload, test")
-tr.Processes.Default.Env = tm.Env
-tr.Processes.Default.Command = 'sleep 2; traffic_ctl rpc invoke get_reload_config_status -f json | jq'
--- a/tests/gold_tests/traffic_ctl/traffic_ctl_config_reload.test.py
+++ b/tests/gold_tests/traffic_ctl/traffic_ctl_config_reload.test.py
@@ -103,7 +103,7 @@ traffic_ctl.config().reload().token(token).validate_with_text(
 
 # Modify ip_allow.yaml and validate the reload status.
 
-tr = Test.AddTestRun("rouch file to trigger ip_allow reload")
+tr = Test.AddTestRun("touch file to trigger ip_allow reload")
 tr.Processes.Default.Command = f"touch {os.path.join(traffic_ctl._ts.Variables.CONFIGDIR, 'ip_allow.yaml')}  && sleep 1"
 tr.Processes.Default.ReturnCode = 0
 


brbzull0:
[approve ci autest 1]

brbzull0:
[approve ci]

brbzull0:
[approve ci]

brbzull0:
[approve ci autest 1]

brbzull0:
[approve ci freebsd clang-analyzer]

PR #12895, opened 2026-02-19
rpc: tolerate unsupported chmod on unix sockets

bneradt:
test_jsonrpcserver can fail on containerized macOS-host mounts where
chmod() on an AF_UNIX socket inode returns EINVAL or ENOTSUP.
When this happens, RPC server init aborts and the client test
fails with connection refused.

Treat unsupported chmod-on-socket errors as non-fatal and keep
running with umask-derived permissions, while preserving fatal
behavior for other chmod failures. Also zero-initialize
sockaddr_un before use to avoid carrying stale bytes into bind().

### DIFF
--- a/src/mgmt/rpc/server/IPCSocketServer.cc
+++ b/src/mgmt/rpc/server/IPCSocketServer.cc
@@ -140,6 +140,7 @@ std::error_code
 IPCSocketServer::init()
 {
   std::error_code ec; // Flag possible errors.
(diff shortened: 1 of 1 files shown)


brbzull0 (approved):
make sense. Thanks.

bneradt:
[approve ci autest 1]

cmcfarlen:
Cherry-picked to 10.2.x

PR #12904, opened 2026-02-20
hrw: fix lost [OR] modifiers, add more autests

zwoop:
This fixes an issue where when modifiers such as [OR] are used with %{GROUP}'s, they can be lost which then defaults to an [AND]. This does not make the customer happy.

### DIFF
--- a/plugins/header_rewrite/condition.h
+++ b/plugins/header_rewrite/condition.h
@@ -85,6 +85,12 @@ public:
     return _mods;
   }
 
+  void
+  set_mods(CondModifiers mods)
+  {
+    _mods = mods;
+  }
+
   // Setters
   virtual void
   set_qualifier(const std::string &q)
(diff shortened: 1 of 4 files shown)


zwoop:
[approve ci]

bryancall (approved):
Clean, well-targeted fix. The bug is clear: when `%{GROUP:END} [OR]` is parsed, the temporary closing `ConditionGroup` object (which holds the `[OR]` modifier) was `delete`d before the modifier could be transferred to the actual group. This silently degrades `[OR]` to the default `[AND]` behavior — a nasty one to diagnose in production since the config loads without errors.

The fix correctly reorders to transfer modifiers before deletion, and the new `set_mods()` setter on `Condition` is the right approach since `_mods` is private with only a getter.

Test coverage is excellent — 22+ cases covering GROUP with OR/AND/NOT, nested groups, two groups with OR between them, groups inside if/elif/else, complex `(A OR B) AND (C OR D)` patterns, and IP conditions in groups. The `hrw4u:` pseudo-code comments in `complex_logics.conf` make the intended boolean logic very readable.

Minor nit: the replay yaml file is missing a trailing newline.

zwoop:
[approve ci]

cmcfarlen:
Cherry-picked to 10.2.x

Issue #12244, opened 2025-05-15
Failed assertion in http tunnel


mrdg:
This is on version 9.2.10

```
Fatal: HttpTunnel.cc:991: failed assertion `c->skip_bytes <= c->buffer_reader->read_avail()`
```

backtrace

```
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x0000152a601e127e in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x0000152a601c48ff in __GI_abort () at ./stdlib/abort.c:79
#5  0x0000152a60e71e1e in ink_abort (message_format=message_format@entry=0x152a60ee7cdb "%s:%d: failed assertion `%s`") at ink_error.cc:99
#6  0x0000152a60e6e04b in _ink_assert (expression=<optimized out>, file=<optimized out>, line=<optimized out>) at ink_assert.cc:37
#7  0x0000560eb7a58d08 in HttpTunnel::producer_run (this=0x560fcfe5db08, p=0x560fcfe5dd10) at HttpTunnel.cc:991
#8  0x0000560eb7a59b40 in HttpTunnel::tunnel_run (this=0x560fcfe5db08, p_arg=0x560fcfe5dd10) at HttpTunnel.cc:818
#9  0x0000560eb79e439e in HttpSM::state_api_callout (this=this@entry=0x560fcfe5cb00, event=event@entry=60000, data=data@entry=0x0) at HttpSM.cc:1648
#10 0x0000560eb79eb989 in HttpSM::state_api_callback (this=this@entry=0x560fcfe5cb00, event=event@entry=60000, data=data@entry=0x0) at HttpSM.cc:1453
#11 0x0000560eb7962dee in TSHttpTxnReenable (txnp=0x560fcfe5cb00, event=TS_EVENT_HTTP_CONTINUE) at traffic_server/InkAPI.cc:6316
#12 0x0000152a540b625d in transform_plugin (contp=0x1528a21f87a0, event=TS_EVENT_HTTP_READ_RESPONSE_HDR, edata=0x560fcfe5cb00) at compress/compress.cc:908
[...]

PR #12905, opened 2026-02-23
fix(http): crash in `HttpTunnel::producer_run` when 100 Continue precedes a transformed response

JakeChampion:
When an origin sends 100 Continue before a compressible response, `setup_100_continue_transfer()` sets `client_response_hdr_bytes` to the intermediate header size.
The subsequent call to `setup_server_transfer_to_transform()` creates the tunnel buffer with `hdr_size=0` - body only, but `client_response_hdr_bytes` is not reset for non-chunked responses.

`perform_cache_write_action()` then passes the stale value as `skip_bytes` to the cache-write consumer, causing `ink_release_assert(c->skip_bytes <= c->buffer_reader->read_avail())` to fire in `producer_run` when the response body is smaller than the 100 Continue headers.

Fix: set `cache_write_skip` to `0` whenever a transform is present, since `setup_server_transfer_to_transform()` never writes headers into the tunnel buffer.

Fixes: #12244


bryancall:
@JakeChampion we weren't going to do another 10.1.x release.

JakeChampion:
> @JakeChampion we weren't going to do another 10.1.x release.

I've also opened this against main, and can back port to 10.2.x instead of 10.1.x if preferred

PR #12906, opened 2026-02-23
fix(http): stale `skip_bytes` in cache-write consumer after 100 Continue with transform

JakeChampion:
When an origin sends 100 Continue before a compressible response, `setup_100_continue_transfer()` sets `client_response_hdr_bytes` to the intermediate header size. `setup_server_transfer_to_transform()` then creates the tunnel buffer with `hdr_size=0` - body only, but does not reset `client_response_hdr_bytes` for non-chunked responses. `perform_cache_write_action()` passes the stale value as `skip_bytes` to the cache-write consumer, causing `ink_release_assert` in `producer_run` when the response body is smaller than the 100 Continue headers.

Fix: Set `cache_write_skip` to `0` when a transform is present, since `setup_server_transfer_to_transform()` never writes headers into the tunnel buffer.

---

Standard flow with no 100 Continue, no compress plugin transform:

ATS creates a tunnel buffer that contains [response headers][response body]
When writing to cache, it needs to skip past the headers so only the body gets stored
It uses a variable called `skip_bytes` for this
`skip_bytes` = size of response headers

What goes wrong with 100 Continue + compress transform:

1 - Origin sends 100 Continue:

Origin -> ATS: HTTP/1.1 100 Continue\r\n\r\n   (25 bytes)
ATS forwards this to the client and records `client_response_hdr_bytes = 25` (the size of those headers)
This is used later to know how much header data was written to the tunnel buffer

2 - Origin sends the real 200 OK response:
Because compress plugin is active, ATS sets up a server-to-transform tunnel
The buffer for this tunnel contains only the body, no headers go into this buffer at all, since the compress transform handles headers separately
`client_response_hdr_bytes` is not reset here for non-chunked responses, it is still 25 bytes from Step 1

3 - Cache write:
`perform_cache_write_action()` says wants to skip past the headers in the tunnel buffer before writing to cache and uses `client_response_hdr_bytes` as it's `skip_bytes`
In this scenario that means it tries to skip 25 bytes

But the tunnel buffer only has the body bytes in it, making the assertion fire and crash ATS

### DIFF
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -6591,8 +6591,16 @@ HttpSM::perform_cache_write_action()
     if (transform_info.entry == nullptr || t_state.api_info.cache_untransformed == true) {
       cache_sm.close_read();
       t_state.cache_info.write_status = HttpTransact::CacheWriteStatus_t::IN_PROGRESS;
-      setup_cache_write_transfer(&cache_sm, server_entry->vc, &t_state.cache_info.object_store, client_response_hdr_bytes,
-                                 "cache write");
+
+      // Decide how many header bytes the cache-write consumer should skip.
+      // When a transform is present, setup_server_transfer_to_transform()
+      // creates the tunnel buffer with hdr_size=0, so the buffer contains
+      // only body data - no response headers to skip.  When no transform is
+      // present, setup_server_transfer() writes the response header into the
+      // buffer, so skip_bytes must equal client_response_hdr_bytes.
+      int64_t cache_write_skip = (transform_info.entry == nullptr) ? client_response_hdr_bytes : 0;
+
+      setup_cache_write_transfer(&cache_sm, server_entry->vc, &t_state.cache_info.object_store, cache_write_skip, "cache write");
     } else {
       // We are not caching the untransformed.  We might want to
       //  use the cache writevc to cache the transformed copy
--- a/tests/gold_tests/pluginTest/compress/compress_100_continue_origin.py
+++ b/tests/gold_tests/pluginTest/compress/compress_100_continue_origin.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python3
+"""Origin server that sends 100 Continue then a compressible 200 OK.
+
+Used to reproduce the crash in HttpTunnel::producer_run when compress.so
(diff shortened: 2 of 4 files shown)


JakeChampion:
I think AuTest 1of4 was a flakey fail

ezelkow1:
[approve ci autest 1]

masaori335:
[approve ci autest]

tests/gold_tests/pluginTest/compress/compress_100_continue_origin.py
```diff
@@ -0,0 +1,126 @@
...
+def handle_connection(conn, addr):
+    """Handle a single client connection."""
+    try:
+        conn.settimeout(10)
+        request = read_request(conn)
+        if request is None:
+            return
+
+        # Send 100 Continue immediately.
+        conn.sendall(b'HTTP/1.1 100 Continue\r\n\r\n')
```
masaori335, line 56: Thank you for adding this test case. But I guess Proxy-Verifier supports `100 Continue` support. ( we can merge this first and change it later )

@bneradt can we cover this case with proxy-verifier?

src/proxy/http/HttpSM.cc
```diff
@@ -6608,8 +6591,16 @@ HttpSM::perform_cache_write_action()
...
-      setup_cache_write_transfer(&cache_sm, server_entry->vc, &t_state.cache_info.object_store, client_response_hdr_bytes,
-                                 "cache write");
+
+      // Decide how many header bytes the cache-write consumer should skip.
+      // When a transform is present, setup_server_transfer_to_transform()
+      // creates the tunnel buffer with hdr_size=0, so the buffer contains
+      // only body data - no response headers to skip.  When no transform is
+      // present, setup_server_transfer() writes the response header into the
+      // buffer, so skip_bytes must equal client_response_hdr_bytes.
+      int64_t cache_write_skip = (transform_info.entry == nullptr) ? client_response_hdr_bytes : 0;
```
masaori335, line 6601: Review note: `transform_info.entry` is set by `HttpSM::do_post_transform_open()` or `HttpSM::do_transform_open()`.

bryancall:
[approve ci]

cmcfarlen:
Cherry-picked to 10.2.x

PR #12916, opened 2026-02-26
`thread_config` test - add visibility and some defensive improvements.

brbzull0:
With **AI** help:

### Problem

The `thread_config` autest intermittently fails on CI with exit code 1
(`[TS_MAIN]` process not found) even though ATS logs show it reached
"fully initialized". The script produces **no diagnostic output** on
failure, making CI failures impossible to investigate.

### Analysis

The CI failure is likely caused by ATS exiting under resource pressure
(12 ATS instances running, one with 100 threads). The `check_threads.py`
script also has defensive gaps:

1. **Race between `process_iter` and method calls**: The script passes
   `attrs=['name', 'cwd', 'threads']` to `psutil.process_iter()` (which
   pre-fetches data into `p.info`), but then calls `p.name()` and `p.cwd()`
   which make **fresh system calls** — creating a window where the process
   can exit between iteration and inspection.

2. **Unhandled `psutil.NoSuchProcess`**: `psutil.Process(t.id).name()`
   can throw if a thread exits between `p.threads()` and the per-thread
   lookup. With 100+ threads, this is not unlikely.

3. **No diagnostic output**: When the process isn't found, the script
   silently returns exit code 1 with nothing on stderr.

### Changes

- Use cached `p.info` dict instead of `p.name()`/`p.cwd()` method calls.
- Handle `psutil.NoSuchProcess` and `psutil.AccessDenied` in process
  iteration, thread listing, and per-thread name lookup.
- Add retry logic (3 attempts, 2s delay) **only** for the "process not
  found" case (exit code 1). Thread count mismatches (exit codes 2–11)
  return immediately and are never masked.
- Print diagnostic output on failure: the searched path and any
  `[TS_MAIN]` processes found with their cwds.

### Limitations

This does **not** address the root cause of the intermittent CI failure,
which is likely ATS exiting under resource pressure. These changes improve
resilience to transient issues and ensure future failures produce
actionable diagnostic output for further investigation.

### DIFF
--- a/tests/gold_tests/thread_config/check_threads.py
+++ b/tests/gold_tests/thread_config/check_threads.py
@@ -20,26 +20,45 @@
 import psutil
 import argparse
 import sys
+import time
 
 
 def count_threads(ts_path, etnet_threads, accept_threads, task_threads, aio_threads):
 
     for p in psutil.process_iter(['name', 'cwd', 'threads']):
 
+        # Use cached info from process_iter attrs to avoid race conditions
+        # where the process exits between iteration and inspection.
+        try:
+            proc_name = p.info.get('name', '')
+            proc_cwd = p.info.get('cwd', '')
+        except (psutil.NoSuchProcess, psutil.AccessDenied):
+            continue
+
         # Find the pid corresponding to the ats process we started in autest.
         # It needs to match the process name and the binary path.
         # If autest can expose the pid of the process this is not needed anymore.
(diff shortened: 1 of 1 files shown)


cmcfarlen (commented):
Can you address or resolve the copilot comments?

brbzull0:
https://github.com/apache/trafficserver/pull/12940

PR #12917, opened 2026-02-26
Fix nullptr crash in RecConfigOverrideFromEnvironment with runroot

brbzull0:
### TL;DR

`RecConfigOverrideFromEnvironment` returned a plain string, so callers could only guess whether a value was overridden by comparing strings — and had no way to tell if the source was an env var or runroot, this was also generating UB because of a stirng created out of a nullptr.
This change makes the function return the resolved value paired with a typed `RecConfigOverrideSource enum`, giving callers explicit knowledge of what happened and why. Now it also returns either the `env` value or the `runroot` value for some runroot records.

### A bit more details:

`RecConfigOverrideFromEnvironment()` previously returned a raw `const char *`
`nullptr` for the runroot case.  All three callers fed that into
`std::string`, which is undefined behaviour and crashed on newer toolchains.

This PR:

1. Changes the return type to `std::pair<std::string, RecConfigOverrideSource>`
   with a new enum (`NONE`, `ENV`, `RUNROOT`) so callers know exactly who
   overrode the value.

2. For runroot-managed path records, returns the **actual resolved Layout path**
   (e.g. `Layout::bindir`, `Layout::logdir`) instead of an empty string.  This
   means validation runs normally on every record — no special cases needed.

3. Replaces the `RecConfigOverrideFromRunroot()` strcmp chain with a `constexpr`
   lookup table that maps record names directly to `Layout::*` pointer-to-member.

4. Adds debug-level log messages at all three call sites:
   `'record_name' overridden with 'value' by environment variable|runroot`

5. Preserves `nullptr` for built-in defaults that are intentionally unset
   (e.g. `proxy.config.ssl.keylog_file`).  This is safe because all
   nullptr-default records are `RECD_STRING` and `RecDataSetFromString`
   handles nullptr for that type.

### DIFF
--- a/include/records/RecCore.h
+++ b/include/records/RecCore.h
@@ -78,7 +78,7 @@ enum class RecConfigOverrideSource {
   RUNROOT, ///< Overridden with the resolved Layout path because runroot manages this record.
 };
 
-/// Human-readable label for the override source (for logging).
+/// Label for the override source (for logging).
 constexpr const char *
 RecConfigOverrideSourceName(RecConfigOverrideSource src)
 {
@@ -87,8 +87,10 @@ RecConfigOverrideSourceName(RecConfigOverrideSource src)
     return "environment variable";
   case RecConfigOverrideSource::RUNROOT:
     return "runroot";
-  default:
+  case RecConfigOverrideSource::NONE:
     return "none";
+  default:
+    return "unknown";
   }
 }
 
--- a/include/records/RecCore.h
+++ b/include/records/RecCore.h
@@ -78,9 +78,24 @@ enum class RecConfigOverrideSource {
   RUNROOT, ///< Overridden with the resolved Layout path because runroot manages this record.
 };
 
-// Test whether the named configuration value is overridden by an environment
-// variable or by the runroot mechanism.  Returns the resolved value together
-// with the source that produced it.
+/// Human-readable label for the override source (for logging).
+constexpr const char *
+RecConfigOverrideSourceName(RecConfigOverrideSource src)
+{
+  switch (src) {
+  case RecConfigOverrideSource::ENV:
+    return "environment variable";
+  case RecConfigOverrideSource::RUNROOT:
+    return "runroot";
+  default:
+    return "none";
+  }
+}
+
+// Test whether the named configuration value is overridden by the execution
+// environment — either a PROXY_CONFIG_* environment variable or the runroot
+// mechanism.  Returns the resolved value together with the source that
+// produced it.
 std::pair<std::string, RecConfigOverrideSource> RecConfigOverrideFromEnvironment(const char *name, const char *value);
 
 //-------------------------------------------------------------------------
--- a/include/records/RecCore.h
+++ b/include/records/RecCore.h
@@ -25,6 +25,8 @@
 
 #include <functional>
 #include <optional>
+#include <string>
+#include <utility>
 
 #include "tscore/Diags.h"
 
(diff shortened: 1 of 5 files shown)


src/records/RecConfigParse.cc
```diff
@@ -123,7 +123,7 @@ RecConfigOverrideFromEnvironment(const char *name, const char *value)
   if (envval) {
     return envval;
   } else if (RecConfigOverrideFromRunroot(name)) {
-    return nullptr;
```
zwoop, line 126: Why this `else if` ? you return "value" regardless of the value of RecConfigOverrideFromRunroot(name)), no?

zwoop (changes requested):

zwoop (dismissed):
I think you should cleanup the else if, I think CoPilot said the same thing.

brbzull0:
> I think you should cleanup the else if, I think CoPilot said the same thing.


thanks for pointing this out. I think I may have introduced a side effect bug.  I'll ammend and put this back for review.

brbzull0:
[approve ci]

src/records/RecYAMLDecoder.cc
```diff
@@ -156,11 +156,12 @@ SetRecordFromYAMLNode(CfgNode const &field, swoc::Errata &errata)
...
-  RecSourceT  source    = (field_value == value_str ? REC_SOURCE_EXPLICIT : REC_SOURCE_ENV);
+  auto [value_str, override_source] = RecConfigOverrideFromEnvironment(record_name.c_str(), field_value.c_str());
+  RecSourceT source                 = (override_source == RecConfigOverrideSource::NONE) ? REC_SOURCE_EXPLICIT : REC_SOURCE_ENV;
 
-  if (source == REC_SOURCE_ENV) {
-    errata.note(ERRATA_DEBUG, "'{}' was override with '{}' using an env variable", record_name, value_str);
+  if (override_source != RecConfigOverrideSource::NONE) {
+    errata.note(ERRATA_DEBUG, "'{}' overridden with '{}' by {}", record_name, value_str,
+                RecConfigOverrideSourceName(override_source));
   }
```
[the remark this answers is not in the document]
brbzull0: The errata is the single structured channel for everything that happens during YAML parsing(overrides, defaults, warnings) each carrying its own severity.  
The logging message can be changed a bit eventually. Not the scope of this PR.

brbzull0:
> I think you should cleanup the else if, I think CoPilot said the same thing.

Ok, there was a deeper issue at the end. My original change was actually causing troubles.
I changed the approach now and updated the code and PR description with the details.
Thanks.

tests/gold_tests/records/records_runroot_precedence.test.py
```diff
@@ -0,0 +1,165 @@
+'''
+'''
```
zwoop: Why is this empty multi-line string here? Typo / leftovers ?

include/records/RecCore.h
```diff
@@ -69,9 +71,32 @@ std::string RecConfigReadConfigPath(const char *file_variable, const char *defau
...
+constexpr const char *
+RecConfigOverrideSourceName(RecConfigOverrideSource src)
+{
+  switch (src) {
+  case RecConfigOverrideSource::ENV:
+    return "environment variable";
+  case RecConfigOverrideSource::RUNROOT:
+    return "runroot";
+  default:
+    return "none";
```
zwoop, line 91: Can this happen? Should it be an assert ? Alternatively, instead of a default:, why not have NONE as the label, future proofing this?

brbzull0: yes, means the rec wasn't overriden.
can't put an assert, this is constexpr.

added a `NONE` case now. For code completeness added a default with `unknown`

tests/gold_tests/records/records_runroot_precedence.test.py
brbzull0: leftovers! thanks.

brbzull0:
[approve ci]

brbzull0:
[approve ci autest 0]

zwoop (approved):

PR #12922, opened 2026-02-27
Add AuTest for connect_attempts_max_retries_down_server

masaori335:
This is an effort to verify `connect_attempts_max_retries_down_server` behavior.

The doc for [`proxy.config.http.connect_attempts_max_retries_down_server`](https://docs.trafficserver.apache.org/admin-guide/files/records.yaml.en.html#proxy-config-http-connect-attempts-max-retries-down-server) says below.

> Maximum number of connection attempts Traffic Server can make while an origin is marked down per request.

Also doc of [`proxy.config.http.connect_attempts_max_retries`](https://docs.trafficserver.apache.org/admin-guide/files/records.yaml.en.html#proxy-config-http-connect-attempts-max-retries) says below

> Once the maximum number of retries is reached, the origin is marked down. After this, the setting `proxy.config.http.connect_attempts_max_retries_down_server` is used to limit the number of retry attempts to the known down origin.

However, I don't see ATS makes connection attempts against down servers. From my understanding, `HostDBRecord::select_best_http` doesn't return `HostDBInfo` which is in down state. How `HttpSM` can try to down servers?

### DIFF
--- a/tests/gold_tests/autest-site/ats_replay.test.ext
+++ b/tests/gold_tests/autest-site/ats_replay.test.ext
@@ -182,6 +182,8 @@ def ATSReplayTest(obj, replay_file: str):
             dns = tr.MakeDNServer(name, **process_config)
         else:
             dns = tr.MakeDNServer(name, default='127.0.0.1')
+        if 'records' in dns_config:
+            dns.addRecords(dns_config['records'])
 
     # Proxy Verifier Server configuration.
     if not 'server' in autest_config:
--- a/tests/gold_tests/dns/connect_attempts.test.py
+++ b/tests/gold_tests/dns/connect_attempts.test.py
@@ -1,0 +1,25 @@
+'''
+Verify Origin Server Connect Attempts Behavior
+'''
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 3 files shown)


masaori335:
I think I got what I'm missing. To test this config, ATS needs to go this line in `HttpTransact::handle_response_from_server`, but current test is trying connect attempt against un-exist servers and ATS doesn't comes here.
https://github.com/apache/trafficserver/blob/5b32738f2388090bdff3b1b6abd855efe63466ed/src/proxy/http/HttpTransact.cc#L3912

bryancall:
[approve ci ubuntu autest 0]

bryancall:
AuTest 0of4 is a real failure in the new `connect_attempts` test — not flaky. It failed on both the original run and a retrigger (build 35759). The proxy-verifier client exits with code 1, meaning the response from ATS didn't match the replay expectations. Server, DNS, and ATS processes all pass fine.

```
FAILED TESTS:
  - connect_attempts

Process: Default: Failed
     Reason: Returned Value 1 != 0
```

Ubuntu passed on retrigger — that one was flaky. Just this test remaining.

bryancall (changes requested):
The test approach needs rework. As you noted in your comment, connecting to unreachable IPs (0.0.0.1/0.0.0.2) exercises the connect-failure path, not the origin-down retry path through `HttpTransact::handle_response_from_server`. A few specific issues:

1. Transaction uuid=10 expects status 500, but `OriginDown()` returns 502 (BAD_GATEWAY) — see `HttpTransact.cc:925` which calls `build_error_response(s, HTTPStatus::BAD_GATEWAY, ...)`.

2. The comment at the top says "this replay file assumes that caching is enabled" but `enable_cache` is set to `false`.

3. The AuTest 0of4 failure is real (not flaky) — the proxy-verifier client gets a response mismatch.

The ATSReplay DNS records extension (`dns.addRecords`) is a nice addition to the harness — worth keeping when the test is reworked.

masaori335:
https://github.com/apache/trafficserver/pull/13102 fixes bugs and also covers this case.

PR #12924, opened 2026-02-27
compress plugin - use configured compression levels for gzip and brotli

JakeChampion:
Previously, gzip and brotli compression used hardcoded levels even when custom levels were configured via `HostConfiguration`

Split initialization into `data_alloc()` and `transform_init()` so
- gzip uses `hc->zlib_compression_level()`
- brotli uses `hc->brotli_compression_level()` and `hc->brotli_lgw_size()`

Also adds proper error handling if encoder initialization fails

### DIFF
--- a/plugins/compress/brotli_compress.cc
+++ b/plugins/compress/brotli_compress.cc
@@ -32,9 +32,6 @@
 
 namespace Brotli
 {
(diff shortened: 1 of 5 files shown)


maskit:
Overall looks good. Please address the comments from Copilot.

PR #12928, opened 2026-02-27
Add option 3 to connect down policy

vmamidi:
Add option 3 to connect down policy to count inactive connections as failures

### DIFF
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -4681,7 +4681,7 @@ HttpSM::track_connect_fail() const
   // Policy 3 additionally marks the server down on transaction inactive timeout,
   // even when had_connect_fail() is false (connect_result was cleared at CONNECTION_ALIVE).
   if (!retval && t_state.txn_conf->connect_down_policy == 3) {
-    retval = (t_state.current.state == HttpTransact::INACTIVE_TIMEOUT);
+    retval = (t_state.current.server->state == HttpTransact::INACTIVE_TIMEOUT);
   }
   return retval;
 }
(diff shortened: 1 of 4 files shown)


src/proxy/http/HttpSM.cc
```diff
@@ -4671,12 +4671,18 @@ HttpSM::track_connect_fail() const
   bool retval = false;
   if (t_state.current.server->had_connect_fail()) {
     // What does our policy say?
-    if (t_state.txn_conf->connect_down_policy == 2) { // Any connection error through TLS handshake
+    if (t_state.txn_conf->connect_down_policy == 2 ||
+        t_state.txn_conf->connect_down_policy == 3) { // Any connection error through TLS handshake
```
masaori335, line 4675: I prefer to make this config bit mask, it'll allow us to choose combination of policies when we have more policies.

bryancall (approved):
LGTM. Clean change — policy 3 correctly extends policy 2 by adding the INACTIVE_TIMEOUT check outside the had_connect_fail() guard, since connect_result is cleared when the connection transitions to CONNECTION_ALIVE. The test is well done with both positive (policy=3 marks down) and negative (policy=2 does not) cases. Docs are clear.

Re: Copilot comment about t_state.current.server->state vs t_state.current.state — both fields get set to INACTIVE_TIMEOUT via different handlers (tunnel handlers set server->state, while handle_server_setup_error and tunnel_handler_post_server set current.state). The check is correct as written.

Re: bitmask suggestion — the failure modes here are hierarchical, not orthogonal (TCP is required for TLS, and a successful connection is required for inactive timeout), so an enum/level approach is the right model. A bitmask would allow expressing meaningless combinations like TLS only without TCP.

cmcfarlen:
Cherry-picked to 10.2.x

PR #12938, opened 2026-03-04
ja4_fingerprint: Add an option to disable logging

maskit:
This adds `--nologging` option to ja4_fingerprint plugin. If it's specified, the plugin does not make a log file.

Current documentation says that the plugin does not have any options, but there is `--preserve` actually. I added documentation for it as well.

### DIFF
--- a/doc/admin-guide/plugins/ja4_fingerprint.en.rst
+++ b/doc/admin-guide/plugins/ja4_fingerprint.en.rst
@@ -78,13 +78,22 @@ Key Differences from JA3
 Plugin Configuration
 ====================
 
-The plugin operates as a global plugin and has no configuration options.
+The plugin operates as a global plugin.
 
 To enable the plugin, add the following line to :file:`plugin.config`::
 
     ja4_fingerprint.so
 
-No additional parameters are required or supported.
+
+.. option:: --preserve
+
+This option controls whether the plugin preserves any existing JA4 header. If the option is specified the plugin keep the header
+intact. If the option is not speficied, the plugins appends a generated fingerprint to the existing header value.
+
+.. option:: --nologging
+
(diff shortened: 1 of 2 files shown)


maskit:
I also started working on adding options to change header names, and now I feel like it may be nice if the default behavior is different.

How about output nothing by default, and have these 3 options to enable each output, instead of introducing `--nologging` that I suggested on this PR.
```
[--log <filename>] [--header=<header-name>] [--via-header <another-header-name>]
```

It would be an incompatible change though.

bneradt:
> I also started working on adding options to change header names, and now I feel like it may be nice if the default behavior is different.
> 
> How about output nothing by default, and have these 3 options to enable each output, instead of introducing `--nologging` that I suggested on this PR.
> 
> ```
> [--log <filename>] [--header=<header-name>] [--via-header <another-header-name>]
> ```
> 
> It would be an incompatible change though.

I'm fine with this. That sounds reasonable.

Just please keep it consistent with ja3_fingerprint as well.

maskit:
> Just please keep it consistent with ja3_fingerprint as well.

Ah, the design came from the ja3 version.... Although the difference from ja3 is noted on README, logging is disabled by default on ja3. The behavior is already inconsistent.

And the ja3 version does not have ways to specify filename and header names. There isn't a way to disable adding the "via" header either.

It doesn't seem like the design is good to me. Should we stick to it? Or should the ja3 version be updated (with incompatible changes) as well? This makes me want to add another ja4 plugin for people who don't use ja3.

maskit:
I don't like the existing ja3 and ja4 plugins' behavior. I made a new plugin.

maskit:
https://github.com/apache/trafficserver/pull/12995

PR #12950, opened 2026-03-09
hrw4u: Adds sandbox protection feature for the compiler

zwoop:
A new sandbox mechanism allows administrators to restrict which hrw4u language features are available at compile time. Policy is defined in a YAML configuration file that can deny specific sections, functions, conditions, operators, and language constructs such as break and variables. Violations are reported as compilation errors with a configurable policy message, and per-input sandbox overrides are also supported for testing purposes.

Co-author and ideas: Miles Libbey


doc/admin-guide/configuration/hrw4u.en.rst
```diff
@@ -494,6 +542,137 @@ Run with `--debug all` to trace:
...
+       conditions:  [ ... ]   # condition keys, e.g. geo.
+       operators:   [ ... ]   # operator keys, e.g. inbound.conn.dscp
+       language:    [ ... ]   # break, variables, in, else, elif
+
+All deny lists are optional. An empty or missing sandbox file permits everything.
+
+Sections
+--------
+
+The ``sections`` list accepts any of the HRW4U section names listed in the
```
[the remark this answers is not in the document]
zwoop: As intended.

zwoop:
I addressed all but one of CoPilot's review comments; One is a FP, the documentation is correct, it just looked in the wrong place.

zwoop:
Addressed the next round of CoPilot comments.

doc/admin-guide/configuration/hrw4u.en.rst
```diff
@@ -434,10 +434,58 @@ Groups
...
+
+.. code-block:: none
+
+   REMAP {
+     if inbound.req.X-Internal != "1" {
+       break;
+     }
+     # Only reached for internal requests
+     inbound.req.X-Debug = "on";
+   }
```
bneradt, line 578: "return;" be a more idiomatic keyword for this behavior. "break" is typically for switch/loops.

Having said that, I don't think that "break" new to this PR. Just documented here, if I'm understanding this correctly. If so, we can obviously address this in a separte PR.

bneradt (changes requested):
The feature seems like a really interesting idea. Could be helpful.

doc/admin-guide/configuration/hrw4u.en.rst
zwoop: Well it breaks all hooks, not just the current one. Possibly exit would make sense. I picked break ;)

bneradt (dismissed):

bneradt (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #12952, opened 2026-03-10
Document OCSP remap requirement

bneradt:
Document that ATS 10 and later fetch OCSP responses through
FetchSM instead of OpenSSL.

Add the remap.config requirement and a Digicert mapping example
to the OCSP stapling admin guide section so operators know they
must map the responder URL.

### DIFF
--- a/doc/release-notes/upgrading.en.rst
+++ b/doc/release-notes/upgrading.en.rst
@@ -182,6 +182,16 @@ The following :file:`records.yaml` changes have been made:
   allocating iobuffers and cache volumes from hugepages if configured in the
   system.
 - The records.yaml entry ``proxy.config.plugin.compiler_path`` has been added to specify an optional compiler tool path for compiling plugins.
+- If OCSP stapling is enabled, ATS 10 and later fetch OCSP responses through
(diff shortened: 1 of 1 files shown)


doc/admin-guide/security/index.en.rst
```diff
@@ -334,6 +334,14 @@ Authority Information Access field of the signed certificate. For example::
                 OCSP - URI:http://ocsp.digicert.com
                 CA Issuers - URI:http://cacerts.digicert.com/DigiCertSHA2SecureServerCA.crt
 
+Before ATS 10, OCSP fetches were handled by OpenSSL. Starting in ATS 10,
+|TS| performs the OCSP fetch itself through FetchSM, so the OCSP responder URL
+must also be reachable through :file:`remap.config`. For example, if the
```
maskit, line 339: This shouldn't be true.
https://github.com/apache/trafficserver/pull/10146

bneradt: Ah, good point. OK, I'll close this PR. The remap.config rules shouldn't be needed for OCSP fetch.

PR #12957, opened 2026-03-12
hrw4u: Add code coverage support for tests

zwoop:
- Adds some new tests, for additional coverage (e.g. debug runs)
- Eliminates some dead code discovered when there's no way to get the coverage over such code.

### DIFF
--- a/tools/hrw4u/Makefile
+++ b/tools/hrw4u/Makefile
@@ -104,7 +104,7 @@ INIT_HRW4U=$(PKG_DIR_HRW4U)/__init__.py
 INIT_U4WRH=$(PKG_DIR_U4WRH)/__init__.py
 INIT_LSP=$(PKG_DIR_LSP)/__init__.py
 
-.PHONY: all gen gen-fwd gen-inv copy-src test clean build package env setup-deps activate update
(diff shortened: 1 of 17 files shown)


tools/hrw4u/Makefile
```diff
@@ -169,6 +169,14 @@ $(PKG_DIR_LSP)/%: src/%
 test:
 	uv run pytest --tb=short tests
 
+coverage:
+	uv run pytest --cov --cov-report=term-missing --cov-report=html tests
+	@echo ""
+	@echo "HTML report: open htmlcov/index.html"
```
bneradt: I hit an issue with the new coverage workflow on a clean checkout: `make coverage` fails before pytest starts because the package is still installed from `build/`, but that directory does not exist until the generated sources have been created.

In my repro, `make coverage` failed with:

error in 'egg_base' option: 'build' does not exist or is not a directory

After generating the sources first, the suite passed. 

Could we make `coverage` ensure `build/` exists first (for example by depending on `gen`, or otherwise preparing the generated package layout), so the documented workflow works from a fresh checkout?

bneradt (changes requested):

bneradt (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #12972, opened 2026-03-13
Fix cache retry assert on ServerAddrSet

### LINKED ISSUE #12971: Assertion failure in set_cache_prepare_write_action_for_new_request on TSHttpTxnServerAddrSet retry
## Description

ATS crashes with a failed assertion when a plugin uses `TSHttpTxnServerAddrSet()` to set a server address, the connection fails, and ATS retries with caching enabled.

```
Fatal: src/proxy/http/HttpTransact.cc:3503: failed assertion `s->redirect_info.redirect_in_process`
```

The crash occurs in `set_cache_prepare_write_action_for_new_request()` which is called from `HandleCacheOpenReadMiss()`.

## Root Cause

The `TSHttpTxnServerAddrSet` retry logic added in #12733 (for issue #12611) does not account for the cache write lock state. The flow is:

1. Cache MISS → `OSDNSLookup` routes to `HandleCacheOpenReadMiss`
2. First pass: `set_cache_prepare_write_action_for_new_request` sets `action = PREPARE_TO_WRITE`
3. SM acquires the cache write lock → `write_lock_state = SUCCESS`, `action = WRITE`
4. Origin connection **fails**
5. Retry logic fires (line 3936): detects `os_addr_style == USE_API`, resets DNS state to `TRY_DEFAULT`, calls `CallOSDNSLookup()`
6. DNS resolves → `OSDNSLookup` → `cache_lookup_result == MISS` → routes back to `HandleCacheOpenReadMiss`
7. Second pass: `write_lock_state` is already `SUCCESS` → `set_cache_prepare_write_action_for_new_request` hits the assertion at line 3503 because `redirect_in_process == false`

The assertion only expects `write_lock_state == SUCCESS` during redirects, but the connection retry is a second legitimate scenario where the write lock is already held.

## Why Tests Didn't Catch This

The autest added in #12733 uses `enable_cache=False`. With caching disabled, `OSDNSLookup` routes to `LookupSkipOpenServer` instead of `HandleCacheOpenReadMiss`, so the cache write lock path is never exercised.

## Fix

The assertion in `set_cache_prepare_write_action_for_new_request` needs to be relaxed to also accept the retry case. When `api_server_addr_set_retried` is true, the cache key is the same (same URL, different server address), so reusing the write lock is safe:

```cpp
ink_release_assert(s->redirect_info.redirect_in_process || s->api_server_addr_set_retried);
```

The existing autest should also add a test run with caching enabled.

## Stack Trace

```
HttpTransact::set_cache_prepare_write_action_for_new_request(HttpTransact::State*)
HttpTransact::HandleCacheOpenReadMiss(HttpTransact::State*)
HttpSM::call_transact_and_set_next_state(void (*)(HttpTransact::State*))
HttpSM::state_api_callout(int, void*)
HttpSM::state_api_callback(int, void*)
TSHttpTxnReenable(tsapi_httptxn*, TSEvent)
<plugin>.so
INKContInternal::handle_event(int, void*)
APIHook::invoke(int, void*) const
HttpSM::state_api_callout(int, void*)
HttpSM::state_hostdb_lookup(int, void*)
```

## Key State at Crash

```
s->redirect_info.redirect_in_process = false
s->cache_info.write_lock_state = SUCCESS
s->cache_info.action = WRITE
s->api_server_addr_set_retried = true
s->response_error = CONNECTION_OPEN_FAILED
s->transact_return_point = HandleCacheOpenReadMiss
```

bneradt:
A TSHttpTxnServerAddrSet retry can re-enter the cache miss path after ATS already holds a cache write lock for the same request. The redirect-only prepared-write reuse logic then asserts because this is not actually a redirect.

Preserve the existing cache write lock and retry context when the retry returns through HandleCacheOpenReadMiss. Add an autest that fails on the unpatched code with the redirect_in_process assertion and passes with this fix.

Fixes: #12971

### DIFF
--- a/src/proxy/http/HttpTransact.cc
+++ b/src/proxy/http/HttpTransact.cc
@@ -3403,6 +3403,12 @@ HttpTransact::HandleCacheOpenReadMiss(State *s)
     TxnDbg(dbg_ctl_http_trans, "READ_RETRY cache read failed, bypassing cache");
     s->cache_info.stale_fallback = nullptr; // Clear unused fallback
     s->cache_info.action         = CacheAction_t::NO_ACTION;
+  } else if (s->cache_info.write_lock_state == CacheWriteLock_t::SUCCESS && s->cache_info.action == CacheAction_t::WRITE) {
+    // Origin retry paths such as TSHttpTxnServerAddrSet() can loop back through
+    // HandleCacheOpenReadMiss after we already own the cache write lock. This
+    // is still the same request, so keep the existing write lock instead of
+    // re-preparing a write as if this were a redirect or a new cache miss.
+    TxnDbg(dbg_ctl_http_trans, "Reusing existing cache write lock while retrying origin selection");
   } else {
     HttpTransact::set_cache_prepare_write_action_for_new_request(s);
   }
--- a/tests/gold_tests/pluginTest/tsapi/test_TSHttpTxnServerAddrSet_retry.test.py
+++ b/tests/gold_tests/pluginTest/tsapi/test_TSHttpTxnServerAddrSet_retry.test.py
@@ -14,31 +14,29 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 '''
-Test to reproduce issue #12611 - retry fails when TSHttpTxnServerAddrSet is used
+Tests for TSHttpTxnServerAddrSet() retry behavior with and without cache enabled.
 '''
 
 import os
 from ports import get_port
 
 Test.Summary = '''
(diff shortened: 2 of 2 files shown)


bryancall (dismissed):

bryancall (approved):

cmcfarlen:
Cherry-picked to 10.2.x

bneradt:
10.1.x via: https://github.com/apache/trafficserver/pull/13155

cmcfarlen:

PR #12985, opened 2026-03-17
Check valid parent selection hash string

### LINKED ISSUE #12973: [BUG] Multiple Intra-Object Overflows in ParentSelection.cc (hash_string)
I am reporting two related memory safety vulnerabilities (intra-object overflows) in src/proxy/ParentSelection.cc. These issues involve an inconsistent length validation
  when parsing parent configuration records, leading to potential buffer overflows in the hash_string field.

  1. Vulnerability Analysis
  In ParentSelection.cc, the code parses parent server strings from the configuration. It correctly validates the length of the hostname field but completely fails to
  validate the length of the hash_string field before performing a memcpy.

  Data Structure (ParentSelection.h):
   1 struct pRecord : ATSConsistentHashNode {
   2     char hostname[MAXDNAME + 1];    // 1025 bytes
   3     ...
   4     char hash_string[MAXDNAME + 1]; // 1025 bytes (Target of overflow)
   5 };

  The Logic Flaw:
  The code includes a guard for hostname (line 552):
  if (tmp - current > MAXDNAME) { ... errPtr = "Parent hostname is too long"; ... }
  However, later in the same function, it performs a memcpy into hash_string using strlen(tmp3) without any bounds check.

  2. Vulnerable Locations
  Location #1: Primary Parents (src/proxy/ParentSelection.cc:572)
   1 if (tmp3) {
   2     // strlen(tmp3) is UNCHECKED and can exceed MAXDNAME
   3     memcpy(this->parents[i].hash_string, tmp3 + 1, strlen(tmp3));
   4     this->parents[i].name = this->parents[i].hash_string;
   5 }

  Location #2: Secondary Parents (src/proxy/ParentSelection.cc:587)
   1 if (tmp3) {
   2     // Symmetrical flaw for secondary_parents array
   3     memcpy(this->secondary_parents[i].hash_string, tmp3 + 1, strlen(tmp3));
   4     this->secondary_parents[i].name = this->secondary_parents[i].hash_string;
   5 }

  3. Impact
  An administrator or an attacker capable of influencing the parent.config (or other parent selection configuration) can provide a string that exceeds MAXDNAME (1024 bytes).
  This results in:
   1. Memory Corruption: memcpy will overflow the hash_string buffer in the pRecord structure.
   2. Heap Overwrite: Since parents is an array of pRecord objects, an overflow in parents[i] will overwrite the metadata or member data of parents[i+1] or adjacent heap
      allocations.
   3. Instability: This corruption can lead to inconsistent consistent-hashing results, crashes (DoS), or potentially arbitrary code execution if critical pointers within
      the ATSConsistentHashNode are corrupted.

  4. Suggested Fix
  Apply a length check to tmp3 similar to the one used for the hostname before performing the memcpy:
   1 if (tmp3) {
   2     size_t hash_len = strlen(tmp3);
   3     if (hash_len > MAXDNAME) {
   4         // Handle error: hash string too long
   5     }
   6     memcpy(this->parents[i].hash_string, tmp3 + 1, hash_len);
   7 }

Best regards,
Pengpeng Hou
ISCAS

serrislew:
Resolves #12973 by checking parent hash string length

### DIFF
--- a/src/proxy/ParentSelection.cc
+++ b/src/proxy/ParentSelection.cc
@@ -557,7 +557,7 @@ ParentRecord::ProcessParents(char *val, bool isPrimary)
       errPtr = "Parent string is empty";
       goto MERROR;
     }
-    if (tmp3 && strlen(tmp3) > MAXDNAME) {
+    if (tmp3 && strlen(tmp3 + 1) > MAXDNAME) {
       errPtr = "Parent hash string is too long";
       goto MERROR;
     }
@@ -1955,12 +1955,12 @@ EXCLUSIVE_REGRESSION_TEST(PARENTSELECTION)(RegressionTest * /* t ATS_UNUSED */, 
   RE(verify(result, ParentResultType::SPECIFIED, "minnie", 80), 213);
 
   // Test 214
-  // Overlong hash_string (MAXDNAME chars) should be rejected by ProcessParents.
+  // Overlong hash_string (exceeding MAXDNAME chars) should be rejected by ProcessParents.
   // The entry is discarded so findParent returns DIRECT.
   {
     tbl[0] = '\0';
     ST(214);
-    std::string long_hash(MAXDNAME, 'a');
+    std::string long_hash(MAXDNAME + 1, 'a');
     std::string cfg = "dest_domain=. parent=host:80&" + long_hash + " round_robin=consistent_hash go_direct=true\n";
     ink_strlcpy(tbl, cfg.c_str(), sizeof(tbl));
     REBUILD;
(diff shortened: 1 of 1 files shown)


src/proxy/ParentSelection.cc
```diff
@@ -556,6 +557,11 @@ ParentRecord::ProcessParents(char *val, bool isPrimary)
       errPtr = "Parent string is empty";
       goto MERROR;
     }
+    if (tmp3 && strlen(tmp3) > MAXDNAME) {
+      errPtr = "Parent hash string is too long";
+      goto MERROR;
```
bryancall, line 562: `tmp3` points at the `&` separator, so `strlen(tmp3)` includes it. The destination buffer is `hash_string[MAXDNAME + 1]`, which can hold a payload of up to `MAXDNAME` characters plus the NUL terminator. The current check rejects a hash of exactly `MAXDNAME` characters even though it fits.

This should check the payload length, not the separator-inclusive length:

```cpp
// tmp3 points at the '&' separator; skip it to measure just the hash payload.
// The destination buffer is hash_string[MAXDNAME + 1], so the payload must fit in MAXDNAME.
if (tmp3 && strlen(tmp3 + 1) > MAXDNAME) {
```

And the test boundaries need to shift accordingly:

```cpp
// Test 214 — hash string exceeding MAXDNAME should be rejected
std::string long_hash(MAXDNAME + 1, 'a');

// Test 215 — hash string of exactly MAXDNAME should be accepted
std::string max_hash(MAXDNAME, 'b');
```

cmcfarlen:
Cherry-picked to 10.2.x

PR #12986, opened 2026-03-17
Align AuTests with latest proxy-verifier checks

bneradt:
The latest proxy-verifier now fails a run when a verifier server or
client is given proxy-side checks for traffic that ATS never produces.
Most tests only needed stale proxy-request or proxy-response nodes
removed, but the shared replay cases below need server-specific files
so each verifier only owns traffic it can actually observe.

- disable_pristine_host_hdr_server_canary_false.replay.yaml keeps
  the canary server on uuid 1 when pristine_host_hdr stays enabled.
- disable_pristine_host_hdr_server_canary_true.replay.yaml keeps the
  canary server on uuid 1 when the Host header is rewritten.
- disable_pristine_host_hdr_server_stable.replay.yaml keeps the
  stable server on uuid 2, which the canary server never receives.
- escalate_original_server_default.replay.yaml keeps only the
  requests that the default-mode origin server really handles.
- escalate_failover_server_default.replay.yaml keeps only the GET
  requests that default-mode escalation sends to failover.
- escalate_original_server_non_get.replay.yaml keeps the origin-side
  subset when --escalate-non-get-methods is enabled.
- escalate_failover_server_non_get.replay.yaml keeps the failover
  subset, including the escalated HEAD request in that mode.
- ja4_fingerprint_basic_server.replay.yaml limits the non-preserve
  test to its one real request instead of preserve-only checks.
- traffic_dump_server.yaml keeps origin verification only for the
  sessions that really reach origin in the main traffic_dump test.
- traffic_dump_ip_filter_server.yaml keeps only the /one request
  used by the traffic_dump IP filter test.

These per-server replays preserve fallback server-response directives
and client-side coverage while dropping only the proxy-side checks
that latest proxy-verifier now correctly reports as unprocessed.

### DIFF
--- a/tests/gold_tests/cache/replay/cache-read-retry.replay.yaml
+++ b/tests/gold_tests/cache/replay/cache-read-retry.replay.yaml
@@ -136,11 +136,6 @@ sessions:
               - [Host, example.com]
               - [X-Request, second-exhausted]
 
-        proxy-request:
-          headers:
-            fields:
-              - [X-Request, {value: 'second-exhausted', as: equal}]
-
         # This request should NOT reach the server if read-while-writer works
         # It should be served from Request 1's response
         server-response:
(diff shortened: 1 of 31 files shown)


zwoop (approved):
Looks good, ship it.

bneradt:
10.2.x backport:
https://github.com/apache/trafficserver/pull/13020

cmcfarlen:
Added to milestone 10.2.0 via #13020

PR #12987, opened 2026-03-17
Update cachedDuplicateHeaders for latest proxy-verifier

bneradt:
The latest proxy-verifier combines duplicate header values during
equality checks unless the replay uses duplicate-aware
verification. cachedDuplicateHeaders still expected the older
rendering, so the AuTest failed even though ATS behavior had not
changed.

Update the replay expectation to match the current duplicate
header matching behavior. This keeps the test validating cached
header handling against the latest proxy-verifier semantics.


bneradt:
I'm going to include this directly in the PV upgrade PR: #12990

zwoop:
Remember to remove Milestones on closed PRs :)

PR #12989, opened 2026-03-17
hrw4u/header_rewrite: Add session-scope state variables

zwoop:
Add SESSION-FLAG, SESSION-INT8, and SESSION-INT16 conditions and their corresponding set-session-flag, set-session-int8, and set-session-int16 operators to the header_rewrite plugin. These mirror the existing transaction-scoped state variables but persist across keep-alive requests on the same connection, using a TS_USER_ARGS_SSN slot. The condition and operator classes are parameterized with a TSUserArgType scope argument to avoid code duplication. The hrw4u transpiler adds a SESSION_VARS section for declaring session-scoped variables, and the reverse transpiler handles both scopes. Documentation and tests are included.

Co-Authored-By: Craig Taylor


bneradt:
Codex review:

I have two follow-up concerns after reviewing and validating this PR.

1. `SESSION_VARS` does not look wired into the LSP declaration scanner yet. `tools/hrw4u/src/lsp/strings.py` still only enters declaration mode for `VARS { ... }`, so once users adopt `SESSION_VARS`, session-scoped variables appear likely to lose normal hover/type metadata in `hrw4u-lsp`. I have not done a live LSP repro yet, so I am flagging this as a tooling concern rather than a confirmed runtime bug.

2. `hrw4u-kg` skips `SESSION_VARS`. In `tools/hrw4u/src/kg_visitor.py`, `visitSection()` handles `ctx.varSection()` but not `ctx.sessionVarSection()`, so the new declaration block is ignored by knowledge-graph extraction. I reproduced this in the `fix_tickets_4` container by running `hrw4u-kg` on `tools/hrw4u/tests/data/vars/session_assign.input.txt`: the output contained the `VARS` block and `bool_0`, but dropped the `SESSION_VARS` block and its session variables entirely.

zwoop:
I've addressed Brian's comments.

zwoop:
[approve ci autest]

bneradt (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #12983, opened 2026-03-17
Merge 11-Dev back into master

cmcfarlen:

PR #12991, opened 2026-03-18
Fix ats_replay.test.ext to use ssl_multicert_yaml

bryancall:
## Summary

- The `ssl_multicert.config` → `ssl_multicert.yaml` migration (#12755) updated `trafficserver.test.ext` and `traffic_replay.test.ext` but missed `ats_replay.test.ext`
- This causes an `AttributeError` exception for any `ATSReplayTest` using `enable_tls: true` (e.g., the `chunked_encoding` `large_chunked` sub-test), which is the root cause of the AuTest 0of4 failure on #12983
- Updates `ats_replay.test.ext` to use `ssl_multicert_yaml` attribute and YAML format for the default cert configuration

```
AttributeError: 'Disk' object has no attribute 'ssl_multicert_config'. Did you mean: 'ssl_multicert_yaml'?
```

**Note:** This fix depends on #12983 (merge of 11-Dev into master) which brings in the `ssl_multicert.yaml` migration. It should be merged after or included in that PR.

## Test plan

- [x] Reproduced the exception in Fedora 42 CI Docker image (`ci.trafficserver.apache.org/ats/fedora:42`)
- [x] Verified `chunked_encoding` test passes (0 exceptions, 2 passed) with the fix
- [x] Verified `cache-auth` test still passes with the fix

### DIFF
--- a/tests/gold_tests/autest-site/ats_replay.test.ext
+++ b/tests/gold_tests/autest-site/ats_replay.test.ext
@@ -43,11 +43,11 @@ def configure_ats(obj: 'TestRun', server: 'Process', ats_config: dict, dns: Opti
     # TLS configs
     enable_tls = process_config.get('enable_tls', False)
     if enable_tls:
-        # Configure ssl_multicert.config if specified.
-        ssl_multicert_config = ats_config.get('ssl_multicert_config', [])
+        # Configure ssl_multicert.yaml if specified.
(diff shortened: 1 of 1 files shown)


bryancall:
Closing - this fix needs to be included in #12983 directly since the issue only manifests in the merge result. Will comment on #12983 with the fix.

PR #12992, opened 2026-03-18
Add TLS support to ats_replay.test.ext using ssl_multicert.yaml

bryancall:
## Summary

- The `ssl_multicert.config` → `ssl_multicert.yaml` migration (#12755) removed the TLS configuration block from `ats_replay.test.ext`, but `ATSReplayTest` still needs it for tests that use `enable_tls: true`
- Without this, when master's `chunked_encoding` test (which includes `large_chunked.replay.yaml` with `enable_tls: true`) is merged in via #12983, it hits `AttributeError: 'Disk' object has no attribute 'ssl_multicert_config'`
- Adds the TLS configuration back using the new `ssl_multicert.yaml` YAML format

## Test plan

- [x] Reproduced the exception in Fedora 42 CI Docker image
- [x] Verified `chunked_encoding` test passes (0 exceptions) with the fix
- [x] Verified `cache-auth` test still passes with the fix

### DIFF
--- a/tests/gold_tests/autest-site/ats_replay.test.ext
+++ b/tests/gold_tests/autest-site/ats_replay.test.ext
@@ -38,6 +38,37 @@ def configure_ats(obj: 'TestRun', server: 'Process', ats_config: dict, dns: Opti
     records_config = ats_config.get('records_config', {})
     ts.Disk.records_config.update(records_config)
 
+    # TLS configs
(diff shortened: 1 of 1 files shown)


bryancall:
[approve ci autest 0]

bryancall:
[approve ci autest 0]

PR #12994, opened 2026-03-18
Clarify hostdb host_file.path reload timing and tip usage

bryancall:
## Summary
- clarify that `proxy.config.hostdb.host_file.path` is checked on the periodic host file interval, not immediately
- include the controlling setting name and default value (`proxy.config.hostdb.host_file.interval`, default `86400` seconds)
- add a properly formatted `.. tip::` block suggesting temporary interval reduction for faster validation

## Test plan
- [x] Build docs locally and render the updated section
- [x] Verify the `host_file.path` section renders the tip as a real admonition block
- [x] Confirm no code behavior changes (docs-only update)


zwoop (approved):
One wonders why our default here is 24h, but not part of this pr.

PR #13001, opened 2026-03-19
AGENTS.md has to be at the root of the repo

bneradt:
I thought Codex would use .codex/AGENTS.md, but it actually only checks for the repo root AGENTS.md.


cmcfarlen:
Cherry-picked to 10.2.x

PR #13003, opened 2026-03-19
Proxy Verifier: use concise stack protocol specification

bneradt:
Proxy Verifier v3.0.0 has a more concise `stack` configurable for `protocol` specification. This makes use of that over the more verbose full `protocol` sequence.


ezelkow1 (approved):

cmcfarlen:
Added to milestone 10.2.0 via #13016

PR #13008, opened 2026-03-20
cmake: limit GENERAL_NAME bssl probe

bneradt:
Only run the GENERAL_NAME namespace probe for BoringSSL-family libraries. On plain OpenSSL builds, the probe can still compile and incorrectly enable HAVE_GENERAL_NAME_IN_BSSL_NAMESPACE.

Force the cache entry off for non-BoringSSL builds so ATS stays on the OpenSSL code path and does not emit bssl::GENERAL_NAME references at link time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -593,19 +593,28 @@ check_cxx_source_compiles(
   }"
   HAVE_CRYPTO_EX_DUP_TYPE1
 )
-check_cxx_source_compiles(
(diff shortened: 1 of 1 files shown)


maskit (approved):
It's so weird, but the change makes sense because the probe is just for BoringSSL family.

cmcfarlen:
Cherry-picked to 10.2.x

cmcfarlen:

Issue #12960, opened 2026-03-12
ATS returns 411 for POST with no body instead of treating body length as zero


bryancall:
## Summary

ATS returns `411 Content Length Required` for a POST request that has neither `Content-Length` nor `Transfer-Encoding` headers. Per RFC 9110 Section 6.4.1, a request with no `Content-Length` and no `Transfer-Encoding` has a message body length of zero — it should not be rejected.

## How to reproduce

```bash
curl -v -X POST http://ats-host/some-endpoint
```

ATS responds:
```
HTTP/1.1 411 Content Length Required
```

Adding `Content-Length: 0` makes it work:
```bash
curl -v -X POST -H "Content-Length: 0" http://ats-host/some-endpoint
```

## Expected behavior

ATS should accept the request and treat the body length as zero, the same way Nginx and Apache httpd do.

## Context

This affects browser `fetch()` calls with `method: "POST"` and no body — browsers send neither `Content-Length` nor `Transfer-Encoding` in this case. Common patterns like snooze/toggle/action endpoints that take no body are broken when proxied through ATS.

## Spec reference

RFC 9110 Section 6.4.1:
> If a message is received without Transfer-Encoding and with an invalid or missing Content-Length, and the message's semantics could include a payload body, the recipient might still be able to determine the body length by other means [...] A server MAY reject a request that contains a message body but not a Content-Length by responding with 411 (Length Required).

The key phrase is "contains a message body" — a POST with no body at all does not contain a message body, so 411 should not apply.

uma-1510:
Hey, I am intrested in working on this issue . can you assign it to me?

bryancall:
@uma-1510 If you want to work on it go for it

bneradt:
I think the `411 Content Length Required` is actually reasonable behavior. And those RFC references look hallucinated. I cannot find them in the RFC 9110.

Having said that, having autest coverage for this is reasonable. I'll add that with a PR and close this issue.

bryancall:
The RFC citations in the issue description are indeed wrong (likely LLM-generated), but the underlying bug is valid.

The correct reference is **RFC 9112 Section 6.3** (Message Body Length), which has two relevant rules:

1. *"If this is a request message and none of the above are true, then the message body length is zero (no message body is present)."*
2. *"A server MAY reject a request that contains a message body but not a Content-Length by responding with 411 (Length Required)."*

[...]

PR #13015, opened 2026-03-23
Add 204 and 308 to heuristically cacheable status codes

bryancall:
### Problem

RFC 9110 Section 15.1 defines 204 (No Content) and 308 (Permanent Redirect) as heuristically cacheable by default, but they were missing from the allowlist in `is_response_cacheable()`. This meant these responses would only be cached via negative caching configuration rather than being treated as cacheable by default like their counterparts (200 and 301).

### Changes

* **Add 204 and 308 to heuristic caching allowlist** -- `HttpTransact::is_response_cacheable()` now includes `HTTP_STATUS_NO_CONTENT` and `HTTP_STATUS_PERMANENT_REDIRECT` alongside existing cacheable codes (200, 203, 300, 301, 410)
* **Add RFC 9110 heuristic caching autest** -- validates that all seven heuristically cacheable status codes (200, 203, 204, 300, 301, 308, 410) are cached when only `Last-Modified` is present, and that non-cacheable codes (302, 307, 400, 403) are not cached. Negative caching disabled to isolate the heuristic path.

### Testing

- [x] New autest: `heuristic_caching` validates all RFC 9110 cacheable codes
- [x] All existing autests pass
- [x] Full CI matrix green

### DIFF
--- a/src/proxy/http/HttpTransact.cc
+++ b/src/proxy/http/HttpTransact.cc
@@ -6564,7 +6564,8 @@ HttpTransact::is_response_cacheable(State *s, HTTPHdr *request, HTTPHdr *respons
   }
 
   if ((response_code == HTTPStatus::OK) || (response_code == HTTPStatus::NOT_MODIFIED) ||
-      (response_code == HTTPStatus::NON_AUTHORITATIVE_INFORMATION) || (response_code == HTTPStatus::MOVED_PERMANENTLY) ||
+      (response_code == HTTPStatus::NON_AUTHORITATIVE_INFORMATION) || (response_code == HTTPStatus::NO_CONTENT) ||
+      (response_code == HTTPStatus::MOVED_PERMANENTLY) || (response_code == HTTPStatus::PERMANENT_REDIRECT) ||
       (response_code == HTTPStatus::MULTIPLE_CHOICES) || (response_code == HTTPStatus::GONE)) {
     TxnDbg(dbg_ctl_http_trans, "YES response code seems fine");
     return true;
--- a/tests/gold_tests/cache/cache-heuristic-status.test.py
+++ b/tests/gold_tests/cache/cache-heuristic-status.test.py
@@ -1,0 +1,28 @@
+'''
+Test heuristic caching of status codes per RFC 9110 Section 15.1.
+'''
+#  Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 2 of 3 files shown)


bryancall:
## RFC 9110 Heuristic Caching Gap Analysis

While adding 204 and 308, I checked the full list of heuristically cacheable status codes from [RFC 9110 Section 15.1](https://www.rfc-editor.org/rfc/rfc9110#section-15.1). ATS is still missing four codes from the allowlist in `is_response_cacheable()`:

| Status | RFC 9110 | ATS (after this PR) |
|---|---|---|
| 200 OK | Cacheable | In allowlist |
| 203 Non-Authoritative Information | Cacheable | In allowlist |
| 204 No Content | Cacheable | In allowlist (this PR) |
| 206 Partial Content | Cacheable | Intentionally excluded -- handled by range caching logic |
| 300 Multiple Choices | Cacheable | In allowlist |
| 301 Moved Permanently | Cacheable | In allowlist |
| 308 Permanent Redirect | Cacheable | In allowlist (this PR) |
| **404 Not Found** | Cacheable | **Missing** -- negative caching only |
| **405 Method Not Allowed** | Cacheable | **Missing** -- negative caching only |
| 410 Gone | Cacheable | In allowlist |
| **414 URI Too Long** | Cacheable | **Missing** -- negative caching only |
| **501 Not Implemented** | Cacheable | **Missing** -- negative caching only |

The four missing codes (404, 405, 414, 501) are currently only cached when `proxy.config.http.negative_caching_enabled` is turned on, and they use a flat TTL (`negative_caching_lifetime`, default 1800s) instead of proper LM-factor heuristic freshness.

Adding these to the heuristic allowlist would be a behavior change for existing deployments -- particularly for 404, where many operators rely on the current default of *not* caching 404s without explicit negative caching configuration. Keeping this PR scoped to just 204 and 308, but will open a separate issue to track the remaining gaps.

bryancall:
@cmcfarlen Friendly ping — this is a small one-liner (milestone version fix). Would appreciate a quick review when you get a chance. Thanks Chris!

moonchen (approved):

bneradt:
> Fixes: #12960 (partial -- 308 caching aspect)

Correct me if I'm wrong, but I don't think this is referencing the correct issue.

PR #13017, opened 2026-03-23
autest: print worker diagnostics for exceptions

bneradt:
Parallel autest runs can fail with worker exceptions that are not tied
to a specific failed test. In CI that currently leaves only the summary
counts, which makes the failure hard to diagnose.

Print worker diagnostics whenever a worker reports exceptions or exits
non-zero without attributing the problem to failed tests. Also accept
plural summary keys while parsing autest output.


maskit (dismissed):
The comment from copilot _sounds_ reasonable but I think this works for the immediate issue (the exception) as is.

maskit (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #13018, opened 2026-03-23
Add log field fallbacks

bneradt:
Allow custom log formats to fall back between header fields with
`??` so configs can log a secondary header when the primary one is
absent. This keeps the feature in the existing `%<...>` syntax and
preserves the current distinction between missing and empty values.

Store fallback chains as a single log field so the selected header
still uses normal escaping and its own slice settings. Update the
logging admin guide and extend the logging AuTest coverage for
primary, secondary, and missing header cases.

### DIFF
--- a/doc/admin-guide/logging/formatting.en.rst
+++ b/doc/admin-guide/logging/formatting.en.rst
@@ -314,8 +314,36 @@ prior to the log field's name, as so::
     Format = '%<{User-agent}cqh>'
 
 The above would insert the User Agent string from the client request headers
-into your log entry (or a blank string if no such header was present, or it did
-not contain a value).
+into your log entry (or ``-`` if no such header was present).
+
+Header fields can also be chained with a fallback operator, ``??``, when you want
+the log to use the first header that exists among n headers. For example::
+
+    Format = '%<{x-primary-id}cqh??{x-secondary-id}cqh??{x-tertiary-id}cqh>'
+
+|TS| evaluates the candidates from left to right and logs the first header that
+exists. If none of the headers exist, |TS| logs ``-`` by default. A header that
+exists but has an empty value is considered present, so |TS| logs the empty
+value instead of falling back. So in the example above, the value of
+x-primary-id of the client request is logged if it exists, otherwise the value
+of x-secondary-id is logged if it exists, otherwise ``-`` is logged if neither
+of the headers is present.
+
+You can provide an explicit quoted default literal as the final term in the
+chain to use instead of the default ``-`` literal::
+
+    Format = '%<{x-primary-id}cqh??{x-secondary-id}cqh??"missing-id">'
+
+If none of the headers exist, |TS| logs the default literal instead,
+``missing-id`` in this case. The default literal must be quoted and must be the
+last term in the chain.
+
+Note that this fallback chain feature is only supported for header field
+expressions (``cqh``, ``pqh``, etc., see the following table).
+
+Slices apply to each candidate in the fallback chain individually::
+
+    Format = '%<{x-primary-id}cqh[0:8]??{x-secondary-id}cqh[0:16]>'
 
 ===== ====================== ==================================================
 Field Source                 Description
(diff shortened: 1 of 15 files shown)


doc/admin-guide/logging/formatting.en.rst
```diff
@@ -314,8 +314,27 @@ prior to the log field's name, as so::
...
+
+|TS| evaluates the candidates from left to right and logs the first header that
+exists. If none of the headers exist, |TS| logs ``-``. A header that exists but
+has an empty value is considered present, so |TS| logs the empty value instead
+of falling back. So in the example above, the value of x-primary-id of the
+client request is logged if it exists, otherwise the value of x-secondary-id is
+logged if it exists, otherwise ``-`` is logged if neither of the headers is
+present.
+
+Note that this fallback chain feature is only supported for header field
```
serrislew: So `%<{x-primary-id}cqh?:Foo>` wouldn't be supported? I guess you can do a workaround to make an internal header with value "Foo" and then use that instead? `%<{x-primary-id}cqh?:{@Foo}>`

src/proxy/logging/LogFormat.cc
```diff
@@ -55,6 +58,94 @@ DbgCtl dbg_ctl_log_format{"log-format"};
...
+}
+
+LogField *
+parse_header_fallback_field(char *symbol)
+{
+  std::string                        original_symbol(symbol);
+  std::vector<LogField::HeaderField> header_fields;
+  char                              *candidate = symbol;
+
+  while (candidate != nullptr) {
```
serrislew: Does this mean there can be multiple fallback candidates?

bneradt: Yes! They can be chained n times:

```
%<{x-primary-id}cqh?:{x-secondary-id}cqh?:{x-tertiary-id}cqh>
```

There is, btw, a dev mailing list email about this if you want to chime in there.

doc/admin-guide/logging/formatting.en.rst
bneradt: We can add a different hard-coded fallback. Would that be useful though? Currently the final fallback is the default `-`, when would some other hard-coded fallback have value over that? Maybe there is a use case for that which I'm not thinking of.

serrislew: I guess it would just make the default customizable

bneradt: Sounds good. Done.

zwoop (dismissed):
Yeah, looks good. I'm ok deferring "general log tag support" for future additions. One could imagine for example a logging of X-Client-IP with a fallback to the connection remote IP log field.

like

```
Client-IP:%<{X-Client-IP}cqh??chi>
```

or some such.

doc/admin-guide/logging/formatting.en.rst
```diff
@@ -314,8 +314,54 @@ prior to the log field's name, as so::
...
+|TS| evaluates the candidates from left to right and logs the first header that
+exists. If none of the headers exist, |TS| logs ``-`` by default. A header that
+exists but has an empty value is considered present, so |TS| logs the empty
+value instead of falling back. So in the example above, the value of
+x-primary-id of the client request is logged if it exists, otherwise the value
+of x-secondary-id is logged if it exists, otherwise ``-`` is logged if neither
+of the headers is present.
+
+The final log field in the chain can be a non-header log field whose value
+specifies the final fallback in the chian. This symbol is used only after all
```
zwoop: Typo, chian -> chain

src/proxy/logging/LogFormat.cc
```diff
@@ -475,6 +501,41 @@ LogFormat::parse_symbol_string(const char *symbol_string, LogFieldList *field_li
...
+      Dbg(dbg_ctl_log_format, "Field fallback symbol: %s", symbol);
+      std::string parse_error;
+      auto        parsed = LogFieldFallback::parse(symbol, parse_error);
+      if (parsed.has_value()) {
+        std::unique_ptr<LogField> fallback_field;
+        bool                      valid_fallback = true;
+
+        if (parsed->fallback_symbol.has_value()) {
+          std::string fallback_symbol = *parsed->fallback_symbol;
+          std::string field_symbol    = fallback_symbol;
```
zwoop: Do we really need two copies here for the string?

zwoop (dismissed):

zwoop (approved):

bneradt:
10.2.x backport PR: https://github.com/apache/trafficserver/pull/13109

cmcfarlen:
Added to milestone 10.2.0 via #13109

PR #13021, opened 2026-03-25
sigusr2 autest: simplify Process and Ready objects

bneradt:
Rewrite the sigusr2 autest to drive each scenario from a single
shel helper script instead of a chain of AuTest Processes.

This removes the Ready-condition races from the old test, keeps the
signal and log rotation ordering in one place, and makes the log file
expectations easier to follow. This should address flakiness in the
test that has been seen in CI.

### DIFF
--- a/tests/gold_tests/logging/sigusr2.test.py
+++ b/tests/gold_tests/logging/sigusr2.test.py
@@ -18,9 +18,12 @@ Verify support of external log rotation via SIGUSR2.
 #  limitations under the License.
 
 import os
+import shlex
 import sys
 
 TS_PID_SCRIPT = 'ts_process_handler.py'
+ROTATE_DIAGS_SCRIPT = 'sigusr2_rotate_diags.sh'
+ROTATE_CUSTOM_LOG_SCRIPT = 'sigusr2_rotate_custom_log.sh'
 
 
 class Sigusr2Test:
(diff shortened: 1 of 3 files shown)


bryancall (changes requested):
The shell-script approach is a big improvement over the old Process chain. Much easier to follow the sequencing, and the explicit polling loops with 60-second timeouts should handle the ASAN/CI slowness that was causing the flakiness.

Two things I noticed:

1. **Missing `StillRunningAfter`** -- The old test had `tr1.StillRunningAfter = diags_test.ts` (and the server) to verify ATS survived the signal. Neither `add_system_log_test` nor `add_configured_log_test` sets this. Without it, if ATS crashes during SIGUSR2 handling, the test could still pass as long as the shell script exits 0 (the shell script's `wait_for_contains` would succeed on files written before the crash). Might be worth adding back.

2. **Typo in commit message** -- "shel helper script" should be "shell helper script" (appears twice).

bneradt:
> The shell-script approach is a big improvement over the old Process chain. Much easier to follow the sequencing, and the explicit polling loops with 60-second timeouts should handle the ASAN/CI slowness that was causing the flakiness.
> 
> Two things I noticed:
> 
> 1. **Missing `StillRunningAfter`** -- The old test had `tr1.StillRunningAfter = diags_test.ts` (and the server) to verify ATS survived the signal. Neither `add_system_log_test` nor `add_configured_log_test` sets this. Without it, if ATS crashes during SIGUSR2 handling, the test could still pass as long as the shell script exits 0 (the shell script's `wait_for_contains` would succeed on files written before the crash). Might be worth adding back.

Actually, with this change, we were able to associate the ATS instance with the more specific TestRun itself rather than the global Test namespace. That's a good thing since it is a cleaner solution - it's easier to think through the processes if they are all neatly associated with their own TestRun. Therefore the TS instance will not be "StillRunningAfter" - it intentionally is cleaned up when the TestRun completes. That said, any crash would manifest itself as a non-zero return in TS and will be seen via that mechanism.

> 2. **Typo in commit message** -- "shel helper script" should be "shell helper script" (appears twice).

Fixed.

bryancall (approved):

PR #13022, opened 2026-03-25
Update traffic ctl exit code

cmcfarlen:
Draft PR against 10.2.x since master branch is quite different here.

We are having an issue where `traffic_ctl` was returning a non-zero exit code when encountering config items that are unused.  This PR adds a new feature `--error-level` that defaults to `error`.  This is used to set the exit code when responses from the server contain errata.  It will find the worst errata severity and then compare that to the error level and if greater, will set the error exit code.

### DIFF
--- a/include/mgmt/rpc/jsonrpc/json/YAMLCodec.h
+++ b/include/mgmt/rpc/jsonrpc/json/YAMLCodec.h
@@ -195,8 +195,7 @@ class yamlcpp_json_encoder
       for (auto const &err : errata) {
         int severity = err.severity(ERRATA_DIAG);
         json << YAML::BeginMap;
-        // using "code" as the key here because this is decoded into a `JSONRPCError`
-        json << YAML::Key << "code" << YAML::Value << severity;
+        json << YAML::Key << "severity" << YAML::Value << severity;
         json << YAML::Key << "message" << YAML::Value << std::string{err.text().data(), err.text().size()};
         json << YAML::EndMap;
       }
--- a/include/shared/rpc/yaml_codecs.h
+++ b/include/shared/rpc/yaml_codecs.h
@@ -64,7 +64,7 @@ template <> struct convert<shared::rpc::JSONRPCError> {
     error.message = helper::try_extract<std::string>(node, "message");
     if (auto data = node["data"]) {
       for (auto &&err : data) {
-        error.data.emplace_back(helper::try_extract<int32_t>(err, "code"), helper::try_extract<std::string>(err, "message"));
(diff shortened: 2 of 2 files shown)


include/mgmt/rpc/jsonrpc/json/YAMLCodec.h
```diff
@@ -193,8 +193,10 @@ class yamlcpp_json_encoder
       json << YAML::Key << "data";
       json << YAML::BeginSeq;
       for (auto const &err : errata) {
+        int severity = err.severity(ERRATA_DIAG);
         json << YAML::BeginMap;
-        json << YAML::Key << "code" << YAML::Value << errata.code().value();
```
brbzull0, line 197: this is a bug in my code. it's printing errata.code all the time. It should be the particular error not the one set in the jsonrpc response.

include/mgmt/rpc/jsonrpc/json/YAMLCodec.h
```diff
@@ -193,8 +193,10 @@ class yamlcpp_json_encoder
...
+        // using "code" as the key here because this is decoded into a `JSONRPCError`
+        json << YAML::Key << "code" << YAML::Value << severity;
```
brbzull0: I think  if use the severity here in the code field we are changing the meaning from "error code" to "severity level" which is not the original intention.

I think we should use severity as part of the error definition instead:

```c++
// the only new stuff would be the severity.
struct JSONRPCError {
  int32_t     code;
  std::string message;
  struct DataEntry {
    int32_t     code;
    int32_t     severity{-1}; // -1 = not present (assume ERROR)
    std::string message;
  };
  std::vector<DataEntry> data;
};
```

This can be set by the handlers as you already did in the test, something like this:

```
resp.errata().assign(std::error_code{errors::Codes::SERVER})
  .note(ERRATA_WARN, "Server already draining.");

// json:

{
  "jsonrpc": "2.0",
  "error": {
    "code": 9,
    "message": "Error during execution",
    "data": [
      {
        "code": 2,
        "severity": 4,
        "message": "Server already draining."
      }
    ]
  },
  "id": "..."
}
```

then we can add the severity if present:

```cpp
for (auto const &err : errata) {
  json << YAML::BeginMap;
  json << YAML::Key << "code" << YAML::Value << errata.code().value();
  if (err.has_severity()) {
    json << YAML::Key << "severity" << YAML::Value
         << static_cast<int>(err.severity());
  }
  json << YAML::Key << "message" << YAML::Value
       << std::string{err.text().data(), err.text().size()};
  json << YAML::EndMap;
}
```

cmcfarlen: The Annotation class doesn't have code, only severity and text.

PR #12126, opened 2025-03-25
Fix header_rewrite loading remap mmdb files

cmcfarlen:
Introduced in #11320

PR #13025, opened 2026-03-27
hrw: Refix loading geodb files

cmcfarlen:
PR #12855 reintroduced a bug that would cause remap rules with header_rewrite plugins configured with mmdb files to never load the mmdb file.  This is because the call_once utilitized to initialize the file is a global to the plugin, so applies to every remap instance.  This means the first remap instance to load the mmdb will also set it for every other remap instance.  If you call `call_once` without a path, then any remap instances that call mmdb after will silently fail to load.

See $11812 and #12126

Edit:

I decided to try my idea for better handling of mmdb files.  The last commit of this PR does just that and allows remap rules with hrw to specify different geo db files rather than the current "first wins" approach.  It also avoids the call_once issue that has cause initialization issues in the past.

### DIFF
--- a/plugins/header_rewrite/header_rewrite.cc
+++ b/plugins/header_rewrite/header_rewrite.cc
@@ -705,17 +705,17 @@ TSRemapNewInstance(int argc, char *argv[], void **ih, char * /* errbuf ATS_UNUSE
     }
   }
 
-  if (!geoDBpath.empty() && !geoDBpath.starts_with('/')) {
-    geoDBpath = std::string(TSConfigDirGet()) + '/' + geoDBpath;
-  }
-
   if (!geoDBpath.empty()) {
+    if (!geoDBpath.starts_with('/')) {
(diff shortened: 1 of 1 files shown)


cmcfarlen:
[approve ci autest 1 2]

cmcfarlen:
We'll have to fix this another way as this breaks the fix for running plugins.  Converting to draft.

cmcfarlen:
This last commit handles the geo db init completely differently, but avoids the call_once issue and allows for rules to vary the mmdb file used.

bneradt (dismissed):
lgtm

cmcfarlen:
I removed the last commit with the improved mmdb loading for another PR.  This one just focuses on fixing the regression.

bneradt (approved):

cmcfarlen:
Cherry-picked to 10.2.x

cmcfarlen:

PR #13031, opened 2026-03-27
Replace storage.config and volume.config with storage.yaml

masaori335:
Address  #10766

# Summary

This PR merges storage.config (span/disk definitions) and volume.config (volume allocation rules) into a single storage.yaml file, and adds comprehensive documentation for the new format.                                             
                                                                                                                                                                                                                                          
# Implementation

- Add YamlStorageConfig class (src/iocore/cache/YamlStorageConfig.cc/h) that parses storage.yaml using yaml-cpp
- Refactor ConfigVol to use a nested Size struct replacing the old flat size/percent/in_percent fields
- Add ConfigVol::Span and ConfigVol::Spans to express per-span volume assignments directly in the config, replacing the old volume= annotation in storage.config
- Add ConfigVolumes::complement() to auto-fill omitted sizes by distributing remaining space evenly among volumes/spans without an explicit size
- Remove Span::forced_volume_num in favor of deriving exclusive span assignments from the YAML structure at startup
- Remove ConfigVolumes::read_config_file() and BuildListFromString() — volume config is now loaded as part of Store::read_config()
- Docs are based on https://github.com/apache/trafficserver/pull/11000 (against 11-Dev branch), but `ram_cache_size` and `ram_cache_cutoff` are added.

### DIFF
--- a/include/config/config_result.h
+++ b/include/config/config_result.h
@@ -59,8 +59,9 @@ namespace config
  * @tparam T The configuration type (e.g., SSLMultiCertConfig).
  */
 template <typename T> struct ConfigResult {
-  T            value;  ///< The parsed configuration value.
-  swoc::Errata errata; ///< Errors or warnings from parsing.
+  T            value;                  ///< The parsed configuration value.
+  swoc::Errata errata;                 ///< Errors or warnings from parsing.
+  bool         file_not_found = false; ///< True if the file was not found (ENOENT).
 
   /**
    * Check if parsing succeeded without errors.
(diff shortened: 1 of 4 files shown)


masaori335:
I moved YAML & legacy config parsers to `src/config/storage.cc`. Now `traffic_ctl config convert` has storage and volume config support. It seems pretty useful for migration. Thank you for suggestion.

```
traffic_ctl config convert storage <storage.config> <volume.config> <output_file>
```

PR #13034, opened 2026-03-28
Use AddAwaitFileContainsTestRun more in autests

bneradt:
This switches the migrated gold tests from condwait file polling to
AddAwaitFileContainsTestRun so they wait for stable log content
rather than raw file creation.

This also:

- Makes the await helper safe for regex needles
- Cleans up stale condwait path leftovers from earlier migrations.

### DIFF
--- a/tests/gold_tests/autest-site/when.test.ext
+++ b/tests/gold_tests/autest-site/when.test.ext
@@ -89,7 +89,7 @@ def AddAwaitFileContainsTestRun(test, name, file_path, needle, desired_count=1) 
     '''
     tr = test.AddTestRun(name)
     p = tr.Processes.Default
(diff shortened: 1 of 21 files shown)


tests/gold_tests/autest-site/when.test.ext
```diff
@@ -89,7 +89,7 @@ def AddAwaitFileContainsTestRun(test, name, file_path, needle, desired_count=1)
     '''
     tr = test.AddTestRun(name)
     p = tr.Processes.Default
-    p.Command = f'echo waiting for {needle} in {file_path}'
+    p.Command = f'echo waiting for file content in {file_path}'
```
bneradt, line 92: Printing `{needle}` had issues when `needle` was a regex in the autest framework.

zwoop (approved):

PR #13035, opened 2026-03-28
Support replay-only AuTest YAML tests

bneradt:
Draft because this depends on the upstream AuTest support landing
first.

Upstream dependency: https://github.com/AuTestSuite/AuTest/pull/1

Allow ATS replay tests to be discovered directly from .test.yaml files
so simple replay-based tests no longer need one-line Python wrappers.

This keeps autest list lightweight by loading replay metadata without
requiring ATS or Proxy Verifier binaries, while keeping existing
ATSReplayTest wrapper-based tests working. The parallel runner now uses
autest list for discovery so replay-only tests are included there too.

This depends on the upstream AuTest custom test format support landing
first: https://github.com/AuTestSuite/AuTest/pull/1


zwoop:
[approve ci autest 2]

PR #13036, opened 2026-03-30
Fix prev_is_cr flag handling in chunked encoding parser

bryancall:
## Summary

The `prev_is_cr` flag in `ChunkedHandler::read_size()` was being set within conditional expressions but not consistently updated, which could cause it to retain stale values across parsing iterations.

This change updates the flag at the end of each loop iteration to ensure it accurately reflects the current character state. This improves the correctness of line break validation when parsing chunked transfer encoding with `strict_chunk_parsing` enabled.

## Changes

- Remove inline assignment of `prev_is_cr` in conditionals
- Update `prev_is_cr` at end of loop iteration
- Add final update when loop exits with remaining data
- Add test case for edge case in chunked encoding

## Testing

- All regression tests pass
- Added test case to verify proper handling of edge cases in chunked encoding

### DIFF
--- a/src/proxy/http/HttpTunnel.cc
+++ b/src/proxy/http/HttpTunnel.cc
@@ -191,7 +191,7 @@ ChunkedHandler::read_size()
             done  = true;
             break;
           } else {
-            if ((prev_is_cr = ParseRules::is_cr(*tmp)) == true) {
+            if (ParseRules::is_cr(*tmp)) {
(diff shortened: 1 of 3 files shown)


bryancall:
[approve ci autest 2]

cmcfarlen:
Cherry-picked to 10.2.x

PR #13040, opened 2026-03-30
HttpSM - make sure we have a valid buffer to write on.

bneradt:
Add a check to make sure we have a valid buffer for tunnel producer to write on.


cmcfarlen:
Cherry-picked to 10.2.x

PR #13042, opened 2026-03-30
Support per-remap geo DB handles in header_rewrite

cmcfarlen:
Replace the single global geo DB handle with a cache of handles keyed by path, allowing different remap rules to use different MMDB files via --geo-db-path. The handle is threaded through RulesConfig -> Resources -> ConditionGeo::get_geo_*() methods. The cache deduplicates when multiple rules share the same path.


zwoop (approved):
I think Copilot's comment is a red herring mostly, the use cases it fears are just not likely to happen.

PR #13044, opened 2026-03-31
Add sni.yaml session ticket overrides (#13006)

### LINKED ISSUE #12953: Session tickets need to be defined in sni.yaml
It should be possible to:
* Enable tickets
* Set the number of tickets

Within sni.yaml, rather than as a system wide default.

bneradt:
Add ssl_ticket_enabled and ssl_ticket_number as sni.yaml overrides, apply them during SNI handling so they affect TLS 1.2 resumption and TLS 1.3 ticket issuance, and add unit and AuTest coverage plus docs and sample config updates.

Fixes #12953

(cherry picked from commit 8d75849e0e3574748f25fd682001e0566d75bd18)

### DIFF
--- a/src/iocore/net/SSLUtils.cc
+++ b/src/iocore/net/SSLUtils.cc
@@ -108,6 +108,12 @@ static DbgCtl dbg_ctl_ssl_session_cache{"ssl.session_cache"};
 static DbgCtl dbg_ctl_ssl_error{"ssl.error"};
 static DbgCtl dbg_ctl_ssl_verify{"ssl_verify"};
 
+#if TS_HAS_TLS_SESSION_TICKET
+static bool ssl_context_enable_ticket_callback(SSL_CTX *ctx);
+static bool ssl_apply_sni_session_ticket_properties(SSL *ssl);
+static bool ssl_set_session_ticket_number(SSL *ssl, size_t num_tickets);
+#endif
+
 /* Using pthread thread ID and mutex functions directly, instead of
  * ATS this_ethread / ProxyMutex, so that other linked libraries
  * may use pthreads and openssl without confusing us here. (TS-2271).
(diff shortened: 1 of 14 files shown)


src/iocore/net/SSLUtils.cc
```diff
@@ -401,13 +407,9 @@ ssl_cert_callback(SSL *ssl, [[maybe_unused]] void *arg)
...
-    // Reset the ticket callback if needed
-    SSL_CTX *ctx = SSL_get_SSL_CTX(ssl);
-#ifdef HAVE_SSL_CTX_SET_TLSEXT_TICKET_KEY_EVP_CB
-    SSL_CTX_set_tlsext_ticket_key_evp_cb(ctx, ssl_callback_session_ticket);
-#else
-    SSL_CTX_set_tlsext_ticket_key_cb(ctx, ssl_callback_session_ticket);
-#endif
+    if (!ssl_apply_sni_session_ticket_properties(ssl)) {
+      retval = 0;
+    }
```
cmcfarlen, line 412: This is where I was confused trying to resolve conflicts.  The old code calls `set_tlsext_ticket_key...` but the new code path doesn't seem to call that.  So I worry this isn't compatible, but maybe the same thing is accomplished?

bneradt: Thanks for checking into this. The logic has changed slightly, I think for the better. Now the callback is set (along with other callbacks) in the `SSLMultiCertConfigLoader::init_server_ssl_ctx` below. You can see the `ssl_context_enable_ticket_callback` in that function further down in this file in this patch. Currently line 1347.

I'm honestly not sure why the callback was set here in the first place. If you look at the other stuff in this `ssl_cert_callback`, it's all cert related, not "add a callback" functionality. It had to go out of its way to reach into the SSL and get the SSL_CTX to set the callback on it. `init_server_ssl_ctx` is the appropriate place to configure the ticket callback.

This discussion also applies to the master patch, btw. (In case that's helpful.)

cmcfarlen: Thank you for the extra info!

bneradt:
10.1.x PR: https://github.com/apache/trafficserver/pull/13297

PR #13046, opened 2026-03-31
Try to change uid/gid on plugin runtime dir before dropping privs, if POSIX not available

ezelkow1:
This is a workaround for systems that don't provide the same POSIX style permissions dropping, i.e. BSD.

This recursively changes the permissions on the runtime (usually var) and logging directories and files to be the user we are dropping to

### DIFF
--- a/src/traffic_server/traffic_server.cc
+++ b/src/traffic_server/traffic_server.cc
@@ -1833,7 +1833,7 @@ chown_owned_dirs(const char *user)
     chown_dir_recursive(logdir.c_str(), pwd->pw_uid, pwd->pw_gid);
   }
 }
-#endif
+#endif // !TS_USE_POSIX_CAP
 
 /*
  * Binds stdout and stderr to files specified by the parameters


src/traffic_server/traffic_server.cc
```diff
@@ -1757,6 +1758,83 @@ change_uid_gid(const char *user)
...
+
+  if (!rundir.empty()) {
+    chown_dir_recursive(rundir.c_str(), pwd->pw_uid, pwd->pw_gid);
+  }
+
+  if (!logdir.empty()) {
+    chown_dir_recursive(logdir.c_str(), pwd->pw_uid, pwd->pw_gid);
+  }
+}
+#endif
```
bneradt: A minor thing, but we often add the comment indicating what the `#endif` matches up with:

```cpp
#endif //  !TS_USE_POSIX_CAP
```

Mind adding that?

ezelkow1: added

cmcfarlen:
Cherry-picked to 10.2.x

PR #13047, opened 2026-03-31
Fix link_libraries for cripts unit tests

cmcfarlen:
This fixes build issues when openssl libs aren't in the default location.


cmcfarlen:
Cherry-picked to 10.2.x

PR #13052, opened 2026-04-02
Fix set-status crash inside if/endif at remap time

moonchen:
Operators inside an if/endif block never had set_hook() called on them, so _hook retained its default value of TS_HTTP_READ_RESPONSE_HDR_HOOK. When set-status executed at REMAP_PSEUDO_HOOK time, exec() took the response-hook branch and called TSHttpHdrStatusSet on a request buffer, corrupting the header union and crashing.

Fix the operator to check TSHttpHdrTypeGet() at runtime instead of trusting get_hook().  Removed _hook and get_hook() from operators.

set_hook() becomes is_hook_valid() in Statement and RuleSet to only validate hook compatibility without mutating _hook.

Upgrade the HTTPHdr::status_set/reason_set asserts to release_assert so that writing response fields into a request buffer is caught immediately rather than silently corrupting memory.

Includes an autest that reproduces the crash scenario: set-status 403 inside an if/endif block under REMAP_PSEUDO_HOOK.

### DIFF
--- a/include/proxy/hdrs/HTTP.h
+++ b/include/proxy/hdrs/HTTP.h
@@ -1097,7 +1097,7 @@ inline void
 HTTPHdr::status_set(HTTPStatus status)
 {
   ink_assert(valid());
-  ink_assert(m_http->m_polarity == HTTPType::RESPONSE);
+  ink_release_assert(m_http->m_polarity == HTTPType::RESPONSE);
 
   http_hdr_status_set(m_http, status);
 }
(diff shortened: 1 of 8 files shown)


zwoop:
Does this still need to be marked as a Draft ?

moonchen:
@zwoop ready for review when you have a chance

zwoop (approved):
Looks good.

cmcfarlen:
Cherry-picked to 10.2.x

PR #13053, opened 2026-04-02
traffic_ctl config convert remap

bneradt:
Move ATS core off the remaining remap-only parsing path and route remap.config loading through the shared src/config remap parser. This lets traffic_ctl and traffic_server use the same conversion logic while leaving only runtime regex-mapping helpers in the remap loader.

This builds off of the work to support remap.yaml here: #12997


src/config/remap.cc
```diff
@@ -0,0 +1,631 @@
...
+#include "tscore/MatcherUtils.h"
+#include "tscore/Tokenizer.h"
+#include "tsutil/ts_diag_levels.h"
+
+namespace
+{
+
+constexpr swoc::Errata::Severity ERRATA_NOTE_SEV{static_cast<swoc::Errata::severity_type>(DL_Note)};
+constexpr swoc::Errata::Severity ERRATA_ERROR_SEV{static_cast<swoc::Errata::severity_type>(DL_Error)};
+
```
masaori335, line 45: I prefer YAML key names definition in the beginning.

include/config/remap.h
```diff
@@ -0,0 +1,92 @@
...
+#include <string_view>
+
+#include <yaml-cpp/yaml.h>
+
+#include "config/config_result.h"
+
+namespace config
+{
+
+using RemapConfig = YAML::Node;
```
masaori335, line 36: Hmm, can we define some struct describes remap.yaml node tree? It'll help us understanding type of node - int, string, list, map.

src/proxy/http/remap/RemapYamlConfig.cc
```diff
@@ -40,16 +40,80 @@
...
 
+extern CacheHostRecord *createCacheHostRecord(const char *volume_str, char *errbuf, size_t errbufsize);
+
 namespace
 {
 DbgCtl dbg_ctl_remap_yaml{"remap_yaml"};
 DbgCtl dbg_ctl_url_rewrite{"url_rewrite"};
+
+swoc::Errata
+parse_yaml_volume(const YAML::Node &node, url_mapping *url_mapping)
```
masaori335, line 59: I feel these parsers in this file should moved to `config/remap.cc`

PR #13054, opened 2026-04-02
proxy-request.expect: absent in PV replays

bneradt:
Proxy Verifier v3.1.2 can now verify that ATS does not send an upstream
request to the origin. Update replay transactions that serve from cache
or otherwise respond locally to use proxy-request.expect: absent.

This makes the no-origin cases explicit across the autests and turns the
old trap server responses into a direct expectation. It also fixes one
stale comment in the HEAD/GET cache replay so the test text matches the
origin behavior.


masaori335 (approved):
Nice! This clarify behavior of ATS.

cmcfarlen:
Cherry-picked to 10.2.x

PR #13055, opened 2026-04-02
Add autest for empty POST framing: 411 response

bneradt:
Add a replay-based autest for a POST request that omits both
Content-Length and Transfer-Encoding. This captures the current
411 response in ATS and gives us a focused reproducer for issue
12960.

411 is in compliance with the RFC:

https://datatracker.ietf.org/doc/html/rfc9112#section-6.3

  A server MAY reject a request that contains a message body but not a
  Content-Length by responding with 411 (Length Required).

Fixes: #12960


cmcfarlen:
Cherry-picked to 10.2.x

PR #13057, opened 2026-04-04
Install rustup instead of cargo to use a newer version of rustc

hnakamur:
This is needed to avoid the following error while building the Cloudflare quiche library:
```
error: package `serde_with v3.18.0` cannot be built because it requires rustc 1.88 or newer, while the currently active rustc version is 1.75.0
```


masaori335 (approved):
Thanks!

cmcfarlen:
Thanks for this fix! We don't need to pick to 10.2 since the master branch Dockerfile can build any tag.

PR #13059, opened 2026-04-04
Log malformed HTTP/2 requests

bneradt:
Unlike HTTP/1 transactions, malformed HTTP/2 requests are rejected before HttpSM creation, so they bypassed the normal transaction logging path. That left malformed h2 traffic out of squid.log even when similar h1 failures were visible.

This adds a pre-transaction LogAccess path for malformed h2 request headers and emits a best-effort access log entry before resetting the stream.

### DIFF
--- a/src/proxy/logging/CMakeLists.txt
+++ b/src/proxy/logging/CMakeLists.txt
@@ -37,7 +37,7 @@ add_library(ts::logging ALIAS logging)
 
 target_include_directories(logging PRIVATE ${SWOC_INCLUDE_DIR})
 
-target_link_libraries(logging PUBLIC ts::inkevent ts::inkutils ts::http ts::hdrs ts::tscore yaml-cpp::yaml-cpp)
+target_link_libraries(logging PUBLIC ts::inkevent ts::inkutils ts::hdrs ts::tscore ts::configmanager yaml-cpp::yaml-cpp)
 
 if(BUILD_TESTING)
   add_executable(test_LogFieldFallback unit-tests/test_LogFieldFallback.cc)
@@ -54,6 +54,10 @@ if(BUILD_TESTING)
   target_compile_definitions(test_RolledLogDeleter PRIVATE TEST_LOG_UTILS)
   target_link_libraries(test_RolledLogDeleter tscore ts::inkevent records Catch2::Catch2WithMain)
   add_catch2_test(NAME test_RolledLogDeleter COMMAND test_RolledLogDeleter)
+
+  add_executable(test_LogAccess unit-tests/test_LogAccess.cc)
+  target_link_libraries(test_LogAccess ts::logging ts::configmanager ts::inkevent records Catch2::Catch2WithMain)
+  add_catch2_test(NAME test_LogAccess COMMAND test_LogAccess)
 endif()
 
 clang_tidy_check(logging)
--- a/tests/gold_tests/connect/h2_malformed_request_logging.test.py
+++ b/tests/gold_tests/connect/h2_malformed_request_logging.test.py
@@ -1,0 +1,207 @@
+'''
+Verify malformed HTTP/2 requests are access logged.
+'''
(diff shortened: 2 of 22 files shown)


tests/gold_tests/connect/h2_malformed_request_logging.test.py
```diff
@@ -0,0 +1,207 @@
...
+    def _keep_support_processes_running(self, tr):
+        if self._processes_started:
+            tr.StillRunningAfter = self._server
+            tr.StillRunningAfter = self._ts
+            return
+
+        tr.Processes.Default.StartBefore(self._server)
+        tr.Processes.Default.StartBefore(self._ts)
+        tr.StillRunningAfter = self._server
+        tr.StillRunningAfter = self._ts
```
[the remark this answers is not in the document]
bneradt: AuTest StillRunningAfter doesn't work that way. Multiple assignments is the idiomatic way to configure multiple StillRunningAfter.

maskit (dismissed):
It's good that this narrows the gap between H1 and H2 in terms of logging.

Suggestions for future improvement:
- I'm pretty sure H3 has the same issue. It would be nice if ProxyTransaction could take care of this pre-transaction scenario . 
- A cleaner interface would be having just `LogAccess(LogData *data)` for both. The subclasses of `LogData` would be `TransactionLogData` that copies data from `HttpSM`, and `PreTransactionLogData` that copies data from `ProxyTransaction` (or `Http2Stream`). The subclasses would be defined in http or http2 module. Then we may be able to remove the the circular dependency (http <-> logging).

bneradt:
Converting to a draft while I work on @maskit 's comments. I'll un-draft when the patch is in better shape.

src/proxy/logging/CMakeLists.txt
```diff
@@ -54,6 +54,10 @@ if(BUILD_TESTING)
   target_compile_definitions(test_RolledLogDeleter PRIVATE TEST_LOG_UTILS)
   target_link_libraries(test_RolledLogDeleter tscore ts::inkevent records Catch2::Catch2WithMain)
   add_catch2_test(NAME test_RolledLogDeleter COMMAND test_RolledLogDeleter)
+
+  add_executable(test_LogAccess unit-tests/test_LogAccess.cc)
+  target_link_libraries(test_LogAccess ts::logging ts::http ts::inkevent records Catch2::Catch2WithMain)
```
maskit: I wonder what requires `ts::http`.

src/proxy/http/TransactionLogData.cc
```diff
@@ -0,0 +1,284 @@
...
+  }
+  return LOG_FINISH_FIN;
+}
+} // end anonymous namespace
+
+TransactionLogData::TransactionLogData(HttpSM *sm) : m_http_sm(sm)
+{
+  ink_assert(sm != nullptr);
+
+  milestones = &sm->milestones;
```
maskit: I think we don't do these copies now, do we? Since `TransactionLogData` has `m_http_sm`, it should be able to provide getters.

maskit (commented):
The copies are concerning. Otherwise, looks good.

src/proxy/http/TransactionLogData.cc
bneradt: That's a good idea. Implemented with getters.

src/proxy/logging/CMakeLists.txt
bneradt: True. We can get rid of `ts:http` now, but that was indirectly pulling in `ts::configmanager` which is needed. Replacing it with `ts::configmanager`.

bneradt: And same with the `logging` target above.

maskit (approved):

bneradt:
10.2.x backport PR: https://github.com/apache/trafficserver/pull/13105

cmcfarlen:
Added to milestone 10.2.0 via #13105

PR #13067, opened 2026-04-07
Exclude http_load sources from RAT

bneradt:
This excludes the third-party http_load sources from RAT checks.
These files carry their own upstream license text rather than ASF
headers, so checking them as Apache-licensed sources causes a false
failure.

This keeps the RAT target consistent across environments and avoids
local failures on http_load.c.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -915,8 +915,7 @@ endif()
 add_custom_target(
   rat
   COMMENT "Running Apache RAT"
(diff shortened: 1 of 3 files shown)


maskit (changes requested):
Can we narrow the scope to tools/http_load/. We may use same file names in another directory.

maskit:
[approve ci autest 2]

maskit:
(Seems like autest sigusr2 is still unstable...)

I thought the change is just prepending the path like:
```
^tests/gold_tests/autest-site/min_cfg$
```

And now I see many entries that only check filenames without the paths. Did we actually chose to have just filename so that we can move files into another directory easily?

bneradt:
I'm going to resolve this by updating from rat 0.13 to 0.17 which supports glob-like paths, making addressing this easier. See: #13071 .

PR #13068, opened 2026-04-07
hrw4u: Fix u4wrh HEADER value resolution missing section context

zwoop:
_handle_set_rm_operation was passing None as the section to _rewrite_inline_percents, causing %{HEADER:...} in operator values to always resolve to the default inbound.resp. prefix instead of the correct context-dependent prefix.


maskit (approved):
Maybe too early for "self-approval" but it looks good to me.

cmcfarlen:
Cherry-picked to 10.2.x

PR #13069, opened 2026-04-08
Add connect_down_policy: 4 for empty reply

masaori335:
# Problem                                                   
                                                                                                                                                                                                                                                           
When an origin server accepts a TCP/TLS connection but closes it without sending any HTTP response, ATS does not mark the origin as down — even after exhausting all retry attempts. This can cause ATS to keep hammering a misbehaving origin on every request.                                                                                                                                                                                                                                                

# Fix 

- Add a new policy, `4`, all failures covered by policy 3, plus the case where `server_response_hdr_bytes == 0` at the end of a transaction — i.e. the origin closed or reset the connection after the TCP handshake without sending any HTTP response.
- The `HttpSM::track_connect_fail()` function was also refactored for clarity. 
- AuTest uses new `on_connect: refuse|reset` feature (thank you, @bneradt )

## notes

Two cases are explicitly excluded:

- Reused keep-alive connections: there is a known race between ATS reusing a keep-alive connection and the origin closing it, described by @moonchen on #12437. Considering this case, the policy 4 is only applied for the first transaction.
- Multiplexed origins ( HTTP/2 ): per-stream error observation (e.g. RST_STREAM) does not indicate the underlying connection is unhealthy.

### DIFF
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -1834,12 +1834,24 @@ Origin Server Connect Attempts
    :overridable:
 
    Controls what origin server connection failures contribute to marking a server down.
-   When set to ``2``, any connection failure during the TCP and TLS handshakes will
-   contribute to marking the server down. When set to ``1``, only TCP handshake failures
-   will contribute to marking a server down. When set to ``0``, no connection failures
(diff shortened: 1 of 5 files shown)


bneradt (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #13075, opened 2026-04-09
Add `Test.AddConfigReload()` autest extension

### LINKED ISSUE #12965: ATS Reload migration task. Autest reload extension
Implement an autest extension that checks reload success/failure via the JSONRPC status API (traffic_ctl config status -t <token>) instead of grepping log files.

brbzull0:
Replace the fragile pattern of fire-and-forget `traffic_ctl config reload` followed by sleep/log-grepping with a deterministic helper that uses monitor mode (`-m`) to block until completion and validates reload tasks via the get_reload_config_status JSONRPC endpoint.

The extension supports expected exit codes (success/fail/timeout/any), task presence and absence assertions, per-task status validation, custom tokens, and a delay_start parameter for filesystem timestamp sensitivity.

Migrate 21 existing tests to use the new helper, removing manual sleep synchronization and When.FileContains polling. Add an optional filename parameter to ConfigReloadTask so sub-tasks carry their associated config file for precise identification in test assertions.

#### Tiny change to the reload core:

Reload sub-tasks previously only carried a description (e.g."SNIConfig"), making it impossible to assert which config file triggereda task. The C++ change adds an optional `filename` parameter to`ConfigReloadTask::add_child()` and `ConfigContext::add_dependent_ctx()`so the JSONRPC status endpoint returns the actual file path (e.g.`sni.yaml`, `remap.yaml`) for each sub-task. This allows tests to validate that the right config handlers ran -- or didn't run.

**Includes RST documentation for the new extension API.**

Fixes https://github.com/apache/trafficserver/issues/12965


tests/gold_tests/autest-site/config_reload.test.ext
```diff
@@ -0,0 +1,228 @@
...
+                    return (False, f"Task '{task_name}' not found in reload. Found: {sorted(all_names)}")
+                if task_status is not None:
+                    task = _find_task(info, task_name)
+                    actual = task.get("status", "unknown") if task else "unknown"
+                    if actual != task_status:
+                        return (False, f"Task '{task_name}' status is '{actual}', expected '{task_status}'")
+
+        if expect_absent_tasks:
+            for task_name in expect_absent_tasks:
+                if task_name in all_names:
```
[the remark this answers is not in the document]
brbzull0: `_collect_task_names()` already normalizes to basenames via `os.path.basename()`, and all call-sites pass short names (e.g. `"sni.yaml"`, not full paths). The JSONRPC endpoint returns whatever was stored on the C++ side, which in this PR is always a filename constant (e.g. `ts::filename::SNI`), not an absolute path. The basename fallback is defensive.

Using `_find_task()` instead would work equivalently but isn't a fix — both approaches resolve to the same result for the inputs this API accepts. Keeping the `set` lookup for presence and reserving `_find_task()` for when we need the actual task object (i.e. status checks) feels cleaner.

tests/gold_tests/autest-site/config_reload.test.ext
```diff
@@ -0,0 +1,228 @@
...
+
+    if timeout is not None:
+        cmd += f" -T {timeout}"
+    if force:
+        cmd += " --force"
+    if data is not None:
+        if data.startswith("@"):
+            cmd += f" --data {data}"
+        else:
+            cmd += f" --data {_shell_quote(data)}"
```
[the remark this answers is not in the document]
brbzull0: Good catch. While `token` is auto-generated (no metacharacters) and `data` paths come from test code, it's cheap to make this defensive. Fixed — both `token` and `data` (including `@path` form) are now shell-quoted via `_shell_quote()`.

tests/gold_tests/tls/tls_sni_yaml_reload.test.py
```diff
@@ -108,20 +106,13 @@
 trupd.Processes.Default.Env = ts.Env
 trupd.Processes.Default.ReturnCode = 0
 
-tr2reload = Test.AddTestRun("Reload config")
-tr2reload.StillRunningAfter = ts
+tr2reload = Test.AddConfigReload(ts, expect="fail", expect_tasks=["sni.yaml"], description="Reload config")
```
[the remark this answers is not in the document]
brbzull0: `expect="fail"` already asserts the overall reload exit code is 2 (failure), and `expect_tasks=["sni.yaml"]` confirms the task was present in the reload tree. Since the overall status is derived from sub-task statuses, adding `{"sni.yaml": "fail"}` would double-assert the same condition.

The dict form is useful when you need to verify mixed per-task statuses independently of the overall result (e.g. `{"sni.yaml": "fail", "SSLConfig": "success"}` to prove the failure was isolated). For this simple single-task case the list form is intentional.

tests/gold_tests/tls/ssl_multicert_loader.test.py
```diff
@@ -78,19 +78,13 @@
 tr2.Processes.Default.Env = ts.Env
 tr2.Processes.Default.ReturnCode = 0
 
-tr2reload = Test.AddTestRun("Reload config")
-tr2reload.StillRunningAfter = ts
+tr2reload = Test.AddConfigReload(ts, expect="fail", expect_tasks=["ssl_multicert.yaml"], description="Reload config")
```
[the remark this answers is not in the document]
brbzull0: Same reasoning as the `tls_sni_yaml_reload` comment — `expect="fail"` covers the failure assertion and the list form confirms the right handler was involved. The dict form would be redundant here. It's available for more complex multi-task scenarios where per-task status matters independently.

brbzull0:
[approve ci]

brbzull0:
@moonchen  just a reminder 😄

tests/gold_tests/traffic_ctl/remap_inc/remap_inc.test.py
```diff
@@ -14,16 +14,12 @@
...
 
-import os
-
 Test.Summary = '''
 Test traffic_ctl config reload with remap.config .include directive
 '''
 
 Test.ContinueOnFail = False
 
-Test.Setup.Copy("wait_reload.sh")
```
moonchen, line 25: Since you're removing this code, can you please also remove the file `wait_reload.sh`

tests/gold_tests/tls/ssl_multicert_loader.test.py
```diff
@@ -78,19 +78,13 @@
...
 tr2reload.StillRunningAfter = server
-tr2reload.Processes.Default.Command = 'traffic_ctl config reload'
-tr2reload.Processes.Default.Env = ts.Env
-tr2reload.Processes.Default.ReturnCode = 0
-ts.Disk.diags_log.Content = Testers.ContainsExpression('ERROR: ', 'ERROR')
```
moonchen, line 87: Was this check removed on purpose?

tests/gold_tests/tls/ssl_multicert_loader.test.py
```diff
@@ -78,19 +78,13 @@
...
 
 # Reload of ssl_multicert.yaml should fail, BUT the old config structure
 # should be in place to successfully answer for the test domain
 tr3 = Test.AddTestRun("Make request again for $sni_domain")
-# Wait for the reload to complete
-tr3.Processes.Default.StartBefore(server2, ready=When.FileContains(ts.Disk.diags_log.Name, 'failed to load certificate ', 1))
```
moonchen, line 93: Was this check removed on purpose?

tests/gold_tests/remap_yaml/remap_reload_yaml.test.py
```diff
@@ -106,15 +105,10 @@ def update_remap_yaml(path: str, lines: list) -> None:
...
     '''.split("\n")))
 
-tr = Test.AddTestRun("remap_yaml reload, fails")
-tr.Processes.Default.Env = tm.Env
-tr.Processes.Default.Command = 'sleep 2; traffic_ctl config reload'
+tr = Test.AddConfigReload(tm, expect="fail", expect_tasks=["remap.yaml"], delay_start=2, description="remap_yaml reload, fails")
 
 tr = Test.AddTestRun("after first reload")
-await_config_reload = tr.Processes.Process('config_reload_failed', 'sleep 30')
-await_config_reload.Ready = When.FileContains(tm.Disk.diags_log.Name, "configuration is invalid")
```
moonchen, line 115: Could this check stay in place?

tests/gold_tests/tls/tls_sni_yaml_reload.test.py
```diff
@@ -84,8 +84,6 @@
 tr.Processes.Default.ReturnCode = 0
 tr.Processes.Default.Streams.stdout = Testers.ExcludesExpression("Could Not Connect", "Verify curl could successfully connect")
 tr.Processes.Default.Streams.stderr = Testers.IncludesExpression(f"CN={sni_domain}", f"Verify curl used the {sni_domain} SNI")
-ts.Disk.diags_log.Content = Testers.IncludesExpression(
-    "SSL negotiation finished successfully", "Verify that the TLS handshake was successful")
```
moonchen, line 88: Was this check removed on purpose?

src/iocore/net/SSLClientCoordinator.cc
```diff
@@ -37,8 +37,8 @@ SSLClientCoordinator::reconfigure(ConfigContext reconf_ctx)
   // The SSLConfig owns the client cert context storage and the SNIConfig will load
   // into it.
   SSLConfig::reconfigure(reconf_ctx.add_dependent_ctx("SSLConfig"));
-  SNIConfig::reconfigure(reconf_ctx.add_dependent_ctx("SNIConfig"));
-  SSLCertificateConfig::reconfigure(reconf_ctx.add_dependent_ctx("SSLCertificateConfig"));
+  SNIConfig::reconfigure(reconf_ctx.add_dependent_ctx("SNIConfig", ts::filename::SNI));
+  SSLCertificateConfig::reconfigure(reconf_ctx.add_dependent_ctx("SSLCertificateConfig", ts::filename::SSL_MULTICERT));
 #if TS_USE_QUIC == 1
   QUICCertConfig::reconfigure(reconf_ctx.add_dependent_ctx("QUICCertConfig"));
```
moonchen: should this also get `ts::filename::SSL_MULTICERT`?

moonchen (changes requested):

moonchen:
@brbzull0 thanks for the PR.  I think it's a good change.  Just a few small areas where I thought it was worth a second look.

tests/gold_tests/traffic_ctl/remap_inc/remap_inc.test.py
brbzull0: Good catch, removed.

tests/gold_tests/tls/ssl_multicert_loader.test.py
brbzull0: Yeah, on purpose. `expect="fail"` already asserts the reload exited with code `2` via monitor mode, and `expect_tasks=["ssl_multicert.yaml"]` confirms the right handler actually ran.

tests/gold_tests/tls/ssl_multicert_loader.test.py
brbzull0: Yes. That `When.FileContains` was sort of a sync gate, it waited for the log line before running the next curl. Now `traffic_ctl config reload -m` blocks until the reload reaches a terminal state, so by the time tr3 runs the reload is definitely done.

tests/gold_tests/remap_yaml/remap_reload_yaml.test.py
brbzull0: This is the same as above: the `await_config_reload` was a sleep 30 placeholder. This was used because the previous reload run was fire-and-forget.  Now with the new `-m` the reload run only completes after the reload terminates, so the next run can start the verifier client straight away. The removed string validation is now asserted by `expect="fail"`.

tests/gold_tests/tls/tls_sni_yaml_reload.test.py
brbzull0: Yes. Now that I was re-checking this It turns out it's a dead check on master, the string itself doesn't appear anywhere in the ATS source.

My understanding is that the successful handshake is already proven by the `CN={sni_domain}` check on curl stderr right above.

src/iocore/net/SSLClientCoordinator.cc
brbzull0: Yes!!  thanks.

brbzull0:
> @brbzull0 thanks for the PR. I think it's a good change. Just a few small areas where I thought it was worth a second look.

Thanks @moonchen  for the review.  Pushed some changes already.  👍

zwoop:
[approve ci]

brbzull0:
[approve ci]

brbzull0:
[approve ci autest 1]

moonchen (approved):

PR #13077, opened 2026-04-09
Allow full CIDR range (1-128) for IPv6 literals in HRW4U grammar

Clendenin:
## Summary

  The `IPV6_CIDR` lexer fragment in the HRW4U grammar rejected values
  1-32, reserving them exclusively for `IPV4_CIDR`. This caused valid
  IPv6 prefixes like `2620:149:a00::/32` to fail with a lexer error.

  The restriction was unnecessary because `IPV4_LITERAL` and
  `IPV6_LITERAL` are already unambiguous at the lexer level: IPv4
  requires dots, IPv6 requires colons. The CIDR fragments are consumed
  inside these parent token rules, so they never compete. This was
  verified by tokenizing overlapping CIDR values through the generated
  ANTLR lexer.

  The fix expands `IPV6_CIDR` from 33-128 to 1-128.


ezelkow1:
[approve ci]

zwoop (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #13078, opened 2026-04-09
Fix use-after-free in ParsedConfigCache for short config strings

maskit:
- Fix use-after-free in ParsedConfigCache when config values are short enough for std::string SSO (Small String Optimization)
- ParsedValue::parse() returned by value, and emplace moved it into the map — relocating the SSO inline buffer while string_views in TargetedCacheControlHeaders::headers[] still pointed to the old address
- Make ParsedValue non-movable and use try_emplace + parse_into() so parsing happens directly in the map node
- Also fixes the same class of bug for HostResData::conf_value and HttpStatusCodeList::conf_value pointers

Reproducer: configure conf_remap with a short targeted header value like ACME-Cache-Control (18 chars, within libc++ SSO threshold of 22). The string_views in the per-transaction override become dangling, causing incorrect cache behavior.

The SSO threshold varies by standard library — libc++ (macOS/clang): 22 bytes, libstdc++ (GCC/Linux): 15 bytes. A value like ACME-Cache-Control (18 chars) triggers SSO on libc++ but uses heap allocation on libstdc++, where the buffer pointer survives the move. This is why the bug may reproduce on macOS but not on Linux CI with GCC.

### DIFF
--- a/include/proxy/http/HttpConfig.h
+++ b/include/proxy/http/HttpConfig.h
@@ -933,6 +933,12 @@ public:
     std::variant<std::monostate, HostResData, HttpStatusCodeList, HttpForwarded::OptionBitSet, MgmtByte,
                  TargetedCacheControlHeaders>
       parsed{};
+
+    ParsedValue()                               = default;
+    ParsedValue(const ParsedValue &)            = delete;
+    ParsedValue &operator=(const ParsedValue &) = delete;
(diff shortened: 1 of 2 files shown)


brbzull0 (approved):
Great!
Thanks.

cmcfarlen:
Cherry-picked to 10.2.x

PR #13082, opened 2026-04-10
jax_fingerprint: Address user arg slot exhaustion

bneradt:
When multiple jax_fingerprint.so instances are loaded (e.g., one per fingerprinting method), each instance was reserving its own user arg slot. ATS has a limited number of slots (~4 per type), causing methods like JA3 and JA4 to fail silently when slots were exhausted.

Solution: Share a single user arg slot per type (TS_USER_ARGS_VCONN or TS_USER_ARGS_TXN) across all jax_fingerprint instances. A ContextMap stores JAxContext instances keyed by method name.

### DIFF
--- a/plugins/experimental/jax_fingerprint/context_map.h
+++ b/plugins/experimental/jax_fingerprint/context_map.h
@@ -1,0 +1,143 @@
+/** @file
+ *
+ * Shared context map for jax_fingerprint plugin instances.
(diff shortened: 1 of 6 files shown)


maskit (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #13083, opened 2026-04-10
Cleanup serving stale while origin server down

masaori335:
`HttpTransact::HandleCacheOpenReadHit` try to return stale contents by checking `is_server_negative_cached(s)`, but actually, this is dead code. Because when all origin servers are marked down, `HttpTransact::OSDNSLookup` serves stale contents.

~~This PR marked as draft because new AuTest depends on https://github.com/apache/trafficserver/pull/13069.~~ (update: this is ready)

### DIFF
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -1948,7 +1948,7 @@ Origin Server Connect Attempts
    :overridable:
 
    Specifies how long (in seconds) |TS| remembers that an origin server was unreachable.
-   During this window, if a stale cached response exists and its age is within
+   During this window, if a stale cached response exists and its age is within the cached response's ``max-age`` plus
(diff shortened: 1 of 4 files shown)


masaori335:
[approve ci]

bryancall (approved):
The removed `is_server_negative_cached()` check in `HandleCacheOpenReadHit` is genuinely redundant. When hostdb has all origin servers marked down, `select_best_http()` returns null, `set_active(null)` sets `resolved_p = false`, and on re-entry `OSDNSLookup` serves the stale cached response directly without contacting origin. The replay test confirms the revalidation request never reaches origin inside the down window and that normal revalidation resumes once it expires. This centralizes the stale-on-all-down path in `OSDNSLookup` instead of duplicating it in the caller, and the freshness boundary in `is_stale_cache_response_returnable()` is unchanged.

masaori335:
[approve ci autest]

cmcfarlen:
Cherry-picked to 10.2.x

PR #13084, opened 2026-04-10
jax_fingerprint: Add --log-field option for custom log fields

maskit:
## Summary
                                                                                                                                                                                                                                             
  - Add `--log-field <symbol>` option to jax_fingerprint plugin that registers a custom log field usable in `logging.yaml` format strings (e.g., `--log-field jaxja4` enables `%<jaxja4>`) 
  - Change `TSLogMarshalCallback` and `LogField::CustomMarshalFunc` from raw function pointers to`std::function` to support capturing lambdas, enabling plugins to register log field callbacks with state
  - Reject `--log-field` in remap plugin context with a clear error, since log fields must be registered globally                                                                                                                            
                                                                                                                                                                                                                                             
  ## Details

  The `--log-field` option requires the plugin to be loaded globally in `plugin.config` because log fields must be registered before log formats are parsed at startup. It works with all fingerprinting methods — both CONNECTION_BASED  (JA3, JA4) and REQUEST_BASED (JA4H). If you use a remap-only setup, you must also load the plugin globally with `--log-field` to register the log field.
                                                                                                                                                                                                                                             
  The plugin captures its config in a lambda passed to `TSLogFieldRegister`, which required widening the callback types from raw function pointers to `std::function`.

  ### Plugin API change
                  
  `TSLogMarshalCallback` in `apidefs.h` changes from a raw function pointer (`int (*)(TSHttpTxn, char *)`) to `std::function<int(TSHttpTxn, char *)>`. This is not a breaking change — existing plugins that pass plain function pointers or non-capturing lambdas to `TSLogFieldRegister` will continue to work as-is, since `std::function` implicitly converts from both. The change simply allows capturing lambdas and other callable objects as well.

### DIFF
--- a/doc/admin-guide/plugins/jax_fingerprint.en.rst
+++ b/doc/admin-guide/plugins/jax_fingerprint.en.rst
@@ -85,6 +85,15 @@ This option specifies the name of the header field where the plugin stores the g
 
 This option specifies the filename for the plugin log file. If not specified, log output will be suppressed.
 
+.. option:: --log-field <symbol>
+
+This option registers a custom log field with the given symbol name that can be used in
+:file:`logging.yaml` log formats. The log field outputs the generated fingerprint value for each
+transaction. If not specified, no custom log field is registered.
+
+For example, if you specify ``--log-field jaxja4``, you can use ``%<jaxja4>`` in your log format
+string in :file:`logging.yaml`.
+
 
 Plugin Behavior
 ===============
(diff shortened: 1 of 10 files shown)


maskit:
> test_cache_CacheAggregateWriteBuffer (Timeout)

Seems unrelated.

maskit:
[approve ci rocky]

bneradt (dismissed):

bneradt (approved):

cmcfarlen:
Removed from project 10.2 as it doesn't support custom log fields.

cmcfarlen:
Cherry-picked to 10.2.x

PR #13087, opened 2026-04-14
Fix AuTest with ssl_multicert

masaori335:
Some AuTests are using `ssl_multicert.config` which is replaced by `ssl_multicert.yaml` recently.
- `h2_malformed_request_logging.test.py‎`
- `tunnel_active_timeout.test.py`


brbzull0 (approved):
lgtm.
Thanks.

PR #13089, opened 2026-04-14
Fix cache_read_vc assertion crash in redirect flow

bneradt:
This builds on top of #12182

This fixes a crash seen in docs where the assertion at 
HttpCacheSM.cc:137 fails:

```
  ink_assert((cache_read_vc == nullptr) ||
             master_sm->t_state.redirect_info.redirect_in_process)
```

The crash occurs when a redirect cache lookup completes after
redirect_in_process has been cleared, but cache_read_vc from the
original request is still set. The root cause is that reset() only
resets captive_action but does not close the existing cache_read_vc
or cancel any pending retry events.

This patch adds the missing cleanup to reset(): it now closes any
existing cache read VC and cancels pending retry events before
starting a new cache operation. This ensures stale state from a
previous cache operation cannot interfere with the new one.

### DIFF
--- a/src/proxy/http/HttpCacheSM.cc
+++ b/src/proxy/http/HttpCacheSM.cc
@@ -70,13 +70,24 @@ HttpCacheAction::cancel(Continuation *c)
 // HttpCacheSM
 //
 /**
-  Reset captive_action and counters for another cache operations.
-  - e.g. following redirect starts over from cache lookup
+  Reset state for another cache operation (e.g., following a redirect).
+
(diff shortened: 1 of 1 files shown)


masaori335 (approved):
Makes sense. Thank you!

Probably, this is what I missed with #12182 and observed a leak of `openDirEntry`, which is fixed by #12205.

masaori335:
[approve ci]

cmcfarlen:
Cherry-picked to 10.2.x

PR #13094, opened 2026-04-16
Guard HTTP/3 test targets with BUILD_TESTING

bneradt:
Guard add_catch2_test for h3 with BUILD_TESTING. Otherwise
-DBUILD_TESTING=off fails when HTTP/3 builds are enabled.


maskit (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #13099, opened 2026-04-17
Replace echo command on autest test cases with printf

maskit:
The -e option is unavailable on echo command on mac.


bneradt (approved):

cmcfarlen:
added to milestone 10.2.0 via #13111

PR #13100, opened 2026-04-17
Fix autest tls_sni_ticket so it can work with BoringSSL

maskit:
The per-SNI override is unavailable with BoringSSL. The expected number of tickets should be different for BoringSSL.


bneradt (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #13110, opened 2026-04-21
Add `_reload` directive support for config reload framework.

brbzull0:
### TL;DR

Adds a generic way to pass operational parameters to config reload handlers.
Instead of per-config flags added to traffic_ctl, handlers now receive directives
via a reserved `_reload` key that the framework extracts automatically.

```bash
traffic_ctl config reload -D virtualhost.id=myhost.example.com
```

Handlers read directives with `ctx.reload_directives()["id"]`, while
`ctx.supplied_yaml()` stays clean (content only, no directives). Also fixes a bug
where stale RPC-supplied config data could replay on subsequent file-based reloads.

---

### Summary

Config handlers sometimes need operational parameters that modify **how** a reload
is performed — for example, scoping a reload to a single entry by ID, or some other key to pass. 
Currently there's no standard way to pass these parameters;
PR #13108 (virtualhost) works around this by adding a per-config `--virtualhost` flag
to `traffic_ctl`. 

This PR adds a generic **reload directive** mechanism:

- A reserved `_reload` key inside each config's YAML node carries directives
- The framework extracts `_reload` before invoking the handler
- `traffic_ctl` gains a `--directive` (`-D`) flag with dot-notation syntax
- Handlers access directives via `ConfigContext::reload_directives()`

### Motivation

The virtualhost PR (#13108) needs to tell the handler *which* virtualhost to reload
from file. Without a generic mechanism, every config that needs parameters would require
its own `traffic_ctl` flag and custom parsing — leading to flag proliferation and
duplicated code.

The `_reload` directive solves this once for all handlers.

#### `traffic_ctl` CLI

```bash
# Reload a specific virtualhost from file
traffic_ctl config reload -D virtualhost.id=myhost

# Multiple directives (space-separated after single -D)
traffic_ctl config reload -D virtualhost.id=myhost virtualhost.foo=bar

# Directives for different handlers in the same reload
traffic_ctl config reload -D virtualhost.id=myhost sni.fqdn=example.com

# Complex directives via -d (full YAML)
traffic_ctl config reload -d 'virtualhost: { _reload: { id: myhost } }'
```

#### Wire format

`-D virtualhost.id=myhost` produces:

```json
{
  "configs": {
    "my_config": {
      "_reload": { "id": "myhost" }
      "param1":{}
    }
  }
}
```

#### Handler API

```cpp
void MyConfig::reconfigure(ConfigContext &ctx) {
    ctx.in_progress();

    if (auto directives = ctx.reload_directives()) {
        if (directives["id"]) {
            std::string id = directives["id"].as<std::string>();
            if (!reload_single_entry(id)) {
                ctx.fail("Unknown entry: " + id);
                return;
            }
            ctx.complete("Reloaded entry: " + id);
            return;
        }
    }

    if (!load_from_file(config_filename)) {
        ctx.fail("Failed to parse " + config_filename);
        return;
    }
    ctx.complete("Loaded from file");
}
```

Key behavior:

- `reload_directives()` returns the extracted `_reload` YAML map
- `supplied_yaml()` is clean — never contains `_reload` keys
- Child contexts inherit directives from their parent
- If `_reload` is not a YAML map, a warning is logged and directives are ignored

#### Framework (`src/mgmt/config/`)

- **`ConfigContext`**: Added `reload_directives()` getter, `set_reload_directives()`
  setter, and propagation to child contexts via `add_dependent_ctx()`
- **`ConfigRegistry::execute_reload()`**: Extracts `_reload` from the config's YAML
  node before passing content to the handler. Also fixes a bug where `_passed_configs`
  entries were not erased after consumption, causing stale RPC data to replay on
  subsequent file-based reloads

#### CLI (`src/traffic_ctl/`)

- **`traffic_ctl.cc`**: Added `--directive` (`-D`) option to `config reload`
- **`CtrlCommands.cc`**: Added `parse_directive()` helper that parses
  `config_key.directive_key=value` and injects into `configs[key]["_reload"][dir] = val`

### DIFF
--- a/doc/appendices/command-line/traffic_ctl.en.rst
+++ b/doc/appendices/command-line/traffic_ctl.en.rst
@@ -454,9 +454,11 @@ Display the current value of a configuration record.
 
       .. note::
 
-         ``-D`` and ``-d`` cannot be combined in the same invocation due to argument
-         parsing constraints. Use ``-d`` with full YAML when you need both directives
-         and inline content in a single reload request.
+         ``-D`` uses variable-argument parsing and must appear as the **last option**
+         on the command line. Any flags placed after ``-D`` will be consumed as directive
+         values. ``-D`` and ``-d`` cannot be combined in the same invocation due to this
+         same constraint. Use ``-d`` with full YAML when you need both directives and
+         inline content in a single reload request.
 
       .. note::
 
--- a/doc/appendices/command-line/traffic_ctl.en.rst
+++ b/doc/appendices/command-line/traffic_ctl.en.rst
@@ -414,6 +414,56 @@ Display the current value of a configuration record.
          will return an error for the corresponding key. The JSONRPC response will contain
          per-key error details.
 
+   .. option:: --directive, -D <config_key.directive_key=value>
+
+      Pass a reload directive to a specific config handler. Directives are operational parameters
+      that modify how the handler performs the reload — for example, scoping a reload to a single
+      entry or enabling a dry-run mode. They are distinct from config content (``-d``).
+
+      The format is ``config_key.directive_key=value``, parsed by splitting on the first ``.``
+      and the first ``=``:
+
+      - ``config_key`` — the registry key (e.g. ``ip_allow``, ``sni``)
+      - ``directive_key`` — the directive name understood by that handler
+      - ``value`` — the directive value (always passed as a string on the wire)
+
+      Multiple directives are passed as space-separated values after a single ``-D``:
+
+      .. code-block:: bash
+
+         # Single directive
+         $ traffic_ctl config reload -D myconfig.id=foo
+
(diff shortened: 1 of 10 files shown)


brbzull0:
[approve ci autest 1]

brbzull0:
[approve ci autest 1]

cmcfarlen (approved):
Seems this is just the initial support, no additional reload functionality that uses this yet.

brbzull0:
> Seems this is just the initial support, no additional reload functionality that uses this yet.

exactly.

cmcfarlen:
Cherry-picked to the 10.2.x branch as a32fadebfa8b01dc0c66323eacfcb7bbcc1b7913 for the 10.2.1 release.

PR #13112, opened 2026-04-21
Add doxygen comment guidance to AGENTS.md

bneradt:
This adds a Doxygen Comments section to AGENTS.md describing the conventions agents should follow when documenting code with doxygen style comments.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -308,6 +308,38 @@ SMDebug(dbg_ctl, "Processing request for URL: %s", url);
 - UPPER_CASE for macros and constants: `HTTP_SM_SET_DEFAULT_HANDLER`
 - Private member variables have the `m_` prefix.
 
+**Doxygen Comments:**
+
+When adding doxygen comments:
+
+- `@brief` is assumed for the first sentence, so give a brief summary right
+  after `/** ` without using `@brief`.
+- In the description of classes, functions, and member variables, convey the
+  responsibility of the item being described (its role and intent), not just
+  what the code obviously does.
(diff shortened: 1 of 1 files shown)


masaori335 (approved):
Cool. Thank you.

AGENTS.md
```diff
@@ -308,6 +308,38 @@ SMDebug(dbg_ctl, "Processing request for URL: %s", url);
...
+When adding doxygen comments:
+
+- `@brief` is assumed for the first sentence, so give a brief summary right
+  after `/** ` without using `@brief`.
+- In the description of classes, functions, and member variables, convey the
+  responsibility of the item being described (its role and intent), not just
+  what the code obviously does.
+- Use `@a <name>` to reference a function argument by name in prose
+  (e.g. "If @a size is zero..."). Use `@c <text>` for inline code or
+  constants (e.g. `@c true`, `@c NULL`, `@c MyEnum::VALUE`).
```
masaori335, line 322: Sometime, too many `@c` make me dizzy to read in text.

```
 * Handles two transitions:
 * - @c UP → @c DOWN: @c _last_failure is @c TS_TIME_ZERO; set via CAS.
 * - @c SUSPECT → @c DOWN: @c fail_window has elapsed since the last failure; @c _last_failure is
 *   refreshed via CAS to restart the fail window.
```

bneradt: Yes, it can be overdone.

Let's move forward with this - it should be an improvement on AI behavior over the current state of things. We can add some more direction later if we see it making an unreasonable amount of decorations.

PR #13116, opened 2026-04-23
header_rewrite: fix set-body origin replacement across hooks and transforms

bryancall:
## Summary

- make `set-body` robust for origin responses by correctly handling internal body replacement before tunneling, including transform-bypass safety in `HttpSM`
- add `READ_RESPONSE_HDR_HOOK` as an allowed `set-body` hook while preserving `SEND_RESPONSE_HDR_HOOK`, and fix API state interaction in `TSHttpTxnErrorBodySet()`
- consolidate AuTests into one replay-driven test covering read/send hooks with and without transforms, with cache disabled and cache-bypass probes; update docs accordingly

## Production validation

Tested with **live traffic on a single production edge/cache host** using a build from this change set: staged **`header_rewrite`**, then **`traffic_server`**, with drain/restart and post-restart checks; exercised **`set-body`** on **`READ_RESPONSE_HDR`** for origin **403** responses via config reload. Behavior matched expectations during rollout observation; the host was returned to rotation and **informal bake/monitoring continues**.

**Ready for review / merge** from the author side.

## Test plan

- [x] `AUTEST_PORT_OFFSET=13000 ./autest.sh --ats-bin /tmp/ts-autest/bin --filter=header_rewrite_set_body_origin` (eris)
- [x] `AUTEST_PORT_OFFSET=15000 ./autest.sh --ats-bin /tmp/ts-autest/bin --filter=header_rewrite_bundle` (eris)

Supersedes #12879.


src/proxy/http/HttpSM.cc
```diff
@@ -1687,9 +1687,23 @@ HttpSM::handle_api_return()
...
   switch (t_state.next_action) {
   case HttpTransact::StateMachineAction_t::TRANSFORM_READ: {
-    HttpTunnelProducer *p = setup_transfer_from_transform();
-    perform_transform_cache_write_action();
-    tunnel.tunnel_run(p);
+    if (t_state.internal_msg_buffer && !t_state.api_server_request_body_set) {
+      SMDbg(dbg_ctl_http, "plugin set internal body, bypassing response transform for internal transfer");
+      transform_info.vc                    = nullptr;
+      t_state.api_info.cache_untransformed = true;
+      if (t_state.hdr_info.client_response.valid() == 0 && t_state.hdr_info.transform_response.valid()) {
```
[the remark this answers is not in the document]
bryancall: Fixed in 6a79af5cb2: when bypassing in TRANSFORM_READ, we now kill any active tunnel and clean up transform vc_table state before switching to internal transfer. This prevents stale active-tunnel state while honoring set-body override.

tests/gold_tests/pluginTest/header_rewrite/header_rewrite_set_body_origin.replay.yaml
```diff
@@ -0,0 +1,345 @@
...
+      url: /cache_probe_read/
+      headers:
+        fields:
+        - [ Host, www.example.com ]
+        - [ uuid, 4 ]
+
+    server-response:
+      status: 200
+      reason: OK
+      headers:
```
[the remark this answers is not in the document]
bryancall: Good catch. Fixed in 6a79af5cb2: cache-probe second responses in blocks 3a/3b now remain non-200 so null_transform does not engage, keeping those probes on the intended no-transform path.

doc/admin-guide/plugins/header_rewrite.en.rst
```diff
@@ -1177,8 +1177,23 @@ set-body
...
+``READ_RESPONSE_HDR_HOOK`` when possible so body replacement happens before
+response body tunneling starts.
+
+.. note::
+
+   When ``set-body`` replaces an origin response body, ATS emits the replacement
+   through its internal error-body path. ``Content-Type`` defaults to
+   ``text/html`` unless you override it with ``set-header Content-Type``.
+   Also, ``set-body ""`` does not suppress an origin response body on this hook;
+   use a non-empty replacement value when sanitizing origin responses.
```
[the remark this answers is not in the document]
bryancall: Clarified in 6a79af5cb2: the docs now explicitly state that set-body "" clears only the internal replacement body and does not suppress an origin response body on this hook.

src/proxy/http/HttpSM.cc
```diff
@@ -7578,12 +7617,19 @@ HttpSM::setup_client_request_plugin_agents(HttpTunnelProducer *p, int num_header
...
-      t_hook = t_hook->m_link.next;
+      APIHook *next = t_hook->m_link.next;
+      // Some transform hooks can already be detached by the time kill_this() runs.
+      // Guard against null continuations while still draining the remaining hooks.
+      if (auto *t_vcon = static_cast<VConnection *>(t_hook->m_cont); t_vcon != nullptr) {
+        t_vcon->do_io_close();
+      }
+      t_hook = next;
     } while (t_hook != nullptr);
   }
```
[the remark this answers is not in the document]
bryancall: Validated this path with AuTests. Removing the info->entry guard reintroduced the same -11 crash in transform_cleanup() for header_rewrite_set_body_origin. Kept the guard, added rationale comments, and re-ran header_rewrite_set_body_origin + filter_body + full ctest + -R 1 regression in this round.

src/proxy/http/HttpSM.cc
```diff
@@ -7664,7 +7710,11 @@ HttpSM::kill_this()
     //   In that case, we need to manually close all the
     //   transforms to prevent memory leaks (INKqa06147)
     if (hooks_set) {
-      transform_cleanup(TS_HTTP_RESPONSE_TRANSFORM_HOOK, &transform_info);
+      bool bypassed_response_transform =
+        t_state.api_info.cache_untransformed && t_state.internal_msg_buffer && !t_state.api_server_request_body_set;
+      if (!bypassed_response_transform) {
+        transform_cleanup(TS_HTTP_RESPONSE_TRANSFORM_HOOK, &transform_info);
+      }
```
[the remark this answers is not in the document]
bryancall: Validated this suggestion directly: always invoking response transform_cleanup() in the bypass case reintroduced a transform_cleanup() segfault (-11). Keeping the bypass-specific skip here is intentional, and the code now documents why. Tests rerun: header_rewrite_set_body_origin, filter_body, full ctest, full -R 1 regression.

tests/gold_tests/pluginTest/header_rewrite/header_rewrite_set_body_origin.replay.yaml
```diff
@@ -0,0 +1,347 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+meta:
+  version: "1.0"
+
+autest:
+  description: 'Test set-body origin replacement without response transforms'
```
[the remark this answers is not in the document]
bryancall: Good catch. Updated replay metadata wording to reflect that null_transform is loaded globally, but selected blocks intentionally keep transform inactive. Included in commit 6f2dfda5e7.

tests/gold_tests/pluginTest/header_rewrite/header_rewrite_set_body_origin.replay.yaml
```diff
@@ -0,0 +1,347 @@
...
+  client:
+    name: 'client'
+    process_config:
+      other_args: '--thread-limit 1'
+
+  ats:
+    name: 'ts'
+
+    plugin_config:
+      - null_transform.so
```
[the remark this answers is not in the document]
bryancall: Addressed with the same clarification update: replay description/comments now explicitly describe transform-loaded vs transform-inactive paths. Included in commit 6f2dfda5e7.

tests/gold_tests/pluginTest/header_rewrite/header_rewrite_set_body_origin.replay.yaml
```diff
@@ -0,0 +1,347 @@
...
+
+    copy_to_config_dir:
+      - 'rules'
+
+    records_config:
+      proxy.config.http.insert_response_via_str: 0
+      proxy.config.http.cache.http: 0
```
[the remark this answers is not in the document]
bryancall: Agreed. Renamed these notes from cache-bypass probes to repeated-URL probes so the terminology matches cache-disabled test behavior. Included in commit 6f2dfda5e7.

tests/gold_tests/pluginTest/header_rewrite/header_rewrite_set_body_origin.replay.yaml
```diff
@@ -0,0 +1,347 @@
...
+
+      # Block 2: SEND_RESPONSE hook replacement path (no transform plugin).
+      - from: "http://www.example.com/set_body_send_resp_403/"
+        to: "http://backend.ex:{SERVER_HTTP_PORT}/origin_send_403/"
+        plugins:
+          - name: "header_rewrite.so"
+            args:
+              - "rules/rule_set_body_origin_send_resp.conf"
+
+      # Block 3a: cache-bypass probe for READ_RESPONSE hook (same URL twice).
```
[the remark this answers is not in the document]
bryancall: Addressed. Updated replay comments to repeated-URL probe wording (cache remains intentionally disabled). Included in commit 6f2dfda5e7.

tests/gold_tests/pluginTest/header_rewrite/header_rewrite_set_body_origin.replay.yaml
```diff
@@ -0,0 +1,347 @@
...
+      - from: "http://www.example.com/cache_probe_read/"
+        to: "http://backend.ex:{SERVER_HTTP_PORT}/cache_probe_read/"
+        plugins:
+          - name: "header_rewrite.so"
+            args:
+              - "rules/rule_set_body_origin_read_resp.conf"
+
+      # Block 3b: cache-bypass probe for SEND_RESPONSE hook (same URL twice).
```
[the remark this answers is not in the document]
bryancall: Addressed in the same replay cleanup: renamed cache-bypass wording to repeated-URL probe for correctness. Included in commit 6f2dfda5e7.

doc/admin-guide/plugins/header_rewrite.en.rst
```diff
@@ -1177,8 +1177,24 @@ set-body
 
   set-body <text>
 
-Sets the body to ``<text>``. Can also be used to delete a body with ``""``. This is only useful when overriding the origin status, i.e.
-intercepting/pre-empting a request so that you can override the body from the body-factory with your own.
+Sets the body to ``<text>``. Can also be used to delete a body with ``""``.
```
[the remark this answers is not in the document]
bryancall: Clarified docs to split contexts explicitly: generic set-body behavior vs origin-response replacement behavior for set-body "". Included in commit 6f2dfda5e7.

doc/admin-guide/plugins/header_rewrite.en.rst
```diff
@@ -1177,8 +1177,24 @@ set-body
...
+response body tunneling starts.
+
+.. note::
+
+   When ``set-body`` replaces an origin response body, ATS emits the replacement
+   through its internal error-body path. ``Content-Type`` defaults to
+   ``text/html`` unless you override it with ``set-header Content-Type``.
+   ``set-body ""`` clears the internal replacement body, but does not suppress an
+   origin response body on this hook; use a non-empty replacement value when
+   sanitizing origin responses.
```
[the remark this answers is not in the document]
bryancall: Addressed via docs wording update that distinguishes synthetic/internal replacement semantics from origin-body suppression semantics for set-body "". Included in commit 6f2dfda5e7.

tests/gold_tests/pluginTest/header_rewrite/header_rewrite_set_body_origin.replay.yaml
```diff
@@ -0,0 +1,347 @@
...
+        data: "Sensitive account info: secret-key-12345"
+
+    proxy-response:
+      status: 200
+      headers:
+        fields:
+        - [ Content-Length, { value: "9", as: equal } ]
+      content:
+        size: 9
+        data: "Sanitized"
```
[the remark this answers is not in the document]
bryancall: Added explicit Content-Type=text/html expectations in the 200/transform-active proxy-response assertions to lock in this contract. Included in commit 6f2dfda5e7.

src/proxy/http/HttpSM.cc
```diff
@@ -1722,6 +1750,17 @@ HttpSM::handle_api_return()
       }
 
       setup_blind_tunnel(true, initial_data);
+    } else if (t_state.internal_msg_buffer && !t_state.api_server_request_body_set && t_state.hdr_info.server_response.valid() &&
+               plugin_tunnel == nullptr &&
+               (api_hooks.get(TS_HTTP_READ_RESPONSE_HDR_HOOK) != nullptr ||
```
zwoop: Is it even possible to get here, without one of these being true ? What's the intent here, that the plugin API wasn't called from somewhere else?  But then we wouldn't have a server_response.valid() either.

I don't know for sure, but feels like either this check on the hook is just superfluous, or possibly should be a assert instead?

src/api/InkAPI.cc
```diff
@@ -4948,6 +4948,9 @@ TSHttpTxnErrorBodySet(TSHttpTxn txnp, char *buf, size_t buflength, char *mimetyp
   s->internal_msg_buffer                     = buf;
   s->internal_msg_buffer_size                = buf ? buflength : 0;
   s->internal_msg_buffer_fast_allocator_size = -1;
+  // TSHttpTxnErrorBodySet() and TSHttpTxnServerRequestBodySet() share the same buffer.
```
zwoop, line 4951: Overall, and I'll start that here:

Claude has a big tendency to add comments that were part of its "thinking" phase. Explaining its reasoning for doing stuff. More often than not, I find those comments to be completely unnecessary, and often distracting, when the code itself is self explanatory. In most cases, the comments can be significantly reduced, and sometimes removed.

I'll refer to this as comment slop :).

src/proxy/http/HttpSM.cc
```diff
@@ -1687,9 +1687,37 @@ HttpSM::handle_api_return()
 
   switch (t_state.next_action) {
   case HttpTransact::StateMachineAction_t::TRANSFORM_READ: {
-    HttpTunnelProducer *p = setup_transfer_from_transform();
-    perform_transform_cache_write_action();
-    tunnel.tunnel_run(p);
+    // A plugin has installed an internal response body (TSHttpTxnErrorBodySet()).
```
zwoop: Comment slop, code is self explanatory.

src/proxy/http/HttpSM.cc
```diff
@@ -1687,9 +1687,37 @@ HttpSM::handle_api_return()
...
+    // In this branch we bypass transform streaming and switch to internal transfer.
+    if (t_state.internal_msg_buffer && !t_state.api_server_request_body_set && t_state.hdr_info.server_response.valid()) {
+      SMDbg(dbg_ctl_http, "plugin set internal body, bypassing response transform for internal transfer");
+      t_state.api_info.cache_untransformed = true;
+      // If a tunnel was already set up for transform I/O, shut it down before we re-route.
```
zwoop: Comment slop, code says exactly the same thing as the comment.

src/proxy/http/HttpSM.cc
```diff
@@ -1687,9 +1687,37 @@ HttpSM::handle_api_return()
...
+        vc_table.cleanup_entry(transform_info.entry);
+        transform_info.entry = nullptr;
+      }
+      transform_info.vc = nullptr;
+      // Some downstream paths still read client_response; seed it from transform_response when missing.
+      if (t_state.hdr_info.client_response.valid() == 0 && t_state.hdr_info.transform_response.valid()) {
+        t_state.hdr_info.client_response.create(HTTPType::RESPONSE);
+        t_state.hdr_info.client_response.copy(&t_state.hdr_info.transform_response);
+      }
+      // The server session is not needed for internal body transfer if it was never tunneled.
```
zwoop: Comment slop.

src/proxy/http/HttpSM.cc
```diff
@@ -1687,9 +1687,37 @@ HttpSM::handle_api_return()
...
+      if (server_entry != nullptr && server_entry->in_tunnel == false) {
+        release_server_session();
+      }
+      // Serve the plugin-provided body through the internal tunnel handler.
```
zwoop, line 1712: Comment slop.

src/proxy/http/HttpSM.cc
```diff
@@ -1722,6 +1750,17 @@ HttpSM::handle_api_return()
       }
 
       setup_blind_tunnel(true, initial_data);
+    } else if (t_state.internal_msg_buffer && !t_state.api_server_request_body_set && t_state.hdr_info.server_response.valid() &&
+               plugin_tunnel == nullptr &&
+               (api_hooks.get(TS_HTTP_READ_RESPONSE_HDR_HOOK) != nullptr ||
+                api_hooks.get(TS_HTTP_SEND_RESPONSE_HDR_HOOK) != nullptr)) {
+      // A plugin replaced the origin response body via TSHttpTxnErrorBodySet().
```
zwoop: Comment slop

src/proxy/http/HttpSM.cc
```diff
@@ -7578,12 +7617,21 @@ HttpSM::setup_client_request_plugin_agents(HttpTunnelProducer *p, int num_header
 inline void
 HttpSM::transform_cleanup(TSHttpHookID hook, HttpTransformInfo *info)
 {
+  // Internal-body bypass can skip transform tunnel setup, leaving no transform
+  // entry to clean up. In that case there is nothing safe/useful to close here.
```
zwoop: Comment slop.

src/proxy/http/HttpSM.cc
```diff
@@ -7578,12 +7617,21 @@ HttpSM::setup_client_request_plugin_agents(HttpTunnelProducer *p, int num_header
...
+  }
   APIHook *t_hook = api_hooks.get(hook);
   if (t_hook && info->vc == nullptr) {
     do {
-      VConnection *t_vcon = t_hook->m_cont;
-      t_vcon->do_io_close();
-      t_hook = t_hook->m_link.next;
+      APIHook *next = t_hook->m_link.next;
+      // Some transform hooks can already be detached by the time kill_this() runs.
+      // Guard against null continuations while still draining the remaining hooks.
```
zwoop: The first comment seems useful, the second is slop.

src/proxy/http/HttpSM.cc
```diff
@@ -7664,7 +7712,14 @@ HttpSM::kill_this()
     //   In that case, we need to manually close all the
     //   transforms to prevent memory leaks (INKqa06147)
     if (hooks_set) {
-      transform_cleanup(TS_HTTP_RESPONSE_TRANSFORM_HOOK, &transform_info);
+      bool bypassed_response_transform =
+        t_state.api_info.cache_untransformed && t_state.internal_msg_buffer && !t_state.api_server_request_body_set;
+      // If we intentionally bypassed response transforms for internal-body
+      // transfer, transform_info may be partially detached; skip cleanup in this
```
zwoop: Partial comment slop, can be reduced to one line.

tests/prepare_proxy_verifier.sh
```diff
@@ -40,7 +40,7 @@ pv_dir="${pv_name}-${pv_version}"
 pv_tar_filename="${pv_dir}.tar.gz"
 pv_tar="${pv_top_dir}/${pv_tar_filename}"
 pv_tar_url="https://ci.trafficserver.apache.org/bintray/${pv_tar_filename}"
-expected_sha1="e11b5867a56c5ffd496b18c901f1273e9c120a47"
+expected_sha1="0a60c646cbc9326abb2fbc397cb9efa8c08a807a"
```
zwoop: What happened here? We bumping proxy verifier version or something as part of this test suite? Feels like that shouldn't be the case, and if we need to, should be a separate PR ?

zwoop (changes requested):
I think this looks good, complex code, so love the tests. Other than all the comment slop, my only question really is around that hook check. That seems unnecessary, albeit I'm not 100% sure. Possibly they should be assertions, but I don't see how you'd get there, if you don't have those plugin hooks.

src/api/InkAPI.cc
bryancall: I had Claude add comments to the code since this is complex code and to document what the code is for and what it is doing.   It didn't have comments before.

src/proxy/http/HttpSM.cc
bryancall: Good catch. Traced `do_api_callout_internal()` (HttpSM.cc:5905) and confirmed `cur_hook_id` is tautologically `TS_HTTP_READ_RESPONSE_HDR_HOOK` when we reach `next_action == SERVER_READ` (set unconditionally on that path), and the `SEND_RESPONSE_HDR_HOOK` arm was dead because `API_SEND_RESPONSE_HDR` never transitions to `SERVER_READ`. The real guards against TSHttpConnect intercept regressions are `plugin_tunnel == nullptr` and `server_response.valid()`. Dropped the entire hook sub-clause in d8ef38db29.

src/api/InkAPI.cc
bryancall: Kept this 2-line comment because it documents a non-obvious cross-API coupling: `TSHttpTxnErrorBodySet` and `TSHttpTxnServerRequestBodySet` share the same `internal_msg_buffer`, with `api_server_request_body_set` as the mode flag. Without the comment, a future reader has no signal that the two TS APIs are coupled. Took the meta point on comment slop elsewhere — see the rest of the diff in d8ef38db29.

src/proxy/http/HttpSM.cc
bryancall: Compressed the 2-line intro to one line in d8ef38db29: `// Bypass transform streaming when a plugin replaced the response body.`

src/proxy/http/HttpSM.cc
bryancall: Dropped in d8ef38db29. Agreed — code is self-evident.

src/proxy/http/HttpSM.cc
bryancall: Compressed to one tight line in d8ef38db29: `// Internal transfer doesn't use the server session.`

src/proxy/http/HttpSM.cc
bryancall: Kept this one — function name says what, comment notes WHY the plugin-provided body is being served through the internal tunnel handler. Open to dropping if you'd rather; the surrounding block context makes it clear enough that I could go either way.

src/proxy/http/HttpSM.cc
bryancall: Compressed to one line alongside dropping the hook sub-clause in d8ef38db29: `// Plugin replaced the origin response body via TSHttpTxnErrorBodySet(); divert to internal transfer.`

src/proxy/http/HttpSM.cc
bryancall: Pre-existing — predates this PR. Same wording at HttpSM.cc:7595-7596 too (response plugin agents twin). Out of scope here; happy to do a separate cleanup PR if you'd like.

src/proxy/http/HttpSM.cc
bryancall: Kept the first line (`// Some transform hooks can already be detached by the time kill_this() runs.`) — it's load-bearing for the `t_vcon != nullptr` guard. Removed the second `// Guard against null continuations...` line in d8ef38db29.

src/proxy/http/HttpSM.cc
bryancall: That 3-line comment was introduced in `6f2dfda5e7` which got dropped in a force-push earlier today (cleaning up unrelated commits that had ended up on this branch). The current HEAD has just the `bypassed_response_transform` bool with no comment. Nothing to do here.

tests/prepare_proxy_verifier.sh
bryancall: Agreed — unintentional local-dev drift, not needed upstream. Reverted to the master SHA1 (`e11b5867...`) in d8ef38db29.

bryancall:
Posting a summary of the design discussion plus the production validation we just ran.

## Why the SERVER_READ divert and not a response transform

`TSHttpTxnErrorBodySet()` from a response-stage hook only worked end-to-end when the origin returned an error path; against a valid origin response the SM took the server-transfer branch and the plugin's body was discarded. The fix in this PR is a precise divert in `HttpSM::handle_api_return()` when the plugin set an internal body, the origin response is valid, no `plugin_tunnel` is in play, and the recorded hook id at `TSHttpTxnErrorBodySet()` time was `READ_RESPONSE_HDR` or `SEND_RESPONSE_HDR`.

The natural alternative is to do what `txn_box`'s `upstream-rsp-body` does — install a `TS_HTTP_RESPONSE_TRANSFORM_HOOK` and rewrite body bytes inline. We considered that and chose the SERVER_READ divert for these reasons.

### Why the SERVER_READ divert

- **Fixes the API, not just one plugin.** Any plugin calling `TSHttpTxnErrorBodySet()` from a response hook gets correct behavior, not only `header_rewrite`.
- **Smaller, contained patch.** The change is localized to `handle_api_return()` plus a captured-hook-id field for scoping. No new plugin code, no new vconn lifecycle.
- **Response headers are not perturbed by the body swap.** Only the body bytes are substituted; status code, Content-Type, Cache-Control, and any other plugin-modified headers flow through the normal response path. A transform forces the plugin to recompute `Content-Length` and decide chunked vs. not-chunked encoding; the divert avoids that.
- **Origin connection is closed for the substituted transaction.** The server session is released before the body is consumed, so the replacement body goes to the client and the origin body is discarded. Drain-and-pool (so the connection can be reused) is a possible follow-up enhancement; this PR takes the simpler close path.

### Trade-offs we accept

- Touches `HttpSM` core, which is what made review careful. The divert is gated on five conditions (`internal_msg_buffer`, `!api_server_request_body_set`, `server_response.valid()`, `plugin_tunnel == nullptr`, captured-hook-id in the response stage) to keep the blast radius narrow.
- `plugin_tunnel == nullptr` is the explicit `TSHttpConnect` intercept guard, replacing the indirect `api_hooks.get()` check.
- The captured-hook-id check (`internal_msg_buffer_set_on`) scopes the divert to plugin-driven response replacement.

### Why not a response transform

- It is a substantially larger plugin rewrite: transform vconn lifecycle, chunked-encoding handling, Content-Length recomputation, abort cleanup.
- It does not address the underlying gap in `TSHttpTxnErrorBodySet()` behavior, so the next plugin to hit this learns the same lesson.
- Cache-write semantics differ (transformed vs untransformed body), which is configuration users have to think about explicitly.

## Production validation

Validated on an internal production edge host running a build that contains this PR's `handle_api_return()` change (the same captured-hook-id divert logic). The host had been running this build for 65+ hours stably prior to the test.

**Method:** appended three header-gated `header_rewrite` rules to an existing per-remap `header_rewrite` config (rules only fire when the client sends a magic `X-CDN-Set-Body-Test: 1` header, so real user traffic is unaffected), bumped `remap.config` mtime, ran `traffic_ctl config reload`. Drove ten curl requests with the magic header against a test endpoint where the origin returns a real response body. Reverted the config and reloaded again.

**Result:**

- All ten test requests returned HTTP 200 with the `header_rewrite` replacement body (41 bytes, the literal test string) instead of the origin body. Status code, response headers, and connection lifecycle were the expected ones for the divert path: the server session for each transaction was released, no transform was installed, and no client-side errors were observed.
- One verification request sent *without* the magic header returned the original origin body unchanged (full size), confirming the rule only fires on the gated condition.
- After reverting the rule and waiting for `traffic_ctl config reload` to propagate, the same test request returned the original origin body. No lingering side effects from the experiment.
- Background traffic on the host (not carrying the magic header) was unaffected throughout.

## Automated test coverage

- `header_rewrite_set_body_origin` AuTest passes, including the `null_transform` interaction case (TRANSFORM_READ branch).
- Full SDK regression suite passes: `SDK_API_TSHttpConnectIntercept`, `SDK_API_TSHttpConnectServerIntercept`, `SDK_API_HttpAltInfo`, `SDK_API_HttpTxnTransform`, `SDK_API_HttpTxnCache`, `SDK_API_HttpSsn`, `SDK_API_HttpHookAdd`.
- Exported symbols of `libtsapi` / `libtscore` / `libtsutil` / `libtscppapi` are unchanged (0 added, 0 removed).

PR #13119, opened 2026-04-27
Restore a shortcut in hot loop in _mime_hdr_field_list_search_by_string

masaori335:
@c-taylor noticed the 10.2.x branch has regression of throughput performance. It looks like several changes made it but the first big drop is introduced by `3b0f0b760387af0c4f6fdd40c4bc98f14dbf0c4d`. The shortcut in the hot loop seems important.

Below is comparison of 1 URL against 1 ET_NET thread performance.

commit | req/sec | note
-- | -- | --
086847590c | 14683.37 | (parent of 3b0f0b7603)
3b0f0b7603 | 11024.33 |  
3b0f0b7603 + this patch | 13535.39 |

### DIFF
--- a/include/tsutil/StringCompare.h
+++ b/include/tsutil/StringCompare.h
@@ -1,0 +1,49 @@
+/** @file
+
+  Helper for std::string_view comparison
+
+  @section license License
+
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+ */
+
+#pragma once
+
+#include <strings.h>
+#include <string_view>
+
+namespace ts
+{
+/**
+  Returns true iff @a lhs and @a rhs compare equal, ignoring case.
+
+  Prefer this over libswoc's @c strcasecmp(std::string_view, std::string_view) when you only need
+  an equality check: this short-circuits on length mismatch, whereas the libswoc version must keep
+  comparing bytes to produce a correct ordering result even when the lengths differ.
+
+  For case-sensitive comparison, use @c std::string_view::operator==.
+ */
+inline bool
+iequals(const std::string_view &lhs, const std::string_view &rhs)
+{
+  if (lhs.size() != rhs.size()) {
+    return false;
+  }
+
+  return ::strncasecmp(lhs.data(), rhs.data(), lhs.size()) == 0;
+}
+} // namespace ts
--- a/include/tsutil/StringCompare.h
+++ b/include/tsutil/StringCompare.h
@@ -38,7 +38,7 @@ namespace ts
   For case-sensitive comparison, use @c std::string_view::operator==.
  */
 inline bool
-iequals(const std::string_view &lhs, const std::string_view &rhs)
+iequals(std::string_view lhs, std::string_view rhs) noexcept
 {
   if (lhs.size() != rhs.size()) {
     return false;
--- a/src/proxy/hdrs/MIME.cc
+++ b/src/proxy/hdrs/MIME.cc
@@ -1160,7 +1160,7 @@ _mime_hdr_field_list_search_by_string(MIMEHdrImpl *mh, std::string_view field_na
 
     too_far_field = &(fblock->m_field_slots[fblock->m_freetop]);
     while (field < too_far_field) {
-      if (field->is_live() &&
+      if (field->is_live() && field->m_len_name == field_name.length() &&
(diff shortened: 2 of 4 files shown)


bryancall:
What strcasecmp() is it calling and look at inlining the function?

masaori335:
This `strcasecmp` in the loop calls our `strcasecmp(const std::string_view &lhs, const std::string_view &rhs)` in the libswoc, which is inlined.

The problem is that this `strcasecmp` almost always call libc's `::strncasecmp`, because the size comparison inside the function is only done to determine the `size_t n` passed as the 3rd argument. ( I thought the size comparison does early return in the beginning, but it doesn't )
https://github.com/apache/trafficserver/blob/a26a2d84595c1038a4061d4a7f4a2c4f55178d81/lib/swoc/src/string_view_util.cc#L29-L47

masaori335:
I introduced a new helper function, `bool iequals(const std::string_view &lhs, const std::string_view &rhs)`, simple equality check case-insensitivly.

I found 3b0f0b760387af0c4f6fdd40c4bc98f14dbf0c4d did the same change in three places, so this PR restores the length check with `ts::iequals`.
```
git show 3b0f0b760387af0c4f6fdd40c4bc98f14dbf0c4d | grep -B 10 strcasecmp
 {
   MIMEFieldBlockImpl *fblock;
   MIMEField          *field, *too_far_field;
@@ -1277,8 +1117,9 @@ _mime_hdr_field_list_search_by_string(MIMEHdrImpl *mh, const char *field_name_st
 
     too_far_field = &(fblock->m_field_slots[fblock->m_freetop]);
     while (field < too_far_field) {
-      if (field->is_live() && (field_name_len == field->m_len_name) &&
-          (strncasecmp(field->m_ptr_name, field_name_str, field_name_len) == 0)) {
+      if (field->is_live() &&
+          strcasecmp(std::string_view{field->m_ptr_name, static_cast<std::string_view::size_type>(field->m_len_name)},
--
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -735,9 +735,9 @@ HttpSM::state_read_client_request_header(int event, void *data)
     if (t_state.hdr_info.client_request.version_get() == HTTP_1_1 &&
         (t_state.hdr_info.client_request.method_get_wksidx() == HTTP_WKSIDX_POST ||
          t_state.hdr_info.client_request.method_get_wksidx() == HTTP_WKSIDX_PUT)) {
-      int         len    = 0;
-      const char *expect = t_state.hdr_info.client_request.value_get(MIME_FIELD_EXPECT, MIME_LEN_EXPECT, &len);
-      if ((len == HTTP_LEN_100_CONTINUE) && (strncasecmp(expect, HTTP_VALUE_100_CONTINUE, HTTP_LEN_100_CONTINUE) == 0)) {
+      auto expect{t_state.hdr_info.client_request.value_get(static_cast<std::string_view>(MIME_FIELD_EXPECT))};
+      if (strcasecmp(expect, std::string_view{HTTP_VALUE_100_CONTINUE,
--
@@ -882,11 +888,11 @@ HttpTransactHeaders::remove_conditional_headers(HTTPHdr *outgoing)
 void
 HttpTransactHeaders::remove_100_continue_headers(HttpTransact::State *s, HTTPHdr *outgoing)
 {
-  int         len    = 0;
-  const char *expect = s->hdr_info.client_request.value_get(MIME_FIELD_EXPECT, MIME_LEN_EXPECT, &len);
+  auto expect{s->hdr_info.client_request.value_get(static_cast<std::string_view>(MIME_FIELD_EXPECT))};
 
-  if ((len == HTTP_LEN_100_CONTINUE) && (strncasecmp(expect, HTTP_VALUE_100_CONTINUE, HTTP_LEN_100_CONTINUE) == 0)) {
-    outgoing->field_delete(MIME_FIELD_EXPECT, MIME_LEN_EXPECT);
+  if (strcasecmp(expect,
```

There're so many other libswoc `strcasecmp` usages we can replace, but it's out of scope from this PR.

include/tsutil/StringCompare.h
```diff
@@ -0,0 +1,49 @@
...
+  Returns true iff @a lhs and @a rhs compare equal, ignoring case.
+
+  Prefer this over libswoc's @c strcasecmp(std::string_view, std::string_view) when you only need
+  an equality check: this short-circuits on length mismatch, whereas the libswoc version must keep
+  comparing bytes to produce a correct ordering result even when the lengths differ.
+
+  For case-sensitive comparison, use @c std::string_view::operator==.
+ */
+inline bool
+iequals(std::string_view lhs, std::string_view rhs) noexcept
```
masaori335, line 41: Standard library doesn't have this, but boost has one. I followed the naming.
https://www.boost.org/doc/libs/latest/libs/beast/doc/html/beast/ref/boost__beast__iequals.html

bryancall: Does libswoc already have a compare like this?

masaori335: CaselessEqual is similar, but it's a functor on TextView (won't accept a plain std::string_view). IMO worth adding ts::iequals for simplicity and cleanness.

https://github.com/apache/trafficserver/blob/4bba67846c05ddb78163a4143deddd7245cb6865/lib/swoc/include/swoc/TextView.h#L963-L970

bneradt (approved):
Looks reasonable. Let's merge this in as is.

There's a lot of `strcasecmp` in our code. Should we more generally replace these?

masaori335:
I do agree with we should replace all `str(n)casecmp() == 0` with this mechanically.

cmcfarlen:
Cherry-picked to 10.2.x

PR #13120, opened 2026-04-27
Propagate PROXY-Protocol src to outbound surfaces

bneradt:
On listeners that receive PROXY-Protocol, the parsed source IP currently does not reach several origin-facing and operator-visible surfaces: the outbound `Client-ip` and `X-Forwarded-For` request headers always carry the immediate TCP peer (the CDN edge), and the outbound TPROXY local-bind address, the HostDB parent-selection affinity hash, and the Slow-Request error log all key off the TCP peer even on listeners that have opted into `:pp-clnt`. The `ssl_has_proxy_v1` debug line is also misleading: it fires for both PPv1 and PPv2 and only logs the destination.

This aligns those outbound surfaces with the PROXY-Protocol source. `HttpTransact::add_client_ip_to_outgoing_request` is updated to prefer `pp_info.src_addr` whenever the user-agent connection has a parsed PROXY-Protocol header, mirroring
`add_forwarded_field_to_request`, so the legacy headers agree with `Forwarded: for=` regardless of `:pp-clnt`. The outbound TPROXY local-bind, HostDB parent-selection affinity, and Slow-Request error log in `HttpSM.cc` are migrated to `t_state.effective_client_addr`, which is `:pp-clnt`-gated by construction so listeners without that flag are unaffected. The PP debug line is rewritten to print the actual PP version together with both `src` and `dst`.

This extends the proxy_protocol autest with a PPv2-over-TLS session using a custom `src-addr` to lock in the new outbound-header behavior, and updates the admin-guide PROXY-Protocol page to clarify which surfaces are `:pp-clnt`-gated and which (`Client-ip`, `X-Forwarded-For`, `Forwarded: for=`) are unconditional.


cmcfarlen (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #13122, opened 2026-04-28
Remove unused error page template

maskit:
The file has never been used since the first commit, and there is no way to use it even via TS API.


brbzull0 (approved):

cmcfarlen:
Cherry-picked to 10.2.x

ezelkow1:
Can you also make a 9.2.x pr for this?

maskit:
I could make ones for older versions but I think having this commit (removing the file) on master and 10.2.x is enough because it's just cleanup.

PR #13123, opened 2026-04-28
Remove virtual dispatch from LogData

bneradt:
This addresses a performance regression added by #13059.

Malformed pre-transaction logging introduced an extra virtual data
interface on the normal access log path. That made every completed
transaction pay for indirection that is only needed for rare
protocol-layer failures.

This replaces the virtual hierarchy with a concrete composed
TransactionLogData wrapper. This keeps LogAccess using one data object
while routing the common HttpSM path through direct getters and falling
back to owned pre-transaction data only when no HttpSM exists.


bneradt:
10.2.x backport:
https://github.com/apache/trafficserver/pull/13124

cmcfarlen (approved):

PR #13130, opened 2026-04-30
Fix HTTP/2 stream (transaction) inactivity timeout

masaori335:
Prior to the change, the activity of the Http2Stream was updated even if `send_a_data_frame` is failed by error like `NO_WINDOW` which breaks inactivity timeout.

It looks like when a write is success, the activity is updated by `Http2Stream::signal_write_event`, so `Http2Stream::send_body` don't need to update it.

https://github.com/apache/trafficserver/blob/83ec495a12b5e8a8795c2feb43289532b5531de7/src/proxy/http2/Http2Stream.cc#L950

### DIFF
--- a/src/proxy/http2/Http2Stream.cc
+++ b/src/proxy/http2/Http2Stream.cc
@@ -975,7 +975,6 @@ void
 Http2Stream::send_body(bool /* call_update ATS_UNUSED */)
 {
   Http2ConnectionState &connection_state = this->get_connection_state();
-  _timeout.update_inactivity();
 
   reentrancy_count++;
 


bryancall (approved):
Removing the unconditional `update_inactivity()` from `send_body()` is the right call: it was refreshing the inactivity timer even when the frame send failed on flow control (`NOT_WRITE_AVAIL` / `NO_WINDOW`), so a stream stalled waiting on a `WINDOW_UPDATE` that never arrives would never time out.

Confirmed the success path still refreshes the timer, so healthy streams won't false-timeout: `send_data_frames()` and `send_data_frames_depends_on_priority()` call `signal_write_event()` with the default `call_update = true`, and that calls `_timeout.update_inactivity()` whenever a write event is actually dispatched (`WRITE_COMPLETE`/`WRITE_READY`). Only the flow-control-stalled path now skips the refresh, which is exactly the case that should be allowed to time out. This also makes the write side consistent with the read path, which already refreshes only on real activity in `signal_read_event()`.

Optional: a replay test that opens a stream, withholds `WINDOW_UPDATE`, and asserts teardown at `transaction_no_activity_timeout_in` would lock this in, though the timing race is admittedly awkward to test deterministically.

cmcfarlen:
Cherry-picked to 10.2.x

PR #13132, opened 2026-04-30
Fix index issue with records in HostDBRecord::select_best_srv

cmcfarlen:
This came up in a copilot review of a fix unrelated to this issue so I created this separate PR to address this.

It seems the loop structure at some point looped over an index, but was changed to a range based loop, neglecting to change one aspect.  But it was not noticed because the index variable is declared before the loop.

The issue is the is_down is always checked on the first target, so could skip up targets or use down targets depending.


masaori335 (approved):
Looks good.

PR #13134, opened 2026-05-01
Add cache key hash logging field and public API

zwoop:
Add 'ckh' log field that emits the base64-encoded cache key digest, and TSHttpTxnCacheKeyDigestGet() for plugin access to the raw hash bytes. Includes admin and API documentation.

Co-Author: Craig Taylor

### DIFF
--- a/include/ts/ts.h
+++ b/include/ts/ts.h
@@ -2776,11 +2776,11 @@ TSReturnCode TSHttpTxnCacheLookupUrlSet(TSHttpTxn txnp, TSMBuffer bufp, TSMLoc o
     for SHA-256 (FIPS mode).  A buffer of at least 32 bytes is
     recommended to accommodate either configuration.
 
-    @param txnp the transaction.
-    @param buffer caller-provided buffer to receive the raw hash bytes.
-      If @c nullptr, only @a length is set (size query).
-    @param length in: capacity of @a buffer in bytes.
-      out: actual digest size in bytes.
+    @param[in] txnp the transaction.
+    @param[out] buffer caller-provided buffer to receive the raw hash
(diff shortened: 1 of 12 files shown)


zwoop:
The use of this new log tag is primarily when the cachekey (or other plugins) modifies the cache key beyond what's in the URL. This allows for data analysis using the unique / identifiable value of the cache key, without spilling any secrets or details into the logs.

include/ts/ts.h
```diff
@@ -2768,6 +2768,25 @@ TSReturnCode TSHttpTxnCachedRespModifiableGet(TSHttpTxn txnp, TSMBuffer *bufp, T
...
+    used for cache lookup or storage on this transaction.  The digest
+    is returned as raw bytes — 16 bytes for MD5 (default) or 32 bytes
+    for SHA-256 (FIPS mode).  A buffer of at least 32 bytes is
+    recommended to accommodate either configuration.
+
+    @param txnp the transaction.
+    @param buffer caller-provided buffer to receive the raw hash bytes.
+      If @c nullptr, only @a length is set (size query).
+    @param length in: capacity of @a buffer in bytes.
+      out: actual digest size in bytes.
```
bneradt: `[in]`, `[out]`, `[in,out]` for `@param` please.

In case it's helpful, we updated the AGENTS.md file for this:
https://github.com/apache/trafficserver/blob/master/AGENTS.md#code-style

But that may have landed after you made this patch.

bneradt (changes requested):
Looks great.

zwoop:
[approve ci]

bneradt (approved):

PR #13138, opened 2026-05-05
Remove cache alternate vector detach

bneradt:
Cache alternate detach is a private helper with no current callers, and
its compaction loop contains a latent out-of-bounds read if the helper is
ever used. Keeping the dead path around preserves a future cache metadata
hazard without providing any current behavior.

This removes the unused helper and its declaration instead of repairing
the unreachable implementation. Existing alternate removal continues to
use the live remove path.


bryancall (approved):

ezelkow1:
Can you also make a 9.2.x pr since i guess this isnt going in sec

cmcfarlen:
Cherry-picked to 10.2.x

cmcfarlen:

PR #13139, opened 2026-05-06
Limit regex remap substitutions

bneradt:
Regex remap targets could repeat valid substitution markers enough times
to exceed the fixed substitution arrays, even when every marker referred
to an allowed capture group.

This rejects targets with more substitution markers than the parser can
store and covers the boundary with remap parser unit tests.


ezelkow1:
[approve ci]

serrislew (approved):

PR #13143, opened 2026-05-07
Fix combo paths with embedded colons

bneradt:
Combo handler query parameters use a colon to select the bucket prefix,
but asset paths can also contain colons. Requests with an embedded path
colon were split at the last colon, producing the wrong internal fetch
URL and failing instead of fetching the requested object.

This keeps the first colon as the combo delimiter and leaves later
colons in the asset path. This also adds an AuTest case with a neutral
asset name so the parser regression is covered.

### DIFF
--- a/plugins/esi/combo_handler.cc
+++ b/plugins/esi/combo_handler.cc
@@ -712,7 +712,7 @@ parseQueryParameters(const char *query, int query_len, ClientRequest &creq)
         }
       }
       param_start_pos = i + 1;
-    } else if (query[i] == ':') {
+    } else if ((query[i] == ':') && (colon_pos < param_start_pos)) {
       colon_pos = i;
     }
   }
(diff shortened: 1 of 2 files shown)


plugins/esi/combo_handler.cc
```diff
@@ -712,7 +712,7 @@ parseQueryParameters(const char *query, int query_len, ClientRequest &creq)
         }
       }
       param_start_pos = i + 1;
-    } else if (query[i] == ':') {
+    } else if ((query[i] == ':') && (colon_pos < param_start_pos)) {
       colon_pos = i;
```
bneradt, line 716: Basically: make sure colon_pos hasn't been set yet before setting it in case there is a `:` in the path too.

shukitchan (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #13144, opened 2026-05-08
Fix XPACK relative index underflow

bneradt:
XPACK dynamic-table lookups convert absolute match indexes into HPACK relative indexes. A missing guard allowed the unsigned subtraction to wrap if an impossible future absolute index reached that conversion path.

This guards the conversion by returning unmatched results before reading the table head and by rejecting absolute indexes beyond the current head. This also adds XPACK unit coverage for empty, missing, and normal relative lookup results.


maskit (approved):

cmcfarlen:
Cherry-picked to 10.2.x

cmcfarlen:

PR #13146, opened 2026-05-08
[TS API] Add plugin API for config reload framework

### LINKED ISSUE #12967: ATS Reload migration task. Expose ConfigRegistry to plugins
Add a plugin API so plugins can register their own config handlers and participate in the reload framework.

https://github.com/apache/trafficserver/pull/12892

brbzull0:
Adds a small `TSCfg*` plugin-facing API that lets plugins register
configuration files with the same reload framework core configs
already use (`ConfigRegistry` / `ConfigContext`). Plugin reloads then
surface in `traffic_ctl config status` with full state tracking,
can receive RPC payloads, and follow the deferred-completion
contract - none of which were available via `TSMgmtUpdateRegister`.

## Motivation

Today, plugins react to `traffic_ctl config reload` only via
`TSMgmtUpdateRegister`. That gives them a notification but nothing
else: no per-key targeting, no payload, no `_reload` directives,
no companion files, no way to surface reload outcome in `config
status`. Every plugin that wants config-reload behaviour ends up
re-implementing pieces of the framework - `regex_revalidate`, for
example, ships its own file-mtime watcher.

This PR exposes the framework directly so plugins integrate with
it instead of re-implementing it.

### Types & enums (in [`include/ts/apidefs.h.in`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/apidefs.h.in))

- [`TSCfgLoadCtx`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/apidefs.h.in#L1595) — opaque per-reload handle.
- [`TSCfgLoadCb`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/apidefs.h.in#L1597-L1613) — plugin reload callback signature.
- [`TSCfgSourceType`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/apidefs.h.in#L1615-L1619) — `FILE_ONLY` / `FILE_AND_RPC`.
- [`TSCfgLogLevel`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/apidefs.h.in#L1621-L1632) — `NOTE` / `WARNING` / `ERROR`.
- [`TSCfgRegistrationInfo`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/apidefs.h.in#L1652-L1689) — option struct for `TSCfgRegister`.
- [`TSCfgFileDependencyInfo`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/apidefs.h.in#L1712-L1735) — option struct for `TSCfgAddFileDependency`.
- [`TSYaml`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/apidefs.h.in#L1739) — opaque alias for `YAML::Node*` (pre-existing JSONRPC type, reused).

### Registration (called from `TSPluginInit`)

| Function | Declaration | Implementation |
|---|---|---|
| `TSCfgRegister` | [`ts.h:1285`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/ts.h#L1285) | [`InkAPI.cc:3335`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/src/api/InkAPI.cc#L3335) |
| `TSCfgAttachReloadTrigger` | [`ts.h:1330`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/ts.h#L1330) | [`InkAPI.cc:3409`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/src/api/InkAPI.cc#L3409) |
| `TSCfgAddFileDependency` | [`ts.h:1347`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/ts.h#L1347) | [`InkAPI.cc:3435`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/src/api/InkAPI.cc#L3435) |

### Per-reload context (used inside the plugin's `TSCfgLoadCb`)

| Function | Declaration | Implementation |
|---|---|---|
| `TSCfgLoadCtxInProgress` | [`ts.h:1371`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/ts.h#L1371) | [`InkAPI.cc:3519`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/src/api/InkAPI.cc#L3519) |
| `TSCfgLoadCtxComplete` | [`ts.h:1384`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/ts.h#L1384) | [`InkAPI.cc:3535`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/src/api/InkAPI.cc#L3535) |
| `TSCfgLoadCtxFail` | [`ts.h:1397`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/ts.h#L1397) | [`InkAPI.cc:3541`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/src/api/InkAPI.cc#L3541) |
| `TSCfgLoadCtxAddLog` | [`ts.h:1412`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/ts.h#L1412) | [`InkAPI.cc:3547`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/src/api/InkAPI.cc#L3547) |
| `TSCfgLoadCtxAddSubtask` | [`ts.h:1427`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/ts.h#L1427) | [`InkAPI.cc:3572`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/src/api/InkAPI.cc#L3572) |
| `TSCfgLoadCtxGetFilename` | [`ts.h:1460`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/ts.h#L1460) | [`InkAPI.cc:3596`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/src/api/InkAPI.cc#L3596) |
| `TSCfgLoadCtxGetReloadToken` | [`ts.h:1476`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/ts.h#L1476) | [`InkAPI.cc:3608`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/src/api/InkAPI.cc#L3608) |
| `TSCfgLoadCtxGetSuppliedYaml` | [`ts.h:1490`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/ts.h#L1490) | [`InkAPI.cc:3620`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/src/api/InkAPI.cc#L3620) |
| `TSCfgLoadCtxGetReloadDirectives` | [`ts.h:1503`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/include/ts/ts.h#L1503) | [`InkAPI.cc:3636`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/src/api/InkAPI.cc#L3636) |

## What this gives plugins, on top of `TSMgmtUpdateRegister`

- **Per-key targeting** - handler runs only when this plugin's
  registered file or trigger record actually changes.
- **RPC payload** - configs registered with `TS_CFG_SOURCE_FILE_AND_RPC`
  receive YAML content via JSONRPC and react to `_reload` directives.
- **Status surface** - success / fail / in-progress / timeout plus
  log entries appear in `traffic_ctl config status` for `config
  reload`, file changes, and record changes during a reload cycle.
- **Subtasks** - handlers can split work into named subtasks that
  aggregate into the parent's status.
- **Companion files** - `TSCfgAddFileDependency` declares an extra
  file whose changes invoke the same handler, optionally routing
  inline RPC content via `dep_key`.
- **Deferred completion** - handlers may stash the context, return,
  and finish on another thread later. Same contract core handlers
  already have.

## Limitations

- Not supported in remap plugins (`TSRemapInit` /
  `TSRemapNewInstance`); the reload framework is centred on global
  plugins.
- `is_required` is propagated to `FileManager` but not enforced at
  reload time today (catalog/inspection only).
- `TSCfgAttachReloadTrigger` is not a free-form record-change
  subscription; it triggers a reload of the registered config and
  nothing else.

## Documentation

- New reference page: [`doc/developer-guide/api/functions/TSCfgRegister.en.rst`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/doc/developer-guide/api/functions/TSCfgRegister.en.rst)
  consolidates `TSCfgRegister`, `TSCfgAttachReloadTrigger`,
  `TSCfgAddFileDependency`, and the full `TSCfgLoadCtx*` family.
- New section in [`doc/developer-guide/config-reload-framework.en.rst`](https://github.com/brbzull0/trafficserver/blob/plugin-config-registry-api-v2/doc/developer-guide/config-reload-framework.en.rst)
  covering plugin reload, with example, lifecycle, and
  `traffic_ctl config status` output.

Fixes: #12967

### DIFF
--- a/doc/developer-guide/api/functions/TSCfgRegister.en.rst
+++ b/doc/developer-guide/api/functions/TSCfgRegister.en.rst
@@ -227,9 +227,13 @@ Registration
 
    Must be called from :func:`TSPluginInit`, **after**
    :func:`TSPluginRegister`. Returns ``TS_ERROR`` if called outside
-   ``TSPluginInit``, before ``TSPluginRegister``, with a null or
-   incomplete ``info`` struct, or if another plugin (or core) has
-   already registered the same key.
+   ``TSPluginInit``, before ``TSPluginRegister``, or with a null or
+   incomplete ``info`` struct.
+
+   If another plugin (or core) has already registered the same key,
+   the duplicate registration is rejected and a warning is logged
+   identifying both owners; ``TS_SUCCESS`` is still returned, since
+   the plugin has no useful recovery path at init time.
 
 :func:`TSCfgAttachReloadTrigger`
    Wires a record so that changing its value re-runs the reload handler
--- a/doc/developer-guide/api/functions/TSCfgRegister.en.rst
+++ b/doc/developer-guide/api/functions/TSCfgRegister.en.rst
@@ -53,10 +53,27 @@ Synopsis
 
    .. var:: std::string_view key
 
-      Unique registry key. Doubles as the YAML node name when an operator
-      reloads via ``traffic_ctl config reload --content '{<key>: {...}}'``.
-      Convention: use the plugin name, or ``<plugin>.<sub>`` when a single
-      plugin owns multiple configs. Required.
+      Unique registry key, also used as the YAML node name when an
+      operator reloads via
+      ``traffic_ctl config reload --data '{<key>: {...}}'``.
+      Must be unique across all registered configs (core and plugin).
+
+      **Convention.** Use the plugin name (``my_plugin``). Keep the key
+      alphanumeric + underscore; avoid dots, since
+      ``traffic_ctl --directive my_plugin.dry_run=true`` parses the
+      first ``.`` as the directive separator.
+
+      **Plugins with multiple files:**
+
+      - If the files share the same logical config and should reload
+        together, register once and add the rest via
+        :func:`TSCfgAddFileDependency` (single task per reload).
+      - If the files are genuinely independent and reloadable on their
+        own, register each separately with a disambiguating key
+        (e.g. ``my_plugin_main``, ``my_plugin_aux``). Each becomes its
+        own task entry in :option:`traffic_ctl config status`.
+
+      Required.
 
    .. var:: std::string_view config_path
 
--- a/doc/developer-guide/api/functions/TSCfgRegister.en.rst
+++ b/doc/developer-guide/api/functions/TSCfgRegister.en.rst
@@ -31,16 +31,16 @@ Synopsis
 
     #include <ts/ts.h>
 
-.. type:: TSCfgSourceType
+.. enum:: TSCfgSourceType
 
    What content sources a plugin's config handler supports.
 
-   .. var:: TS_CFG_SOURCE_FILE_ONLY
+   .. enumerator:: TS_CFG_SOURCE_FILE_ONLY
 
       Handler reloads only from a file on disk. RPC-supplied YAML is
       rejected.
 
-   .. var:: TS_CFG_SOURCE_FILE_AND_RPC
+   .. enumerator:: TS_CFG_SOURCE_FILE_AND_RPC
 
       Handler can also process YAML content supplied via RPC.
 
@@ -95,7 +95,7 @@ Synopsis
    .. var:: TSCfgSourceType source
 
       Whether ``handler`` can also process YAML content supplied via RPC.
-      Defaults to :var:`TS_CFG_SOURCE_FILE_ONLY`.
+      Defaults to :enumerator:`TSCfgSourceType::TS_CFG_SOURCE_FILE_ONLY`.
 
    .. var:: bool is_required
 
@@ -355,7 +355,7 @@ Inputs:
    framework-reserved ``_reload`` directives are stripped before
    delivery. The returned node is undefined when the reload was driven
    by file change rather than RPC. Only meaningful when the entry was
-   registered with :var:`TS_CFG_SOURCE_FILE_AND_RPC`.
+   registered with :enumerator:`TSCfgSourceType::TS_CFG_SOURCE_FILE_AND_RPC`.
 
 :func:`TSCfgLoadCtxGetReloadDirectives`
    Returns the YAML map extracted from the ``_reload`` key in the
--- a/doc/developer-guide/api/functions/TSCfgRegister.en.rst
+++ b/doc/developer-guide/api/functions/TSCfgRegister.en.rst
@@ -99,8 +99,12 @@ Synopsis
 
    .. var:: bool is_required
 
-      ``true`` if the file must exist on disk for a reload to be
-      considered successful. Defaults to ``false``.
+      Hint propagated to FileManager: marks the file as "required" in
+      catalog/inspection output and emits a ``Dbg`` line if the file
+      is missing at registration time. The framework does **not**
+      enforce this flag at reload-time today: the handler is still
+      invoked even when the file is missing or invalid, and the
+      plugin chooses whether to fail the reload. Defaults to ``false``.
 
 .. type:: TSCfgFileDependencyInfo
 
@@ -131,8 +135,9 @@ Synopsis
 
    .. var:: bool is_required
 
-      ``true`` if the file must exist on disk for a reload to be
-      considered successful. Defaults to ``false``.
+      Hint propagated to FileManager (catalog/inspection only); see the
+      equivalent field on :type:`TSCfgRegistrationInfo`. Defaults to
+      ``false``.
 
 .. type:: TSCfgLoadCtx
 
@@ -147,14 +152,35 @@ Synopsis
    Plugin reload callback signature. ``data`` is the opaque pointer the
    plugin supplied via :var:`TSCfgRegistrationInfo::data`.
 
+.. enum:: TSCfgLogLevel
+
+   Severity for log entries emitted via :func:`TSCfgLoadCtxAddLog`.
+
+   .. enumerator:: TS_CFG_LOG_NOTE
+
+      Informational. Default level. Maps to ``DL_Note`` internally.
+
+   .. enumerator:: TS_CFG_LOG_WARNING
+
+      Concerning; the reload may still complete. Maps to ``DL_Warning``.
+
+   .. enumerator:: TS_CFG_LOG_ERROR
+
+      Failure cause; pair with a subsequent :func:`TSCfgLoadCtxFail`.
+      Maps to ``DL_Error``.
+
+   The framework deliberately does not expose the fatal-class diagnostics
+   levels here: a reload handler should not be able to terminate the
+   process. Plugins that truly need debug-only output should use
+   ``Dbg(ctl, ...)`` instead of this API.
+
 .. function:: TSReturnCode TSCfgRegister(const TSCfgRegistrationInfo *info)
-.. function:: TSReturnCode TSCfgAttachTrigger(std::string_view key, std::string_view record_name)
+.. function:: TSReturnCode TSCfgAttachReloadTrigger(std::string_view key, std::string_view record_name)
 .. function:: TSReturnCode TSCfgAddFileDependency(const TSCfgFileDependencyInfo *info)
-.. function:: TSReturnCode TSCfgSetEnabled(std::string_view key, int enabled)
 .. function:: void TSCfgLoadCtxInProgress(TSCfgLoadCtx ctx, std::string_view msg)
 .. function:: void TSCfgLoadCtxComplete(TSCfgLoadCtx ctx, std::string_view msg)
 .. function:: void TSCfgLoadCtxFail(TSCfgLoadCtx ctx, std::string_view msg)
-.. function:: void TSCfgLoadCtxAddLog(TSCfgLoadCtx ctx, DiagsLevel level, std::string_view msg)
+.. function:: void TSCfgLoadCtxAddLog(TSCfgLoadCtx ctx, TSCfgLogLevel level, std::string_view msg)
 .. function:: TSCfgLoadCtx TSCfgLoadCtxAddSubtask(TSCfgLoadCtx ctx, std::string_view description)
 .. function:: std::string_view TSCfgLoadCtxGetFilename(TSCfgLoadCtx ctx)
 .. function:: std::string_view TSCfgLoadCtxGetReloadToken(TSCfgLoadCtx ctx)
@@ -188,13 +214,20 @@ Registration
    incomplete ``info`` struct, or if another plugin (or core) has
    already registered the same key.
 
-:func:`TSCfgAttachTrigger`
+:func:`TSCfgAttachReloadTrigger`
    Attaches a fully-qualified record name (e.g.
    ``proxy.config.my_plugin.enabled``) to a previously registered key.
-   When the record value changes, the plugin's handler is invoked, the
-   same way it would be if the file changed on disk. May be called
-   multiple times to attach more than one trigger record to the same
-   entry.
+   When the record value changes, the plugin's reload handler is
+   invoked, the same way it would be if the file changed on disk. May
+   be called multiple times to attach more than one trigger record to
+   the same entry.
+
+   This is **not** a general record-change subscription primitive: the
+   only callback the plugin gets is the existing config-reload
+   handler, the reload is treated as file-driven (no RPC payload), and
+   standalone record changes (i.e. ``traffic_ctl config set`` outside
+   an active reload cycle) invoke the handler with an empty context
+   that does not surface in :option:`traffic_ctl config status`.
 
 :func:`TSCfgAddFileDependency`
    Adds a companion file to a previously registered key. When the
@@ -202,12 +235,6 @@ Registration
    YAML under :var:`TSCfgFileDependencyInfo::dep_key` via JSONRPC, when
    set - the plugin's handler is invoked. With ``dep_key`` empty the
    dependency is file-change-only.
-
-:func:`TSCfgSetEnabled`
-   Toggles a registered config at runtime. Disabled entries are skipped
-   during reloads (their tasks are short-circuited to a
-   ``"skipped (disabled)"`` completion). Unlike the other three, this
-   function may be called at any time, not only during ``TSPluginInit``.
 
 Per-reload context (TSCfgLoadCtx)
 ---------------------------------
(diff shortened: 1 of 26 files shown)


cmcfarlen (changes requested):
What happens when a plugin is unloaded or reinitialized? I couldn't find where the config context is unregistered when a plugin is unloaded.

brbzull0:
> What happens when a plugin is unloaded or reinitialized? I couldn't find where the config context is unregistered when a plugin is unloaded.

This is only for global plugins, IIRC we do not unload/reinitialize global plugins. (unless we have a new feature that I am not aware of).

brbzull0:
@cmcfarlen  ping. 😄

cmcfarlen (commented):
Nice work — the design is sound and the docs are unusually thorough. Two things I think should be sorted before merge, plus a handful of smaller items.

## Blocking

### 1. Handler-less registry entries can now abort the server

`do_register()` now passes the **registry key** as `configName` to `FileManager::addFile()` whenever `filename_record` is empty, and `process_config_update()` routes an unknown `configName` that exists in the registry straight to `schedule_reload()`.

Two core entries are registered with an empty `filename_record` **and no handler**:

```cpp
// src/traffic_server/traffic_server.cc:795,797
reg.register_static_file("storage", ts::filename::STORAGE, {}, true);
reg.register_static_file("plugin",  ts::filename::PLUGIN);
```

Repro: edit `storage.config` (or `plugin.config`), then `traffic_ctl config reload` → `rereadConfig()` sees the mtime change → `fileChanged("storage.config", "storage")` → `RecGetRecordType("storage")` fails → new branch → `contains("storage")` is true → `schedule_reload("storage")` → `execute_reload()` → `ink_release_assert(entry_copy.handler)` → traffic_server aborts.

Today `configName` is `""` for these two files, so `process_config_update()` is a no-op — this PR introduces the path.

Suggest fixing both ends:

- In the new `FileManager` branch, resolve the entry and require a handler before scheduling (`find(configName)` + `entry->handler`), otherwise keep the existing `Dbg`.
- Downgrade `ink_release_assert(entry_copy.handler)` in `execute_reload()` to a `Warning` + `return`. A release assert reachable by editing a config file is a liability independent of this specific case.

### 2. The `TSCfgLoadCtx` liveness guard is itself a use-after-free

`finalize_plugin_ctx()` ends with `delete pctx`. Every subsequent call on that handle — `plugin_ctx_alive()`, or a second `Complete`/`Fail` reaching `pctx->consumed.exchange(true)` and `Warning(..., pctx->reload_token.c_str())` — dereferences freed memory. So the `consumed` flag doesn't actually protect anything: the diagnostic meant to catch plugin misuse is a heap-use-after-free read, and under the ASAN autest builds it will be reported as such rather than printing the warning.

Two ways out:

- **Make it real:** don't `delete` on finalize. Keep the handle in a process-lifetime side table (or intrusive list under a lock) keyed by an opaque id/generation, mark it consumed, and reclaim at reload-cycle end. Post-finalize calls then become genuine no-ops.
- **Make it honest:** delete as today, but remove `plugin_ctx_alive()`'s dereference-based check and the `ink_assert`, and document that use after `Complete`/`Fail` is undefined behaviour — which is what the code actually implements.

As written, the header comment ("After the completing call, the handle is invalid") and the code's apparent tolerance of post-finalize calls contradict each other, and plugin authors will trust the wrong one.

## Significant

**`TSCfgRegister` reports success when the registration was dropped.** `do_register()` only `Warning`s on a duplicate key, `register_plugin_config()` returns `void`, and `TSCfgRegister()` returns `TS_SUCCESS` for "registration attempt". A plugin whose key collides is never called and has no way to find out. Suggest having `do_register`/`register_plugin_config` return `bool` and propagating it. Also: `TSCfgIsRegistered` is declared and implemented but appears in neither the new `.rst` nor the PR summary table.

**No versioning on the option structs.** `TSCfgRegister.en.rst` says "new fields may be appended in future |TS| versions without breaking source compatibility". True for *source*, not for *binary*: a plugin built against today's header passes a smaller `TSCfgRegistrationInfo`, and a newer core reads past its end. Either add a leading `size_t struct_size` (or version tag) and check it, or scope the doc claim to recompiled plugins.

**The wrapper lambda rediscovers what it already knows.** In `TSCfgRegister`:

```cpp
auto const &key = ctx.get_description();
if (auto const *entry = config::ConfigRegistry::Get_Instance().find(key); entry != nullptr) {
  handle->filename = entry->resolve_filename();
}
```

`key_str` is in scope at registration time — capturing it avoids depending on `description == key`, which is an incidental invariant of the current `create_config_context()` call sites, and drops a shared-lock registry lookup per reload.

**`TSCfgLoadCtxGetFilename` semantics are ambiguous.** `resolve_config_filename()` deliberately returns the bare filename (no `sysconfdir` prefix). A plugin registering `"myplugin.yaml"` gets `"myplugin.yaml"` back and has to prepend `TSConfigDirGet()`, while one registering an absolute path gets a directly openable path. Both new test plugins register absolute paths, so the relative case is untested. Either resolve to an absolute path in the plugin layer, or state it explicitly — the doc currently reads "the resolved file path the framework expects this handler to read".

Related: for a companion file added via `TSCfgAddFileDependency`, the handler gets the *parent's* filename and no indication of which file changed. Worth documenting, and arguably worth exposing.

## Smaller items

- **Global `namespace detail`** in `include/mgmt/config/ConfigContext.h` plus `friend class detail::RecordTriggeredReloadContinuation` is namespace pollution in a shared header. `config::detail` would be better — or thread `plugin_name` through `ReloadCoordinator::create_config_context()` and drop the friendship and forward declaration entirely.
- **`attach()` lock split:** check under `shared_lock`, wire, then re-find under `unique_lock`. The `if (it != _entries.end())` on the second pass silently swallows the case where the entry vanished. Init-time-only today, so benign, but a `Dbg`/`Warning` beats dropping the trigger record silently.
- In `RecordTriggeredReloadContinuation`, `ctx.set_plugin_name()` in the `!ctx` standalone branch is a guaranteed no-op (`_task.lock()` is null). Harmless, but it reads as though attribution works on that path.
- **`regex_revalidate`:** the `TSCfgRegister()` return value is discarded while every other call in that file logs; `config_cont` can now be declared inside the `if (!disable_timed_reload)` block since that's its only use. Also worth a look: a registry-triggered reload right after a timed one legitimately reports "unchanged" (via `last_load`), which may read as a failed reload in `config status`.
- The reworded `execute_reload` non-terminal warning is terser but drops the actionable guidance the old text had about calling `ctx.complete()`/`ctx.fail()` from the deferred path. I'd keep that.
- **Diff noise:** the en-dash→hyphen rewrites across 8 files and the `Entry::has_handler()` removal are cosmetic churn inside a 3k-line feature PR. Splitting them into a separate commit would make re-review considerably easier.
- Docs could state that the `std::string_view` fields are copied during the `TSCfgRegister` call, so temporaries are safe — otherwise authors will assume they must keep backing storage alive.

## Test coverage

Good breadth on the happy paths: file reload, RPC YAML, `_reload` directives, subtask success/failure, and two-stage deferred completion, all asserted through `traffic_ctl config status`. Gaps that line up with the findings above:

- Nothing exercises the new `FileManager` → `schedule_reload` routing, which is where #1 lives — a test that touches `storage.config` and reloads would fail today.
- No test for handle misuse (double `Complete`, use-after-`Fail`); a small Catch2 test around `finalize_plugin_ctx` would have surfaced #2 without a running server.
- No coverage for duplicate-key registration, empty `plugin_name` refusal, or a relative `config_path`.

## Security / performance

No concerns. The API is gated to `TSPluginInit` after a successful `TSPluginRegister` via `validate_plugin_init()`, message text is correctly passed as a `%.*s` argument rather than a format string, and the RPC content path is unchanged. Reload is a cold path, so the per-reload allocation and registry lookup don't matter.

brbzull0:
[approve ci autest 2]

brbzull0:
[approve ci freebsd]

brbzull0:
Hi @cmcfarlen, thanks for the review, I  made some changes base on your feedback.  thanks.

cmcfarlen (commented):
Re-reviewed the three commits since my last pass (`ede51fca7c`, `d427d269e5`, `c04093a524`). Both blockers are genuinely fixed, and fixed the right way rather than papered over. What's left is small.

## Verified fixed

**Handler-less entries can no longer abort the server.** `is_reloadable()` does the existence-and-handler check atomically under the registry lock, `process_config_update()` routes on that instead of `contains()`, and `execute_reload()`'s `ink_release_assert(entry_copy.handler)` is now a `Warning` + `return`. Both ends, as suggested. The `storage.config` / `plugin.config` repro no longer reaches `schedule_reload()`.

**The `TSCfgLoadCtx` handle is now actually safe.** Opaque monotonic ids instead of raw pointers (so no ABA on a stale handle), `shared_ptr` ownership with `lookup()` returning a reference the caller holds for the duration of its call, and `extract()` removing under the same lock so a second finalize can't double-free. This resolves Copilot's cross-thread finding as well — the deferred-completion contract is now honored rather than nominally supported.

Also confirmed fixed: `TSCfgRegister` returns `TS_ERROR` on a dropped registration, with `do_register`/`register_plugin_config` returning `bool`; the wrapper lambda captures `key_str` rather than rediscovering it via `get_description()`; `apply_passed_config()` uses a const `cfg_view` so `_reload` is no longer inserted into the payload; `add_file_dependency()` wires the record callback before touching `FileManager`, with a comment explaining why; and `regex_revalidate` gained a real `reload_mutex` taken on both the timed and framework paths, created and destroyed with the plugin state.

New unit tests for duplicate-key rejection and `is_reloadable` cover the logic behind both blockers.

## Correction to my previous review

I claimed `TSCfgLoadCtxGetFilename` hands back a bare filename for a relative `config_path`, forcing plugins to prepend `TSConfigDirGet()`. That was wrong — `Entry::resolve_filename()` prepends `sysconfdir` for any non-absolute path, and has done so on master all along. I was looking at the free `resolve_config_filename()`. Plugins get an openable path either way; disregard that item.

It does leave a small doc inaccuracy, though. `TSCfgRegister.en.rst:363` says step 2 "returns ``config_path`` as-registered", which isn't true for a relative path — it comes back `sysconfdir`-prefixed. Worth correcting since it's the difference between a plugin calling `open()` directly or not.

## Remaining items

### 1. `ink_assert` makes the documented no-ops fatal in the builds CI uses

`ts.h` now says of `TSCfgLoadCtxComplete`: *"A later call on the same handle (double finalize, or any accessor) is ignored and logged as misuse, not acted on. Null/unknown ``ctx`` is a no-op."* The registry comment makes the same promise — *"a genuine no-op instead of a use-after-free read."*

Both paths then call `ink_assert(!"...")`, and `_ink_assert` is `TS_NORETURN`. `ink_assert` is compiled in under `DEBUG` (`ink_assert.h:42`), which is exactly what the autest and ASAN presets build. So:

```c
TSCfgLoadCtxComplete(nullptr, {});   // -> id 0, never valid (_next_id starts at 1)
                                     // -> extract() misses -> Warning + ink_assert -> abort
```

A documented no-op aborts `traffic_server` in a debug build. Same for any post-finalize accessor. The safety property the new registry buys you is only observable in release builds.

Pick one contract and make both the code and the docs say it:

- **Supported no-op** (matches what's written): drop the two `ink_assert`s, keep the `Warning`. Misuse is diagnosable without being fatal, and the property becomes testable.
- **Undefined behavior with a diagnostic**: keep the asserts, and drop "is a no-op" / "not acted on" from `ts.h` and the `.rst`.

I'd take the first — it's what the registry rewrite was for, and it's a two-line change. Null-handle tolerance in particular is worth keeping, since it lets a plugin's error path call `Complete` unconditionally.

### 2. Option-struct ABI is still unversioned

`TSCfgRegistrationInfo` and `TSCfgFileDependencyInfo` have no size or version field, and the docs claim (`:47-52`) that *"new fields may be appended in future |TS| versions without breaking source compatibility"*, with a nearby note that the struct "stays ABI-stable".

Source compatibility is right. Binary compatibility isn't: a plugin compiled against today's header allocates today's struct and passes a pointer; a newer core that appended a field reads past the end of that allocation. Given how carefully the rest of the API surface is kept ABI-stable, this seems worth closing now rather than at the first append:

```c
struct TSCfgRegistrationInfo {
  size_t struct_size{sizeof(TSCfgRegistrationInfo)};  // set by the plugin's compiler
  ...
};
```

with `TSCfgRegister` refusing anything smaller than the fields it reads. Alternatively, scope the doc claim to recompiled plugins and say appending is an ABI break — that's a valid choice, just not the one currently documented.

### 3. Minor leftovers

- **Global `namespace detail`** in `include/mgmt/config/ConfigContext.h:46`, sitting right beside a properly nested `namespace config`, plus `friend class detail::RecordTriggeredReloadContinuation`. `config::detail` costs nothing and stops a very generic name from leaking out of a shared header.
- **`ctx.set_plugin_name(entry->plugin_name)` at `ConfigRegistry.cc:125`** is still a guaranteed no-op on the standalone branch — `_task.lock()` is null there. Harmless, but it reads as though attribution works on that path.
- **`regex_revalidate`**: `TSCfgRegister(&cfg_info);` still discards the return with a comment that errors are logged internally. That's now true, but the plugin proceeds believing it's registered when it isn't and silently loses framework reloads — worth at least a `Dbg`. `config_cont` can also move inside the `if (!disable_timed_reload)` block, its only use.
- **The en-dash → hyphen rewrites are still in the diff** across several files. Cosmetic churn inside a 3k-line feature PR makes re-review harder than it needs to be; a follow-up commit would be cleaner, though not worth another round-trip at this point.
- The `_live` entry for a handler that never finalizes persists for process lifetime. The comment is honest about it being one entry per un-finalized reload, and it's no worse than the old leak, but reaping at reload-cycle end would bound it for a buggy plugin.

## Test coverage

The unit tests added since my last pass cover the two blockers' logic well. Two gaps remain:

- **No misuse-path coverage** — double `Complete`, accessor after `Fail`, null handle. This is the code that got rewritten, and its whole point is defined behavior under misuse. Note these tests can't be written while item 1 stands, since a debug build would abort; fixing item 1 unblocks them.
- Nothing exercises a file change mapping to a handler-less registry key end to end. The `is_reloadable` unit test covers the decision, which is the important half, so this is optional.

## One coordination note

This adds five autests as `tests/gold_tests/jsonrpc/*.test.py` plus three test plugins. #13545 converts the entire tree to Uranium/pytest manifests under `tests/uranium_tests/`. Whichever lands second will need its tests ported, and the plugin `CMakeLists.txt` under `tests/gold_tests/jsonrpc/plugins/` moves with it. Worth agreeing on an order now rather than discovering it in a conflict — the five tests here are mostly `traffic_ctl config status` output assertions, which map onto replay manifests cleanly enough, but not for free.

Nice work on the rework — the handle registry in particular is a real improvement over what was there.

cmcfarlen (approved):

PR #13151, opened 2026-05-09
Fix AIO callback from_api completion lifetime

bneradt:
The AIOCallback io_complete handler used a member variable, from_api, to
determine whether to delete itself. The problem is that in the situation
where in the course of processing the function `this` was already
deleted, the use of the from_api variable was, by definition, a use
after free. If built under ASan, this resulted in a use-after-free
assertion.

Concretely, we were seeing this in docs during cache stripe
initialization after an unclean shutdown.  If the on-disk cache
directory was dirty, startup recovery scanned the data area, cleared
directory entries for the uncertain range, and wrote the repaired
directory back out.  The temporary AIO callbacks for that recovery wrote
live in StripeInitInfo.  When the recovery write completion was
delivered, StripeSM::handle_recover_write_dir() deleted StripeInitInfo,
which destroyed the AIOCallback object whose AIOCallback::io_complete()
frame was still returning. At that point, the use of from_api was a use
after free.

This snapshots the API-owned callback flag before dispatching the
completion and uses that local value for the post-callback cleanup.
This also adds a focused regression test for completion handlers that
release the callback owner before AIOCallback::io_complete() returns.

Introduced in #13027

### DIFF
--- a/include/iocore/aio/AIO.h
+++ b/include/iocore/aio/AIO.h
@@ -79,10 +79,10 @@ struct AIOCallback : public Continuation {
   EThread     *thread = AIO_CALLBACK_THREAD_ANY;
   AIOCallback *then   = nullptr;
   // set on return from aio_read/aio_write
-  int64_t    aio_result = 0;
-  AIO_Reqs  *aio_req    = nullptr;
-  ink_hrtime sleep_time = 0;
-  bool       from_api   = false;
+  int64_t    aio_result  = 0;
+  AIO_Reqs  *aio_req     = nullptr;
+  ink_hrtime sleep_time  = 0;
+  bool       from_ts_api = false;
   SLINK(AIOCallback, alink); /* for AIO_Reqs::aio_temp_list */
 #if TS_USE_LINUX_IO_URING
   iovec        iov     = {}; // this is to support older kernels that only support readv/writev
--- a/src/iocore/aio/unit_tests/test_AIOCallback.cc
+++ b/src/iocore/aio/unit_tests/test_AIOCallback.cc
@@ -1,0 +1,115 @@
+/** @file
+
+  Catch based unit tests for AIOCallback.
+
+  @section license License
+
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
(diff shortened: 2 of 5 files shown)


bryancall (dismissed):

include/iocore/aio/AIO.h
```diff
@@ -82,8 +82,8 @@ struct AIOCallback : public Continuation {
   int64_t    aio_result = 0;
   AIO_Reqs  *aio_req    = nullptr;
   ink_hrtime sleep_time = 0;
-  bool       from_api   = false;
-  SLINK(AIOCallback, alink); /* for AIO_Reqs::aio_temp_list */
+  bool       from_api   = false; ///< Whether the TS API created this callback.
```
zwoop: Extremely nitpick, but this doxygen comment feels misplaced. Use a regular comment, but honestly, I think the variable name itself is self documenting.

src/iocore/aio/unit_tests/test_AIOCallback.cc
```diff
@@ -0,0 +1,67 @@
...
+
+} // namespace
+
+TEST_CASE("AIOCallback completion tolerates owner deletion", "[iocore][aio]")
+{
+  bool completed = false;
+  auto owner     = new AIOCompletionOwner(completed);
+  auto callback  = &owner->callback;
+
+  CHECK(callback->io_complete(EVENT_NONE, nullptr) == EVENT_DONE);
```
zwoop, line 112: Somewhat interesting thought from Claude, a comment may be useful? Again, nitpick.

```
The test only catches the regression under ASan/MSan. The owner is destroyed inside 
handle_aio_complete, so on broken code the post-dispatch if (from_api) reads freed memory — but since 
from_api == false by default, the branch happens to be skipped and the test passes. Without sanitizers, 
this test would not have caught the original bug. The CI runs ASan, so this is acceptable, but consider 
noting it in a comment so someone running the test locally without ASan understands why a passing result 
isn't conclusive. Even better: also assert the from_api == true path (e.g., a second test where the owner 
does not delete itself and the callback is dynamically allocated with from_api = true, verifying the 
snapshot still triggers the self-delete).
```

include/iocore/aio/AIO.h
bneradt: I guess I'm embarrassing myself here, but I added that comment explicitly because it wasn't clear to me when I was first studying this use-after-free bug that `from_api` meant "from the external TS API".  It wasn't obvious what API was being referred to. I suppose I can rename to `from_ts_api`.

src/iocore/aio/unit_tests/test_AIOCallback.cc
bneradt: Good points...

zwoop (approved):

PR #13155, opened 2026-05-11
10.1.x: Fix cache retry assert on ServerAddrSet (#12972)

bneradt:
A TSHttpTxnServerAddrSet retry can re-enter the cache miss path after
ATS already holds a cache write lock for the same request. The
redirect-only prepared-write reuse logic then asserts because this is
not actually a redirect.

Preserve the existing cache write lock and retry context when the
retry returns through HandleCacheOpenReadMiss. Extend the retry autest
to cover the cached retry path that fails without this fix.

Fixes: #12971
(cherry picked from commit bf909edf88a4c35cef5f4ff86c7590a867f2ab7e)


cmcfarlen:
picked normally with the addition PR.  Thanks!

PR #12792, opened 2026-01-08
plugins(stats_over_http): Add Prometheus v2 output with metric labels

itsh3sam:
The current Prometheus output in stats_over_http dumps flattened metric names (e.g., `proxy.process.http.200_responses`), which makes it difficult to aggregate data in Grafana or Prometheus. This adds a new "v2" output format that parses these names into labels like `status="200"` or `method="GET"`.

The new format can be accessed via /_stats/prometheus_v2 or by passing an Accept header with version=2.0.0.

To keep this efficient, the parsing uses `swoc::TextView` for a zero- allocation approach. It recognizes common ATS patterns such as status codes, HTTP methods, cache results, and thread/volume indices. I've also added deduplication for the HELP and TYPE metadata lines, so that metrics sharing the same base name (but different labels) are grouped correctly in the output.

Includes compile-time unit tests (static_assert) to verify the parsing logic and updated AuTests for the new endpoint.

Related issue: #12778

PR #13156, opened 2026-05-12
Add Prometheus v2 labeled stats

bneradt:
Prometheus consumers currently only get flat ATS stat names, which makes
related counters hard to aggregate and can hide broken output behind
parser leniency. The draft v2 output also exposed that risk by letting
`completed` look like a request method and by interleaving samples from
the same family.

This adds a Prometheus v2 response format that groups samples by metric
family and derives labels for methods, directions, status codes, cache
results, time buckets, and cache volumes. This keeps lifecycle counters
such as completed requests as their own metrics while preserving the
existing v1 output.

This extends the stats_over_http AuTest and Prometheus ingester to
validate both the raw v2 exposition and the parser's view of it. This
catches split families, missing TYPE metadata, malformed labels, and
regressions in the expected labeled samples.

---

Taking over from https://github.com/apache/trafficserver/pull/12792

### DIFF
--- a/doc/admin-guide/plugins/stats_over_http.en.rst
+++ b/doc/admin-guide/plugins/stats_over_http.en.rst
@@ -105,20 +105,28 @@ if you wish to have it in CSV format you can do so by passing an ``Accept`` head
 
 .. option:: Accept: text/csv
 
-Prometheus formatted output is also supported via the ``Accept`` header:
+Prometheus formatted output is also supported via the ``Accept`` header. Version 0.0.4
+(flat metric names) and version 2.0.0 (labeled metrics for better aggregation)
+are supported:
 
 .. option:: Accept: text/plain; version=0.0.4
+.. option:: Accept: text/plain; version=2.0.0
 
 Alternatively, the output format can be specified as a suffix to the configured
 path in the HTTP request target.  The supported suffixes are ``/json``,
-``/csv``, and ``/prometheus``.  For example, if the path is set to ``/_stats``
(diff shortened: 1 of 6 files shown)


bryancall (approved):
This looks good and the coverage is better than most feature PRs I review. The v2 parsing, label derivation, and the family-splitting regression are all well exercised, and CI is green. Approving.

A few non-blocking items for a follow-up or a quick touch-up here:

- In `prometheus_v2_out_stat`, when `format_prometheus_v2_sample` returns false the sample is dropped with a bare return and no `Dbg`, unlike the v1 path which logs its skips. Add a log naming the metric and reason so a missing series is debuggable.
- The new `TSIOBufferWrite == TS_ERROR` branch in `stats_add_data_to_resp_buffer` returns 0 silently. Even though this effectively never fires on a growable IOBuffer, log it so a real write failure is not invisible.
- The new compressed v2 output (gzip/deflate/br) has no test. One autest run with `Accept-Encoding: gzip` against `/_stats/prometheus_v2` that decompresses and re-validates would lock down the header strings and the compression branch.
- The label escape/unescape round trip (`escape_prometheus_v2_label_value` / `decode_label_value`) is dead code under the current fixtures since no metric name yields a quote, backslash, or newline. A focused case feeding one of those through both sides would close that gap.

None of these block merge.

cmcfarlen:
Cherry-picked to the 10.2.x branch as b2bc2c22719f8e38608d1c489ce254469e7525d5 for the 10.2.0 release.

PR #13157, opened 2026-05-12
Fix build with newer Apple clang

maskit:
Four issues from newer Apple clang (version 21):

1. TextView.h: -Wdeprecated-literal-operator flags the whitespace between `""` and the suffix identifier in user-defined literal operator declarations. Remove the space in `operator"" _tv` and `operator"" _sv`.

2. configure.ac: the version check added in PR #12834 only applied -Wno-cast-function-type-mismatch when Apple clang was exactly version 17, so the flag was silently skipped on 21+. Extend the check to match version >= 17.

3. CryptoHash.h: memset/memcpy on non-trivially-copyable union type trip -Wnontrivial-memcall. Cast `this` to `void *`.

4. HostStatus.cc: same issue with bzero on HostStatRec.

### DIFF
--- a/include/tscore/CryptoHash.h
+++ b/include/tscore/CryptoHash.h
@@ -44,7 +44,7 @@ union CryptoHash {
   uint8_t u8[CRYPTO_HASH_SIZE / sizeof(uint8_t)];
 
   /// Default constructor - init to zero.
-  CryptoHash() { memset(this, 0, sizeof(*this)); }
+  CryptoHash() { memset(static_cast<void *>(this), 0, sizeof(*this)); }
   /// Copy constructor.
   CryptoHash(CryptoHash const &that) = default;
 
@@ -53,7 +53,7 @@ union CryptoHash {
   operator=(CryptoHash const &that)
   {
     if (this != &that) {
-      memcpy(this, &that, sizeof(*this));
+      memcpy(static_cast<void *>(this), &that, sizeof(*this));
     }
     return *this;
   }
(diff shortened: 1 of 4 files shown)


maskit:
[approve ci autest 2]

include/tscore/CryptoHash.h
```diff
@@ -44,7 +44,7 @@ union CryptoHash {
   uint8_t u8[CRYPTO_HASH_SIZE / sizeof(uint8_t)];
 
   /// Default constructor - init to zero.
-  CryptoHash() { memset(this, 0, sizeof(*this)); }
+  CryptoHash() { memset(static_cast<void *>(this), 0, sizeof(*this)); }
```
zwoop, line 47: Claude had an interesting thought on this, not for now, but for future. Maybe he's right, and we could simplify this "core" class.

```
  The warning symptom is that a user-defined default constructor and
  user-defined copy-assignment make the union non-trivially-copyable even
  though every member is a `uint*_t` array. A cleaner long-term fix
  (master only — not for 9.2.x) would be to initialize `u64{}` in-class
  and default `operator=`, making the union genuinely trivially copyable
  and retiring the memset/memcpy entirely. Filing a follow-up on master is
  the right venue for that — this PR is the correct minimum for the
  release branch.
```

zwoop (approved):

zwoop:
One thought / comment (not sure what we did for master), and likely for @bneradt  : 

Does the changes to TextView also need to go into the libswoc code repository ?

PR #13159, opened 2026-05-13
Avoid confusing AIO callback lifetime test

bneradt:
Coverity reports the AIO callback lifetime regression test as a leak
and bad free because the test heap-allocates an owner object whose
member callback is destroyed indirectly during completion. The runtime
behavior is intentional, but the ownership pattern makes the regression
test look invalid to static analysis.

This updates the fixture so the owner has normal stack lifetime and
owns the callback through a unique_ptr. The completion handler still
destroys the callback before AIOCallback::io_complete() returns,
preserving the lifetime coverage while making the allocation and free
visible to analysis.

### DIFF
--- a/src/iocore/aio/unit_tests/test_AIOCallback.cc
+++ b/src/iocore/aio/unit_tests/test_AIOCallback.cc
@@ -23,6 +23,8 @@
 
 #include <catch2/catch_test_macros.hpp>
 
+#include <memory>
+
 #include "iocore/aio/AIO.h"
 #include "iocore/eventsystem/Event.h"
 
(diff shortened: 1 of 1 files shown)


src/iocore/aio/unit_tests/test_AIOCallback.cc
```diff
@@ -84,11 +87,11 @@ struct DeletionTrackedAIOCallback : AIOCallback {
...
 {
-  bool completed = false;
-  auto owner     = new AIOCompletionOwner(completed);
-  auto callback  = &owner->callback;
+  bool               completed = false;
+  AIOCompletionOwner owner(completed);
+  auto              *callback = owner.callback.get();
 
   // Without ASan, a broken implementation can still pass because the stale value of from_ts_api is typically false.
   CHECK(callback->io_complete(EVENT_NONE, nullptr) == EVENT_DONE);
```
zwoop, line 97: Not important, but after completing this, I think callback is in an undefined state. We don't use it after this, but maybe we should NULL it out or something ? In case someone looks at this after we retire.

zwoop (approved):

PR #13161, opened 2026-05-14
Fix undercount of cp_list_len

masaori335:
When ATS enqueue a CacheVol to the list, the counter of the list have to be incremented.


bneradt (approved):
Prefix increment would be better, but postfix occurs elsewhere in this file already.

cmcfarlen:
Cherry-picked to 10.2.x

PR #7382, opened 2020-12-10
Switch to Standard atomic library for free/atomic lists in ink_queue.h.

ywkaras:

PR #13170, opened 2026-05-18
Convert ink_queue implementation to std::atomic

### LINKED ISSUE #7398: Need to switch from ink_atomic.h to Standard lib <atomic>
This is not simply a cosmetic issue.  ink_atomic.h effectively assumes there is no need for memory fence instructions on any architecture that TS runs on:

https://github.com/apache/trafficserver/blob/77714f253f3a3ab4ce39877da52e14b6270f6868/include/tscore/ink_atomic.h#L171

For x86 architectures for example, this is not true for any version after the 386.  It may also be true, for some architectures (or compiler options), that the alignment requirements when doing atomic access are stricter than those for non-atomic access.

### LINKED ISSUE #11640: TSan: ink atomic queue not so atomic
```
WARNING: ThreadSanitizer: data race (pid=9915)
  Atomic write of size 8 at 0xffff933d5bb0 by main thread (mutexes: write M0, write M1, write M2, write M3, write M4):
    #0 ink_atomiclist_push(InkAtomicList*, void*) /trafficserver/src/tscore/ink_queue.cc:559 (traffic_server+0x93d17c) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #1 ProtectedQueue::enqueue(Event*) /trafficserver/src/iocore/eventsystem/ProtectedQueue.cc:52 (traffic_server+0xecee84) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #2 EventProcessor::schedule(Event*, int) /trafficserver/src/iocore/dns/../eventsystem/P_UnixEventProcessor.h:125 (traffic_server+0xc2c5d0) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #3 EventProcessor::schedule_every(Continuation*, long, int, int, void*) /trafficserver/src/iocore/dns/../eventsystem/P_UnixEventProcessor.h:200 (traffic_server+0xc2ca90) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #4 RecProcessStart() /trafficserver/src/iocore/eventsystem/RecProcess.cc:189 (traffic_server+0xedc834) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #5 main /trafficserver/src/traffic_server/traffic_server.cc:2157 (traffic_server+0x8fcca4) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)

  Previous read of size 8 at 0xffff933d5bb0 by thread T4 (mutexes: write M5):
    #0 ProtectedQueue::wait(long) /trafficserver/src/iocore/eventsystem/ProtectedQueue.cc:94 (traffic_server+0xecf0b4) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #1 EThread::DefaultTailHandler::waitForActivity(long) /trafficserver/include/iocore/eventsystem/EThread.h:393 (traffic_server+0xed33c0) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #2 EThread::execute_regular() /trafficserver/src/iocore/eventsystem/UnixEThread.cc:300 (traffic_server+0xed2510) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #3 EThread::execute() /trafficserver/src/iocore/eventsystem/UnixEThread.cc:350 (traffic_server+0xed2780) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #4 spawn_thread_internal /trafficserver/src/iocore/eventsystem/Thread.cc:75 (traffic_server+0xed0220) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #5 <null> <null> (libtsan.so.2+0x555e4) (BuildId: 4fc6d4b1d08e85e4139b0e07fb335a013192e924)

  Location is heap block of size 1101696 at 0xffff932d5000 allocated by main thread:
    #0 operator new(unsigned long) <null> (libtsan.so.2+0xa7418) (BuildId: 4fc6d4b1d08e85e4139b0e07fb335a013192e924)
    #1 EventProcessor::spawn_event_threads(int, int, unsigned long) /trafficserver/src/iocore/eventsystem/UnixEventProcessor.cc:451 (traffic_server+0xed7b7c) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #2 TasksProcessor::start(int, unsigned long) /trafficserver/src/iocore/eventsystem/Tasks.cc:42 (traffic_server+0xecfa08) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #3 main /trafficserver/src/traffic_server/traffic_server.cc:2155 (traffic_server+0x8fcca0) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)

  Mutex M0 (0xffff97803f90) created at:
    #0 pthread_mutex_init <null> (libtsan.so.2+0x5f218) (BuildId: 4fc6d4b1d08e85e4139b0e07fb335a013192e924)
    #1 ink_mutex_init(pthread_mutex_t*) /trafficserver/src/tscore/ink_mutex.cc:54 (traffic_server+0x93aa0c) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #2 ProxyMutex::init() /trafficserver/include/iocore/eventsystem/Lock.h:245 (traffic_server+0x8e4df4) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #3 new_ProxyMutex() /trafficserver/include/iocore/eventsystem/Lock.h:677 (traffic_server+0x8e5490) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #4 Thread::Thread() /trafficserver/src/iocore/eventsystem/Thread.cc:44 (traffic_server+0xecfff4) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #5 EThread::EThread() /trafficserver/src/iocore/eventsystem/UnixEThread.cc:84 (traffic_server+0xed0958) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #6 main /trafficserver/src/traffic_server/traffic_server.cc:1936 (traffic_server+0x8fb688) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)

  Mutex M1 (0xffff97803d10) created at:
    #0 pthread_mutex_init <null> (libtsan.so.2+0x5f218) (BuildId: 4fc6d4b1d08e85e4139b0e07fb335a013192e924)
    #1 ink_mutex_init(pthread_mutex_t*) /trafficserver/src/tscore/ink_mutex.cc:54 (traffic_server+0x93aa0c) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #2 ProxyMutex::init() /trafficserver/include/iocore/eventsystem/Lock.h:245 (traffic_server+0x8e4df4) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #3 new_ProxyMutex() /trafficserver/include/iocore/eventsystem/Lock.h:677 (traffic_server+0x8e5490) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #4 Thread::Thread() /trafficserver/src/iocore/eventsystem/Thread.cc:44 (traffic_server+0xecfff4) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #5 EThread::EThread(ThreadType, int) /trafficserver/src/iocore/eventsystem/UnixEThread.cc:89 (traffic_server+0xed0c38) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #6 EventProcessor::spawn_event_threads(int, int, unsigned long) /trafficserver/src/iocore/eventsystem/UnixEventProcessor.cc:451 (traffic_server+0xed7bbc) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #7 EventProcessor::start(int, unsigned long) /trafficserver/src/iocore/eventsystem/UnixEventProcessor.cc:550 (traffic_server+0xed84f0) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #8 main /trafficserver/src/traffic_server/traffic_server.cc:2110 (traffic_server+0x8fc9ac) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)

  Mutex M2 (0xffff97803c90) created at:
    #0 pthread_mutex_init <null> (libtsan.so.2+0x5f218) (BuildId: 4fc6d4b1d08e85e4139b0e07fb335a013192e924)
    #1 ink_mutex_init(pthread_mutex_t*) /trafficserver/src/tscore/ink_mutex.cc:54 (traffic_server+0x93aa0c) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #2 ProxyMutex::init() /trafficserver/include/iocore/eventsystem/Lock.h:245 (traffic_server+0x8e4df4) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #3 new_ProxyMutex() /trafficserver/include/iocore/eventsystem/Lock.h:677 (traffic_server+0x8e5490) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #4 Thread::Thread() /trafficserver/src/iocore/eventsystem/Thread.cc:44 (traffic_server+0xecfff4) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #5 EThread::EThread(ThreadType, int) /trafficserver/src/iocore/eventsystem/UnixEThread.cc:89 (traffic_server+0xed0c38) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #6 EventProcessor::spawn_event_threads(int, int, unsigned long) /trafficserver/src/iocore/eventsystem/UnixEventProcessor.cc:451 (traffic_server+0xed7bbc) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #7 EventProcessor::start(int, unsigned long) /trafficserver/src/iocore/eventsystem/UnixEventProcessor.cc:550 (traffic_server+0xed84f0) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #8 main /trafficserver/src/traffic_server/traffic_server.cc:2110 (traffic_server+0x8fc9ac) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)

  Mutex M3 (0xffff97803990) created at:
    #0 pthread_mutex_init <null> (libtsan.so.2+0x5f218) (BuildId: 4fc6d4b1d08e85e4139b0e07fb335a013192e924)
    #1 ink_mutex_init(pthread_mutex_t*) /trafficserver/src/tscore/ink_mutex.cc:54 (traffic_server+0x93aa0c) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #2 ProxyMutex::init() /trafficserver/include/iocore/eventsystem/Lock.h:245 (traffic_server+0x8e4df4) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #3 new_ProxyMutex() /trafficserver/include/iocore/eventsystem/Lock.h:677 (traffic_server+0x8e5490) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #4 Thread::Thread() /trafficserver/src/iocore/eventsystem/Thread.cc:44 (traffic_server+0xecfff4) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #5 EThread::EThread(ThreadType, int) /trafficserver/src/iocore/eventsystem/UnixEThread.cc:89 (traffic_server+0xed0c38) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #6 EventProcessor::spawn_event_threads(int, int, unsigned long) /trafficserver/src/iocore/eventsystem/UnixEventProcessor.cc:451 (traffic_server+0xed7bbc) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #7 TasksProcessor::start(int, unsigned long) /trafficserver/src/iocore/eventsystem/Tasks.cc:42 (traffic_server+0xecfa08) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #8 main /trafficserver/src/traffic_server/traffic_server.cc:2155 (traffic_server+0x8fcca0) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)

  Mutex M4 (0xffff97803910) created at:
    #0 pthread_mutex_init <null> (libtsan.so.2+0x5f218) (BuildId: 4fc6d4b1d08e85e4139b0e07fb335a013192e924)
    #1 ink_mutex_init(pthread_mutex_t*) /trafficserver/src/tscore/ink_mutex.cc:54 (traffic_server+0x93aa0c) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #2 ProxyMutex::init() /trafficserver/include/iocore/eventsystem/Lock.h:245 (traffic_server+0x8e4df4) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #3 new_ProxyMutex() /trafficserver/include/iocore/eventsystem/Lock.h:677 (traffic_server+0x8e5490) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #4 Thread::Thread() /trafficserver/src/iocore/eventsystem/Thread.cc:44 (traffic_server+0xecfff4) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #5 EThread::EThread(ThreadType, int) /trafficserver/src/iocore/eventsystem/UnixEThread.cc:89 (traffic_server+0xed0c38) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #6 EventProcessor::spawn_event_threads(int, int, unsigned long) /trafficserver/src/iocore/eventsystem/UnixEventProcessor.cc:451 (traffic_server+0xed7bbc) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #7 TasksProcessor::start(int, unsigned long) /trafficserver/src/iocore/eventsystem/Tasks.cc:42 (traffic_server+0xecfa08) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #8 main /trafficserver/src/traffic_server/traffic_server.cc:2155 (traffic_server+0x8fcca0) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)

  Mutex M5 (0xffff933d5bd0) created at:
    #0 pthread_mutex_init <null> (libtsan.so.2+0x5f218) (BuildId: 4fc6d4b1d08e85e4139b0e07fb335a013192e924)
    #1 ink_mutex_init(pthread_mutex_t*) /trafficserver/src/tscore/ink_mutex.cc:54 (traffic_server+0x93aa0c) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #2 ProtectedQueue::ProtectedQueue() /trafficserver/src/iocore/dns/../eventsystem/P_ProtectedQueue.h:38 (traffic_server+0xc2bbd4) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #3 EThread::EThread(ThreadType, int) /trafficserver/src/iocore/eventsystem/UnixEThread.cc:89 (traffic_server+0xed0cc0) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #4 EventProcessor::spawn_event_threads(int, int, unsigned long) /trafficserver/src/iocore/eventsystem/UnixEventProcessor.cc:451 (traffic_server+0xed7bbc) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #5 TasksProcessor::start(int, unsigned long) /trafficserver/src/iocore/eventsystem/Tasks.cc:42 (traffic_server+0xecfa08) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #6 main /trafficserver/src/traffic_server/traffic_server.cc:2155 (traffic_server+0x8fcca0) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)

  Thread T4 '[ET_TASK 0]' (tid=9924, running) created by main thread at:
    #0 pthread_create <null> (libtsan.so.2+0x6073c) (BuildId: 4fc6d4b1d08e85e4139b0e07fb335a013192e924)
    #1 ink_thread_create /trafficserver/include/tscore/ink_thread.h:129 (traffic_server+0xecfdb4) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #2 Thread::start(char const*, void*, unsigned long, std::function<void ()> const&) /trafficserver/src/iocore/eventsystem/Thread.cc:92 (traffic_server+0xed0354) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #3 EventProcessor::spawn_event_threads(int, int, unsigned long) /trafficserver/src/iocore/eventsystem/UnixEventProcessor.cc:469 (traffic_server+0xed7e70) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #4 TasksProcessor::start(int, unsigned long) /trafficserver/src/iocore/eventsystem/Tasks.cc:42 (traffic_server+0xecfa08) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
    #5 main /trafficserver/src/traffic_server/traffic_server.cc:2155 (traffic_server+0x8fcca0) (BuildId: 7206bc40bbaaebae6ec8626a5203b22a73f3cecf)
```

JosiahWI:
The PR title is fairly self-explanatory, but the design choices here deserve explicit mention.

* There is a new `TS_HAS_128BIT_ATOMIC` so that 128-bit head objects depends on both `__sync` and `__atomic`

In our build pipeline, OSX and FreeBSD support this.

* The head_p type is no longer a union with a {pointer, version} field

This has been done to eliminate type punning, which was done all over the implementation, and is UB. The pointer and version are now always set through the macros `FREELIST_POINTER`, `FREELIST_VERSION`, and `SET_FREELIST_POINTER_VERSION`, which use a separate {pointer, version} struct type and `memcpy` on platforms where this is appropriate (see preprocessor defs for the list).

* The head_p type has been changed into a type alias of the data type

This is necessary so that the head of the list will be an atomic integer type instead of an atomic class type to be sure it can use 128bit atomic hardware instructions on platforms that support them.

* Freelist alignment is now adjusted to be satisfy `void*` alignment requirements

This is a minor bug in master.

* The freelist and atomiclist pop operations (called freelist_new for the freelist) are now locked to provide mutual exclusion

This particular change was made to fully fix #11640 - there is a minor data race without it in that the second pointer from the list head can be overwritten by an allocator's placement new before it is read without synchronization in `freelist_new` (a similar argument applies to `atomiclist_pop`; `atomiclist_popall` is unaffected) by another thread, which is going to subsequently find out the list head is stale and retry. Thus, the garbage pointer is not dereferenced, but this is still UB. Benchmarking suggests this has not caused any performance regression (see below). In fact, I wonder if the addition of the lock helped performance under contention.

I have been thinking about other approaches here. One approach is to add an atomic flag to the list head that is set by any thread popping from the list. A thread that has successfully popped can spin on that flag to wait for the completion of any other threads still reading the memory it is about to return. My intuition is that this will be better, but I don't know without benchmarking, and it's a lot more complex than the lock.

Fixes #7398 
Fixes #11640 in release mode only (dummy_forced_read calls still race)
Partially, #13572 

## Previous Work

See #7382. This PR is only a step in the direction of #7382; it retains a lot of the old code structure along with most of its design flaws. If this change is accepted, it should thereafter be possible to apply other design improvements from #7382, such as the fleshed out versioned pointer type, with greater confidence.

## A Few Comments About Assertions

This PR adds a hoard of assertions that check alignment requirements. Most of them are debug assertions - the alignment check on the pointer passed to `freelist_push` is a release assert for now, because it would almost certainly indicate a major issue if it triggered. According to the comment from @bryancall, this assertion was in fact failing before (it was previously a debug assert, and he commented it out). I am hopeful that that issue is now resolved.

## Performance Implications (from tools/benchmark/benchmark_FreeList)

This change represents a significant performance improvement as number of threads increases. The following benchmarks are from tools/benchmark/benchmark_FreeList. The first one is from the old code without this patch. The second one is a benchmark of the code with this patch, but the size of the `head_p` pointer is different because my system does not support 128bit atomics in hardware.

### Before (without this PR) (`-DCMAKE_BUILD_TYPE=Release`, 128 bit `head_p`)

```
nthreads = 1                                   100             1     3.43135 s
                                        34.1753 ms    34.0414 ms      34.32 ms
                                        708.547 us    634.286 us    798.867 us

nthreads = 2                                   100             1     48.6705 s
                                        362.513 ms      336.2 ms    388.466 ms
                                        133.529 ms    124.707 ms     143.15 ms

nthreads = 6                                   100             1     2.44896 m
                                         1.44199 s     1.43396 s     1.45083 s
                                        42.9103 ms    37.2258 ms    50.7251 ms
```

### After (with this PR) (`-DCMAKE_BUILD_TYPE=Release`, 64 bit `head_p`)

```
benchmark name                       samples       iterations    est run time
nthreads = 1                                   100             1     3.11346 s
                                        30.9922 ms    30.8757 ms    31.1318 ms
                                        646.479 us    546.436 us     845.25 us

nthreads = 2                                   100             1     19.0999 s
                                        182.391 ms      180.4 ms    183.905 ms
                                        8.80696 ms    6.94193 ms     13.516 ms
                                        
nthreads = 6                                   100             1     1.50301 m
                                        975.659 ms     960.88 ms    990.707 ms
                                        75.9773 ms    71.2177 ms    81.3398 ms
```


include/tscore/ink_queue.h
```diff
@@ -71,30 +75,62 @@ void ink_queue_load_64(void *dst, void *src);
...
+};
+
+#elif defined(__x86_64__) || defined(__ia64__) || defined(__powerpc64__) || defined(__mips64)
+
+struct head_p_view {
+  int vaddr      : 48;
+  int version    : 15;
+  int vaddr_mode : 1;
+};
+#endif
```
JosiahWI: I put this here as documentation but didn't use it anywhere. Should I remove it?

src/tscore/ink_queue.cc
```diff
@@ -519,115 +576,127 @@ ink_atomiclist_init(InkAtomicList *l, const char *name, uint32_t offset_to_next)
...
+  item = l->head.load();
   do {
-    INK_QUEUE_LD(item, l->head);
     if (TO_PTR(FREELIST_POINTER(item)) == nullptr) {
       return nullptr;
     }
     SET_FREELIST_POINTER_VERSION(next, FROM_PTR(nullptr), FREELIST_VERSION(item) + 1);
-    result = ink_atomic_cas(&l->head.data, item.data, next.data);
-  } while (result == 0);
-  {
```
JosiahWI, line 560: I removed this unnecessary block nesting.

src/tscore/ink_queue.cc
```diff
@@ -519,115 +576,127 @@ ink_atomiclist_init(InkAtomicList *l, const char *name, uint32_t offset_to_next)
...
+  } while (result == false);
+
+  void *ret = TO_PTR(FREELIST_POINTER(item));
+  void *e   = ret;
+  /* fixup forward pointers */
+  while (e) {
+    head_p *e_ = to_head_p(e, l->offset);
+    void   *n  = TO_PTR(FREELIST_POINTER(*e_));
+    SET_FREELIST_POINTER_VERSION(*e_, n, FREELIST_VERSION(*e_));
+    e = n;
```
JosiahWI: If the offset is applied before checking for nullptr, it is possible to pass the tscore tests but fail a bunch of other unit tests (event system and cache tests).

src/tscore/ink_queue.cc
```diff
@@ -222,10 +251,13 @@ freelist_new(InkFreeList *f)
 {
   head_p item;
   head_p next;
-  int    result = 0;
+  bool   result = false;
+
+  std::lock_guard guard{f->m};
+
+  item = f->head.load();
```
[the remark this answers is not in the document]
JosiahWI: This is not a bad idea, but I think the FreeList benchmark stresses this logic heavily and is sufficient validation.

ci/asan_leak_suppression/unit_tests.txt
```diff
@@ -2,3 +2,8 @@
 # not destroyed because it holds a reference to a stack-allocated
 # TestRefCountObj whose free() override calls exit(1).
 leak:test_http_hdr_print_and_copy_aux
+
+# on 64-bit architectures, freelist pointers are stored in a special bitwise
+# format, so LSan cannot find link pointers. It will erroneously determine
+# that the list tail is not reachable.
+leak:freelist_new
```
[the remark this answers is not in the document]
JosiahWI: Good feedback, but this is intentional.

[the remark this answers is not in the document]
JosiahWI: Good feedback, but this is intentional.

zwoop:
[approve ci]

JosiahWI:
Rebased on 0847dc6

src/tscore/ink_queue.cc
```diff
@@ -248,13 +291,26 @@ freelist_new(InkFreeList *f)
...
+      // LSan root-region registration. Each cell on the free-chain stores its
+      // successor as a bitwise-tagged pointer (version bits merged into the high
+      // bits of the word by SET_FREELIST_POINTER_VERSION / FROM_PTR). LSan's
+      // pointer scanner sees the tagged word and cannot recognize it as a heap
+      // pointer, so it classifies the cells as direct leaks. Registering the
+      // entire chunk as a root region tells LSan to scan the chunk's bytes for
+      // pointers; through that scan every cell on the chain becomes reachable and
+      // is reclassified as "still reachable." Registration happens once per chunk
+      // (amortized over chunk_size cells) and is a no-op in non-ASan builds.
+#if defined(__SANITIZE_ADDRESS__) || (defined(__has_feature) && __has_feature(address_sanitizer))
```
JosiahWI: Looks correct; not sure why this is failing

JosiahWI: Interesting: conditional operators don't necessarily short-circuit in preprocessor definitions.

src/tscore/ink_queue.cc
```diff
@@ -517,76 +608,96 @@ ink_freelists_dump(FILE *f)
...
 
 void *
 ink_atomiclist_popall(InkAtomicList *l)
 {
   head_p item;
   head_p next;
-  int    result = 0;
+  bool   result = false;
+
+  item = l->head.load();
```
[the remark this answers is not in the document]
JosiahWI: Nice catch. I believe than analysis is correct.

phongn:
Apologies for stepping on your toes with #13571. My change is relatively constrained, mostly to get RISC-V support working; I was going to think about doing exactly what you do in this PR next.

cmake/Check128BitAtomic.cmake
```diff
@@ -0,0 +1,60 @@
...
+  set(CMAKE_REQUIRED_FLAGS "-Werror -mcx16")
+  check_cxx_source_compiles("${CHECK_PROGRAM}" TS_HAS_128BIT_ATOMIC)
+  set(NEED_MCX16 ${TS_HAS_128BIT_ATOMIC})
+  unset(CMAKE_REQUIRED_FLAGS)
+endif()
+
+set(TS_NEEDS_MCX16_FOR_128BIT_ATOMIC
+    ${NEED_MCX16}
+    CACHE BOOL "Whether -mcx16 is needed to compile 128bit atomics"
+)
```
phongn, line 55: Note: `-mcx16` will not assume that `cmpxchg16b` is available, which is important for if we ever replace `__sync_val_compare_and_swap` with something more modern: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80878

phongn (commented):
Thanks for this — the direction is right, and several pieces here are genuinely valuable: the `std::atomic`/`memcpy`/placement-`new` conversion removes real UB, the alignment assertions are sound, and the `.load()` conversions fix part of #13572. I ran your branch through a fairly deep review (stress tests on both head configurations, a memory-order audit, and benchmarks on the configurations your published numbers could not cover). The implementation itself held up well under stress; the problems I found are in the configuration gate and in what it silently changes. Details below, roughly in severity order.

## 1. The `TS_HAS_128BIT_ATOMIC` probe fails on every GCC build (blocker)

The probe in `Check128BitAtomic.cmake` compiles and links `std::atomic<__int128>` without `-latomic`. GCC never inlines 16-byte atomics, even with `-mcx16` — it always emits `__atomic_compare_exchange_16` library calls ([GCC PR80878](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80878)) — so the probe's link step fails on GCC regardless of flags. Only clang with `-mcx16` passes. Verified on gcc 11.5 / clang 21:

```
g++     plain  : undefined reference to `__atomic_compare_exchange_16'
g++     -mcx16 : undefined reference to `__atomic_compare_exchange_16'
clang++ plain  : undefined reference to `__atomic_compare_exchange_16'
clang++ -mcx16 : LINKS
```

Building this branch confirms it: the gcc build gets `#define TS_HAS_128BIT_ATOMIC 0`, the clang build gets `1`. I suspect this is also why your own "after" benchmark ran with a 64-bit `head_p` — it is not a hardware limitation, it is this probe.

Consequence: `TS_128BIT_QUEUE` is 0 on Linux x86-64 **and aarch64** GCC builds — essentially every distro/production build — which silently drop from the 128-bit head (64-bit version field) to the packed head (15-bit version on x86-64, 11-bit on aarch64), with behavior now diverging by compiler.

## 2. The packed head's narrow version field breaks LogObject's reference counting (high, cascades from 1)

`LogObject::_checkout_write` increments the head version once per checkout, and on buffer swap credits `m_references += FREELIST_VERSION(old_h) - 1` — the version field is a **cumulative checkout count per buffer lifetime**, not just an ABA tag. A default 9 MB log buffer with 100–200-byte entries absorbs roughly 45–90k checkouts before it fills. That exceeds a 15-bit version (32,768) and dwarfs an 11-bit one (2,048). A wrapped version undercounts `m_references`, so the flush thread can free a `LogBuffer` while writers still hold references.

This hazard already exists on master for the packed-tier platforms (ppc64/mips64), where nobody runs logging-heavy ATS. Because of finding 1, this PR would promote it to mainstream Linux gcc builds on x86-64 and aarch64.

## 3. With the head width held constant, the locks are a significant regression

_Ed. note_: This benchmark was run with a Debug release build and is invalid; see next comment below for a valid benchmarking run. 

<del>

The PR's before/after comparison changes two variables at once (128-bit lock-free head → 64-bit packed head + locks). I benchmarked `tools/benchmark/benchmark_FreeList` with the head width held constant (clang builds, where `TS_128BIT_QUEUE=1` on both sides), plus the gcc pairing. Ice Lake, 8 vCPUs, `--benchmark-samples 30`, dev preset:

| Configuration | 1 thread | 2 threads | 6 threads |
|---|---|---|---|
| master, clang (128-bit head, lock-free) | 231.5 ms | 661.7 ms | 2.157 s |
| **this PR, clang (128-bit head + pop locks)** | **365.9 ms (+58%)** | **901.0 ms (+36%)** | **3.323 s (+54%)** |
| master, gcc (128-bit head, `__sync`, lock-free) | 245.8 ms | 759.7 ms | 2.719 s |
| this PR, gcc (packed-64 head + pop locks) | 247.0 ms (+0.5%) | 756.6 ms (−0.4%) | 3.048 s (+12%) |

With the same 128-bit head, the locked version is 36–58% slower at every thread count on this host. Even the packed-64 configuration is neutral at 1–2 threads and ~12% slower at 6. The "significant performance improvement as number of threads increases" in the description appears to be the 128-bit→64-bit head change (finding 1), not the locks. One hypothesis for why the hybrid regime does poorly: locked pops and lock-free CAS pushes contend for the same head cache line, so pushers extend the poppers' critical sections.

</del>

## 4. `ink_atomiclist_remove` is the one mutation left unlocked (medium)

Pops and popall now serialize on `l->m`, but `ink_atomiclist_remove` still CASes the head lock-free, so a remove racing a locked pop can recreate exactly the use-after-free window this PR closes for #11640. Today this is safe only by the legacy discipline (remove callers must be the sole popper) — which the current callers happen to satisfy (`NetHandler` removes from the enable lists only on the owning net thread). Since remove is rare, taking the lock there too is cheap defense and makes the #11640 fix unconditional.

## Smaller notes

* Chunk allocation in `freelist_new` now happens while holding the pop mutex, so a slow hugepage allocation (which can stall in compaction for tens of ms) blocks all poppers of that freelist. In exchange it eliminates the old racing-duplicate-chunk behavior — probably a fine trade, but worth stating in the description.
* `alignment = std::lcm(alignment, alignof(InkFreeList))` conflates the alignment of the `InkFreeList` bookkeeping struct with the alignment (and hence padded `type_size`) of user objects: an 8-byte-aligned small-object freelist now gets 16-byte object stride. Allocating the struct with `alignof(InkFreeList)` separately from the user alignment would avoid the bloat.
* `new (item) void *{}` value-initializes (a dead nullptr store) before the real next-pointer store; `new (item) void *` avoids it. In `freelist_bulkfree` the placement-new also re-runs on every CAS retry.

## What held up under testing

For completeness, the things I tried to break and could not: a 6-thread conservation/double-reachability stress test (each popped item is claimed via an atomic flag, so an item reachable from two places is detected) passes repeatedly on **both** head configurations of this branch; the memory orders are correctly paired (pop CAS acquire/acquire, free CAS release/relaxed, push seq-cst) with the one DEBUG-only `dummy_forced_read` race you already document; all structs that gained a `std::mutex` are `new`-allocated so constructors run; and a full tree build passes with gcc. The core implementation is solid — the issues are all in the gate and its consequences.

## Suggested fix, and relation to #13571

#13571 (just opened for the riscv64 build break, #13555) adds exactly the missing piece: a probe that retries 128-bit atomics **with `-latomic`** and links it when needed (`TS_HAS_128BIT_CAS_LIBATOMIC` / `TS_NEEDS_LIBATOMIC_FOR_CAS`). If `TS_128BIT_QUEUE` accepted that tier, then:

* gcc/x86-64 keeps a 128-bit head: libatomic's ifunc resolves to `cmpxchg16b` at runtime, so it stays lock-free with a small call overhead (~2 ns/op in my microbenchmarks) instead of silently downgrading to 15-bit versions — findings 1 and 2 disappear.
* riscv64 gets a correct lock-based head, so this PR would also fix #13555.
* Your design already tolerates lock-based heads on the pop side — the mutex is there anyway.

Happy to coordinate: if #13571 lands first, this PR rebases onto it and folds `TS_HAS_128BIT_CAS_LIBATOMIC` into the `TS_128BIT_QUEUE` gate, then re-run the benchmark with head width held constant to evaluate the locks on their own merits (or make the pop lock conditional on the platform lacking lock-free atomics).

phongn:
Following up on my earlier review with a correction, a root-cause analysis of the regression, and a prototype that removes it.

## Correction to my benchmark table

My earlier numbers were built with the `dev` preset, which is a Debug/low-optimization build. At low optimization the `std::atomic` wrapper layers in this PR do not inline, which unfairly penalized this PR's code structure relative to master's macro/union code. Your Release methodology was right and my contended numbers were wrong — apologies for the noise. Re-run with `--preset release`, interleaved A/B/C (each binary alternated within each rep, 3 reps, 20 samples; 1-thread means reproduce within 1%):

| Configuration | 1 thread | 6 threads |
|---|---|---|
| master, clang (128-bit head, lock-free) | 56.7 ms | 2.19 s |
| this PR, clang (128-bit head + pop locks) | **73.1 ms (+29%)** | **1.60 s (−27%)** |
| prototype, clang (128-bit head, lock-free + `atomic_ref`) | **56.7 ms (= master)** | 2.02 s |
| master, gcc (128-bit head, `__sync`, lock-free) | 76.7 ms | 2.58 s |
| this PR, gcc (packed-64 head + pop locks) | 47.2 ms | 1.44 s |
| prototype, gcc (packed-64 head, lock-free + `atomic_ref`) | 30.1 ms | 1.59 s |

So in Release, the locks genuinely **win under heavy contention** — your instinct there was correct and my earlier claim of a contended regression was an artifact. The real cost is on the **uncontended fast path**: +29% per op at 1 thread on the 128-bit tier.

## Why this explains the end-to-end regression

I tested three mechanisms in isolation (microbenchmark with a versioned 128-bit Treiber stack):

* **Mutex fixed cost, uncontended** — confirmed: the lock/unlock pair adds ~30-45% per op at 1 thread, matching the table above.
* **False sharing** (the `std::mutex` and the head share a cache line: mutex occupies bytes 0-39, head sits at offset 48) — **exonerated**: padding them onto separate lines changes nothing measurable.
* **Lock convoy under contention** — exonerated: the lock is *better* than CAS retry storms when contended.

In production, ProxyAllocator thread caches absorb most freelist traffic and ProtectedQueue operations rarely collide, so the fleet mostly runs the *uncontended* path — paying the +29% tax on every op while rarely collecting the contended win. That is consistent with the 5-10% end-to-end regression you measured internally.

## Prototype: same UB fix, no locks

The #11640 bug is a *data race* on the speculative next-pointer read — the version-tagged CAS has always correctly rejected stale values; the algorithm does not need mutual exclusion. The prototype keeps everything else in this PR (memcpy views, placement-new lifetime handling, alignment asserts, LSan annotations) and:

* removes the three `lock_guard`s and the now-unused `std::mutex` members, and
* converts every next-slot access (pop's speculative read, push/free/bulkfree stores, popall's fixup walk, remove's reads/writes) to relaxed `std::atomic_ref<void *>`, so the race is defined behavior.

Relaxed slot accesses compile to plain loads/stores; the release/acquire edge stays on the head CAS (slot store *sequenced-before* the release CAS; popper's acquire on the head *synchronizes-with* it). Two details worth noting because they are easy to miss: the value-initializing placement-news (`new (p) void *{}`) were themselves plain stores into raced slots, so the prototype uses default-init (`new (p) void *`, no store) followed by an atomic store; and `ink_atomiclist_remove`'s slot accesses are converted too (it still requires the documented single-popper discipline, as it always has, but its accesses are now defined).

**Branch:** https://github.com/phongn/trafficserver/commit/988a139a57 (stacked on this PR's head, cherry-pick friendly)

**Validation:** passes a 6-thread conservation/double-reachability stress test repeatedly on both head configurations, in both Debug and Release; performance is exactly master at 1 thread (56.7 ms) and ≈ master at 6 threads on the 128-bit tier.

**Honest residuals:** the Debug-only races remain (`dummy_forced_read`, which you already documented, and `DEADBEEF`'s plain fills). And for `InkAtomicList` specifically, a stale popper can touch caller-owned memory after the pop — the same invariant master has relied on for decades (ironclad for the freelist, whose chunks are never unmapped; theoretical for arbitrary atomiclist items). The locks close that fully; `atomic_ref` only removes the UB. If you want maximum conservatism, a hybrid is coherent: lock-free freelist (the measured hot path) + locked atomiclist pops.

Given your internal 5-10% end-to-end regression was likely the uncontended mutex tax, the prototype should recover it — it would be great to see it through your internal benchmarking. Also worth noting from the table: gcc's packed-64 lock-free configuration (30 ms) is by far the fastest, so once the LogObject version-width hazard from my earlier review is addressed, deliberately selecting the packed head could be a legitimate follow-up optimization rather than a probe accident.

phongn:
A follow-up on the `atomic_ref` prototype, because it deserves a direct answer to the strongest objection to any lock-free fix here: **what stops a popper from reading memory the list no longer owns?** An external allocator does not synchronize with a lock-free reader, so if a popped node can be returned to an allocator that unmaps it, a stale reader can fault, and no atomic access fixes a page fault. That is the memory-reclamation problem, and the general solutions (hazard pointers, epochs) are heavyweight — or, as a minimal reader-indication scheme, end up resembling a lock.

The prototype does not solve the general problem. It relies on two narrower invariants, one structural and one already load-bearing in master and in this PR:

**1. `InkFreeList`: the memory is never returned to any allocator.** Chunks are allocated once and are immortal — in master, in this PR, and in the prototype. There is no "memory the freelist no longer owns"; a caller holding an object still sits inside a live chunk. A stale popper's relaxed read can only observe garbage from a caller's live object, which the versioned CAS rejects. This is the type-stable-memory setting the Treiber stack was designed for (same reasoning as the kernel's `SLAB_TYPESAFE_BY_RCU`). No reclamation machinery is needed because there is no reclamation. Notably, the freelist is where the measured regression lives — `benchmark_FreeList` exercises only `InkFreeList`.

**2. `InkAtomicList`: the hazard requires two threads popping the same list concurrently.** A stale reader can only exist if another pop races it. An audit of every consumer in the tree shows all atomic lists are multi-producer/single-consumer: `NetHandler` drains the enable lists on the owning net thread (`NetHandler.cc:228,238`), the AIO thread drains `aio_todo`/`aio_temp_list` for its fd (`AIO.cc:362,513`), the UDP net thread drains its queues (`UnixUDPNet.cc:1341,1885`), `QUICClosedConCollector` drains `closedQueue` on its own continuation, each `LogBufferManager` is drained by its flush thread, and each `ProtectedQueue` by its ethread. Single consumer means no concurrent pop, means the reclamation window is unreachable — which is also why master's racy version never faulted in 25 years of production.

So the honest framing of lock-vs-prototype is not "correct vs. fast": both are correct for the code as it exists. The difference is **guarantee versus discipline**. The lock makes `InkAtomicList` safe for any future caller, including a hypothetical multi-consumer list whose items go to `free()`; the prototype keeps the discipline implicit, exactly as master always has. Two considerations for that judgment:

* This PR already relies on the same discipline: `ink_atomiclist_remove` is not locked, so a remove racing a locked pop reopens the identical window. It is safe today because removers are the sole poppers of their lists — i.e., the single-consumer discipline is load-bearing either way.
* There is a middle path that prices the disagreement cleanly: **lock-free freelist** (structurally safe, and where the uncontended regression is) **plus locked `InkAtomicList` pops** (caller-owned memory keeps the hard guarantee). And the discipline can become a *checked* discipline for free: a DEBUG-only assertion that records the first popper's thread ID per list and asserts subsequent pops match. That is a stronger guardrail than master ever had, at zero production cost.

If ATS ever genuinely needs multi-consumer atomic lists over memory that can be unmapped, the escalation path is the reader-indication scheme sketched in this PR's description (a per-popper announce slot that a reclaiming thread waits on) — a single-slot hazard pointer. Nothing today needs it.

Either way, the deciding number is end-to-end: if the regression seen in production benchmarking is the uncontended mutex tax (the 1-thread microbenchmarks strongly suggest it is), the prototype — or the hybrid — should recover most of it while keeping this PR's UB fixes intact.

JosiahWI:
A general Treiber stack is thread safe for access to the data because the data is part of the stack node. This might be the case for `InkAtomicList` - I would have to review it again. For `InkFreeList` specifically, the general Treiber stack reasoning breaks down, because the _data does not live on the stack_. What lives on the stack are stack bookkeeping objects which are overwritten with some other object by an `Allocator`.  That ends the lifetime of the bookkeeping metadata. Reading the version or pointer out of the freelist node, even through a `std::atomic_ref`, is a data race. But it's more than a data race. It's also very likely to be an aliasing violation, because the memory the freelist is reading from may be in the middle of some subobject in an `HttpHdr` or what have you, and the effective type may no longer be the same as the freelist node metadata type. That metadata object no longer exists. Whether or not it's acceptable to leave that in the program anyway may be an interesting question, but I do not think there is a doubt as to whether or not it is UB. It is UB. And a comparison to a general Treiber stack is a subtle pitfall - for the `InkFreeList` case, the properties that make a Treiber stack thread safe do not all hold. The `InkAtomicList` case may be worth another look - if no one touches the stack metadata of a popped node, then the correctness does hold according to the Treiber stack design.

PR #13171, opened 2026-05-18
Quiet ESI streaming gunzip zero-output logs

bneradt:
ESI streaming gunzip can receive valid gzip chunks that consume input
without producing output, such as an initial header-only chunk. Those
chunks currently emit a `buf below zero` error even though the stream
continues successfully, which creates noisy Traffic Server logs.

This treats zero-output progress as normal streaming inflate behavior
and reserves error logging for actual zlib failures. It also extends the
ESI gzip unit test support to capture error logs and covers the
header-only chunk case.


JosiahWI (commented):
Looks reasonable

shukitchan (approved):

shukitchan:
This looks fine. Thanks.

PR #13174, opened 2026-05-19
proxy.config.ssl.client.CA.cert.filename: overridable

bneradt:
This adds proxy.config.ssl.client.CA.cert.filename, the client CA
certificate path, to the overridable config.


masaori335 (approved):
Looks good. 

This is not blocker, but I guess we want to make `proxy.config.ssl.client.private_key.path` overridable too.

PR #13177, opened 2026-05-19
Correct records.yaml record drift

bneradt:
Several records.yaml entries had drifted from the names or
reloadability registered in RecordsConfig, which leaves generated
config references pointing at records that do not exist or implying
runtime updates that are not supported.

This corrects the obvious doc-only mismatches: typoed directives,
stale QUIC record names and defaults, an unregistered deprecated
admin record, and reloadable annotations that disagree with the
registered update type.


serrislew (approved):

cmcfarlen:

cmcfarlen:
Cherry-picked to 10.2.x

PR #13181, opened 2026-05-20
slice: fix stpcpy off-by-one for header value extraction

traeak:
The off by one wastes a single character, it doesn't overrun.


zwoop:
[approve ci]

JosiahWI (approved):
This simplifies the logic flow.

PR #13182, opened 2026-05-20
Stabilize parallel AuTest helpers

bneradt:
Parallel AuTest runs can execute redirect and redirect_actions at the
same time. Both tests wrote request files into the same generated data
directory, so one test could overwrite the other's inputs and make
redirect fail with unrelated redirect_actions requests.

This gives each redirect test its own generated-data directory and
keeps those directories ignored. This also makes the SIGUSR2 helper
tolerate processes whose psutil cmdline is unavailable, so process
scanning does not crash before signaling Traffic Server.

### DIFF
--- a/tests/gold_tests/logging/ts_process_handler.py
+++ b/tests/gold_tests/logging/ts_process_handler.py
@@ -36,7 +36,10 @@ class GetPidError(Exception):
 def get_ts_process_pid(ts_identifier):
     processes = []
     for proc in psutil.process_iter(['cmdline']):
-        commandline = ' '.join(proc.info['cmdline'])
+        cmdline = proc.info.get('cmdline', [])
+        if not cmdline:
+            continue
+        commandline = ' '.join(cmdline)
         if '/traffic_server' in commandline and ts_identifier in commandline:
             return proc
     raise GetPidError("Could not find a traffic_server process")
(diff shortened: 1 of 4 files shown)


tests/gold_tests/redirect/.gitignore
```diff
@@ -1 +1,3 @@
 generated_test_data/
```
JosiahWI: Should `generated_test_data` be removed from .gitignore now?

tests/gold_tests/logging/ts_process_handler.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, message):
 def get_ts_process_pid(ts_identifier):
     processes = []
     for proc in psutil.process_iter(['cmdline']):
-        commandline = ' '.join(proc.info['cmdline'])
+        commandline = ' '.join(proc.info.get('cmdline') or [])
```
JosiahWI: Would a typical Python dev understand at a glance what's going on here? You could use the default value for `get` to make this more conventional for Python.

```suggestion
commandline = ' '.join(proc.info.get('cmdline', [])
```

JosiahWI (changes requested):
Good catch. I have a few minor concerns.

tests/gold_tests/redirect/.gitignore
bneradt: Good point.

tests/gold_tests/logging/ts_process_handler.py
bneradt: Yes, that is better. Thanks.

JosiahWI (approved):

cmcfarlen:

cmcfarlen:
Cherry-picked to 10.2.x

PR #13186, opened 2026-05-20
HTTP/3 support via OpenSSL 3.5

bneradt:
# Overview

This now contains two commits based directly on `master`. The prerequisite
HTTP/2 fix from #13363 and H3 quiche/BoringSSL fixes and tests from #13213
have landed and are no longer part of this PR's commit stack.

# First Commit: HTTP/3 via OpenSSL 3.5 + quiche

Fedora now ships OpenSSL 3.5 with the third-party QUIC TLS callback API, but
quiche still links against the older quictls/BoringSSL symbols. ATS therefore
could not use the system OpenSSL library for downstream HTTP/3 without
bringing in a different TLS stack.

This adds CMake detection for the OpenSSL callback API and provides a private
compatibility layer that maps quiche's legacy hooks onto
`SSL_set_quic_tls_cbs`. This requires static quiche in that mode so ATS
resolves the shim symbols locally and links the final binaries against the
system OpenSSL libraries.

This also relaxes verifier-only HTTP/3 AuTest gates that do not execute curl,
so those tests can run when ATS has QUIC support but the installed curl lacks
HTTP/3.

# Second Commit: HTTP/3 via OpenSSL 3.5 QUIC

OpenSSL 3.5 can terminate QUIC connections directly, but ATS only had a
quiche-backed HTTP/3 listener. Operators who want to use the system OpenSSL
QUIC stack needed a separate downstream backend without changing the existing
quiche path or origin HTTP/3 scope.

This adds an optional `ENABLE_OPENSSL_QUIC` backend that uses OpenSSL's native
QUIC listener and stream APIs for downstream HTTP/3. This keeps the backend
mutually exclusive with quiche, exposes `TS_HAS_OPENSSL_QUIC`, and shares
ATS's existing HTTP/3 stream handling above the transport.

This also installs native-QUIC TLS callbacks for ALPN and SNI certificate
selection before ATS has a QUIC NetVC to bind. OpenSSL native QUIC does not
make a selected `SSL_CTX` certificate active via `SSL_set_SSL_CTX` alone, so
this applies the selected certificate, key, and chain to the connection SSL.

This also broadens client-side HTTP/3 tests to run with either backend and
hardens transaction cleanup when OpenSSL closes stream state before ATS
finishes teardown. This caches stream identifiers, declines listener-time
QUIC tickets until a NetVC is bound, and adds focused H3 lifecycle and
session-ticket coverage.

# Testing

Validated in the Fedora 44 `asfats3` container:

- Generic Fedora CI build with QUIC disabled.
- System OpenSSL 3.5 + static quiche build and HTTP/3 unit tests.
- Native OpenSSL 3.5 QUIC build and HTTP/3 unit tests.
- All ten `h3*` AuTests with both OpenSSL 3.5 backends.
- The complete 10.2.x backport stack with BoringSSL/quiche and native OpenSSL
  3.5. The BoringSSL run passes all nine applicable tests; the OpenSSL-only
  session-ticket test is skipped as designed.

# Production Testing

This is running in docs for
https://docs.trafficserver.apache.org/ targets via Alt-Svc header rewrite
rules to negotiate `h3` when supported.

### DIFF
--- a/src/iocore/net/OpenSSLQUICNetVConnection.cc
+++ b/src/iocore/net/OpenSSLQUICNetVConnection.cc
@@ -1,6 +1,6 @@
 /** @file
 
-  A brief file description
+  OpenSSL native QUIC NetVConnection support.
 
   @section license License
 
@@ -28,23 +28,53 @@
 #include "api/APIHook.h"
 #include "iocore/eventsystem/EThread.h"
 #include "iocore/net/QUICMultiCertConfigLoader.h"
-#include "iocore/net/UDPPacket.h"
-#include "iocore/net/quic/QUICEvents.h"
-#include "iocore/net/quic/QUICStream.h"
-#include "iocore/net/quic/QUICGlobals.h"
 #include "iocore/net/SSLAPIHooks.h"
+#include "iocore/net/quic/QUICApplicationMap.h"
+#include "iocore/net/quic/QUICEvents.h"
+#include "iocore/net/quic/QUICGlobals.h"
+#include "iocore/net/quic/QUICStream.h"
+#include "iocore/net/quic/QUICStreamManager.h"
 #include "tscore/ink_config.h"
 
-#include <netinet/in.h>
-#include <quiche.h>
+#include <openssl/err.h>
+#include <openssl/quic.h>
+#include <openssl/ssl.h>
+
+#include <algorithm>
+#include <cerrno>
 #include <cstring>
+#include <vector>
 
 namespace
 {
-constexpr ink_hrtime WRITE_READY_INTERVAL = HRTIME_MSECONDS(2);
+constexpr ink_hrtime OPENSSL_QUIC_EVENT_INTERVAL = HRTIME_MSECONDS(2);
 
 DbgCtl dbg_ctl_quic_net{"quic_net"};
 DbgCtl dbg_ctl_v_quic_net{"v_quic_net"};
+
+class OpenSSLQUICStreamManager : public QUICStreamManager
+{
+public:
+  OpenSSLQUICStreamManager(QUICContext *context, QUICApplicationMap *app_map, QUICNetVConnection *vc)
+    : QUICStreamManager(context, app_map), _vc(vc)
+  {
+  }
+
+  QUICConnectionErrorUPtr
+  create_uni_stream(QUICStreamId &new_stream_id) override
+  {
+    return this->_vc->create_openssl_stream(SSL_STREAM_FLAG_UNI | SSL_STREAM_FLAG_NO_BLOCK, new_stream_id);
+  }
+
+  QUICConnectionErrorUPtr
+  create_bidi_stream(QUICStreamId &new_stream_id) override
+  {
+    return this->_vc->create_openssl_stream(SSL_STREAM_FLAG_NO_BLOCK, new_stream_id);
+  }
+
+private:
+  QUICNetVConnection *_vc = nullptr;
+};
 
 } // end anonymous namespace
 
@@ -67,34 +97,65 @@ QUICNetVConnection::QUICNetVConnection()
 QUICNetVConnection::~QUICNetVConnection() {}
 
 void
-QUICNetVConnection::init(QUICVersion /* version ATS_UNUSED */, QUICConnectionId /* peer_cid ATS_UNUSED */,
-                         QUICConnectionId /* original_cid ATS_UNUSED */, UDPConnection *, QUICPacketHandler *)
+QUICNetVConnection::init(SSL *ssl, QUICPacketHandler *packet_handler)
 {
+  SET_HANDLER((NetVConnHandler)&QUICNetVConnection::acceptEvent);
+
+  this->_ssl            = ssl;
+  this->_packet_handler = packet_handler;
+  this->_quic_connection_id.randomize();
+  this->_initial_source_connection_id = this->_quic_connection_id;
+  this->_cid_text                     = this->_quic_connection_id.hex();
+
+  SSL_set_ex_data(ssl, QUIC::ssl_quic_qc_index, static_cast<QUICConnection *>(this));
+  SSL_set_blocking_mode(ssl, 0);
+  SSL_set_event_handling_mode(ssl, SSL_VALUE_EVENT_HANDLING_MODE_IMPLICIT);
+  SSL_set_incoming_stream_policy(ssl, SSL_INCOMING_STREAM_POLICY_ACCEPT, 0);
+  SSL_set_default_stream_mode(ssl, SSL_DEFAULT_STREAM_MODE_NONE);
+
+  QUICConfig::scoped_config params;
+  SSL_set_feature_request_uint(ssl, SSL_VALUE_QUIC_IDLE_TIMEOUT, params->no_activity_timeout_in());
+  SSL_set_feature_request_uint(ssl, SSL_VALUE_QUIC_STREAM_BIDI_LOCAL_AVAIL, params->initial_max_streams_bidi_in());
+  SSL_set_feature_request_uint(ssl, SSL_VALUE_QUIC_STREAM_UNI_LOCAL_AVAIL, params->initial_max_streams_uni_in());
+
+  this->_bindSSLObject();
+  if (auto const servername = quic_sni_server_name(ssl); !servername.empty()) {
+    this->set_sni_server_name(servername);
+  }
 }
 
 void
-QUICNetVConnection::init(QUICVersion /* version ATS_UNUSED */, QUICConnectionId /* peer_cid ATS_UNUSED */,
-                         QUICConnectionId original_cid, QUICConnectionId /* first_cid ATS_UNUSED */,
-                         QUICConnectionId /* retry_cid ATS_UNUSED */, UDPConnection *udp_con, quiche_conn *quiche_con,
-                         QUICPacketHandler *packet_handler, QUICConnectionTable *ctable, SSL *ssl)
+QUICNetVConnection::set_quic_endpoints(IpEndpoint const &local, IpEndpoint const &remote)
 {
-  SET_HANDLER((NetVConnHandler)&QUICNetVConnection::acceptEvent);
-  this->_udp_con                     = udp_con;
-  this->_quiche_con                  = quiche_con;
-  this->_packet_handler              = packet_handler;
-  this->_original_quic_connection_id = original_cid;
-  this->_quic_connection_id.randomize();
-  this->_initial_source_connection_id = this->_quic_connection_id;
+  this->local_addr      = local;
+  this->remote_addr     = remote;
+  this->got_local_addr  = true;
+  this->got_remote_addr = true;
+}
 
-  if (ctable) {
-    this->_ctable = ctable;
-    this->_ctable->insert(this->_quic_connection_id, this);
-    this->_ctable->insert(this->_original_quic_connection_id, this);
+QUICConnectionErrorUPtr
+QUICNetVConnection::create_openssl_stream(uint64_t flags, QUICStreamId &new_stream_id)
+{
+  if (this->_ssl == nullptr) {
+    return std::make_unique<QUICConnectionError>(QUICTransErrorCode::INTERNAL_ERROR, "QUIC connection is not initialized");
   }
 
-  this->_ssl = ssl;
-  SSL_set_ex_data(ssl, QUIC::ssl_quic_qc_index, static_cast<QUICConnection *>(this));
-  this->_bindSSLObject();
+  SSL *stream_ssl = SSL_new_stream(this->_ssl, flags);
+  if (stream_ssl == nullptr) {
+    return std::make_unique<QUICConnectionError>(QUICTransErrorCode::INTERNAL_ERROR, "Failed to create QUIC stream");
+  }
+
+  SSL_set_blocking_mode(stream_ssl, 0);
+  SSL_set_event_handling_mode(stream_ssl, SSL_VALUE_EVENT_HANDLING_MODE_IMPLICIT);
+  SSL_set_mode(stream_ssl, SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER | SSL_MODE_ENABLE_PARTIAL_WRITE);
+
+  new_stream_id = SSL_get_stream_id(stream_ssl);
+  this->_openssl_streams.emplace(new_stream_id, stream_ssl);
+
+  [[maybe_unused]] QUICConnectionError err;
+  this->_stream_manager->create_stream(new_stream_id, err);
+
+  return nullptr;
 }
 
 void
@@ -103,17 +164,11 @@ QUICNetVConnection::free()
   this->free_thread(this_ethread());
 }
 
-// called by ET_UDP
 void
 QUICNetVConnection::remove_connection_ids()
 {
-  if (this->_ctable) {
-    this->_ctable->erase(this->_quic_connection_id, this);
-    this->_ctable->erase(this->_original_quic_connection_id, this);
-  }
 }
 
-// called by ET_UDP
 void
 QUICNetVConnection::destroy(EThread *t)
 {
@@ -135,13 +190,18 @@ QUICNetVConnection::free_thread(EThread * /* t ATS_UNUSED */)
 {
   QUICConDebug("Free connection");
 
-  this->_udp_con = nullptr;
+  this->_unschedule_openssl_event();
 
-  quiche_conn_free(this->_quiche_con);
-  this->_quiche_con = nullptr;
+  for (auto &[stream_id, stream_ssl] : this->_openssl_streams) {
+    SSL_free(stream_ssl);
+  }
+  this->_openssl_streams.clear();
 
-  this->_unschedule_quiche_timeout();
-  this->_unschedule_packet_write_ready();
+  if (this->_ssl != nullptr) {
+    this->_unbindSSLObject();
+    SSL_free(this->_ssl);
+    this->_ssl = nullptr;
+  }
 
   this->_application_map.reset();
   this->_stream_manager.reset();
@@ -152,8 +212,10 @@ QUICNetVConnection::free_thread(EThread * /* t ATS_UNUSED */)
   TLSEventSupport::clear();
   TLSCertSwitchSupport::_clear();
 
-  this->_packet_handler->close_connection(this);
-  this->_packet_handler = nullptr;
+  if (this->_packet_handler != nullptr) {
+    this->_packet_handler->close_connection(this);
+    this->_packet_handler = nullptr;
+  }
 }
 
 void
@@ -164,30 +226,22 @@ QUICNetVConnection::reenable(VIO * /* vio ATS_UNUSED */)
 int
 QUICNetVConnection::state_handshake(int event, Event *data)
 {
-  if (quiche_conn_is_established(this->_quiche_con)) {
-    this->_switch_to_established_state();
-    return this->handleEvent(event, data);
+  if (data == this->_packet_write_ready) {
+    this->_packet_write_ready = nullptr;
   }
 
   switch (event) {
-  case QUIC_EVENT_PACKET_READ_READY:
-    this->_handle_read_ready();
-    break;
-  case QUIC_EVENT_PACKET_WRITE_READY:
-    this->_close_packet_write_ready(data);
-    this->_handle_write_ready();
-    // Reschedule WRITE_READY
-    this->_schedule_packet_write_ready(true);
-    break;
+  case EVENT_IMMEDIATE:
   case EVENT_INTERVAL:
-    this->_close_quiche_timeout(data);
-    this->_handle_interval();
+  case QUIC_EVENT_PACKET_READ_READY:
+  case QUIC_EVENT_PACKET_WRITE_READY:
+    this->_handle_openssl_events();
     break;
   case VC_EVENT_EOS:
   case VC_EVENT_ERROR:
   case VC_EVENT_ACTIVE_TIMEOUT:
   case VC_EVENT_INACTIVITY_TIMEOUT:
-    _unschedule_packet_write_ready();
+    this->_unschedule_openssl_event();
     this->_propagate_event(event);
     this->closed = 1;
     break;
@@ -196,8 +250,15 @@ QUICNetVConnection::state_handshake(int event, Event *data)
     break;
   }
 
-  if (this->closed != 1 && (quiche_conn_is_closed(this->_quiche_con) || quiche_conn_is_draining(this->_quiche_con))) {
+  if (this->closed != 1 && SSL_is_init_finished(this->_ssl)) {
+    this->_switch_to_established_state();
+    this->_handle_openssl_events();
+  }
+
+  if (this->closed != 1 && this->_openssl_connection_closed()) {
     this->_schedule_closing_event();
+  } else if (this->closed != 1) {
+    this->_schedule_openssl_event();
   }
 
   return EVENT_DONE;
@@ -206,29 +267,26 @@ QUICNetVConnection::state_handshake(int event, Event *data)
 int
 QUICNetVConnection::state_established(int event, Event *data)
 {
-  if (!this->_quiche_con) {
-    // Connection has been closed.
+  if (this->_ssl == nullptr) {
     return EVENT_DONE;
   }
+
+  if (data == this->_packet_write_ready) {
+    this->_packet_write_ready = nullptr;
+  }
+
   switch (event) {
-  case QUIC_EVENT_PACKET_READ_READY:
-    this->_handle_read_ready();
-    break;
-  case QUIC_EVENT_PACKET_WRITE_READY:
-    this->_close_packet_write_ready(data);
-    this->_handle_write_ready();
-    // Reschedule WRITE_READY
-    this->_schedule_packet_write_ready(true);
-    break;
+  case EVENT_IMMEDIATE:
   case EVENT_INTERVAL:
-    this->_close_quiche_timeout(data);
-    this->_handle_interval();
+  case QUIC_EVENT_PACKET_READ_READY:
+  case QUIC_EVENT_PACKET_WRITE_READY:
+    this->_handle_openssl_events();
     break;
   case VC_EVENT_EOS:
   case VC_EVENT_ERROR:
   case VC_EVENT_ACTIVE_TIMEOUT:
   case VC_EVENT_INACTIVITY_TIMEOUT:
-    _unschedule_packet_write_ready();
+    this->_unschedule_openssl_event();
     this->_propagate_event(event);
     this->closed = 1;
     break;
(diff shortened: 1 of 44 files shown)


tests/gold_tests/h3/h3_session_ticket.test.py
```diff
@@ -0,0 +1,111 @@
...
+        return (
+            'session_path="{session_file}"; '
+            'timeout 5 bash -c \'sleep 1 | '
+            'openssl s_client -quic -alpn h3 '
+            '-connect 127.0.0.1:{port} -servername foo.com '
+            '{session_option} "$$0" -brief -ign_eof\' "$$session_path"; '
+            'status=$$?; '
+            'if [ "$$status" -ne 0 ] && [ "$$status" -ne 124 ]; then exit "$$status"; fi; '
+            'test -s "$$session_path"'.format(
+                session_file=self.session_file, port=self._ts.Variables.ssl_port, session_option=session_option))
```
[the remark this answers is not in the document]
bneradt: No change here. The doubled dollars are intentional because AuTest processes this command through Python `string.Template`, where `$$` escapes to a literal `$`. After template substitution, the shell receives `$0`, `$?`, `$status`, etc.

I demonstrated that you suggested single-dollar form fails locally: AuTest fails before shell execution with `ValueError: Invalid placeholder`.

tests/gold_tests/timeout/quic_no_activity_timeout.test.py
```diff
@@ -117,18 +117,19 @@ def run(self, check_for_max_idle_timeout=False):
...
+    test1 = Test_quic_no_activity_timeout(
+        "Test ts.quic.no_activity_timeout_in(quic max_idle_timeout) with a 5s delay",
+        no_activity_timeout_in=3000,  # 3s `max_idle_timeout`
+        replay_keys="delay5s",
+        gold_file="gold/quic_no_activity_timeout.gold")
+    test1.run(check_for_max_idle_timeout=True)
+
+    # QUIC Ignores the default_inactivity_timeout config, so the ts.quic.no_activity_timeout_in
+    # should be honor
+    test2 = Test_quic_no_activity_timeout(
```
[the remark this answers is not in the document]
bneradt: Fixed. Changed “should be honor” to “should be honored”.

include/iocore/net/quic/QUICStreamAdapter.h
```diff
@@ -39,6 +39,7 @@ class QUICStreamAdapter
 
   virtual int64_t    write(QUICOffset offset, const uint8_t *data, uint64_t data_length, bool fin) = 0;
   Ptr<IOBufferBlock> read(size_t len);
+  void               consume(size_t len);
   virtual bool       is_eos()     = 0;
```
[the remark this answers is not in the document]
bneradt: Fixed. The test helper now calls `consume(nread)` after copying bytes out of the `QUICStreamAdapter`, so the test path matches the intended read/consume contract.

src/iocore/net/OpenSSLQUICPacketHandler.cc
```diff
@@ -0,0 +1,281 @@
...
+      SSL_free(ssl);
+      continue;
+    }
+
+    IpEndpoint remote_addr;
+    remote_addr.setToAnyAddr(this->server.accept_addr.sa.sa_family);
+
+    vc->init(ssl, this);
+    vc->id = net_next_connection_number();
+    vc->set_quic_endpoints(this->server.accept_addr, remote_addr);
```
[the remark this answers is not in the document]
bneradt: Fixed. The OpenSSL QUIC listener now extracts the peer address from the datagram BIO via `BIO_dgram_get_peer()` and converts it to `IpEndpoint`, falling back to the old wildcard address only if that lookup fails.

src/iocore/net/OpenSSLQUICNetVConnection.cc
```diff
@@ -0,0 +1,1042 @@
...
+QUICNetVConnection::cids() const
+{
+  return this->_cid_text;
+}
+
+QUICFiveTuple const
+QUICNetVConnection::five_tuple() const
+{
+  return {};
+}
```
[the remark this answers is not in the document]
bneradt: Fixed. `five_tuple()` now returns the actual remote/local UDP tuple for OpenSSL QUIC connections, and `QUICFiveTuple` default members are initialized so the default constructor is safe as well.

src/proxy/http3/Http3StreamDataVIOAdaptor.cc
```diff
@@ -53,17 +57,17 @@ Http3StreamDataVIOAdaptor::handle_frame(std::shared_ptr<const Http3Frame> frame,
...
 
-  this->_buffer->dealloc_reader(reader);
-  this->_sink_vio->nbytes = this->_total_data_length;
+  SCOPED_MUTEX_LOCK(lock, this->_sink_vio->mutex, this_ethread());
+  MIOBuffer *writer    = this->_sink_vio->get_writer();
+  int64_t    delivered = writer->write(this->_reader, this->_reader->read_avail());
+  this->_reader->consume(delivered);
+  this->_sink_vio->nbytes  = this->_total_data_length;
+  this->_sink_vio->ndone  += delivered;
+  this->_finalized         = true;
```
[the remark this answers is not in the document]
bneradt: Fixed. The H3 header adaptor now accounts for bytes written into the sink VIO, and DATA finalization no longer sets nbytes to DATA payload length only. The read side now completes with nbytes == ndone after both headers and DATA have been produced, preserving the VIO accounting contract.

phongn (commented):
🤖 _Automated review by **Claude Code**, posted on behalf of @phongn._

Inline comments below are from an offline analysis pass over the diff (correctness + security focus). They range from a couple of suggested fixes to confirmations and things worth verifying. Summary findings, in rough priority: (1) `stream_write_capacity` error path, (2) `create_*_stream` quiche behavior change to regression-test, (3) whole-DATA-frame buffering, (4) redundant SNI/cert callbacks, (5) a maintenance note on stream-map iteration, and (6) a suggestion to give `OpenSSLQuicCompat.cc` a dedicated security pass. Nice work overall — the `QUICStreamIO` abstraction and test coverage are solid.

src/iocore/net/OpenSSLQUICNetVConnection.cc
```diff
@@ -0,0 +1,1046 @@
...
+  if (it == this->_openssl_streams.end()) {
+    return -1;
+  }
+
+  uint64_t avail = 0;
+  if (SSL_get_stream_write_buf_avail(it->second, &avail) == 1) {
+    return std::min<uint64_t>(avail, 16 * 1024);
+  }
+
+  return 16 * 1024;
```
phongn: 🤖 _Automated review by **Claude Code**, posted on behalf of @phongn._

When `SSL_get_stream_write_buf_avail()` fails (stream reset / concluded / closed), this returns a fabricated 16 KB of capacity. `QUICStream::send_data()` treats that as positive "room to write" and will attempt a write that then fails. Suggest returning `0` (or `-1`) on the failure path so a non-writable stream isn't reported as writable.

src/iocore/net/OpenSSLQUICNetVConnection.cc
```diff
@@ -0,0 +1,1046 @@
...
+      [[maybe_unused]] QUICConnectionError err;
+      this->_stream_manager->create_stream(stream_id, err);
+    }
+  }
+}
+
+void
+QUICNetVConnection::_process_openssl_streams()
+{
+  for (auto &[stream_id, stream_ssl] : this->_openssl_streams) {
```
phongn: 🤖 _Automated review by **Claude Code**, posted on behalf of @phongn._

Maintenance heads-up: this range-`for` over `_openssl_streams` is only safe because nothing in `receive_data()`/`send_data()` synchronously re-enters `create_openssl_stream()` — which does `_openssl_streams.emplace(...)` and could rehash, invalidating this iterator (UB). That invariant holds today (server-initiated streams are all created in `Http3App::start()` before the first call here, and the data paths only `schedule_imm()`), but it's implicit. Consider snapshotting the stream ids before the loop, or adding a comment documenting the invariant, so a future change that opens a stream from within a data callback doesn't silently introduce UB.

src/iocore/net/OpenSSLQUICNetVConnection.cc
```diff
@@ -0,0 +1,1046 @@
...
+}
+
+QUICConnectionErrorUPtr
+QUICNetVConnection::create_openssl_stream(uint64_t flags, QUICStreamId &new_stream_id)
+{
+  if (this->_ssl == nullptr) {
+    return std::make_unique<QUICConnectionError>(QUICTransErrorCode::INTERNAL_ERROR, "QUIC connection is not initialized");
+  }
+
+  SSL *stream_ssl = SSL_new_stream(this->_ssl, flags | SSL_STREAM_FLAG_NO_BLOCK);
```
phongn: 🤖 _Automated review by **Claude Code**, posted on behalf of @phongn._

Minor: `SSL_STREAM_FLAG_NO_BLOCK` is already OR'd in by both callers (`OpenSSLQUICStreamManager::create_uni_stream`/`create_bidi_stream`), so OR-ing it again here is redundant. Separately, on `SSL_new_stream` failure this always reports `STREAM_LIMIT_ERROR`, even when the cause isn't a stream-limit condition.

src/iocore/net/quic/QUICStreamManager.cc
```diff
@@ -113,14 +113,38 @@ QUICStreamManager::create_stream(QUICStreamId stream_id, QUICConnectionError & /
 }
 
 QUICConnectionErrorUPtr
-QUICStreamManager::create_uni_stream(QUICStreamId /* new_stream_id ATS_UNUSED */)
+QUICStreamManager::create_uni_stream(QUICStreamId &new_stream_id)
```
phongn, line 116: 🤖 _Automated review by **Claude Code**, posted on behalf of @phongn._

Behavior change worth flagging on the **quiche** path. Previously `create_uni_stream`/`create_bidi_stream` were no-op stubs (param unused, returned `nullptr` without setting the id), so `Http3App::start()` ended up registering the control stream under id `0`. This now allocates a real server-initiated id (`SERVER_UNI + 4n`) and calls `create_stream()`. That's an improvement, but it modifies the shared base class used by the quiche backend — not just the new OpenSSL path. Please confirm the quiche backend (`ci-fedora-quiche`) is regression-tested: the H3 control stream now genuinely opens where it effectively didn't before.

src/proxy/http3/Http3Frame.cc
```diff
@@ -251,6 +251,12 @@ Http3DataFrame::payload_length() const
   return this->_length;
 }
 
+bool
+Http3DataFrame::_parse()
```
phongn, line 270: 🤖 _Automated review by **Claude Code**, posted on behalf of @phongn._

This gates a DATA frame on its *entire* payload being buffered before dispatch (the dispatcher's cloned reader pins the stream buffer blocks until then, and `handle_frame` copies the whole payload into `_buffer` in one shot). Since a peer may send a body as a single arbitrarily-large DATA frame, peak per-stream memory now scales with the largest single frame (held transiently in both the stream buffer and `_buffer`).

It does fix a real bug — the old inherited `_parse() { return true; }` re-dispatched a DATA frame once per arriving byte (that's the 5→1 change in `test_Http3FrameDispatcher`). And `finalize()` already flushed the whole body at once in both old and new code, so this isn't replacing streaming with buffering. Just worth confirming the whole-frame buffering is acceptable for large uploads/downloads — the tests top out at 300 KB.

src/iocore/net/QUICMultiCertConfigLoader.cc
```diff
@@ -88,12 +90,191 @@ QUICCertConfig::release(SSLCertLookup *lookup)
...
+           SSL_TLSEXT_ERR_OK :
+           SSL_TLSEXT_ERR_ALERT_FATAL;
+}
+} // end anonymous namespace
+
+void
+QUICMultiCertConfigLoader::_set_handshake_callbacks(SSL_CTX *ctx)
+{
+  Dbg(dbg_ctl_quic, "installing OpenSSL QUIC cert callback on SSL_CTX %p", ctx);
+  SSL_CTX_set_client_hello_cb(ctx, quic_client_hello_callback, nullptr);
```
phongn, line 278: 🤖 _Automated review by **Claude Code**, posted on behalf of @phongn._

`_set_handshake_callbacks` installs three overlapping callbacks that all funnel into `select_quic_sni_context`: `SSL_CTX_set_client_hello_cb`, `SSL_CTX_set_tlsext_servername_callback`, and `SSL_CTX_set_cert_cb`. They'll each fire per handshake, repeating the `SSL_set_SSL_CTX` + manual cert re-application up to three times. The client-hello path also uses a hand-rolled SNI-extension parser (`servername_from_client_hello`) that re-implements what OpenSSL already provides via the servername callback.

Recommend collapsing to a single callback (the `cert_cb` or `client_hello_cb` alone is sufficient for cert switching) and dropping the bespoke parser — smaller attack surface and no risk of the two SNI paths diverging. (The parser's bounds checks look correct on my read; this is about redundancy, not a parsing defect.)

src/iocore/net/quic/OpenSSLQuicCompat.cc
```diff
@@ -0,0 +1,477 @@
...
+  {OSSL_FUNC_SSL_QUIC_TLS_YIELD_SECRET,         reinterpret_cast<void (*)(void)>(yield_secret_cb)        },
+  {OSSL_FUNC_SSL_QUIC_TLS_GOT_TRANSPORT_PARAMS, reinterpret_cast<void (*)(void)>(got_transport_params_cb)},
+  {OSSL_FUNC_SSL_QUIC_TLS_ALERT,                reinterpret_cast<void (*)(void)>(alert_cb)               },
+  {0,                                           nullptr                                                  },
+};
+
+} // namespace
+
+extern "C" int
+SSL_set_quic_method(SSL *ssl, SSL_QUIC_METHOD const *quic_method)
```
phongn, line 394: 🤖 _Automated review by **Claude Code**, posted on behalf of @phongn._

This compatibility shim is the most security-sensitive part of the PR — it bridges TLS secret material (`yield_secret_cb`, the crypto send/recv/release callbacks) between quiche's legacy QUIC-TLS API and OpenSSL 3.5's third-party callback API. On my read the read/write secret pairing is correct for TLS 1.3 (equal-length secrets, paired delivery via `deliver_paired_secrets`), and `SSL_set_quic_use_legacy_codepoint` being a no-op is fine for 3.5's standard transport-param codepoint.

No specific defect to report here — flagging it for a dedicated, focused review pass plus an interop matrix (quiche peer ↔ this shim, both client and server roles), given the blast radius if secret translation is subtly wrong.

src/iocore/net/OpenSSLQUICPacketHandler.cc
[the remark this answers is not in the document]
phongn: 🤖 _Automated review by **Claude Code**, posted on behalf of @phongn._

Confirming this against OpenSSL 3.5.5 (the system lib here). The literal "set to any" is now just the fallback — the code does try `get_quic_peer_addr()` first — but the underlying concern holds:

- OpenSSL 3.5 exposes **no per-connection server-side peer-address getter** (only the client-side setter `SSL_set1_initial_peer_addr`); there is no `SSL_get*peer*addr`.
- `SSL_get_rbio()` on a listener-accepted connection returns the listener's **shared** UDP socket BIO, so `BIO_dgram_get_peer()` returns whatever peer was last recorded on that shared BIO — not reliably *this* connection's. Under concurrency it can return the wrong peer, and on failure it falls back to `0.0.0.0`/`::`.

Either way the client IP that reaches access logs, ACLs, transparency, and plugins can be wrong. The curl/proxy-verifier tests run over loopback and don't assert on `%<chi>`, so they won't catch this — a test that checks the logged client IP would. Worth confirming whether OpenSSL intends any supported way to recover the per-connection peer here.

src/proxy/http3/Http3StreamDataVIOAdaptor.cc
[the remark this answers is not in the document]
phongn: 🤖 _Automated review by **Claude Code**, posted on behalf of @phongn._

Traced this one. Both the header adaptor and this data adaptor sink to the same VIO (`&_read_vio`, the consumer-facing read VIO from `do_io_read`). I don't think it produces a negative `ntodo()` in practice: that `ndone` is producer-side accounting, `do_io_read()` resets it to `0` per read cycle, and the raw QUIC-stream consumption is tracked on a separate VIO (`_info.read_vio->ndone`). The varied body-size tests (0 / 100 / 120 KB / 300 KB, GET and POST) passing is consistent with that.

That said, the broader point stands: `finalize()` overwriting the consumer-set `nbytes` with `ndone` (and both adaptors hand-incrementing `ndone`) leans on a non-obvious ordering invariant — it's only safe because of the `do_io_read` reset. A short comment explaining why `nbytes = ndone` is safe here, or deriving completion from end-of-stream rather than overwriting `nbytes`, would make this more robust against future changes.

zwoop:
[approve ci autest 2]

src/iocore/net/OpenSSLQUICNetVConnection.cc
bneradt: Fixed. stream_write_capacity() now returns 0 if SSL_get_stream_write_buf_avail() fails, so a reset/closed stream is not reported as writable.

src/iocore/net/OpenSSLQUICNetVConnection.cc
bneradt: Fixed. _process_openssl_streams() now snapshots stream IDs and re-finds each stream before processing, so stream map mutation cannot invalidate the active iteration.

src/iocore/net/quic/QUICStreamVCAdapter.cc
```diff
@@ -84,18 +86,40 @@ QUICStreamVCAdapter::_read(size_t len)
     SCOPED_MUTEX_LOCK(lock, this->_write_vio.mutex, this_ethread());
 
     IOBufferReader *reader = this->_write_vio.get_reader();
-    block                  = make_ptr<IOBufferBlock>(reader->get_current_block()->clone());
+    if (reader->get_current_block() == nullptr || reader->block_read_avail() <= 0) {
+      return block;
+    }
```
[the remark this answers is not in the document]
bneradt: Fixed. _read() now treats a null write-side reader like no available data instead of dereferencing it.

src/iocore/net/quic/QUICStreamVCAdapter.cc
```diff
@@ -84,18 +86,40 @@ QUICStreamVCAdapter::_read(size_t len)
...
+  }
+
+  SCOPED_MUTEX_LOCK(lock, this->_write_vio.mutex, this_ethread());
+
+  IOBufferReader *reader      = this->_write_vio.get_reader();
+  const size_t    consume_len = std::min(len, static_cast<size_t>(std::max<int64_t>(reader->read_avail(), 0)));
+  if (consume_len > 0) {
+    reader->consume(consume_len);
+    this->_write_vio.ndone += consume_len;
+  }
```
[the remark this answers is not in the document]
bneradt: Fixed. _consume() now returns if the write-side reader is null, and I hardened unread_len() the same way.

tests/gold_tests/autest-site/conditions.test.ext
```diff
@@ -217,6 +231,29 @@ def HasProxyVerifierVersion(self, version):
     return self.EnsureVersion([verifier_path, "--version"], min_version=version)
 
 
+def HasGoVersion(self, version):
+    """Check whether the go command is available at the requested version."""
+
+    def version_tuple(value):
+        return tuple(int(part) for part in value.split('.'))
```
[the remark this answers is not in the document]
bneradt: Fixed via changes in #13213

src/iocore/net/UnixUDPNet.cc
```diff
@@ -666,14 +666,16 @@ UDPNetProcessorInternal::read_multiple_messages_from_net(UDPNetHandler *nh, UDPC
...
 }
 #endif
 
 void
 UDPNetProcessorInternal::udp_read_from_net(UDPNetHandler *nh, UDPConnection *xuc)
 {
 #if HAVE_RECVMMSG
-  read_multiple_messages_from_net(nh, xuc);
+  while (read_multiple_messages_from_net(nh, xuc)) {}
 #else
```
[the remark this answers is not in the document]
bneradt: Fixed in #13213 before it landed. udp_read_from_net() now limits each event to MAX_RECEIVE_MSG_BATCHES_PER_EVENT—currently eight batches—so the net thread cannot drain indefinitely. UnixUDPNet.cc is no longer changed by this PR.

src/proxy/http3/Http3App.cc
```diff
@@ -286,6 +303,10 @@ Http3App::_handle_uni_stream_on_read_ready(int /* event */, VIO *vio)
   default:
     break;
   }
+
+  if (error && error->cls != Http3ErrorClass::UNDEFINED) {
+    this->_handle_error(*error);
+  }
```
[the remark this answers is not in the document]
bneradt: Fixed. Unknown unidirectional HTTP/3 stream data is now discarded as it arrives, following RFC 9114’s discard option, so unread data cannot accumulate on repeated READ_READY events.

src/proxy/http3/Http3ProtocolEnforcer.cc
```diff
@@ -30,7 +30,7 @@ Http3ProtocolEnforcer::interests()
   return {Http3FrameType::DATA,         Http3FrameType::HEADERS,      Http3FrameType::X_RESERVED_1, Http3FrameType::CANCEL_PUSH,
           Http3FrameType::SETTINGS,     Http3FrameType::PUSH_PROMISE, Http3FrameType::X_RESERVED_2, Http3FrameType::GOAWAY,
           Http3FrameType::X_RESERVED_3, Http3FrameType::X_RESERVED_4, Http3FrameType::MAX_PUSH_ID,  Http3FrameType::X_MAX_DEFINED,
-          Http3FrameType::UNKNOWN};
+          Http3FrameType::RESERVED,     Http3FrameType::UNKNOWN};
```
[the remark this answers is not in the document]
bneradt: Fixed. Aliased frame interests are now de-duplicated during handler registration, and the redundant X_MAX_DEFINED entries were removed from the protocol enforcer and frame counter. A regression test verifies that a MAX_PUSH_ID frame is dispatched exactly once.

src/iocore/net/QUICMultiCertConfigLoader.cc
```diff
@@ -88,12 +92,256 @@ QUICCertConfig::release(SSLCertLookup *lookup)
...
+  }
+
+  *out    = nullptr;
+  *outlen = 0;
+  return SSL_TLSEXT_ERR_ALERT_FATAL;
+}
+} // end anonymous namespace
+
+bool
+QUICMultiCertConfigLoader::_set_alpn_callback(SSL_CTX *ctx)
```
maskit, line 312: I don't get why we need another callback function.

src/iocore/net/QUICMultiCertConfigLoader.cc
```diff
@@ -88,12 +92,256 @@ QUICCertConfig::release(SSLCertLookup *lookup)
...
+
+  return true;
+}
+#endif
+
+bool
+QUICMultiCertConfigLoader::_setup_session_cache(SSL_CTX *ctx)
+{
+#if TS_HAS_OPENSSL_QUIC
+  SSL_CTX_set_min_proto_version(ctx, TLS1_3_VERSION);
```
maskit: It's done in `quic_new_ssl_ctx`. Why we need to do this again here?

src/iocore/net/quic/QUICConfig.cc
```diff
@@ -54,6 +57,24 @@ quic_new_ssl_ctx()
   return ssl_ctx;
 }
 
+SSL_CTX *
+quic_new_server_ssl_ctx()
```
maskit, line 61: Why a new function? The code surrounded by ifdef could be placed in quic_new_ssl_ctx. And the version settings are done there.

src/iocore/net/QUICMultiCertConfigLoader.cc
bneradt: The existing ssl_alpn_select_callback retrieves ALPNSupport from the SSL object’s ex-data, which is bound when the QUICNetVConnection is initialized. With OpenSSL’s native QUIC listener, ALPN negotiation happens before SSL_accept_connection() returns, while ATS allocates and initializes the QUICNetVConnection only after that call succeeds, so the existing callback has no ALPNSupport available. This context-level callback therefore performs only the early HTTP/3 ALPN selection needed by the listener; after acceptance, the SSL object is attached to the NetVC and normal ATS protocol handling takes over.

src/iocore/net/QUICMultiCertConfigLoader.cc
bneradt: Good catch—the TLS 1.3 bounds were being applied twice. Both QUIC context constructors already establish TLS 1.3 as an invariant when creating their SSL_CTX, so _setup_session_cache() should not repeat that policy while configuring an unrelated feature. I removed the duplicate calls, leaving protocol-version configuration in the context constructors where it applies consistently regardless of session-cache behavior.

src/iocore/net/quic/QUICConfig.cc
bneradt: quic_new_ssl_ctx() cannot unconditionally use OSSL_QUIC_server_method() because it is also called by quic_init_client_ssl_ctx(). Native OpenSSL requires the server-specific method for the listener, while the client and non-native quiche paths must retain the existing TLS_method() context, so moving the #ifdef into the shared function would make that function return a server-only context to client callers in native builds. Keeping quic_new_server_ssl_ctx() makes the role distinction explicit; its non-native path delegates to quic_new_ssl_ctx(), and the duplicated TLS-version settings noted separately have now been removed.

src/iocore/net/QUICMultiCertConfigLoader.cc
maskit: I already dislike the OpenSSL API (the inconsistency between TCP and QUIC)... but now I understand the necessity. This is fine for now but we should revisit this later.

src/iocore/net/quic/QUICConfig.cc
maskit: I forgot we already have `quic_init_client_ssl_ctx`. I still feel like these functions could be reorganized but this is not a blocker.

maskit (approved):
Overall looks good. No blockers.

We probably should reorganize functions so both implementations sit together nicely, but I may want to add QMux (QUIC over TCP) support before doing it to have a better picture.

bneradt:
10.2.x version:
https://github.com/apache/trafficserver/pull/13463

cmcfarlen:
Already present on the 10.2.x branch for the 10.2.0 release: the cherry-pick is empty because these changes landed via #13463 (939915eeee8a5cf0e5ffb9e51def40603cddb317), the 10.2.x adaptation of this work. The files it adds are byte-identical to master on that branch.

PR #13188, opened 2026-05-21
Add TSMutex lock guard

bneradt:
Plugin code that protects small critical sections with TSMutex has to pair
every early return with a matching unlock. That pattern is easy to get wrong
and makes the intended lock lifetime harder to see.

This adds a small TSMutexLockGuard helper to the plugin API and uses it in
plugin code where the mutex naturally stays locked until a return path.

### DIFF
--- a/include/ts/ts.h
+++ b/include/ts/ts.h
@@ -1168,6 +1168,26 @@ TSReturnCode TSMutexLockTry(TSMutex mutexp);
 
 void TSMutexUnlock(TSMutex mutexp);
 
+/** Scoped lock guard for a @c TSMutex.
+
+    Locks @a mutexp on construction and unlocks it when the guard leaves scope.
+ */
+class [[nodiscard]] TSMutexLockGuard
+{
+public:
+  explicit TSMutexLockGuard(TSMutex mutexp) : m_mutex(mutexp) { TSMutexLock(m_mutex); }
+
+  TSMutexLockGuard(const TSMutexLockGuard &)            = delete;
+  TSMutexLockGuard &operator=(const TSMutexLockGuard &) = delete;
+  TSMutexLockGuard(TSMutexLockGuard &&)                 = delete;
+  TSMutexLockGuard &operator=(TSMutexLockGuard &&)      = delete;
+
+  ~TSMutexLockGuard() { TSMutexUnlock(m_mutex); }
+
+private:
+  TSMutex m_mutex = nullptr;
+};
+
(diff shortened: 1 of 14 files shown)


include/ts/ts.h
```diff
@@ -1168,6 +1168,26 @@ TSReturnCode TSMutexLockTry(TSMutex mutexp);
...
 void TSMutexUnlock(TSMutex mutexp);
 
+/** Scoped lock guard for a @c TSMutex.
+
+    Locks @a mutexp on construction and unlocks it when the guard leaves scope.
+ */
+class TSMutexLockGuard
+{
+public:
+  explicit TSMutexLockGuard(TSMutex mutexp) : m_mutex(mutexp) { TSMutexLock(m_mutex); }
```
zwoop:
```
[[nodiscard("Did you forget to name your lock guard variable?")]]
explicit TSMutexLockGuard(TSMutex mutexp) : m_mutex(mutexp) { TSMutexLock(m_mutex); }
```

or some such ? This prevents a user from accidentally making a no-op lock guard (which almost certainly is not what they want). Albeit nitpicky, it feels useful in a public (plugin) API.

zwoop (dismissed):

include/ts/ts.h
bneradt: Good idea, thanks. I added [[nodiscard]] to the constructor. I left off the
custom message because ts.h still permits C++17 plugin builds, and Clang treats
the string-literal form as a C++20 extension in strict C++17 mode.

include/ts/ts.h
```diff
@@ -1168,6 +1168,26 @@ TSReturnCode TSMutexLockTry(TSMutex mutexp);
...
 void TSMutexUnlock(TSMutex mutexp);
 
+/** Scoped lock guard for a @c TSMutex.
+
+    Locks @a mutexp on construction and unlocks it when the guard leaves scope.
+ */
+class TSMutexLockGuard
+{
+public:
+  [[nodiscard]] explicit TSMutexLockGuard(TSMutex mutexp) : m_mutex(mutexp) { TSMutexLock(m_mutex); }
```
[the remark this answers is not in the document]
bneradt: Applied. [[nodiscard]] is now on the TSMutexLockGuard class declaration, so the public header uses the C++17-supported placement while keeping the guard constructor explicit.

masaori335 (approved):
Looks good to me. No leak of lock anymore.

PR #13192, opened 2026-05-22
yaml-cpp-0.9.0

bneradt:
Addresses the following g++ 16 build issue:

```
[8/1560] Building CXX object lib/yamlcpp/CMakeFiles/yaml-cpp.dir/src/emitterutils.cpp.o
FAILED: [code=1] lib/yamlcpp/CMakeFiles/yaml-cpp.dir/src/emitterutils.cpp.o
/usr/sbin/ccache /usr/lib64/ccache/c++ -DDEBUG -DOPENSSL_API_COMPAT=10002 -DOPENSSL_IS_OPENSSL3 -DPACKAGE_NAME="\"Apache Traffic Server\"" -DPACKAGE_VERSION=\"11.0.0\" -D_DEBUG -Dlinux -Dyaml_cpp_EXPORTS -I/home/bneradt/shared/ts_asf/include -I/home/bneradt/shared/ts_asf/build/include -I/home/bneradt/shared/ts_asf/lib/yamlcpp/include -I/home/bneradt/shared/ts_asf/lib/yamlcpp/src -pthread -g -std=c++20 -fPIC -fdiagnostics-color=always -DENABLE_SYSTEMTAP_PROBES -Wno-invalid-offsetof -Wno-shadow -Werror -MD -MT lib/yamlcpp/CMakeFiles/yaml-cpp.dir/src/emitterutils.cpp.o -MF lib/yamlcpp/CMakeFiles/yaml-cpp.dir/src/emitterutils.cpp.o.d -o lib/yamlcpp/CMakeFiles/yaml-cpp.dir/src/emitterutils.cpp.o -c /home/bneradt/shared/ts_asf/lib/yamlcpp/src/emitterutils.cpp
/home/bneradt/shared/ts_asf/lib/yamlcpp/src/emitterutils.cpp:221:11: error: ‘uint16_t’ was not declared in this scope
  221 | std::pair<uint16_t, uint16_t> EncodeUTF16SurrogatePair(int codePoint) {
      |           ^~~~~~~~
/home/bneradt/shared/ts_asf/lib/yamlcpp/src/emitterutils.cpp:13:1: note: ‘uint16_t’ is defined in header ‘<cstdint>’; this is probably fixable by adding ‘#include <cstdint>’
   12 | #include "yaml-cpp/null.h"
  +++ |+#include <cstdint>
   13 | #include "yaml-cpp/ostream_wrapper.h"
[...]

PR #13193, opened 2026-05-22
GCC 16: Regex header integer includes

bneradt:
GCC 16 no longer exposes fixed-width integer types through the current transitive include path. Include cstdint directly so Regex.h remains self-contained.

### DIFF
--- a/include/tsutil/Regex.h
+++ b/include/tsutil/Regex.h
@@ -23,10 +23,11 @@
 
 #pragma once
 
+#include <cstdint>
+#include <memory>
 #include <string_view>
 #include <string>
 #include <vector>
-#include <memory>
 
 /// @brief Match flags for regular expression evaluation.
 ///


bryancall (approved):
Verified on Fedora 44 / GCC 16.1.1 (eris):

- Fresh clone of master at `09c4f8a` (current tip), `gh pr checkout 13193`, configured with the `dev` preset (`-Werror` active on all 995 compile units), built with `ninja -j32 -k 0`.
- `Regex.h`, `Regex.cc`, and `test_Regex.cc` all compile clean with zero warnings.
- The fix is correct: GCC 16's libstdc++ no longer transitively exposes the fixed-width integer types, so the header has to include `<cstdint>` directly.
- Build progresses through 1054/1167 targets before stopping on the unrelated `lib/yamlcpp/src/emitterutils.cpp` break that is owned by PR #13192. With #13192 also applied, the full build completes clean in ~46s (separately verified).

LGTM. CI is also fully green other than one AuTest 0of4 flake (3 of 4 shards passed and the change is a header-only `#include` addition; can't plausibly cause a runtime test failure.
EOF
)

bryancall:
[approve ci autest 0]

cmcfarlen:
Cherry-picked to 10.2.x

cmcfarlen:

PR #13197, opened 2026-05-23
Update cert compression reporting

bneradt:
Some OpenSSL builds expose certificate-compression preference APIs while
disabling the built-in compression algorithms. In that shape, ATS reported
cert compression support based on API availability even though configured
algorithms could not actually run.

This reports certificate-compression support from the algorithms ATS can use
in the selected OpenSSL API path, and skips the callback-only AuTest unless
callbacks are available. This also makes the supported algorithm table
contain only usable entries, so configuration fails cleanly when an
unavailable algorithm is requested.

---

This is required for our migration to the fedora:44 image since that has updated build tools that exposes this issue.

### DIFF
--- a/src/iocore/net/TLSCertCompression.cc
+++ b/src/iocore/net/TLSCertCompression.cc
@@ -31,7 +31,9 @@ namespace
 DbgCtl dbg_ctl_ssl_cert_compress{"ssl_cert_compress"};
 }
 
+#if HAVE_SSL_CTX_ADD_CERT_COMPRESSION_ALG || HAVE_SSL_CTX_SET1_CERT_COMP_PREFERENCE
 constexpr unsigned int N_ALGORITHMS = 3;
+#endif
 
 #if HAVE_SSL_CTX_ADD_CERT_COMPRESSION_ALG
 #include "TLSCertCompression_zlib.h"
@@ -45,6 +47,9 @@ constexpr unsigned int N_ALGORITHMS = 3;
 #endif
 #endif
 
+#if HAVE_SSL_CTX_ADD_CERT_COMPRESSION_ALG || HAVE_SSL_CTX_SET1_CERT_COMP_PREFERENCE
+namespace
+{
 struct alg_info {
   const char *name;
   int32_t     number;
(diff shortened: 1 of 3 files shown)


src/traffic_layout/info.cc
```diff
@@ -53,6 +54,34 @@
...
+static constexpr int ts_has_cert_compression_brotli = 1;
+#else
+static constexpr int ts_has_cert_compression_brotli = 0;
+#endif
+
+#if (HAVE_SSL_CTX_ADD_CERT_COMPRESSION_ALG && HAVE_ZSTD_H) || (HAVE_SSL_CTX_SET1_CERT_COMP_PREFERENCE && !defined(OPENSSL_NO_ZSTD))
+static constexpr int ts_has_cert_compression_zstd = 1;
+#else
+static constexpr int ts_has_cert_compression_zstd = 0;
+#endif
```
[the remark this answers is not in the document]
bneradt: Callback-capable builds now report the algorithms ATS can provide locally, and preference-only builds fall back to OpenSSL’s built-in algorithm availability.

src/iocore/net/TLSCertCompression.cc
```diff
@@ -107,16 +121,24 @@ register_certificate_compression_preference(SSL_CTX *ctx, const std::vector<std:
...
       if (strcmp(specified_algs[i].c_str(), supported_algs[j].name) == 0) {
-        algs[n++] = supported_algs[j].number;
-        Dbg(dbg_ctl_ssl_cert_compress, "Enabled %s", supported_algs[j].name);
+        info = &supported_algs[j];
+        break;
       }
     }
+    if (info == nullptr || !info->available) {
+      Dbg(dbg_ctl_ssl_cert_compress, "Unrecognized algorithm: %s", specified_algs[i].c_str());
+      return 0;
```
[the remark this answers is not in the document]
bneradt: Unknown algorithm names still log as unrecognized, while recognized-but-unavailable algorithms now log that the algorithm was disabled by the OpenSSL build.

src/iocore/net/TLSCertCompression.cc
```diff
@@ -95,28 +109,43 @@ register_certificate_compression_preference(SSL_CTX *ctx, const std::vector<std:
...
-      if (SSL_CTX_add_cert_compression_alg(ctx, info->number, info->compress_func, info->decompress_func) == 0) {
-        return 0;
-      }
-      Dbg(dbg_ctl_ssl_cert_compress, "Enabled %s", info->name);
-    } else {
+    if (info == nullptr) {
       Dbg(dbg_ctl_ssl_cert_compress, "Unrecognized algorithm: %s", alg.c_str());
       return 0;
     }
+    if (!info->available) {
```
maskit: This line is gated by `HAVE_SSL_CTX_ADD_CERT_COMPRESSION_ALG`, so this condition is always false, unnecessary.

src/iocore/net/TLSCertCompression.cc
```diff
@@ -48,25 +46,41 @@ constexpr unsigned int N_ALGORITHMS = 3;
...
   const char *name;
   int32_t     number;
+  bool        available;
 #if HAVE_SSL_CTX_ADD_CERT_COMPRESSION_ALG
   ssl_cert_compression_func_t   compress_func;
   ssl_cert_decompression_func_t decompress_func;
 #endif
 } supported_algs[] = {
   {"zlib",   1,
+#if HAVE_SSL_CTX_ADD_CERT_COMPRESSION_ALG || !defined(OPENSSL_NO_ZLIB)
```
maskit: Can't we surround the entire entry `HAVE_SSL_CTX_ADD_CERT_COMPRESSION_ALG || !(defined(OPENSSL_NO_ZLIB)` instead of adding `available` flag?

maskit (changes requested):
Looks like the code is overcomplicated. I don't see the necessity of `available` flag.

src/iocore/net/TLSCertCompression.cc
bneradt: Sounds good. Removed the `available` flag.

maskit:
I removed this from 10.1 and 10.2 projects because the initial commit is for certain compression is only on master.

src/iocore/net/TLSCertCompression.cc
```diff
@@ -53,70 +55,81 @@ struct alg_info {
...
 #endif
 #if HAVE_ZSTD_H
-  {"zstd",   3,
-#if HAVE_SSL_CTX_ADD_CERT_COMPRESSION_ALG
-   compression_func_zstd,   decompression_func_zstd
+  {"zstd",   3, compression_func_zstd,   decompression_func_zstd  },
+#endif
+  {nullptr,  0, nullptr,                 nullptr                  },
+#else
+#if HAVE_SSL_CTX_SET1_CERT_COMP_PREFERENCE && !defined(OPENSSL_NO_ZLIB)
```
maskit: `HAVE_SSL_CTX_SET1_CERT_COMP_PREFERENCE` is repeated. We could have it as `#elif HAVE_SSL_CTX_ADD_CERT_COMPRESSION_ALG`, and `#else` would be just `{nullptr, 0}`.

bneradt:
> I removed this from 10.1 and 10.2 projects because the initial commit is for certain compression is only on master.

Oh, thanks. I didn't realize that was master-only.

src/iocore/net/TLSCertCompression.cc
bneradt: Good point.

src/iocore/net/TLSCertCompression.cc
```diff
@@ -53,70 +55,81 @@ struct alg_info {
...
 
+alg_info const *
+find_algorithm(std::string const &name)
+{
+  for (auto const &alg : supported_algs) {
+    if (alg.name != nullptr && name == alg.name) {
+      return &alg;
+    }
+  }
+  return nullptr;
```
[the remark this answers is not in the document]
bneradt: Done. I moved the helper/table into an anonymous namespace so find_algorithm() has internal linkage.

maskit (approved):

PR #13200, opened 2026-05-26
Single source of truth for Proxy Verifier metadata

bneradt:
#13200 updated the Proxy Verifier versions and one of the checksum
references correctly, but missed the CMakeLists checksum. Currently, the
version and checksums for the Proxy Verifier release are referenced in
multiple places, making it easy to miss updating the various required
locations. This is bad.

This patch adds a checksum companion file next to the version file and
teaches all the consumers to read the shared version/checksum sources
when downloading Proxy Verifier.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -676,8 +676,11 @@ if(ENABLE_AUTEST)
   # the autest command. The original AUTEST_OPTIONS string is used in the
   # autest.sh script.
   separate_arguments(AUTEST_OPTIONS_LIST UNIX_COMMAND "${AUTEST_OPTIONS}")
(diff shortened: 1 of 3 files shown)


masaori335 (approved):
Looks good

grahamsedman:
Hi Brian,

Centralizing the Proxy Verifier version and checksums into a single source of truth eliminates the error-prone manual updates across multiple files. This change will prevent future mismatches like the missed CMakeLists checksum and makes the release process more robust. Great improvement for the projects maintainability.

I'm glad that I found the reason why the pull requests was stuck at the AuTests stage see a few more greens.

cmcfarlen:
Cherry-picked to 10.2.x

cmcfarlen:

PR #13202, opened 2026-05-26
Reduce TLS write-path overhead

zwoop:
Coalesce a response's plaintext across IOBuffer blocks into a per-thread staging buffer and issue a single `SSL_write` per chunk (bounded to one max TLS record), instead of one `SSL_write` per block. For responses fragmented across many small blocks this cuts the number of `SSL_write` calls, TLS records, and `write()` syscalls; contiguous blocks still write in place with no copy.

Portable across OpenSSL and BoringSSL (no filter BIO), and preserves dynamic record sizing and the partial-write / `WANT_WRITE` retry path.

### DIFF
--- a/src/iocore/net/SSLNetVConnection.cc
+++ b/src/iocore/net/SSLNetVConnection.cc
@@ -747,11 +747,9 @@ SSLNetVConnection::load_buffer_and_write(int64_t towrite, MIOBufferAccessor &buf
   do {
     IOBufferReader *reader = buf.reader();
 
-    // Unlike the per-block original, l may span blocks so it can be coalesced below.
-    int64_t avail  = reader->read_avail();
-    int64_t wavail = towrite - total_written;
-
-    l = (wavail < avail) ? wavail : avail;
+    // towrite is already bounded by read_avail() in the caller, so this never exceeds
+    // what's queued -- avoid re-walking the block chain (O(n)) on the write hot path.
+    l = towrite - total_written;
 
     // TS-2365: If the SSL max record size is set and we have
     // more data than that, break this into smaller write
--- a/src/iocore/net/SSLNetVConnection.cc
+++ b/src/iocore/net/SSLNetVConnection.cc
@@ -739,16 +739,17 @@ SSLNetVConnection::load_buffer_and_write(int64_t towrite, MIOBufferAccessor &buf
 
   Dbg(dbg_ctl_ssl, "towrite=%" PRId64, towrite);
 
-  // Per-thread scratch to coalesce fragmented blocks into one SSL_write. Reuse across
-  // connections is safe: a WANT_WRITE retry flushes SSL's own record buffer, not this.
+  // SSL_write retries (WANT_WRITE/READ) must reuse the same address; thread_local gives the
+  // coalesce path a stable one. SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER is off.
   static thread_local char gather_buf[SSL_MAX_TLS_RECORD_SIZE];
+
+  // Caller bounds towrite by read_avail(); asserting lets the loop skip an O(n) chain walk.
+  ink_assert(towrite <= buf.reader()->read_avail());
 
   ERR_clear_error();
   do {
     IOBufferReader *reader = buf.reader();
 
-    // towrite is already bounded by read_avail() in the caller, so this never exceeds
-    // what's queued -- avoid re-walking the block chain (O(n)) on the write hot path.
     l = towrite - total_written;
 
     // TS-2365: If the SSL max record size is set and we have
@@ -781,8 +782,7 @@ SSLNetVConnection::load_buffer_and_write(int64_t towrite, MIOBufferAccessor &buf
       break;
     }
 
-    // Coalesce into the scratch buffer only when the chunk spans blocks and fits one
-    // record; otherwise write in place capped to the block.
+    // Coalesce across blocks only when it fits one record; else write in place, capped to the block.
     const char *write_block;
     int64_t     block_avail = reader->block_read_avail();
 
--- a/src/iocore/net/SSLNetVConnection.cc
+++ b/src/iocore/net/SSLNetVConnection.cc
@@ -739,18 +739,19 @@ SSLNetVConnection::load_buffer_and_write(int64_t towrite, MIOBufferAccessor &buf
 
   Dbg(dbg_ctl_ssl, "towrite=%" PRId64, towrite);
 
+  // Per-thread scratch to coalesce fragmented blocks into one SSL_write. Reuse across
+  // connections is safe: a WANT_WRITE retry flushes SSL's own record buffer, not this.
+  static thread_local char gather_buf[SSL_MAX_TLS_RECORD_SIZE];
+
   ERR_clear_error();
   do {
-    // What is remaining left in the next block?
-    l                   = buf.reader()->block_read_avail();
-    char *current_block = buf.reader()->start();
+    IOBufferReader *reader = buf.reader();
 
-    // check if to amount to write exceeds that in this buffer
+    // Unlike the per-block original, l may span blocks so it can be coalesced below.
+    int64_t avail  = reader->read_avail();
     int64_t wavail = towrite - total_written;
 
-    if (l > wavail) {
-      l = wavail;
-    }
+    l = (wavail < avail) ? wavail : avail;
 
     // TS-2365: If the SSL max record size is set and we have
     // more data than that, break this into smaller write
@@ -782,15 +783,30 @@ SSLNetVConnection::load_buffer_and_write(int64_t towrite, MIOBufferAccessor &buf
       break;
     }
 
+    // Coalesce into the scratch buffer only when the chunk spans blocks and fits one
+    // record; otherwise write in place capped to the block.
+    const char *write_block;
+    int64_t     block_avail = reader->block_read_avail();
+
+    if (block_avail < l && l <= static_cast<int64_t>(sizeof(gather_buf))) {
+      reader->memcpy(gather_buf, l, 0);
+      write_block = gather_buf;
+    } else {
+      if (l > block_avail) {
+        l = block_avail;
+      }
+      write_block = reader->start();
+    }
+
     try_to_write       = l;
     num_really_written = 0;
-    Dbg(dbg_ctl_v_ssl, "b=%p l=%" PRId64, current_block, l);
-    err = this->_ssl_write_buffer(current_block, l, num_really_written);
+    Dbg(dbg_ctl_v_ssl, "b=%p l=%" PRId64, write_block, l);
+    err = this->_ssl_write_buffer(write_block, l, num_really_written);
 
     // We wrote all that we thought we should
     if (num_really_written > 0) {
       total_written += num_really_written;
-      buf.reader()->consume(num_really_written);
+      reader->consume(num_really_written);
     }
 
     Dbg(dbg_ctl_ssl, "try_to_write=%" PRId64 " written=%" PRId64 " total_written=%" PRId64, try_to_write, num_really_written,


src/iocore/net/TLSBasicSupport.cc
```diff
@@ -52,6 +52,9 @@ TLSBasicSupport::initialize()
 TLSBasicSupport *
 TLSBasicSupport::getInstance(SSL *ssl)
 {
+  if (auto *exd = SSLExDataGet(ssl)) {
+    return exd->basic;
+  }
   return static_cast<TLSBasicSupport *>(SSL_get_ex_data(ssl, _ex_data_index));
```
maskit, line 55: Why do we need this fallback?

src/iocore/net/SSLNetVConnection.cc
```diff
@@ -111,47 +111,53 @@ SSLNetVConnection::_make_ssl_connection(SSL_CTX *ctx)
...
+  this->_ssl_ex_data.basic       = static_cast<TLSBasicSupport *>(this);
+  this->_ssl_ex_data.event       = static_cast<TLSEventSupport *>(this);
+  this->_ssl_ex_data.alpn        = static_cast<ALPNSupport *>(this);
+  this->_ssl_ex_data.resumption  = static_cast<TLSSessionResumptionSupport *>(this);
+  this->_ssl_ex_data.sni         = static_cast<TLSSNISupport *>(this);
+  this->_ssl_ex_data.early_data  = static_cast<TLSEarlyDataSupport *>(this);
+  this->_ssl_ex_data.tunnel      = static_cast<TLSTunnelSupport *>(this);
+  this->_ssl_ex_data.cert_switch = static_cast<TLSCertSwitchSupport *>(this);
+  SSLNetVCAttach(this->ssl, &this->_ssl_ex_data);
   TLSEventSupport::bind(this->ssl, this);
```
maskit, line 145: Why do we need to keep this one?

maskit (dismissed):
First of all, doing multiple things on one PR is a bad idea. It's difficult to review 👎 

What's the motivation for the consolidated ExData? The initialize functions still register an index for each,  the number of `SSL_get_ex_data` calls is unchanged since every `getInstance` calls it (getting the same ExData object again and again).

I didn't look at the other changes.

zwoop:
@maskit I removed the SSL_get_ex_data stuff, it does reduce the number of reallocations, but it's pretty marginal impact on performance (maybe 1 microsecond / connection).

The performance on the BIO buffer is real though, I benchmarked it again, and still get a 15% performance improvement:

  ## Summary

  BIO write buffering collapses 2 writes into 1 even for tiny (~317 B) responses, yielding ~26% more throughput and ~20–35% lower latency across the distribution. Both runs are stable (CV ≤ 0.2%, n=3) with non-overlapping 95%
  CIs, so the gain is real.

  | Metric | Before | With BIO buffering | Δ |
  |--------|-------:|-------------------:|:--|
  | Req/sec (median) | 541,518 | 681,681 | **+25.9%** |
  | Throughput | 171.7 MB/s | 216.1 MB/s | **+25.9%** |
  | p50 latency | 2.15 ms | 1.71 ms | **−20.5%** |
  | p95 latency | 5.74 ms | 4.17 ms | **−27.4%** |
  | p99 latency | 8.31 ms | 5.42 ms | **−34.8%** |
  | Mean latency | 2.76 ms | 2.19 ms | −20.7% |

## ATS master

| Run | Req/sec | Bytes/sec | Reqs (200 OK) | Total | Avg Size | Errors | p50 (ms) | Duration |
|-----|---------|-----------|---------------|-------|----------|--------|----------|----------|
| 1 | 541102 | 171.5 MB/s | 24.376M | 24.376M | 317 B | 0.00% | 2.13 | 45.0s |
| 2 | 542248 | 171.9 MB/s | 24.423M | 24.423M | 317 B | 0.00% | 2.14 | 45.0s |
| 3 | 541518 | 171.7 MB/s | 24.389M | 24.389M | 317 B | 0.00% | 2.15 | 45.0s |
| | | | | | | | | |
| **Median (x̃)** | 541518 | 171.7 MB/s | 24.389M | 24.389M | — | — | — | — |
| **Mean (μ)** | 541623 | 171.7 MB/s | 24.396M | 24.396M | — | — | — | — |

**RPS Statistics** (n=3): x̃=541518, μ=541623, σ=580, CV=0.1%, range=[541102–542248]
**95% CI**: [540182–543063] req/s *(wide interval — increase `runs` for tighter bounds)*

### Timing (ms)

| Metric | Latency | TTFB | Connect |
|--------|--------:|--------:|--------:|
| Min | 0.373 | 0.366 | 0.350 |
| μ (Mean) | 2.76 | 2.76 | 5.38 |
| σ (StdDev) | 1.51 | 1.51 | 9.74 |
| p50 | 2.15 | 2.14 | 2.83 |
| p95 | 5.74 | 5.74 | 27.25 |
| p99 | 8.31 | 8.31 | 54.44 |
| Max | 132.5 | 132.5 | 62.69 |
| Skew (γ₁) | 4.55 | 4.57 | 3.86 |
| Samples | 1.524M | 1.524M | 1.5K |

### Latency Distribution

```
  p50 │█ 2.15
  p95 │█ 5.74
  p99 │██ 8.31
  min=0.373  μ=2.76  max=132.5
```

## With BIO buffering

| Run | Req/sec | Bytes/sec | Reqs (200 OK) | Total | Avg Size | Errors | p50 (ms) | Duration |
|-----|---------|-----------|---------------|-------|----------|--------|----------|----------|
| 1 ⚠ | 679041 | 215.3 MB/s | 30.594M | 30.594M | 317 B | 0.00% | 1.71 | 45.1s |
| 2 | 681681 | 216.1 MB/s | 30.702M | 30.702M | 317 B | 0.00% | 1.70 | 45.0s |
| 3 | 682115 | 216.2 MB/s | 30.719M | 30.719M | 317 B | 0.00% | 1.71 | 45.0s |
| | | | | | | | | |
| **Median (x̃)** | 681681 | 216.1 MB/s | 30.702M | 30.702M | — | — | — | — |
| **Mean (μ)** | 680946 | 215.9 MB/s | 30.671M | 30.671M | — | — | — | — |

**RPS Statistics** (n=3): x̃=681681, μ=680946, σ=1664, CV=0.2%, range=[679041–682115]
**95% CI**: [676813–685078] req/s *(wide interval — increase `runs` for tighter bounds)*

⚠ **Outlier detected**: Run 1 (modified Z-score |Z| > 2.5)

### Timing (ms)

| Metric | Latency | TTFB | Connect |
|--------|--------:|--------:|--------:|
| Min | 0.307 | 0.307 | 0.423 |
| μ (Mean) | 2.19 | 2.19 | 6.09 |
| σ (StdDev) | 0.995 | 0.995 | 10.87 |
| p50 | 1.71 | 1.71 | 3.04 |
| p95 | 4.17 | 4.17 | 35.07 |
| p99 | 5.42 | 5.42 | 56.97 |
| Max | 88.34 | 88.34 | 62.11 |
| Skew (γ₁) | 3.69 | 3.69 | 3.37 |
| Samples | 1.920M | 1.920M | 1.5K |

### Latency Distribution

```
  p50 │█ 1.71
  p95 │█ 4.17
  p99 │██ 5.42
  min=0.307  μ=2.19  max=88.34
```

src/iocore/net/SSLNetVConnection.cc
```diff
@@ -123,11 +123,26 @@ SSLNetVConnection::_make_ssl_connection(SSL_CTX *ctx)
...
+        BIO *wbio = BIO_new_socket(this->get_socket(), BIO_NOCLOSE);
+        BIO_set_mem_eof_return(wbio, -1);
+        SSL_set_bio(ssl, rbio, wbio);
+      } else {
+        // No blind-tunnel: direct socket BIO, skip handshake buffers. On OpenSSL a
+        // write-side buffer BIO coalesces TLS records into fewer write() syscalls;
+        // BoringSSL has no filter BIOs, so it uses the plain socket BIO.
+        BIO *rbio = BIO_new_socket(this->get_socket(), BIO_NOCLOSE);
+        BIO *wbio = BIO_new_socket(this->get_socket(), BIO_NOCLOSE);
+#if !defined(OPENSSL_IS_BORINGSSL)
```
maskit: Not a blocker but I'd gate it by the availability of `BIO_f_buffer`. This would cause a build error if ATS is built with AWS-LC.

moonchen (changes requested):
I like the changes here, but I think this probably should be two PRs.  One being the flush changes, and the other one being the changes to buffering for the non-blind-tunnel use case.  This would make it easier to attribute the performance gains, and bisect if needed.

zwoop:
@moonchen I redid the entire PR, because the filter BIO APIs aren't available in Boring. This simpler version is still a significant improvement.

src/iocore/net/SSLNetVConnection.cc
```diff
@@ -737,18 +737,19 @@ SSLNetVConnection::load_buffer_and_write(int64_t towrite, MIOBufferAccessor &buf
...
 
-    // check if to amount to write exceeds that in this buffer
+    // Unlike the per-block original, l may span blocks so it can be coalesced below.
+    int64_t avail  = reader->read_avail();
     int64_t wavail = towrite - total_written;
 
-    if (l > wavail) {
-      l = wavail;
-    }
+    l = (wavail < avail) ? wavail : avail;
```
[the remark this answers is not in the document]
zwoop: Good catch — fixed. `towrite` is already bounded by `read_avail()` in the caller (write_to_net_io), so the per-iteration `read_avail()` walk was redundant; now using `towrite - total_written` directly with no O(n) block-chain walk.

src/iocore/net/SSLNetVConnection.cc
```diff
@@ -737,18 +737,19 @@ SSLNetVConnection::load_buffer_and_write(int64_t towrite, MIOBufferAccessor &buf
 
   Dbg(dbg_ctl_ssl, "towrite=%" PRId64, towrite);
 
+  // Per-thread scratch to coalesce fragmented blocks into one SSL_write. Reuse across
+  // connections is safe: a WANT_WRITE retry flushes SSL's own record buffer, not this.
+  static thread_local char gather_buf[SSL_MAX_TLS_RECORD_SIZE];
```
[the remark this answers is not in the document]
zwoop: Updated the PR description to match the implementation — this is a userspace staging-buffer gather (coalesce plaintext, one SSL_write per chunk), not the buffer-BIO approach the original description described.

zwoop:
Re-ran the tests again, and this time, for small objects, this patch was 25% more RPS. :)

zwoop:
[approve ci]

maskit (commented):
Looks good to me. I'll leave the approval to Mo.

src/iocore/net/SSLNetVConnection.cc
```diff
@@ -737,18 +737,17 @@ SSLNetVConnection::load_buffer_and_write(int64_t towrite, MIOBufferAccessor &buf
...
-    char *current_block = buf.reader()->start();
-
-    // check if to amount to write exceeds that in this buffer
-    int64_t wavail = towrite - total_written;
+    IOBufferReader *reader = buf.reader();
 
-    if (l > wavail) {
-      l = wavail;
-    }
+    // towrite is already bounded by read_avail() in the caller, so this never exceeds
```
JosiahWI: Could you (debug) assert this at the top of the function to enforce the precondition?

JosiahWI (changes requested):
The new code is very well commented and the logic makes sense. I have one request.

zwoop:
[approve ci]

src/iocore/net/SSLNetVConnection.cc
```diff
@@ -737,18 +737,17 @@ SSLNetVConnection::load_buffer_and_write(int64_t towrite, MIOBufferAccessor &buf
 
   Dbg(dbg_ctl_ssl, "towrite=%" PRId64, towrite);
 
+  // Per-thread scratch to coalesce fragmented blocks into one SSL_write. Reuse across
+  // connections is safe: a WANT_WRITE retry flushes SSL's own record buffer, not this.
```
moonchen: I think we should rephrase why the gather_buf is thread_local.  Isn't it this way because we don't set `SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER`?  Maybe something like:

```
// SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER is off, so SSL_write retries after WANT_WRITE/WANT_READ 
// require the same buffer address.  We ensure the same buffer address between retries by reusing a
// thread_local buffer.
```

moonchen:
One comment suggestion, but otherwise LGTM.

src/iocore/net/SSLNetVConnection.cc
zwoop: Done — rephrased the comment as suggested.

src/iocore/net/SSLNetVConnection.cc
zwoop: Done — added `ink_assert(towrite <= buf.reader()->read_avail())` at the top of the function.

zwoop:
Addressed the review comments

moonchen (dismissed):

JosiahWI (dismissed):
Looks good.

bneradt (approved):

bryancall (dismissed):
_(duplicate — disregard; see my other approval)_

bryancall (approved):
Approve. Coalescing multi-block chunks into the thread-local `gather_buf` before `SSL_write` is correct: overflow gated by `l <= sizeof(gather_buf)`, oversized chunks fall back in-place, and the moving-write-buffer retry holds since the buffer is `static thread_local` and `redoWriteSize` pins the length. Real ~25% RPS win on small objects, CI green, two approvals already.

moonchen (approved):

c-taylor:
Should this be 10.2 tagged?

cmcfarlen:
Cherry-picked to 10.2.x

PR #13204, opened 2026-05-27
hrw4u: Fix u4wrh emitting HTTP-section operators into TXN_CLOSE

Clendenin:
## Summary
u4wrh bypassed section access validation when emitting statements, so operators
that declare `sections=HTTP_SECTIONS` could land inside a generated `TXN_CLOSE`
block. The output then failed re-parse in hrw4u.

This change validates the operator's declared sections at emit time and surfaces
a clear error when the originating HRW places it in an incompatible hook.

### DIFF
--- a/tools/hrw4u/src/hrw_symbols.py
+++ b/tools/hrw4u/src/hrw_symbols.py
@@ -183,8 +183,7 @@ class InverseSymbolResolver(SymbolResolverBase):
 
                 if tag == "HEADER":
                     return f"{self.get_prefix_for_context('header_condition', section)}{suffix}", False
-                else:
-                    return f"{lhs_prefix}{suffix.replace(':', '.')}", False
+                return f"{lhs_prefix}{suffix.replace(':', '.')}", False
         return None
 
     def _should_lowercase_suffix(self, tag_match: str, lhs_prefix: str) -> bool:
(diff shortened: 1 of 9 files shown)


zwoop:
Not for this PR, but while reviewing, I filed two more issues. @Clendenin do you want to take a look at these two as well ?

https://github.com/apache/trafficserver/issues/13618
https://github.com/apache/trafficserver/issues/13619

tools/hrw4u/src/hrw_symbols.py
zwoop, line 478: I think there's a use case here as well which slips through:

```
cond %{TXN_CLOSE_HOOK}
    rm-destination QUERY
```

This produces

```
TXN_CLOSE {
    inbound.url.query = "";
}
```

Which is not legal:

```
leif@hel ~/apache/reviews/tools/hrw4u ±master⚡ » uv run ./scripts/u4wrh /tmp/bar.hrw|  uv run ./scripts/hrw4u
cond %{TXN_CLOSE_HOOK} [AND]
<stdin>:2:4: error: inbound.url.query is not available in the TXN_CLOSE section
   2 |     inbound.url.query = "";
     |     ^
```

zwoop (approved):
If you don't mind, can you address the "rm-destination" case as well ?

PR #13205, opened 2026-05-27
hrw4u: Fix u4wrh cidr() emitting 1-arg form missing IPv6 bits

Clendenin:
## Summary
HRW `%{CIDR:a}` supplies only IPv4 bits and relies on implicit IPv6 defaults,
while HRW4U `cidr()` requires both `ipv4_bits` and `ipv6_bits`. When u4wrh sees
the single value form, pad with the canonical HRW IPv6 default (48) so the
emitted HRW4U has valid arity and round-trips stably.

This change also extends the CIDR reverse branch to handle the bare `%{CIDR}`
no-payload form, padding to `cidr(24, 48)` using HRW's canonical defaults so
the forward round-trip satisfies the 2-arg signature.


zwoop:
[approve ci autest]

zwoop:
[approve ci autest 1]

zwoop:
Looking at this, I have some concerns about the fact that this changes behavior from how header_rewrite works. I think this is actually also a bug in header_rewrite itself, so I filed #13620.

zwoop (approved):

PR #13207, opened 2026-05-27
hrw4u: Fix u4wrh emitting quoted header names in HRW4U output

Clendenin:
## Summary
Header qualifiers that contain only characters legal in the HRW4U `IDENT` rule
were being wrapped in double quotes at emission time, which the grammar rejects
on re-parse. Quote a qualifier only when it actually contains characters that
require quoting.

Adds a parametrized unit test for `Validator.unquote_if_ident` covering
IDENT-safe stripping, space-containing qualifiers, empty strings, and
leading-digit rejection, symmetric with the existing `needs_quotes` and
`quote_if_needed` tests.

### DIFF
--- a/tools/hrw4u/src/common.py
+++ b/tools/hrw4u/src/common.py
@@ -50,6 +50,9 @@ class RegexPatterns:
         re.VERBOSE | re.DOTALL,
     )
 
+    # Grammar IDENT rule from hrw4u.g4 (header names, qualifiers)
+    GRAMMAR_IDENT: Final = re.compile(r'^[a-zA-Z_][a-zA-Z0-9_@.-]*$')
+
     # Additional performance patterns
     IDENTIFIER: Final = re.compile(r'^[a-zA-Z_][a-zA-Z0-9_]*$')
     WHITESPACE: Final = re.compile(r'\s+')
--- a/tools/hrw4u/src/hrw_symbols.py
+++ b/tools/hrw4u/src/hrw_symbols.py
@@ -233,6 +233,7 @@ class InverseSymbolResolver(SymbolResolverBase):
 
     def _handle_set_rm_operation(
             self, cmd: str, toks: list[str], prefix: str, qualifier: str, section: SectionType | None = None) -> str:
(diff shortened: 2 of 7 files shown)


zwoop:
While looking at this, these fixes actually unravel another weakness in the u4wrh code. Look at this example below, it'd be good to address this here, since the fixes in this PR opens up potentially really bad results in some odd cases:

```
leif@hel ~/apache/reviews/tools/hrw4u ±master⚡ » cat /tmp/foo.hrw
cond %{SEND_RESPONSE_HDR_HOOK}
    set-header "body" "v"
leif@hel ~/apache/reviews/tools/hrw4u ±master⚡ » uv run ./scripts/u4wrh /tmp/foo.hrw
SEND_RESPONSE {
    inbound.resp.body = "v";
}
leif@hel ~/apache/reviews/tools/hrw4u ±master⚡ » uv run ./scripts/u4wrh /tmp/foo.hrw | uv run ./scripts/hrw4u
cond %{SEND_RESPONSE_HDR_HOOK} [AND]
    set-body "v"
```

This is specific to the body header, but I think it'd be good to address this here, since the fixes makes it possible now to get into a really bad result.

zwoop:
I think there's a second issue, with @-headers, and potentially other headers, e.g.

```
leif@hel ~/apache/reviews/tools/hrw4u ±master⚡ » cat /tmp/at.hrw
    set-header "@Foo" "v"
leif@hel ~/apache/reviews/tools/hrw4u ±master⚡ » uv run ./scripts/u4wrh /tmp/at.hrw
REMAP {
    inbound.req."@Foo" = "v";
}
```

That should produce:

```
REMAP {
    inbound.req.@Foo = "v";
}
```

zwoop:
Here's a suggestion from Claude
[hrw4u-qualifier-guard.patch](https://github.com/user-attachments/files/31763497/hrw4u-qualifier-guard.patch)

But, please take a closer look of course. Also, this still reserves

inbound.resp.body

for 'body' (i.e. can't have that as a header).

zwoop (changes requested):
See inline comments.

PR #13208, opened 2026-05-27
hrw4u: Fix u4wrh set-config emitting 3-arg form with HRW type hint

Clendenin:
## Summary
Some legacy header_rewrite configs spell `set-config` with an explicit HRW type
token (`INT`, `STRING`, `FLOAT`) between name and value. The native ATS parser
has no concept of that token, and HRW4U `set-config` takes exactly two
arguments. When u4wrh encounters the three-argument form, drop the type token
so the emitted HRW4U call is `set-config(name, value)`.

Adds INT, STRING, and FLOAT reverse fixtures to guard each branch of the HRW
type-hint strip.

### DIFF
--- a/tools/hrw4u/src/hrw_symbols.py
+++ b/tools/hrw4u/src/hrw_symbols.py
@@ -304,6 +304,9 @@ class InverseSymbolResolver(SymbolResolverBase):
             quoted_value = f'"{value}"' if not (value.startswith('"') and value.endswith('"')) else value
             qargs = [qualifier, quoted_value]
         else:
+            if name == "set-config" and len(args) == 3 and args[1].upper() in {"INT", "STRING", "FLOAT"}:
+                # strip HRW type hint; canonical HRW4U is 2-arg (type resolved via TSHttpTxnConfigFind)
+                args = [args[0], args[2]]
             qargs = [self._rewrite_inline_percents(Validator.quote_if_needed(a), section) for a in args]
 
         return f"{name}({', '.join(qargs)})" if qargs else f"{name}()"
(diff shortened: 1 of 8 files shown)


ezelkow1:
[approve ci osx]

zwoop:
Hmmm, so this feels like a bandaid on a broken behavior/use of header_rewrite. It feels to me that "fixing" u4wrh to accept this broken input is not making it better, meaning it'd be better to let it give an error such that the source of this can be fixed.

Also, we may want to consider trying to fix the header_rewrite parser to make this an error on config load/reload. It clearly does not behave as you expect.

So my preference here would be to make this input to u4wrh an error, rather than handling it.

@mlibbey Wdyt ?

mlibbey:
Agree with zwoop. I think the bug here is that when u4wrh runs on 
```
cond %{CLIENT-HEADER:X-Thisdoesntcache} ="" [NOT]
  set-config proxy.config.http.cache.http INT 0
```
it yields the invalid
```
REMAP {
    if inbound.req.X-Thisdoesntcache {
        set-config("proxy.config.http.cache.http", "INT", 0);
    }
}
```
if you do the double conversion, hrw4u correctly errors:
```
<stdin>:3:8: error: Invalid number of arguments (expected 2, got 3)
   3 |         set-config("proxy.config.http.cache.http", "INT", 0);
     |         ^
```
Also interesting adding the [L] after set-config and giving to u4wrh... 
```
cond %{CLIENT-HEADER:X-Thisdoesntcache} ="" [NOT]
  set-config proxy.config.http.cache.http 0 [L]
```
does not yield a `break;`

zwoop (changes requested):
Lets address these concerns.

PR #13209, opened 2026-05-28
redo_cache_lookup: move to examples; fix `fallback` lifetime

bneradt:
The redo_cache_lookup plugin kept the fallback URL as a pointer into the plugin.config argv storage. That storage can be released after plugin initialization, leaving cache-lookup-complete callbacks to dereference stale memory.

This copies the parsed fallback URL into plugin-owned storage and passes its owned bytes to TSHttpTxnRedoCacheLookup.

Also, while investigating this, it looks like this plugin was made simply to demonstrate the use of TSHttpTxnRedoCacheLookup rather than being a production-useful plugin. The initial commit says as much and there is no customer-facing documentation for this plugin. As such, I'm moving this to the examples plugin.


bneradt:
[approve ci autest]

cmcfarlen (approved):

PR #13210, opened 2026-05-28
tscore: fix out-of-bounds read in ats_base64_decode

phongn:
## Problem

`ats_base64_decode()` reads out of bounds when the decodable (base64
alphabet) prefix of the input is not a multiple of four. The main loop

```c
for (size_t i = 0; i < inBytes; i += 4) {
  ... DECODE(inBuffer[2]) ... DECODE(inBuffer[3]) ...
}
```

iterates while `i < inBytes`, so for a prefix of length 1–3 (mod 4) it
runs one extra iteration that reads `inBuffer[2]`/`inBuffer[3]` past the
valid prefix — out of bounds of the caller's buffer when the input ends
there. The trailing adjustment then reads `inBuffer[-2]`.

This is reachable from `TSBase64Decode()` and other callers whenever a
caller passes a buffer sized to the input and the input is not 4-aligned
(e.g. unpadded base64).

## Fix

Decode only complete 4-character groups in the loop, then handle a
trailing 2- or 3-character group explicitly (a lone trailing character
does not encode a full byte and is dropped, as RFC 4648 requires). No
read ever passes the alphabet prefix.

The decoded length and bytes are **unchanged for every well-defined
input** — only the out-of-bounds reads are removed. Encoding and the
lookup table are untouched.

## Testing

This also adds `src/tscore/unit_tests/test_ink_base64.cc`, since the
base64 codec had no unit coverage upstream. Beyond the specific
regression, it exercises the scalar codec fairly broadly:

- RFC 4648 known vectors and random-data round-trips (sizes 0–256);
- the URL-safe alphabet (`-`/`_`) and both alphabets mixed in one input;
- truncation at the first non-alphabet byte (whitespace, `=`, or other
  garbage injected at every position);
- in-place decode (`dst == src`) and undersized-output-buffer rejection;
- a sweep of every non-4-aligned prefix length (the regression case).

The decode helper holds the input in an exact-size heap buffer, so under
AddressSanitizer:

- against the previous decoder the tests fault —
  `heap-buffer-overflow ... READ of size 1 in ats_base64_decode`;
- with this fix all 14,505 assertions pass clean (verified in both a
  normal build and an `-DENABLE_ASAN=ON` build).

### DIFF
--- a/src/tscore/CMakeLists.txt
+++ b/src/tscore/CMakeLists.txt
@@ -158,6 +158,7 @@ if(BUILD_TESTING)
     unit_tests/test_Throttler.cc
     unit_tests/test_Tokenizer.cc
     unit_tests/test_arena.cc
+    unit_tests/test_ink_base64.cc
     unit_tests/test_ink_inet.cc
     unit_tests/test_ink_memory.cc
     unit_tests/test_ink_string.cc
--- a/src/tscore/ink_base64.cc
+++ b/src/tscore/ink_base64.cc
@@ -33,8 +33,6 @@
 #include "tscore/ink_base64.h"
 #include "tscore/ink_assert.h"
 
-// TODO: The code here seems a bit klunky, and could probably be improved a bit.
-
 bool
 ats_base64_encode(const unsigned char *inBuffer, size_t inBufferSize, char *outBuffer, size_t outBufSize, size_t *length)
 {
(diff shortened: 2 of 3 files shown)


phongn:
Thanks for the review. Addressed in 30aa774: the test output buffers were allocated as `*_dstlen() + 1`, but those helpers already include the trailing NUL, so the extra byte would have masked a one-byte overwrite at the documented bound. They now allocate exactly the `dstlen`-reported size (verified the encoder/decoder write their NUL at `dstlen - 1`, so the tests still pass clean in both a normal and an `-DENABLE_ASAN=ON` build).

ezelkow1:
[approve ci autest 2]

masaori335:
We can ignore autest 2 failure, that's broken for now.

ezelkow1 (approved):
lgtm, just waiting on the fix for the tests to go in (unrelated to this PR, the fix is in another PR)

zwoop:
[approve ci autest 2]

bryancall (approved):
LGTM. I verified the logic, not just the description:

- The OOB read is real: the old `i < inBytes` loop ran an extra iteration on non-4-aligned prefixes, reading `inBuffer[2]`/`inBuffer[3]` past the input and then `inBuffer[-2]` in the fixup.
- Decoded length and bytes stay the same for valid input. `buf[0]` depends only on `in[0..1]`, and in the `mod==3` case `in[2]` is present, so the retained bytes match the old non-OOB path.
- No negative-index regression: `DECODE(x)` casts to `unsigned char`, so replacing the explicit `uint8_t` cast in the counting loop is equivalent.
- The trailing NUL write stays in bounds given the `decode_dstlen` guard, and in-place decode (`dst == src`) is preserved since writes trail reads.

Test coverage is excellent for a memory-safety fix: inputs in exact-size heap buffers so ASan catches any over-read, a full prefix-length sweep for the regression, plus round-trips, URL-safe and mixed alphabets, terminator injection at every position, in-place, and undersized-buffer rejection. Good to finally have unit coverage on the codec.

bryancall (approved):
Re-confirming approval. Verified again via a multi-angle review pass: the out-of-bounds fix is correct, behavior-preserving for well-defined input, and the test coverage is solid.

cmcfarlen:
Cherry-picked to 10.2.x

PR #13212, opened 2026-05-29
slice: Adjest AuTest for slice_prefetch

masaori335:
It looks like recently merged https://github.com/apache/trafficserver/pull/12949 broke `slice_prefetch` AuTest.

PR #13215, opened 2026-05-29
slice: Avoid redundant prefetch re-scheduling

zwoop:
schedule_prefetch() ran on every server response and re-issued blocks already scheduled this request, which the in-flight dedup set does not catch. Track a per-request high-water mark so each block is prefetched at most once; update the slice_prefetch gold. Supersedes #13212.


cmcfarlen (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #13216, opened 2026-05-29
header_rewrite: Fix IPv6 CIDR masking for non-aligned prefixes

zwoop:
ConditionCidr masked the byte that was just zeroed instead of the partial-prefix byte, and kept the low bits instead of the high bits, so any IPv6 prefix not a multiple of 8 (e.g. /60) was silently rounded up to the next byte boundary. Mask the byte above the zeroed region with a high-bit mask, and define the IPv4 /0 case, which was a shift-by-32 (undefined behavior).

The masking math now lives in a header-only, TS-API-free cidr.h so it can be unit-tested directly. Adds unit tests for IPv4 /0,/24,/32 and IPv6 /0,/48,/52,/60,/63,/64,/128, plus IPv4 CIDR cases in the header_rewrite bundle replay; the verifier client is IPv4 only, so IPv6 masking is covered by the unit test.

### DIFF
--- a/plugins/header_rewrite/cidr.h
+++ b/plugins/header_rewrite/cidr.h
@@ -1,0 +1,54 @@
+/*
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


zwoop:
[approve ci autest 2]

plugins/header_rewrite/cidr.h
```diff
@@ -0,0 +1,54 @@
...
+// CIDR masking helpers for %{CIDR:...}, kept header-only for unit testing.
+#pragma once
+
+#include <netinet/in.h>
+#include <arpa/inet.h>
+#include <cstdint>
+#include <cstring>
+
+// /0 yields a 0 mask, avoiding the undefined `<< 32`.
+inline in_addr_t
```
bneradt, line 28: I think we decided that we were better off not including explicit `inline` uses.

bneradt (approved):
Looks good. I made a comment about the use of `inline` - I think we decided as a community not to add more of those. But on the other hand it shouldn't hurt anything so I'm fine with it personally.

plugins/header_rewrite/cidr.h
zwoop: Good point. I did this, because in order to test this with catch2, I had to pull it out, and this really wants to be inline.

PR #13223, opened 2026-06-02
Fix mismatched sINT/dINT log field types

masaori335:
Seven boolean/counter fields were declared dINT, but their marshal functions write a single int, and proxy_protocol_version (ppv) was declared dINT while it actually marshals a string. The dINT type wrongly excludes these fields from log filters and aggregates, and the ppv mislabeling misrepresents variable-length string bytes as two fixed ints to any type-driven consumer. Retype the single-int fields as sINT and ppv as STRING so the declared type matches what each marshal function emits.

What happens on mismatch

- the field can't be used in a filter (dINT rejected)
- can't be aggregated (requires sINT). So fields like cqssl (is-client-SSL, a clean 0/1) were silently un-filterable

I'm going to introduce a new binary logging format with self description of type, it depends on this fix.

### DIFF
--- a/include/proxy/logging/LogField.h
+++ b/include/proxy/logging/LogField.h
@@ -96,10 +96,10 @@ public:
   using VarUnmarshalFunc          = std::variant<decltype(nullptr), UnmarshalFunc, UnmarshalFuncWithSlice, UnmarshalFuncWithMap>;
 
   enum Type {
-    sINT = 0,
-    dINT,
-    STRING,
-    IP, ///< IP Address.
+    sINT = 0, ///< Single INT
+    dINT,     ///< Double INT
+    STRING,   ///< String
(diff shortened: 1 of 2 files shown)


masaori335:
[approve ci autest 2of4]

bneradt (approved):
Looks good. Thanks for fixing.

Is there any way to add compile time or other verification to catch these? Apparently it's an easy mistake for devs to make. We can keep this as a separate PR if you wish, if this is even possible.

masaori335:
Yeh, it's easy to mistake. To find the error, we need to count `marshal_int/unmarshal_int` function calls, once -> `sINT`, twice -> `dINT`.

cmcfarlen:
Cherry-picked to 10.2.x

cmcfarlen:

PR #13226, opened 2026-06-02
ocsp: replace stapling_mutex with std::shared_mutex and fix leaks

c-taylor:
Replace the per-certinfo ink_mutex with a std::shared_mutex so readers on the TLS handshake hot path (ssl_callback_ocsp_stapling) and the refresh scan (ocsp_update) can run concurrently; only the cache update takes an exclusive lock. The handshake reader now copies the DER staple to a stack buffer under the shared lock and allocates outside the critical section.

Because the mutex is no longer a C type, certinfo is allocated with new/delete instead of OPENSSL_malloc.

Also fix three pre-existing bugs:
- certinfo_map_free never freed cinf->cid (leak on every CTX teardown).
- On BoringSSL the X509 key ref taken via X509_up_ref was never released; free it in certinfo_map_free.
- ssl_stapling_init_cert's error path could delete a certinfo_map still owned by the SSL_CTX. Only delete a map this call created.

This is a rework of https://github.com/apache/trafficserver/pull/13097 to separate into 2x independent PRs

### std::shared_mutex vs. bravo
The lock pressure for new TLS connections is less than volume or other event/txn mutex. Favoring the reduced memory overhead in the implementation over perfect hypothetical performance. The majority of the benefit is realised from using ANY shared type.

I would suggest that should bravo be favored it become an isolated PR.

### DIFF
--- a/src/iocore/net/OCSPStapling.cc
+++ b/src/iocore/net/OCSPStapling.cc
@@ -22,6 +22,8 @@
 #include "P_OCSPStapling.h"
 
 #include <memory>
+#include <mutex>
+#include <shared_mutex>
 
 #include <openssl/ssl.h>
 #include <openssl/x509v3.h>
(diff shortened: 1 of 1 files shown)


c-taylor:
Expecting minor conflicts after merging https://github.com/apache/trafficserver/pull/13229
Otherwise this is ready for review. I always had to resolve conflicts when separating the donor PR.

bneradt (dismissed):

c-taylor:
Rebase + resolve conflicts

bneradt:
Picking up this work here:
https://github.com/apache/trafficserver/pull/13490

PR #13228, opened 2026-06-02
Add in a vendor copy of Google Highway

phongn:
[Google Highway](https://github.com/google/highway) is a C++ library that provides portable SIMD/vector intrinsics. It currently supports all targets that ATS intends to build for (notably x86_64 and aarch64), and offers runtime dynamic dispatch.

Profiling of ATS indicates we have a number of string processing hotspots that are amenable to vector acceleration. However, each microarchitecture has different intrinsics and on x86_64 this varies wildly. Rather than take up the maintenance burden of maintaining and updating hand-written intrinsic code, we instead can use Highway to generate it for us. 

This pull request adds a snapshot of Highway 1.4.0, but without tests, documentation or non-CMake build infrastructure.


masaori335:
Do we really need to copy the lib in our repo? I thought let CMake to find the external library like what we do for jemalloc, hwloc, ...etc.

https://github.com/apache/trafficserver/blob/master/cmake/Findjemalloc.cmake
https://github.com/apache/trafficserver/blob/master/cmake/Findhwloc.cmake

masaori335:
NVM, we agreed with vendor copy at weekly bug scrub.

masaori335 (dismissed):
Thank you!

masaori335 (approved):

PR #13229, opened 2026-06-02
ocsp: add single-cert stapling fast path and certinfo RAII

c-taylor:
Skip the SSL_get_certificate() lookup and X509_cmp() DER re-parse in the stapling callback when an SSL_CTX has a single certificate. The shortcut is gated to non-dual-cert builds; under HAVE_NATIVE_DUAL_CERT_SUPPORT a CTX can hold multiple certs where only one has OCSP info, so map size alone cannot identify the negotiated cert.

Give certinfo a constructor/destructor so its resources are managed by RAII, and allocate it with make_unique. This consolidates the cleanup that was duplicated across certinfo_map_free and the init error path, and fixes two pre-existing leaks (cid and the BoringSSL cert ref) plus an error path that could delete a certinfo_map still owned by the SSL_CTX.

### DIFF
--- a/src/iocore/net/OCSPStapling.cc
+++ b/src/iocore/net/OCSPStapling.cc
@@ -21,6 +21,8 @@
 
 #include "P_OCSPStapling.h"
 
+#include <memory>
+
 #include <openssl/ssl.h>
 #include <openssl/x509v3.h>
 #include <openssl/asn1.h>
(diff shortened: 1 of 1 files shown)


src/iocore/net/OCSPStapling.cc
```diff
@@ -949,24 +956,17 @@ ssl_stapling_init_cert(SSL_CTX *ctx, X509 *cert, const char *certname, const cha
   X509_up_ref(cert);
 #endif
 
-  map->insert(std::make_pair(cert, cinf));
+  map->insert(std::make_pair(cert, cinf_ptr.release()));
```
bneradt: Should the map be changed to take the unique_ptr's? (And thus this would be a std::move rather than a release() here.)

bneradt (changes requested):

bneradt (approved):

bryancall:
@c-taylor Do you want us to merge this in to master?

c-taylor:
Local validations were in progress since my last commit for @bneradt - Good here!

c-taylor:
All my validations are complete; Good to Merge.

cmcfarlen:
Cherry-picked to 10.2.x

PR #11733, opened 2024-08-22
Fix C-style cast in Cache subsystem

masaori335:
Will be rebased when other cleanup PRs are merged.

PR #13233, opened 2026-06-03
Fix CLFUS RAM cache value metric broken by integer division

phongn:
## Summary

The CLFUS RAM cache ranks objects by a value density, `value = (hits + 1) / (size + overhead)`, compared against a running `double` average on every eviction/admission decision. Since #11733 this has been computed with **integer** division and truncates to `0` for essentially every object, silently disabling CLFUS's frequency/size/recency logic and degrading it to FIFO.

## Root cause

#11733 ("Fix C-style cast in Cache subsystem") modernized the macro:

```diff
-#define CACHE_VALUE_HITS_SIZE(_h, _s) ((float)((_h) + 1) / ((_s) + ENTRY_OVERHEAD))
+#define CACHE_VALUE_HITS_SIZE(_h, _s) (static_cast<float>(((_h) + 1) / ((_s) + ENTRY_OVERHEAD)))
```

The original C cast bound only to the numerator, so the division was floating point. The `static_cast<float>(...)` wraps the **whole** quotient, so `(hits + 1) / (size + overhead)` is now integer division. `hits` is on the order of tens while `size + overhead` is thousands+, so the result truncates to `0` for any normal object.

With `value ≡ 0`, `_average_value` stays `0` and every value-gated branch is dead (`0 > 0` is false):

- no promote-to-MRU on hit (`get()`),
- no CLOCK second-chance / requeue (`put()`),
- no value-based re-admission from the history (ghost) list.

CLFUS effectively becomes FIFO, which generally performs worse than the simpler LRU it is meant to beat. This has affected every CLFUS deployment since the regression merged on 2024-08-28.

## Fix

Bind the cast to the numerator only, restoring floating-point division (the pre-#11733 semantics). In addition, convert to modern C++ `constexpr` for type-safety:

```c++
constexpr double
cache_value_hits_size(const uint64_t hits, const uint32_t size)
{
  return static_cast<double>(hits + 1) / (size + entry_overhead);
}

constexpr double
cache_value(const RamCacheCLFUSEntry *const e)
{
  return cache_value_hits_size(e->hits, e->size);
}
```

## Test

Adds `ram_cache_clfus_value`, a focused regression test asserting the value metric is a non-zero fraction and is monotonic in hits and in size. It fails on the pre-fix macro and passes after:

- buggy macro → `FAILED` ("value metric truncated to zero", etc.)
- fixed macro → `PASSED`

The existing `ram_cache` regression still passes. With the fix, CLFUS's size-aware behavior is restored — in the bundled `ram_cache` test at the 16 MB size, its variable-size hit rate rises to ~0.82, slightly ahead of LRU (~0.80) on the same workload.

## Drive-By Changes

@moonchen suggested that the new regression test could be run at compile time via use of `static_assert`; to facilitate this change the prior C/C++ preprocessor macros and constants have been converted to modern C++ `constexpr` constants and functions. 

## Notes

- Severity: correctness regression affecting all users of the non-default `proxy.config.cache.ram_cache.algorithm = 0` (CLFUS) since #11733; a backport may be warranted.
- This restores the intended algorithm. A separate, long-standing limitation — resident entries' hit counts are only aged on the ghost list, never while resident — is **not** addressed here and is tracked separately.

Fixes a regression introduced in #11733.

### DIFF
--- a/src/iocore/cache/RamCacheCLFUS.cc
+++ b/src/iocore/cache/RamCacheCLFUS.cc
@@ -37,19 +37,16 @@
 #include <lzma.h>
 #endif
 
-#define REQUIRED_COMPRESSION 0.9 // must get to this size or declared incompressible
-#define REQUIRED_SHRINK      0.8 // must get to this size or keep original buffer (with padding)
-#define HISTORY_HYSTERIA     10  // extra temporary history
-#define ENTRY_OVERHEAD       256 // per-entry overhead to consider when computing cache value/size
-#define LZMA_BASE_MEMLIMIT   (64 * 1024 * 1024)
 // #define CHECK_ACOUNTING 1 // very expensive double checking of all sizes
 
-#define REQUEUE_HITS(_h)              ((_h) ? ((_h) - 1) : 0)
-#define CACHE_VALUE_HITS_SIZE(_h, _s) (static_cast<float>((_h) + 1) / ((_s) + ENTRY_OVERHEAD))
-#define CACHE_VALUE(_x)               CACHE_VALUE_HITS_SIZE((_x)->hits, (_x)->size)
+constexpr float    required_compression = 0.9f;
+constexpr float    required_shrink      = 0.8f;
+constexpr uint32_t history_hysteria     = 10;
+constexpr uint32_t entry_overhead       = 256; // per-entry overhead to consider when computing cache value/size
+constexpr uint32_t lzma_base_memlimit   = 64 * 1024 * 1024;
 
-#define AVERAGE_VALUE_OVER 100
-#define REQUEUE_LIMIT      100
+constexpr uint32_t average_value_over = 100;
+constexpr uint32_t requeue_limit      = 100;
 
 #ifdef DEBUG
 
@@ -83,6 +80,24 @@ struct RamCacheCLFUSEntry {
   LINK(RamCacheCLFUSEntry, hash_link);
   Ptr<IOBufferData> data;
 };
+
+constexpr uint64_t
+requeue_hits(const uint64_t hits)
+{
+  return hits ? (hits - 1) : 0;
+}
+
+constexpr double
+cache_value_hits_size(const uint64_t hits, const uint32_t size)
+{
+  return static_cast<double>(hits + 1) / (size + entry_overhead);
+}
+
+constexpr double
+cache_value(const RamCacheCLFUSEntry *const e)
+{
+  return cache_value_hits_size(e->hits, e->size);
+}
 
 class RamCacheCLFUS : public RamCache
 {
@@ -231,7 +246,7 @@ check_accounting(RamCacheCLFUS *c)
   RamCacheCLFUSEntry *y = c->lru[0].head;
   while (y) {
     x++;
-    xsize += y->size + ENTRY_OVERHEAD;
+    xsize += y->size + entry_overhead;
     y      = y->lru_link.next;
   }
   y = c->lru[1].head;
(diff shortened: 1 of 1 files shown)


moonchen (dismissed):
Change looks good.  One optional suggestion:

in your regression test, you could use `static_assert(CACHE_VALUE_HITS_SIZE(1u, 16384u) > 0.0f, ...)` and similar statements, so that the failures are caught at compile time without running the test.

paulo:
> A backport may be warranted

A backport would be ideal, I'm seeing the same behavior 🙏

phongn:
> Change looks good. One optional suggestion:
> 
> in your regression test, you could use `static_assert(CACHE_VALUE_HITS_SIZE(1u, 16384u) > 0.0f, ...)` and similar statements, so that the failures are caught at compile time without running the test.

I took the liberty of replacing the old C/C++ preprocessor macros and constants with modern `constexpr` equivalents, in addition to altering the regression test to use `static_assert`.

src/iocore/cache/RamCacheCLFUS.cc
```diff
@@ -31,24 +31,22 @@
...
 // #define CHECK_ACOUNTING 1 // very expensive double checking of all sizes
 
-#define REQUEUE_HITS(_h)              ((_h) ? ((_h) - 1) : 0)
-#define CACHE_VALUE_HITS_SIZE(_h, _s) (static_cast<float>(((_h) + 1) / ((_s) + ENTRY_OVERHEAD)))
-#define CACHE_VALUE(_x)               CACHE_VALUE_HITS_SIZE((_x)->hits, (_x)->size)
+constexpr float    required_compression = 0.9f;
+constexpr float    required_shrink      = 0.8f;
+constexpr uint32_t history_hysteria     = 10;
+constexpr uint32_t entry_overhead       = 256; // per-entry overhead to consider when computing cache value/size
+constexpr uint32_t lzma_base_memlimit   = 64 * 1024 * 1024;
```
moonchen: This is unused when building without LibLZMA, resulting in a warning/error.

src/iocore/cache/RamCacheCLFUS.cc
```diff
@@ -31,24 +31,22 @@
...
-#define HISTORY_HYSTERIA     10  // extra temporary history
-#define ENTRY_OVERHEAD       256 // per-entry overhead to consider when computing cache value/size
-#define LZMA_BASE_MEMLIMIT   (64 * 1024 * 1024)
 // #define CHECK_ACOUNTING 1 // very expensive double checking of all sizes
 
-#define REQUEUE_HITS(_h)              ((_h) ? ((_h) - 1) : 0)
-#define CACHE_VALUE_HITS_SIZE(_h, _s) (static_cast<float>(((_h) + 1) / ((_s) + ENTRY_OVERHEAD)))
-#define CACHE_VALUE(_x)               CACHE_VALUE_HITS_SIZE((_x)->hits, (_x)->size)
+constexpr float    required_compression = 0.9f;
+constexpr float    required_shrink      = 0.8f;
```
moonchen: Did you change these from double to float on purpose?

moonchen (changes requested):
👍 to converting macros to constexpr.  Just a couple of artifacts from the conversion.

src/iocore/cache/RamCacheCLFUS.cc
phongn: I'm not sure what I was thinking here, though I don't think the precision is particularly important. I reverted this back to `double`.

src/iocore/cache/RamCacheCLFUS.cc
phongn: Guarded this with the `HAVE_LZMA_H` preprocessor define.

moonchen (approved):
Both fixed — `required_compression`/`required_shrink` are back to `double`, and `lzma_base_memlimit` is now scoped inside `#ifdef HAVE_LZMA_H`. Confirmed the no-LibLZMA config compiles clean under clang `-Wall -Wextra -Werror`. LGTM.

paulo:
Any chance we can get a new minor release with this fix? 🙏

PR #13236, opened 2026-06-04
Fix truncated HTTP version in log field unmarshalling

masaori335:
I found `LogAccess::unmarshal_http_version` has a bug, it's truncated - e.g. "HTTP/1." instead of "HTTP/1.1".
This affects only `sshv` and `csshv`. (Clarification: `cqpv` and `sqpv` are not affected)

Also, remove several unused marshal/unmarshal functions.

### DIFF
--- a/src/proxy/logging/unit-tests/test_LogAccess.cc
+++ b/src/proxy/logging/unit-tests/test_LogAccess.cc
@@ -26,6 +26,7 @@
 #include "proxy/NonHttpSmLogData.h"
 #include "proxy/logging/LogAccess.h"
 #include "proxy/logging/TransactionLogData.h"
+#include "tscore/ink_align.h"
 #include "tscore/ink_inet.h"
 
 #include <string_view>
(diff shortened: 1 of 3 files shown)


src/proxy/logging/unit-tests/test_LogAccess.cc
```diff
@@ -213,3 +214,21 @@ TEST_CASE("LogAccess non-HttpSM client host port is null-safe", "[LogAccess]")
...
+
+    char  dest[64] = {};
+    char *src      = marshalled;
+    int   len      = LogAccess::unmarshal_http_version(&src, dest, sizeof(dest));
+    REQUIRE(len > 0);
+    return std::string(dest, len);
+  };
+
+  CHECK(render(1, 1) == "HTTP/1.1");
+  CHECK(render(1, 0) == "HTTP/1.0");
```
bneradt, line 233: Minor: can you also test HTTP/2.0 and HTTP/3.0 to make sure those are logged as "HTTP/2.0" and "HTTP/3.0". I think `.0` should be suffixes for each of those.

bneradt (changes requested):

src/proxy/logging/unit-tests/test_LogAccess.cc
masaori335: Right, this field adds the `.0` for HTTP/2 and HTTP/3.

masaori335: Test cases are added. If we want to change the behavior to remove the suffix, it should be done in another PR.

bneradt (approved):
lgtm

masaori335:
[approve ci freebsd]

masaori335:
[approve ci autest 1of4 2of4]

JosiahWI:
[approve ci freebsd autest 2]

cmcfarlen:
Cherry-picked to 10.2.x

PR #13239, opened 2026-06-04
curl 8.20 test update: curl PROXY destination changes

bneradt:
curl 8.20 intentionally mirrors --haproxy-clientip into both PROXY
addresses to keep the header address family consistent. The
TSVConnPPInfo AuTest still expected the older destination address, so
jobs with newer curl failed even though ATS preserved the PROXY metadata
it received.

This relaxes the destination-address expectation to accept either curl
behavior while continuing to verify the source address and PROXY
metadata. This also wraps the long curl command strings while leaving
the test's request flow unchanged.


maskit (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #13243, opened 2026-06-06
jax_fingerprint: Emit type-default for custom log fields with no transaction

maskit:
When a log entry has no backing HttpSM (e.g. a malformed H2/H3 request rejected before transaction creation), passing nullptr to a TSLogFieldRegister'd callback dereferences null inside the plugin.  Skip the callback in that case and emit the type-appropriate default the way built-in fields do: 0 for ints, "-" for strings, AF_UNSPEC for IPs.

### DIFF
--- a/include/proxy/logging/LogAccess.h
+++ b/include/proxy/logging/LogAccess.h
@@ -328,7 +328,7 @@ public:
   void set_http_header_field(LogField::Container container, char *field, char *buf, int len);
 
   // Plugin
-  int marshal_custom_field(char *buf, const LogField::CustomMarshalFunc &plugin_marshal_func);
+  int marshal_custom_field(char *buf, LogField::Type type, const LogField::CustomMarshalFunc &plugin_marshal_func);
 
   //
   // unmarshalling routines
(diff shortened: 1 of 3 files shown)


amirnoshian:
I’m trying to build the project using version 8, but I’m encountering some issues.

When building traffic_opc with the Dockerfile, there seems to be a problem with pdgp:13. It looks like this dependency or image might need to be changed.

Additionally, I’m also having issues with the Traffic Ops Dockerfile during the build process.

Does anyone have any suggestions on how to resolve these problems?

JosiahWI:
@amirnoshian If you could open an issue that would help us track it. This is not the right thread for your comment. Please ask if you would like help opening an issue or have questions about it.

amirnoshian:
thanks for your feedback now i make the new issue do you have idea for
thats problem ?

bneradt:
[approve ci osx freebsd]

bneradt (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #13244, opened 2026-06-06
[10.2.x] backport #13210: fix out-of-bounds read in ats_base64_decode

dkgkdfg65:
10.2.x is missing the base64-decode bounds fix that's on master (#13210). On 10.2.x ats_base64_decode walks the input in 4-char groups but doesn't guard the tail when the length isn't a multiple of 4, so a crafted non-aligned base64 string reads past the input buffer.

checked it actually over-reads on 10.2.x (not just a missing line): built ats_base64_decode with -fsanitize=address on ubuntu:22.04 and fed a non-4-multiple input. pre-fix ASan reports a heap-buffer-overflow READ just past the buffer and aborts; with the master fix cherry-picked it decodes cleanly. the upstream unit test (test_ink_base64.cc) comes along in the cherry-pick.

clean cherry-pick (-x), original author (Phong Nguyen) preserved. happy to rebase if you'd prefer.

upstream: https://github.com/apache/trafficserver/commit/1e1dd1aabeef958d3d876296d0411b8250cfbfcb

### DIFF
--- a/src/tscore/CMakeLists.txt
+++ b/src/tscore/CMakeLists.txt
@@ -158,6 +158,7 @@ if(BUILD_TESTING)
     unit_tests/test_Throttler.cc
     unit_tests/test_Tokenizer.cc
     unit_tests/test_arena.cc
+    unit_tests/test_ink_base64.cc
     unit_tests/test_ink_inet.cc
     unit_tests/test_ink_memory.cc
     unit_tests/test_ink_string.cc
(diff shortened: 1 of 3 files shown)


bryancall:
@cmcfarlen Can you check to see if the origin PR cherry picks in clean, if not you can use the PR.

cmcfarlen:
This picked clean.  No need for back port PR.

PR #13254, opened 2026-06-10
Break diags log rotation into subroutines

JosiahWI:
This change reduces a small amount of duplication and introduces meaningful function names. I'm cleaning this up now because there are (very minor) bugs in the log rotation logic (I have not filed issues yet), and my hope is that this refactoring makes the code easier to understand and review for those bugfixes.

This change restructures conditional flow to use guard clauses instead of nesting all the logic in conditional checks.

* Extract `Diags::roll_diags_on_filesize`

* Extract `Diags::roll_diags_on_interval`

* Extract `Diags::roll_streams_on_filesize`

* Extract `Diags::roll_streams_on_interval`

* Extract `Diags::remake_logfile`

* Extract `Diags::swap_diagslog`

### DIFF
--- a/src/tscore/Diags.cc
+++ b/src/tscore/Diags.cc
@@ -48,6 +48,8 @@
 #include "tscore/Regression.h"
 #include "tscore/Diags.h"
 #include "ts/ts.h"
+
+#include <memory>
 #include <string_view>
 
 int DiagsConfigState::_enabled[2] = {0, 0};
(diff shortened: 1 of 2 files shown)


src/tscore/Diags.cc
```diff
@@ -634,83 +591,19 @@ Diags::should_roll_outputlog()
...
-  // (and useful) when traffic_server is NOT running in stand alone mode. If traffic_server
-  // is NOT running in stand alone mode, then stderr and stdout SHOULD ALWAYS be pointing
-  // to the same file (traffic.out).
-  //
-  // If for some reason, someone wants the feature to have stdout pointing to some file on
-  // disk, and stderr pointing to a different file on disk, and then also wants both files to
-  // rotate according to the (same || different) scheme, it would not be difficult to add
-  // some more config options in records.yaml and said feature into this function.
-  if (ret_val) {
-    ink_assert(!need_consider_stderr);
```
JosiahWI, line 711: Instead of asserting that this flag was set, I am asserting at an earlier point that the condition for setting the flag is met. The code was originally designed to be a little easier to change if more output stream flexibility was needed. Since that obviously hasn't happened for a long time, and was only a small amount of code, I have removed the extra logic flow and this flag entirely, simplifying the code.

cmcfarlen:
Maybe we should deprecate these features for 11.  @zwoop what say you?

PR #13256, opened 2026-06-11
Fix mismatched log field types more

masaori335:
Following https://github.com/apache/trafficserver/pull/13223. Four more LogFields are initialized with unmatched types with marshal/unmarshal function.

Seems we're easily make mistake with this, we need something to check it at compile time.


cmcfarlen (approved):

cmcfarlen:
Cherry-picked to 10.2.x

cmcfarlen:

PR #13258, opened 2026-06-11
authproxy: Release client request handle in StateAuthorized

maskit:
TSHttpTxnClientReqGet was not paired with TSHandleMLocRelease, unlike every other call site in this file. The handle is a top-level HTTP_HEADER mloc so the release is effectively a no-op today, but matching the documented API contract avoids surprises if the SDK implementation ever changes.


JosiahWI (approved):
Fixes a lock leak

PR #13263, opened 2026-06-12
Fix bounds check in CacheVC::scanObject

traeak:
Fix bounds check in CacheVC::scanObject, requires on disk corruption.

The prior check compared doc->data() - buf->data() against io.aiocb.aio_nbytes cast to int. The pointer arithmetic could wrap when doc->hlen (read from disk) was very large, and the int cast truncated size_t buffer sizes, so the check could pass while doc->hdr() and the subsequent HTTPInfo::unmarshal walked memory past the I/O buffer.                                                         
                                                                                
Re-express the check in the unsigned size_t domain and validate doc->hlen directly against the bytes remaining after the Doc header, guarding each subtraction so it cannot underflow.

### DIFF
--- a/src/iocore/cache/CacheVC.cc
+++ b/src/iocore/cache/CacheVC.cc
@@ -766,9 +766,17 @@ CacheVC::scanObject(int /* event ATS_UNUSED */, Event * /* e ATS_UNUSED */)
       }
       break;
     }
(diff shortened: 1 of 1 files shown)


masaori335 (approved):
Looks good. Thank you!

cmcfarlen:

cmcfarlen:
Cherry-picked to 10.2.x

PR #13264, opened 2026-06-12
Handle OpenSSL without dynamic ENGINE

bneradt:
Some OpenSSL builds expose the async APIs ATS can use while disabling the legacy ENGINE interface. On those platforms, such as RHEL 10, the old test-only `async_engine` plugin fails to compile even though runtime TLS async support can still be built and exercised without ENGINE.

This removes the dynamic ENGINE dependency from the AuTest coverage. The replacement test plugin runs from a TLS certificate hook, installs an OpenSSL async wait fd, pauses the current async job, and wakes it from a worker thread. That keeps the test focused on ATS's `SSL_ERROR_WANT_ASYNC` handshake handling without relying on the deprecated ENGINE API.


tests/gold_tests/tls/tls_engine.test.py
```diff
@@ -25,16 +25,19 @@
 Test tls via the async interface with the sample async_engine
 '''
 
+async_engine = os.path.join(Test.Variables.AtsTestPluginsDir, 'async_engine.so')
+
 Test.SkipUnless(
     Condition.HasOpenSSLVersion('1.1.1'),
     Condition.IsOpenSSL(),
+    Condition(lambda: os.path.isfile(async_engine), async_engine + " not found."),
```
JosiahWI: Good. We can't use `Condition.PluginExists` because the test plugin isn't already in the runtime plugin directory.

JosiahWI (dismissed):
Looks good.

maskit (dismissed):
This is fine for now but we should rewrite it with new API (Provider) or remove it.

bneradt:
> This is fine for now but we should rewrite it with new API (Provider) or remove it.

This is a really good point. I reworked this to remove the ENGINE dependency entirely. The replacement `tls_async_handshake` AuTest uses OpenSSL's async job/wait-fd API from a TLS cert hook, so it still validates ATS's `SSL_ERROR_WANT_ASYNC` handshake path without requiring dynamic ENGINE support.

maskit (approved):

cmcfarlen:

cmcfarlen:
Cherry-picked to 10.2.x

PR #13269, opened 2026-06-15
Cripts: make URL::Query copyable via deep-copy of _state

serrislew:
`Url::Query` holds its parsed-parameter scratch space in a `std::unique_ptr<State>`, which implicitly deletes the copy ctor and copy-assign. This PR defines the copy operations explicitly to deep-copy `*_state`, and `= default`s the moves so they aren't suppressed by the user-declared copies.

### DIFF
--- a/include/cripts/Urls.hpp
+++ b/include/cripts/Urls.hpp
@@ -335,6 +335,25 @@ public:
 
     using Component::Component;
 
+    // _state is deep-copied, but the segment `string_view`s (and Component::_owner) still
+    // reference the source URL. The copy's views dangle if the source URL's path is rewritten,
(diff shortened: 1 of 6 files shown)


zwoop (changes requested):
Looks good! On further inspection, I'm pretty sure we want the same fixes for the path URL component as well.

Note that this is technically still not a deep copy, but it's close enough for now, and this patch restore previous (pre-PIMPL) behavior. [It shares ownership with the source URL still, but that's a non-issue unless you explicitly .Flush() the copy].

zwoop (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #13276, opened 2026-06-15
Fix tsapi build with ENABLE_PROBES=ON

Clendenin:
Cache headers transitively pull in <ts/ats_probe.h> via P_CacheDir.h. When ENABLE_PROBES=ON, ENABLE_SYSTEMTAP_PROBES is defined and ats_probe.h `#include <sys/sdt.h>`. tsapi's include path didn't have lib/systemtap.


cmcfarlen:
This was introduced in #13245, but the root issue is including a private header file from a public one and exposing impl details.

cmcfarlen (approved):

cmcfarlen:
[approve ci autest 3]

bneradt (approved):

PR #13280, opened 2026-06-16
Enable probes in Fedora C++20 CI

bneradt:
Fedora CI runs the C++20 preset rather than the base Fedora preset, so
probe coverage drifted out of that build even though ci-fedora had the
option enabled.

This makes the C++20 preset inherit the Fedora preset and keeps only the
C++ standard override locally, so shared Fedora CI settings stay in one
place.


masaori335 (approved):

JosiahWI:
[approve ci osx]

cmcfarlen:

cmcfarlen:
Cherry-picked to 10.2.x

PR #13283, opened 2026-06-16
Add unit tests for `Continuation` logic

JosiahWI:
This change was generated by Claude Opus 4.7 from API contracts I have not yet opened PRs for. It does not touch existing files other than one CMakeLists.txt, so it is low-risk.

I reviewed each test case and they are good. I have not reviewed code semantics carefully for correctness. The default delay for the `CountingContinuation` wait is five seconds; I think a wait with a timeout is an acceptable way to test this, and five seconds should be plenty. If an event scheduled to run immediately doesn't call back the associated continuation within five seconds, that's probably unacceptable performance and worth failing the test for. :D


src/iocore/eventsystem/unit_tests/inkevent_test_fixtures.h
```diff
@@ -0,0 +1,157 @@
...
+private:
+  std::atomic<bool> flag{false};
+};
+
+/**
+  Continuation whose handler counts every dispatch into a public atomic
+  counter. The first consumer is the Continuation tests in
+  test_Continuation.cc; later inkevent tests reuse it whenever they need
+  to observe handler invocations as an externally-visible side effect
+  (FR-003).
```
JosiahWI: Need to remove this. 😡

moonchen (dismissed):
Saw your "need to remove this" 😡 — I assume the duplicate EventProcessorListener gets reconciled in a follow-up? Happy to leave it for a later PR since this is test-only.

JosiahWI:
@moonchen I was mad at the FR-003 in the comment text. The agent sometimes refers to various parts of its spec documents in the code comments, which is unhelpful, since you don't have access to the spec documents.

Do you mean the duplicate listeners in the existing unit test files? I can file a PR for that following this one. 👍

moonchen (approved):

PR #13286, opened 2026-06-16
header_rewrite: count operators and conditions run

moonchen:
Add proxy.process.plugin.header_rewrite.{operators,conditions} to track how much work header_rewrite does.  These increment each time an operator or condition is executed.

### DIFF
--- a/plugins/header_rewrite/condition.h
+++ b/plugins/header_rewrite/condition.h
@@ -49,6 +49,10 @@ public:
   bool
   do_eval(const Resources &res)
   {
+    // In do_eval, so AND/OR short-circuited conditions aren't counted.
+    if (hrw_stat_conditions != nullptr) {
+      ts::Metrics::Counter::increment(hrw_stat_conditions);
+    }
     bool rt = eval(res);
 
     if (has_modifier(_mods, CondModifiers::NOT)) {
--- a/plugins/header_rewrite/statement.cc
+++ b/plugins/header_rewrite/statement.cc
@@ -21,6 +21,24 @@
 //
 #include "statement.h"
 
+ts::Metrics::Counter::AtomicType *hrw_stat_operators  = nullptr;
+ts::Metrics::Counter::AtomicType *hrw_stat_conditions = nullptr;
+
+void
+init_hrw_work_stats()
+{
+  // createPtr() is idempotent and internally synchronized: it returns the existing counter when
+  // one already has the given name. header_rewrite.so can be loaded both globally (TSPluginInit)
+  // and as a remap plugin (TSRemapInit), so init runs more than once; every call resolves to the
+  // same process-wide counter regardless of ordering or which thread runs it.
+  if (nullptr == hrw_stat_operators) {
+    hrw_stat_operators = ts::Metrics::Counter::createPtr("proxy.process.plugin.header_rewrite.operators");
+  }
+  if (nullptr == hrw_stat_conditions) {
+    hrw_stat_conditions = ts::Metrics::Counter::createPtr("proxy.process.plugin.header_rewrite.conditions");
+  }
+}
+
 void
 Statement::append(Statement *stmt)
 {
--- a/plugins/header_rewrite/statement.h
+++ b/plugins/header_rewrite/statement.h
@@ -27,10 +27,18 @@
 #include <vector>
 
 #include "ts/ts.h"
+#include "tsutil/Metrics.h"
 
 #include "resources.h"
 #include "parser.h"
 #include "lulu.h"
+
+// Counters for the number of header_rewrite operators executed and conditions evaluated; the
+// bare hook invocation count cannot tell a small ruleset from a large one. Created once via
+// init_hrw_work_stats(); both are null until then, so increments are guarded.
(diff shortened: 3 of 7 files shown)


tests/gold_tests/autest-site/ats_replay.test.ext
```diff
@@ -323,14 +323,28 @@ def ATSReplayTest(obj, replay_file: str):
...
+                    f'^{re.escape(metric_name)}\\s+{expected_value}$', f'{metric_name} should be {expected_value}')
+                check_tr.StillRunningAfter = ts
+            elif 'min' in check:
+                minimum = check['min']
+                check_tr = obj.AddTestRun(f'Verify {metric_name} >= {minimum}')
+                # Fail if the metric value (field 2) is below the minimum, or absent.
+                # $$ escapes the awk field ref from autest's command templating.
+                check_tr.Processes.Default.Command = (
+                    f"traffic_ctl metric get {metric_name} | "
+                    f"awk 'BEGIN {{ rc = 1 }} {{ if ($$2 + 0 >= {minimum}) rc = 0 }} END {{ exit rc }}'")
```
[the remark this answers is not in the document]
moonchen: `$$` here is the escape for autest's command templating (Python `Template`-style `$` substitution), so `$$2` reaches awk as `$2` (the second field) at run time — as noted in the inline comment on the line above. Writing `$2` directly would be consumed by the command templating rather than passed through to awk, so the current form is intentional (and the metric check passes).

plugins/header_rewrite/condition.h
```diff
@@ -49,6 +49,10 @@ class Condition : public Statement
   bool
   do_eval(const Resources &res)
   {
+    // In do_eval, so AND/OR short-circuited conditions aren't counted.
```
JosiahWI, line 52: Good comment.

plugins/header_rewrite/statement.h
```diff
@@ -32,6 +32,13 @@
 #include "parser.h"
 #include "lulu.h"
 
+// Counters for the number of header_rewrite operators executed and conditions evaluated; the
+// bare hook invocation count cannot tell a small ruleset from a large one. Created once via
+// init_hrw_work_stats(); both are TS_ERROR until then, so increments are guarded.
```
JosiahWI: Is `init_hrw_work_stats` not serialized with the hook registration?

plugins/header_rewrite/statement.cc
```diff
@@ -21,6 +21,24 @@
...
+int hrw_stat_operators  = TS_ERROR;
+int hrw_stat_conditions = TS_ERROR;
+
+void
+init_hrw_work_stats()
+{
+  // Plugin init is single-threaded, so the create-once guard is safe. header_rewrite.so can
+  // be loaded both globally and as a remap plugin; both must share one process-wide stat.
+  if (TS_ERROR == hrw_stat_operators) {
+    hrw_stat_operators = TSStatCreate("proxy.process.plugin.header_rewrite.operators", TS_RECORDDATATYPE_INT,
```
JosiahWI: Our API documentation says that `TsStatCreate` is deprecated since ATS 10, and that we should use the Metrics.h APIs instead. I think we should heed that if possible.

JosiahWI (changes requested):

plugins/header_rewrite/statement.cc
moonchen: Thanks for catching the deprecated API. I've moved the counters to ts::Metrics.

plugins/header_rewrite/statement.h
moonchen: You're right. It is serialized. The guard was to protect from the case where header_rewrite is loaded as both a remap plugin and a global plugin.

moonchen:
[approve ci autest 0]

plugins/header_rewrite/statement.cc
```diff
@@ -21,6 +21,24 @@
...
 #include "statement.h"
 
+ts::Metrics::Counter::AtomicType *hrw_stat_operators  = nullptr;
+ts::Metrics::Counter::AtomicType *hrw_stat_conditions = nullptr;
+
+void
+init_hrw_work_stats()
+{
+  // createPtr() is idempotent and internally synchronized: it returns the existing counter when
+  // one already has the given name. header_rewrite.so can be loaded both globally (TSPluginInit)
```
JosiahWI, line 31: A reader should infer from the first sentence that `createPtr` may be invoked twice on the init path for the same metric. It is a justification for omitting a null guard. But there is a null guard, so the comment may cause confusion.

The comment does cover the important fact, which is that the plugin may be loaded as both a global and a remap plugin simultaneously. I think it's not worth rerunning CI to reword in this case, so I'm not requesting a change.

moonchen:
[approve ci autest 0]

JosiahWI (approved):
The change has been upgraded to the newer metrics API and includes an AuTest addition for the new metrics. Approving.

bneradt (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #13289, opened 2026-06-17
TLS: count handshake signatures by key type

moonchen:
A full TLS handshake runs an asymmetric signature whose cost depends on the server key type -- RSA being far heavier than ECDSA.  Count those signatures by key type to make that work visible; resumed handshakes skip the signature.

Also add proxy.process.ssl.connections_closed, incremented once per TLS connection teardown, to track teardown volume.

New counters:
  * proxy.process.ssl.handshake_sign_rsa / _ecdsa / _other
  * proxy.process.ssl.connections_closed

### DIFF
--- a/src/iocore/net/SSLNetVConnection.cc
+++ b/src/iocore/net/SSLNetVConnection.cc
@@ -991,6 +991,12 @@ SSLNetVConnection::clear()
   client_sess.reset();
 
   if (ssl != nullptr) {
+    // clear() runs from free() once per VC recycle, so this is the single chokepoint where a TLS
+    // connection's SSL object is torn down -- count it as one connection close here. Blind-tunnel
+    // conversions free their SSL earlier (before any data) and continue as a tunnel rather than a
+    // close, so by here ssl is already null for them and they are not counted (they are tracked by
+    // the tunnel metrics instead).
+    Metrics::Counter::increment(ssl_rsb.connections_closed);
     SSL_free(ssl);
     ssl = nullptr;
   }
--- a/src/iocore/net/TLSBasicSupport.cc
+++ b/src/iocore/net/TLSBasicSupport.cc
@@ -319,6 +319,28 @@ TLSBasicSupport::_update_end_of_handshake_stats()
 {
   Metrics::Counter::increment(ssl_rsb.total_success_handshake_count_in);
 
+  // Only a full handshake runs the asymmetric signature (a resumed one skips it); count those
+  // signatures by the server private-key type.
+  {
+    const SSL *ssl = this->_get_ssl_object();
+    if (ssl != nullptr && !SSL_session_reused(ssl)) {
+      if (EVP_PKEY *pkey = SSL_get_privatekey(ssl); pkey != nullptr) {
+        switch (EVP_PKEY_id(pkey)) {
+        case EVP_PKEY_RSA:
+        case EVP_PKEY_RSA_PSS:
+          Metrics::Counter::increment(ssl_rsb.handshake_sign_rsa);
+          break;
+        case EVP_PKEY_EC:
+          Metrics::Counter::increment(ssl_rsb.handshake_sign_ecdsa);
+          break;
+        default:
+          Metrics::Counter::increment(ssl_rsb.handshake_sign_other);
+          break;
+        }
+      }
+    }
+  }
+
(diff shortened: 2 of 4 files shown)


moonchen:
[approve ci freebsd]

src/iocore/net/SSLNetVConnection.cc
```diff
@@ -991,6 +991,7 @@ SSLNetVConnection::clear()
   client_sess.reset();
 
   if (ssl != nullptr) {
+    Metrics::Counter::increment(ssl_rsb.connections_closed);
     SSL_free(ssl);
     ssl = nullptr;
```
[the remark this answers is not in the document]
moonchen: `connections_closed` counts terminated TLS connections (the `clear()` teardown path). The other two `SSL_free(this->ssl)` sites are early blind-tunnel conversions — they free the SSL before any data is read and the connection continues as a blind tunnel rather than being torn down, so they are not connection closes. That blind-tunnel volume is already tracked separately by `proxy.process.tunnel.total_client_connections_*` (per tunnel type). Updated the PR description to say "once per TLS connection teardown" instead of "once per SSL_free".

src/iocore/net/TLSBasicSupport.cc
```diff
@@ -319,6 +319,28 @@ TLSBasicSupport::_update_end_of_handshake_stats()
 {
   Metrics::Counter::increment(ssl_rsb.total_success_handshake_count_in);
 
+  // Only a full handshake runs the asymmetric signature (a resumed one skips it); count those
+  // signatures by the server private-key type.
+  {
+    SSL *ssl = this->_get_ssl_object();
```
JosiahWI: Can this be pointer to const? This is a bit of a nitpick and it's not blocking, but the local on line 325 aliases locals in blocks below. If the pointed-to type is const-qualified, it should be easier for a compiler to determine that it can hoist all the `_get_ssl_object` calls and cache the pointer in a register. It wouldn't be out of scope in my opinion to make the same change to all instances in this method at the same time. If you don't want to do this now, provide a justification and mark resolved; I'll re-review.

src/iocore/net/SSLNetVConnection.cc
```diff
@@ -991,6 +991,7 @@ SSLNetVConnection::clear()
   client_sess.reset();
 
   if (ssl != nullptr) {
+    Metrics::Counter::increment(ssl_rsb.connections_closed);
```
JosiahWI, line 999: Can this happen at an earlier point where it would be clearer the TLS connection is being terminated? The name of the method (`clear`) does not suggest this codepath is only used for TLS termination, which could be confusing to a future reader.

JosiahWI (changes requested):

src/iocore/net/TLSBasicSupport.cc
moonchen: Done for the new block. I left the other `_get_ssl_object()` locals as-is: the group-name branch calls `SSLGetGroupName(SSL *)`, which is non-const because the OpenSSL APIs it uses (`SSL_group_to_name`, `SSL_get_negotiated_group`) take a non-const `SSL *` -- that's also why the negotiated-group branch already needs a `const_cast`. Converting them all would scatter `const_cast`s rather than remove them, so I'd rather leave those. Marking resolved.

src/iocore/net/SSLNetVConnection.cc
moonchen: `clear()` is reached exactly once per VC, from `UnixNetVConnection::free()` on recycle, so it's the single chokepoint where a connection's SSL object is torn down. `do_io_close()` can be called more than once or skipped on abort paths, so counting there would over/under-count. Blind-tunnel conversions free their SSL earlier and continue as a tunnel (counted by the tunnel metrics), so by the time `clear()` runs `ssl` is already null for them and they're excluded. I added a comment at the increment spelling this out, since the concern is readability for a future reader -- but happy to move it if you'd prefer.

JosiahWI: The comment is a good alternative to moving it. Thanks for adding it!

JosiahWI (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #13306, opened 2026-06-21
tests: add TLS gold tests

moonchen:
Add autest coverage for several TLS behaviors:

- TLS renegotiation: rejected by default, and answered promptly when
  `proxy.config.ssl.allow_client_renegotiation` is enabled.
- The `proxy.config.ssl.max_record_size` clamp, verified by measuring the
  application-data record sizes on the wire, not just the response body.
- A failed outbound TLS origin connection, surfaced to the client as a 5xx.
- A TLS origin that resets the connection mid-request-body, which must fail the
  transaction promptly instead of waiting for the inactivity timeout.
- HTTP tunnel flow control (`proxy.config.http.flow_control`) on a TLS client
  connection: a slow reader makes the tunnel throttle the origin and unthrottle
  as the buffered data drains, and the full response body must still arrive. This
  path had no prior gold coverage.
- Certificate/SNI reload (`ssl_multicert.yaml` via `traffic_ctl config reload`)
  while concurrent TLS handshakes are in flight: the swapped-in certificate must
  take effect, every handshake must succeed, and ATS must not crash.

### DIFF
--- a/tests/gold_tests/tls/tls_origin_post_abort.test.py
+++ b/tests/gold_tests/tls/tls_origin_post_abort.test.py
@@ -1,0 +1,114 @@
+'''
+When a TLS origin resets the connection while ATS is still sending it a POST
+request body, ATS must fail the transaction promptly -- the transport error must
+reach the request-body tunnel right away as VC_EVENT_ERROR, not sit until the
+outbound inactivity timeout rescues it.
+'''
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import os
+import sys
+
+Test.Summary = __doc__
+
+
+class TestOriginPostAbort:
+    '''Verify a TLS origin RST mid-POST-body fails the transaction promptly.'''
+
+    _ts_counter: int = 0
+    _origin_counter: int = 0
+
+    def __init__(self) -> None:
+        '''Declare the test Processes.'''
+        Test.GetTcpPort("origin_port")
+        self._ts = self._configure_trafficserver()
+        self._origin = self._configure_origin()
+
+    def _configure_trafficserver(self) -> 'Process':
+        '''Configure Traffic Server with a TLS origin and a generous rescue timeout.
+
+        :return: The Traffic Server Process.
+        '''
+        ts = Test.MakeATSProcess(f'ts-{TestOriginPostAbort._ts_counter}', enable_cache=False)
+        TestOriginPostAbort._ts_counter += 1
+
+        ts.Disk.records_config.update(
+            {
+                'proxy.config.url_remap.remap_required': 1,
+                'proxy.config.http.connect_attempts_max_retries': 0,
+                'proxy.config.http.connect_attempts_timeout': 15,
+                # The rescue timeout: a prompt failure must beat this by a wide margin.
+                'proxy.config.http.transaction_no_activity_timeout_out': 15,
+                'proxy.config.http.transaction_no_activity_timeout_in': 30,
+                # Keep the kernel send buffer small so the request body cannot be
+                # absorbed by the kernel: ATS must still be mid-send at the RST.
+                'proxy.config.net.sock_send_buffer_size_out': 65536,
+                'proxy.config.ssl.client.verify.server.policy': 'DISABLED',
+                'proxy.config.diags.debug.enabled': 0,
+                'proxy.config.diags.debug.tags': 'http|ssl|ssl_io',
+            })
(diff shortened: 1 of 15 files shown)


tests/gold_tests/tls/tls_record_size_client.py
```diff
@@ -0,0 +1,143 @@
...
+import argparse
+import socket
+import ssl
+import sys
+
+TLS_APPLICATION_DATA = 23
+# A clamped plaintext record becomes ciphertext of plaintext + per-record overhead
+# (AEAD nonce/tag, or CBC IV/MAC/padding). 256 covers any cipher's overhead while
+# staying far below an unclamped ~16 KB record, so the clamp check stays decisive.
+RECORD_OVERHEAD = 256
```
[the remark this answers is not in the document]
moonchen: Good catch — fixed in the latest push. The client now pins AEAD (GCM) suites, so the per-record overhead is a fixed 24 bytes (8-byte explicit nonce + 16-byte tag) rather than a CBC suite's variable IV/MAC/padding. Verified on the wire: clamped records measure 4120 bytes = 4096 clamp + 24, well within the 256 allowance. This also makes the docstring's "AEAD overhead" claim actually hold regardless of cipher negotiation.

tests/gold_tests/tls/tls_record_size.test.py
```diff
@@ -0,0 +1,116 @@
...
+
+        The client downloads the object and measures the TLS records on the wire,
+        asserting both that the body is intact and that no application-data record
+        exceeds the configured clamp.
+        '''
+        tr = Test.AddTestRun("max_record_size>0 clamps records on a large TLS download")
+        tr.Processes.Default.StartBefore(self._server)
+        tr.Processes.Default.StartBefore(self._ts)
+        tr.Processes.Default.Command = (
+            f'python3 {os.path.join(Test.TestDirectory, "tls_record_size_client.py")} '
```
bneradt: For this and the other occurrences of this in this PR: instead of `python3` for these, we typically use `{sys.executable}`. Again, mention this in the AGENTS.md file as well.

tests/gold_tests/tls/tls_origin_post_abort.test.py
```diff
@@ -0,0 +1,99 @@
...
+        the connection mid-body (SO_LINGER 0). Proxy Verifier can only close cleanly
+        (close_notify), so it cannot express a mid-body reset; hence the small
+        raw-socket origin helper rather than an ATSReplayTest.
+
+        :return: The origin server Process.
+        '''
+        origin = Test.Processes.Process(
+            f'origin-{TestOriginPostAbort._origin_counter}', 'python3 {0} -p {1} -c {2} -d 1.0'.format(
+                os.path.join(Test.TestDirectory, 'tls_post_abort_origin.py'), Test.Variables.origin_port,
+                os.path.join(Test.Variables.AtsTestToolsDir, 'ssl', 'server.pem')))
```
bneradt: Please replace all .format() calls with `f'...'` strings in this PR.

While we're add it, please add a note about this in the AGENTS.md file in the section about writing autests.

tests/gold_tests/tls/tls_record_size_client.py
```diff
@@ -0,0 +1,154 @@
...
+TLS_APPLICATION_DATA = 23
+# A clamped plaintext record becomes ciphertext of plaintext + AEAD overhead
+# (TLS1.2 GCM: 8-byte explicit nonce + 16-byte tag = 24 bytes; the cipher is pinned
+# to AEAD below). 256 is a generous ceiling over that, far below an unclamped ~16 KB
+# record, so the clamp check stays decisive and cannot be tripped by the larger,
+# variable expansion of a CBC suite.
+RECORD_OVERHEAD = 256
+
+
+def iter_record_lengths(buf):
```
bneradt: Can you add type declarations for this?

bneradt (changes requested):

bneradt:
One additional issue I noticed:

- `tests/gold_tests/tls/tls_post_abort_client.py:67`: the helper treats any response header from ATS as a prompt failure and returns PASS as long as it arrives before the threshold. That can make `tls_origin_post_abort.test.py` pass without exercising the TLS origin reset path: if the origin never starts, the remap breaks, ATS rejects the request before forwarding the body, or the TLS handshake fails immediately, the client still sees `outcome = 'response received'` and passes. Please tighten this so the test proves the intended scenario, for example by checking a 5xx status plus origin-side markers (`request headers received`, `connection reset sent`) and/or requiring some request-body bytes to be sent before the failure.

Signed, Codex (GPT-5)

tests/gold_tests/tls/tls_record_size.test.py
moonchen: Done — switched all the helper-launch `Command` strings in the PR to `{sys.executable}`. Added a note to the Writing Autests section of AGENTS.md.

tests/gold_tests/tls/tls_origin_post_abort.test.py
moonchen: Done — converted every `.format()` call in the PR's files to f-strings, and added a note to the Writing Autests section of AGENTS.md.

tests/gold_tests/tls/tls_record_size_client.py
moonchen: Added: `iter_record_lengths(buf: bytes | bytearray) -> Iterator[tuple[int, int]]`, plus the `flush`/`feed` closures in the same function. Also annotated `main()` in `tls_renegotiation_client.py` for consistency, and added a line about annotating helper functions to the AGENTS.md note above.

moonchen:
Tightened on both sides. The client now requires (a) some request-body bytes were sent, (b) a prompt failure, and (c) a 5xx status — a bare connection drop or any non-5xx now fails instead of passing. The test additionally asserts the origin printed `request headers received` and `connection reset sent`.

bneradt (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #13309, opened 2026-06-22
Fix redirected cache write without write VC

bneradt:
Redirected transactions can retain CACHE_WL_SUCCESS after the concrete
cache write VC has been cleared. When a later response reaches cache
write setup, we crash dereferencing the missing write VC.

This only reuses a redirected cache write when the prepared write VC is
still available. Otherwise, this resets the write state so ATS prepares
a fresh cache write. The added test simply adds some coverage around the
scenario but is not a true regression test. That is, the test still
passes without the src/ code change.

This addresses a crash in the cache write setup path:

```
#0 HttpSM::setup_cache_write_transfer(...)
#1 HttpSM::perform_cache_write_action()
#2 HttpSM::handle_api_return()
#3 HttpSM::state_api_callout()
#4 HttpSM::state_api_callback()
#5 TSHttpTxnReenable()
#6 EscalateResponse()
```

### DIFF
--- a/src/proxy/http/HttpTransact.cc
+++ b/src/proxy/http/HttpTransact.cc
@@ -3565,14 +3565,9 @@ HttpTransact::set_cache_prepare_write_action_for_new_request(State *s)
   // This method must be called no more than one time per request. It should
   // not be called for non-cacheable requests.
   if (s->cache_info.write_lock_state == CacheWriteLock_t::SUCCESS) {
-    // If and only if this is a redirected request, we may have already
-    // prepared a cache write (during the handling of the previous request
-    // which got the 3xx response) and can safely re-use it. Otherwise, we
-    // risk storing the response under the wrong cache key. This is a release
-    // assert because the correct behavior would be to prepare a new write,
-    // but we can't do that because we failed to release the lock. To recover
-    // we would have to tell the state machine to abort its write, and we
-    // don't have a state for that.
+    // If and only if this is a redirected request, we may have already prepared
+    // a cache write during the handling of the previous request and can re-use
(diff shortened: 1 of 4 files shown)


src/proxy/http/HttpTransact.cc
```diff
@@ -3565,16 +3565,18 @@ HttpTransact::set_cache_prepare_write_action_for_new_request(State *s)
...
-    // don't have a state for that.
+    // If and only if this is a redirected request, we may have already prepared
+    // a cache write during the handling of the previous request and can safely
+    // re-use it while the prepared write VC is still available. Otherwise, we
+    // risk storing the response under the wrong cache key.
     ink_release_assert(s->redirect_info.redirect_in_process);
-    s->cache_info.action = CacheAction_t::WRITE;
+    if (s->state_machine->get_cache_sm().cache_write_vc != nullptr) {
+      s->cache_info.action = CacheAction_t::WRITE;
+    } else {
```
JosiahWI: I think it would be better to release the cache lock and reset `write_lock_state` at the point where we close the cache VConnection.

bneradt: Thanks, sounds good. I moved the write_lock_state reset to the cache write VC close paths: the normal tunnel cache-write handler and the direct chain abort path. HttpTransact now keeps the original redirected SUCCESS reuse behavior, while the state is cleared when the prepared cache write is actually closed.

JosiahWI (approved):

moonchen (approved):

bneradt:
10.1.x PR: https://github.com/apache/trafficserver/pull/13336

cmcfarlen:
Cherry-picked to 10.2.x

PR #11214, opened 2024-04-04
Refactor Metrics storage to preserve lifetime

cmcfarlen:
Asan builds were showing use after free on metrics.  This refactors Metrics so the storage will outlive all threads that use metrics.

PR #13310, opened 2026-06-22
Introduce Clang Thread Safety Analysis, and apply it to two subsystems

moonchen:
### Motivation

A lot of ATS's concurrency contracts live only in comments and `DEBUG`-only
`ink_assert`s — "must hold this mutex," "only call on this thread." The compiler
can't see them, so the easy path is the one that silently violates them, and the
result shows up in production as a data race.

Clang's Thread Safety Analysis (`-Wthread-safety`) lets us put those contracts in
the type system: annotate which mutex guards which data, and the compiler proves
every access holds the lock. The annotations are compile-time only — zero runtime
cost, and nothing on GCC.

To show this pays for itself rather than just adding ceremony, this PR also uses
the new analysis to **find and fix an existing data race** (see below).

### What's in this PR

1. **Annotation infrastructure** (`tsutil`)
   - `tsutil/ts_thread_safety.h`: `TS_*` macros wrapping the Clang attributes;
     they expand to nothing off Clang.
   - Annotated lock types for the analysis to track: `ts::mutex` with
     `ts::lock_guard` (the annotated counterparts to `std::mutex` /
     `std::lock_guard`), and `ts::shared_mutex` with `ts::write_guard` /
     `ts::read_guard`.
   - A unit test that is itself compiled with the analysis as a worked example.
   - A CMake lane: `-Wthread-safety` is on by default for Clang as a **warning**.

2. **Annotate `SSLOriginSessionCache`** — its session map and queue are reached
   from every thread; they're now marked guarded by the cache mutex, so the
   locking discipline is compiler-enforced instead of conventional. The
   hand-rolled lock witness on `remove_oldest_session` (a `std::unique_lock`
   parameter checked with `owns_lock()`) becomes a compile-time `TS_REQUIRES`.

3. **Fix an unsynchronized read race in `Metrics::Storage`** — the analysis
   caught it (details below).

4. **Gate it in CI** — the `ci`/`branch` presets set
   `THREAD_SAFETY_ANALYSIS_AS_ERROR=ON` so findings are errors on the Clang
   lanes and gate merges, while local and `dev` builds keep them warnings.

### The bug the analysis caught

`Metrics::Storage::create()`, `createSpan()`, and `current()` access
`_cur_blob` / `_cur_off` / `_blobs` under `_mutex`, but `valid()`,
`lookup(IdType)`, and `name()` read the same fields with **no lock held**
(`rename()` likewise read them before taking the lock). A single `Storage` is
shared by all threads; metrics registered at runtime — a plugin `TSStatCreate`,
or a config reload — advance those fields while live traffic reads them through
the plugin stat APIs. That's a data race / UB.

The fix marks the fields guarded by the mutex and takes it on every access, so
the analysis enforces the locking from here on. The reads that were missing a
lock take it exclusively, matching the existing locked paths; `Storage` keeps its
plain mutex (now `ts::mutex`), so its runtime locking is otherwise unchanged.

### Why ATS-owned guards instead of `std::lock_guard`/`std::unique_lock`

Clang's analysis can only track lock state through types that carry its
annotations, and the standard RAII lock wrappers don't fit: `std::unique_lock`,
`std::shared_lock`, and `std::scoped_lock` are intentionally flexible — deferred
locking, move, `release()`, adopt — dynamic state the analysis can't follow, and
ATS doesn't need that flexibility (the common case is a plain scoped lock).

So ATS provides its own simple, scoped, non-movable guards — `ts::lock_guard`
over `ts::mutex`, and `ts::write_guard` / `ts::read_guard` over
`ts::shared_mutex` — which are exactly the shape the analysis can verify. Reach
for `std::unique_lock`/`shared_lock` where the dynamic features are genuinely
needed, accepting that those sites sit outside the analysis.

### Developer impact

- **Default Clang builds**: warnings only. `-Wno-error=thread-safety` keeps it a
  warning even in `-Werror` builds, so an in-progress annotation never blocks a
  local build.
- **CI** (`THREAD_SAFETY_ANALYSIS_AS_ERROR=ON`, set in the presets): findings
  are errors on the Clang lanes and gate the merge, which keeps `master` clean —
  so the default warnings only ever flag a developer's own in-progress change.
- **GCC**: unaffected. The flag is Clang-only on purpose (GCC doesn't know
  `-Wthread-safety` and would itself error if passed it), and the macros compile
  to nothing.
- **Going forward**: prefer the annotated `ts::` guards for scoped locking and
  mark the data they protect; reach for `std::` lock guards only where the
  dynamic features are actually needed.

### Testing

- `test_tsutil` includes `test_thread_safety.cc`, compiled with
  `-Werror=thread-safety` under Clang as a live example.
- The annotated translation units build clean under `clang -Wthread-safety` and
  under GCC; removing a lock from an annotated reader produces the expected
  diagnostic.

### Scope

This is a deliberately small seed: the infrastructure plus two annotated
subsystems, one of which surfaced a real fix. It is meant to establish the
pattern and the CI lane; further annotation (e.g. NetHandler/cache/HostDB state,
and enforcing "no blocking lock on ET_NET" via a thread-role capability) can
follow incrementally as each subsystem is brought to zero warnings.

### DIFF
--- a/include/tsutil/Metrics.h
+++ b/include/tsutil/Metrics.h
@@ -338,7 +338,7 @@ private:
     std::pair<int16_t, int16_t>
     current() const
     {
-      ts::scoped_lock lock(_mutex);
+      ts::lock_guard lock(_mutex);
       return {_cur_blob, _cur_off};
     }
 
@@ -347,7 +347,7 @@ private:
     {
       auto [blob, entry] = _splitID(id);
 
-      ts::scoped_lock lock(_mutex);
+      ts::lock_guard lock(_mutex);
       return (id >= 0 && ((blob < _cur_blob && entry < MAX_SIZE) || (blob == _cur_blob && entry <= _cur_off)));
     }
   };
--- a/include/tsutil/Metrics.h
+++ b/include/tsutil/Metrics.h
@@ -38,6 +38,7 @@
 #include "swoc/MemSpan.h"
 
 #include "tsutil/Assert.h"
+#include "tsutil/TsMutex.h"
 
 namespace ts
 {
@@ -304,17 +305,17 @@ private:
 
   class Storage
   {
-    BlobStorage        _blobs;
-    uint16_t           _cur_blob = 0;
-    uint16_t           _cur_off  = 0;
-    LookupTable        _lookups;
-    mutable std::mutex _mutex;
+    BlobStorage _blobs   TS_GUARDED_BY(_mutex);
+    uint16_t _cur_blob   TS_GUARDED_BY(_mutex) = 0;
+    uint16_t _cur_off    TS_GUARDED_BY(_mutex) = 0;
+    LookupTable _lookups TS_GUARDED_BY(_mutex);
+    mutable ts::mutex    _mutex;
 
   public:
     Storage(const Storage &)            = delete;
     Storage &operator=(const Storage &) = delete;
 
-    Storage()
+    Storage() TS_NO_THREAD_SAFETY_ANALYSIS // single-threaded construction; not yet shared
     {
       _blobs[0] = std::make_unique<NamesAndAtomics>();
       release_assert(_blobs[0]);
@@ -325,7 +326,7 @@ private:
     ~Storage() {}
 
     IdType           create(const std::string_view name, const MetricType type = MetricType::COUNTER);
-    void             addBlob();
+    void             addBlob() TS_REQUIRES(_mutex);
     IdType           lookup(const std::string_view name) const;
     AtomicType      *lookup(const std::string_view name, IdType *out_id, MetricType *out_type = nullptr) const;
     AtomicType      *lookup(Metrics::IdType id, std::string_view *out_name = nullptr, MetricType *out_type = nullptr) const;
@@ -337,7 +338,7 @@ private:
     std::pair<int16_t, int16_t>
     current() const
     {
-      std::lock_guard lock(_mutex);
+      ts::scoped_lock lock(_mutex);
       return {_cur_blob, _cur_off};
     }
 
@@ -346,6 +347,7 @@ private:
     {
       auto [blob, entry] = _splitID(id);
 
+      ts::scoped_lock lock(_mutex);
       return (id >= 0 && ((blob < _cur_blob && entry < MAX_SIZE) || (blob == _cur_blob && entry <= _cur_off)));
     }
   };
--- a/include/tsutil/TsSharedMutex.h
+++ b/include/tsutil/TsSharedMutex.h
@@ -27,6 +27,7 @@
 #include <pthread.h>
 #include "tsutil/Strerror.h"
 #include "tsutil/Assert.h"
+#include "tsutil/ts_thread_safety.h"
 
 #ifdef X
 #error "X preprocessor symbol defined"
@@ -48,7 +49,7 @@ namespace ts
 {
 // A class with the same interface as std::shared_mutex, but which is not prone to writer starvation.
 //
-class shared_mutex
+class TS_CAPABILITY("shared_mutex") shared_mutex
 {
 public:
   shared_mutex() {}
@@ -58,8 +59,13 @@ public:
   shared_mutex(shared_mutex const &)            = delete;
   shared_mutex &operator=(shared_mutex const &) = delete;
 
+  // The lock/unlock methods are the trusted implementation of this capability:
+  // their bodies drive the raw pthread_rwlock_t, which some libc headers (e.g.
+  // FreeBSD's <pthread.h>) annotate as a capability in its own right. Exempt the
+  // bodies from analysis so only the capability contract on each signature is
+  // checked; the actual data-race checking happens at the call sites.
   void
-  lock()
+  lock() TS_ACQUIRE() TS_NO_THREAD_SAFETY_ANALYSIS
   {
     int error = pthread_rwlock_wrlock(&_lock);
     if (error != 0) {
@@ -69,7 +75,7 @@ public:
   }
 
   bool
-  try_lock()
+  try_lock() TS_TRY_ACQUIRE(true) TS_NO_THREAD_SAFETY_ANALYSIS
   {
     int error = pthread_rwlock_trywrlock(&_lock);
     if (EBUSY == error) {
@@ -84,7 +90,7 @@ public:
   }
 
   void
-  unlock()
+  unlock() TS_RELEASE() TS_NO_THREAD_SAFETY_ANALYSIS
   {
     X(debug_assert(_exclusive);)
     X(_exclusive = false;)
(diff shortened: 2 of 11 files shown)


include/tsutil/TsSharedMutex.h
```diff
@@ -27,6 +27,7 @@
 #include <pthread.h>
 #include "tsutil/Strerror.h"
 #include "tsutil/Assert.h"
+#include "tsutil/ts_thread_safety.h"
```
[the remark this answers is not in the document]
moonchen: Addressed: added `tsutil/ts_thread_safety.h` to `TSUTIL_PUBLIC_HEADERS` so it is installed alongside the public headers that include it (`TsSharedMutex.h`, and `Metrics.h` transitively).

src/tsutil/CMakeLists.txt
```diff
@@ -80,9 +80,19 @@ if(BUILD_TESTING)
...
   )
 
+  # The thread-safety test doubles as a compile-time check: under Clang, build it
+  # with the analysis enabled and treat any finding as an error so a broken
+  # annotation chain fails the build. A no-op for GCC (the macros expand to
+  # nothing there).
+  set_source_files_properties(
+    unit_tests/test_thread_safety.cc PROPERTIES COMPILE_OPTIONS
+                                                "$<$<CXX_COMPILER_ID:Clang>:-Wthread-safety;-Werror=thread-safety>"
+  )
```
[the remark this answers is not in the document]
moonchen: Verified this works as written. CMake keeps the `;` inside the `$<...>` generator expression and emits a two-element list; the generated compile command shows Clang getting `-Wthread-safety -Werror=thread-safety` and GCC getting neither (no stray `-Werror=thread-safety>`). Leaving as-is.

src/tsutil/Metrics.cc
```diff
@@ -57,8 +57,8 @@ Metrics::Storage::addBlob() // The mutex must be held before calling this!
 Metrics::IdType
 Metrics::Storage::create(std::string_view name, const MetricType type)
 {
-  std::lock_guard lock(_mutex);
-  auto            it = _lookups.find(name);
+  ts::scoped_writer_lock lock(_mutex);
+  auto                   it = _lookups.find(name);
```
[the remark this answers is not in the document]
moonchen: Good catch, and confirmed real: with `MAX_BLOBS == 8192`, `debug_assert(_cur_blob < MAX_BLOBS)` permits `_cur_blob == 8191`, and the pre-increment in `_blobs[++_cur_blob]` then writes `_blobs[8192]`, one past the array. It is also silent in release builds since `debug_assert` is a no-op under `NDEBUG`. This is pre-existing on master (unchanged since #11214/#12238) and outside the scope of this thread-safety PR, so I will address it in a separate PR rather than expand this one. Thanks!

include/tsutil/Metrics.h
```diff
@@ -304,17 +305,17 @@ class Metrics
 
   class Storage
   {
-    BlobStorage        _blobs;
-    uint16_t           _cur_blob = 0;
-    uint16_t           _cur_off  = 0;
-    LookupTable        _lookups;
-    mutable std::mutex _mutex;
+    mutable ts::shared_mutex _mutex;
```
masaori335: Replacing `mutex` with `shared_mutex` seems discussable. I'd like spawn this to another PR for performance comparison.

moonchen: Agreed -- and rather than defer it, I removed the `shared_mutex` from `Metrics` entirely (046c451578).

The swap only happened because the annotation infrastructure shipped just `ts::shared_mutex`; there was no annotated plain mutex to reach for, and you can't put `TS_GUARDED_BY` on data behind a bare `std::mutex`. So this branch now adds `ts::mutex` and `ts::scoped_lock` -- annotated, drop-in counterparts to `std::mutex` / `std::lock_guard`. `ts::mutex` just wraps `std::mutex`, so the runtime behavior is identical, and `Metrics::Storage` is back on a plain exclusive mutex. There's no reader/writer lock left to performance-compare.

The unsynchronized-read fix stays: those reads were always meant to run under the exclusive lock and were simply missing it.

Thanks for catching this!

include/tsutil/TsSharedMutex.h
```diff
@@ -178,6 +184,38 @@ class shared_mutex
   X(std::atomic<int> _shared{0};)
 };
 
+// RAII guards for ts::shared_mutex that carry Clang thread-safety capability
+// state. Prefer these over std::unique_lock / std::shared_lock in code annotated
+// for -Wthread-safety: the analysis does not reliably track the std wrappers
+// (see tsutil/ts_thread_safety.h).
```
masaori335, line 190: The renaming from `std::unique_lock/std::shared_lock` to `ts::scoped_writer_lock/ts::scoped_reader_lock` is explicitly telling that it's adding thread-safety capability but losing flexibility? 

I thought we should name them `ts:unique_lock/ts::shared_lock` but Claude pointed out below.

> std::unique_lock/std::shared_lock are the flexible wrappers: movable, deferrable (std::defer_lock), .unlock()/.lock()/.release()/adopt. The ts:: versions are deliberately rigid — acquire-in-ctor, release-in-dtor, non-movable, no manual unlock. Someone who writes ts::unique_lock l(m); …; l.unlock(); (valid on std::unique_lock) gets a compile error. The name advertises flexibility the type doesn't have.

moonchen: I avoided `unique_lock/shared_lock` because of that flexibility.  We won't be able to supply an annotated class that is equivalent to those `std` classes.  In terms of capability, I think what I'm adding is closes to `std::lock_guard`.

What do you think about these names: `ts::lock_guard`, `ts::write_guard`, `ts::read_guard`?

masaori335: Sounds nice! Shorter is better!

include/tsutil/Metrics.h
```diff
@@ -304,17 +305,17 @@ class Metrics
...
-    BlobStorage        _blobs;
-    uint16_t           _cur_blob = 0;
-    uint16_t           _cur_off  = 0;
-    LookupTable        _lookups;
-    mutable std::mutex _mutex;
+    BlobStorage _blobs   TS_GUARDED_BY(_mutex);
+    uint16_t _cur_blob   TS_GUARDED_BY(_mutex) = 0;
+    uint16_t _cur_off    TS_GUARDED_BY(_mutex) = 0;
+    LookupTable _lookups TS_GUARDED_BY(_mutex);
+    mutable ts::mutex    _mutex;
```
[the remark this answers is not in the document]
moonchen: I tested this with Clang, and it works as written.  This isn't actually the rule for class members.

include/tsutil/TsSharedMutex.h
moonchen: Names have been changed.

masaori335 (approved):
Looks good. Thank you!

moonchen:
[approve ci autest]

cmcfarlen:
Cherry-picked to 10.2.x

PR #13312, opened 2026-06-23
Throttle OCSP cert-status error instead of logging on every handshake

ezelkow1:
This message can get a bit spammy

PR #12951 Changed it from Debug to Error

### DIFF
--- a/src/iocore/net/OCSPStapling.cc
+++ b/src/iocore/net/OCSPStapling.cc
@@ -1421,7 +1421,7 @@ ssl_callback_ocsp_stapling(SSL *ssl, void *)
   time_t current_time = time(nullptr);
   if ((cinf->resp_derlen == 0 || cinf->is_expire) || (cinf->expire_time < current_time && !cinf->is_prefetched)) {
     ink_mutex_release(&cinf->stapling_mutex);
-    Error("ssl_callback_ocsp_stapling: failed to get certificate status for %s", cinf->certname);
+    SiteThrottledError("ssl_callback_ocsp_stapling: failed to get certificate status for %s", cinf->certname);
     return SSL_TLSEXT_ERR_NOACK;
   } else {
     unsigned char *p = static_cast<unsigned char *>(OPENSSL_malloc(cinf->resp_derlen));


JosiahWI (dismissed):

src/iocore/net/OCSPStapling.cc
```diff
@@ -1421,7 +1421,7 @@ ssl_callback_ocsp_stapling(SSL *ssl, void *)
   time_t current_time = time(nullptr);
   if ((cinf->resp_derlen == 0 || cinf->is_expire) || (cinf->expire_time < current_time && !cinf->is_prefetched)) {
     ink_mutex_release(&cinf->stapling_mutex);
-    Error("ssl_callback_ocsp_stapling: failed to get certificate status for %s", cinf->certname);
+    Dbg(dbg_ctl_ssl_ocsp, "ssl_callback_ocsp_stapling: failed to get certificate status for %s", cinf->certname);
```
bneradt: Is this OK to downgrade? The other cert related messages in this function are Error. I'm not saying not to downgrade, but I just want to double check this is OK. Doesn't the user need to be alerted to issues like this?

bneradt (changes requested):

src/iocore/net/OCSPStapling.cc
bneradt: Can we turn this into a `SiteThrottledError` instead? That will compromise quieting the loud log that can emit a ton while still alerting ops of some issue with this.

ezelkow1: Good call — switched it to `SiteThrottledError` in the latest push. It stays at Error severity so ops are still alerted, but it's now throttled per call site (default `proxy.config.log.throttling_interval_msec` = 60s) and emits a "suppressed N times" count, so a flapping cert won't flood error.log. Also updated the commit message to reflect the new behavior.

bneradt (approved):

ezelkow1:
[approve ci fedora]

cmcfarlen:
Cherry-picked to 10.2.x

PR #13314, opened 2026-06-23
fix: reject @ character in header names

### LINKED ISSUE #12082: `@` is incorrectly accepted within header names
RFC 9110 states that header names must consist only of `tchar`s, which are defined as follows:
```
  tchar          = "!" / "#" / "$" / "%" / "&" / "'" / "*"
                 / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
                 / DIGIT / ALPHA
                 ; any VCHAR, except delimiters
```

`@` is not in this list, but ATS accepts it within header names, as long as it is not the first character.

To test this, send ATS a request with `@` in a header name, and observe that it forwards the invalid header without modification.

For example, sending the following request
```
GET / HTTP/1.1\r\n
Host: whatever\r\n
To@st: test\r\n
\r\n
```
elicits the following response:
```
GET / HTTP/1.1\r\n
Host: echo\r\n
To@st: test\r\n
Client-ip: 172.18.0.1\r\n
X-Forwarded-For: 172.18.0.1\r\n
Via: http/1.1 traffic_server[503cce9c-f4c1-4ee9-bd19-63c6ae774944] (ApacheTrafficServer/10.1.0)\r\n
\r\n
```

The correct behavior is to respond 400, since `@` is not a valid character in this context.

mmustafasenoglu:
## Summary

Remove the exception that allowed '@' character in header names.

## Changes

- Removed the special case that allowed '@' in header names
- Updated the comment to reference RFC 9110 Section 5.1

## Why

RFC 9110 Section 5.1 states that header field names must consist only of tchar characters:

```
tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*"
      / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
      / DIGIT / ALPHA
```

'@' is not in this list, so header names containing '@' should be rejected with a 400 response.

The previous fix for INKqa09141 incorrectly allowed '@' in header names.

Fixes #12082

### DIFF
--- a/src/proxy/hdrs/MIME.cc
+++ b/src/proxy/hdrs/MIME.cc
@@ -2426,12 +2426,11 @@ mime_parser_parse(MIMEParser *parser, HdrHeap *heap, MIMEHdrImpl *mh, const char
     /////////////////////////////////////////////
 
     /**
-     * Fix for INKqa09141. The is_token function fails for '@' character.
-     * Header names starting with '@' signs are valid headers. Hence we
-     * have to add one more check to see if the first parameter is '@'
-     * character then, the header name is valid.
+     * RFC 9110 Section 5.1: Header field names must consist only of
+     * tchar characters. The '@' character is not a valid tchar, so
+     * header names containing '@' should be rejected as invalid.
      **/
-    if ((!ParseRules::is_token(*parsed)) && (*parsed != '@')) {
+    if (!ParseRules::is_token(*parsed)) {
       continue; // toss away garbage line
     }
 


mmustafasenoglu:
Hello! This is my first contribution to Apache Traffic Server. Could you please trigger CI for this bug fix PR? Thank you!

JosiahWI:
@mmustafasenoglu This is a bug fix from a standards point of view, but anyone relying on the `@` character in internal headers could perceive this as a breaking change. Would you like to label it with Bug, Incompatible, or both? I expect it to be discussed by the community within a week, but I'm interested in your opinion on it.

mmustafasenoglu:
Hi @JosiahWI, thanks for asking! I think both labels make sense here. While this is a bug fix for RFC 9110 compliance, the  character could indeed be used in internal headers, so marking it as Incompatible is appropriate. I've added both Bug and Incompatible labels. Thanks!

bneradt:
[approve ci]

mmustafasenoglu:
Hi, following up on this bug fix PR. It fixes the @ character validation in header names per RFC 9110. Would appreciate a review. Thank you!

bneradt:
We are actually working on this issue internally. Thank you for the contribution, but we will be posting a fix for this soon.

PR #13318, opened 2026-06-24
TLS: Fix memory leaks in cert load and OCSP stapling

moonchen:
Several error and early-return paths in the TLS server certificate loading and OCSP stapling code omit cleanup and leak native objects.  Most are gated on misconfiguration (bad or unreadable cert files) or allocation failure, so they accumulate slowly rather than on every handshake.

1. setClientCertCACerts leaked the X509_STORE when X509_STORE_load_locations failed; the store is only adopted by SSL_set0_verify_cert_store on success. This fires per handshake on the sni.yaml verify_client path when the CA file or directory is bad.
2. SSLNetVConnection::populate leaked the migrated SSL object when the base populate failed: the source VC has already relinquished ownership, so the SSL must be freed when this VC does not adopt it.
3. load_certs leaked the leaf certificate on the mid-chain failure returns (bad intermediate chain, private key, or CA file).  The leaf is now held in a scoped_X509 so every path releases it.
4. _prep_ssl_ctx and SSLCreateServerContext leaked already-parsed leaves when load_certs_and_cross_reference_names failed partway, because the X509_free drain was skipped on the early return.
5. set_session_id_context leaked the client CA name stack on the digest error paths that run before ownership is handed to the SSL_CTX.
6. ssl_callback_ocsp_stapling leaked the OCSP response buffer when SSL_set_tlsext_status_ocsp_resp failed; it only takes ownership of the buffer on success and the return value was ignored.
7. stapling_refresh_response leaked the duplicated OCSP CERTID when TS_OCSP_request_add0_id failed.

### DIFF
--- a/src/iocore/net/OCSPStapling.cc
+++ b/src/iocore/net/OCSPStapling.cc
@@ -1241,6 +1241,8 @@ stapling_refresh_response(certinfo *cinf, TS_OCSP_RESPONSE **prsp)
     goto err;
   }
   if (!TS_OCSP_request_add0_id(req, id)) {
+    // add0 only takes ownership of id on success.
+    TS_OCSP_CERTID_free(id);
     goto err;
   }
 
@@ -1427,6 +1429,11 @@ ssl_callback_ocsp_stapling(SSL *ssl, void *)
     Error("ssl_callback_ocsp_stapling: failed to get certificate status for %s", cinf->certname);
     return SSL_TLSEXT_ERR_NOACK;
   } else {
+#ifdef OPENSSL_IS_BORINGSSL
+    // SSL_set_ocsp_response copies the response, so hand it the cached buffer directly.
+    int set_ok = SSL_set_ocsp_response(ssl, cinf->resp_der, cinf->resp_derlen);
+    ink_mutex_release(&cinf->stapling_mutex);
+#else
     unsigned char *p = static_cast<unsigned char *>(OPENSSL_malloc(cinf->resp_derlen));
     if (p == nullptr) {
       ink_mutex_release(&cinf->stapling_mutex);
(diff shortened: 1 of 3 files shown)


src/iocore/net/SSLUtils.cc
```diff
@@ -1167,6 +1167,8 @@ setClientCertCACerts(SSL *ssl, const char *file, const char *dir)
     X509_STORE *ctx = X509_STORE_new();
     if (X509_STORE_load_locations(ctx, file && file[0] != '\0' ? file : nullptr, dir && dir[0] != '\0' ? dir : nullptr)) {
       SSL_set0_verify_cert_store(ssl, ctx);
+    } else {
+      X509_STORE_free(ctx);
     }
```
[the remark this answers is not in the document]
moonchen: suggestion taken

src/iocore/net/OCSPStapling.cc
```diff
@@ -1435,7 +1437,11 @@ ssl_callback_ocsp_stapling(SSL *ssl, void *)
     }
     memcpy(p, cinf->resp_der, cinf->resp_derlen);
     ink_mutex_release(&cinf->stapling_mutex);
-    SSL_set_tlsext_status_ocsp_resp(ssl, p, cinf->resp_derlen);
+    if (SSL_set_tlsext_status_ocsp_resp(ssl, p, cinf->resp_derlen) == 0) {
+      // On success the call takes ownership of p and frees it; on failure it does not.
+      OPENSSL_free(p);
```
maskit, line 1448: This is a right thing to do. We need this fix.

However, with BoringSSL, `SSL_set_ocsp_response` may work nicer. No need for `OPENSSL_malloc`. 

`SSL_set_tlsext_status_ocsp_resp`:
https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_set_tlsext_status_ocsp_resp

`SSL_set_ocsp_response`:
https://commondatastorage.googleapis.com/chromium-boringssl-docs/ssl.h.html#SSL_set_ocsp_response

Just a suggestion. Not a blocker.

maskit (dismissed):
Looks good to me. Just one suggestion that can be done as an improvement later.

src/iocore/net/OCSPStapling.cc
moonchen: Thanks for pointing out the better function to use in BoringSSL.  I've updated the PR to use this in BoringSSL.  OpenSSL builds stay unchanged since `SSL_set_ocsp_response` is not available.

maskit:
Rocky build failed but the failed test looks unrelated to this change.

> =================================================================
> ==5494==ERROR: AddressSanitizer: heap-use-after-free on address 0x62f0000004f8 at pc 0x000000493994 bp 0x7f5372eddc10 sp 0x7f5372eddc00
> WRITE of size 8 at 0x62f0000004f8 thread T1 ([ET_NET 0])
> 
> 
> 99% tests passed, 1 tests failed out of 162
> 
> Total Test time (real) = 1523.51 sec
> 
> The following tests FAILED:
> 	 29 - test_cache_CacheStripe (Timeout)

maskit:
[approve ci rocky]

maskit (approved):
Thanks for the additional improvement. Looks good.

cmcfarlen:
Cherry-picked to 10.2.x

cmcfarlen:

PR #13324, opened 2026-06-24
jax_fingerprint.test.py: address log wait flakiness

### LINKED ISSUE #13300: Intermittent AuTest failure: jax_fingerprint (expected fingerprint missing from log)
The `jax_fingerprint` gold test (`tests/gold_tests/pluginTest/jax_fingerprint/jax_fingerprint.test.py`) is flaky on CI: `jax_fingerprint.log` intermittently does not contain the expected fingerprint when checked, even though the test awaits the log content. It passes on re-run with no code change.

```
file .../jax_fingerprint.log : Verify the log contains a JA4H fingerprint. - Failed
file .../jax_fingerprint.log : Verify the log contains a JA3 fingerprint. - Failed
```

bneradt:
The jax_fingerprint AuTest split its log verification between a
broad await for the method name and a later file-content assertion for
the fingerprint line. On CI, that can still leave the final assertion
observing the log before the fingerprint entry has been proven present.

This makes the await check the fingerprint-shaped log entry directly
and keeps the file-content assertion on the same pattern, so each
variant both waits for and verifies the exact log content it needs.

Fixes: #13300

### DIFF
--- a/tests/gold_tests/pluginTest/jax_fingerprint/jax_fingerprint.test.py
+++ b/tests/gold_tests/pluginTest/jax_fingerprint/jax_fingerprint.test.py
@@ -101,6 +101,7 @@ class JaxFingerprintTest:
         # HTTP/2 always runs over TLS (h2 requires TLS).
         self._needs_tls = method in ('JA3', 'JA4') or http2
         self._replay_file = self._choose_replay_file()
+        self._fingerprint_pattern = rf'{self._method}: [a-z0-9_-]+'
 
         tr = Test.AddTestRun(name)
         self._configure_dns(tr)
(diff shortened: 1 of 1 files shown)


tests/gold_tests/pluginTest/jax_fingerprint/jax_fingerprint.test.py
```diff
@@ -101,14 +101,15 @@ def __init__(
...
         self._needs_tls = method in ('JA3', 'JA4') or http2
         self._replay_file = self._choose_replay_file()
+        self._fingerprint_pattern = rf'{self._method}: [a-z0-9_-]+'
 
         tr = Test.AddTestRun(name)
         self._configure_dns(tr)
         self._configure_server(tr)
         self._configure_trafficserver()
         self._configure_client(tr)
         Test.AddAwaitFileContainsTestRun(
```
maskit, line 111: I’m not sure how this helps. `AddAwaitFileContainsTestRun` does not wait enough because it finds a wrong line that contains a method name?

bneradt: Right. Or a partial string maybe. If you look at @moonchen 's description in the corresponding issue, the Await succeeded, but then the ContainsExpression failed. This patch makes the two of those to correspond to the same string so they don't get out of sync. That seems like a reasonable thing for me to do, we'll see whether that actually addresses the test flakiness. Sometimes multiple incremental steps need to be done in CI over time to address flakey issues like this.

maskit (approved):

cmcfarlen:
Cherry-picked to 10.2.x

PR #13328, opened 2026-06-25
cache: shared-memory-backed Dir for fast restart

masaori335:
# Motivation

Cold-start cache initialization rebuilds each stripe's in-memory directory from disk on every
restart — multi-minute on large caches.

# Approach

Host the directory (`raw_dir`) in POSIX shared memory so the next process start attaches the
existing segment in milliseconds instead of rebuilding it. Recovery is binary and fail-safe:
anything untrustworthy (crash, reboot, ABI/schema mismatch, failed validation, bad disk) falls
back to the existing disk-rebuild path, and reads still validate `Doc` magic + key so a stale
entry is a miss, never served corruption.

# New configs & traffic_ctl commands

Opt-in behind `proxy.config.cache.shm.enabled` (default `0`, a functional no-op). Also adds
`proxy.config.cache.shm.{name_prefix,use_hugepages,purge_stale_on_start}` and a
`traffic_ctl cache shm status|clear` command.

# Details

Design, recovery model, configuration, and platform notes are in
`doc/developer-guide/cache-architecture/shm-fast-restart.en.rst` (added in this PR) and the
`proxy.config.cache.shm.*` entries in `records.yaml`.

# Testing
- Unit: `test_CacheShm` (ABI hash, storage signature, control round-trip, name length, prefix
  normalization, process liveness).
- AuTest: 7 `cache_shm_*` suites — fast restart, unclean shutdown, schema/storage mismatch,
  bad-disk drop + orphan reclaim, concurrent-attach refusal, purge-on-disable.
  
# Caveats
- No automated test yet exercises the `_shm_directory_is_valid()` rejection branches or the
  bad-disk / flush-failure `invalidate_stripe_directory()` paths (planned follow-up).
- macOS concurrent-attach is best-effort: the `flock` guard is authoritative on Linux but a
  no-op on macOS POSIX shm, where it falls back to an owner-pid liveness check.

### DIFF
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -3068,6 +3068,108 @@ RAM Cache
    Compression runs on task threads. To use more cores for RAM cache
    compression, increase :ts:cv:`proxy.config.task_threads`.
 
+.. _admin-cache-shm-fast-restart:
+
+Shared Memory Fast Restart
+==========================
+
+|TS| can optionally keep the cache directory -- the in-memory index that maps
+cached objects to their location on disk -- in POSIX shared memory so that it
+survives a process restart. On a normal start the directory is read from disk
+and, for a large cache, rebuilt in memory before the cache comes online. When
+this feature is enabled and the previous instance shut down cleanly, the new
+instance attaches the existing shared memory segments and skips that work,
+bringing the cache online much faster.
+
+The shared memory directory is only an optimization for restart time; the
+on-disk cache always remains the source of truth. A new instance discards the
+segments and falls back to reading the directory from disk whenever they cannot
+be trusted, including when:
+
+- the previous instance did not shut down cleanly (for example, it crashed),
+- the |TS| binary's directory structures changed (an ABI mismatch, such as
+  after an upgrade), or
+- the shared memory schema version changed.
+
+A change to the on-disk storage layout in :file:`storage.yaml` does *not*
+discard everything. Each cache stripe attaches by its own identity, so stripes
+that still exist are fast-attached, added or resized stripes are rebuilt from
+disk, and segments for stripes that no longer exist are reclaimed.
+
+Segments left over from a crash can be inspected or removed with
+``traffic_ctl cache shm status`` and ``traffic_ctl cache shm clear``, which act
+directly on the shared memory objects whether or not |TS| is running.
+
+.. note::
+
+   This is an experimental feature, disabled by default. All of its settings
+   take effect only on a restart of |TS|.
+
+.. ts:cv:: CONFIG proxy.config.cache.shm.enabled INT 0
+
+   Enables the shared memory cache directory described above. When ``0`` (the
+   default), the cache directory is always read from disk on start.
+
+   The feature requires ``shm_open()`` in libc, which |TS| checks for at build
+   time. Where it is absent -- glibc older than 2.34, which keeps it in
+   ``librt`` -- the feature is compiled out and this setting has no effect: |TS|
+   logs a warning at startup and every stripe uses a heap directory. That
+   applies to RHEL/Rocky 8, Ubuntu 20.04 and Debian 11 among others; run
+   ``traffic_layout info | grep TS_USE_CACHE_SHM`` to check a given build.
+
+.. ts:cv:: CONFIG proxy.config.cache.shm.name_prefix STRING ats
+
+   The word used to name the POSIX shared memory objects, which on Linux appear
+   under ``/dev/shm``. Set only the middle word (default ``ats``); |TS| frames it
+   as ``/<word>-`` so the leading ``/`` that POSIX requires and the trailing
+   ``-`` separator cannot be mis-typed. With the default the control segment is
+   named ``/ats-control`` and each per-stripe directory segment ``/ats-s<N>``
+   (for example ``/ats-s0``). Any stray framing characters are trimmed, so a
+   value carried over from an older release (such as ``/ats-``) still resolves to
+   the same names. Give each |TS| instance sharing a host a distinct word so
+   their segments do not collide.
+
+   Renaming this value does not remove segments created under the old prefix:
+   |TS| only manages segments under the *current* prefix, so the old ``/dev/shm``
+   objects linger until cleared manually with ``traffic_ctl cache shm clear
+   --prefix <old-word>`` (or a host reboot).
+
+.. ts:cv:: CONFIG proxy.config.cache.shm.use_hugepages INT 0
+
+   When enabled (``1``), |TS| attempts to back the shared memory directory with
+   huge pages, which cut the cost of tearing down the large directory's page
+   tables at process exit (and ease TLB pressure). This requires the shared
+   memory to be eligible for huge pages (for example, ``/dev/shm`` mounted with
+   huge page support on Linux). When it is not, |TS| logs a debug message under
+   the ``cache_shm`` tag and transparently falls back to ordinary pages, so
+   enabling this is always safe.
+
+   This advises transparent huge pages (``MADV_HUGEPAGE``) on the mapping; the
+   reserved ``MAP_HUGETLB`` pages used by the global hugepage allocator cannot
+   back a ``tmpfs``-backed shared memory segment. To avoid silently downgrading a
+   box that already runs with :ts:cv:`proxy.config.allocator.hugepages` enabled,
+   |TS| turns this on automatically when the global allocator is enabled and this
+   record is left at its default; set it to ``0`` explicitly to opt out.
+
+.. ts:cv:: CONFIG proxy.config.cache.shm.purge_stale_on_start INT 0
+
+   When enabled (``1``) and :ts:cv:`proxy.config.cache.shm.enabled` is ``0``,
+   |TS| removes any leftover shared memory segments for
+   :ts:cv:`proxy.config.cache.shm.name_prefix` at startup (the ``<prefix>control``
+   segment and the per-stripe segments it lists). This guards against two
+   hazards of running with the feature disabled after it had been enabled:
+
+   - the leftover segments keep consuming memory (for example ``/dev/shm`` on
+     Linux) even though the disabled instance never reads them, and
+   - a later run with the feature re-enabled would otherwise fast-attach a
+     directory that went stale while |TS| ran disabled and wrote only to disk.
+
+   The purge is skipped if a live process still owns the segments (a concurrent
+   instance using the same prefix), and it never blocks startup. It has no
+   effect when the feature is enabled, when no ``<prefix>control`` segment
+   exists, or when set to ``0`` (the default). ``traffic_ctl cache shm clear``
+   performs the same cleanup on demand.
+
 .. _admin-heuristic-expiration:
 
 Heuristic Expiration
--- a/doc/developer-guide/cache-architecture/shm-fast-restart.en.rst
+++ b/doc/developer-guide/cache-architecture/shm-fast-restart.en.rst
@@ -374,11 +374,12 @@ Validating a trusted segment
 
 The magic/version checks confirm the segment *looks like* a directory, but a
 stale-yet-magic-valid segment could still present offsets that would turn into
-out-of-bounds disk I/O. And ``CacheShm::mark_clean_shutdown`` runs while the
-event threads are still live (see `Shutdown`_), so a directory mutation torn at
-process exit can sit behind a ``clean_shutdown = 1`` flag.
-``Stripe::_shm_directory_is_valid`` therefore validates both the header fields
-and the shape of the entry graph before the attach:
+out-of-bounds disk I/O. And ``CacheShm::mark_clean_shutdown`` runs after
+``TSSystemState::shut_down_event_system`` but joins no event thread (see
+`Shutdown`_), so a directory mutation torn at process exit can still sit behind a
+``clean_shutdown = 1`` flag. ``Stripe::_shm_directory_is_valid`` therefore
+validates both the header fields and the shape of the entry graph before the
+attach:
 
 * ``sector_size`` is non-zero and no larger than ``STORE_BLOCK_SIZE``;
 * ``write_pos``, ``last_write_pos`` and ``agg_pos`` all lie within the stripe's
@@ -418,13 +419,39 @@ and the shape of the entry graph before the attach:
   destination. Note the bound is on where the entry *starts*, not on its extent:
   ``dir_approx_size`` rounds up, so the last object in a stripe legitimately
   overhangs ``skip + len``, which is precisely what that truncation exists for;
-* each segment's free list is walked structurally, not just bounds-checked: at most
-  one node per entry, every node empty, and ``prev`` pointing back at the node
-  nearer the head. An in-range cycle such as ``A -> B -> A`` satisfies every
-  per-entry check above and is invisible to ``check_segment()``, which walks bucket
-  chains. Attached, it lets ``freelist_pop()`` hand out an entry twice and write a
-  free-list link over the second holder's ``tag``/``phase``/``head``/``pinned``
-  word, until the ``dir_offset(e)`` guard in ``freelist_pop`` resets the segment;
+* every entry in the segment is accounted for exactly once, by
+  ``_shm_segment_membership_is_valid``. This is a membership proof, not a
+  reachability walk: the free list is walked (every node empty, ``prev`` pointing
+  back at the node nearer the head), then every bucket chain is walked (every node
+  in use, no node a bucket root), each visit marked in a scratch bitmap, and
+  finally *every* index in the segment must have been marked. Anything reached
+  twice, and anything reached at all, fails.
+
+  Reachability alone is not enough, because the state that matters is the one no
+  walk visits. ``Directory::insert()`` takes an empty row off the free list with
+  ``unlink_from_freelist()`` and only fills it several statements later; torn in
+  between, the row is empty, off the free list, not yet in any bucket chain, and
+  still holding the free-list ``prev``/``next`` it had. Every per-entry check above
+  passes, the remaining free list is self-consistent, and ``check_segment()`` walks
+  bucket chains. The next insert to scan that bucket finds the row empty, unlinks it
+  a second time, and writes ``dir_set_next`` into whatever its stale ``prev`` now
+  names -- truncating or cross-linking a live chain, or clobbering
+  ``header->freelist[s]`` when that ``prev`` is 0 -- and ``dir_set_prev`` into the
+  ``tag``/``phase``/``head``/``pinned`` word of whatever its stale ``next`` names.
+  The mirror-image tear, after the row is chained but before ``dir_assign_data``
+  fills it, leaves an empty entry inside a bucket chain; it *is* reached exactly
+  once, which is why membership is paired with the empty/in-use polarity of where
+  each entry was reached.
+
+  The same pass subsumes what a free-list-only walk caught: an in-range cycle such
+  as ``A -> B -> A`` satisfies every per-entry check and is invisible to
+  ``check_segment()``, and attached it lets ``freelist_pop()`` hand out an entry
+  twice. The invariant being proved is the one the live code maintains --
+  ``init_segment()`` frees rows 1..``DIR_DEPTH``-1 of every bucket onto the free
+  list and never row 0, ``insert()`` moves an entry from the free list to a chain,
+  and ``delete_entry()`` moves it back -- so a clean segment always satisfies it,
+  and the cost is one ``segment_entries``-bit bitmap on a walk that was already
+  linear in the segment;
 * ``Directory::check_segment()`` passes for the segment: bucket chain lengths, no
   chained-but-empty entries, no chain loops. This is the ``CHECK_DIR`` walk that is
   otherwise debug-only, and it is what turns a torn directory into a rebuild rather
@@ -483,20 +510,28 @@ Wiring
 On a clean exit, ``AutoStopCont::mainEvent`` calls
 ``sync_cache_dir_on_shutdown()`` whenever the cache is initialized.
 ``sync_cache_dir_on_shutdown`` snapshots every stripe (taking each stripe mutex,
-which excludes writers *concurrent with* the snapshot), and only then calls
-``CacheShm::mark_clean_shutdown``, which sets ``clean_shutdown = 1`` and
-``msync``\ s the header. ``owner_pid`` is deliberately left alone; see
-`Concurrent-attach guard`_. When the feature is disabled,
-``mark_clean_shutdown`` is a no-op (there is no control segment), so the shutdown
-path is unchanged for a stock |TS|.
+which excludes writers *concurrent with* the snapshot). Marking the segment clean
+is deliberately **not** part of it: ``AutoStopCont::mainEvent`` calls
+``CacheShm::mark_clean_shutdown`` itself, after ``shut_down_event_system()``, and
+that is what sets ``clean_shutdown = 1`` and ``msync``\ s the header. ``owner_pid``
+is deliberately left alone; see `Concurrent-attach guard`_. When the feature is
+disabled, ``mark_clean_shutdown`` is a no-op (there is no control segment), so the
+shutdown path is unchanged for a stock |TS|.
 
-It does not stop *future* writers: this runs before ``shut_down_event_system()``
-and the main thread exits without joining the event threads, so a stripe can
-still be written after its snapshot. That is why the fast-attach path re-validates
-the directory structurally rather than trusting ``clean_shutdown`` alone -- see
-`Validating a trusted segment`_. A late write that only lands a directory entry
-is harmless either way: the read path checks ``Doc`` magic and key before serving,
-so a stale entry resolves to a miss.
+Ordering it after ``shut_down_event_system()`` matters because the mark is a
+promise about the *whole* directory, not just the snapshot. Marked before it, any
+event thread still running could be torn mid-``Directory::insert()`` and the
+resulting half-linked directory would be published as trusted -- see
+`Validating a trusted segment`_ for what that costs.
+
+It is not a hard barrier, though: ``shut_down_event_system()`` sets a flag and the
+main thread exits without joining the event threads, so a straggler can still
+write after the mark. That is why the fast-attach path re-validates the directory
+structurally rather than trusting ``clean_shutdown`` alone. A late write that only
+lands a directory *entry* is harmless either way: the read path checks ``Doc``
+magic and key before serving, so a stale entry resolves to a miss. A late write
+that tears the directory's *links* is not, and the membership check in
+``Stripe::_shm_directory_is_valid`` is what catches it.
 
 The on-disk directory is still written
 --------------------------------------
--- a/doc/developer-guide/cache-architecture/shm-fast-restart.en.rst
+++ b/doc/developer-guide/cache-architecture/shm-fast-restart.en.rst
@@ -1,0 +1,814 @@
+.. Licensed to the Apache Software Foundation (ASF) under one
+   or more contributor license agreements.  See the NOTICE file
+   distributed with this work for additional information
+   regarding copyright ownership.  The ASF licenses this file
+   to you under the Apache License, Version 2.0 (the
+   "License"); you may not use this file except in compliance
+   with the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing,
+   software distributed under the License is distributed on an
+   "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+   KIND, either express or implied.  See the License for the
+   specific language governing permissions and limitations
+   under the License.
+
+.. include:: ../../common.defs
+.. default-domain:: cpp
+
+.. _cache-shm-fast-restart:
+
+Shared-Memory Cache Directory (Fast Restart)
+********************************************
+
+.. note::
+
+   This is an experimental feature, disabled by default. It is controlled by
+   the ``proxy.config.cache.shm.*`` settings (see :ref:`configuration
+   <cache-shm-configuration>`). The administrator-facing description lives at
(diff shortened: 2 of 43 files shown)


src/iocore/cache/CacheShmPurge.h
```diff
@@ -0,0 +1,240 @@
...
+  };
+  if (::fstat(fd, &sb) < 0 || static_cast<std::size_t>(sb.st_size) < CONTROL_SIZE) {
+    // Too small to hold a valid header/table: there is no table to walk, so just
+    // unlink the control object itself (it still occupies memory).
+    report.segment_size = static_cast<long long>(sb.st_size);
+    report.outcome      = PurgeOutcome::TooSmall;
+    int e               = ::shm_unlink(report.control_name.c_str()) == 0 ? 0 : errno;
+    report.unlinked.push_back({report.control_name, true, e});
+    return report;
+  }
```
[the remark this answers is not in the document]
masaori335: `sb` is initialized with zero by `struct stat sb{};`. I think it's format issue.

src/iocore/cache/CacheShmPurge.h
```diff
@@ -0,0 +1,241 @@
...
+  // EWOULDBLOCK is the only errno meaning "another process holds it"; anything else
+  // means flock is unusable here -> fall back to the owner_pid backstop.
+  if (errno == EWOULDBLOCK) {
+    return LockResult::HeldByOther;
+  }
+  if (unexpected_errno != nullptr) {
+    *unexpected_errno = errno;
+  }
+  return LockResult::Unsupported;
+}
```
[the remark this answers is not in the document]
masaori335: Not an issue - calling `flock` with `LOCK_NB` never returns `EINTR`.

src/iocore/cache/CacheDir.cc
```diff
@@ -947,6 +948,11 @@ sync_cache_dir_on_shutdown()
     thr.join();
   }
 
+  // All writers are now stopped (every stripe's mutex was held and released by the
+  // shutdown threads above), so the directory is final -- only now is it safe to
+  // mark the shm control segment clean.
+  CacheShm::mark_clean_shutdown();
```
moonchen: I don't think "All writers are now stopped" is right here. `shutdown()` only holds/releases each stripe's mutex, which excludes concurrent writers, not future ones. An ET_NET thread can still complete a write after a stripe's snapshot, and this runs before `shut_down_event_system()`.

Was the intention that a host would need to be fully drained before the cache dir can be persisted?

src/iocore/cache/StripeSM.cc
```diff
@@ -1342,7 +1366,15 @@ StripeSM::shutdown(EThread *shutdown_thread)
   // directories have not been inserted for these writes
   if (!this->_write_buffer.is_empty()) {
     Dbg(dbg_ctl_cache_dir_sync, "Dir %s: flushing agg buffer first", this->hash_text.get());
-    this->flush_aggregate_write_buffer(this->fd);
+    if (!this->flush_aggregate_write_buffer(this->fd)) {
```
moonchen, line 1372: What happens if the agg buffer write is pending in AIO?  Before this PR, the double write was probably harmless, but now the wrong write_pos would get picked up by the next ATS pid.

src/iocore/cache/Stripe.cc
```diff
@@ -153,15 +154,22 @@ Stripe::_init_directory(std::size_t directory_size, int header_size, int footer_
...
-    this->directory.raw_dir = static_cast<char *>(ats_alloc_hugepage(directory_size));
-    if (this->directory.raw_dir != nullptr) {
-      this->directory.raw_dir_huge = true;
-    }
-  }
-  if (nullptr == this->directory.raw_dir) {
-    this->directory.raw_dir      = static_cast<char *>(ats_memalign(ats_pagesize(), directory_size));
+  // Try a shared-memory-backed directory first; fall back to the hugepage /
+  // aligned-heap path when shm is disabled or the attach/create fails.
+  this->directory.raw_dir = CacheShm::attach_or_create_stripe(hash_text.get(), directory_size);
```
moonchen, line 159: This preempts the `ats_alloc_hugepage` fallback, so a box relying on real `MAP_HUGETLB` hugepages for the dir silently drops to 4K tmpfs pages once shm is enabled — the `MADV_HUGEPAGE` replacement is a separate default-off knob.

src/traffic_ctl/CacheShmCommand.cc
```diff
@@ -0,0 +1,261 @@
...
+    _invoked_func = [this]() { clear(); };
+  }
+}
+
+std::string
+CacheShmCommand::get_prefix()
+{
+  // The operator gives only the middle word (e.g. --prefix ats); frame it the
+  // same way the server does so the two agree on segment names.
+  std::string configured = DEFAULT_PREFIX;
```
moonchen, line 103: `records.yaml` already has the prefix and `traffic_ctl` reads it for the `config` commands — why require the operator to pass `--prefix` at all?

doc/admin-guide/files/records.yaml.en.rst
```diff
@@ -3003,6 +3003,89 @@ RAM Cache
...
+instance attaches the existing shared memory segments and skips that work,
+bringing the cache online much faster.
+
+The shared memory directory is only an optimization for restart time; the
+on-disk cache always remains the source of truth. A new instance discards the
+segments and falls back to reading the directory from disk whenever they cannot
+be trusted, including when:
+
+- the previous instance did not shut down cleanly (for example, it crashed),
+- the on-disk storage layout described by :file:`storage.yaml` changed,
```
moonchen: Does storage.yaml changing actually cause segments to be discarded?

The code comments say:

```
  /// Fingerprint of the storage topology. Not a trust gate (see initialize()):
  /// informational, drives the "storage changed" log wording.
  static uint64_t storage_signature(const Store &store);```

File: doc/developer-guide/cache-architecture/shm-fast-restart.en.rst [hunk 8]
```diff
@@ -0,0 +1,656 @@
...
+
+.. list-table::
+   :header-rows: 1
+   :widths: 42 58
+
+   * - Test
+     - Scenario
+   * - ``cache_shm_fast_restart``
+     - Directory survives a clean shutdown and is fast-attached.
+   * - ``cache_shm_data_integrity``
```
moonchen: I didn't find any data integrity tests.  Did you forget to add them to the PR?

bryancall (dismissed):
Approving. This is well built and safe to land. It is opt-in and off by default, the shared-memory attach path validates the mapping size before use and bounds the stripe count, and every failing system call is logged with its name and error string and then propagated back to the caller.

One correction to my earlier note. I wrote that a fast-attach bug returning a wrong offset or length would still pass CI. That is not accurate, and I withdraw it. The fast_restart test wires the origin to return a 502 on the hit transaction, so if the shared-memory directory failed to resolve the cached object, ATS would fall through to the origin and the 502 would fail the test. The real gap is narrower than I first stated.

The items below are all non-blocking for an off-by-default feature, but worth closing out before this is relied on in production.

1. The testing table in `doc/developer-guide/cache-architecture/shm-fast-restart.en.rst` lists a `cache_shm_data_integrity` autest that is not in this PR. Please either add the test or drop the row.

2. The objects the tests serve are small and single-fragment. A multi-fragment or large-object case after a fast attach would cover the one path the 502 trick does not fully exercise.

3. The rejection branches in `_shm_directory_is_valid()` and the invalidate-on-shutdown paths have no direct unit coverage, as the caveats note. They are close to pure header math, so each branch is a cheap unit test.

I also agree with the open Copilot threads on this PR.

4. In `CacheShm.cc:229`, `unlink_all_known_segments()` unlinks stripe names read from the control segment. On the path that drops an untrusted or corrupt control segment those names are untrusted, so a corrupt segment could lead to `shm_unlink` on unrelated objects. Since the names are always built from `name_prefix` via `build_stripe_shm_name()`, gating the unlink loop on a prefix match, and only running it when the control magic is valid, closes this with no effect on the normal path.

5. The three `storage.config` references in the design doc at lines 132, 340, and 430 should be `storage.yaml` to match the tests and the rest of this PR.

Thanks for the clean, well-documented change. The approval stands.

src/iocore/cache/CacheShm.cc
```diff
@@ -0,0 +1,748 @@
...
+unlink_all_known_segments()
+{
+  if (g_control != nullptr) {
+    for (uint32_t i = 0; i < g_control->stripe_count && i < MAX_STRIPES; ++i) {
+      std::string name = read_shm_name(g_control->stripes[i].shm_name);
+      if (!name.empty()) {
+        Dbg(dbg_ctl, "shm_unlink stripe %s", name.c_str());
+        shm_unlink(name.c_str());
+      }
+    }
```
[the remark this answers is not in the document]
bryancall: Agreeing with this. Since stripe names are always built from `name_prefix` via `build_stripe_shm_name()`, gating this unlink loop on a prefix match, and only running it when the control magic is valid, would keep a corrupt control segment from calling `shm_unlink` on unrelated objects, with no effect on the normal drop path. Worth doing before this is relied on in production.

src/iocore/cache/CacheShm.cc
```diff
@@ -0,0 +1,748 @@
...
+void *
+open_and_map_shm(const std::string &name, std::size_t size, ShmAccess access, [[maybe_unused]] HugePages hugepages,
+                 int *out_fd = nullptr, int *out_errno = nullptr)
+{
+  if (out_errno != nullptr) {
+    *out_errno = 0;
+  }
+  int oflags = O_RDWR;
+  if (access == ShmAccess::Create) {
+    oflags |= O_CREAT;
```
moonchen: Please use O_EXCL for creation, to prevent the possibility of inheriting permissions from a malicious pre-existing file.

src/iocore/cache/CacheShm.cc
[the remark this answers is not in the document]
masaori335: done

masaori335:
Thank you for taking a look. I addressed 1. 4. and 5. by 0cc3b52.

moonchen (changes requested):

doc/developer-guide/cache-architecture/shm-fast-restart.en.rst
masaori335: This is removed by https://github.com/apache/trafficserver/commit/0cc3b52eff792b7e736f2d0bfd0d3b0b0d589f13.

src/traffic_ctl/CacheShmCommand.cc
masaori335: In the case of the prefix in `record.yaml` is changed, but want to remove the old shm from `traffic_ctl` cmd.

src/iocore/cache/CacheDir.cc
masaori335: That's true, this comment is wrong. The intention is set the "clean flag" to indicate next process that previous process did not crash (did shutdown) and cache dir on shm is available.

src/iocore/cache/StripeSM.cc
masaori335: Good point. When double write (flush agg bufer + in-flight AIO) happens, `write_pos` is advanced twice. So next process will start writing from there. This is wasting disk space. Let me figure out.

src/iocore/cache/CacheShm.cc
masaori335: Addressed by 161f39c7b282933fae3596666f1452538407a448

src/iocore/cache/Stripe.cc
masaori335: Addressed by 161f39c7b282933fae3596666f1452538407a448

doc/admin-guide/files/records.yaml.en.rst
masaori335: Code is correct. Addressed by 47d89f3f73968a0cae4b4db5e8b019d4499b4161

src/iocore/cache/StripeSM.cc
masaori335: I'm going to change shutdown sequence of cache, but it'll be big change, so as a workaround, I invalidate the stripe when AIO is in progress. ( 54a2ec3402ba071787d8006cfb1cc7b0f5704aad )
```
  if (shm_backed && this->is_io_in_progress()) {
    Dbg(dbg_ctl_cache_dir_sync, "Dir %s: AIO write in flight -- invalidating shm copy, syncing dir to disk", this->hash_text.get());
    CacheShm::invalidate_stripe_directory(this->directory.raw_dir);
    shm_backed = false;
  }
```

src/iocore/cache/AggregateWriteBuffer.cc
```diff
@@ -49,7 +49,6 @@ AggregateWriteBuffer::flush(int fd, off_t write_pos) const
 {
   int r = pwrite(fd, this->_buffer, this->_buffer_pos, write_pos);
```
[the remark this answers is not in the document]
masaori335: Out of scope.

src/iocore/cache/CacheShm.cc
```diff
@@ -0,0 +1,776 @@
...
+      return nullptr;
+    }
+  } else {
+    // The kernel rounds shm size up to a page boundary (16 KiB on macOS / Apple
+    // Silicon), so accept any size in [requested, page-up].
+    struct stat sb {
+    };
+    std::size_t expected_max = INK_ALIGN(size, ats_pagesize());
+    if (fstat(fd, &sb) < 0 || sb.st_size < 0 || static_cast<std::size_t>(sb.st_size) < size ||
+        static_cast<std::size_t>(sb.st_size) > expected_max) {
```
moonchen, line 278: This wedges shm off permanently after an upgrade that changes `sizeof(CacheShmControl)` (a MAX_STRIPES bump, a longer shm_name, a new StripeEntry field): the open fails here, `initialize()` never reaches the abi_hash drop path (that needs a successful map), and the O_EXCL create then fails with EEXIST on every restart -- until someone runs `traffic_ctl cache shm clear`. The CacheShmLayout.h comment ("Bumping it changes the ABI hash, so old segments are dropped automatically") only holds for same-size changes.

Could `initialize()` treat a wrong-size (or any non-ENOENT) open failure as a stale segment -- take the flock/owner-pid guard, unlink, and fall through to the fresh create? (Alternatively: a fixed-size preamble that is mapped and checked before the full-size map, so the version fields stay readable across layout changes.)

src/iocore/cache/Stripe.cc
```diff
@@ -170,6 +178,52 @@ Stripe::_init_directory(std::size_t directory_size, int header_size, int footer_
...
+// disk I/O. On failure the caller falls through to the disk read + recover_data().
+//
+// Trust model: the shm segment is trusted to the same degree as the on-disk
+// directory (same-uid, mode 0600). Stripe geometry (segments/buckets) is recomputed
+// locally each run and raw_dir_size is exact-matched before attach, so this only
+// validates the header cursor fields and per-segment freelist heads; it does not
+// re-validate individual Dir entries -- the read path already checks Doc magic + key
+// before serving, so a stale entry resolves to a miss, never served corruption.
+bool
+Stripe::_shm_directory_is_valid() const
```
moonchen: Since these rejection code branches are critical for preventing cache corruption, I think they should be tested. `shm_poke.py` can poke the stripe segment header in /dev/shm after a clean shutdown (e.g. set `write_pos` past the stripe, or `freelist[0]` out of range); a restart should then log "shm directory invalid ... falling back to disk read" and still serve a HIT after disk recovery.

src/iocore/cache/Stripe.cc
```diff
@@ -170,6 +178,52 @@ Stripe::_init_directory(std::size_t directory_size, int header_size, int footer_
...
+
+  // write_pos/last_write_pos/agg_pos must point into the data region.
+  const off_t data_lo = this->start;
+  const off_t data_hi = this->skip + this->len;
+
+  if (this->directory.header->write_pos < data_lo || this->directory.header->write_pos > data_hi ||
+      this->directory.header->last_write_pos < data_lo || this->directory.header->last_write_pos > data_hi ||
+      this->directory.header->agg_pos < data_lo || this->directory.header->agg_pos > data_hi) {
+    return false;
+  }
```
moonchen, line 202: One optional suggestion: a clean shutdown also guarantees the write cursor is quiesced (`flush_aggregate_write_buffer` and `aggWriteDone` both maintain it, and an in-flight AIO invalidates the segment), so the validator can require it -- the runtime asserts that would otherwise catch this are debug-only.
```suggestion
  }

  // A clean shutdown leaves the write cursor quiesced; a mismatch means this is
  // not a clean-shutdown artifact.
  if (this->directory.header->agg_pos != this->directory.header->write_pos ||
      this->directory.header->last_write_pos > this->directory.header->write_pos) {
    return false;
  }
```

src/traffic_ctl/CacheShmCommand.cc
```diff
@@ -0,0 +1,266 @@
...
+
+void
+CacheShmCommand::status()
+{
+  const std::string prefix       = get_prefix();
+  const std::string control_name = cache_shm::control_segment_name(prefix);
+
+  int fd = shm_open(control_name.c_str(), O_RDONLY, 0);
+  if (fd < 0) {
+    std::cerr << "cache shm: control segment '" << control_name << "' not found: " << std::strerror(errno) << '\n';
```
moonchen: One optional suggestion: an operator with a custom `proxy.config.cache.shm.name_prefix` gets "not found" against the default `ats` prefix and may conclude shm was never enabled. Worth a hint here (and in `clear()`'s NotPresent message):
```suggestion
    std::cerr << "cache shm: control segment '" << control_name << "' not found: " << std::strerror(errno)
              << " (if proxy.config.cache.shm.name_prefix is set, pass --prefix <word>)\n";
```

src/iocore/cache/StripeSM.cc
```diff
@@ -178,6 +179,26 @@ StripeSM::init(bool clear)
     return clear_dir_aio();
   }
 
+  // shm fast restart: a clean shutdown left the in-shm directory authoritative, so
+  // skip both the disk read and recover_data() (which would re-scan the tail and
+  // discard the entries the shm copy preserved). After validating the in-shm
+  // header/footer, jump straight to dir_init_done() in the normal post-recovery
+  // state. Validation failure falls through to disk read + recover_data().
+  if (CacheShm::mode() == CacheShm::Mode::AttachExisting && CacheShm::is_shm_pointer(this->directory.raw_dir)) {
+    if (this->directory.header->magic == STRIPE_MAGIC && this->directory.footer->magic == STRIPE_MAGIC &&
```
moonchen, line 185: `mark_clean_shutdown()` runs while event threads are still live, and the main thread exits without joining them (the wait loop in traffic_server.cc wakes on the same flag and calls exit) -- so a dir mutation torn at process exit can leave a structurally inconsistent directory behind a `clean_shutdown=1` flag. Since there's no cheap way to quiesce writers first, could the fast path also run `Directory::check()` (the CHECK_DIR walk: bucket bounds, chain consistency, loop detection) and fall back to the disk read when it fails? It's an in-memory walk -- small next to the rebuild it replaces -- and it turns a torn dir into a rebuild instead of an attach.

moonchen (dismissed):
Went through the whole PR again -- the earlier threads are all addressed (resolved them). Three asks below: the wrong-size control-segment wedge, tests for the `_shm_directory_is_valid` rejection branches, and a structural check on the fast-attach path. Plus two optional suggestions.

src/iocore/cache/CacheShm.cc
masaori335: Confirmed — the size-checked open returns nullptr, so initialize() skips the whole drop block and the O_EXCL create then fails with EEXIST on every restart.

Took the second option. CONTROL_HEADER_SIZE = offsetof(CacheShmControl, stripes) is now a documented frozen prefix pinned by a static_assert, and a new open_control_segment() maps just that much when the segment's size is not this build's. The liveness guard (flock, or owner_pid where flock is unsupported) still applies, and then the segment is dropped and recreated in the same start:


cache shm: control segment /ats-control is 20001 bytes, not this build's 12336; dropping it
cache shm: creating fresh control segment /ats-control (owner pid ...)
One thing your scenario also exposed: with the table unreadable the old drop path skipped the unlink loop entirely and leaked every stripe segment — a whole directory each, so potentially GBs of shm. Both the server drop path and purge_segments() now sweep <prefix>s0..MAX_STRIPES-1 by name when the table cannot be trusted (those names are ours by construction, so nothing outside the prefix is touched).

New autest cache_shm_control_size_mismatch covers it: ts2 must drop + recreate, and then ts3 must fast-attach the segment ts2 created — that last step is what proves the wedge is gone rather than deferred by one restart.

Addressed by 80cff77c80.

src/iocore/cache/Stripe.cc
masaori335: Added cache_shm_dir_invalid. ts1 fills the cache and clean-shuts-down, then shm_poke.py tampers with the stripe segment between runs: ts2 sees write_pos pushed past the end of the stripe, ts3 sees freelist[0] pushed past the segment's entry count.

Each instance must attach the shm segments (the control segment is untouched, so this is specifically the per-stripe gate), log


shm directory invalid for '<stripe>'; falling back to disk read
not log the fast-attach line, and still serve the object out of cache (X-Cache: hit-fresh) after the disk read + recover_data(). The replay's origin response for those transactions is a 502 sentinel, so a wrongly forwarded request fails the run.

Linux-only, like the other poke tests.

Addressed by 80cff77c80.

src/iocore/cache/Stripe.cc
masaori335: Took the agg_pos half, left last_write_pos out.

agg_pos == write_pos is a real post-clean-shutdown invariant: flush_aggregate_write_buffer() and aggWriteDone() both leave them equal, an in-flight AIO invalidates the segment, and it is exactly what the debug-only asserts at Stripe.cc:464 and StripeSM.cc:746 check. Added.

last_write_pos > write_pos is a legitimate state though — agg_wrap() sets write_pos = start and agg_pos = write_pos but never touches last_write_pos, so right after a wrap last_write_pos is the pre-wrap tail. handle_recover_from_data() handles precisely that case:


recover_pos = directory.header->last_write_pos;
if (recover_pos >= skip + len) {
  recover_wrapped = true;
  recover_pos     = start;
}
So that clause would send every restart in the post-wrap window back to the disk read. On a low-traffic stripe that window can be long, and it fails silently (just a slower start), which is the worst kind of regression to notice. Left it out and noted why in the code and the design doc.

Addressed by 80cff77c80.

src/traffic_ctl/CacheShmCommand.cc
masaori335: Added, in status() and in clear()'s NotPresent. Suppressed when --prefix was actually given, so the hint only appears when the default prefix was in play.

Addressed by 80cff77c80.

src/iocore/cache/StripeSM.cc
masaori335: Added, with one prerequisite pass in front of it.

Directory::check() follows next through dir_from_offset(), which does no bounds checking, and dir_next is a uint16_t. On a stripe with fewer than 16384 buckets — so segment_entries < 65536, which a small stripe is — a torn link can walk check() off the end of the mapping and segfault instead of rejecting. That would turn "reject and rebuild" into a crash loop, so _shm_directory_is_valid() now does one sequential pass bounds-checking every entry's next and prev first, and only then runs the structural walk. prev is included because the runtime writes through it (unlink_from_freelist, delete_entry) even though check() never follows it.

Both passes are in-memory and sequential, so they are small next to the disk read + recover_data() they replace, as you say.

Also fixed the design doc, which repeated the same "all writers are quiesced" claim you flagged in CacheDir.cc — it now states that this runs before shut_down_event_system(), that the main thread does not join the event threads, and that the structural re-validation is the reason clean_shutdown alone is not trusted.

Addressed by 80cff77c80.

masaori335:
Thank you for taking a look. Issues and suggestions are addressed by 80cff77c80.

src/iocore/cache/StripeSM.cc
```diff
@@ -1326,9 +1347,26 @@ StripeSM::shutdown(EThread *shutdown_thread)
...
+  // In-flight AIO write: aggWriteDone advances write_pos again after we drop this mutex, so
+  // the shm header is not final to trust verbatim. Invalidate it and fall through to the
+  // on-disk write so the next start recovers from a fresh snapshot (short recover_data()
+  // scan) instead of a stale periodic-sync copy -- the same path a non-shm stripe takes,
+  // and recover_data() rescans the tail so the imprecise write_pos is reconciled.
+  if (shm_backed && this->is_io_in_progress()) {
+    Dbg(dbg_ctl_cache_dir_sync, "Dir %s: AIO write in flight -- invalidating shm copy, syncing dir to disk", this->hash_text.get());
+    CacheShm::invalidate_stripe_directory(this->directory.raw_dir);
+    shm_backed = false;
+  }
```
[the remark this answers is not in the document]
masaori335: Good catch, and thanks — I went with the second option you suggested: a separate per-stripe invalidation marker that never touches the disk-write buffer.

`CacheShm::invalidate_stripe_directory()` no longer stores through `raw_dir` at all. It sets a `dir_untrusted` byte on the stripe's `StripeEntry` in the control segment, which is never a source buffer for a disk write. On the next start `attach_or_create_stripe()` refuses a marked entry and creates a fresh segment; that segment's zero magic sends `StripeSM::init()` down the ordinary disk-read + `recover_data()` path, so the recovery outcome is unchanged — only the mechanism for signalling it moved.

Worth noting your diagnosis was broader than my first attempt at it. Deferring the invalidation to function exit stopped `shutdown()`'s own pwrite from carrying the zeroed magic, but the periodic dir sync reaches the same buffer from a different continuation (`memcpy(buf, stripe->directory.raw_dir, dirlen)` in `CacheDir.cc`), so that route was still open — and on that one the consequence is worse than extended recovery: a zeroed magic reaching the on-disk A/B copy makes the next start *clear* the stripe instead of recovering it. Moving the marker out of `raw_dir` closes both routes by construction, and let the deferral be deleted rather than kept as an ordering constraint.

Regression test: `"StripeSM::shutdown marks the stripe untrusted without touching the directory"` in `test_CacheShmShutdown.cc` enters this exact `is_io_in_progress()` path and asserts both that the in-shm magic is still `STRIPE_MAGIC` and that the on-disk copy reads back `STRIPE_MAGIC`. I checked it fails if the change is reverted.

src/traffic_ctl/traffic_ctl.cc
```diff
@@ -315,6 +317,20 @@ main([[maybe_unused]] int argc, const char **argv)
...
+  // arguments, which would make an omitted --prefix indistinguishable from an explicit
+  // one. CacheShmCommand supplies the runtime default and keys its "did you set
+  // name_prefix?" hint off the option being absent.
+  shm_command.add_option("--prefix", "-p", "shm name prefix word, framed as /<word>- (default 'ats')", "", 1, "");
+  shm_command.add_command("status", "Show the cache shared-memory control segment and stripe table", [&]() { command->execute(); })
+    .add_example_usage("traffic_ctl cache shm status")
+    .add_example_usage("traffic_ctl cache shm status --prefix ats-t");
+  shm_command.add_command("clear", "Unlink the cache shared-memory control and stripe segments", [&]() { command->execute(); })
+    .add_example_usage("traffic_ctl cache shm clear")
+    .add_example_usage("traffic_ctl cache shm clear --prefix ats-t");
```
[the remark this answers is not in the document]
masaori335: Fixed by [55d5019](https://github.com/apache/trafficserver/pull/13328/commits/55d501947a886d1ba74e184982c9415aaef63481)

bneradt (commented):
I found four issues in the current head (`9c754f2`):

1. **[P1] Fast-attach accepts out-of-stripe `Dir` offsets**

   In `src/iocore/cache/Stripe.cc:218-229`, `_shm_directory_is_valid()` bounds `next` and `prev`, but never validates a live entry’s `dir_offset()` or extent. An out-of-phase entry with a huge offset can pass both this loop and `check_segment()`. `CacheVC::handleRead()` then computes an out-of-stripe `aio_offset`, and its end truncation can convert a negative length to a huge unsigned `aio_nbytes`. Since `clean_shutdown` is set before event threads stop, a torn directory mutation can reach this state without malicious tampering.

   Please validate every nonempty entry’s `vol_offset()` and approximate extent against `[data_lo, data_hi]` before permitting fast attach.

2. **[P2] Cyclic or inconsistent free lists pass validation**

   The same validation loop only bounds free-list links; `Directory::check_segment()` checks bucket chains, not the free list. An in-range cycle such as `A -> B -> A` therefore passes. Subsequent `freelist_pop()` operations can overwrite an already-allocated entry’s tag bits and eventually reset the segment.

   Please walk each free list with a node limit or visited set, verify `prev` consistency, and reject cycles before attaching it.

3. **[P2] Clearing `owner_pid` creates a concurrent-attach window**

   In `src/iocore/cache/CacheShm.cc:881-892`, `mark_clean_shutdown()` clears the PID before the process exits, while event threads may still modify directories. On platforms where POSIX-shm `flock` is unsupported (notably macOS and FreeBSD), another process can observe `clean_shutdown == 1` and `owner_pid == 0` and attach concurrently.

   Please retain the PID until process exit. The next process’s liveness check will naturally see it as dead; while it is still alive, refusing the attach is the correct behavior.

4. **[P2] Clearing a smaller foreign control segment bypasses the live-owner lock**

   In `include/shared/cache_shm/Purge.h:246-253`, `purge_segments()` sweeps and unlinks any control segment smaller than the current build before calling `try_lock_control()` at line 272. A newer `traffic_ctl` can therefore unlink segments belonging to a still-running older build—the exact upgrade case the frozen header is intended to support.

   Please acquire the lock before every unlink path. Where locking is unsupported, inspect the frozen header when possible or fail closed rather than sweeping a potentially live owner’s namespace.

bneradt (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as abfbfcd9e6e82677562478c3d7f0925dd3b3f51e for the 10.2.0 release.

PR #13330, opened 2026-06-25
Annotate BRAVO locks for thread-safety analysis

masaori335:
1. Add annotations to suppress warnings with `-Wthread-safety` in BRAVO implementation. 
2. Cleanups


moonchen (approved):
Thanks for adding these annotations. Confirmed the `token = 0` reset is correct and doesn't affect existing callers.

cmcfarlen:
Cherry-picked to 10.2.x

PR #13340, opened 2026-06-26
[10.2.x] cache: apply per-volume settings on first start after clear (#13252)

masaori335:
Per-volume tuning (ram_cache, ram_cache_size, ram_cache_cutoff, avg_obj_size, fragment_size) was only copied onto the CacheVol when matching an existing on-disk volume, so volumes created fresh (e.g. the first start after a cache clear) ignored the config until the next restart. Apply the settings in one pass after all CacheVols exist.

Extend cache_volume_features to verify the settings take effect via per-volume metrics.

(cherry picked from commit ec72cf41dadf4876be0e83addbb91d8ddf104290)

Conflicts:
  tests/gold_tests/cache/cache_volume_features.replay.yaml


cmcfarlen:
[approve ci]

PR #13342, opened 2026-06-26
Refresh default TLS context on secret update

### LINKED ISSUE #9562: In what cases are updates by TSSslSecretSet() actually used?
I'm running a test with this `ssl_multicert.config` file:
```
dest_ip = 0 ssl_cert_name=2050.crt ssl_key_name=private.key
```
The cert in 2050.crt expired in the year 2050.  The steps of the test are:

1. Execute a `curl` that makes a TLS connection, but does not authenticate the received server cert, with verbose output enabled.
2. Use TSSslSecretSet()/TSSslSecretUpdate() to change the 2050.crt to one that is identical, except for expiring in 2060.
3. Repeat the same `curl` command.

Here is the debug trace output with the debug regex `ssl`: https://gist.github.com/ywkaras/a7b511a290d62627be18ba301653944c
The update of the cert can be seen starting a line 201.  However, the expiration date in the verbose output in the second `curl` does not change.  Presumably this means the new cert is not used.

I notice that, in this function:  https://github.com/apache/trafficserver/blob/de8e2439fdd4dc491bb7af512b8f8603a81a7be1/iocore/net/SSLNetVConnection.cc#L1129
getCTX() is only called in the client case, not the server case.  Does this limit the cases where updates to secrets are actually used?

bneradt:
Default server certificate secret updates could rebuild the TLS
contexts for CN/SAN lookups while leaving the default/no-SNI context
pointing at the old SSL_CTX. Operators could update cert material on
disk and through the secret API, but new handshakes without a more
specific match could still serve the stale certificate.

This updates runtime context refresh to cover address/default lookup
entries owned by the same ssl_multicert policy and retains the
default context while callers create new TLS sessions. This also adds
an AuTest that updates a plugin-loaded default certificate and
verifies the next no-SNI handshake sees the new certificate.

Fixes: #9562

### DIFF
--- a/src/iocore/net/P_SSLCertLookup.h
+++ b/src/iocore/net/P_SSLCertLookup.h
@@ -26,6 +26,8 @@
 #include "iocore/eventsystem/ConfigProcessor.h"
 #include "iocore/net/SSLTypes.h"
 #include "records/RecCore.h"
+#include "tsutil/Bravo.h"
+#include <mutex>
 #include <shared_mutex>
 
 #include <set>
@@ -134,8 +136,7 @@ struct SSLCertLookup : public ConfigInfo {
   std::unique_ptr<SSLContextStorage> ssl_storage;
   std::unique_ptr<SSLContextStorage> ec_storage;
 
-  shared_SSL_CTX ssl_default;
-  bool           is_valid = true;
+  bool is_valid = true;
 
   int insert(const char *name, SSLCertContext const &cc);
   int insert(const IpEndpoint &address, SSLCertContext const &cc);
@@ -146,6 +147,7 @@ struct SSLCertLookup : public ConfigInfo {
       @return @c A pointer to the matched context, @c nullptr if no match is found.
   */
   SSLCertContext *find(const IpEndpoint &address) const;
+  SSLCertContext *find(const IpEndpoint &address, [[maybe_unused]] SSLCertContextType ctxType) const;
 
   /** Find certificate context by name (FQDN).
       Exact matches have priority, then wildcards. Only destination based matches are checked.
@@ -154,10 +156,18 @@ struct SSLCertLookup : public ConfigInfo {
   SSLCertContext *find(const std::string &name, SSLCertContextType ctxType = SSLCertContextType::GENERIC) const;
 
   // Return the last-resort default TLS context if there is no name or address match.
-  SSL_CTX *
+  shared_SSL_CTX
   defaultContext() const
   {
-    return ssl_default.get();
+    ts::bravo::shared_lock<ts::bravo::shared_mutex> lock(default_ctx_mutex);
+    return ssl_default;
+  }
+
+  void
+  setDefaultContext(shared_SSL_CTX ctx) const
+  {
+    std::lock_guard<ts::bravo::shared_mutex> lock(default_ctx_mutex);
+    ssl_default = std::move(ctx);
   }
 
   unsigned        count(SSLCertContextType ctxType = SSLCertContextType::GENERIC) const;
@@ -170,6 +180,9 @@ struct SSLCertLookup : public ConfigInfo {
   ~SSLCertLookup() override;
 
 private:
+  mutable ts::bravo::shared_mutex default_ctx_mutex;
+  mutable shared_SSL_CTX          ssl_default;
+
   // Map cert_secret name to lookup keys
   std::unordered_map<std::string, std::vector<std::string>> cert_secret_registry;
 };
--- a/src/iocore/net/QUICPacketHandler.cc
+++ b/src/iocore/net/QUICPacketHandler.cc
@@ -281,7 +281,8 @@ QUICPacketHandlerIn::_recv_packet(int /* event ATS_UNUSED */, UDPPacket *udp_pac
     QUICConnectionId new_cid;
 
     QUICCertConfig::scoped_config server_cert;
-    SSL                          *ssl = SSL_new(server_cert->defaultContext());
+    auto                          default_ctx = server_cert->defaultContext();
+    SSL                          *ssl         = SSL_new(default_ctx.get());
 
     quiche_conn *quiche_con = quiche_conn_new_with_tls(
       new_cid, new_cid.length(), retry_token.original_dcid(), retry_token.original_dcid().length(), &udp_packet->to.sa,
--- a/src/iocore/net/SSLUtils.cc
+++ b/src/iocore/net/SSLUtils.cc
@@ -1733,6 +1733,27 @@ SSLMultiCertConfigLoader::update_ssl_ctx(const std::string &secret_name)
       if (!ctx) {
         retval = false;
       } else {
+        if ((*policy_iter)->addr) {
+          SSLCertContext *cc = nullptr;
+          if (strcmp((*policy_iter)->addr, "*") == 0) {
+            this->_set_handshake_callbacks(ctx.get());
+            cc = lookup->find("*", loadingctx.ctx_type);
+          } else {
+            IpEndpoint ep;
+            if (ats_ip_pton((*policy_iter)->addr, &ep) == 0) {
+              cc = lookup->find(ep, loadingctx.ctx_type);
+            } else {
+              Error("'%s' is not a valid IPv4 or IPv6 address", (const char *)(*policy_iter)->addr);
+              retval = false;
+            }
+          }
+          if (cc && cc->userconfig.get() == policy_iter->get()) {
+            cc->setCtx(ctx);
+            if (strcmp((*policy_iter)->addr, "*") == 0) {
+              lookup->setDefaultContext(ctx);
+            }
+          }
+        }
         for (auto const &name : common_names) {
           SSLCertContext *cc = lookup->find(name, loadingctx.ctx_type);
           if (cc && cc->userconfig.get() == policy_iter->get()) {
@@ -1787,8 +1808,8 @@ SSLMultiCertConfigLoader::_store_single_ssl_ctx(SSLCertLookup *lookup, const sha
     if (strcmp(sslMultCertSettings->addr, "*") == 0) {
       Dbg(dbg_ctl_ssl_load, "Addr is '*'; setting %p to default", ctx.get());
       if (lookup->insert(sslMultCertSettings->addr, SSLCertContext(ctx, ctx_type, sslMultCertSettings, keyblock)) >= 0) {
-        inserted            = true;
-        lookup->ssl_default = ctx;
+        inserted = true;
+        lookup->setDefaultContext(ctx);
         this->_set_handshake_callbacks(ctx.get());
       }
     } else {
@@ -1890,7 +1911,7 @@ SSLMultiCertConfigLoader::load(SSLCertLookup *lookup, bool firstLoad)
   // We *must* have a default context even if it can't possibly work. The default context is used to
   // bootstrap the SSL handshake so that we can subsequently do the SNI lookup to switch to the real
   // context.
-  if (lookup->ssl_default == nullptr) {
+  if (lookup->defaultContext() == nullptr) {
     shared_SSLMultiCertConfigParams sslMultiCertSettings(new SSLMultiCertConfigParams);
     sslMultiCertSettings->addr = ats_strdup("*");
     if (!this->_store_ssl_ctx(lookup, sslMultiCertSettings)) {
--- a/tests/gold_tests/tls/tls_secret_update_default.test.py
+++ b/tests/gold_tests/tls/tls_secret_update_default.test.py
@@ -1,0 +1,234 @@
+'''
+Test default server certificate updates through TSSslSecretSet/TSSslSecretUpdate.
+'''
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import os
+import re
+
+Test.Summary = '''
+Verify that secret updates refresh the default server SSL_CTX used without SNI.
+'''
+
+
+class TestDefaultSecretUpdate:
+    '''Verify default server certificate updates through the secret API.'''
+
+    initial_cert: str = 'signed-bar.pem'
+    updated_cert: str = 'signed2-bar.pem'
+    key_file: str = 'signed-bar.key'
+    shadowed_cert: str = 'signed-foo.pem'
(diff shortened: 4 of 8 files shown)


src/iocore/net/SSLUtils.cc
```diff
@@ -1733,6 +1733,18 @@ SSLMultiCertConfigLoader::update_ssl_ctx(const std::string &secret_name)
       if (!ctx) {
         retval = false;
       } else {
+        // Address and default-context lookups are owned by the same policy, but
+        // are not part of the certificate CN/SAN name sets below.
+        for (unsigned i = 0; i < lookup->count(loadingctx.ctx_type); ++i) {
```
moonchen: This predicate matches every entry the policy owns, including the unique-SAN entries — everything from one `ssl_multicert` line shares the same `userconfig`. Those get stamped with the common ctx here and only repaired by the `unique_names` loop below, and if that rebuild fails, `init_server_ssl_ctx` returns an empty vector (`goto fail`), so `if (!unique_ctx)` never runs and they permanently keep the wrong cert while `update_ssl_ctx` returns true. Since the goal is just the addr-keyed entry, `find("*", ...)` (or `ats_ip_pton` + `find(ep)` for a non-`*` addr) would update exactly that entry and keep the "old ctx or new ctx, never a wrong one" behavior. The comment above also doesn't match the code — these entries *are* the CN/SAN name sets below.

src/iocore/net/SSLUtils.cc
```diff
@@ -1733,6 +1733,18 @@ SSLMultiCertConfigLoader::update_ssl_ctx(const std::string &secret_name)
...
+          SSLCertContext *cc = lookup->get(i, loadingctx.ctx_type);
+          if (cc && cc->ctx_type == loadingctx.ctx_type && cc->userconfig.get() == policy_iter->get()) {
+            cc->setCtx(ctx);
+          }
+        }
+        if ((*policy_iter)->addr && strcmp((*policy_iter)->addr, "*") == 0) {
+          lookup->setDefaultContext(ctx);
+          this->_set_handshake_callbacks(ctx.get());
```
moonchen: Unlike the load path, this lookup is live, so the new ctx is reachable by net threads (`cc->setCtx` above, `setDefaultContext` here) before the callbacks are installed. `SSL_new` copies `cert_cb`, so a handshake landing in that window runs with no cert selection at all. Installing the callbacks right after the ctx is created, before any `setCtx`, closes the window.

src/iocore/net/QUICPacketHandler.cc
```diff
@@ -281,7 +281,8 @@ QUICPacketHandlerIn::_recv_packet(int /* event ATS_UNUSED */, UDPPacket *udp_pac
     QUICConnectionId new_cid;
 
     QUICCertConfig::scoped_config server_cert;
-    SSL                          *ssl = SSL_new(server_cert->defaultContext());
+    auto                          default_ctx = server_cert->defaultContext();
```
moonchen, line 284: Not for this PR, but worth a note in the description or a follow-up issue: `QUICCertConfig` builds its own `SSLCertLookup` that only a full reload rebuilds, so secret updates still leave QUIC serving the stale default cert.

src/iocore/net/P_SSLCertLookup.h
```diff
@@ -154,10 +154,18 @@ struct SSLCertLookup : public ConfigInfo {
   SSLCertContext *find(const std::string &name, SSLCertContextType ctxType = SSLCertContextType::GENERIC) const;
 
   // Return the last-resort default TLS context if there is no name or address match.
-  SSL_CTX *
+  shared_SSL_CTX
   defaultContext() const
   {
-    return ssl_default.get();
+    std::lock_guard<std::mutex> lock(default_ctx_mutex);
```
moonchen: This needs `#include <mutex>` — the header only includes `<shared_mutex>`, so `std::mutex`/`std::lock_guard` currently compile through transitive includes.

src/iocore/net/P_SSLCertLookup.h
```diff
@@ -134,8 +134,8 @@ struct SSLCertLookup : public ConfigInfo {
   std::unique_ptr<SSLContextStorage> ssl_storage;
   std::unique_ptr<SSLContextStorage> ec_storage;
 
-  shared_SSL_CTX ssl_default;
-  bool           is_valid = true;
+  mutable shared_SSL_CTX ssl_default;
```
moonchen: Can this move to the private section next to `default_ctx_mutex`? Nothing accesses it directly anymore, and keeping it public leaves the door open for the kind of unlocked read this PR fixed in SSLStats.cc.

src/iocore/net/P_SSLCertLookup.h
```diff
@@ -154,10 +154,18 @@ struct SSLCertLookup : public ConfigInfo {
   SSLCertContext *find(const std::string &name, SSLCertContextType ctxType = SSLCertContextType::GENERIC) const;
 
   // Return the last-resort default TLS context if there is no name or address match.
-  SSL_CTX *
+  shared_SSL_CTX
   defaultContext() const
```
moonchen, line 160: One optional suggestion: the default ctx is now stored twice (the `"*"` entry and `ssl_default`) under two different locks, and this one is exclusive on every no-SNI accept. Caching a pointer to the `"*"` entry's `SSLCertContext` and delegating to its `getCtx()`/`setCtx()` (shared_mutex) would drop both the duplicate state and the exclusive lock.

tests/gold_tests/tls/tls_secret_update_default.test.py
```diff
@@ -0,0 +1,151 @@
...
+
+class TestDefaultSecretUpdate:
+    '''Verify default server certificate updates through the secret API.'''
+
+    initial_cert = 'signed-bar.pem'
+    updated_cert = 'signed2-bar.pem'
+    key_file = 'signed-bar.key'
+    plugin = 'ssl_secret_load_test.so'
+
+    def __init__(self):
```
moonchen: Type annotations on these methods, please — matching the other class-style tests in this directory (e.g. `def __init__(self) -> None:` in tls_origin_post_abort.test.py).

tests/gold_tests/tls/tls_secret_update_default.test.py
```diff
@@ -0,0 +1,151 @@
...
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import os
+
+Test.Summary = '''
```
moonchen, line 23: Optional: tls_check_cert_select_plugin.test.py already has the same plugin/wildcard-policy/CopyAs+touch scaffolding — a no-SNI run added there might cover this without a second copy of the update dance.

moonchen (changes requested):
👍 to covering the default/no-SNI context on secret updates. A couple of issues with the new refresh loop, plus some smalls.

tests/gold_tests/tls/tls_secret_update_default.test.py
bneradt: Added

src/iocore/net/SSLUtils.cc
bneradt: Fixed. The refresh now updates only the policy’s exact address entry: find("*", ctx_type) for the wildcard/default policy, or ats_ip_pton followed by the typed address lookup for an explicit address. It no longer temporarily overwrites every entry sharing the policy, so unique-SAN contexts retain either their old context or a successfully rebuilt one.

src/iocore/net/QUICPacketHandler.cc
bneradt: Agreed. This change updates the SSLConfig lookup used by TCP TLS; QUICCertConfig maintains a separate lookup and still requires a full reload. I’m treating QUIC secret-update support as follow-up work rather than expanding this fix into that separate lifecycle.

src/iocore/net/SSLUtils.cc
bneradt: Fixed. For the wildcard/default context, the handshake callbacks are now installed before the context is published through either SSLCertContext::setCtx() or setDefaultContext(), closing the window where a net thread could create an SSL without the certificate callback.

src/iocore/net/P_SSLCertLookup.h
bneradt: Fixed. I added the direct <mutex> include rather than relying on a transitive include.

src/iocore/net/P_SSLCertLookup.h
bneradt: Good suggestion. I kept the existing default-context storage model in this PR to avoid broadening the fix into an ownership refactor. Caching the wildcard SSLCertContext and delegating to its synchronized accessors would remove the duplicate state and is a worthwhile follow-up.

src/iocore/net/P_SSLCertLookup.h
bneradt: Fixed. ssl_default is now private and located next to default_ctx_mutex, so callers must use the synchronized accessor methods.

tests/gold_tests/tls/tls_secret_update_default.test.py
```diff
@@ -0,0 +1,151 @@
...
+        ts.addSSLfile(f'ssl/{self.updated_cert}')
+        ts.addSSLfile(f'ssl/{self.key_file}')
+        Test.PrepareTestPlugin(os.path.join(Test.Variables.AtsTestPluginsDir, self.plugin), ts)
+
+        ts.Disk.records_config.update(
+            {
+                'proxy.config.diags.debug.enabled': 1,
+                'proxy.config.diags.debug.tags': 'ssl_secret_load_test|ssl|ssl_config_updateCTX',
+                'proxy.config.ssl.server.cert.path': f'{ts.Variables.SSLDir}/../',
+                'proxy.config.ssl.server.private_key.path': f'{ts.Variables.SSLDir}/../',
```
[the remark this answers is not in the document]
bneradt: Meh, the ../ is fine and using just SSLDir as you suggests breaks the test.

src/iocore/net/P_SSLCertLookup.h
```diff
@@ -154,10 +155,18 @@ struct SSLCertLookup : public ConfigInfo {
...
   {
-    return ssl_default.get();
+    std::lock_guard<std::mutex> lock(default_ctx_mutex);
+    return ssl_default;
+  }
```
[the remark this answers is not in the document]
bneradt: Makes sense. Using bravo lock though.

tests/gold_tests/tls/tls_secret_update_default.test.py
bneradt: Thanks for the suggestion, but I'd prefer to pass on this change for now at least.

src/iocore/net/SSLUtils.cc
```diff
@@ -1733,6 +1733,27 @@ SSLMultiCertConfigLoader::update_ssl_ctx(const std::string &secret_name)
...
+              Error("'%s' is not a valid IPv4 or IPv6 address", (const char *)(*policy_iter)->addr);
+              retval = false;
+            }
+          }
+          if (cc && cc->userconfig.get() == policy_iter->get()) {
+            cc->setCtx(ctx);
+          }
+          if (strcmp((*policy_iter)->addr, "*") == 0) {
+            lookup->setDefaultContext(ctx);
+          }
```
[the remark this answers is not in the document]
bneradt: Fixed. setDefaultContext() is now gated by the same userconfig ownership check as the active "*" lookup entry. The AuTest now includes a shadowed wildcard policy and verifies that refreshing its secret does not replace the no-SNI default certificate.

tests/gold_tests/tls/tls_secret_update_default.test.py
```diff
@@ -0,0 +1,234 @@
...
+        return tr
+
+    def _add_secret_update_wait_run(self, secret_name: str) -> 'TestRun':
+        '''Wait for the test plugin to update the certificate secret.'''
+        tr = Test.AddTestRun(f'Wait for the {secret_name} secret update')
+        tr.Processes.Default.Command = 'echo awaiting secret update'
+        tr.Processes.Default.ReturnCode = 0
+        await_update = tr.Processes.Process('await_update', 'sleep 30')
+        await_update.Ready = When.FileContains(
+            self._ts.Disk.traffic_out.Name, f'updated cert for secret .*{re.escape(secret_name)}')
```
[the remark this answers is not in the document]
bneradt: Nah, db logs are emitted via traffic_out

tests/tools/plugins/ssl_secret_load_test.cc
```diff
@@ -159,8 +159,8 @@ CB_Update_Secret(TSCont cont, TSEvent /* event ATS_UNUSED */, void * /* edata AT
     }
   }
   for (auto name : updates) {
-    Dbg(dbg_ctl, "update cert for secret %s", name.c_str());
     TSSslSecretUpdate(name.c_str(), name.length());
+    Dbg(dbg_ctl, "updated cert for secret %s", name.c_str());
   }
```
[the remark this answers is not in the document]
bneradt: The plugin now checks TSSslSecretUpdate() and emits the updated cert synchronization marker only on success; failures are reported through TSError.

tests/gold_tests/tls/tls_secret_update_default.test.py
```diff
@@ -0,0 +1,187 @@
...
+        self._keep_processes_running(tr)
+        return tr
+
+    def _add_secret_update_wait_run(self, secret_name: str) -> 'TestRun':
+        '''Wait for the test plugin to update the certificate secret.'''
+        tr = Test.AddTestRun(f'Wait for the {secret_name} secret update')
+        tr.Processes.Default.Command = 'echo awaiting secret update'
+        tr.Processes.Default.ReturnCode = 0
+        await_update = tr.Processes.Process('await_update', 'sleep 30')
+        await_update.Ready = When.FileContains(self._ts.Disk.traffic_out.Name, f'update cert for secret .*{secret_name}')
```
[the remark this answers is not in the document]
bneradt: Fixed. The secret name is now escaped with re.escape() before embedding it in the When.FileContains regular expression.

tests/gold_tests/tls/tls_secret_update_default.test.py
```diff
@@ -0,0 +1,187 @@
...
+            self, name: str, expected_cn: str, unexpected_cn: str, expected_description: str,
+            unexpected_description: str) -> 'TestRun':
+        '''Add a curl run against ATS without SNI.'''
+        tr = Test.AddTestRun(name)
+        self._keep_processes_running(tr)
+        tr.MakeCurlCommand(
+            f"-k -v --http1.1 -H 'host: doesnotmatter' https://127.0.0.1:{self._ts.Variables.ssl_port}/", ts=self._ts)
+        tr.Processes.Default.ReturnCode = 0
+        tr.Processes.Default.Streams.All = Testers.ContainsExpression(f'issuer:.*CN={expected_cn}', expected_description)
+        tr.Processes.Default.Streams.All += Testers.ExcludesExpression(f'issuer:.*CN={unexpected_cn}', unexpected_description)
```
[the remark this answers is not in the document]
bneradt: Fixed. Expected and unexpected issuer CN values are now escaped with re.escape() before constructing the assertion expressions.

moonchen (approved):
Verified that previous findings are addressed. CI is green.

cmcfarlen:
Cherry-picked to 10.2.x

Issue #13347, opened 2026-06-29
Remove remaining deprecated OpenSSL 3.0 APIs in SSLUtils.cc (DH and ENGINE)


bryancall:
Apache Traffic Server already builds, runs, and is CI-tested against OpenSSL 3.x (CI covers OpenSSL 3.0.8 and 3.5.6, plus quictls). On OpenSSL 3.0 and newer the build defines `OPENSSL_API_COMPAT=10002` (see CMakeLists.txt), which keeps a few deprecated calls compiling through the 1.0.2 compatibility layer.

This issue tracks removing that residual deprecated usage so we can eventually drop the compatibility shim and stay clean as these APIs are removed in future OpenSSL releases. It is the remaining work split out from the umbrella #7341, which I am closing because the compatibility goal (build, run, and pass the test suite on OpenSSL 3.x) is met.

All of the remaining deprecated-in-3.0 calls are in `src/iocore/net/SSLUtils.cc`:

Diffie-Hellman parameter helpers (deprecated in favor of the EVP_PKEY interface):
- `DH_get_2048_256()` ([line 389](https://github.com/apache/trafficserver/blob/a2c55e3dbe918b229981074c6c81a92b5534c88f/src/iocore/net/SSLUtils.cc#L389), [line 499](https://github.com/apache/trafficserver/blob/a2c55e3dbe918b229981074c6c81a92b5534c88f/src/iocore/net/SSLUtils.cc#L499))
- `DH_new()` ([line 423](https://github.com/apache/trafficserver/blob/a2c55e3dbe918b229981074c6c81a92b5534c88f/src/iocore/net/SSLUtils.cc#L423))
- `PEM_read_bio_DHparams()` ([line 497](https://github.com/apache/trafficserver/blob/a2c55e3dbe918b229981074c6c81a92b5534c88f/src/iocore/net/SSLUtils.cc#L497))

ENGINE API (deprecated in 3.0 in favor of providers):
- `ENGINE_load_dynamic()` ([line 832](https://github.com/apache/trafficserver/blob/a2c55e3dbe918b229981074c6c81a92b5534c88f/src/iocore/net/SSLUtils.cc#L832))
- `ENGINE_get_default_RSA()` ([line 919](https://github.com/apache/trafficserver/blob/a2c55e3dbe918b229981074c6c81a92b5534c88f/src/iocore/net/SSLUtils.cc#L919))
[...]

PR #13349, opened 2026-06-29
Support OpenSSL 3.0 APIs for Diffie-Hellman

JosiahWI:
This is work for #13347. It does not remove the OpenSSL 1.1.1 compatibility, but it switches the implementation to use OpenSSL 3.x APIs when available.

The change is split into three commits. The first one is a suite of unit tests generated by Claude Opus. They are written to test through a high-level API for now, but target the DH key loading. I reviewed them and fixed some issues in the generated code. The tests use OpenSSL 3.x APIs and are only added to the build if those APIs are available.

The second commit is the OpenSSL 3.x implementation. This is the easiest place to view the meat of the change without distraction.

Finally, I split most of the work here into a new file called SSLKeyUtils.cc in the third commit. SSLUtils.cc needs refactoring according to CodeScene (the file is massive and not cohesive), and I saw an opportunity to make a small chip in it here in that direction, with the special benefit of isolating all the messy precompiler directives in the new file so that the SSLUtils.cc function remains sleek and build-configuration-independent.

When reviewing, please take a look at `EVP_PKEY_CTX_set_dh_rfc5114`. In the OpenSSL 3.x implementation of `gen_dh_2048_258` I would rather use that helper function than manually set params. When I tried it, OpenSSL happily failed the corresponding unit test. Closer docs inspection reveals that the key type has to be `EVP_PK_DHX`, even for the `dh` version of the function. I don't know the difference between DH and DHX and whether it would be behavior-preserving to switch, so I'm asking for help here.

I have been thorough in error-checking and memory management, referring to OpenSSL's API documentation, but I am not familiar with OpenSSL APIs yet. I would appreciate a review from someone with a lot of OpenSSL experience.


src/iocore/net/SSLKeyUtils.cc
```diff
@@ -0,0 +1,186 @@
...
+
+  if (EVP_PKEY_keygen_init(pctx.get()) <= 0) {
+    Error("failed to initialize OpenSSL keygen");
+    return nullptr;
+  }
+
+  char             prime_group[]{"dh_2048_256"};
+  OSSL_PARAM const params[]{OSSL_PARAM_utf8_string("group", prime_group, sizeof(prime_group) - 1), OSSL_PARAM_END};
+
+  if (!EVP_PKEY_CTX_set_params(pctx.get(), params)) {
```
JosiahWI, line 63: This is where I want to use the `EVP_PKEY_CTX_set_dh_rfc5114` function instead of manually setting parameters.

src/iocore/net/SSLKeyUtils.cc
```diff
@@ -0,0 +1,186 @@
...
+
+bool
+set_ctx_dh(SSL_CTX *ctx, dh_key_t *pkey)
+{
+  bool result{SSL_CTX_set_options(ctx, SSL_OP_SINGLE_DH_USE) && SSL_CTX_set0_tmp_dh_pkey(ctx, pkey)};
+  if (!result) {
+    EVP_PKEY_free(pkey);
+  }
+  return result;
+}
```
JosiahWI, line 106: Of special note: the OpenSSL 3.0 API takes ownership of the key, whereas the older API does not. Therefore, the control flow with respect to memory management is different between the two `set_ctx_dh` implementations, which could be surprising.

src/iocore/net/unit_tests/test_SSLDHParams.cc
```diff
@@ -0,0 +1,183 @@
...
+{
+
+std::string
+make_valid_dh_pem()
+{
+  EVP_PKEY_CTX *pctx = EVP_PKEY_CTX_new_from_name(nullptr, "DH", nullptr);
+  REQUIRE(pctx != nullptr);
+  REQUIRE(EVP_PKEY_paramgen_init(pctx) > 0);
+  OSSL_PARAM params[2] = {
+    OSSL_PARAM_construct_utf8_string("group", const_cast<char *>("ffdhe2048"), 0),
```
JosiahWI: Segfault waiting to happen.

src/iocore/net/unit_tests/test_SSLDHParams.cc
```diff
@@ -0,0 +1,186 @@
...
+}
+
+} // namespace
+
+TEST_CASE("ssl_context_enable_dhe: nullptr dhparams file falls back to built-in DH parameters")
+{
+  CHECK(init_with_dhparams(nullptr));
+}
+
+TEST_CASE("ssl_context_enable_dhe: valid ffdhe2048 DH PEM file is accepted")
```
JosiahWI: Name needs to be corrected. I changed the format to dh_2048_256.

src/iocore/net/SSLKeyUtils.cc
```diff
@@ -0,0 +1,186 @@
...
+
+  if (!EVP_PKEY_CTX_set_params(pctx.get(), params)) {
+    Error("SSL dhparams source returned invalid parameters");
+    return nullptr;
+  }
+
+  EVP_PKEY *pkey{};
+  EVP_PKEY_generate(pctx.get(), &pkey);
+
+  return pkey;
```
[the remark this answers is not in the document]
JosiahWI: I think this is nonsense.

maskit (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as a8329a928257e3685e1ca61264b19670775c000a for the 10.2.0 release.

PR #13352, opened 2026-07-01
prefetch: don't drop replacements for non-participating optional capture groups

moonchen:
## Problem

`Pattern::replace()` in the prefetch plugin rejects any replacement reference `$N` whose index is `>=` the value returned by the match. But that value is *one past the highest capture group that **participated*** in the match, not the number of groups the pattern defines. A trailing optional group such as `(\?.*)?` that does not participate — e.g. a request with no query string — lowers that count and makes a perfectly valid `$N` look out of range.

Concretely, with a remap like:

```
@plugin=prefetch.so @pparam=--fetch-path-pattern=/(.*-)(\d+)(\?.*)?$/$1{$2+1}$3/
```

every matching request **without a query string** logs:

```
ERROR: (prefetch) invalid reference in replacement string: $3
```

and the prefetch is silently dropped (the client is still served).

## Fix

- **`replace()`**: at match time, substitute an empty string for a group that did not participate instead of failing the whole replacement — the documented PCRE2 semantics for an unmatched group.
- **`compile()`**: validate `$N` references once, at config-load time, against the pattern's actual capture-group count (`Regex::get_capture_count()`) — the correct place to catch a genuinely out-of-range reference such as `$5` against a 3-group pattern, and a pattern defining more groups than can be captured (`$0..$9`).

## Additional hardening

- Treat an unusable `--fetch-path-pattern`, and invalid `--fetch-count` / `--fetch-max` / `--fetch-overflow` values, as **configuration errors**: the remap instance fails to load instead of silently running with prefetch disabled. On a running server, `traffic_ctl config reload` safely rejects the bad config and keeps the current one.
- Skip an empty expanded path instead of scheduling a zero-length fetch (which `BgFetch` turns into a self-prefetch of the original request path).
- Remove the now-dead `Pattern::process()` / `Pattern::capture()` helpers (no callers).

> **Note for reviewers:** the "refuse to load on invalid config" behavior change is intentional and aligns with how `--fetch-policy` already behaves, but it *is* a behavior change — happy to split it into a separate PR if preferred.

## Tests

Adds three autests under `tests/gold_tests/pluginTest/prefetch/`:
- `prefetch_optional_group` — the non-participating optional-group case (fails on the pre-fix code).
- `prefetch_bad_pattern_refused` — an over-limit (10-group) pattern makes ATS refuse to load the remap.
- `prefetch_empty_replacement` — an empty-collapsing replacement is logged and skipped, not self-prefetched.

Built locally (macOS, PCRE2); all prefetch autests pass. Draft pending CI.

### DIFF
--- a/plugins/prefetch/configs.cc
+++ b/plugins/prefetch/configs.cc
@@ -21,6 +21,8 @@
  * @brief Plugin configuration.
  */
 
+#include <charconv>  /* std::from_chars() */
+#include <cstring>   /* strlen() */
 #include <fstream>   /* std::ifstream */
 #include <getopt.h>  /* getopt_long() */
 #include <sstream>   /* std::istringstream */
@@ -90,20 +92,25 @@ PrefetchConfig::setFetchOverflow(const char *optarg)
 }
 
 /**
- * @brief Whether @a optarg is a non-empty string of decimal digits (a valid unsigned integer option).
+ * @brief Parses @a optarg as an unsigned integer option value.
+ * @param optarg the option value to parse.
+ * @param value set to the parsed value on success, untouched on failure.
+ * @return true if @a optarg is a non-empty string of decimal digits that fits in @a value.
  */
 static bool
-isUnsignedInt(const char *optarg)
+parseUnsignedInt(const char *optarg, unsigned &value)
 {
-  if (nullptr == optarg || '\0' == *optarg) {
+  if (nullptr == optarg) {
     return false;
   }
-  for (const char *p = optarg; '\0' != *p; ++p) {
-    if (*p < '0' || *p > '9') {
-      return false;
-    }
-  }
-  return true;
+
+  const char *const end = optarg + strlen(optarg);
+  auto const [parsed, ec]{std::from_chars(optarg, end, value)};
+
+  /* from_chars() reports a leading sign or a non-digit as invalid_argument and a value too large
+   * for @a value as result_out_of_range. Requiring it to consume the whole string rejects trailing
+   * characters such as "10abc". */
+  return std::errc{} == ec && parsed == end;
 }
 
 /**
@@ -172,8 +179,8 @@ PrefetchConfig::init(int argc, char *argv[])
       break;
 
     case 'c': /* --fetch-count */
-      if (isUnsignedInt(optarg)) {
-        setFetchCount(optarg);
+      if (unsigned count = 0; parseUnsignedInt(optarg, count)) {
+        setFetchCount(count);
       } else {
         PrefetchError("invalid --fetch-count '%s': expected a non-negative integer", optarg ? optarg : "");
         status = false;
@@ -199,8 +206,8 @@ PrefetchConfig::init(int argc, char *argv[])
     } break;
 
     case 'x': /* --fetch-max */
-      if (isUnsignedInt(optarg)) {
-        setFetchMax(optarg);
+      if (unsigned max = 0; parseUnsignedInt(optarg, max)) {
+        setFetchMax(max);
       } else {
         PrefetchError("invalid --fetch-max '%s': expected a non-negative integer", optarg ? optarg : "");
         status = false;
--- a/plugins/prefetch/configs.cc
+++ b/plugins/prefetch/configs.cc
@@ -71,14 +71,39 @@ iequals(const StringView lhs, const StringView rhs)
                     [](const char a, const char b) { return tolower(a) == tolower(b); });
 }
 
-void
+bool
 PrefetchConfig::setFetchOverflow(const char *optarg)
 {
-  if (StringView("64") == optarg) {
+  if (nullptr == optarg) {
+    return false;
+  }
+  if (StringView("32") == optarg) {
+    _fetchOverflow = EvalPolicy::Overflow32;
+  } else if (StringView("64") == optarg) {
     _fetchOverflow = EvalPolicy::Overflow64;
   } else if (iequals("bignum", optarg)) {
     _fetchOverflow = EvalPolicy::Bignum;
+  } else {
+    return false;
   }
+  return true;
+}
+
+/**
+ * @brief Whether @a optarg is a non-empty string of decimal digits (a valid unsigned integer option).
+ */
+static bool
+isUnsignedInt(const char *optarg)
+{
+  if (nullptr == optarg || '\0' == *optarg) {
+    return false;
+  }
+  for (const char *p = optarg; '\0' != *p; ++p) {
+    if (*p < '0' || *p > '9') {
+      return false;
+    }
+  }
+  return true;
 }
 
 /**
@@ -147,7 +172,12 @@ PrefetchConfig::init(int argc, char *argv[])
       break;
 
     case 'c': /* --fetch-count */
-      setFetchCount(optarg);
+      if (isUnsignedInt(optarg)) {
+        setFetchCount(optarg);
+      } else {
+        PrefetchError("invalid --fetch-count '%s': expected a non-negative integer", optarg ? optarg : "");
+        status = false;
+      }
       break;
 
     case 'e': /* --fetch-path-pattern */ {
@@ -156,7 +186,10 @@ PrefetchConfig::init(int argc, char *argv[])
         if (pattern->init(optarg)) {
           _nextPaths.add(std::move(pattern));
         } else {
-          PrefetchError("failed to initialize next object pattern");
+          /* An unusable fetch-path-pattern is a configuration error; fail instance creation so ATS
+           * refuses to load the remap rule rather than silently running with prefetch disabled. */
+          PrefetchError("failed to initialize fetch-path-pattern '%s'", optarg ? optarg : "");
+          status = false;
         }
       }
     } break;
@@ -166,11 +199,19 @@ PrefetchConfig::init(int argc, char *argv[])
     } break;
 
     case 'x': /* --fetch-max */
-      setFetchMax(optarg);
+      if (isUnsignedInt(optarg)) {
+        setFetchMax(optarg);
+      } else {
+        PrefetchError("invalid --fetch-max '%s': expected a non-negative integer", optarg ? optarg : "");
+        status = false;
+      }
       break;
 
     case 'o': /* --fetch-overflow */
-      setFetchOverflow(optarg);
+      if (!setFetchOverflow(optarg)) {
+        PrefetchError("invalid --fetch-overflow '%s': expected 32, 64, or bignum", optarg ? optarg : "");
+        status = false;
+      }
       break;
 
     case 'r': /* --replace-host */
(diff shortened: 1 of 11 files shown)


plugins/prefetch/configs.cc
```diff
@@ -71,14 +71,39 @@ iequals(const StringView lhs, const StringView rhs)
...
+/**
+ * @brief Whether @a optarg is a non-empty string of decimal digits (a valid unsigned integer option).
+ */
+static bool
+isUnsignedInt(const char *optarg)
+{
+  if (nullptr == optarg || '\0' == *optarg) {
+    return false;
+  }
+  for (const char *p = optarg; '\0' != *p; ++p) {
```
serrislew: nitpicky but does this need any range-checking?

plugins/prefetch/pattern.cc
```diff
@@ -253,22 +181,18 @@ Pattern::replace(const String &subject, String &result)
...
-    std::string_view dst       = matches[replIndex];
+    int replIndex = _tokens[i];
 
-    String src(_replacement, _tokenOffset[i], 2);
+    /* $replIndex was validated at config-load time against the number of groups the pattern defines, but
+     * the group may still not have participated in *this* match (e.g. a trailing optional group such as
+     * "(\?.*)?" when the subject has no query string).  pcre2_match() returns one past the highest
+     * participating group, so substitute an empty string for a group at or beyond that -- the documented
+     * PCRE2 semantics for an unmatched group -- rather than failing the whole replacement. */
+    std::string_view dst = (replIndex < matchCount) ? matches[replIndex] : std::string_view{};
```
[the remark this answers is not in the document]
moonchen: Confirmed against PCRE2 10.47: for `(a)?(b)` on subject `b`, the match count is 3 and group 1 is `PCRE2_UNSET`, so `replIndex < matchCount` passes and the subject pointer is advanced by an unset offset.

Two corrections. This is not introduced here — master reaches `matches[replIndex]` under the identical `_tokens[i] >= matchCount` precondition, so the exposure is the same before and after this PR. And it does not crash: the length is `end - start` with both unset, so the view is zero-length and callers see an empty string. The pointer is invalid, which is worth fixing, but nothing observably breaks.

The real defect is in `RegexMatches::operator[]`, which every ovector caller shares (`regex_remap`, `cachekey`, `regex_revalidate`). Fixing it there rather than in this plugin: #13441.

plugins/prefetch/configs.cc
moonchen: Good catch.

Switched both `--fetch-count` and `--fetch-max` to `std::from_chars()` parsing directly into the destination `unsigned`, which reports out-of-range and trailing junk in the same call. Added `prefetch_bad_count_refused` to cover it; it fails against the previous code.

plugins/prefetch/plugin.cc
```diff
@@ -634,6 +634,14 @@ contHandleFetch(const TSCont contp, TSEvent event, void *edata)
               String expandedPath;
 
               if (config.getNextPath().replace(workingPath, expandedPath)) {
+                if (expandedPath.empty()) {
+                  /* A replacement that collapses to empty (e.g. every referenced group was optional and
+                   * absent) would otherwise be scheduled with a zero-length path, which BgFetch skips --
+                   * leaving the original request path in place and prefetching the pristine URL itself.
+                   * Stop rather than issue that self-prefetch. */
+                  PrefetchError("prefetch pattern produced an empty path; check the fetch-path-pattern replacement");
```
serrislew: If this is per transaction, will this flood our logs with this error?

serrislew:
since validation behavior changes, does this need a doc update?

plugins/prefetch/pattern.cc
```diff
@@ -253,22 +181,18 @@ Pattern::replace(const String &subject, String &result)
...
-      PrefetchError("invalid reference in replacement string: $%d", _tokens[i]);
-      return false;
-    }
-  }
-
   int previous = 0;
   for (int i = 0; i < _tokenCount; i++) {
-    int              replIndex = _tokens[i];
-    std::string_view dst       = matches[replIndex];
+    int replIndex = _tokens[i];
```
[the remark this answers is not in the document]
moonchen: Fixed. The view is now built from `""` so `data()` is never null, and the append uses the `std::string_view` overload rather than the pointer/length pair, so neither the `"%.*s"` nor the append passes a null pointer.

plugins/prefetch/pattern.cc
```diff
@@ -253,22 +181,18 @@ Pattern::replace(const String &subject, String &result)
...
+     * "(\?.*)?" when the subject has no query string).  pcre2_match() returns one past the highest
+     * participating group, so substitute an empty string for a group at or beyond that -- the documented
+     * PCRE2 semantics for an unmatched group -- rather than failing the whole replacement. */
+    std::string_view dst = (replIndex < matchCount) ? matches[replIndex] : std::string_view{};
 
-    PrefetchDebug("replacing '%s' with '%.*s'", src.c_str(), static_cast<int>(dst.length()), dst.data());
+    PrefetchDebug("replacing '$%d' with '%.*s'", replIndex, static_cast<int>(dst.length()), dst.data());
 
     result.append(_replacement, previous, _tokenOffset[i] - previous);
     result.append(dst.data(), dst.length());
```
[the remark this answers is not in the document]
moonchen: Fixed. The view is now built from `""` so `data()` is never null, and the append uses the `std::string_view` overload rather than the pointer/length pair, so neither the `"%.*s"` nor the append passes a null pointer.

plugins/prefetch/plugin.cc
moonchen: Good catch, and worse than per transaction: whether the replacement collapses depends on the request, so for a misconfigured pattern it recurs indefinitely rather than being a one-time complaint at load.

`PrefetchConfig::shouldReportEmptyPath()` now reports it at error level once per remap instance and drops to the debug tag afterwards. A config reload builds a new instance, so a genuine misconfiguration stays visible after each reload.

moonchen:
> since validation behavior changes, does this need a doc update?

Yes, added. There is now a paragraph after the parameter reference in `doc/admin-guide/plugins/prefetch.en.rst` stating that an invalid parameter value is a configuration error: `--fetch-count` and `--fetch-max` must be decimal numbers that fit in an unsigned integer, `--fetch-overflow` must be `32` or `64`, and `--fetch-path-pattern` must compile and may only reference capture groups the pattern defines. It also says the remap rule fails to load rather than loading with prefetch silently disabled, and that `traffic_ctl config reload` rejects such a configuration and keeps the running one.

Rebased onto master while here, which picked up #71d40c137c (`Constrain prefetch relative paths`). That reworks the `--fetch-query` relative-path branch, so it does not overlap the `--fetch-path-pattern` branch this PR touches. All nine prefetch autests pass locally, including the `prefetch_query_path_traversal` test that came with it.

cmcfarlen:
The core diagnosis is right and I verified the pieces it rests on. `PCRE2_INFO_CAPTURECOUNT` does report the groups the pattern *defines* (excluding group 0), so validating `$N` against `Regex::get_capture_count()` at compile time is the correct bound, and `_tokens[i] > captureCount` is the right comparison — `$3` against a 3-group pattern is legal, `$4` is not. Moving that check out of the per-match path into `compile()` is the real fix: the old code was comparing a config-time constant against a per-request value, which is why a request without a query string could invalidate a `$3` that is perfectly valid for the pattern.

The hardening is solid too. `std::errc{} == ec && parsed == end` is the right `from_chars()` idiom — it rejects trailing junk and out-of-range in one call, which a `strtoul`-style check usually misses. And `shouldReportEmptyPath()` via `_reportedEmptyPath.exchange(true, relaxed)` correctly makes the empty-path complaint once per instance rather than per transaction, which matters here precisely because whether the replacement collapses is request-dependent.

**One thing I think is still open, and it is an ordering hazard with #13441.**

The null-pointer guard covers only one of the two ways `dst` can end up empty:

```cpp
std::string_view dst = (replIndex < matchCount) ? matches[replIndex] : std::string_view{""};
```

The `""` protects the `replIndex >= matchCount` branch — the trailing-optional-group case this PR is about. But your own PCRE2 analysis in the thread above establishes the other case: for `(a)?(b)` on subject `b`, `matchCount` is 3 and group 1 is `PCRE2_UNSET`, so `replIndex < matchCount` holds and the *first* branch is taken.

Today that yields a garbage pointer with zero length. Once #13441 lands, `RegexMatches::operator[]` returns `std::string_view()` for an unset group — its unit test asserts `matches[1].data() == nullptr` explicitly. So after #13441, this line hands a genuine `nullptr` to:

```cpp
PrefetchDebug("replacing '$%d' with '%.*s'", replIndex, static_cast<int>(dst.length()), dst.data());
result.append(dst);
```

which is the exact `%.*s`-with-null concern raised at pattern.cc:186/198. The reply there says the view is now built from `""` so `data()` is never null; that holds for the branch it was aimed at, but not for this one. Benign on the libcs we build against, still undefined, and #13441 turns it from "invalid pointer" into "null pointer" — the case implementations are least likely to tolerate.

It is reachable with an ordinary pattern, not just a contrived one: any optional group that precedes a participating group, e.g. `--fetch-path-pattern=/(v\d+/)?(.*-)(\d+)$/$1$2{$3+1}/` on a request without the version prefix.

Normalizing regardless of which branch produced the view closes it and makes the guard say what the comment already claims:

```cpp
std::string_view dst = (replIndex < matchCount) ? matches[replIndex] : std::string_view{};
if (nullptr == dst.data()) {
  dst = std::string_view{""};
}
```

That also means this PR stops depending on #13441's choice of return value either way, which seems worth having given both are in flight — right now the correctness of this line is coupled to a decision made in another PR.

Two smaller notes:

- Dropping `Pattern::process()` and `Pattern::capture()` is fine — I confirmed no remaining callers. Worth keeping in the description since removing public-looking plugin methods tends to raise eyebrows.
- The "refuse to load on invalid config" change is the right call and matching `--fetch-policy`'s existing behavior is a good argument, but it is a genuine behavior change: an operator whose `--fetch-count` was silently ignored now fails the remap. You already flagged it and offered to split; I would keep it here, since shipping the validation without the enforcement would leave the bad config running. Just make sure it lands in a release note.

Nothing else from me — with the null normalization I would be happy with this.

cmcfarlen (approved):

moonchen:
Holding this until #13517 merges.

As it stands here, `replace()` still takes the `matches[replIndex]` branch for a non-participating group, which returns a null `data()` since #13441 landed. #13517 fixes that in `RegexMatches::operator[]` itself, after which this collapses to a plain `matches[replIndex]` with no guard needed. Rebasing once it's in.

cmcfarlen:
Cherry-picked to the 10.2.x branch as 2fe4afb523d447b9ec190ba5f4154b526aefc9ac for the 10.2.0 release.

PR #13360, opened 2026-07-03
traffic_crashlog: emit a well-formed report when the backtrace is empty

moonchen:
#### Problem
`ServerBacktrace()` can succeed with an empty backtrace — the target's thread list was unreadable, e.g. a fast-aborting target already exited before the forked helper attached. The success check only tests for a null trace, so the crash log ends up with no backtrace section and the existing fallbacks (in-process backtrace, diagnostic message) never run.

#### Fix
Require a non-empty trace before taking the success path, so the empty case falls through to the fallbacks; free the trace when falling through.

Easy to hit with any startup `Fatal` (e.g. a listen-port bind failure): the process is gone before the helper attaches.

### DIFF
--- a/src/traffic_crashlog/traffic_crashlog.cc
+++ b/src/traffic_crashlog/traffic_crashlog.cc
@@ -145,14 +145,18 @@ crashlog_write_backtrace(FILE *fp, const crashlog_target &target)
   // can also happen without a debugger. Possibly in that case, there is a race with the
   // kernel locking the process information?
 
-  if (mgmterr == 0 && trace != nullptr) {
+  if (mgmterr == 0 && trace != nullptr && trace[0] != '\0') {
     // ServerBacktrace succeeded - this gives us backtraces for all threads.
     fprintf(fp, "%s", trace);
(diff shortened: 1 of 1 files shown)


JosiahWI (approved):

src/traffic_crashlog/traffic_crashlog.cc
```diff
@@ -145,14 +145,18 @@ crashlog_write_backtrace(FILE *fp, const crashlog_target &target)
...
   // kernel locking the process information?
 
-  if (mgmterr == 0 && trace != nullptr) {
+  if (mgmterr == 0 && trace != nullptr && trace[0] != '\0') {
     // ServerBacktrace succeeded - this gives us backtraces for all threads.
     fprintf(fp, "%s", trace);
     free(trace);
     return true;
   }
+  free(trace);
```
JosiahWI, line 154: This is a use case for `free` accepting `nullptr`. Interesting!

cmcfarlen:
Cherry-picked to 10.2.x

PR #12397, opened 2025-07-28
Retry Connect with Exponential Backoff

masaori335:
When connect attempt to the origin server is failed, ATS retries immediately. In some case, this pile up connections to the origin servers and hard to make it back. This change introduce a new config to do exponential backoff on retry.

PR #13365, opened 2026-07-06
10.1.x: USDT: normalize names for STATE_ENTER

moonchen:
Backport of #13346.

`STATE_ENTER` passed `&HttpSM::<state>` straight through as the USDT probe
name, so probes were emitted as e.g. `&HttpSM::state_read_client_request_header`.
The `&` and `::` characters are illegal in USDT probe names: bpftrace and perf
treat `:` as the provider/name field separator and reject `&`, so these probes
could not be attached by name.

Pass the bare handler name from each call site and add the prefix in the
macro, producing valid names like `state_enter_state_read_client_request_header`
that share a common `state_enter_*` wildcard. Also fixes a `&HttpSM :` typo
that had collapsed one probe to a bare `&HttpSM`, and a call site that passed
the wrong label (`tunnel_handler_plugin_client` from `tunnel_handler_plugin_agent`)
so the probe name matches its handler.

Did not cherry-pick cleanly:
- `state_http_server_open` conflicted with the exponential-backoff retry
  logic from #12397, which isn't in 10.1.x; resolved by keeping 10.1.x's
  version of that function.
- Two call sites that 10.1.x still carries but master no longer has
  (`state_pre_resolve`, `state_mark_os_down`) also needed converting to the
  new macro syntax, or the file wouldn't compile.

Verified `traffic_server` builds clean on 10.1.x with this applied.

### DIFF
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -199,7 +199,7 @@ initialize_thread_for_connecting_pools(EThread *thread)
   {                                                                                      \
     /*ink_assert (magic == HTTP_SM_MAGIC_ALIVE); */ REMEMBER(event, reentrancy_count);   \
     SMDbg(dbg_ctl_http, "[%s, %s]", #state_name, HttpDebugNames::get_event_name(event)); \
-    ATS_PROBE1(state_name, sm_id);                                                       \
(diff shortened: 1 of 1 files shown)


moonchen:
@cmcfarlen

PR #12648, opened 2025-11-06
ATSReplayTest: autests via replay.yaml

bneradt:
For context, this continues the work started in #12371. The goal is to take full advantage of the autest extension capabilities to make writing ATS autests easier.

This adds the Test.ATSReplayTest autest extension so that tests can be written through a replay.yaml file that specifies both the configuration of the test processes as well as the HTTP Proxy Verifier traffic for the test.

This patch updates the header_rewrite_bundle to use this new framework.

PR #13366, opened 2026-07-06
10.1.x: Fix set-status crash inside if/endif at remap time

moonchen:
Backport of #13052.

Operators inside an if/endif block never had set_hook() called on them, so
_hook retained its default value of TS_HTTP_READ_RESPONSE_HDR_HOOK. When
set-status executed at REMAP_PSEUDO_HOOK time, exec() took the
response-hook branch and called TSHttpHdrStatusSet on a request buffer,
corrupting the header union and crashing.

Fix the operator to check TSHttpHdrTypeGet() at runtime instead of trusting
get_hook(). Removed _hook and get_hook() from operators.

set_hook() becomes is_hook_valid() in Statement and RuleSet to only
validate hook compatibility without mutating _hook.

Upgrade the HTTPHdr::status_set/reason_set asserts to release_assert so
that writing response fields into a request buffer is caught immediately
rather than silently corrupting memory.

Did not cherry-pick cleanly: three master-only refactors not in 10.1.x
touch the same lines (enum class HTTPType in #12249, string_view header
accessors in #12243, and the txn/setter overloads for
TSHttpHdrStatusSet/TSHttpTxnStatusSet in #12484). Resolved using 10.1.x's
existing API surface; the logic change is otherwise identical to master.

Not included: the accompanying autest, since the bundle-style replay.yaml
test infrastructure it relies on (#12648) isn't in 10.1.x.

Verified header_rewrite.so and traffic_server build clean on 10.1.x with
this applied.

### DIFF
--- a/include/proxy/hdrs/HTTP.h
+++ b/include/proxy/hdrs/HTTP.h
@@ -1091,7 +1091,7 @@ inline void
 HTTPHdr::status_set(HTTPStatus status)
 {
   ink_assert(valid());
-  ink_assert(m_http->m_polarity == HTTP_TYPE_RESPONSE);
+  ink_release_assert(m_http->m_polarity == HTTP_TYPE_RESPONSE);
 
   http_hdr_status_set(m_http, status);
 }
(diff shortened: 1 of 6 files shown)


moonchen:
@cmcfarlen

PR #13379, opened 2026-07-13
Preserve client port in transaction logs

bneradt:
Fixes behavior introduced via #13123

---

Client source ports could be logged as zero when the live inbound
connection was cleared before transaction log marshalling. This
obscured the effective client endpoint in access logs.

This reads the port from the effective client endpoint retained by
HttpSM, keeping it available after connection cleanup.


moonchen (approved):

cmcfarlen:
Cherry-picked to 10.2.x

cmcfarlen:
Cherry-picked to the 10.1.x branch as 158a9e5b0cccf91579be26985015f792f521dcea for the 10.1.5 release.

PR #13380, opened 2026-07-14
Honor the RamCache copy flag in LRU and S3-FIFO

phongn:
## Summary

`put(..., copy = true)` is the caller's contract that buffers may not be shared between the cache and the caller in either direction. `CacheVC` requests it whenever RAM cache compression is configured (`http_copy_hdr`, `CacheVC.cc`), because it unmarshals HTTP headers in place in its own buffer *after* the put, and again on every RAM-cache hit.

The LRU and S3-FIFO RAM caches ignored the flag and shared buffers both ways. With `proxy.config.cache.ram_cache.compress` set and `algorithm = 1` (the default) or `2`, put stores the caller's buffer and CacheVC's subsequent in-place unmarshal silently mutates the cached bytes, so every later RAM hit hands out a buffer the cache no longer controls and two CacheVCs can hold the same live `HdrHeap`.

**What that breaks today is narrower than this PR originally claimed.** `HTTPInfo::unmarshal` returns early on an already-unmarshaled heap ("Already unmarshaled, must be a ram cache", `HTTP.cc`), so under default settings the mutation is masked. With `proxy.config.cache.enable_checksum` on, the recheck in `handleReadDone` walks the mutated bytes and marks the doc `DOC_CORRUPT` on every RAM hit after the first. The earlier description said the re-unmarshal asserts; it does not.

The `http_copy_hdr` gate keys off the compress config only, never the algorithm, so this is reachable purely by configuration. This is a contract fix and a backport candidate.

## Changes

- The copy contract is implemented once, as `RamCache::copy_data_in()` / `RamCache::copy_data_out()` on the base class; CLFUS's put/get sites use them (no behavior change there).
- `RamCacheLRU` and `RamCacheS3FIFO` honor `copy` the way CLFUS always has: copy on put, copy on get. A `copy = true` put refreshes a resident entry that is still sharing a caller's buffer; a put onto an entry that already holds a private copy leaves it alone (the concurrent-miss case). CLFUS gets the same guard, narrowed to uncompressed private entries since for a compressed one the swap is what decompresses it.
- `RamCacheLRU` records the data length on its entry, as CLFUS and S3-FIFO already did; CLFUS also refreshes `e->len` on a resident re-put.
- `copy = false` behavior is unchanged: zero-copy sharing, pinned by a test.
- Catch2 test `test_RamCacheCopy`, eight cases across all three policies: caller mutation after put and after get, the shared-to-private refresh, `copy=false` sharing, byte accounting across the refresh (gauges *and* the internal counters eviction runs off), S3-FIFO's ghost readmit, and the already-private no-op.

## Notes for reviewers

- **`ram_cache.compress` changes at runtime despite `RECU_RESTART_TS`.** The record is marked restart-required, but `cache_config_ram_cache_compress` is linked with `RecEstablishStaticConfigInt32`, and the records reload path fires update callbacks regardless of update type (the type is only reported back so `traffic_ctl` can say a restart is needed). Two consequences. The shared-to-private refresh is reachable: a reload landing between two racing misses' `handleReadDone` calls puts a `copy=true` put onto a `copy=false` entry. And a **1→0 reload is a hazard this PR extends to LRU/S3-FIFO**: private copies hold marshaled header bytes, but `CacheVC::handleRead` decides whether to unmarshal a RAM hit from the *live* config, so after the reload it hands up marshaled bytes and `get_handle` fails until the entry is evicted. CLFUS has always had this. The fix is in CacheVC (decide from the buffer's state, not the config) and is a follow-up.
- **Cost.** LRU and S3-FIFO cannot compress, so with `compress` set they now pay copy-in on every header-bearing HTTP fill and copy-out on every hit, under the stripe mutex, for a compression they cannot perform. Gating `http_copy_hdr` on the algorithm in CacheVC would avoid that; it is left as a follow-up because LRU compression is planned and will need exactly this copy support.
- `copy = true` now dereferences `len` for two more policies. `doc->len` is not bounded against the read size in `handleReadDone` (pre-existing for CLFUS); a bound there is a follow-up.
- The refresh needs no eviction pass: a private copy is an exact-size allocation while a shared buffer was charged its rounded `block_size()`, and same key and auxkey name one on-disk doc, so `len` does not grow. Recorded in a comment at both sites.
- S3-FIFO's ghost-readmit and stale-auxkey paths go through the single insert site, so they inherit the copy handling.
- The test keeps constructed caches reachable rather than destroying them: the policies have no destructors (entries are pool-allocated), so destroying a cache object would strand its entries under leak checkers.

## Follow-ups (separate PRs)

Done: #13669 applied the LRU seen filter to new keys only. Without it the filter ran before the resident-entry lookup and could decline the copy refresh below; this branch is rebased on top of it.

- CacheVC: gate `http_copy_hdr` on the algorithm, and decide RAM-hit unmarshalling from the buffer rather than the live config.
- CacheVC: bound `doc->len` against the read size before the put.
- CLFUS `get()`: the decompress-on-hit path charges `compressed_len` where it now holds `len` bytes.

### DIFF
--- a/src/iocore/cache/RamCacheLRU.cc
+++ b/src/iocore/cache/RamCacheLRU.cc
@@ -33,6 +33,7 @@
 struct RamCacheLRUEntry {
   CryptoHash key;
   uint64_t   auxkey;
+  bool       copy; // copy-in-copy-out: buffers are never shared with callers
   LINK(RamCacheLRUEntry, lru_link);
   LINK(RamCacheLRUEntry, hash_link);
   Ptr<IOBufferData> data;
@@ -147,7 +148,11 @@ RamCacheLRU::get(CryptoHash *key, Ptr<IOBufferData> *ret_data, uint64_t auxkey)
     if (e->key == *key && e->auxkey == auxkey) {
       lru.remove(e);
       lru.enqueue(e);
-      (*ret_data) = e->data;
+      if (e->copy) {
+        (*ret_data) = copy_data_out(e->data.get(), e->data->block_size());
+      } else {
+        (*ret_data) = e->data;
+      }
       DDbg(dbg_ctl_ram_cache, "get %X %" PRIu64 " HIT", key->slice32(3), auxkey);
       ts::Metrics::Counter::increment(cache_rsb.ram_cache_hits);
       ts::Metrics::Counter::increment(stripe->cache_vol->vol_rsb.ram_cache_hits);
@@ -181,9 +186,8 @@ RamCacheLRU::remove(RamCacheLRUEntry *e)
   return ret;
 }
 
-// ignore 'copy' since we don't touch the data
 int
-RamCacheLRU::put(CryptoHash *key, IOBufferData *data, [[maybe_unused]] uint32_t len, bool, uint64_t auxkey)
+RamCacheLRU::put(CryptoHash *key, IOBufferData *data, uint32_t len, bool copy, uint64_t auxkey)
 {
   if (!max_bytes) {
     return 0;
@@ -213,6 +217,18 @@ RamCacheLRU::put(CryptoHash *key, IOBufferData *data, [[maybe_unused]] uint32_t 
       if (e->auxkey == auxkey) {
         lru.remove(e);
         lru.enqueue(e);
+        if (copy) {
+          // The entry may still be sharing a caller's buffer from a put made
+          // while copy semantics were not requested; refresh it with a
+          // private copy before the caller mutates its buffer.
+          int64_t delta = static_cast<int64_t>(len) - e->data->block_size();
+
+          e->data  = copy_data_in(data, len);
+          e->copy  = true;
+          bytes   += delta;
+          ts::Metrics::Gauge::increment(cache_rsb.ram_cache_bytes, delta);
+          ts::Metrics::Gauge::increment(stripe->cache_vol->vol_rsb.ram_cache_bytes, delta);
+        }
         return 1;
       } else { // discard when aux keys conflict
         e = remove(e);
@@ -224,13 +240,18 @@ RamCacheLRU::put(CryptoHash *key, IOBufferData *data, [[maybe_unused]] uint32_t 
   e         = THREAD_ALLOC(ramCacheLRUEntryAllocator, this_ethread());
   e->key    = *key;
   e->auxkey = auxkey;
-  e->data   = data;
+  e->copy   = copy;
+  if (copy) {
+    e->data = copy_data_in(data, len);
+  } else {
+    e->data = data;
+  }
   bucket[i].push(e);
   lru.enqueue(e);
-  bytes += ENTRY_OVERHEAD + data->block_size();
+  bytes += ENTRY_OVERHEAD + e->data->block_size();
   objects++;
-  ts::Metrics::Gauge::increment(cache_rsb.ram_cache_bytes, ENTRY_OVERHEAD + data->block_size());
-  ts::Metrics::Gauge::increment(stripe->cache_vol->vol_rsb.ram_cache_bytes, ENTRY_OVERHEAD + data->block_size());
+  ts::Metrics::Gauge::increment(cache_rsb.ram_cache_bytes, ENTRY_OVERHEAD + e->data->block_size());
+  ts::Metrics::Gauge::increment(stripe->cache_vol->vol_rsb.ram_cache_bytes, ENTRY_OVERHEAD + e->data->block_size());
   while (bytes > max_bytes) {
     RamCacheLRUEntry *ee = lru.dequeue();
     if (ee) {
--- a/src/iocore/cache/RamCacheLRU.cc
+++ b/src/iocore/cache/RamCacheLRU.cc
@@ -221,6 +221,12 @@ RamCacheLRU::put(CryptoHash *key, IOBufferData *data, uint32_t len, bool copy, u
           // The entry may still be sharing a caller's buffer from a put made
           // while copy semantics were not requested; refresh it with a
           // private copy before the caller mutates its buffer.
+          //
+          // A refresh only ever gives bytes back, so unlike an insert it needs
+          // no eviction pass: the private copy is charged its exact length
+          // while a shared buffer was charged the rounded block_size() (always
+          // >= len), and a re-put under the same key and auxkey names the same
+          // on-disk doc, so len itself does not grow.
           int64_t delta = static_cast<int64_t>(len) - e->data->block_size();
 
           e->data  = copy_data_in(data, len);
--- a/src/iocore/cache/RamCacheLRU.cc
+++ b/src/iocore/cache/RamCacheLRU.cc
@@ -33,6 +33,7 @@
 struct RamCacheLRUEntry {
   CryptoHash key;
   uint64_t   auxkey;
+  uint32_t   len;  // data length; block_size() only bounds it, so copy entries need it recorded
   bool       copy; // copy-in-copy-out: buffers are never shared with callers
   LINK(RamCacheLRUEntry, lru_link);
   LINK(RamCacheLRUEntry, hash_link);
@@ -149,7 +150,7 @@ RamCacheLRU::get(CryptoHash *key, Ptr<IOBufferData> *ret_data, uint64_t auxkey)
       lru.remove(e);
       lru.enqueue(e);
       if (e->copy) {
-        (*ret_data) = copy_data_out(e->data.get(), e->data->block_size());
+        (*ret_data) = copy_data_out(e->data.get(), e->len);
       } else {
         (*ret_data) = e->data;
       }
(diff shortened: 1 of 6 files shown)


phongn:
Tracking issue: #13381

moonchen:
Read through this and ran the tests locally. The bug reproduces as described: `http_copy_hdr` at `CacheVC.cc:403` is computed from `cache_config_ram_cache_compress` and the doc properties with no reference to the algorithm, and `RecordsConfig.cc:863` defaults `ram_cache.algorithm` to 1, so the default policy was the one ignoring the flag.

I checked a couple of other cases:

Tested on a 5000 byte object to exercise the size rounding in `iobuffer_size_to_index()`. The tests still pass. Worth using a non-power-of-two payload anyway, so the tests do not depend on 8192 landing exactly on a size index.

Checked the `ram_cache_bytes` gauge across the resident refresh: N objects put with `copy=false`, re-put with `copy=true`, then the cache overflowed to force eviction. The gauge rises on the puts and does not go negative through eviction, for all three policies. That path is the only place the fix does arithmetic rather than just swapping the buffer, so it is worth a permanent test.

I also recommend an S3-FIFO ghost admit test. The copy handling is inherited -- the ghost is removed and the fresh entry goes through the same insert site -- but the readmit lands in `SEG_MAIN` and accounts against `_m_bytes`, and with `copy` set the charged size is `len` rather than `data->block_size()`. That combination is not covered.

Happy to approve once it comes out of draft.

phongn:
Thanks for the careful read — all three are in, and the first turned out to be load-bearing rather than hygiene.

**Non-power-of-two payload.** Switched to 5000 bytes, with a helper deriving the rounding difference instead of hardcoding it. You were right that the tests shouldn't depend on 8192 landing on a size index: at 8192 every size computation this fix adds is degenerate. `size = copy ? len : data->block_size()` returns the same value on both arms, and `delta` is 0 in both refresh paths. Confirmed by mutation — changing LRU's refresh to charge `data->block_size()` instead of `len` now fails (`66560 == 41024`), and at 8192 that same mutation passes silently.

**`ram_cache_bytes` across the refresh.** Added for all three policies, asserting the exact rounding given back per object rather than a range. It's uniform because CLFUS turns out to share the same refresh arithmetic (`delta = size - e->size`), so all three give back exactly `block_size(5000) - 5000` per object. It also checks that the per-volume gauge tracks the global one, and that neither goes negative through the eviction flood.

Chasing that turned up something the gauge alone cannot see: the gauge and the internal counter that admission and eviction actually run off (`bytes`, `_s_bytes`/`_m_bytes`) are updated independently, so deleting the counter update while keeping the gauge passed every assertion. Covered both ways now. For S3-FIFO `size()` derives from those counters, so it is checked against the gauge at a point where no ghosts exist yet. For LRU `size()` recomputes from the entry list with a different per-entry formula and structurally cannot witness the counter, so that one is behavioral: fill the cache, refresh every entry, then insert more objects than there was room for beforehand — they fit only if the refresh really returned the bytes, and nothing resident is evicted if it did.

**S3-FIFO ghost admit.** Added. I wanted to be sure it wasn't quietly degenerating into a plain insert, so I instrumented the branch temporarily and confirmed it fires exactly once with `copy=1 len=5000 old_size=8192 seg=2` (`SEG_GHOST`) — so it does cross from a `block_size()`-charged entry to a `len`-charged one, which was your point. It also floods a second time afterwards and checks the object survives, as evidence the readmit landed in main rather than small.

Two things fell out of this that are worth flagging:

- LRU was the only policy storing no data length, recovering it from `e->data->block_size()` on get. That is correct only because `copy_data_in()` happens to allocate with `new_xmalloc_IOBufferData`, an invariant maintained in a different file with nothing checking it. It now stores `len` like CLFUS and S3-FIFO, and the refresh delta is derived from the two allocations, so neither the length handed back nor the accounting depends on how `copy_data_in()` allocates.
- The description was wrong about the resident-refresh path. I had justified it as handling a configuration change, but `ram_cache.compress` is `RECU_RESTART_TS`, so `cache_config_ram_cache_compress` cannot change within a process, and `http_copy_hdr`'s other inputs (`doc_type`, `hlen`) are fixed for a given doc — so the copy flag cannot flip for an entry that is already resident. Both `fixup()` callsites are evacuations that relocate a doc byte-identically, so they do not flip it either. I've kept the path, since the contract is per-put, but rewrote the description to call it defensive, and noted the consequence: `delta` is never positive, which is why the refresh needs no eviction pass. That reasoning is now a comment at both sites.

src/iocore/cache/RamCacheLRU.cc
```diff
@@ -213,6 +218,26 @@ RamCacheLRU::put(CryptoHash *key, IOBufferData *data, [[maybe_unused]] uint32_t
       if (e->auxkey == auxkey) {
         lru.remove(e);
         lru.enqueue(e);
+        if (copy) {
```
moonchen, line 221: Please guard this with `copy && !e->copy`.

When two requests both get a RAM-cache miss, both read the same object from disk, and then both try to fill the RAM cache, whichever finishes first inserts a private copy. The other request finds that entry resident and currently allocates and copies the entire object again. Both requests allocate a RAM-cache buffer, but only the first allocation is necessary.

If `e->copy` is already true, the stored buffer is private and `get()` never exposes it directly, so replacing it is unnecessary. The guard preserves the shared-to-private transition while avoiding the redundant allocation and memcpy. The same applies to the corresponding resident-refresh block in S3-FIFO.

moonchen (changes requested):
Please address the inline comment.

src/iocore/cache/RamCacheLRU.cc
phongn: Done in f3286882a7 — LRU and S3-FIFO both guard on `copy && !e->copy` now, so the refresh is reserved for the shared-to-private transition. Thanks; this race also corrects my framing from the last round. I'd described the refresh branch as unreachable, but that's true only of the shared-to-private *transition*. The branch itself fires on exactly the concurrent-miss case you describe, which is why the guard has real value.

I also applied it to CLFUS, since its pre-existing resident path has the same redundant copy — it replaces `e->data` unconditionally on every re-put, both arms. Under the default seen filter it's actually the most exposed of the three: CLFUS gates its filter on new entries only and S3-FIFO has none, so both hit the refresh on the second racing put, while LRU filters *before* the resident walk and toggles per slot, so it needs a fourth.

The CLFUS guard is narrower, though: `copy && e->flag_bits.copy && !e->flag_bits.compressed`. The plain form is unsafe there. A compressed entry holds `compressed_len` bytes behind `flag_bits.copy`, and the resident path clears `flag_bits.compressed` after the swap — skipping the swap while still clearing the flag would have `get()` memcpy `e->len` bytes out of a `compressed_len`-byte buffer (`compressed_len < e->len` is the compressor's admission condition). So a compressed entry still takes the swap, which is what decompresses it. The narrow guard is exact rather than approximate: every path that produces an uncompressed private entry sets `size == len` (Linsert; the compressor's no-gain branch; `get()`'s in-place decompress is `!copy`-only), so the `delta` it skips is 0 and the flag writes it skips are no-ops. Only the allocation and memcpy go away.

One honest caveat: the `!compressed` term isn't covered by a unit test. It needs a compressed entry, which requires the background compressor on `ET_TASK`, and that isn't reachable from the Catch2 harness. It rests on the reasoning above. In the race itself the entry is freshly inserted and the compressor hasn't run, so the term is there for correctness rather than for the common path.

The guard is invisible to the gauge — a private copy is already charged `len`, so `delta` is 0 either way — so the new test pins it by re-putting *different* bytes under the same key and auxkey and checking the cache kept the first, for all three policies. Real callers never do that (same key and auxkey name one on-disk doc), but it's the only observable. It also confirms the shared-to-private refresh still happens, so the guard is on the entry's state and not on how many puts it has seen. Disabling any one of the three guards fails that test for its policy.

phongn:
Two corrections to things I said earlier in this PR, both from a closer read after your guard comment. I've rewritten the description accordingly.

**`ram_cache.compress` is not fixed for the life of the process.** I'd claimed the shared-to-private refresh was unreachable because the record is `RECU_RESTART_TS`. That marking is advisory: `cache_config_ram_cache_compress` is linked via `RecEstablishStaticConfigInt32`, and `RecExecConfigUpdateCbs` fires every pending callback regardless of update type -- the type is only returned so `traffic_ctl` can report that a restart is needed. So a reload changes the value live. That makes the refresh reachable (a reload landing between two racing misses). One thing that surfaced from that and that I've deliberately left out of this PR: LRU runs its seen filter before the resident-entry lookup, so it can decline exactly that refresh (CLFUS gates its filter on new entries only). That's a pre-existing LRU behavior rather than part of the copy contract, so it will come as its own PR. It also means a 1→0 reload leaves LRU/S3-FIFO holding marshaled private copies that `handleRead` no longer routes through the unmarshal -- CLFUS has always had that exposure, this PR extends it, and the right fix (decide from the buffer's state, not the config) is in CacheVC. Called out in the description as a follow-up.

**The failure mode was overstated.** The description said the double unmarshal asserts. `HTTPInfo::unmarshal` returns early on an already-unmarshaled heap ("must be a ram cache"), so under default config the shared-buffer mutation is masked. The observable failure is with `enable_checksum` on, where the recheck in `handleReadDone` walks the mutated bytes and marks the doc corrupt on every RAM hit after the first. The contract violation is real either way and the fix stands; the severity framing did not.

Also in the two new commits: the CLFUS resident re-put now refreshes `e->len` alongside `e->data` (defensive, matching the other two), `copy_data_in` allocates through `IOBufferData::alloc` instead of setting `_mem_type` by hand, and the test keys now spread across hash buckets -- they all landed in bucket 0 before. Two pre-existing items surfaced that I've left for separate PRs: `doc->len` isn't bounded against the read size before the put, and CLFUS's decompress-on-hit path charges `compressed_len` for an entry that now holds `len` bytes.

moonchen (dismissed):
Thanks for addressing the review. Verified the guards and the two corrections locally; tests pass ASAN-clean. LGTM.

src/iocore/cache/RamCacheLRU.cc
```diff
@@ -230,16 +235,60 @@ RamCacheLRU::put(CryptoHash *key, IOBufferData *data, [[maybe_unused]] uint32_t
     }
   }
 
+  RamCacheLRUEntry *e = bucket[i].head;
```
moonchen: The rebase left two resident-entry loops. This one redeclares `e` (`error: redeclaration of 'RamCacheLRUEntry* e'`), and with that fixed it's still unreachable: the loop above from #13669 returns 1 on a resident hit, so the `copy && !e->copy` refresh never runs and 4 of the 8 `test_RamCacheCopy` cases fail. Move the refresh into the first loop and drop this one. With that change the RAM cache tests pass locally.

moonchen (changes requested):
The rebase over #13669 broke `RamCacheLRU::put`; see inline. Also, the LRU seen-filter item in the description's follow-ups has landed as #13669.

moonchen (approved):
Change looks good. Thanks for fixing the merge.

PR #13388, opened 2026-07-15
Hard-enforce max_active_streams_in at HTTP/2 stream creation

maskit:
proxy.config.http2.max_active_streams_in previously only adjusted the advertised SETTINGS_MAX_CONCURRENT_STREAMS, leaving the proxy unable to bound buffered response memory against clients that open streams faster than the advisory throttle reacts. It now also carries a finite default so the cap is no longer effectively unlimited.

The new knob proxy.config.http2.max_active_streams_policy_in selects enforcement. Value 0 keeps the advisory behavior that lowers advertised concurrency to min_concurrent_streams_in. Value 1 refuses new inbound streams with REFUSED_STREAM once the process-wide active-stream count reaches the limit and leaves the advertised value untouched. Refusal happens after HPACK decoding so the dynamic table stays in sync and the connection survives; max_active_streams_exceeded_in counts each refusal.

max_active_streams_in now defaults to 200000 rather than 0.

(cherry picked from commit c76a7c81220fca009ff4dc7de91647c3cf6d72d7)

### DIFF
--- a/doc/admin-guide/files/records.config.en.rst
+++ b/doc/admin-guide/files/records.config.en.rst
@@ -4180,14 +4180,43 @@ HTTP/2 Configuration
    This is used when :ts:cv:`proxy.config.http2.max_active_streams_in` is set
    larger than ``0``.
 
-.. ts:cv:: CONFIG proxy.config.http2.max_active_streams_in INT 0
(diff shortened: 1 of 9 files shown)


ezelkow1:
[approve ci]

ezelkow1 (approved):

PR #13394, opened 2026-07-16
reject over-long integers in xpack_decode_integer

dxbjavid:
xpack_decode_integer accumulates the continuation bytes of an HPACK/QPACK integer by shifting each 7-bit group left by an amount that grows seven per byte and is never bounded, so a header field carrying a long run of continuation bytes drives that shift width past 64. Once it does the guard on the preceding line that is meant to reject over-large encodings shifts by the same out-of-range amount and quietly stops rejecting anything, which is undefined behaviour on the wire-facing path shared by both the HTTP/2 and HTTP/3 header decoders. The loop now stops as soon as the shift width reaches the width of the accumulator so an over-long encoding is treated as a compression error, and I have added a decoding case to test_XPACK.cc that fails without the change.

### DIFF
--- a/src/proxy/hdrs/XPACK.cc
+++ b/src/proxy/hdrs/XPACK.cc
@@ -78,7 +78,7 @@ xpack_decode_integer(uint64_t &dst, const uint8_t *buf_start, const uint8_t *buf
       }
 
       uint64_t added_value = *p & 0x7f;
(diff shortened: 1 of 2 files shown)


JosiahWI:
We are actually working on this issue internally. Thank you for the contribution, but we will be posting a fix for this soon.

PR #13398, opened 2026-07-16
Avoid Diags lock for syslog output

### LINKED ISSUE #7374: Avoid locks in DiagsError when target is syslog
Various thread wide warn/note messages cause excessive lock contention.

I note from:
include/tscore/Diags.h
src/tscore/Diags.cc
https://github.com/apache/trafficserver/blob/fef47d7919e0f87e4f650dfb26742c3a0482091b/src/tscore/Diags.cc#L250

That note() and other messages have an explicit lock. This is a problem on busy systems when there are:
* Many threads
* Messages that cause all thread to log simultaneously occur.

In many cases, the types of messages that occur simultaneously are due to error or load events. Ironically this is made worse by the current structure. Change here may improve Diag::Log overhead/behaviour also...

Such messages include:
* Log disks full
* NOTE: Failure threshold met failcount

A lock should not be required when we know the target is syslog (possibly other types also).
A few suggestions:
* Move the syslog conditional to the head of the logic
* Apply conditionals to the lock semantics
* Move the locks into the existing conditionals

bneradt:
Syslog-only diagnostics take the same mutex used to serialize file
output, adding avoidable contention when many threads report a shared
failure.

This limits the diagnostics mutex to FILE-backed destinations and
removes the obsolete FreeBSD exception because syslog provides its own
thread safety.

Fixes: #7374

---

I have this running on a box in production now and it seems to be running well.

### DIFF
--- a/src/tscore/Diags.cc
+++ b/src/tscore/Diags.cc
@@ -238,8 +238,16 @@ Diags::print_va(const char *debug_tag, DiagsLevel diags_level, const SourceLocat
   // now, finally, output the message //
   //////////////////////////////////////
 
-  lock();
-  if (config.outputs[diags_level].to_diagslog) {
+  auto const &output = config.outputs[diags_level];
+
+  // FILE output must be serialized, but syslog provides its own thread safety.
+  bool const serialize_file_output = output.to_diagslog || output.to_stdout || output.to_stderr || regression_testing_on;
+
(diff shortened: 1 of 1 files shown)


JosiahWI (approved):
There are existing race conditions in this part of the diags code I think, but I don't think this change adds any new races beyond what already existed.

PR #13403, opened 2026-07-18
Remove unused RemapProcessor.h include from RemapPlugins.h

blurman-ai:
RemapPlugins.h includes RemapProcessor.h, and RemapProcessor.h includes RemapPlugins.h back, forming an include cycle. RemapPlugins.h does not use RemapProcessor: the class derives from Continuation and its members are HttpTransact::State, URL and HTTPHdr, none of which come from RemapProcessor.h.

Everything RemapPlugins.h needs already arrives through its other includes (EventSystem.h, HttpTransact.h, RemapPluginInfo.h), which are also the only things RemapProcessor.h contributed to the include closure. Removing the include breaks the cycle with no call-site changes.

### DIFF
--- a/include/proxy/http/remap/RemapPlugins.h
+++ b/include/proxy/http/remap/RemapPlugins.h
@@ -26,7 +26,6 @@
 
 #include "tscore/ink_platform.h"
 #include "iocore/eventsystem/EventSystem.h"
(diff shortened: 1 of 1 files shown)


JosiahWI (approved):
Thank you for the fix.

PR #13405, opened 2026-07-18
Cache: fix cached-header HdrHeap growth on repeated 304 revalidation

moonchen:
## Problem

A hot, frequently-revalidated cached object slowly grows the `HdrHeap` of its
**cached response header** without bound. Once the header's marshalled size
crosses the aggregation-buffer fragment limit (`AGG_SIZE`, 4 MiB), the cache
write of the updated header can never be admitted again
(`Stripe::add_writer` rejects `agg_len > AGG_SIZE` → `AIO_SOFT_FAILURE` →
`openWriteCloseHeadDone` bails before `dir_overwrite`). The stored object is
then **frozen**: its headers never update, it is always stale, and every request
re-runs the doomed revalidation write + serve-stale path.

The growth is invisible to `fields_count()` — the *live* field count stays
constant. It shows up only in `HdrHeap::marshal_length()` (the value that gates
the cache write): dead `MIMEField`/field-block objects accumulate in the
bump-allocated object area.

## Root cause

Two independent contributors, both in the 304 cache-update path:

1. **`merge_and_update_headers_for_cache_update`** unconditionally deletes
   `Age`/`ETag`/`Expires` from the cached header before re-merging the 304's
   values. A deleted `MIMEField` slot is only reclaimed when its entire 16-slot
   block becomes empty, and `HdrHeap::allocate_obj` never reuses an EMPTY slot —
   it only bump-allocates. So each revalidation of an object whose 304 carries a
   changing `Expires` strands one dead slot. `HdrHeap::marshal` is a raw region
   `memcpy` (it does not compact), so the dead slots are persisted to disk, read
   back, and re-marshalled larger on the next revalidation — unbounded growth.

2. **`merge_response_header_with_cached_header`** used a sticky `dups_seen` flag:
   once *any* response field had a duplicate, every *later* single-valued field
   took the `field_create`/`field_attach` **append** path instead of an in-place
   `value_set`, stranding a slot per merge for those fields too.

## Fix

1. In `merge_and_update_headers_for_cache_update`, delete a caching header only
   when the server response **omits** it (guarded on `presence()`). When the 304
   carries it, the merge overwrites the cached field in place — no delete +
   recreate, so no stranded slot. Final header state is unchanged.

2. Rewrite `merge_response_header_with_cached_header` to reconcile each field
   name **independently** at its dup-list head: overwrite existing cached values
   in place (reusing the slot via `field_value_set`), append only genuine extra
   values, and delete any surplus cached values. This makes the merge idempotent
   — re-merging an unchanged response mutates existing fields instead of
   stranding dead ones — and removes the sticky-`dups_seen` behavior. It also
   preserves the documented `Warning`/`Content-Type` handling.

## Tests

`test_HttpTransact.cc` gains a "Repeated revalidation does not grow the cached
header heap" section asserting `HdrHeap::marshal_length()` stays bounded across
up to 4000 merges for: no-dup, leading-dup, response-only-dup, and
changing-length-value cases; plus a correctness case (dropping a duplicate value
leaves exactly the response's values) and an end-to-end case that mirrors the
caller's conditional-delete sequence over the cooked WKS caching headers
(`Cache-Control`/`Expires`/`Age`), checking the cooked `max-age` and bounded
heap. All existing merge assertions are unchanged.

Verified locally: full `test_http` passes under `dev`, `dev-asan`
(ASAN+UBSAN, no errors), and `release`.

## Performance

Because the merge no longer deletes-then-recreates (and the caller skips
unnecessary deletes), the hot path is **faster**, not slower. Measured with a
local A/B microbenchmark (not included in this PR) that timed the new merge
against a copy of the previous implementation head-to-head in one process, on a
realistic ~14-field cached 200 + a 304 (release build, ns/op, best of 5, 3
runs):

```
merge only:
  typical 304 (no dup):  ~ -2% to -8%   (in-place replace, ~neutral)
  dup 304 (Cache-Ctrl):  ~ -27% to -29% (no more append+delete churn)
full revalidation unit (caller delete + merge):
  typical 304 (no dup):  ~ -18% to -21% (caller skips redundant deletes)
  dup 304 (Cache-Ctrl):  ~ -26% to -28%
```

## Compatibility / risk

- The resulting cached header field set is equivalent to the previous code for
  realistic 304/200 responses (names, values, dup count, dup value order). Only
  the physical ordering of *distinct* field names can differ when appending
  extra dup values, which is not significant (RFC 7230 §3.2.2) and no cached-
  response consumer depends on it.
- One corner-case divergence, which is a fix: once a duplicated field was seen,
  the old code's second sweep also deleted cached fields for names the copy loop
  never re-adds (`Set-Cookie`, `Warning`, hop-by-hop headers, `Content-Length`,
  `Transfer-Encoding`; only `Content-Type` was exempted). The new code never
  touches those cached fields, restoring `merge_warning_header`'s documented
  precondition that the cached `Warning` codes reach it untouched.
- This prevents the bloat from forming. It does **not** shrink or unfreeze an
  object that is *already* bloated on disk: an in-place update cannot reclaim
  existing dead object-area bytes (only a fresh full store / copy does). Objects
  already wedged need to leave the cache (purge, full 200 re-store, or eviction).

## Notes for reviewers

- A broader hardening, not done here: give the `HdrHeap` object area a
  compaction valve analogous to the existing string-heap `coalesce_str_heaps`
  (which already exists precisely to bound "infinite build up of dead strings on
  header merge", see the `INKqa08287` comment in `HdrHeap::allocate_str`). The
  object area has no such valve today.

### DIFF
--- a/src/proxy/http/HttpTransact.cc
+++ b/src/proxy/http/HttpTransact.cc
@@ -5035,9 +5035,26 @@ HttpTransact::merge_and_update_headers_for_cache_update(State *s)
   // 10.3.5), but RFC 7232) is clear that the 304 and 200 responses
   // must be identical (see section 4.1). This code attempts to strike
   // a balance between the two.
-  cached_hdr->field_delete(static_cast<std::string_view>(MIME_FIELD_AGE));
-  cached_hdr->field_delete(static_cast<std::string_view>(MIME_FIELD_ETAG));
-  cached_hdr->field_delete(static_cast<std::string_view>(MIME_FIELD_EXPIRES));
+  //
+  // Only delete a caching header that the server response does NOT carry.
+  // If the response carries it, the merge below overwrites the cached field
+  // in place (value_set reuses the existing slot). Deleting it here first
+  // would force the merge to recreate the field, and because a deleted
+  // MIMEField slot is never reused until its whole block is empty, a
+  // long-lived cached object that is revalidated repeatedly (e.g. a hot
+  // max-age=60 object whose Expires changes every 304) would accrete a dead
+  // slot per revalidation until its header heap crosses the aggregation
+  // fragment limit and the cache write can no longer commit (frozen stale).
+  HTTPHdr *server_response = &s->hdr_info.server_response;
+  if (!server_response->presence(MIME_PRESENCE_AGE)) {
+    cached_hdr->field_delete(static_cast<std::string_view>(MIME_FIELD_AGE));
+  }
+  if (!server_response->presence(MIME_PRESENCE_ETAG)) {
+    cached_hdr->field_delete(static_cast<std::string_view>(MIME_FIELD_ETAG));
+  }
+  if (!server_response->presence(MIME_PRESENCE_EXPIRES)) {
+    cached_hdr->field_delete(static_cast<std::string_view>(MIME_FIELD_EXPIRES));
+  }
 
   merge_response_header_with_cached_header(cached_hdr, &s->hdr_info.server_response);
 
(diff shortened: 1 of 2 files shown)


bneradt (approved):
Looks reasonable to me. I had gpt-5.6-sol review also and it had no concerning findings.

cmcfarlen:
Cherry-picked to 10.2.x

PR #13406, opened 2026-07-19
Deliver VCONN_CLOSE for parked TLS hooks; fix SNI queue accounting

moonchen:
Rebased onto master after 508c1bea26 and 7c0dfb03d5 landed. Those already fixed the inverted sweep
loop condition and the `max_age` expiry detach, so this PR no longer contains either. What remains is
a core hook-dispatch bug and the queued-connection half of the rate_limit accounting, which depends
on it.

**A connection that closes while parked in a TLS handshake hook never reaches the plugin's close
hook.** `TLSEventSupport::callHooks()` moves the hook state to `DONE` when a connection closes, but it
kept `curHook` pointing into whichever handshake hook list the connection was parked in. Each hook id
owns a separate list, so advancing `curHook` walked the handshake list rather than the close list: the
close event was dropped once that list ran out, and delivered to the next handshake plugin when it did
not. The fix restarts from the head of the close list unless we are already iterating it, and routes
`TS_EVENT_VCONN_OUTBOUND_CLOSE` through the same path, which previously invoked nothing at all for a
connection parked in the outbound pre-handshake hook.

In the rate_limit SNI queue the consequence is a freed `TSVConn` left on the queue and a leaked
selector lease, after which the next sweep reenables freed memory.

**A connection that closes while still queued gives back a slot it never took.** A connection parked
at the ClientHello hook holds no reservation; the sweep reserves the slot only when it resumes the
connection. The `VCONN_CLOSE` handler nonetheless called `limiter->free()` unconditionally, so a close
in that state decremented `_active` with no matching `reserve()`. Once the counter wraps, the next
`reserve()` trips `TSReleaseAssert(_active <= _limit)` and the server aborts. `RateLimiter::remove()`
now dequeues the connection and reports whether it was queued, and the handler frees a slot only when
it was not. Dequeuing also drops what would otherwise be a stale entry.

These land together because neither is complete alone: the plugin fix is unreachable until close
delivery works, and the core fix on its own makes the unmatched `free()` reachable and reintroduces
the abort. The plugin commits come first so no bisect point has the core fix without its prerequisite.

The sweep takes no lock. `reserve()`, `free()`, `pop()` and `remove()` are each internally
synchronized, and the one composite failure, a slot reserved when the queue turns out to be empty,
hands the slot straight back. An earlier revision of this PR serialized the sweep against close under
a plugin-global mutex; that is gone.

**Tests.** `tls_hooks_close_while_parked` drives the core fix with the existing `ssl_hook_test.so`,
using a delayed ClientHello hook and a close hook plus a 1s handshake timeout that fires inside the 2s
park. Reverting the core fix makes it fail with `iterated to curHook=0x0` and no close callback. Three
rate_limit SNI autests cover queue-then-resume, `max_age` expiry and reject teardown.
`rate_limit_sni_reject` is new coverage rather than a regression test for this change.
`rate_limit_sni_expiry` overlaps `rate_limit_sni.test.py` in scenario, but asserts positively that
expiry ran and reads `traffic.out`, where `ink_abort` writes.

**Not tested.** The outbound close path is reasoned from the code, not exercised. No config-reload,
selector-teardown or concurrent-load coverage. `RateLimiter::remove()` is an O(queue-depth) scan on
every rate-limited close, measured by microbenchmark at 3ns for an empty queue, 270ns at depth 100 and
2.2us at depth 1000, but not measured under load. The rate_limit autests need coreutils `timeout` and
fail rather than skip without it.

### DIFF
--- a/plugins/experimental/rate_limit/limiter.h
+++ b/plugins/experimental/rate_limit/limiter.h
@@ -327,6 +327,25 @@ public:
     return item;
   }
 
+  // Remove a still-queued element (e.g. a connection that closed before it was resumed).
+  // Returns true if it was found in the queue, so the caller can tell a queued element
+  // (which never reserved a slot) from one that was already resumed.
+  bool
+  remove(T elem)
+  {
+    std::lock_guard<std::mutex> lock(_queue_lock);
+
+    for (auto it = _queue.begin(); it != _queue.end(); ++it) {
+      if (std::get<0>(*it) == elem) {
+        _queue.erase(it);
+        --_size;
+        return true;
+      }
+    }
+
+    return false;
+  }
+
   void
   incrementMetric(uint metric)
   {
--- a/plugins/experimental/rate_limit/limiter.h
+++ b/plugins/experimental/rate_limit/limiter.h
@@ -330,8 +330,13 @@ public:
   // Remove a still-queued element (e.g. a connection that closed before it was resumed).
   // Returns true if it was found in the queue, so the caller can tell a queued element
   // (which never reserved a slot) from one that was already resumed.
+  //
+  // Linear in the queue depth, and only reached when an element closes while queued, which
+  // requires the limiter to be at its limit. The depth is bounded by the configured queue size;
+  // note that a "queue" without a "size" leaves _max_queue at UINT32_MAX, in which case the only
+  // bound is proxy.config.net.connections_throttle.
   bool
-  remove(T elem)
+  remove(const T &elem)
   {
     std::lock_guard<std::mutex> lock(_queue_lock);
 
--- a/plugins/experimental/rate_limit/sni_selector.cc
+++ b/plugins/experimental/rate_limit/sni_selector.cc
@@ -219,9 +219,17 @@ sni_queue_cont(TSCont cont, TSEvent /* event ATS_UNUSED */, void * /* edata ATS_
     QueueTime now     = std::chrono::system_clock::now(); // Only do this once per limiter
 
     if (owner) { // Don't operate on the aliases
-      // Try to enable some queued VCs (if any) if there are slots available
+      // Try to enable some queued VCs (if any) if there are slots available. Reserving before
+      // dequeuing means a resumed VC owns the slot it was granted, so its VCONN_CLOSE releases
+      // exactly that slot.
       while (limiter->size() > 0 && limiter->reserve() == ReserveStatus::RESERVED) {
-        auto [vc, contp, start_time]    = limiter->pop();
+        auto [vc, contp, start_time] = limiter->pop();
+
+        if (nullptr == vc) { // A concurrent close emptied the queue; give the slot back
+          limiter->free();
+          break;
+        }
+
         std::chrono::milliseconds delay = std::chrono::duration_cast<std::chrono::milliseconds>(now - start_time);
 
         (void)contp; // Ugly, but silences some compilers.
@@ -236,11 +244,18 @@ sni_queue_cont(TSCont cont, TSEvent /* event ATS_UNUSED */, void * /* edata ATS_
 
         while (limiter->size() > 0 && limiter->hasOldEntity(now)) {
           // The oldest object on the queue is too old on the queue, so "kill" it.
-          auto [vc, contp, start_time]  = limiter->pop();
+          auto [vc, contp, start_time] = limiter->pop();
+
+          if (nullptr == vc) { // A concurrent close emptied the queue
+            break;
+          }
+
           std::chrono::milliseconds age = std::chrono::duration_cast<std::chrono::milliseconds>(now - start_time);
 
           (void)contp;
           Dbg(dbg_ctl, "Queued VC is too old (%ldms), erroring out", static_cast<long>(age.count()));
+          // This VC never reserved a slot; detach it (clear the arg and release the selector
+          // lease) so its VCONN_CLOSE does not release a slot it never held.
           TSUserArgSet(vc, gVCIdx, nullptr);
           limiter->selector()->release();
           TSVConnReenableEx(vc, TS_EVENT_ERROR);
--- a/plugins/experimental/rate_limit/sni_limiter.cc
+++ b/plugins/experimental/rate_limit/sni_limiter.cc
@@ -171,7 +171,12 @@ sni_limit_cont(TSCont contp, TSEvent event, void *edata)
 
     if (limiter) {
       TSUserArgSet(vc, gVCIdx, nullptr);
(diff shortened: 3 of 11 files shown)


plugins/experimental/rate_limit/sni_selector.cc
```diff
@@ -22,6 +22,13 @@
 
 std::atomic<SniSelector *> SniSelector::_instance = nullptr;
 
+// The VC user-arg index, defined in sni_limiter.cc, used to detach an expired queued VC.
+extern int gVCIdx;
+
+// Shared lock (defined in sni_limiter.cc) serializing the queue/slot transactions below
+// against the net-thread VCONN_CLOSE handler.
+extern std::mutex gQueueMutex;
```
cmcfarlen: Could the lock be per-limiter instead of global?

plugins/experimental/rate_limit/limiter.h
```diff
@@ -327,6 +327,25 @@ template <class T> class RateLimiter
...
 
+  // Remove a still-queued element (e.g. a connection that closed before it was resumed).
+  // Returns true if it was found in the queue, so the caller can tell a queued element
+  // (which never reserved a slot) from one that was already resumed.
+  bool
+  remove(T elem)
+  {
+    std::lock_guard<std::mutex> lock(_queue_lock);
+
+    for (auto it = _queue.begin(); it != _queue.end(); ++it) {
```
cmcfarlen, line 343: how big can queue get?

cmcfarlen (changes requested):
A bit concerned that the rate_limit ops are now globally serialized.  Consider finer grained locks.

plugins/experimental/rate_limit/limiter.h
moonchen: > how big can queue get?

Unbounded by default, which I agree is not sensible.

`_max_queue` is `0` (no queue) until a `queue:` block appears, and then
`limiter.h:214` is:

```cpp
_max_queue = queue["size"] ? queue["size"].as<uint32_t>() : UINT32_MAX;
```

A `queue:` block without a `size:` gets `UINT32_MAX`, and
`full()` (`_size >= max_queue()`) can then never trip. The practical ceiling
becomes `proxy.config.net.connections_throttle`, 30000 by default.

`_queue` is a `std::deque`, so an erase from the middle is O(n).  I'm filing an issue: #13511 to update defaults, and revisit this queue data structure.

plugins/experimental/rate_limit/sni_selector.cc
moonchen: > Could the lock be per-limiter instead of global?

I removed the global lock, so the only locking left is per-limiter: the
`_queue_lock` and `_active_lock` that `RateLimiter` already had.

cmcfarlen:
Both of my earlier points are addressed, and I verified them rather than taking the replies at face value. From my side this is clear.

**Global lock → per-limiter.** Confirmed: the only lock acquisition this PR adds is `std::lock_guard<std::mutex> lock(_queue_lock)` inside the new `remove()`, and `_queue_lock`/`_active_lock` are pre-existing per-limiter members (`limiter.h:407`, already used by `reserve()`, `free()`, `push()` and `pop()`). Grepping the diff for any `static`/global mutex turns up nothing. So `remove()` is exactly as granular as the methods it sits next to, and my serialization concern does not apply.

**Queue depth.** The inline comment on `remove()` now records the caveat where someone reading the O(n) erase will actually see it — that a `queue:` block without a `size:` leaves `_max_queue` at `UINT32_MAX` so `full()` never trips, leaving `proxy.config.net.connections_throttle` as the only real bound. With #13511 filed to fix the default and revisit the container, that is the right split: this PR stays a bug fix and the questionable default gets handled on its own.

While re-reading I went looking for a residual race and convinced myself it is closed, which is worth recording because it is the subtle part of the change. My worry was the check-then-act at the close site:

```cpp
if (!limiter->remove(vc)) {
  limiter->free();
}
```

`remove()` works under `_queue_lock` and `free()` under `_active_lock`, so the decision and the decrement are not atomic with respect to each other. If the sweep could dequeue a VC and only afterwards reserve its slot, a close landing in that window would see "not queued", call `free()`, and decrement a slot nothing had taken — the unmatched decrement that wraps the counter and makes the next `reserve()` trip `TSReleaseAssert(_active <= _limit)`, exactly the abort your test drives.

Reversing the sweep to reserve *before* dequeuing is what closes it, and the `nullptr == vc` give-back handles the case where the queue emptied underneath:

```cpp
while (limiter->size() > 0 && limiter->reserve() == ReserveStatus::RESERVED) {
  auto [vc, contp, start_time] = limiter->pop();
  if (nullptr == vc) { // A concurrent close emptied the queue; give the slot back
    limiter->free();
    break;
  }
```

That makes the three cases add up: still queued, `remove()` returns true and no slot is released because none was held; already resumed, `remove()` returns false and `free()` releases precisely the slot the sweep granted; sweep raced and got nothing, the slot goes straight back. The `max_age` expiry loop is consistent too, since it pops without reserving and detaches with `TSUserArgSet(vc, gVCIdx, nullptr)` so the eventual `VCONN_CLOSE` cannot release a slot it never had. The "Reserving before dequeuing means a resumed VC owns the slot it was granted" comment is doing real work — worth keeping verbatim.

The `TLSEventSupport.cc` half reads correctly as well: with a connection closing while parked in a handshake hook, `curHook` still points into the handshake hook's list, and since each hook id owns a separate list, advancing it would walk the handshake list instead of the close list — dropping the close event or handing it to the wrong plugin. Resetting to `nullptr` unless already in `HANDSHAKE_HOOKS_DONE` (i.e. unless already iterating the close list) is the right guard, and picking up `TS_EVENT_VCONN_OUTBOUND_CLOSE` alongside `TS_EVENT_VCONN_CLOSE` closes the same gap on the outbound side.

I will leave the formal approval to a separate action so my `CHANGES_REQUESTED` stops blocking. One process note, purely about release timing rather than the code: this grew from a focused rate_limit accounting fix to +621/-5 across 11 files with the VCONN_CLOSE delivery work folded in, and 10.2.0 is days from its RC. The two halves are genuinely related — the parked-hook close is what produces the queued-close accounting path — so I am not asking for a split on technical grounds. It is worth an explicit RM call on whether this lands in 10.2.0 or 10.2.1 given the Core/TLS blast radius.

cmcfarlen (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as 6b006336ca87bf51deb93bc1bbf146f326c374c9 for the 10.2.0 release.

PR #13409, opened 2026-07-20
Add live HostDB cache clearing

### LINKED ISSUE #9022: clear hostdb cache without restarting ATS
--clear-hostdb
Clear the DNS resolver cache upon startup. 

Requires a restart

bneradt:
Operators cannot discard stale HostDB entries without restarting Traffic
Server. A restart interrupts service and delays recovery from incorrect
DNS data.

This adds a traffic_ctl command and restricted JSON-RPC method to clear
HostDB safely under partition locks while preserving in-flight records.
This also uses scoped status snapshots and verifies clearing and
repopulation without a restart.

Fixes: #9022

### DIFF
--- a/doc/appendices/command-line/traffic_ctl.en.rst
+++ b/doc/appendices/command-line/traffic_ctl.en.rst
@@ -1290,6 +1290,13 @@ traffic_ctl hostdb
 
    If ``HOSTNAME`` is specified, the output is filtered to show only records whose names contain the given string.
 
+.. option:: clear
+
+   :ref:`admin_hostdb_clear`
+
+   Remove all DNS resolution records from HostDB without restarting |TS|. In-flight transactions retain their current resolution,
+   while subsequent transactions resolve their upstream hosts again.
+
 traffic_ctl rpc
 ---------------
 .. program:: traffic_ctl rpc
--- a/doc/appendices/command-line/traffic_ctl.en.rst
+++ b/doc/appendices/command-line/traffic_ctl.en.rst
@@ -1316,6 +1316,13 @@ traffic_ctl hostdb
 
    If ``HOSTNAME`` is specified, the output is filtered to show only records whose names contain the given string.
 
+.. option:: clear
+
+   :ref:`admin_hostdb_clear`
+
+   Remove all cached DNS resolution records from HostDB without restarting |TS|. Entries loaded from the host file remain. In-flight
+   transactions retain their current resolution, while subsequent transactions resolve their upstream hosts again.
+
 traffic_ctl rpc
 ---------------
 .. program:: traffic_ctl rpc
--- a/doc/developer-guide/jsonrpc/jsonrpc-api.en.rst
+++ b/doc/developer-guide/jsonrpc/jsonrpc-api.en.rst
@@ -453,6 +453,8 @@ JSONRPC API
 
 * `admin_host_set_status`_
 
+* `admin_hostdb_clear`_
+
 * `admin_server_stop_drain`_
 
 * `admin_server_start_drain`_
@@ -1393,6 +1395,58 @@ Response:
          ]
          ,"errorList":[]
       }
+   }
+
+
+HostDB
+======
+
+.. _admin_hostdb_clear:
+
+admin_hostdb_clear
+------------------
+
+|method|
+
+Description
+~~~~~~~~~~~
+
+Remove all cached DNS resolution records from HostDB. In-flight transactions retain their current resolution, while subsequent
+transactions resolve their upstream hosts again.
+
+Parameters
+~~~~~~~~~~
+
+* ``params``: Omitted
+
+Result
+~~~~~~
+
+The response will contain the default `success_response` or an error. :ref:`jsonrpc-node-errors`.
+
+Examples
+~~~~~~~~
+
+Request:
+
+.. code-block:: json
+   :linenos:
+
+   {
+      "id": "dc8b7f1e-2521-4bc0-b0ed-5e3ec751599d",
+      "jsonrpc": "2.0",
+      "method": "admin_hostdb_clear"
+   }
+
+Response:
+
+.. code-block:: json
+   :linenos:
+
+   {
+      "jsonrpc": "2.0",
+      "result": "success",
+      "id": "dc8b7f1e-2521-4bc0-b0ed-5e3ec751599d"
    }
 
 
--- a/doc/developer-guide/jsonrpc/jsonrpc-api.en.rst
+++ b/doc/developer-guide/jsonrpc/jsonrpc-api.en.rst
@@ -455,6 +455,8 @@ JSONRPC API
 
 * `admin_host_set_status`_
 
+* `admin_hostdb_clear`_
+
 * `admin_server_stop_drain`_
 
 * `admin_server_start_drain`_
@@ -1450,6 +1452,58 @@ Response:
          ]
          ,"errorList":[]
       }
+   }
+
+
+HostDB
+======
+
+.. _admin_hostdb_clear:
+
+admin_hostdb_clear
+------------------
+
+|method|
+
+Description
+~~~~~~~~~~~
+
+Remove all cached DNS resolution records from HostDB. Entries loaded from the host file remain. In-flight transactions retain their
+current resolution, while subsequent transactions resolve their upstream hosts again.
+
+Parameters
+~~~~~~~~~~
+
+* ``params``: Omitted
+
+Result
+~~~~~~
+
+The response will contain the default `success_response` or an error. :ref:`jsonrpc-node-errors`.
+
+Examples
+~~~~~~~~
+
+Request:
+
+.. code-block:: json
+   :linenos:
+
+   {
+      "id": "dc8b7f1e-2521-4bc0-b0ed-5e3ec751599d",
+      "jsonrpc": "2.0",
+      "method": "admin_hostdb_clear"
+   }
+
+Response:
+
+.. code-block:: json
+   :linenos:
+
+   {
+      "jsonrpc": "2.0",
+      "result": "success",
+      "id": "dc8b7f1e-2521-4bc0-b0ed-5e3ec751599d"
    }
 
 
--- a/include/iocore/hostdb/HostDBProcessor.h
+++ b/include/iocore/hostdb/HostDBProcessor.h
@@ -652,6 +652,9 @@ struct HostDBProcessor : public Processor {
    */
   int start(int no_of_additional_event_threads = 0, size_t stacksize = DEFAULT_STACKSIZE) override;
 
+  /// Remove all cached DNS resolution records.
+  void clear();
+
   // Private
   HostDBCache *cache();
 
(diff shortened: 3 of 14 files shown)


masaori335:
I'm just forwarding message, but P0 and P1 seems blocker. P4 is pretty interesting, this fixed a leak silently, which I introduced, thank you :)

----

# Review: `Add live HostDB cache clearing` (aa9fabc9c6)

The design is sound: `probe()` returns a `Ptr<HostDBRecord>` and `ResolveInfo::record`
holds it, so in-flight transactions genuinely survive a clear. The status-handler
rewrite also silently fixes a real refcount leak. But the new `clear()` exposes a
pre-existing lock-discipline hole in a way that is now operator-triggerable.

Findings are ranked P0 (blocking, memory-unsafe) through P4 (nit).

---

## P0 — `clear()` races with unlocked `erase()`: data race / use-after-free

`RefCountCache::clear()` now takes the partition's exclusive lock
(`src/iocore/hostdb/P_RefCountCache.h:513`), which correctly excludes `probe()`
(shared lock, `src/iocore/hostdb/HostDB.cc:482`) and `put()` (unique lock,
`src/iocore/hostdb/HostDB.cc:1000`).

It does **not** exclude the two `erase()` call sites that take no partition lock at all:

- `src/iocore/hostdb/HostDB.cc:887` — `dnsEvent()` NXDOMAIN path,
  `hostDB.refcountcache->erase(old_r->key)`. This is a common path.
- `src/iocore/hostdb/HostDB.cc:405` — `reply_to_cont()`, reached unlocked from
  `dnsEvent` (`:1030`) and `probeEvent` (`:1109`).

`erase()` mutates the intrusive hash map, mutates the `PriorityQueue expiry_queue`,
and frees the `RefCountCacheHashEntry` back to the ClassAllocator. Concurrently the
RPC thread walks and destroys that same map under a lock the ET_NET thread never
takes. Result: corrupted bucket/expiry lists, double-free, or UAF.

Before this commit `clear()` was only ever called from unit tests, so the unlocked
`erase()` sites raced only against each other and against `put()` on the same
partition — a much narrower window. This commit turns it into "operator runs
`traffic_ctl hostdb clear` during an NXDOMAIN storm."

**This is not a one-line fix.** `getby()` holds a *shared* lock at
`src/iocore/hostdb/HostDB.cc:580` and calls `reply_to_cont()` at `:600`, which can
call `erase()`. `ts::shared_mutex` wraps a non-recursive, writer-preferring
`pthread_rwlock_t`, so naively adding a `unique_lock` inside `erase()` self-deadlocks
that path. Either hoist the erase out of `reply_to_cont()`, or restructure so `getby()`
releases the shared lock before calling it.

**Strongly recommended:** annotate the partition state so the compiler enumerates the
violations instead of relying on review —

```cpp
hash_type item_map                                    TS_GUARDED_BY(lock);
PriorityQueue<RefCountCacheHashEntry *> expiry_queue  TS_GUARDED_BY(lock);
void erase(uint64_t key, ink_time_t expiry_time = -1) TS_REQUIRES(lock);
```

Note this will only flag on macOS/libc++ — on Fedora CI the std lock types are not
capabilities, so `-Wthread-safety` will stay quiet even with `AS_ERROR=ON`.

## P1 — Null deref if `clear` is called during startup

`initialize_jsonrpc_server()` runs at `src/traffic_server/traffic_server.cc:2143`;
`hostDBProcessor.start()` at `:2365`. Between them: `eventProcessor.start()`, thread
spawn, `Log::init()`. `hostDB.refcountcache` is `nullptr` until `HostDBCache::start()`
(`src/iocore/hostdb/HostDB.cc:289`). A `traffic_ctl hostdb clear` in that window
segfaults traffic_server.

`get_hostdb_status` has the same pre-existing hole (its `try`/`catch` does not catch
SIGSEGV), so one shared guard fixes both:

```cpp
if (hostDB.refcountcache == nullptr) {
  return {errata: "HostDB is not initialized"};
}
```

## P2 — The headline behavior is untested

`tests/gold_tests/dns/hostdb_clear.test.py` covers only the happy path. Nothing
verifies that an in-flight transaction survives a clear — the one behavior the commit
message and both doc changes promise, and the one most likely to silently regress.

## P3 — Emptiness assertion is a weak negative

The test asserts `ExcludesExpression(hostname)` on `traffic_ctl hostdb status` output.
That passes for the wrong reasons too (RPC error, empty response, changed formatting).
Assert on the metric instead: `proxy.process.hostdb.cache.current_items == 0`.

## P3 — No audit log for a destructive operation

Wiping the entire DNS cache emits nothing to `diags.log`. Add a
`Note("HostDB cleared via JSONRPC")` in `clear_hostdb()`
(`src/mgmt/rpc/handlers/hostdb/HostDB.cc:212`).

## P3 — `RefCountCachePartition::copy()` is now dead code

The status handler was its only caller. Delete the declaration
(`src/iocore/hostdb/P_RefCountCache.h:176`) and definition (`:346-355`).

## P3 — Prefer `ts::write_guard` over `std::unique_lock`

`include/tsutil/TsSharedMutex.h:188` explicitly directs new code to the
capability-carrying guards. The surrounding `std::unique_lock` uses in `HostDB.cc` are
legacy, not convention. Relevant to the P0 annotation work above.

## P3 — Writer-lock stall on ET_NET

Freeing up to `max_count / partitions` entries per partition under an exclusive lock
blocks ET_NET threads in `probe()`, and Linux's writer-preferring rwlock queues new
readers behind it. Bounded per partition, so likely fine — but consider scheduling the
per-partition clears on ET_TASK rather than running them inline on the RPC thread.

## P4 — Leak fix is real but unmentioned

The old status handler called `partition.copy()`, which `alloc()`s a
`RefCountCacheHashEntry` per record and `make_ptr()`s the item (refcount++), then
dropped the raw pointers on the floor. Every `traffic_ctl hostdb status` leaked one
hash entry and one permanent refcount per record — meaning those records could never
be freed. The new `Ptr`-based collection fixes it. Worth a line in the commit message
so it is not lost.

## P4 — Docs imply host-file entries are cleared

Host-file entries are correctly left alone (they are config, not cache), but "Remove
all DNS resolution records from HostDB" reads like they would be affected. A
clarifying clause in both `doc/appendices/command-line/traffic_ctl.en.rst` and
`doc/developer-guide/jsonrpc/jsonrpc-api.en.rst` would help.

## P4 — Argument-count rejection untested

`add_command("clear", ..., 0, ...)` declares arity 0, but no test exercises
`traffic_ctl hostdb clear extra_arg`.

## P4 — Inconsistent `StillRunningAfter` in the last test run

The final test run sets only `tr.StillRunningAfter = ts`, dropping `dns` and `server`
unlike every preceding run. Harmless, but inconsistent.

---

**Merge gate:** P0 and P1 should block. P2 is worth doing before merge, since the
in-flight guarantee is the feature's main safety claim. P3/P4 are cleanups.

bneradt:
Thanks @masaori335 for forwarding the concurrency and test audit. I amended the commit to address the actionable items:

- RefCountCache now owns its public read/write locking, guarded state is annotated, and the unlocked erase/callback paths no longer race or self-deadlock.
- HostDB RPCs return a server error before cache initialization, and successful clears write an audit note.
- The AuTest now checks the exact empty-cache metric, a delayed in-flight transaction across a clear, repopulation without restart, extra-argument rejection, and the audit log. I also corrected the final process-liveness assertions.
- The docs now clarify that host-file entries are not cleared, and the dead copy helper was removed.

I considered moving the clear loop to ET_TASK. Running the same loop there would move the work but would not shorten any partition writer-lock hold, so I kept the operation synchronous and bounded one partition at a time. A chunked/asynchronous clear would be a reasonable follow-up if production measurements show material control-plane stalls.

The branch is rebased on current master at 052d4a4984.

masaori335:
@bneradt Sorry for late reply on this. The fedora build error might be introduced.
```
FAILED: [code=1] src/mgmt/rpc/CMakeFiles/rpcpublichandlers.dir/handlers/hostdb/HostDB.cc.o 
/usr/sbin/ccache /usr/sbin/clang++ -DDEBUG -DOPENSSL_API_COMPAT=10002 -DOPENSSL_IS_AT_LEAST_OPENSSL3 -DPACKAGE_NAME="\"Apache Traffic Server\"" -DPACKAGE_VERSION=\"11.0.0\" -DTS_OPENSSL_QUIC_TLS_CBS_COMPAT -D_DEBUG -Dlinux -I/home/jenkins/workspace/Github_Builds/fedora/src/include -I/home/jenkins/workspace/Github_Builds/fedora/src/build/include -I/home/jenkins/workspace/Github_Builds/fedora/src/lib -I/home/jenkins/workspace/Github_Builds/fedora/src/lib/swoc/include -I/home/jenkins/workspace/Github_Builds/fedora/src/lib/yamlcpp/include -isystem /home/jenkins/workspace/Github_Builds/fedora/src/lib/systemtap -pthread -g -std=c++20 -Wthread-safety -Werror=thread-safety -DENABLE_SYSTEMTAP_PROBES -Wno-invalid-offsetof -pipe -Wall -Wextra -Wno-noexcept-type -Wsuggest-override -Wno-vla-extension -fno-strict-aliasing -mcx16 -Wno-deprecated-declarations -Werror -MD -MT src/mgmt/rpc/CMakeFiles/rpcpublichandlers.dir/handlers/hostdb/HostDB.cc.o -MF src/mgmt/rpc/CMakeFiles/rpcpublichandlers.dir/handlers/hostdb/HostDB.cc.o.d -o src/mgmt/rpc/CMakeFiles/rpcpublichandlers.dir/handlers/hostdb/HostDB.cc.o -c /home/jenkins/workspace/Github_Builds/fedora/src/src/mgmt/rpc/handlers/hostdb/HostDB.cc
In file included from ../src/mgmt/rpc/handlers/hostdb/HostDB.cc:25:
In file included from ../include/../src/iocore/hostdb/P_HostDBProcessor.h:33:
../include/../src/iocore/hostdb/P_RefCountCache.h:348:16: error: returning variable 'item_map' by reference requires holding shared_mutex 'lock' exclusively [-Werror,-Wthread-safety-reference-return]
  348 |   return this->item_map;
      |                ^
../src/mgmt/rpc/handlers/hostdb/HostDB.cc:93:39: note: in instantiation of member function 'RefCountCachePartition<HostDBRecord>::get_map' requested here
   93 |         for (auto &&entry : partition.get_map()) {
      |                                       ^
1 error generated.
```

bneradt:
Thanks @masaori335 — I addressed the Fedora issue against a local build using the CI container and amended the fix into 2e38442db0.

`get_map()` returned a mutable reference to guarded state despite requiring only a shared lock. It now returns a const reference and is a const member function, matching the status handler's read-only iteration.

I also addressed the replay-fixture feedback: the delayed response is now in `hostdb_clear.replay.yaml`, and the shared `single_transaction.replay.yaml` used by `dns_ttl` has its original timing restored.

I am retaining per-partition locking for clear. Concurrent DNS resolutions may repopulate a partition once it has been cleared; the operation does not suspend lookups or promise that the cache remains empty under active traffic. Holding all partition locks would create a global pause and still allow repopulation immediately after unlocking, before the RPC response reaches the caller. A global barrier/generation policy would require a separately defined stronger contract.

The amended branch is pushed.

PR #13411, opened 2026-07-20
txn_box: Log text block reload failures

bneradt:
Text blocks silently lose their file-backed content when a periodic
reload can no longer read the configured file. Requests then see an
empty block without an operator-visible diagnostic.

This logs the transition to unavailable content once and adds
regression coverage for a file that becomes unreadable. This also
documents the error and fallback behavior.

The original txn_box repository is defunct, so fixes for the plugin are
now tracked in the Traffic Server repository. This patch addresses:
https://github.com/SolidWallOfCode/txn_box/issues/90


brbzull0 (approved):
lgtm.
Thanks.

PR #13412, opened 2026-07-21
Add APIs to set a firewall mark with a mask

JosiahWI:
This adds two new overloads to the existing firewall mark APIs, per the proposal I sent to the dev mailing list:

- `TSHttpTxnClientPacketMarkSet(TsHttpTxn, int, int)`
- `TSHttpTxnServerPacketMarkSet(TsHttpTxn, int, int)`

### DIFF
--- a/include/ts/ts.h
+++ b/include/ts/ts.h
@@ -1643,6 +1643,32 @@ TSReturnCode TSHttpTxnClientPacketMarkSet(TSHttpTxn txnp, int mark, int mask);
 */
 TSReturnCode TSHttpTxnServerPacketMarkSet(TSHttpTxn txnp, int mark);
 
+/** Change selected bits of the packet firewall mark for the server side connection
+
+    Only the bits selected by @a mask are modified. Bits set in @a mask are updated from the
+    corresponding bits of @a mark; bits clear in @a mask retain their previous value. Equivalently,
+    the resulting mark is computed as:
+
+    @code
+    result = (current_mark & ~mask) | (mark & mask);
+    @endcode
+
+    @a mark and @a mask are interpreted as 32-bit unsigned bit patterns. A @a mask of 0xFFFFFFFF
+    replaces the entire mark and is equivalent to the two-argument overload. A @a mask of 0 leaves
+    the mark unchanged. The current mark in this formula is the value recorded on @a txnp, not
+    socket state.
+
+    @note The firewall mark is only honored on platforms whose OS supports it, specifically Linux via
+    @c SO_MARK. On platforms without @c SO_MARK support the call still returns TS_SUCCESS, but setting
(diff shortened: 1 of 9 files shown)


src/api/InkAPI.cc
```diff
@@ -4900,6 +4900,26 @@ TSHttpTxnClientPacketMarkSet(TSHttpTxn txnp, int mark)
...
+    return TS_ERROR;
+  }
+
+  NetVConnection *vc = sm->get_ua_txn()->get_netvc();
+  if (nullptr == vc) {
+    return TS_ERROR;
+  }
+
+  vc->options.packet_mark =
+    (vc->options.packet_mark & ~static_cast<uint32_t>(mask)) | (static_cast<uint32_t>(mark) & static_cast<uint32_t>(mask));
```
shinrich, line 4918: Shouldn't we

src/api/InkAPI.cc
```diff
@@ -4921,6 +4941,21 @@ TSHttpTxnServerPacketMarkSet(TSHttpTxn txnp, int mark)
...
+  sdk_assert(sdk_sanity_check_txn(txnp) == TS_SUCCESS);
+
+  // The current mark lives on the transaction config so the masked update composes
+  // correctly even before an origin connection exists.
+  TSMgmtInt current = 0;
+  TSHttpTxnConfigIntGet(txnp, TS_CONFIG_NET_SOCK_PACKET_MARK_OUT, &current);
+  uint32_t new_mark =
+    (static_cast<uint32_t>(current) & ~static_cast<uint32_t>(mask)) | (static_cast<uint32_t>(mark) & static_cast<uint32_t>(mask));
+
+  return TSHttpTxnServerPacketMarkSet(txnp, static_cast<int>(new_mark));
```
shinrich, line 4956: We should

include/ts/ts.h
```diff
@@ -1601,6 +1601,32 @@ TSReturnCode           TSHttpSsnClientFdGet(TSHttpSsn ssnp, int *fdp);
...
+    @c SO_MARK. On platforms without @c SO_MARK support the call still returns TS_SUCCESS when a
+    client connection is present, but setting the mark has no effect at the OS layer (it is a safe
+    no-op).
+
+    @note The change takes effect immediately on the live client connection
+
+    @return TS_SUCCESS if the client connection was modified, TS_ERROR if there is no client
+    connection to modify
+*/
+TSReturnCode TSHttpTxnClientPacketMarkSet(TSHttpTxn txnp, int mark, int mask);
```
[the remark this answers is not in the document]
JosiahWI: I'll have to think about this. This is an interesting observation.

JosiahWI:
After further consideration, we realized it is not useful to have a mask for this API, because no other processes should set the socket-level default mark on the socket.

PR #13417, opened 2026-07-21
Logging: add cache freshness fields

### LINKED ISSUE #13127: Add ability to log "freshness_limit" and "current_age"
I want to log “freshness_limit” in ATS to allow us to better model the cache contents.
It would be good to enable the logging of “current_age” also.

This allows us to create logs for modeling cache age and churn over extended periods.

These would be new log fields:
* freshness_limit written when objects are served from or written to the cache.
* current_age when serving

[Ideally avoiding header use]

bneradt:
Cache modeling cannot currently observe an object’s effective
freshness limit or its age when ATS serves it, leaving access logs
unable to describe cache lifetime and churn.

This adds cfl and cca log fields and records their values when ATS
prepares cache metadata or serves a cached response. It resets the
values for each lookup and adds replay coverage for writes, hits, and
non-cacheable responses.

Fixes: #13127

Co-authored-by: sheikh-saifi <p240592@pwr.nu.edu.pk>

### DIFF
--- a/doc/admin-guide/logging/formatting.en.rst
+++ b/doc/admin-guide/logging/formatting.en.rst
@@ -157,6 +157,8 @@ Cache Details
 .. _crra:
 .. _cwra:
 .. _cccs:
+.. _cfl:
+.. _cca:
 
 These log fields reveal details of the |TS| proxy interaction with its own
 cache while attempting to service incoming client requests.
@@ -191,6 +193,17 @@ cccs  Proxy Cache    Cache collapsed connection success;
                      -1: collapsing was attempted but failed, request went upstream
                      0: collapsing was unnecessary
                      1: attempted to collapse and got a cache hit on subsequent read attempts
+
+cfl   Proxy Cache    Freshness limit, in seconds, for an object served from or
+                     written to cache. The limit reflects parsed cache response
+                     metadata and the effective cache configuration. The value
+                     is ``-1`` when not applicable.
+
+cca   Proxy Cache    Current age, in seconds, of the cached object when served.
+                     The age is calculated from the cached response metadata
+                     and cache timing state. The value is ``-1`` when not
+                     applicable.
+
 ===== ============== ==========================================================
 
 .. _admin-logging-fields-txn:
--- a/doc/admin-guide/logging/formatting.en.rst
+++ b/doc/admin-guide/logging/formatting.en.rst
@@ -157,6 +157,8 @@ Cache Details
 .. _crra:
 .. _cwra:
 .. _cccs:
+.. _cfl:
+.. _cca:
 
 These log fields reveal details of the |TS| proxy interaction with its own
 cache while attempting to service incoming client requests.
@@ -191,6 +193,17 @@ cccs  Proxy Cache    Cache collapsed connection success;
                      -1: collapsing was attempted but failed, request went upstream
                      0: collapsing was unnecessary
                      1: attempted to collapse and got a cache hit on subsequent read attempts
+
+cfl   Proxy Cache    Freshness limit, in seconds, for an object served from or
+                     written to cache. The limit is calculated from response
+                     headers and the configured guaranteed and heuristic
+                     lifetime bounds. The value is ``-1`` when not applicable.
+
+cca   Proxy Cache    Current age, in seconds, of the cached object when served.
+                     The age is calculated from the cached response metadata
+                     and cache timing state. The value is ``-1`` when not
+                     applicable.
+
 ===== ============== ==========================================================
 
 .. _admin-logging-fields-txn:
--- a/src/proxy/http/HttpTransact.cc
+++ b/src/proxy/http/HttpTransact.cc
@@ -2139,6 +2139,9 @@ HttpTransact::HandleRequestAuthorized(State *s)
 void
 HttpTransact::DecideCacheLookup(State *s)
 {
+  s->cache_info.freshness_limit = -1;
+  s->cache_info.current_age     = -1;
+
   // Check if a client request is lookupable.
   if (s->redirect_info.redirect_in_process) {
     // for redirect, we want to skip cache lookup and write into
@@ -3059,6 +3062,7 @@ HttpTransact::build_response_from_cache(State *s, HTTPWarningCode warning_code)
     obj = s->cache_info.object_read;
   }
   cached_response = obj->response_get();
+  set_cache_freshness_info(s, cached_response, obj->request_sent_time_get(), obj->response_received_time_get(), true);
 
   // If the client request is conditional, and the cached copy meets
   // the conditions, do not need to send back the full document,
@@ -4835,6 +4839,7 @@ HttpTransact::handle_cache_operation_on_forward_server_response(State *s)
     // unset warning revalidation failed header if it set
     // (potentially added by negative revalidating)
     delete_warning_value(base_response, HTTPWarningCode::REVALIDATION_FAILED);
+    set_cache_freshness_info(s, base_response, s->request_sent_time, s->response_received_time, true);
   }
   ink_assert(base_response->valid());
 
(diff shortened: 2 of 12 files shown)


sheikh-saifi:
Thanks for fixing this! the reset in DecideCacheLookup and splitting the write vs hit call sites was exactly what my version was missing. Went through the code and the three test cases (write/hit/uncacheable) and it all makes sense.
Sorry for disappearing after the first review, got pulled into other stuff and should've followed up. Happy to help test or review anything else on this.

bneradt:
> Thanks for fixing this! the reset in DecideCacheLookup and splitting the write vs hit call sites was exactly what my version was missing. Went through the code and the three test cases (write/hit/uncacheable) and it all makes sense. Sorry for disappearing after the first review, got pulled into other stuff and should've followed up. Happy to help test or review anything else on this.

Sounds good. Thanks for the comments. I definitely plan to attribute your work with the commit, in addition to the PR here.

moonchen (commented):
Superseded. Please read my later review on this PR instead. This one had two incorrect line references and one inaccurate claim, both fixed there.

src/proxy/http/HttpTransact.cc
```diff
@@ -3059,6 +3062,7 @@ HttpTransact::build_response_from_cache(State *s, HTTPWarningCode warning_code)
     obj = s->cache_info.object_read;
   }
   cached_response = obj->response_get();
+  set_cache_freshness_info(s, cached_response, obj->request_sent_time_get(), obj->response_received_time_get(), true);
```
moonchen: These should use `s->request_sent_time` and `s->response_received_time` rather than the object's raw timestamps.

`what_is_document_freshness()` uses the transaction timestamps, which are capped against `s->client_request_time`. The `Age` header sent to the client uses those values as well. Using the object's raw timestamps here can therefore make `cca` differ from both the freshness decision and the response's `Age` header.

There is also a problem on the `api_update_cached_object == CONTINUE` path. There, `obj` is `&s->cache_info.object_store`, whose timestamps can still be zero after a plugin creates it with `TSHttpTxnCachedRespModifiableGet()`. Passing those values to `calculate_document_age()` makes `cca` roughly the current Unix timestamp.

Passing `s->request_sent_time` and `s->response_received_time` fixes both cases.

src/proxy/http/HttpTransact.cc
```diff
@@ -4835,6 +4839,7 @@ HttpTransact::handle_cache_operation_on_forward_server_response(State *s)
     // unset warning revalidation failed header if it set
     // (potentially added by negative revalidating)
     delete_warning_value(base_response, HTTPWarningCode::REVALIDATION_FAILED);
+    set_cache_freshness_info(s, base_response, s->request_sent_time, s->response_received_time, true);
```
moonchen, line 4845: This call is skipped by the negative revalidation path earlier in the function. That branch sets `action = SERVE_AND_UPDATE` and returns before reaching here.

Those transactions serve the stale cached object and extend its stored `Expires`, but leave both freshness fields at -1. Negative revalidation is enabled by default for 500, 502, 503, and 504 responses. There is a reproducer in the review summary.

doc/admin-guide/logging/formatting.en.rst
```diff
@@ -191,6 +193,17 @@ cccs  Proxy Cache    Cache collapsed connection success;
                      -1: collapsing was attempted but failed, request went upstream
                      0: collapsing was unnecessary
                      1: attempted to collapse and got a cache hit on subsequent read attempts
+
+cfl   Proxy Cache    Freshness limit, in seconds, for an object served from or
+                     written to cache. The limit reflects parsed cache response
+                     metadata and the effective cache configuration. The value
```
moonchen: "the effective cache configuration" is broader than what `calculate_document_freshness_limit()` currently includes.

The calculation uses the response headers plus `guaranteed_min/max_lifetime` and `heuristic_min/max_lifetime`, but not `s->cache_control`. That excludes `cache.config` directives such as `ttl-in-cache` and `revalidate-after`.

For example, with `ttl-in-cache=300` and an origin response containing `max-age=1`, a hit at age 3 logs `cfl=1 cca=3`.

moonchen (changes requested):

doc/admin-guide/logging/formatting.en.rst
bneradt: Updated in 2460007d79. The documentation now says the limit is calculated from response headers and the configured guaranteed/heuristic lifetime bounds, without implying that `cache.config` directives such as `ttl-in-cache` or `revalidate-after` are included.

src/proxy/http/HttpTransact.cc
bneradt: Fixed in 2460007d79. `build_response_from_cache()` now passes `s->request_sent_time` and `s->response_received_time`, so `cca` uses the same capped transaction timestamps as freshness evaluation and the Age header. This also avoids zero timestamps from `object_store` on the modifiable cached-response path.

src/proxy/http/HttpTransact.cc
bneradt: Fixed in 2460007d79. The negative-revalidation early-return path now calls `set_cache_freshness_info()` after extending the cached response. I also added a replay regression that exercises a `TCP_REFRESH_FAIL_HIT` and verifies that `cfl` and `cca` are populated.

moonchen (changes requested):
One remaining correctness issue:

[P2] Do not publish cfl before the cache write succeeds

set_headers_for_cache_write() sets the transaction-wide cfl while merely preparing cache metadata. For transformed responses, this happens before CACHE_ISSUE_WRITE_TRANSFORM acquires the write lock (HttpTransact.cc:5145), and the lock-failure branch at HttpTransact.cc:5104-5108 leaves cfl set. Cache body or disk errors can likewise happen later in HttpSM::tunnel_handler_cache_write().

Consequently, cfl can describe an object that was never stored. On a transformed cache hit, the failed transform attempt can also overwrite the freshness limit of the cached object actually served. This contradicts the documented "served from or written to cache" semantics and makes cache-content modeling inaccurate. Please preserve the write-side value as pending until the write succeeds, without overwriting a serve-side value on failure, and add coverage for a failed-write path.

The three findings from my previous review are resolved.

bneradt:
Addressed the remaining review issue and force-pushed `18e63ce1c8`.

Write-side freshness limits are now kept pending separately for the original and transformed cache objects. `cfl` is published only after `tunnel_handler_cache_write()` reports a successful write; failures discard the pending value without overwriting freshness information for a cached object already selected for serving. If both object variants succeed, the transformed object retains precedence.

I also added deterministic failed-write coverage by exceeding `proxy.config.cache.max_doc_size`. The replay verifies that the write result is `ERR` and both `cfl` and `cca` remain `-1`; the existing successful-write, hit, negative-revalidation, and non-cacheable cases still pass.

Validation in `asfats5`:

- Full configure, format, build, and install: passed.
- Focused `cache-freshness-fields` AuTest: passed.

tests/gold_tests/logging/verify_cache_freshness_fields.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def load_log_entries(log_path: pathlib.Path) -> dict[str, tuple[int, int, str, str]]:
+    entries: dict[str, tuple[int, int, str, str]] = {}
+
+    for line in log_path.read_text(encoding='utf-8').splitlines():
+        uuid, freshness_limit, current_age, cache_result, cache_write_result = line.split()
+        entries[uuid] = (int(freshness_limit), int(current_age), cache_result, cache_write_result)
+
+    return entries
```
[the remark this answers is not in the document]
bneradt: Addressed in e9988d6c89. The verifier now reports the file and line for rows with the wrong field count, rejects duplicate UUIDs explicitly, and converts those validation errors into concise stderr diagnostics. Targeted malformed-row and duplicate assertions pass, as does the full `cache-freshness-fields` AuTest.

moonchen (dismissed):
Thanks for addressing the review. The P2 is fixed: `set_headers_for_cache_write()` now stages into `pending_write_freshness_limit` / `pending_transform_freshness_limit`, and `tunnel_handler_cache_write()` only publishes on `COMPLETE`, with the transform value winning when both objects are written.

Ran the autest against an installed build. Log rows: `failed-cache-write -1 -1 TCP_MISS ERR` and `negative-revalidation 1 2 TCP_REFRESH_FAIL_HIT`. Also mutation-tested it -- restoring the old behavior (publishing at prepare time) makes the failed-write row log `cfl=120` and the test fails, so the fix is pinned. LGTM.

bneradt:
[approve ci autest 2]

PR #13418, opened 2026-07-21
Add traffic_ctl cache clear command

### LINKED ISSUE #9399: Provide ability to clear cache with traffic_ctl
Feature request:

It would certainly save a load of time for workload validation if you could get ATS to re-initialise its cache without restarting.

e.g. traffic_ctl cache clear

When load testing for example you need to ensure the same state between comparative runs.
Ideally something such as the above would, as best as ATS can tell, have the same state as it would when restarted.

This would really help with load validation automation for example. Reducing the number of states that need tracking by the orchestrator. Perhaps I would go further than say just clear the cache, and clear all running state.

bneradt:
Cache resets currently require restarting Traffic Server or manually
changing the cache generation. This complicates repeatable workload
comparisons and their automation.

This adds a traffic_ctl cache clear command backed by a restricted
JSON-RPC method that advances the global HTTP cache generation. This
also documents the logical purge semantics and exercises the command in
the cache-generation AuTest.

Fixes: #9399

### DIFF
--- a/doc/appendices/command-line/traffic_ctl.en.rst
+++ b/doc/appendices/command-line/traffic_ctl.en.rst
@@ -43,6 +43,8 @@ of subcommands that control different aspects of Traffic Server:
 
 :program:`traffic_ctl config`
    Manipulate and display configuration records
+:program:`traffic_ctl cache`
+   Invalidate cached objects
 :program:`traffic_ctl metric`
    Manipulate performance and status metrics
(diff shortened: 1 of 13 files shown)


cmcfarlen (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as 0f4a5da34453421d38463dc40310375189abc472 for the 10.2.0 release.

PR #13419, opened 2026-07-21
Clarify SRV origin selection documentation

### LINKED ISSUE #12440: Feature request for remapping rules via SRV records
I wrote a very crude C program for fetching SRV records to generate a `remap.confg` file.  I called it [trafficmap](https://github.com/tuaris/trafficmap/blob/master/src/trafficmap.c).  It takes in a text file with a list of domains.  One would run the program and then reload traffic server.

SRV records look like this:

```
_http._tcp		IN	SRV	5 0 8002 web03.domain.tld.
_http._tcp.www	IN	SRV	5 0 8002 web03.domain.tld.
```

The resulting remap.config looks like this:

```
map http://mywebsite.tld http://web03.domain.tld:8002
map http://www.mywebsite.tld  http://web03.domain.tld:8002
```

I'd like to see a feature in Traffic Server where this would happen in real time with caching (while respecting TTL).  Without having the need to pre-generate any configuration.  Changes and new additions would propagate automatically through DNS.  A plugin would be fine, but I'd prefer to not have that via Lua scripting/plugin.

bneradt:
SRV origin discovery is difficult to find outside the records
reference, and the existing port description contradicts Traffic Server
behavior. This can cause operators to miss the feature or constrain SRV
records unnecessarily.

This documents lookup-name construction, SRV target and port selection,
fallback behavior, HostDB caching, and the feature's use from remap
replacement URLs.

Fixes: #12440

### DIFF
--- a/doc/admin-guide/files/remap.config.en.rst
+++ b/doc/admin-guide/files/remap.config.en.rst
@@ -111,6 +111,14 @@ Traffic Server recognizes three space-delimited fields: ``type``,
 
     where ``scheme`` is ``http``, ``https``, ``ws`` or ``wss``.
 
+    To select an origin host and port dynamically through DNS, enable
+    :ts:cv:`proxy.config.srv_enabled`. |TS| then queries the SRV
+    service derived from the replacement URL before resolving the replacement
+    host directly. For example, an ``https://origin.example.com/`` replacement
+    causes a query for ``_https._tcp.origin.example.com``. When a usable SRV
+    record is found, its target and port are used for the origin connection;
+    otherwise, |TS| falls back to the replacement host and port.
+
    .. note:: A remap rule for requests that upgrade from HTTP to WebSocket still require a remap rule with the ``ws`` or ``wss`` scheme.
 
 
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -3211,13 +3211,23 @@ DNS
    :overridable:
 
    Enables (``1``) or disables (``0``) the use of SRV records for origin server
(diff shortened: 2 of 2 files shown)


doc/admin-guide/files/remap.config.en.rst
```diff
@@ -111,6 +111,14 @@ Traffic Server recognizes three space-delimited fields: ``type``,
 
     where ``scheme`` is ``http``, ``https``, ``ws`` or ``wss``.
 
+    To select an origin host and port dynamically through DNS, enable
+    :ts:cv:`proxy.config.srv_enabled`. Traffic Server then queries the SRV
```
randall: (I don't know the rules (if I ever did here) so feel free to ignore). 

Should `Traffic Server` be  `|TS|` here?

doc/admin-guide/files/remap.config.en.rst
```diff
@@ -111,6 +111,14 @@ Traffic Server recognizes three space-delimited fields: ``type``,
...
+    service derived from the replacement URL before resolving the replacement
+    host directly. For example, an ``https://origin.example.com/`` replacement
+    causes a query for ``_https._tcp.origin.example.com``. When a usable SRV
+    record is found, its target and port are used for the origin connection;
+    otherwise, Traffic Server falls back to the replacement host and port.
```
randall: and here?

doc/admin-guide/files/remap.config.en.rst
bneradt: Thank you @randall . You're right, that should be `|TS|`.

bneradt:
[approve ci]

randall (approved):

PR #13420, opened 2026-07-23
Reduce H2 header allocations

zwoop:
This reduces per-request HTTP/2 overhead in two independent steps. First, the HPACK header-block decode happens in place when the whole block is contiguous in the frame reader (avoiding a per-request malloc+memcpy+free), and HEADERS-frame encoding uses a stack buffer for typical header sizes. Second — the larger win — the decoded request header is handed directly to HttpSM instead of being serialized and reparsed: the 2→1.1 converter now normalizes the URL (splitting host:port and path?query the way the reparse did), so the pre-parsed header can be copy()'d into the state machine, skipping a full parse_req per request. In local benchmarking this roughly doubles small-request throughput (~800K → ~1.68M req/s).

Co-Author: Masakazu Kitajo

### DIFF
--- a/include/proxy/hdrs/HdrHeap.h
+++ b/include/proxy/hdrs/HdrHeap.h
@@ -175,6 +175,7 @@ struct StrHeapDesc {
 class HdrHeap
 {
 public:
+  // Also sizes HTTP/2's on-stack HEADERS encode buffer (2*this).
   static constexpr int DEFAULT_SIZE = 2048;
 
   void init();
--- a/include/proxy/hdrs/URL.h
+++ b/include/proxy/hdrs/URL.h
@@ -215,6 +215,9 @@ ParseResult url_parse_http(HdrHeap *heap, URLImpl *url, const char **start, cons
                            bool verify_host_characters);
 ParseResult url_parse_http_regex(HdrHeap *heap, URLImpl *url, const char **start, const char *end, bool copy_strings);
 
+// strict_uri_parsing 0 = no check; 1/2 apply the same test url_parse() does.
+bool url_is_uri_compliant(int strict_uri_parsing, std::string_view value);
+
 char *url_unescapify(Arena *arena, const char *str, int length);
 
 void unescape_str(char *&buf, char *buf_e, const char *&str, const char *str_e, int &state);
--- a/include/proxy/http/HttpSM.h
+++ b/include/proxy/http/HttpSM.h
@@ -205,20 +205,6 @@ public:
 
   void attach_client_session(ProxyTransaction *txn);
 
-  // Borrowed: must outlive the copy in state_read_client_request_header. Lets HTTP/2 skip serialize+reparse.
-  void
-  set_pre_parsed_ua_request(HTTPHdr *hdr)
-  {
-    _pre_parsed_ua_request = hdr;
-  }
-
-  // Non-null until state_read_client_request_header copies the borrow; lets send_headers assert synchronous delivery.
-  bool
-  has_pending_pre_parsed_ua_request() const
-  {
-    return _pre_parsed_ua_request != nullptr;
-  }
-
   // Called after the network connection has been completed
   //  to set the session timeouts and initiate a read while
   //  holding the lock for the server session
@@ -624,6 +610,12 @@ private:
   IOBufferReader *_netvc_reader      = nullptr;
   MIOBuffer      *_netvc_read_buffer = nullptr;
 
+  bool _client_response_header_is_ready = false;
+  bool _server_request_header_is_ready  = false;
+
+  // Direct-passed headers bypass the tunnel, so client_response_hdr_bytes stays 0.
+  int _direct_response_hdr_bytes = 0;
+
   void kill_this();
   void update_stats();
   void transform_cleanup(TSHttpHookID hook, HttpTransformInfo *info);
@@ -640,6 +632,23 @@ public:
   int     client_transaction_priority_weight() const;
   int     client_transaction_priority_dependence() const;
 
+  HTTPHdr *get_client_response_header();
+  HTTPHdr *get_server_request_header();
+
+  // For logging/SDK: client_response_hdr_bytes counts only what the tunnel wrote.
+  int
+  reported_client_response_hdr_bytes() const
+  {
+    return client_response_hdr_bytes > 0 ? client_response_hdr_bytes : _direct_response_hdr_bytes;
+  }
+
+  void
+  clear_pending_send_header()
+  {
+    _client_response_header_is_ready = false;
+    _server_request_header_is_ready  = false;
+  }
+
   ink_hrtime get_server_inactivity_timeout();
   ink_hrtime get_server_active_timeout();
   ink_hrtime get_server_connect_timeout();
@@ -647,8 +656,6 @@ public:
 
 private:
   void cancel_pending_server_connection();
-
-  HTTPHdr *_pre_parsed_ua_request = nullptr;
 };
 
 ////
@@ -729,11 +736,28 @@ HttpSM::get_cache_sm()
 inline int
 HttpSM::write_response_header_into_buffer(HTTPHdr *h, MIOBuffer *b)
 {
-  if (t_state.client_info.http_version == HTTPVersion(0, 9)) {
+  if (_ua.get_txn()->supports_direct_header_passing()) {
+    // Nothing lands in the buffer, so 0 keeps the tunnel's byte math honest.
+    _client_response_header_is_ready = true;
+    _direct_response_hdr_bytes       = h->length_get();
+    return 0;
+  } else if (t_state.client_info.http_version == HTTPVersion(0, 9)) {
     return 0;
   } else {
     return write_header_into_buffer(h, b);
   }
+}
+
+inline HTTPHdr *
+HttpSM::get_client_response_header()
+{
+  return _client_response_header_is_ready ? &t_state.hdr_info.client_response : nullptr;
+}
+
+inline HTTPHdr *
+HttpSM::get_server_request_header()
+{
+  return _server_request_header_is_ready ? &t_state.hdr_info.server_request : nullptr;
 }
 
 inline int
--- a/include/proxy/http/HttpSM.h
+++ b/include/proxy/http/HttpSM.h
@@ -205,6 +205,20 @@ public:
 
   void attach_client_session(ProxyTransaction *txn);
 
+  // Borrowed: must outlive the copy in state_read_client_request_header. Lets HTTP/2 skip serialize+reparse.
+  void
+  set_pre_parsed_ua_request(HTTPHdr *hdr)
+  {
+    _pre_parsed_ua_request = hdr;
+  }
+
+  // Non-null until state_read_client_request_header copies the borrow; lets send_headers assert synchronous delivery.
+  bool
+  has_pending_pre_parsed_ua_request() const
+  {
+    return _pre_parsed_ua_request != nullptr;
+  }
+
   // Called after the network connection has been completed
   //  to set the session timeouts and initiate a read while
   //  holding the lock for the server session
@@ -633,6 +647,8 @@ public:
 
 private:
   void cancel_pending_server_connection();
+
+  HTTPHdr *_pre_parsed_ua_request = nullptr;
 };
 
 ////
--- a/include/proxy/http2/Http2Stream.h
+++ b/include/proxy/http2/Http2Stream.h
@@ -82,6 +82,11 @@ public:
   bool expect_receive_trailer() const override;
   void set_expect_receive_trailer() override;
 
+  bool           supports_direct_header_passing() const override;
+  bool           is_parsed_receive_header_ready() const override;
+  const HTTPHdr *parsed_receive_header() const override;
+  bool           has_pending_send_header() const override;
+
   Http2ErrorCode decode_header_blocks(HpackHandle &hpack_handle, uint32_t maximum_table_size, uint32_t header_field_max_size);
   Http2ErrorCode decode_header_blocks(HpackHandle &hpack_handle, uint32_t maximum_table_size, uint32_t header_field_max_size,
                                       const uint8_t *block, uint32_t block_len);
@@ -222,6 +227,7 @@ private:
   int              _sent_request_method{-1};
 
   HTTPHdr _receive_header;
+  bool    _is_parsed_receive_header_ready = false;
 #if TS_USE_MALLOC_ALLOCATOR
   MIOBuffer _receive_buffer{BUFFER_SIZE_INDEX_FOR_XMALLOC_SIZE(4096)};
 #else
--- a/include/proxy/http2/Http2Stream.h
+++ b/include/proxy/http2/Http2Stream.h
@@ -83,6 +83,8 @@ public:
   void set_expect_receive_trailer() override;
 
   Http2ErrorCode decode_header_blocks(HpackHandle &hpack_handle, uint32_t maximum_table_size, uint32_t header_field_max_size);
+  Http2ErrorCode decode_header_blocks(HpackHandle &hpack_handle, uint32_t maximum_table_size, uint32_t header_field_max_size,
+                                      const uint8_t *block, uint32_t block_len);
   void           send_headers(Http2ConnectionState &cstate);
   void           initiating_close();
   bool           is_outbound_connection() const;
--- a/src/proxy/hdrs/URL.cc
+++ b/src/proxy/hdrs/URL.cc
@@ -1212,6 +1212,22 @@ url_is_mostly_compliant(const char *start, const char *end)
 } // namespace UrlImpl
 using namespace UrlImpl;
 
+bool
+url_is_uri_compliant(int strict_uri_parsing, std::string_view value)
+{
+  const char *start = value.data();
+  const char *end   = start + value.length();
+
+  switch (strict_uri_parsing) {
+  case 1:
+    return url_is_strictly_compliant(start, end);
+  case 2:
+    return url_is_mostly_compliant(start, end);
+  default:
+    return true;
+  }
+}
+
 ParseResult
 url_parse(HdrHeap *heap, URLImpl *url, const char **start, const char *end, bool copy_strings_p, int strict_uri_parsing,
           bool verify_host_characters)
--- a/src/proxy/hdrs/VersionConverter.cc
+++ b/src/proxy/hdrs/VersionConverter.cc
@@ -201,7 +201,15 @@ VersionConverter::_convert_req_from_2_to_1(HTTPHdr &header) const
   if (MIMEField *field = header.field_find(PSEUDO_HEADER_AUTHORITY);
       field != nullptr && field->value_is_valid(is_control_BIT | is_ws_BIT)) {
     auto authority{field->value_get()};
-    header.m_http->u.req.m_url_impl->set_host(header.m_heap, authority, true);
+
+    // Match url_parse_http(): require full consumption, else a stray '/', '?' or '#' is dropped.
+    const char *astart = authority.data();
+    const char *aend   = authority.data() + authority.length();
+
+    if (url_parse_internet(header.m_heap, header.m_http->u.req.m_url_impl, &astart, aend, true, true) != ParseResult::DONE ||
+        astart != aend) {
+      return ParseResult::ERROR;
+    }
 
     if (!is_connect_method) {
       MIMEField *host = header.field_find(static_cast<std::string_view>(MIME_FIELD_HOST));
@@ -229,14 +237,26 @@ VersionConverter::_convert_req_from_2_to_1(HTTPHdr &header) const
     // :path
     if (MIMEField *field = header.field_find(PSEUDO_HEADER_PATH);
         field != nullptr && field->value_is_valid(is_control_BIT | is_ws_BIT)) {
-      auto path{field->value_get()};
+      auto  path{field->value_get()};
+      auto *url = header.m_http->u.req.m_url_impl;
 
-      // cut first '/' if there, because `url_print()` add '/' before printing path
-      if (path.starts_with("/"sv)) {
+      // Split as url_parse_http() would, so cache keys and remap see the same fields.
+      if (auto hpos = path.find('#'); hpos != std::string_view::npos) {
+        url->set_fragment(header.m_heap, path.substr(hpos + 1), true);
+        path = path.substr(0, hpos);
+      }
+
+      if (auto qpos = path.find('?'); qpos != std::string_view::npos) {
+        url->set_query(header.m_heap, path.substr(qpos + 1), true);
+        path = path.substr(0, qpos);
+      }
+
+      // url_parse_http() strips every leading '/'; url_print() re-adds exactly one.
+      while (path.starts_with("/"sv)) {
         path.remove_prefix(1);
       }
 
-      header.m_http->u.req.m_url_impl->set_path(header.m_heap, path, true);
+      url->set_path(header.m_heap, path, true);
 
       header.field_delete(field);
     } else {
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -600,7 +600,6 @@ HttpSM::state_read_client_request_header(int event, void *data)
   case VC_EVENT_ACTIVE_TIMEOUT:
     // The user agent is hosed.  Close it &
     //   bail on the state machine
-    _pre_parsed_ua_request = nullptr; // the stream is going away; the borrow would dangle
     vc_table.cleanup_entry(_ua.get_entry());
     _ua.set_entry(nullptr);
     set_ua_abort(HttpTransact::ABORTED, event);
@@ -621,23 +620,22 @@ HttpSM::state_read_client_request_header(int event, void *data)
   // tokenize header //
   /////////////////////
 
-  ParseResult state;
+  ParseResult       state;
+  ProxyTransaction *ua_txn = _ua.get_txn();
 
-  if (_pre_parsed_ua_request != nullptr) {
+  if (ua_txn->supports_direct_header_passing() && ua_txn->is_parsed_receive_header_ready()) {
     // UA_FIRST_READ never fires here: the read buffer stays empty.
     if (milestones[TS_MILESTONE_UA_FIRST_READ] == 0) {
       ATS_PROBE1(milestone_ua_first_read, sm_id);
       milestones[TS_MILESTONE_UA_FIRST_READ] = ink_get_hrtime();
     }
-    t_state.hdr_info.client_request.copy(_pre_parsed_ua_request);
-    _pre_parsed_ua_request = nullptr;
-    bytes_used             = t_state.hdr_info.client_request.length_get();
-    state                  = ParseResult::DONE;
+    t_state.hdr_info.client_request.copy(ua_txn->parsed_receive_header());
+    bytes_used = t_state.hdr_info.client_request.length_get();
+    state      = ParseResult::DONE;
   } else {
-    state = t_state.hdr_info.client_request.parse_req(&http_parser, _ua.get_txn()->get_remote_reader(), &bytes_used,
-                                                      _ua.get_entry()->eos, t_state.http_config_param->strict_uri_parsing,
-                                                      t_state.http_config_param->http_request_line_max_size,
-                                                      t_state.http_config_param->http_hdr_field_max_size);
+    state = t_state.hdr_info.client_request.parse_req(
+      &http_parser, ua_txn->get_remote_reader(), &bytes_used, _ua.get_entry()->eos, t_state.http_config_param->strict_uri_parsing,
+      t_state.http_config_param->http_request_line_max_size, t_state.http_config_param->http_hdr_field_max_size);
   }
 
   client_request_hdr_bytes += bytes_used;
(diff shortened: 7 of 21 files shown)


src/proxy/hdrs/URL.cc
```diff
@@ -1188,6 +1188,22 @@ url_is_mostly_compliant(const char *start, const char *end)
 } // namespace UrlImpl
 using namespace UrlImpl;
 
+bool
+url_is_uri_compliant(int strict_uri_parsing, std::string_view value)
+{
+  const char *start = value.data();
+  const char *end   = start + value.length();
```
JosiahWI, line 1219: You could alternatively use `value.begin` and `value.end`, but what you did is correct.

include/proxy/hdrs/URL.h
```diff
@@ -215,6 +215,9 @@ ParseResult url_parse_http(HdrHeap *heap, URLImpl *url, const char **start, cons
                            bool verify_host_characters);
 ParseResult url_parse_http_regex(HdrHeap *heap, URLImpl *url, const char **start, const char *end, bool copy_strings);
 
+// strict_uri_parsing 0 = no check; 1/2 apply the same test url_parse() does.
+bool url_is_uri_compliant(int strict_uri_parsing, std::string_view value);
```
JosiahWI, line 219: It's out of scope for this PR, but follow-up work should update the `url_parse` API to take a `std::string_view` for consistency (and then it can call this helper).

include/proxy/http/HttpSM.h
```diff
@@ -199,6 +199,20 @@ class HttpSM : public Continuation, public PluginUserArgs<TS_USER_ARGS_TXN>
 
   void attach_client_session(ProxyTransaction *txn);
 
+  // Borrowed: must outlive the copy in state_read_client_request_header. Lets HTTP/2 skip serialize+reparse.
+  void
+  set_pre_parsed_ua_request(HTTPHdr *hdr)
```
JosiahWI: It's out of scope for this PR, but if performance allows, it would be nice to take a `std::shared_ptr` here. The comment is good.

src/proxy/hdrs/VersionConverter.cc
```diff
@@ -201,7 +201,16 @@ VersionConverter::_convert_req_from_2_to_1(HTTPHdr &header) const
   if (MIMEField *field = header.field_find(PSEUDO_HEADER_AUTHORITY);
       field != nullptr && field->value_is_valid(is_control_BIT | is_ws_BIT)) {
     auto authority{field->value_get()};
-    header.m_http->u.req.m_url_impl->set_host(header.m_heap, authority, true);
+
+    // Match the reparse so url_print() stays byte-identical (legacy path
+    // unchanged); require full consumption, else a stray '/', '?' or '#' is dropped.
+    const char *astart = authority.data();
```
JosiahWI: It's still not quite clear to me why the `url_print` output would have changed.

src/proxy/http2/Http2Stream.cc
```diff
@@ -326,6 +335,37 @@ Http2Stream::send_headers(Http2ConnectionState & /* cstate ATS_UNUSED */)
...
+      this->signal_read_event(VC_EVENT_READ_READY);
+    }
+    // Delivery is synchronous (stream mutex): the borrow is copied, or the txn
+    // finished and _sm is null. A still-pending borrow is unreachable; recover anyway.
+    if (this->_sm == nullptr || !this->_sm->has_pending_pre_parsed_ua_request()) {
+      return;
+    }
+    ink_assert(!"pre-parsed handoff was deferred");
+    this->_sm->set_pre_parsed_ua_request(nullptr);
+  }
```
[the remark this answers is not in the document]
JosiahWI: This one needs to be checked.

src/proxy/http/HttpSM.cc
```diff
@@ -627,10 +628,24 @@ HttpSM::state_read_client_request_header(int event, void *data)
...
+    t_state.hdr_info.client_request.copy(_pre_parsed_ua_request);
+    _pre_parsed_ua_request = nullptr;
+    bytes_used             = t_state.hdr_info.client_request.length_get();
+    state                  = ParseResult::DONE;
+  } else {
+    state = t_state.hdr_info.client_request.parse_req(&http_parser, _ua.get_txn()->get_remote_reader(), &bytes_used,
+                                                      _ua.get_entry()->eos, t_state.http_config_param->strict_uri_parsing,
+                                                      t_state.http_config_param->http_request_line_max_size,
+                                                      t_state.http_config_param->http_hdr_field_max_size);
+  }
```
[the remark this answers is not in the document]
JosiahWI: Validation should be applied at parse-time, not on the use of the cached value. The cached value should invariantly already be validated.

include/proxy/hdrs/HdrHeap.h
```diff
@@ -175,6 +175,7 @@ struct StrHeapDesc {
 class HdrHeap
 {
 public:
+  // NOTE: also sizes an on-stack HEADERS encode buffer (2*this) in HTTP/2; a static_assert bounds the growth.
```
JosiahWI: Good comment.

src/proxy/http2/Http2ConnectionState.cc
```diff
@@ -452,11 +452,22 @@ Http2ConnectionState::rcv_headers_frame(const Http2Frame &frame)
   if (stream->trailing_header_is_possible()) {
     // Don't leak the header_blocks from the initial, non-trailing headers.
     ats_free(stream->header_blocks);
+    stream->header_blocks = nullptr;
```
JosiahWI, line 458: Setting `nullptr` is necessary semantically, because `header_blocks` might not be touched after this if the header block can be read in-place.

src/proxy/http2/Http2Stream.cc
```diff
@@ -283,10 +283,16 @@ Http2Stream::main_event_handler(int event, void *edata)
...
+}
+
+Http2ErrorCode
+Http2Stream::decode_header_blocks(HpackHandle &hpack_handle, uint32_t maximum_table_size, const uint8_t *block, uint32_t block_len)
 {
   Http2ErrorCode error =
-    http2_decode_header_blocks(&_receive_header, (const uint8_t *)header_blocks, header_blocks_length, nullptr, hpack_handle,
-                               _trailing_header_is_possible, maximum_table_size, this->is_outbound_connection());
+    http2_decode_header_blocks(&_receive_header, block, block_len, nullptr, hpack_handle, _trailing_header_is_possible,
+                               maximum_table_size, this->is_outbound_connection());
```
JosiahWI: What's the reason for adding the 4-argument overload?

src/proxy/http2/Http2ConnectionState.cc
```diff
@@ -2472,9 +2487,11 @@ Http2ConnectionState::send_headers_frame(Http2Stream *stream)
...
   }
 
-  uint32_t        buf_len = send_hdr->length_get() * 2; // Make it double just in case
-  ts::LocalBuffer local_buffer(buf_len);
-  uint8_t        *buf = local_buffer.data();
+  uint32_t buf_len = send_hdr->length_get() * 2; // Make it double just in case
+  // Keep typical header sets off the heap; oversized ones spill to it.
+  static_assert(HdrHeap::DEFAULT_SIZE * 2 <= 8192, "keep HEADERS encode stack buffer within the event-thread stack budget");
+  ts::LocalBuffer<uint8_t, HdrHeap::DEFAULT_SIZE * 2> local_buffer(buf_len);
+  uint8_t                                            *buf = local_buffer.data();
```
JosiahWI, line 2536: Do I understand correctly that you are increasing the buffer capacity here to increase the chance that typical response headers will fit in the buffer?

src/proxy/http/HttpSM.cc
```diff
@@ -627,10 +628,24 @@ HttpSM::state_read_client_request_header(int event, void *data)
...
+  if (_pre_parsed_ua_request != nullptr) {
+    // The UA_FIRST_READ gate above never fires on this path: the read buffer stays empty.
+    if (milestones[TS_MILESTONE_UA_FIRST_READ] == 0) {
+      ATS_PROBE1(milestone_ua_first_read, sm_id);
+      milestones[TS_MILESTONE_UA_FIRST_READ] = ink_get_hrtime();
+    }
+    t_state.hdr_info.client_request.copy(_pre_parsed_ua_request);
+    _pre_parsed_ua_request = nullptr;
+    bytes_used             = t_state.hdr_info.client_request.length_get();
+    state                  = ParseResult::DONE;
```
JosiahWI: This computation of `bytes_used` is good. It is more accurate than the `parse_req` path, even.

src/proxy/hdrs/VersionConverter.cc
JosiahWI: It seems to me like these are fixes to the HTTP/2 to HTTP/1.1 header conversion, which aren't directly part of the optimization itself. Is that right?

zwoop:
@JosiahWI Note that most of this got rewritten again, from suggestions from Masakazu. We'll have to rerun copilot etc.

src/proxy/http2/Http2ConnectionState.cc
zwoop: no, it's setting the local (stack) buffer to 2x the default header heap size. So, for the "common" case, where things (hopefully) would have fit in a freelist header_heap, it now stays on the stack. The assert it to assure that future changes to the header_heap default size doesn't blow the stack

zwoop:
[approve ci autest]

src/proxy/http2/Http2Stream.cc
```diff
@@ -830,15 +896,36 @@ Http2Stream::update_write_request(bool call_update)
...
+        this->_send_header.field_attach(f);
+      }
+      this->_sm->clear_pending_send_header();
+      state = ParseResult::DONE;
+    } else if (this->is_outbound_connection()) {
       state = this->_send_header.parse_req(&http_parser, this->_send_reader, &bytes_used, false);
     } else {
-      state = this->_send_header.parse_resp(&http_parser, this->_send_reader, &bytes_used, false);
+      state = ParseResult::CONT;
     }
```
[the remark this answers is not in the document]
zwoop: Reopening — confirmed regression, good catch. On the inbound path I dropped the `parse_resp()` fallback in favor of the new direct-passing setter (`get_client_response_header()`), but `HttpSM::setup_100_continue_transfer()` still serializes the interim response via `write_header_into_buffer()` instead of `write_response_header_into_buffer()`. So `_client_response_header_is_ready` is never set, `send_hdr == nullptr` in `update_write_request()`, and the serialized 100/103 sitting in `_send_reader` falls into the new `else { state = ParseResult::CONT; }` branch and is never emitted as a HEADERS frame.

Since 103 Early Hints and Expect:100-continue both route through `handle_100_continue_response` -> `INTERNAL_100_RESPONSE` -> `setup_100_continue_transfer`, this is the root cause of the `post-continue`, `early_hints`, and `httpbin` (Expect:100-continue) autest failures. Fix: restore the `parse_resp()` fallback for serialized interim responses (or route `setup_100_continue_transfer` through the direct-passing setter).

src/proxy/http/HttpSM.cc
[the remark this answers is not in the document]
zwoop: Reopening to close the loop on validation parity. The fast path does re-apply the checks that matter most: `strict_uri_parsing` via `url_is_uri_compliant()` on path/query/fragment, host-character verification via `url_parse_internet(..., verify_host_characters=true)`, and a failed 2->1.1 conversion falls back to the serialize+parse path (invalid method -> 400).

Two `parse_req` checks are *not* reproduced on the fast path: the full `ParseRules::is_token` method check (we only reject control/WS chars via `value_is_valid(is_control_BIT | is_ws_BIT)`, not separators) and `validate_hdr_content_length` (duplicate/conflicting Content-Length). CR/LF are control chars and are rejected, so this is not a request-smuggling vector, but the method-token and Content-Length parity gaps are real. Need to confirm the H2 layer already rejects conflicting Content-Length before this point; will tighten the method check to `is_token` otherwise.

src/proxy/http/HttpSM.cc
```diff
@@ -627,10 +627,23 @@ HttpSM::state_read_client_request_header(int event, void *data)
...
+    t_state.hdr_info.client_request.copy(ua_txn->parsed_receive_header());
+    bytes_used = t_state.hdr_info.client_request.length_get();
+    state      = ParseResult::DONE;
+  } else {
+    state = t_state.hdr_info.client_request.parse_req(
+      &http_parser, ua_txn->get_remote_reader(), &bytes_used, _ua.get_entry()->eos, t_state.http_config_param->strict_uri_parsing,
+      t_state.http_config_param->http_request_line_max_size, t_state.http_config_param->http_hdr_field_max_size);
+  }
 
   client_request_hdr_bytes += bytes_used;
```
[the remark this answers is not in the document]
zwoop: Confirmed. On the fast path `bytes_used = client_request.length_get()` is the full header-set size, and the ParseResult::ERROR handler then uses `bytes_used` as a request-line-length surrogate, so an oversized H2 header set gets mapped to 414 (URI Too Long) instead of a generic 400. Fix: keep the `REQUEST_URI_TOO_LONG` heuristic off the fast path — track request-line length separately, or gate that branch on the serialize path.

src/proxy/http2/Http2Stream.cc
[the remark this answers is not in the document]
zwoop: This no longer applies after the rewrite. The `set_pre_parsed_ua_request()` mechanism and its `ink_assert(!"pre-parsed handoff was deferred")` are gone; handoff is now a flag on the stream (`_is_parsed_receive_header_ready`) that the SM pulls via `parsed_receive_header()`. A deferred `signal_read_event()` (MUTEX_TRY_LOCK miss -> reschedule via `_read_vio_event`) just leaves the flag set for the next read — no assert, no crash. Keeping resolved.

src/proxy/hdrs/VersionConverter.cc
zwoop: The fast path skips the serialize+reparse round-trip, and that reparse used to normalize the URL a particular way. `url_print()` output — and therefore cache keys and remap matching — has to stay byte-identical, so the 2->1.1 converter now reproduces that normalization directly: full `:authority` consumption via `url_parse_internet` (erroring on a stray `/ ? #`), splitting `:path` into path/query/fragment, and stripping the leading `/` that `url_print` re-adds. And yes to your follow-up: these are 2->1.1 conversion fixes required to make the optimization behavior-preserving, not the optimization itself.

src/proxy/http2/Http2Stream.cc
zwoop: The explicit `(block, block_len)` overload lets us decode the header block in place from the contiguous HPACK buffer, avoiding the per-request malloc/memcpy/free into `header_blocks` that the member-based signature forced. That in-place decode is one of the main allocation savings in this PR.

include/proxy/http/HttpSM.h
zwoop: Moot after the rewrite — `set_pre_parsed_ua_request(HTTPHdr*)` is gone. The handoff is now via the `ProxyTransaction` interface, and the stream owns `_receive_header` for the lifetime of the borrow, so there is no ownership question left to solve with `shared_ptr`. Keeping resolved.

include/proxy/hdrs/URL.h
zwoop: Agreed — moving `url_parse` to `std::string_view` (so it can reuse `url_is_uri_compliant`) is a good cleanup, but out of scope for this PR. Good follow-up.

src/proxy/hdrs/URL.cc
```diff
@@ -1188,6 +1188,22 @@ url_is_mostly_compliant(const char *start, const char *end)
...
+
+  switch (strict_uri_parsing) {
+  case 1:
+    return url_is_strictly_compliant(start, end);
+  case 2:
+    return url_is_mostly_compliant(start, end);
+  default:
+    return true;
+  }
+}
```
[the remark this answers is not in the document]
zwoop: Fixed. `url_is_uri_compliant()` now returns early on an empty view before touching `value.data()`, so the `{nullptr, 0}` that `URLImpl::get_query()`/`get_fragment()` return for an unset component no longer reaches the pointer arithmetic.

src/proxy/http2/Http2Stream.cc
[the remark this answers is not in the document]
zwoop: Fixed in the amended commit — the inbound `else` branch calls `parse_resp()` again instead of `state = ParseResult::CONT`, so serialized interim/legacy responses are emitted as HEADERS frames. Final responses still take the direct-passing `send_hdr` path.

src/proxy/http/HttpSM.cc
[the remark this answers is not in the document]
zwoop: Fixed. The fast path now sets a `direct_header_passed` flag, and the `REQUEST_URI_TOO_LONG` branch in the `ParseResult::ERROR` handler is gated on `!direct_header_passed` — an oversized H2 header set now returns a generic 400 rather than 414.

src/proxy/http/HttpSM.cc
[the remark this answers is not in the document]
zwoop: Addressed. The fast-path gate in `Http2Stream::send_headers()` now also requires `parse_req_would_accept()`: the method must be all-token (`ParseRules::is_token`), and any `Content-Length` must be a single `1*DIGIT` value with no differing duplicate and no accompanying `Transfer-Encoding`. Anything else falls back to serialize+parse, where `parse_req`/`validate_hdr_content_length` adjudicate it exactly as before. Request-target and host-character checks were already re-applied via `url_is_uri_compliant()` and `url_parse_internet(..., verify_host_characters=true)`.

src/proxy/hdrs/URL.cc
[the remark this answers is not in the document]
JosiahWI: Dowvoting Copilot - Copilot is wrong. Adding a 0 offset to `nullptr` is defined behavior. See [this SO answer](https://stackoverflow.com/a/59409094)  which references the standard.

include/proxy/http/HttpSM.h
```diff
@@ -707,13 +730,30 @@ HttpSM::get_cache_sm()
...
 HttpSM::write_response_header_into_buffer(HTTPHdr *h, MIOBuffer *b)
 {
-  if (t_state.client_info.http_version == HTTPVersion(0, 9)) {
+  if (_ua.get_txn()->supports_direct_header_passing()) {
+    // Nothing lands in the buffer, so 0 keeps the tunnel's byte math honest.
+    _client_response_header_is_ready = true;
+    _direct_response_hdr_bytes       = h->length_get();
+    return 0;
+  } else if (t_state.client_info.http_version == HTTPVersion(0, 9)) {
     return 0;
```
JosiahWI, line 745: Is this a dead code path now? Can we remove it?

include/proxy/http2/Http2Stream.h
```diff
@@ -81,7 +81,14 @@ class Http2Stream : public ProxyTransaction
   bool expect_receive_trailer() const override;
   void set_expect_receive_trailer() override;
 
+  bool           supports_direct_header_passing() const override;
+  bool           is_parsed_receive_header_ready() const override;
+  const HTTPHdr *parsed_receive_header() const override;
+  bool           has_pending_send_header() const override;
+
   Http2ErrorCode decode_header_blocks(HpackHandle &hpack_handle, uint32_t maximum_table_size);
+  Http2ErrorCode decode_header_blocks(HpackHandle &hpack_handle, uint32_t maximum_table_size, const uint8_t *block,
```
JosiahWI: Nitpick: you could consider using `std::span` here.

JosiahWI (changes requested):

include/proxy/http2/Http2Stream.h
zwoop: I think I did, and then it didn't compile on some of our older platforms. But I'll try again.

src/proxy/hdrs/URL.cc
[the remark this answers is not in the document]
zwoop: You're right, and I was wrong to act on this. `nullptr + 0` is well-defined — [expr.add]/4 says adding 0 to a null pointer yields a null pointer, so `start + value.length()` with an empty view was never UB. The guard is reverted; the empty case already fell out of the `i < end` loop conditions in `url_is_strictly_compliant()` / `url_is_mostly_compliant()`, so behavior is identical either way and the branch was pure noise.

My mistake was taking the Copilot finding at face value instead of checking the standard. Thanks for the citation.

src/proxy/http2/Http2Stream.cc
```diff
@@ -326,6 +365,50 @@ Http2Stream::send_headers(Http2ConnectionState & /* cstate ATS_UNUSED */)
...
+    }
+    if (MIMEField *cl = _receive_header.field_find(static_cast<std::string_view>(MIME_FIELD_CONTENT_LENGTH)); cl != nullptr) {
+      auto value{cl->value_get()};
+      if (cl->has_dups() || value.empty() || std::any_of(value.begin(), value.end(), [](char c) { return c < '0' || c > '9'; }) ||
+          _receive_header.field_find(static_cast<std::string_view>(MIME_FIELD_TRANSFER_ENCODING)) != nullptr) {
+        return false;
+      }
+    }
+    return true;
+  };
```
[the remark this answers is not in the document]
zwoop: Confirmed and fixed. `url_parse_internet()` splits `user@host` into userinfo + host and validates only the host part, so the conversion succeeded and the Host header kept the raw `:authority` value — which then never reached `validate_hdr_host()`. The legacy path 400s it because `http_parse_host_header()` → `validate_host_name()` rejects '@' (not in `is_host_char`).

Rather than hand-roll the character check, `parse_req_would_accept()` now calls `http_parse_host_header()` on the Host value and rejects duplicates, which is exactly what `validate_hdr_host()` does — so parity is by construction rather than by a second implementation that can drift. Such requests fall back to serialize+parse and get the same 400 as before.

Worth noting the userinfo is itself an RFC 9113 8.3.1 violation (`:authority` MUST NOT include it), so rejecting is correct on two counts.

include/proxy/http/HttpSM.h
zwoop: Which of the two branches do you mean? Neither looks dead to me, so I'd rather ask than guess:

- The `supports_direct_header_passing()` branch is the new H2 path — it's what replaces the serialize for HTTP/2 responses.
- The `HTTPVersion(0, 9)` branch below is still reachable: `HttpSM.cc:7545` branches on `t_state.client_info.http_version == HTTP_0_9`, and `:7208` asserts it is *not* 0.9, so the value does occur for inbound clients.

If you meant something else — e.g. that the `MIOBuffer *b` parameter is now unused on the direct path — say so and I'll take another look.

include/proxy/http2/Http2Stream.h
zwoop: I tried again, and I'd rather not in this PR. `std::span<const uint8_t>` does compile locally (clang, C++20), but there is currently **no** `std::span` anywhere in `src/`, `include/`, or `plugins/` — this would be the first use in the tree, which matches my recollection of it not surviving the older platforms in CI.

Introducing it here means a portability question decided by a perf PR rather than on its own merits, so I've left the `(block, block_len)` pair. Happy to see it done tree-wide as its own change — the two-arg overload is a mechanical conversion at that point.

zwoop:
[approve ci autest]

JosiahWI (approved):

bryancall (changes requested):
I read the whole diff and spent most of the time on the parts that could go wrong quietly. The core of this is sound, and a couple of pieces are better than I expected:

- The riskiest change, decoding HPACK straight out of the frame reader's IOBuffer instead of a stream-owned allocation that lived until stream teardown, is actually safe. `MIMEFieldWrapper::value_set`/`name_set` copy into the HdrHeap and `XpackDynamicTable::insert_entry` copies via `_storage.write()`, so nothing retains a pointer into the frame buffer. The in-place branch is correctly restricted to `end_headers`, CONTINUATION still takes the realloc path, and `header_blocks` is explicitly nulled on the trailer path so the destructor's `ats_free` stays correct.
- The URL normalization parity work is careful and I could not find a divergence. The `#` split before the `?` split reproduces `url_parse_http`'s scan for inputs like `/p#f?q`, and `while (path.starts_with("/"))` matches the remove-preceding-slashes loop, where a single `remove_prefix` would have changed the cache key for `//foo`.
- Replacing `set_host(authority)` with `url_parse_internet()` plus a full-consumption check closes a real hole. Previously `:authority: example.com/foo` was stored whole as the host and reparsed into host plus path. It now returns `ParseResult::ERROR` and gets a 400, and it brings H2 CONNECT in line with how HTTP/1.1 parses authority-form targets.
- The byte-accounting split is thought through rather than a blanket rename, with readers that mean "bytes the client received" using the reported accessor and readers doing tunnel arithmetic correctly keeping the raw counter.
- The unrelated `http2.test.py` fix is worth having on its own. `|&` is bash 4 syntax and autest runs through `/bin/sh`, which is bash 3.2 on macOS, so that case was dying on a syntax error before ever reaching ATS.

Requesting changes on two things, neither of which is a defect in the merged code path.

## 1. Leftover debugging file

`tests/gold_tests/h2/http2_fc_iso.test.py`

This is a 348-line scratch copy. I reconstructed it from the patch and diffed it against `tests/gold_tests/h2/http2_flow_control.test.py`: only 85 lines differ out of 348, and every difference is a deletion. The docstring is "Isolation repro: HTTP/2 flow control, policy 0, 500-byte window", the whole `Http2FlowControlTest` class is byte-identical including the "paramenter" typos, and the single surviving invocation is character-for-character the same one that already exists at the tail of the original file. So it is pure duplication with zero new coverage, it tests flow control rather than header handling, and it carries `proxy.config.diags.debug.enabled: 3` with the `http` tag, which will produce large `traffic.out` files on every CI run and for every developer running the h2 gold tests. It is mentioned in neither the PR body nor any of the five commits.

It should just be deleted.

## 2. The fast-path rejection gates have no negative test

`tests/gold_tests/h2/replay/h2_request_handling.replay.yaml`

The gates added in 86afd77 are the entire safety argument for skipping `parse_req`: the token method check, `http_parse_host_header` on the synthesized Host, the single `1*DIGIT` Content-Length with no Transfer-Encoding, and `strict_uri_parsing`. The Host gate in particular is load-bearing, since `http_parse_host_header` calls `validate_host_name`, which rejects the `@` in `user@host`, which is exactly the case your commit message says it exists for.

Every proxy-response in the added replay is 200, 204 or 304. There is no expected 400 anywhere in the PR's test additions, and grepping the diff for `user@`, `userinfo` or `Transfer-Encoding` in test files finds nothing but Content-Length values on happy-path bodies. So if someone later reorders the `&&` chain or hoists `parse_req_would_accept()` behind a flag, a request with `:authority: attacker@origin.example.com` reaches the origin unvalidated and CI stays fully green. That is the bug Copilot found and you fixed, silently reintroducible.

One negative scenario per gate would close it.

## Also worth a note

**`src/proxy/http/HttpSM.cc:724`** `proxy.config.http.request_line_max_size` silently stops applying to HTTP/2 on the fast path, since `parse_req` never runs and its limit argument is never consulted. An operator who sets it to 1024 gets a 4KB path rejected over HTTP/1.1 and proxied over HTTP/2, same URL and same ATS, with no log line explaining the difference. `records.yaml.en.rst:1465` still states the limit without qualification. Documentation or a release note rather than a code change, I think, but the scope change should be written down somewhere.

The branch is also conflicting with master and will need a rebase.

## Smaller observations, not blocking

- `include/proxy/http/HttpSM.h:646` `clear_pending_send_header()` zeroes both the client-response and server-request ready flags, and the inbound and outbound streams share one HttpSM. Nothing in the current sequencing lets both be set at once, so this is latent, but it is a sharp edge for whoever next adds an internal retry or redirect after the response header is written.
- `include/proxy/http/HttpSM.h:640` `reported_client_response_hdr_bytes()` uses "raw counter is nonzero" as its sentinel for "the header went through the tunnel", which `setup_blind_tunnel` violates by adding prewarm body bytes to that same counter. With `proxy.config.tunnel.prewarm` on, `pshl` and `TSHttpTxnClientRespHdrBytesGet` can report origin body bytes as response header size.
- `src/proxy/hdrs/VersionConverter.cc:207` `url_parse_internet()` is called with source pointers into `header.m_heap` while it writes host and port back into that same heap. `set_host(heap, host, true)` runs before `set_port(heap, port, true)`, and `HdrHeap::allocate_str` can reach `coalesce_str_heaps()`, which drops the last reference to the old storage. I could not construct a reachable case, since it needs more than `MAX_LOST_STR_SPACE` of freed string space accumulated during HPACK decode, but the aliasing shape is there and worth a second opinion from someone who knows the heap better than I do.

bryancall:
Checking in on this one. It has been a month since I requested changes, and the branch has gone to conflicting with master in the meantime, so it needs a rebase before it can be evaluated again.

Both things I asked for are small:

1. Delete `tests/gold_tests/h2/http2_fc_iso.test.py`. It is a 348-line scratch copy of `http2_flow_control.test.py` with no new coverage, and it carries `proxy.config.diags.debug.enabled: 3`, so it grows `traffic.out` on every CI run and for everyone running the h2 gold tests locally.
2. One negative scenario per fast-path gate in `h2_request_handling.replay.yaml`. Every proxy-response in the added replay is 200, 204 or 304, so there is no test that fails if the `&&` chain is later reordered and `:authority: attacker@origin.example.com` reaches the origin unvalidated.

None of that is a defect in the merged code path, and I want to be clear the review was positive on the substance: I read the whole diff, the in-place HPACK decode out of the frame reader is genuinely safe, and the authority parsing change closes a real hole.

I would rather this land than rot. If you are short of time, say so and I will look at whether the test gap can reasonably move to a follow-up issue so the rest is not held up. Are you still planning to take this one forward?

PR #13424, opened 2026-07-23
header_rewrite: add sort-destination operator

juanthropic:
Add a `sort-destination QUERY` operator to header_rewrite that stably
sorts the request URL's query parameters by name. Normalizing parameter
order lets varied client query strings map to one origin request, which
helps origin-side caching and dedup.

The sort logic lives in a new pure, TS-API-free `sort_query()` helper
with its own unit test, kept separate from the operator so the
cache-key and parent-key operators can reuse it in a later PR.

- `sort_query()` uses a stable sort, so duplicate-named parameters keep
  their relative order.
- Empty tokens from leading, trailing, or consecutive `&` are dropped.
- Docs note that `QUERY` is the only supported part; any other part is a
  no-op, matching the sibling `rm-destination` wording.

Covered by a Catch2 unit test for the helper and an end-to-end Proxy
Verifier test that checks an unsorted client query reaches the origin
sorted.

### DIFF
--- a/doc/admin-guide/plugins/header_rewrite.en.rst
+++ b/doc/admin-guide/plugins/header_rewrite.en.rst
@@ -1421,6 +1421,17 @@ the operator will effectively be a no-op.
     This operator is deprecated, use the `set-http-cntl`_ operator instead,
     with the ``SKIP_REMAP`` control.
 
+sort-destination
+~~~~~~~~~~~~~~~~
+::
+
+  sort-destination QUERY
+
+Sorts the query parameters of the remapped destination's URL by parameter
+name. Sorting is stable, so parameters that share the same name keep their
+relative order. Currently ``QUERY`` is the only valid part for
+sort-destination.
+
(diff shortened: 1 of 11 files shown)


moonchen (commented):
Do you only want a cache key that doesn't change depending on the ordering of parameters, or do you also want the origin server to see sorted parameters?

If it's only the first one you need, I think an additive hash of parameters will drop the ordering of parameters more efficiently, with fewer allocations.

If you also need the sorted parameters to go to the origin, then one suggestion is to check if the parameters are already in sorted order before allocating memory to sort them.

In either case, I want to also point out that cachekey.so with --sort-params=true deduplicates parameters by name, so this behavior is different.

zwoop:
@moonchen can you please shepherd this PR through the reviews and changes we discussed please?

juanthropic:
Sorry for the delay in moving this PR forward. I'll get to this next week.

PR #13425, opened 2026-07-24
header_rewrite: reject bad run-plugin at config load

zwoop:
A run-plugin whose target plugin failed to load left a null instance that tripped a release assert and aborted the server on the first request. Propagate the load failure as an exception so the rule is rejected at config load time (a reload simply keeps the running config), and guard exec() so a stray bad rule can never abort the process.

### DIFF
--- a/plugins/header_rewrite/operators.cc
+++ b/plugins/header_rewrite/operators.cc
@@ -23,6 +23,7 @@
 #include <cstring>
 #include <algorithm>
 #include <iomanip>
+#include <stdexcept>
 
 #include "records/RecCore.h"
 #include "ts/ts.h"
@@ -1282,8 +1283,7 @@ OperatorRunPlugin::initialize(Parser &p)
   auto plugin_args = p.get_value();
 
   if (plugin_name.empty()) {
-    TSError("[%s] missing plugin name", PLUGIN_NAME);
-    return;
+    throw std::runtime_error("run-plugin missing plugin name");
   }
 
   std::vector<std::string> tokens;
@@ -1321,7 +1321,7 @@ OperatorRunPlugin::initialize(Parser &p)
   delete[] argv;
 
   if (!_plugin) {
-    TSError("[%s] Unable to load plugin '%s': %s", PLUGIN_NAME, plugin_name.c_str(), error.c_str());
+    throw std::runtime_error("run-plugin unable to load plugin '" + std::string{plugin_name} + "': " + error);
   }
 }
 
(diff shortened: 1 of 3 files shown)


zwoop:
[approve ci]

plugins/header_rewrite/operators.cc
```diff
@@ -1282,8 +1283,7 @@ OperatorRunPlugin::initialize(Parser &p)
   auto plugin_args = p.get_value();
 
   if (plugin_name.empty()) {
-    TSError("[%s] missing plugin name", PLUGIN_NAME);
-    return;
+    throw std::runtime_error("run-plugin missing plugin name");
```
cmcfarlen, line 1286: Did Claude do this? We typically don't throw exceptions.

bneradt (commented):
I found two exception-safety/resource-lifetime issues that should be addressed:

### Make the temporary argv exception-safe

`OperatorRunPlugin::initialize()` allocates `argv` with `new[]`, calls `getRemapPlugin()`, and only then reaches `delete[]`. The factory and its filesystem/allocation helpers can throw standard exceptions; those are now caught by `RuleSet::add_operator()` or `OperatorIf::add_operator()`, but unwinding skips the manual delete. A rejected reload can therefore leak this allocation. Please use RAII here, such as `std::vector<char *>` or `std::unique_ptr<char *[]>`.

https://github.com/apache/trafficserver/blob/079936f921b58b8b2b216b42bf403ad0b41cd9c5/plugins/header_rewrite/operators.cc#L1297-L1324

### Clean up active nested-if frames on failure

When a bad `run-plugin` appears inside an open `if`, `OperatorIf::add_operator()` catches the initialization exception and returns false. `RulesConfig::parse_config()` then takes its runtime-error return path, but `if_stack` still contains raw `OperatorIf *` values created for active frames. Those frames have not yet been transferred into the `RuleSet`, so returning without draining the stack leaks them and any successfully initialized operators or plugin instances they already own. Repeated rejected reloads can accumulate these leaks. Please make the stack RAII-owned or clean it on all error exits; adding a nested `run-plugin` case to this test would cover the second initialization path.

https://github.com/apache/trafficserver/blob/079936f921b58b8b2b216b42bf403ad0b41cd9c5/plugins/header_rewrite/header_rewrite.cc#L369-L417

bneradt:
Superseded by #13493, which takes over this change with Leif's original authored commit and message preserved, applies the review feedback, and avoids the new exception path. Thanks, Leif!

PR #13426, opened 2026-07-24
header_rewrite: add POST_REMAP_HOOK support

juanthropic:
`header_rewrite` can attach rulesets to most transaction hooks, but not to
`TS_HTTP_POST_REMAP_HOOK`. For a global (`plugin.config`) configuration, that
leaves no hook that sees the *remapped* request *before the cache lookup*:

- `READ_REQUEST_HDR_HOOK` / `READ_REQUEST_PRE_REMAP_HOOK` run before remapping,
  so they only ever see the pristine request.
- `REMAP_PSEUDO_HOOK` sees the remapped request before the lookup, but is valid
  only in a remap context.
- `SEND_REQUEST_HDR_HOOK` sees the remapped request, but fires after the lookup
  and only when the request is forwarded to an origin. It cannot influence the
  lookup, and it never runs on a cache hit.

That window is where anything feeding the cache lookup has to run. The `cachekey`
plugin registers `TS_HTTP_POST_REMAP_HOOK` for exactly this reason
(`plugins/cachekey/plugin.cc:114`): its remap-time path (`TSRemapDoRemap`)
covers the per-remap case, and a global instance needs the remapped request
before `TSCacheUrlSet` is consumed by the lookup.

This is the hook a planned `header_rewrite` cache-key operator needs, for the
same reason: setting the cache key is only meaningful between remapping and the
cache lookup, and `SEND_REQUEST_HDR_HOOK` is already too late.

This PR adds a `POST_REMAP_HOOK` hook condition and wires it end to end:

- Recognize the `POST_REMAP_HOOK` keyword in the parser.
- Handle the POST_REMAP event so rulesets run at that hook.
- Gather the post-remap request headers for the hook.
- Allow operators and conditions on the hook.

Covered by a parser unit test and an end-to-end autest.

### DIFF
--- a/doc/admin-guide/plugins/header_rewrite.en.rst
+++ b/doc/admin-guide/plugins/header_rewrite.en.rst
@@ -1738,14 +1738,17 @@ POST_REMAP_HOOK
 ~~~~~~~~~~~~~~~
 
 Forces evaluation of the ruleset immediately after remapping has completed, but
-before |TS| contacts the origin server (or fetches the object from cache). There
-is no response data yet, so context-adapting conditions and operators match
-against the request.
+before |TS| looks the request up in the cache. There is no response data yet, so
+context-adapting conditions and operators match against the request, which at
+this point is the remapped request.
 
-This hook is available to both global (:file:`plugin.config`) and per-remap
-(:file:`remap.config`) configurations, and is primarily useful for
-globally-configured ``header_rewrite`` instances that need to act on the request
-after remapping.
+For rulesets in :file:`remap.config`, `REMAP_PSEUDO_HOOK`_ already covers this
+window. This hook exists for globally-configured rulesets, which otherwise have
+no hook that sees the remapped request before the cache lookup:
+`READ_REQUEST_HDR_HOOK`_ and `READ_REQUEST_PRE_REMAP_HOOK`_ run before
+remapping, and `SEND_REQUEST_HDR_HOOK`_ runs after the lookup, only when the
+request is forwarded to an origin. Anything that has to influence the lookup
+itself belongs at this hook.
 
 SEND_REQUEST_HDR_HOOK
 ~~~~~~~~~~~~~~~~~~~~~
--- a/doc/admin-guide/plugins/header_rewrite.en.rst
+++ b/doc/admin-guide/plugins/header_rewrite.en.rst
@@ -1734,6 +1734,19 @@ files shared by both the global :file:`plugin.config` and individual remapping
 entries in :file:`remap.config`, this hook condition will force the subsequent
 ruleset(s) to be valid only for remapped transactions.
 
+POST_REMAP_HOOK
+~~~~~~~~~~~~~~~
+
+Forces evaluation of the ruleset immediately after remapping has completed, but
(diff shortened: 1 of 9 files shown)


bneradt (commented):
Codex had one comment. Otherwise the PR looks good to me.

bneradt:
Adding TS_HTTP_POST_REMAP_HOOK only to Statement::initialize_hooks() does not affect derived operators that override initialize_hooks(). In particular, OperatorSetPluginCntl and OperatorSetStateFlag/Int8/Int16 each say they should be allowed everywhere, but their explicit hook lists omit the new hook. The session-state operators reuse those state classes, so they are affected too.

As a result, a POST_REMAP_HOOK ruleset using set-plugin-cntl, set-state-*, or set-session-* is rejected during configuration parsing. Could these overrides add TS_HTTP_POST_REMAP_HOOK (or delegate to Statement::initialize_hooks()), with a configuration-level test covering one of these operators?

juanthropic:
Thanks for the review @bneradt!

> Could these overrides add TS_HTTP_POST_REMAP_HOOK (or delegate to Statement::initialize_hooks())

Yeah definitely one or the other. The comment does say "should be allowed everywhere", so I wonder why the list repeats the base instead of delegating to `Statement::initialize_hooks()`. Perhaps @zwoop or @maskit could comment on whether it makes sense to remove the redundant override and rely on the base.

maskit:
I don't see any reasons to have separate lists for those operators. The one I added is just a copy-paste of the other one. Those lists should be removed but the removal is unrelated to this change. I'd make another PR for it.

bneradt:
I agree that removing these redundant overrides would be a useful separate cleanup. My concern does not require that cleanup in this PR, though.

`initialize_hooks()` is virtual, so when `Statement::initialize()` invokes it for `OperatorSetPluginCntl` or one of the state operators, C++ calls the derived override instead of `Statement::initialize_hooks()`. The new POST_REMAP entry in the base implementation therefore has no effect on those operators.

For example, this currently fails configuration parsing:

```text
cond %{POST_REMAP_HOOK}
  set-state-flag 0 true
```

`RuleSet::add_operator()` rejects it because `OperatorSetStateFlag::_allowed_hooks` does not contain `TS_HTTP_POST_REMAP_HOOK`. The session operators are affected too because the factory implements them using the same state classes.

The existing AuTest does not expose this because `set-header` inherits the base hook list.

If removing the overrides is out of scope, that is fine. The minimal fix here is to add `TS_HTTP_POST_REMAP_HOOK` to the four overrides marked "allowed everywhere": `OperatorSetPluginCntl`, `OperatorSetStateFlag`, `OperatorSetStateInt8`, and `OperatorSetStateInt16`. Adding `set-state-flag 0 true` to the existing POST_REMAP test ruleset would cover the parsing path.

bneradt (approved):

bneradt:
Picking up the initialize() override fix here:
https://github.com/apache/trafficserver/pull/13486

Thank you @juanthropic for working on this!

cmcfarlen:
Cherry-picked to the 10.2.x branch as 6c874f29a971326c157c8f0fa7b887b423aaa997 for the 10.2.0 release.

PR #13428, opened 2026-07-25
Restrict HTTP/2 TLS cipher suites

### LINKED ISSUE #5287: Don't allow blacklisted HTTP/2 ciphers to use HTTP/2
Migrate from JIRA: https://issues.apache.org/jira/browse/TS-3787

bneradt:
HTTP/2 can currently be negotiated with TLS 1.2 cipher suites that
RFC 9113 prohibits. Clients that enforce the restriction can reject
those connections instead of falling back to HTTP/1.1.

This checks the pending cipher during ALPN selection and removes
HTTP/2 when its IANA identifier is prohibited. This also adds AuTest
coverage for permitted negotiation and HTTP/1.1 fallback.

Fixes: #5287

### DIFF
--- a/src/iocore/net/SSLUtils.cc
+++ b/src/iocore/net/SSLUtils.cc
@@ -71,6 +71,7 @@
 #endif
 
 #include <algorithm>
+#include <cstdint>
 #include <thread>
 #include <utility>
 #include <string>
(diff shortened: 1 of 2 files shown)


src/iocore/net/SSLUtils.cc
```diff
@@ -369,6 +371,53 @@ ssl_next_protos_advertised_callback(SSL *ssl, const unsigned char **out, unsigne
...
+is_http2_prohibited_cipher(const SSL_CIPHER *cipher)
+{
+  struct CipherRange {
+    uint16_t first;
+    uint16_t last;
+  };
+
+  // RFC 9113 Appendix A lists 276 prohibited TLS 1.2 cipher suites. The IANA
+  // identifiers for those suites form these 24 contiguous ranges.
+  static constexpr CipherRange prohibited_ranges[] = {
```
masaori335, line 384: I checked with IANA registration.
https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-4

masaori335 (approved):
Looks good to me.

PR #13429, opened 2026-07-25
doc: clarify run-plugin argument is fixed at load time

maskit:
The run-plugin action's plugin-argument is parsed once when the rule loads, like a remap.config argument, and is never re-evaluated per request, so variable interpolation (e.g. %{HEADER:bar}) does not work there.


zwoop (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as a823957a6f675fc5b4f5d8505721a85f32f4f9ed for the 10.2.0 release.

PR #13431, opened 2026-07-25
Make log pipe test tolerate restricted kernels

bneradt:
Some CI environments do not permit unprivileged processes to increase
FIFO buffer sizes, even when the requested size is below the system
maximum. The log_pipe AuTest treats this kernel policy as a product
failure and cannot run in those environments.

This makes the test accept only the corresponding `Operation not
permitted` diagnostic when the FIFO remains unchanged, while continuing
to reject other errors and fatal diagnostics. It also waits briefly for
asynchronous diagnostic output before deciding whether the size check
failed.


maskit (approved):
Looks reasonable.

cmcfarlen:
Cherry-picked to the 10.2.x branch as 9bafe093f5580eccf7f2a5d6222ae9faa68fbaae for the 10.2.0 release.

PR #13433, opened 2026-07-27
Add autest for http2.incomplete_header_timeout_in

### LINKED ISSUE #11391: Add AuTest to exercise http2.incomplete_header_timeout_in
It looks like we need to find or write a client to exercise the timeout.
https://github.com/apache/trafficserver/pull/11354

bneradt:
No off-the-shelf client can withhold the END_HEADERS flag, so this adds
an ad-hoc HTTP/2 client that fragments the request header block and
optionally never sends the CONTINUATION frame. The test also covers a
CONTINUATION that arrives in time and a slow origin response, so that a
regression in canceling the timeout at transaction start is caught too.

Fixes: #11391


masaori335 (approved):
Thank you for covering the case.

PR #13438, opened 2026-07-28
Fix exact URL cache config matching

### LINKED ISSUE #13421: cache.config: Exact `url=` rules are dropped after successful parsing
## Description

`cache.config` accepts `url=` as a primary destination. A valid rule passes parsing, but `UrlMatcher` does not add it to the URL hash table. The configuration loads without an error, and the rule never matches a request.

The insertion code is in `UrlMatcher<Data, Result>::NewEntry()` in `src/proxy/ControlMatcher.cc`:

```cpp
cur_d = data_array + num_el;
error = cur_d->Init(line_info);

if (error.failed()) {
  url_str[num_el]   = ats_strdup(pattern);
  url_value[num_el] = num_el;
  url_ht.emplace(url_str[num_el], url_value[num_el]);
  num_el++;
}

return error;
```

`error.failed()` is `true` when initialization fails. This condition is backwards: a successfully initialized rule skips the insertion into `url_ht`, while the function still returns success.

## Expected behavior

When `cur_d->Init(line_info)` succeeds, ATS should add the rule to `url_ht`. The rule's cache action should apply when a request URL exactly matches the configured value.

## Cause

The exact URL matcher was added for TS-1280. The original code used a string pointer for errors and inserted the rule only after successful initialization:

```cpp
errBuf = cur_d->Init(line_info);

if (errBuf == NULL) {
  // Insert into url_ht.
}
```

Commit `70cc2d4ce1762b182dd52f43a6313c61772c3a8a` replaced the error string with `config_parse_error`. That type evaluates to `true` when an error is present, but the same change made the `UrlMatcher` condition `if (error)`, reversing the old success check.

Commit `39f8e5c057386bffd51f5ba2284d2e715543e100` later changed the condition to `if (error.failed())` without correcting the reversed logic.

Related commits:

- [[Original implementation](https://github.com/apache/trafficserver/commit/2d9ad48c5075b40ebfd864b1e9ce4f097d84b85c)](https://github.com/apache/trafficserver/commit/2d9ad48c5075b40ebfd864b1e9ce4f097d84b85c)
- [[Commit that introduced the regression](https://github.com/apache/trafficserver/commit/70cc2d4ce1762b182dd52f43a6313c61772c3a8a)](https://github.com/apache/trafficserver/commit/70cc2d4ce1762b182dd52f43a6313c61772c3a8a)
- [Conversion to `Result`](https://github.com/apache/trafficserver/commit/39f8e5c057386bffd51f5ba2284d2e715543e100)

## Suggested fix

```diff
 error = cur_d->Init(line_info);

-if (error.failed()) {
+if (!error.failed()) {
   url_str[num_el]   = ats_strdup(pattern);
   url_value[num_el] = num_el;
   url_ht.emplace(url_str[num_el], url_value[num_el]);
   num_el++;
 }
```

Regression tests should cover:

- A valid `url=` rule is inserted and matches the complete URL.
- A different URL does not match the rule.
- A failed `Init()` call does not insert a rule and returns a configuration error.
- A duplicate `url=` value still returns the duplicate-rule error.

bneradt:
Exact `url=` entries in `cache.config` are currently inserted only when record initialization fails. Successfully parsed rules are silently dropped, so their cache actions never apply, while failed rules incorrectly advance matcher state.

Correct the success check, add focused `UrlMatcher` unit coverage, and add an exact-URL replay AuTest. The replay helper now accepts `cache_config` entries with dynamic server ports so this behavior can be tested end to end.

Testing:
- Original code: `cache-exact-url` fails with `expected 201 got 200`
- Fixed code: `test_proxy "[ControlMatcher]"` passes 13 assertions
- Fixed code: `cache-exact-url` passes

Fixes #13421

### DIFF
--- a/doc/developer-guide/testing/autests.en.rst
+++ b/doc/developer-guide/testing/autests.en.rst
@@ -356,6 +356,7 @@ The ``autest`` section configures the test environment:
   - **process_config**: Parameters passed to ``MakeATSProcess`` (e.g., ``enable_cache``)
   - **records_config**: Dictionary of records.config settings
   - **remap_config**: List of remap rules (string or dict format)
+  - **cache_config**: List of cache.config rules
   - **copy_to_config_dir**: List of files/directories to copy to ATS config directory
   - **log_validation**: Log validation rules for ``traffic_out`` and ``diags_log``
   - **metric_checks**: List of metric name/value pairs to verify after traffic completes
--- a/src/proxy/ControlMatcher.cc
+++ b/src/proxy/ControlMatcher.cc
@@ -313,7 +313,7 @@ UrlMatcher<Data, MatchResult>::NewEntry(matcher_line *line_info)
   // Fill in the parameter info
   cur_d = data_array + num_el;
   error = cur_d->Init(line_info);
-  if (error.failed()) {
+  if (!error.failed()) {
     url_str[num_el]   = ats_strdup(pattern);
     url_value[num_el] = num_el;
     url_ht.emplace(url_str[num_el], url_value[num_el]);
--- a/src/proxy/unit_tests/CMakeLists.txt
+++ b/src/proxy/unit_tests/CMakeLists.txt
@@ -16,7 +16,14 @@
 #######################
 
 add_executable(
-  test_proxy main.cc test_ControlBase.cc test_FetchSM.cc test_ParentHashConfig.cc test_PluginYAML.cc stub.cc
+  test_proxy
+  main.cc
+  test_ControlBase.cc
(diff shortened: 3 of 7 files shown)


maskit (dismissed):

maskit:
Please resolve the merge conflict.

maskit (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as 92f3f3c8f2d5f41d3ad6c20a71fc09cd25e33a17 for the 10.2.0 release.

PR #13444, opened 2026-07-28
Raise HTTP/2 SETTINGS limits

### LINKED ISSUE #13443: Default HTTP/2 SETTINGS limits reject extension settings
### Problem

The default value of `proxy.config.http2.max_settings_per_frame` is 7.
That limit was introduced in 2019 and it seems too low for today's clients.

We are now seeing a large volume of production connections from Meta
forward proxies whose initial SETTINGS frame contains 11 parameters. The
66-byte frame contains the usual HTTP/2 settings plus five settings from
the WebTransport-over-HTTP/2 draft (`0x2b61` through `0x2b65`).

Traffic Server accepts the first seven entries, rejects the eighth, and
closes the connection with `ENHANCE_YOUR_CALM`:

```
HTTP/2 connection error code=0x0b stream_id=0 recv settings too many settings in a frame
```

These connections proceed to ordinary HTTP requests when the limit is
raised, so this is an interoperability problem rather than malformed
traffic.

[RFC 9113 section 6.5.2](https://www.rfc-editor.org/rfc/rfc9113.html#section-6.5.2)
allows a SETTINGS payload to contain any number of parameters and requires
unsupported identifiers to be ignored. The extension identifiers are
documented in the
[WebTransport-over-HTTP/2 draft](https://datatracker.ietf.org/doc/html/draft-ietf-webtrans-http2-14#section-11.2).

### Proposed change

Raise the default per-frame limit from 7 to 16, retaining a conservative
bound while allowing room for protocol extensions. Raise
`proxy.config.http2.max_settings_per_minute` from 14 to 32 to preserve
the existing two-frame allowance and leave room for a legitimate settings
update after the initial frame.

Add an HTTP/2 AuTest that sends an 11-parameter initial SETTINGS frame and
verifies that a request succeeds with the defaults.

bneradt:
HTTP/2 extensions can add enough SETTINGS parameters to exceed the
defaults, causing Traffic Server to close otherwise valid connections
before a request. This occurs with Meta forward proxy traffic that
advertises WebTransport-over-HTTP/2 settings.

This raises the per-frame limit to 16 and the per-minute limit to
32, retaining abuse protections while allowing protocol growth. This
also adds an AuTest that sends an extension-rich SETTINGS frame and
verifies that the subsequent request succeeds.

Fixes: #13443

### DIFF
--- a/doc/admin-guide/files/records.yaml.en.rst
+++ b/doc/admin-guide/files/records.yaml.en.rst
@@ -5078,7 +5078,7 @@ HTTP/2 Configuration
 
    This is the threshold of sampling stream number to start checking the stream error rate.
 
-.. ts:cv:: CONFIG proxy.config.http2.max_settings_per_frame INT 7
+.. ts:cv:: CONFIG proxy.config.http2.max_settings_per_frame INT 16
    :reloadable:
 
    Specifies how many settings in an HTTP/2 SETTINGS frame |TS| accepts.
@@ -5086,7 +5086,7 @@ HTTP/2 Configuration
    code of ENHANCE_YOUR_CALM.
    Any negative value configures no limit to the number of settings received.
 
-.. ts:cv:: CONFIG proxy.config.http2.max_settings_per_minute INT 14
+.. ts:cv:: CONFIG proxy.config.http2.max_settings_per_minute INT 32
    :reloadable:
 
    Specifies how many settings in HTTP/2 SETTINGS frames |TS| accept for a minute.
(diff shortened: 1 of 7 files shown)


maskit (approved):
I"m fine with the new limits (still small enough), but at the same time I'm not sure what are reasonable limits. We probably don't want to make the default values 1000 even if there is a legitimate case. The default values should be safe, and settings should be adjusted by users for minor use cases.

cmcfarlen:
Cherry-picked to the 10.2.x branch as 53f07be0aeb05eb7ef2a77f36f8d7a5077b3a614 for the 10.2.0 release.

PR #13447, opened 2026-07-29
HTTP/2 to origin session pool and reliability fixes

bneradt:
Long-lived HTTP/2 origin connections could remain discoverable after
entering half-close, become unreachable when global pooling was
configured, or be poisoned by a single request's Connection: close
header. Requests assigned to those sessions could fail immediately.
Valid responses could also be rejected when headers used CONTINUATION,
when their payload length followed HEAD or 304 rules, or when normal
SETTINGS acknowledgments crossed the receive-rate limit. Large-session
flow control also advertised the connection allowance as each stream's
receive window.

This makes outbound HTTP/2 session lifecycle consistently thread-local.
It always probes the thread pool before configured global pools, ignores
Connection: close for outbound session shutdown, refuses to re-pool
half-closed sessions, and evicts sessions as soon as they enter local
half-close. HTTP/1.x global pooling behavior remains unchanged.

This also carries the peer's not-processed assertion from GOAWAY and
REFUSED_STREAM into retry selection. Retry admission requires no
observed response and either no body or a complete replayable copy,
including during upload failures. Solicited SETTINGS acknowledgments are
excluded from the receive limit in both directions because their rate is
bounded by settings sent by Traffic Server. This patch keeps configured
stream receive windows separate from the connection allowance.

This accepts CONTINUATION frames in valid outbound stream states without
creating a second transaction and preserves request metadata needed to
validate HEAD and conditional 304 responses. Replay-based AuTests and
custom HTTP/2 origins cover session reuse, half-close handling, retries,
split response headers, payload validation, and SETTINGS accounting.
Negative cases enforce retry and frame-validation boundaries, and
counters distinguish admitted retries from unavailable body copies.

Co-authored-by: GPT-6 Astra Medium


bneradt:
[approve ci autest 1]

bneradt:
Updated the PR with further retry and SETTINGS-accounting fixes:

- Retry preparation guards a missing client transaction and records response evidence before replacing a closed stream for decoding.
- A failed thread-local session probe can fall back to the configured global pool while preserving failure accounting.
- Solicited SETTINGS ACKs are excluded from the receive limit in both directions. A new inbound test keeps three streams open and verifies the resulting ACKs are accepted.
- All 15 retry scenarios assert the admission/body-availability counters. Invalid-frame tests retain targeted log checks, and origin diagnostics identify the actual reset and GOAWAY values.

Validation: build, formatting, all 965 CTest cases, six focused AuTest files, and the Sphinx docs build pass. Mutation checks confirm the negative body test rejects an omitted replay-body gate and the inbound ACK test rejects restoration of inbound ACK counting. The exact body-length check remains a defensive invariant; a new end-to-end truncated-copy scenario is not included.

bneradt:
This PR now also fixes the stream/session receive-window initialization bug. SETTINGS_INITIAL_WINDOW_SIZE uses the configured stream window; the enlarged session window is used only for the connection WINDOW_UPDATE. Policy 1 retains the fixed stream limit, and policy 2 still adjusts stream windows dynamically after stream creation.

A new wire regression checks policies 0/1/2 at 65,535 and 32,768 bytes, including the initial SETTINGS, independent connection window, and one/two/three concurrent open streams. All four policy-1/2 cases fail against the old binary; policy-0 controls pass. All six pass with the fix.

Validation: build, formatting, 151 selected CTest cases, and eight focused AuTest files pass, including the full flow-control suite and existing H2-to-origin tests. Tests that encountered port collisions during simultaneous branch runs passed on rerun.

PR #13450, opened 2026-07-29
h2origin autest: honor the 30s squid.log wait budget

### LINKED ISSUE #13445: Intermittent AuTest failure: h2origin (squid.log not written within 10s)
The `h2origin` gold test (`tests/gold_tests/h2/h2origin.test.py`) is flaky on CI: the "Wait for the squid.log to be written" run intermittently fails because the log line does not appear within a 10-second readiness window. It passes on re-run with no code change.

```
Run: Wait for the squid.log to be written: Failed
  Starting TestRun 2-tr : Checking that Process is ready within 10.0 seconds so we can start process: Default - Failed
     Reason: Process failed to become ready in time: 10.0 seconds. Waited 10.000008344650269 seconds
  Process: watcher: Passed
       Reason: Returned value: 9.50020146369934 < 30.0
  Process: Default: Skipped
     Reason: Was not started
```

Seen on https://ci.trafficserver.apache.org/job/Github_Builds/job/autest/40871/ (AuTest 0of4), where the other 65 test files in the shard passed.

The test intends a 30-second budget for the wait:

https://github.com/apache/trafficserver/blob/master/tests/gold_tests/h2/h2origin.test.py#L87-L98

`watcher.TimeOut` and `tr.TimeOut` are both set to 30, but the readiness gate that `StartBefore` applies to the watcher is a separate 10-second default that the test does not raise. The watcher was still waiting at 9.5s when the readiness check expired at 10.0s, so the 30-second budget is never reached. Squid log entries appear only when the log buffer is flushed, so on a loaded CI worker the entry for the last transaction can take longer than 10 seconds to show up.

Passing `ready=` with a longer timeout to `StartBefore`, or lowering the log flush interval for this test, would make the wait honor the intended budget.

mandarwagh9:
Fixes #13445.

### Problem

`tests/gold_tests/h2/h2origin.test.py` hand-rolled its "wait for
squid.log" step:

```python
tr = Test.AddTestRun("Wait for the squid.log to be written")
timeout = 30
watcher = tr.Processes.Process("watcher")
watcher.Command = f"sleep {timeout}"
watcher.Ready = When.FileContains(ts.Disk.squid_log.Name, r'14 http/1.1 http/2')
watcher.TimeOut = timeout
...
```

The intent is a 30-second budget, but neither `watcher.TimeOut` nor
`tr.TimeOut` governs the readiness gate. That gate is `StartupTimeout`,
which `Processes.Process()` defaults to 10 seconds:

- `autest/testenities/processes.py`: `startup_timeout=10,  # default to 10 second as most things should be ready by this time`
- `autest/runlogic/runlogic.py`: `if ready_item.object.StartupTimeout < curr_temp: ... "Checking that {typename} is ready within {ready_item.object.StartupTimeout} seconds ..."`

So the run failed at 10.0s while the watcher was still waiting at 9.5s,
exactly as reported in the issue. On a loaded CI worker the squid.log
entry for the last transaction had not been flushed yet.

Worth noting: simply raising `StartupTimeout` to 30 would still race,
because the watcher's own `sleep 30` would exit at about the same
moment, tripping autest's "Process finished before it was ready" path.

### Fix

Two complementary changes:

1. **Use the existing `AddAwaitFileContainsTestRun` helper**
   (`tests/gold_tests/autest-site/when.test.ext`) instead of the
   hand-rolled watcher. It sets `StartupTimeout = 30` against a
   `sleep 60`, so the 30-second budget is actually honored with no
   sleep/gate race. This helper is already used at 20+ call sites;
   `logging/log-milestone-fields.test.py::_waitForLog` is the same
   shape, `StillRunningAfter` lines included. h2origin was simply a
   leftover from before the helper existed.

2. **Set `proxy.config.log.max_secs_per_buffer: 1`** so the log entry
   appears promptly rather than waiting out the default 5-second buffer
   flush. This addresses the underlying latency the issue describes
   rather than only widening the window. 13 other gold tests that assert
   on log contents already do this.

The needle, test-run name, and `StillRunningAfter` assertions are
unchanged, so coverage is identical.

### Testing

Verified locally: `yapf` 0.43.0 with `.style.yapf` reports no
formatting changes, and the file compiles. I develop on Windows and
cannot build or run ATS there, so I have not executed the autest
myself — relying on Jenkins for that. The change is confined to test
configuration and contains no production code.


cmcfarlen:
[approve ci]

cmcfarlen:
Actually just ran into this on the 10.2.x branch.  The h2origin autest is flaky waiting for the log update.  Thanks so much for the fix!

bneradt:
Can you please rebase, @mandarwagh9 ? I've also add you to the CI allow list, so CI should run for your PR on a future push.

cmcfarlen (commented):
The diagnosis is right, and the references check out: `StartupTimeout` defaults to 10 (`autest/testenities/processes.py:41`), which is what failed the run at 10.0s, and it's not governed by either `watcher.TimeOut` or `tr.TimeOut`.

`AddAwaitFileContainsTestRun` is the correct fix rather than just raising the timeout — it pairs `StartupTimeout = 30` with a `sleep 60` (`when.test.ext:93-96`), so the process outlives the gate and can't trip "Process finished before it was ready". Dropping `tr.TimeOut = 30` is consistent with all 34 existing call sites of the helper.

`proxy.config.log.max_secs_per_buffer: 1` is a good addition rather than only widening the window; the default is 5 (`RecordsConfig.cc:1081`) and 15 other gold tests that assert on log contents already set it.

I confirmed the needle survived the conflict resolution as `r'16 http/2 http/2'` — that was the one thing worth double-checking, since keeping the older `14 http/1.1 http/2` would have passed CI while waiting on the wrong entry. All 15 checks are green, including all four AuTest shards.

Two optional notes, neither introduced here:

- The Problem section still quotes the pre-conflict needle `14 http/1.1 http/2`. Worth refreshing since the PR body becomes the commit message.
- `tr.StillRunningAfter = ts` followed by `tr.StillRunningAfter = server` only checks `server`: `TesterSet.Assign` replaces the tester list where `+=` appends (`autest/core/testerset.py:47-52`). So an ATS crash during this run would go unnoticed. It's pre-existing in this file and in ~160 other gold tests, so not yours to fix — but `+=` on the second line is free if you're touching these lines anyway.

cmcfarlen (approved):
Thanks for the fix!

cmcfarlen:
Cherry-picked to the 10.2.x branch as e00a0da685532f5f15bf83dfbf4238f9abc91911 for the 10.2.1 release.

PR #13451, opened 2026-07-29
Add ats-build skill for Claude harness

JosiahWI:
This is to clean up the CLAUDE.md file and hopefully save context. Most users will have their own custom build instructions, so the skill will not have to be loaded.


ezelkow1 (dismissed):

.claude/skills/ats-build/SKILL.md
```diff
@@ -0,0 +1,96 @@
...
+1. Determine build directory name, which is based on the current branch, tag, or commit SHA. This will be used for all builds on this branch/tag/commit.
+2. If that build directory exists, skip this step, otherwise, initialize the build.
+3. Determine the correct target to use.
+4. Launch the build.
+
+## Command Reference
+
+```bash
+ref=$(git symbolic-ref --short -q HEAD || git describe --tags --exact-match 2>/dev/null || git rev-parse --short HEAD)
+BUILD_DIR="build-${ref##*/}"
```
cmcfarlen: I prefer to just use the build directory specified in the preset, not have one for each branch.  Can we make this optional some-how or opt-in?  My workflow makes use of custom presets with their own build directories.

JosiahWI: I think it might be fine just to use the default build directory from the preset, so I can remove this.

ezelkow1 (approved):

PR #13452, opened 2026-07-29
Backport Release 2 security bundle (#13442) to 10.2.x

cmcfarlen:
## Draft — CI verification only

Backports the **#13442 "[master] Release 2"** coordinated security/hardening bundle to
`10.2.x` (these fixes are what the 10.2.0 release is waiting on). Cherry-picked with `-x`
from the range `f910eed635^..e7051e1583` — **149 of 150 commits**.

Security changes are documented in the separate security document (not the
milestone-driven changelog), so there is no per-PR milestone/project bookkeeping here.

### Included: remap-table refcount race fix
- **`709443e870` "Fix race in remap table refcount during reload"** (the
  `std::atomic<UrlRewrite*>` → `AtomicSharedPtr<UrlRewrite>` migration) is now folded into
  this PR, adapted to 10.2.x's `register_config`/`CfgLoad` reload code and scoped enums. The
  shutdown-teardown guard it pairs with is already in this bundle.

### Commits that required manual conflict resolution
Master SHA → on-branch SHA. The first two are code-logic and worth review; the rest are
new-unit-test registration / test-file merges (10.2.x lacked master-only tests
`test_PluginYAML.cc` / `benchmark_TLSCertCompression.cc`, which were excluded).

- `2ce51ba62c` → `94f2069783` — **SSLNetVConnection**: applied the proxy-protocol
  src/dst-addr null-guard onto 10.2.x's existing control flow (did not take master's
  refactored parse-failure branch).
- `e8d92a51f2` → `0a9b3a3dd5` — **test_Hdrs.cc**: appended only the new `http_parse_status`
  TEST_CASE; excluded the master-only `unmarshal_v24_1` block.
- `8102a6b6e3` → `e6f8b84551` — proxy unit_tests CMakeLists: added `test_ControlBase.cc`.
- `6a4ecdad19` → `21b786a1ec` — net unit_tests CMakeLists: added `test_OCSPStapling.cc`.
- `6a8c575f3f` → `de964687af` — proxy unit_tests CMakeLists: added `test_FetchSM.cc`.
- `6916a91bde` → `63d5c95765` — test_proxy: the macOS link-duplicates fix (drop
  libinknet_stub.cc, add `appVersionInfo`).
- `b5510fb26d` → `93657ab769` and `43a5d7dec7` → `c17b724772` —
  http2_crlf_header_validation.test.py (records_config format interplay with the ssl fixup).

### AuTest fixups (10.2.x harness adaptation)
[...]

PR #13454, opened 2026-07-30
Fix race in remap table refcount during reload (10.2.x)

cmcfarlen:
## Draft — CI verification

Backports the remap-table refcount race fix to `10.2.x`, adapted to this branch's
config-reload code. The fix is on master as `709443e870`; it was deferred from the
Release 2 backport (#13452) because master's version is entangled with remap.yaml and
the `ConfigRegistry::attach()` API, neither of which is on 10.2.x.

### The race
The global remap table's `load()` and `acquire()` ran as two unsynchronized steps, while
config reload swapped the table and released the old one without a mutex. A reader
preempted between load and acquire could revive a table whose refcount reload had just
driven to zero — after the deleter was already scheduled — a use-after-free.

### The fix
- Retire the bespoke `acquire()/release()` refcount on `UrlRewrite`; use
  `std::atomic<std::shared_ptr<UrlRewrite>>` (via a new `AtomicSharedPtr` helper) for
  publish-and-replace.
- Each transaction snapshots the current table into `HttpSM::m_remap` at session start;
  reload `exchange()`s in a new `shared_ptr` and drops its ref, so the old table destructs
  only after the last in-flight `HttpSM` releases its snapshot.
- `shutdown_url_rewrite()` drains and inhibits further drops so plugin `doneInstance()`
  runs while `this_ethread()` is still valid.
- Second commit guards remap consumers against a null table during the shutdown window
  (quiet defensive remap miss), preventing a crash in `RemapProcessor::setup_for_remap`
  when a transaction is created after the table is torn down.

### 10.2.x adaptation
Kept this branch's `register_config(…)` + `CfgLoad*` reload instrumentation and scoped
enums; grafted the `AtomicSharedPtr` lifecycle onto them.

### Testing
Local build clean; `ctest` 157/158 (the one failure is the pre-existing macOS
`test_jsonrpcserver` unix-socket flake, unrelated). Note: the unit tests do not exercise
the reload race itself — a config-reload-under-load check is the meaningful CI signal here.

### DIFF
--- a/include/proxy/ReverseProxy.h
+++ b/include/proxy/ReverseProxy.h
@@ -43,12 +43,14 @@
 #include "proxy/http/remap/UrlMapping.h"
 #include "mgmt/config/ConfigContext.h"
 
+#include "tsutil/AtomicSharedPtr.h"
+
 #define EMPTY_PORT_MAPPING (int32_t) ~0
 
 class url_mapping;
 struct host_hdr_info;
 
-extern std::atomic<UrlRewrite *> rewrite_table;
+extern AtomicSharedPtr<UrlRewrite> rewrite_table;
 
 // API Functions
 int init_reverse_proxy();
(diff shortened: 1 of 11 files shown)


cmcfarlen:
Folded into #13452 (Release 2 bundle) — the remap-table refcount race fix belongs with the security bundle it was split from. Superseded by this branch's inclusion; closing.

PR #13455, opened 2026-07-30
Reuse deleted MIME field slots

### LINKED ISSUE #8466: MIME Field Free List

bneradt:
Long-lived MIME headers allocate field blocks as fields are added and
removed. HPACK exercises this path heavily, while the current attempt to
bound the block chain scans and destroys empty blocks during deletion.

This maintains a deleted-slot free list through m_next_dup and consumes
it before allocating another field block. Appending within a partially
filled tail preserves established field order. The free list is rebuilt
for copied or unmarshaled headers, and same-name duplicate order is
preserved.

This adds high-water churn, insertion-order, duplicate, copy, and
marshal coverage.

Fixes: #8466

### DIFF
--- a/src/proxy/hdrs/MIME.cc
+++ b/src/proxy/hdrs/MIME.cc
@@ -542,6 +542,7 @@ mime_hdr_sanity_check(MIMEHdrImpl *mh)
   MIMEField          *field, *next_dup;
   uint32_t            slot_index, index;
   uint64_t            masksum;
+  size_t              deleted_count = 0;
 
   ink_assert(mh != nullptr);
 
@@ -617,13 +618,15 @@ mime_hdr_sanity_check(MIMEHdrImpl *mh)
           ink_release_assert(found);
         }
         // re-find the field --- should always find the head dup
-        MIMEField *mf = mime_hdr_field_find(mh, field->m_ptr_name, field->m_len_name);
+        MIMEField *mf = mime_hdr_field_find(mh, {field->m_ptr_name, field->m_len_name});
         ink_release_assert(mf != nullptr);
         if (mf == field) {
           ink_release_assert((field->m_flags & MIME_FIELD_SLOT_FLAGS_DUP_HEAD) != 0);
         } else {
           ink_release_assert((field->m_flags & MIME_FIELD_SLOT_FLAGS_DUP_HEAD) == 0);
         }
+      } else if (field->m_readiness == MIME_FIELD_SLOT_READINESS_DELETED) {
+        ++deleted_count;
       }
 
       ++slot_index;
@@ -633,6 +636,19 @@ mime_hdr_sanity_check(MIMEHdrImpl *mh)
 
   ink_release_assert(last_fblock == mh->m_fblock_list_tail);
   ink_release_assert(masksum == mh->m_presence_bits);
+
+  if (mh->m_free_slot != MIME_FIELD_FREE_SLOT_UNINITIALIZED) {
+    size_t free_count = 0;
+
+    field = mh->m_free_slot == MIME_FIELD_FREE_SLOT_NONE ? nullptr : mime_hdr_field_get_slotnum(mh, mh->m_free_slot);
+    while (field != nullptr) {
+      ink_release_assert(field->m_readiness == MIME_FIELD_SLOT_READINESS_DELETED);
+      ink_release_assert(mime_hdr_field_slotnum(mh, field) >= 0);
+      ink_release_assert(++free_count <= deleted_count);
+      field = field->m_next_dup;
+    }
+    ink_release_assert(free_count == deleted_count);
+  }
 }
 #endif
 
@@ -949,6 +965,8 @@ mime_hdr_cooked_stuff_init(MIMEHdrImpl *mh, MIMEField *changing_field_or_null)
 void
 mime_hdr_init(MIMEHdrImpl *mh)
 {
+  mh->m_free_slot = MIME_FIELD_FREE_SLOT_NONE;
+
   mime_hdr_init_accelerators_and_presence_bits(mh);
 
   mime_hdr_cooked_stuff_init(mh, nullptr);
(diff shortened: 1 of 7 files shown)


src/proxy/hdrs/unit_tests/test_Hdrs.cc
```diff
@@ -657,6 +657,175 @@ test_arena_aux(Arena &arena, int len)
...
+
+    MIMEField *deleted = add_field(hdr, "X-Deleted");
+    MIMEField *kept    = add_field(hdr, "X-Kept");
+
+    hdr.field_delete(deleted, false);
+    MIMEField *appended = add_field(hdr, "X-Appended");
+
+    CHECK(appended != deleted);
+    CHECK(mime_hdr_field_slotnum(hdr.m_mime, kept) < mime_hdr_field_slotnum(hdr.m_mime, appended));
+  }
```
moonchen, line 759: This passes on master and fails here -- `field_find` returns the newly added field, slot 15 vs slot 0.

```suggestion
  }

  SECTION("A reused earlier slot preserves same-name field order")
  {
    HTTPHdr hdr;
    hdr.create(HTTPType::RESPONSE);
    ts::PostScript cleanup([&]() -> void { hdr.destroy(); });

    MIMEField *victim = add_field(hdr, "X-Victim");
    for (unsigned index = 1; index < MIME_FIELD_BLOCK_SLOTS - 1; ++index) {
      add_field(hdr, "X-Filler-" + std::to_string(index));
    }
    MIMEField *first = add_field(hdr, "Set-Cookie");
    REQUIRE(hdr.m_mime->m_first_fblock.m_freetop == MIME_FIELD_BLOCK_SLOTS);

    hdr.field_delete(victim, false);
    MIMEField *second = add_field(hdr, "Set-Cookie");

    CHECK(hdr.field_find("Set-Cookie") == first);
    CHECK(mime_hdr_field_slotnum(hdr.m_mime, first) < mime_hdr_field_slotnum(hdr.m_mime, second));
  }
```

src/proxy/hdrs/MIME.cc
```diff
@@ -1338,21 +1359,63 @@ mime_field_init(MIMEField *field)
...
+  if (mh->m_free_slot == MIME_FIELD_FREE_SLOT_UNINITIALIZED) {
+    mime_hdr_rebuild_field_free_list(mh);
   }
 
+  if (mh->m_free_slot != MIME_FIELD_FREE_SLOT_NONE) {
+    field = mime_hdr_field_get_slotnum(mh, mh->m_free_slot);
+    ink_release_assert(field != nullptr);
+    ink_release_assert(field->m_readiness == MIME_FIELD_SLOT_READINESS_DELETED);
+
+    mh->m_free_slot = field->m_next_dup ? mime_hdr_field_slotnum(mh, field->m_next_dup) : MIME_FIELD_FREE_SLOT_NONE;
```
moonchen: Each field delete, and each insert that reuses a slot, now traverses the field block list twice -- once in `mime_hdr_field_get_slotnum` and once in `mime_hdr_field_slotnum`. The chain is bounded by this change so it's not a blocker, but may be worth a quick benchmark against alternative approaches. Copilot flagged this too; its confidence filter suppressed the comment into the review summary.

src/proxy/hdrs/MIME.cc
```diff
@@ -1002,6 +1020,8 @@ mime_hdr_destroy(HdrHeap *heap, MIMEHdrImpl *mh)
   // heap->deallocate_obj(mh);
 }
 
+static void mime_hdr_rebuild_field_free_list(MIMEHdrImpl *mh);
```
moonchen: Optional suggestion: `mime_hdr_rebuild_field_free_list` could be moved up here.

moonchen (changes requested):

src/proxy/hdrs/unit_tests/test_Hdrs.cc
bneradt: Thanks, this exposed that arbitrary slot reuse could reorder same-name fields. I added this regression test and made allocation name-aware so duplicate fields skip earlier free slots.

src/proxy/hdrs/MIME.cc
bneradt: Agreed. Deleted fields now store the next free slot number directly, eliminating the second field-block traversal from deletion and reuse.

src/proxy/hdrs/MIME.cc
bneradt: Done. I moved the full mime_hdr_rebuild_field_free_list() definition to the earlier location.

src/proxy/hdrs/MIME.cc
```diff
@@ -1361,10 +1407,59 @@ mime_field_create(HdrHeap *heap, MIMEHdrImpl *mh)
...
+
+  if (MIMEField *last_dup = name.empty() ? nullptr : mime_hdr_field_find(mh, name); last_dup != nullptr) {
+    while (last_dup->m_next_dup != nullptr) {
+      last_dup = last_dup->m_next_dup;
+    }
+    last_dup_slot = mime_hdr_field_slotnum(mh, last_dup);
+    ink_release_assert(last_dup_slot >= 0);
+  }
+
+  if (mh->m_free_slot == MIME_FIELD_FREE_SLOT_UNINITIALIZED) {
```
moonchen: Could we avoid `mime_hdr_field_find()` and walking the duplicate chain when `m_free_slot == MIME_FIELD_FREE_SLOT_NONE`? Since the tail is full here, `mime_field_create()` must allocate a new block, and its slot will necessarily follow every existing duplicate. This would mean rebuilding first when `m_free_slot == MIME_FIELD_FREE_SLOT_UNINITIALIZED`, then returning early if the result is `NONE`.

bneradt: Implemented in `5d91d73dbc`. The free list is now rebuilt before any name lookup, and an empty result returns directly to normal block allocation. That avoids both `mime_hdr_field_find()` and the duplicate-chain walk when no reusable slot exists.

Validated with a full configure/format/build/install in `asfats5`, the focused MIME slot-reuse test (29 assertions), and the HPACK regression executable.

moonchen (approved):
Thanks for addressing the review. Verified the no-free-slot early return preserves duplicate order, and `test_proxy_hdrs` passes with `ENABLE_MIME_SANITY_CHECK` on (53 cases / 437k assertions), which exercises the free-list invariants. LGTM.

PR #13458, opened 2026-07-30
Replace ENGINE support with PROVIDER support on OpenSSL3+

### LINKED ISSUE #13347: Remove remaining deprecated OpenSSL 3.0 APIs in SSLUtils.cc (DH and ENGINE)

JosiahWI:
Resolves #13347

### DIFF
--- a/src/iocore/net/SSLKeyUtils.cc
+++ b/src/iocore/net/SSLKeyUtils.cc
@@ -194,11 +194,6 @@ use_pkey_from_file(SSL_CTX *ctx, const char *keyPath)
 {
   ink_assert(keyPath && keyPath[0] != '\0');
   int const result{SSL_CTX_use_PrivateKey_file(ctx, keyPath, SSL_FILETYPE_PEM)};
-  if (1 != result) {
-    char err_buf[256]{};
-    ERR_error_string_n(ERR_get_error(), err_buf, sizeof(err_buf));
-    Error("failed to load private key %s: %s", keyPath, err_buf);
-  }
   return 1 == result;
 }
 
--- a/src/iocore/net/SSLKeyUtils.cc
+++ b/src/iocore/net/SSLKeyUtils.cc
@@ -39,6 +39,8 @@
 #ifdef OPENSSL_IS_OPENSSL3
 #include <openssl/decoder.h>
 #include <openssl/params.h>
+#include <openssl/store.h>
+#include <openssl/ui.h>
 #else
 #include <openssl/bn.h>
 #endif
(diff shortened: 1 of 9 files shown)


bneradt:
[approve ci autest 1]

src/iocore/net/unit_tests/test_SSLDHParams.cc
```diff
@@ -184,3 +299,65 @@ TEST_CASE("ssl_context_enable_dhe: truncated DH PEM (missing END marker) is reje
...
+//
+// This is the case an ENGINE-based ATS handled via ENGINE_load_private_key.
+TEST_CASE("SSLPrivateKeyHandler: a key named by an OpenSSL provider URI is loaded from the provider")
+{
+  CertAndKey                           ck = make_cert_and_key();
+  MockHardwareProvider::ScopedProvider provider{ck.key_pem};
+  REQUIRE(provider.is_loaded());
+
+  TempFile cert{ck.cert_pem};
+  CHECK(load_key_via_load_certs(cert.get_path(), MockHardwareProvider::URI));
```
bneradt, line 362: **[P1] Preserve provider URIs before filesystem path resolution**

This test injects the raw URI directly into `CertLoadData`, bypassing the production configuration path. `load_certs_and_cross_reference_names()` passes every `ssl_key_name` through `Layout::relative_to(params->serverKeyPathOnly, keyname)`, so the default key directory turns `pkcs11:...` into `/.../pkcs11:...`. `OSSL_STORE_open()` consequently receives a filesystem path instead of the provider scheme, making the new provider path unreachable from normal `ssl_multicert.yaml` configuration. Please preserve recognized store URIs before filesystem resolution and exercise the full normalization path in this test.

src/iocore/net/SSLUtils.cc
```diff
@@ -856,43 +856,24 @@ SSLMultiCertConfigLoader::default_server_ssl_ctx()
...
-    pkey = ENGINE_load_private_key(e, keyPath, nullptr, nullptr);
-    if (pkey) {
-      if (!SSL_CTX_use_PrivateKey(ctx, pkey)) {
-        Dbg(dbg_ctl_ssl_load, "failed to load server private key from engine");
-        EVP_PKEY_free(pkey);
-        return false;
-      }
-    }
+  bool result{false};
+  if (keyPath && keyPath[0] != '\0') {
```
bneradt, line 860: **[P1] Keep SSL-secret plugin data authoritative**

Trying `keyPath` before the `secret_data` returned by `SSLSecret::getOrLoadSecret()` reverses the documented `TS_LIFECYCLE_SSL_SECRET_HOOK` contract, where plugin data replaces disk loading. If the configured path contains a stale but parseable key, this helper succeeds and skips the plugin key; the later certificate check can then fail without ever trying the matching secret. This also defeats consistent updates made through `TSSslSecretSet()`. Please retain secret-data precedence for ordinary file names and use the path-first store lookup only for actual provider URIs.

bneradt (commented):
**[P2] Remove the remaining OpenSSL 3 ENGINE call**

This PR says it resolves #13347, but `SSLPostConfigInitialize()` still calls [`ENGINE_load_dynamic()`](https://github.com/apache/trafficserver/blob/5b2408011208c053d1c785f3bd930144776c57bf/src/iocore/net/SSLUtils.cc#L769-L775), one of the deprecated calls explicitly listed in that issue. With the current OpenSSL 3 compatibility setting, CMake still detects this API and ATS invokes it whenever `proxy.config.ssl.engine.conf_file` is configured. Please exclude this call under `OPENSSL_IS_OPENSSL3` or replace the initialization path before resolving the issue.

src/iocore/net/SSLUtils.cc
```diff
@@ -856,43 +856,24 @@ SSLMultiCertConfigLoader::default_server_ssl_ctx()
...
+    result = use_pkey_from_secret_data(ctx, secret_data, secret_data_len);
+  }
+
+  if (!result) {
+    Dbg(dbg_ctl_ssl_load, "failed to load server private key (%.*s) from %s", secret_data_len < 50 ? secret_data_len : 50,
+        secret_data, (!keyPath || keyPath[0] == '\0') ? "[empty key path]" : keyPath);
+    return false;
+  }
+
+  if (!SSL_CTX_check_private_key(ctx)) {
```
bneradt, line 874: **[P1] Do not require this comparison for opaque hardware keys**

The previous `e == nullptr` guard was deliberate: hardware-backed ENGINE keys, including HSM/TPM implementations, may support signing without exposing enough key material for `SSL_CTX_check_private_key()` to compare them. Making the check unconditional rejects those keys in the retained pre-OpenSSL-3 ENGINE path and can likewise reject opaque provider keys. The mock provider does not cover this because it returns an ordinary exportable PEM key. Please preserve the hardware-key distinction or use a validation path that provider/ENGINE-backed opaque keys can support.

src/iocore/net/SSLKeyUtils.cc
```diff
@@ -188,3 +192,113 @@ set_ctx_dh(SSL_CTX *ctx, dh_key_t *pkey)
...
+
+bool
+use_pkey_from_file(SSL_CTX *ctx, const char *keyPath)
+{
+  ink_assert(keyPath && keyPath[0] != '\0');
+
+#if HAVE_ENGINE_GET_DEFAULT_RSA && HAVE_ENGINE_LOAD_PRIVATE_KEY
+  // Before providers, a hardware key store was reached through an ENGINE.
+  // Absent a configured engine there is nothing to ask.
+  if (ENGINE *e = ENGINE_get_default_RSA(); e != nullptr) {
```
bneradt, line 265: **[P2] Release the functional ENGINE reference**

`ENGINE_get_default_RSA()` returns an incremented functional reference that must be released with `ENGINE_finish()` before it is discarded. Neither the successful early return nor the failure/fallback path releases `e`, so repeated certificate loads and configuration reloads leak ENGINE references. Please manage this with an RAII deleter or explicitly finish the ENGINE on every path.

JosiahWI:
Opting to leave out provider support until there is a need.

PR #13459, opened 2026-07-30
CI: enable WAMR in Fedora builds

bneradt:
Fedora CI disabled WAMR because its image carried an older runtime. The
presets also point to the custom OpenSSL and curl installations removed
from the Fedora 44 image, so dependency discovery relies on stale hints.

This restores AUTO detection for WAMR in shared CI and AuTest presets
and removes the obsolete Fedora OpenSSL and curl roots. The Fedora
presets retain their explicit OpenTelemetry and WAMR roots, while quiche
retains its dedicated BoringSSL and quiche locations.


bneradt:
Superseded by #13423, which already landed the Fedora OTEL/WAMR preset requirements and WASM linkage fixes.

PR #13461, opened 2026-07-30
Fix JSONRPC server shutdown race

bneradt:
JSONRPC server shutdown can race with worker thread startup. When the
worker starts after stop_thread(), it restores the running flag and
polls a closed socket indefinitely. This causes test_jsonrpcserver and
process shutdown to hang.

This marks the socket server as running before the worker is created,
so a concurrent stop cannot be overwritten. It also passes the owning
server to the worker instead of relying on the mutable global server
pointer.

### DIFF
--- a/include/mgmt/rpc/server/IPCSocketServer.h
+++ b/include/mgmt/rpc/server/IPCSocketServer.h
@@ -144,7 +144,7 @@ private:
   void close();
   void late_check_peer_credentials(int peedFd, TSRPCHandlerOptions const &options, swoc::Errata &errata) const;
 
-  std::atomic_bool _running;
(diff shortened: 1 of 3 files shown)


cmcfarlen (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as 1427ebf71ad9abef22c8c4fd40b391caa935591a for the 10.2.0 release.

cmcfarlen:
Cherry-picked to the 10.1.x branch as 629c96bd4f0a1fb6d6fb878d0b8a5c8c136fab22 for the 10.1.5 release.

PR #13462, opened 2026-07-30
Stabilize proxy protocol access log test

bneradt:
Access log records for independent PROXY Protocol replay sessions can be
finalized on different event threads. The test assumed replay order and
failed intermittently when two otherwise correct records were reversed.

This waits for all expected records and sorts them before comparing
against the full gold output. It preserves content and count coverage
without requiring transaction completion order.


maskit (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as 0909f575cfed0441be15fa260ee203d98ff02611 for the 10.2.0 release.

PR #13463, opened 2026-07-30
10.2.x: HTTP/3 via OpenSSL QUIC

bneradt:
OpenSSL 3.5 can terminate QUIC connections directly, but ATS only had a
quiche-backed HTTP/3 listener. Operators who want to use the system
OpenSSL QUIC stack needed a separate downstream backend without changing
the existing quiche path or origin HTTP/3 scope.

This adds an optional ENABLE_OPENSSL_QUIC backend that uses OpenSSL's
native QUIC listener and stream APIs for downstream HTTP/3. This keeps
the backend mutually exclusive with quiche, exposes TS_HAS_OPENSSL_QUIC,
and shares ATS's existing HTTP/3 stream handling above the transport.

This also installs native-QUIC TLS callbacks for ALPN and SNI
certificate selection before ATS has a QUIC NetVC to bind. OpenSSL
native QUIC does not make a selected SSL_CTX certificate active via
SSL_set_SSL_CTX alone, so this applies the selected cert, key, and chain
to the connection SSL.

This also broadens client-side HTTP/3 tests to run with either backend,
keeps H3 streams open across informational responses, and hardens
transaction cleanup when OpenSSL closes stream state before ATS finishes
teardown. This caches stream identifiers, declines listener-time QUIC
tickets until a NetVC is bound, and adds focused H3 lifecycle and
session-ticket coverage.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -214,6 +214,7 @@ set(ENABLE_TPROXY
         'X' where X is a number to use as the IP_TRANSPARENT sockopt,
         anything else to enable."
 )
+option(ENABLE_OPENSSL_QUIC "Use OpenSSL native QUIC (default OFF)")
 option(ENABLE_QUICHE "Use quiche (default OFF)")
 
 option(ENABLE_EXAMPLE "Build example directory (default OFF)")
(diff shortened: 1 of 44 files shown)


cmcfarlen (approved):

cmcfarlen:
Landed directly on the 10.2.x branch as 939915eeee8a5cf0e5ffb9e51def40603cddb317 for the 10.2.0 release (this PR was merged against 10.2.x, so no cherry-pick was needed).

PR #13464, opened 2026-07-31
TLS: Fix EVP_PKEY leak in SSLPrivateKeyHandler

moonchen:
`SSL_CTX_use_PrivateKey()` takes its own reference on the key, so the reference from `PEM_read_bio_PrivateKey()` belongs to the caller. `SSLPrivateKeyHandler()` released it only when attaching the key failed, so a successful load leaked one `EVP_PKEY` per certificate. This repeats at startup, on every config reload, and on every secret or certificate update.

Hold the key in a `scoped_PKEY` so every exit releases it.

### DIFF
--- a/src/iocore/net/P_SSLUtils.h
+++ b/src/iocore/net/P_SSLUtils.h
@@ -114,6 +114,14 @@ namespace detail
     }
   };
 
+  struct PKEYDeleter {
+    void
+    operator()(EVP_PKEY *p)
+    {
+      EVP_PKEY_free(p);
+    }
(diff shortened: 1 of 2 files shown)


JosiahWI (approved):
Thanks for this fix. This is also fixed in #13458; I'll rebase that PR on this.

cmcfarlen:
Cherry-picked to the 10.2.x branch as d14c803426dab1f15423204f4d91993bee1ddd8b for the 10.2.0 release.

PR #13466, opened 2026-07-31
Release logging thread continuations

bneradt:
Dedicated logging continuations remain allocated after their worker loops return during shutdown. Once shutdown reliably completes, LeakSanitizer can report the flush continuation as an 80-byte leak.

Release both continuations after their loops finish. This matches the ownership model used by other dedicated event threads.


cmcfarlen (approved):
Ah, ez fix

cmcfarlen:
Cherry-picked to the 10.2.x branch as ca5f3a3e24cb48ecfcf2a8edeb142ab807df6407 for the 10.2.0 release.

PR #13467, opened 2026-07-31
10.1.x: ocsp: add single-cert stapling fast path and certinfo RAII (#13229)

moonchen:
Backport of #13229 (`30d8fcad44`) to 10.1.x, because this fixes a leak this fixes a leak in the OCSP stapling code: under BoringSSL the certinfo map's reference on the leaf certificate is never released, so a leaf leaks per certificate on every config reload.

---

Skip the SSL_get_certificate() lookup and X509_cmp() DER re-parse in the stapling callback when an SSL_CTX has a single certificate. The shortcut is gated to non-dual-cert builds; under HAVE_NATIVE_DUAL_CERT_SUPPORT a CTX can hold multiple certs where only one has OCSP info, so map size alone cannot identify the negotiated cert.

Give certinfo a constructor/destructor so its resources are managed by RAII, and allocate it with make_unique. This consolidates the cleanup that was duplicated across certinfo_map_free and the init error path, and fixes two pre-existing leaks (cid and the BoringSSL cert ref) plus an error path that could delete a certinfo_map still owned by the SSL_CTX.

### DIFF
--- a/src/iocore/net/OCSPStapling.cc
+++ b/src/iocore/net/OCSPStapling.cc
@@ -21,6 +21,8 @@
 
 #include "P_OCSPStapling.h"
 
+#include <memory>
+
 #include <openssl/ssl.h>
 #include <openssl/x509v3.h>
 #include <openssl/asn1.h>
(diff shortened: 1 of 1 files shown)


moonchen:
[approve ci fedora autest 2]

moonchen:
[approve ci autest 2]

moonchen:
[approve ci autest 2]

moonchen:
[approve ci autest 2]

moonchen:
[approve ci autest 2]

PR #13468, opened 2026-07-31
Enable OpenSSL QUIC when available

bneradt:
OpenSSL 3.5 provides the QUIC implementation ATS needs to accept
HTTP/3 connections from downstream clients. This affects client-to-ATS
traffic, not ATS-to-origin traffic. Requiring an extra opt-in leaves
capable builds without client-facing HTTP/3, but merely compiling QUIC
should not make deployments without QUIC listeners load every TLS
certificate twice.

This makes native OpenSSL QUIC an automatic CMake option. It activates
only with upstream OpenSSL's native server API, remains disabled when
quiche is selected, and preserves explicit ON and OFF overrides.

This also starts the QUIC processor and maintains its certificate
table only when a QUIC listener is configured, avoiding the duplicate
load when HTTP/3 is unused.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -215,7 +215,6 @@ set(ENABLE_TPROXY
         'X' where X is a number to use as the IP_TRANSPARENT sockopt,
         anything else to enable."
 )
-option(ENABLE_OPENSSL_QUIC "Use OpenSSL native QUIC (default OFF)")
 option(ENABLE_QUICHE "Use quiche (default OFF)")
 
 option(ENABLE_EXAMPLE "Build example directory (default OFF)")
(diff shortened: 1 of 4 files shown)


maskit:
If the OpenSSL version works in a similar way as Quiche version, there is a downside to enabling QUIC automatically. All TLS certs would be loaded twice in total, once by SSLMultiCertConfigLoader and once by QUICMultiCertConfigLoader, even if you don't enable QUIC on records.config. The behavior would be consistent though.

Edit: Actually ENABLE_QUIC may be necessary for Quiche? Regardless, there's that downside.

bneradt:
> If the OpenSSL version works in a similar way as Quiche version, there is a downside to enabling QUIC automatically. All TLS certs would be loaded twice in total, once by SSLMultiCertConfigLoader and once by QUICMultiCertConfigLoader, even if you don't enable QUIC on records.config. The behavior would be consistent though.

Great point. This is pretty easy to address actually.

Before: quic_NetProcessor.start() was loading a second certificate table for every QUIC-capable build, and coordinated TLS reloads repeated that work even without a :quic listener. 

After: this patch changes both paths to check HttpProxyPort::hasQUIC() before doing them.

maskit (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as 944fb6d97d7b7759cde178dc0e0ab57f21ea8f3f for the 10.2.0 release.

PR #13470, opened 2026-08-01
Fix some copy instead of move Coverity CIDs

JosiahWI:
CIDs fixed in this patch:

1664317
1664316
1664286
1664284
1663965
1660036
1658876
1658839
1658837
1658799
1654583
1654213
1654211
1654210
1654207
1654206
1645817
1645816
1645814
1645813
1645811
1645809
1645808
1645806
1645804
1645802
1645801
1645799
1645798
1645796
1645795
1644303
1644300
1644293
1644255
1644242
1644227
1644207

### DIFF
--- a/include/iocore/net/ConnectionTracker.h
+++ b/include/iocore/net/ConnectionTracker.h
@@ -32,6 +32,7 @@
 #include <mutex>
 #include <sstream>
 #include <tuple>
+#include <utility>
 #include "records/RecCore.h"
 #include "tscore/ink_platform.h"
 #include "tscore/ink_config.h"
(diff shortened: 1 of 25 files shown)


JosiahWI:
Copilot's comment about moving the ternary operand blocking NRVO/copy elision is possibly the most interesting dubious comment I've seen from Copilot to date. It has low confidence, which is good, because it should. This is a very subtle trap in C++: the original ternary expression has an l-value operand and an r-value operand, so it is guaranteed to evaluate to an r-value. Therefore, the ternary expression is actually forced to cause a temporary copy of `metric_name` if that branch is taken. Coverity is right - this is a special edge case where the `std::move` is actually appropriate.

cmcfarlen (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as ebad97a1e37d094211e171a90c18cf7bb901f5be for the 10.2.0 release. This is a partial pick; see the commit message for the omitted hunks.

PR #13472, opened 2026-08-01
Guard pre-init log wakeups

bneradt:
Plugins can fill a text log buffer during TSPluginInit before logging
preprocessing threads and their notification objects exist. This causes
a null dereference that crashes traffic_server during startup.

This problem is addressed in this patch by queuing those buffers without
signaling until the logging workers exist. Their initial queue scan then
flushes the pending data. A regression test forces a rollover during
plugin initialization and verifies ATS stays up and preserves the log
entry.

With the production fix removed, the regression test caused
traffic_server to crash with SIGSEGV at address zero while signaling the
preprocessing notification. With the fix applied, the identical test
passes.

### DIFF
--- a/src/proxy/logging/LogObject.cc
+++ b/src/proxy/logging/LogObject.cc
@@ -395,7 +395,9 @@ LogObject::_checkout_write(size_t *write_offset, size_t bytes_needed)
         int idx = m_buffer_manager_idx++ % m_flush_threads;
         Dbg(dbg_ctl_log_logbuffer, "adding buffer %d to flush list after checkout", buffer->get_id());
         m_buffer_manager[idx].add_to_flush_queue(buffer);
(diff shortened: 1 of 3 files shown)


bneradt:
I verified the regression test in an A/B run. With the test and test-plugin changes present but the production LogObject guards removed, traffic_server crashed during TSPluginInit with SIGSEGV at address 0x0. The backtrace passed through EventNotify::signal() and LogObject::_checkout_write(), and AuTest failed. Restoring the production fix made the identical test pass.

maskit (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as 9960332e3369867a1affe6fc99b4bdaad1d281a6 for the 10.2.0 release.

PR #13475, opened 2026-08-03
slice: purge every block of an object, not just those before a gap

masaori335:
## Motivation

  A sliced PURGE only removed the blocks before the first gap. The walk stopped at
  the first block that was not in cache, so any object whose cached blocks were not
  a contiguous run from block 0 was partially purged — and the client still got a
  `200`. A gap in the middle left every block behind it cached; an uncached first
  block purged nothing and relayed that block's `404`; and `bytes=-N` deleted the
  head while leaving the tail it had named.

  The stop was load-bearing rather than a stray early return: the walk's only other
  terminator needs the object length, which slice learns from a `206`'s
  `Content-Range`, and a PURGE response has none.

  ## Changes

  **ATS core** — report the purged object's extent as `X-Purged-Content-Range` on a
  PURGE cache hit, so a caller holding one piece of a larger resource can learn its
  extent. Deliberately not `Content-Range`: on a `200` that is meaningless under RFC
  9110, and `cache_range_requests` reads the pair as a stored `206` being served as
  `200` and rewrites the status.

  **Slice plugin** — give PURGE its own state machine, so it no longer routes through
  `handleFirstServerHeader` and cannot leak a block's `404` to the client. It walks
  every block, steps over a `404`, takes the extent as the largest any block reports
  (blocks disagree when the origin object was replaced in place), and synthesizes the
  response only once the walk finishes: `200` if any block was removed, `404` if none.
  Until some block reports an extent the walk is bounded by `--purge-probe-blocks`
  (default 8), overridable per request via `--purge-probe-header`. A suffix range is
  widened to the whole object; an unparseable range is refused with `400`.

  ## Tests

  `slice_purge_gaps` covers the traversal over gaps, both open-ended range forms, the
  miss bound and its override, and the refusal. `slice_stale_generation` reproduces
  the client-visible failures of an origin object replaced in place, which is how
  this was found.

### DIFF
--- a/doc/admin-guide/storage/index.en.rst
+++ b/doc/admin-guide/storage/index.en.rst
@@ -304,7 +304,7 @@ from any other IP, we connect to the daemon via localhost: ::
       > Host: example.com
       > Accept: */*
       >
-      < HTTP/1.1 200 Ok
+      < HTTP/1.1 200 OK
       < Date: Thu, 08 Jan 2010 20:32:07 GMT
       < Connection: keep-alive
 
@@ -316,7 +316,7 @@ If the removed object was stored as a partial response, that is if it carried a
 ``Content-Range``, then the ``200 OK`` also reports that range back in a
 ``X-Purged-Content-Range`` header::
 
-      < HTTP/1.1 200 Ok
+      < HTTP/1.1 200 OK
       < X-Purged-Content-Range: bytes 0-1048575/9437184
 
 This lets a caller that holds one piece of a larger resource learn the whole
--- a/doc/admin-guide/storage/index.en.rst
+++ b/doc/admin-guide/storage/index.en.rst
@@ -312,6 +312,21 @@ The next time Traffic Server receives a request for the removed object,
 it will contact the origin server to retrieve a new copy, which will replace
 the previously cached version in Traffic Server.
 
+If the removed object was stored as a partial response, that is if it carried a
+``Content-Range``, then the ``200 OK`` also reports that range back in a
+``X-Purged-Content-Range`` header::
+
+      < HTTP/1.1 200 Ok
+      < X-Purged-Content-Range: bytes 0-1048575/9437184
+
+This lets a caller that holds one piece of a larger resource learn the whole
+resource's extent without a second lookup. It is what allows the
+:ref:`admin-plugins-slice` plugin to purge an object block by block and know which
+block is the last one. The range is reported under its own header name rather than
+as ``Content-Range``, because ``Content-Range`` on a ``200`` response has no
+meaning under :rfc:`9110` and is read by other components as a sign that a stored
+partial response is being served.
+
 This procedure only removes the index to the object from a specific Traffic Server
 cache. While the object remains on disk, Traffic Server will no longer able to find
 the object. The next request for that object will result in a fresh copy of the
--- a/plugins/slice/client.cc
+++ b/plugins/slice/client.cc
@@ -32,7 +32,7 @@ namespace
 {
 // Miss bound for this purge: the request header when usable, else the config value.
 int
-purge_miss_bound(HttpHeader const &header, Config const *const conf)
+purge_miss_bound(HttpHeader const &header, Config *const conf)
 {
   char probestr[64];
   int  probelen = sizeof(probestr);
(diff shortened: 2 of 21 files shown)


doc/admin-guide/storage/index.en.rst
```diff
@@ -312,6 +312,21 @@ The next time Traffic Server receives a request for the removed object,
 it will contact the origin server to retrieve a new copy, which will replace
 the previously cached version in Traffic Server.
 
+If the removed object was stored as a partial response, that is if it carried a
+``Content-Range``, then the ``200 OK`` also reports that range back in a
+``X-Purged-Content-Range`` header::
+
+      < HTTP/1.1 200 Ok
```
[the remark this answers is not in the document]
masaori335: Nice catch. It comes from existing example in above. These are fixed.

plugins/slice/client.cc
```diff
@@ -19,9 +19,46 @@
...
+  swoc::TextView value{probestr, static_cast<size_t>(probelen)};
+  // isspace is only defined for values representable as unsigned char
+  value.trim_if([](char c) { return 0 != isspace(static_cast<unsigned char>(c)); });
+
+  swoc::TextView parsed;
+  intmax_t const blocks = swoc::svtoi(value, &parsed, 10);
+
+  // parsed must cover the whole value: "8abc" is a mistake, not eight blocks
+  if (parsed.size() != value.size() || blocks <= 0 || std::numeric_limits<int>::max() < blocks) {
+    ERROR_LOG("Ignoring invalid %.*s value '%.*s'", static_cast<int>(conf->m_purge_probe_header.size()),
```
serrislew: Should this use `canLogError()`? Can this spam our error logs?

plugins/slice/client.cc
```diff
@@ -94,6 +131,28 @@ handle_client_req(TSCont contp, TSEvent event, Data *const data)
 
     data->m_req_range = rangebe;
 
+    if (data->is_purge()) {
+      // The substituted range covers block 0, so walking it would delete the head
+      if (TS_HTTP_STATUS_REQUESTED_RANGE_NOT_SATISFIABLE == data->m_statustype) {
+        ERROR_LOG("Refusing PURGE with an unparseable range");
```
serrislew: Same comment here

plugins/slice/server.cc
```diff
@@ -499,12 +502,173 @@ handleNextServerHeader(Data *const data)
...
+  int64_t const firstblock = range.firstBlockFor(blockbytes);
+  if (data->m_blocknum < firstblock) {
+    data->m_blocknum = firstblock;
+  }
+
+  if (data->m_contentlen < 0) {
+    // With no extent reported yet, the miss bound is the only end condition
+    if (data->m_purge_miss_bound <= data->m_purge_misses) {
+      DEBUG_LOG("purge gave up after %d consecutive uncached block(s)", data->m_purge_misses);
+      finish_purge(contp, data);
```
serrislew: shouldn't `range.blockIsInside` be checked regardless?

plugins/slice/client.cc
masaori335: Done by [1fda6ba](https://github.com/apache/trafficserver/pull/13475/commits/1fda6bab5ff685912031a16c1e171deb91721e58)

plugins/slice/server.cc
masaori335: Fixed by [15df451](https://github.com/apache/trafficserver/pull/13475/commits/15df451ddd70e7913adc8aed667756b4fb8a4a9f)

plugins/slice/server.cc
```diff
@@ -499,12 +502,174 @@ handleNextServerHeader(Data *const data)
...
+// downstream before this, so one uncached block cannot leak a 404 to the client.
+// A non-NONE status overrides the outcome of the walk.
+void
+finish_purge(TSCont const contp, Data *const data, TSHttpStatus const status)
+{
+  data->m_upstream.close();
+  data->m_blockstate = BlockState::Done;
+
+  TSHttpStatus const reply =
+    (TS_HTTP_STATUS_NONE != status) ? status : (0 < data->m_purge_hits ? TS_HTTP_STATUS_OK : TS_HTTP_STATUS_NOT_FOUND);
```
serrislew: I noticed that we return 200 to client as long as we have at least one purge hit since we assume that purge miss means block is not cached but we continue the path to purge. But what if there was a 5xx error from a later purge block and we stop purging. Even though we purged the beginning blocks, we could leave the later blocks cached but tell client 200 (assuming the whole object is gone). 

I think we should distinguish the non-200 responses so 404 is purge miss and that is fine, we can continue to purge down the object but if its 5xx or non-404, the client should not get a 200 for incomplete PURGE

masaori335: Good point. This assumes only 200 and 404. Let me cover error cases.

masaori335: I changed to stop walking at error.

serrislew (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as 430ba4b8b35864c38ccab31f394f280218f6548b for the 10.2.0 release.

PR #13440, opened 2026-07-28
Resolve OpenSSL 4.0 build issues

jeredfloyd:
Fedora 45 updates to OpenSSL 4.0, which removes several long-deprecated capabilities and breaks ATS build.   Specific issues fixed are lack of const tagging of pointers to immutable structures, and disallowed introspection of ASN1_STRING. Resolves #13427

PR #13476, opened 2026-08-03
Resolve OpenSSL 4.0 build issues

maskit:
## Summary

Continues #13440 ("Resolve OpenSSL 4.0 build issues"), fixing the remaining CI failures on Ubuntu 20.04 and FreeBSD 13.1 (OpenSSL 1.1.1) flagged in review.

`X509_NAME_get_index_by_NID()` takes a non-const `X509_NAME *` on OpenSSL 1.1.1 but a `const X509_NAME *` on OpenSSL 3.0+ and post-constify OpenSSL/BoringSSL. The original PR widened several `X509_NAME`-typed variables to `const` to build against the newer, constified signature, which broke the older signature at three call sites:

- `src/tscore/X509HostnameValidator.cc` — `validate_hostname()`
- `src/iocore/net/SSLUtils.cc` — `SSLMultiCertConfigLoader::load_certs_and_cross_reference_names()`
- `plugins/experimental/txn_box/plugin/src/ts_util.cc` — `ssl_value_for()`

Each call site now does `const_cast<X509_NAME *>(...)`, matching the existing precedent in `OCSPStapling.cc` for the same kind of OpenSSL-version const mismatch.

## Test plan

- [x] Built `tscore`, `inknet`, and `txn_box` targets locally against BoringSSL — all compile clean
- [x] Ran `cmake --build build -t format` — no additional changes
- [ ] CI to confirm the Ubuntu 20.04 / FreeBSD 13.1 (OpenSSL 1.1.1) jobs that failed on #13440 now pass
```

### DIFF
--- a/example/plugins/c-api/verify_cert/verify_cert.cc
+++ b/example/plugins/c-api/verify_cert/verify_cert.cc
@@ -37,7 +37,7 @@ namespace
 DbgCtl dbg_ctl{PLUGIN_NAME};
 
 static void
-debug_certificate(const char *msg, X509_NAME *name)
+debug_certificate(const char *msg, const X509_NAME *name)
 {
   BIO *bio;
 
--- a/plugins/certifier/certifier.cc
+++ b/plugins/certifier/certifier.cc
@@ -401,12 +401,18 @@ mkcrt(const std::string &commonName, int serial)
   X509_gmtime_adj(X509_get_notAfter(cert.get()), static_cast<long>(3650) * 24 * 3600);
 
   // Get handle to subject name
-  X509_NAME *n = X509_get_subject_name(cert.get());
+  X509_NAME *n = X509_NAME_dup(X509_get_subject_name(cert.get()));
   // Set common name field
   if (X509_NAME_add_entry_by_txt(n, "CN", MBSTRING_ASC, (unsigned char *)commonName.c_str(), -1, -1, 0) != 1) {
     TSError("[%s] %s: failed to add certificate subject CN", PLUGIN_NAME, __func__);
     return nullptr;
   }
+  if (X509_set_subject_name(cert.get(), n) != 1) {
+    TSError("[%s] %s: failed to set certificate subject", PLUGIN_NAME, __func__);
+    X509_NAME_free(n);
+    return nullptr;
+  }
+  X509_NAME_free(n);
 
   // Set Traffic Server public key
   if (X509_set_pubkey(cert.get(), ca_pkey_scoped.get()) == 0) {
--- a/plugins/certifier/certifier.cc
+++ b/plugins/certifier/certifier.cc
@@ -402,9 +402,14 @@ mkcrt(const std::string &commonName, int serial)
 
   // Get handle to subject name
   X509_NAME *n = X509_NAME_dup(X509_get_subject_name(cert.get()));
+  if (n == nullptr) {
+    TSError("[%s] %s: failed to duplicate certificate subject", PLUGIN_NAME, __func__);
+    return nullptr;
+  }
   // Set common name field
   if (X509_NAME_add_entry_by_txt(n, "CN", MBSTRING_ASC, (unsigned char *)commonName.c_str(), -1, -1, 0) != 1) {
     TSError("[%s] %s: failed to add certificate subject CN", PLUGIN_NAME, __func__);
+    X509_NAME_free(n);
     return nullptr;
   }
   if (X509_set_subject_name(cert.get(), n) != 1) {
--- a/plugins/certifier/certifier.cc
+++ b/plugins/certifier/certifier.cc
@@ -88,10 +88,11 @@ template <> struct default_delete<SSL_CTX> {
 } // namespace std
 
 /// Name aliases for unique pts to openSSL objects
-using scoped_X509     = std::unique_ptr<X509>;
-using scoped_X509_REQ = std::unique_ptr<X509_REQ>;
-using scoped_EVP_PKEY = std::unique_ptr<EVP_PKEY>;
-using scoped_SSL_CTX  = std::unique_ptr<SSL_CTX>;
+using scoped_X509      = std::unique_ptr<X509>;
+using scoped_X509_REQ  = std::unique_ptr<X509_REQ>;
+using scoped_EVP_PKEY  = std::unique_ptr<EVP_PKEY>;
+using scoped_SSL_CTX   = std::unique_ptr<SSL_CTX>;
+using scoped_X509_NAME = std::unique_ptr<X509_NAME, decltype(&X509_NAME_free)>;
 
 class SslLRUList
 {
@@ -401,23 +402,20 @@ mkcrt(const std::string &commonName, int serial)
   X509_gmtime_adj(X509_get_notAfter(cert.get()), static_cast<long>(3650) * 24 * 3600);
 
   // Get handle to subject name
-  X509_NAME *n = X509_NAME_dup(X509_get_subject_name(cert.get()));
+  scoped_X509_NAME n{X509_NAME_dup(X509_get_subject_name(cert.get())), X509_NAME_free};
   if (n == nullptr) {
     TSError("[%s] %s: failed to duplicate certificate subject", PLUGIN_NAME, __func__);
     return nullptr;
   }
   // Set common name field
-  if (X509_NAME_add_entry_by_txt(n, "CN", MBSTRING_ASC, (unsigned char *)commonName.c_str(), -1, -1, 0) != 1) {
+  if (X509_NAME_add_entry_by_txt(n.get(), "CN", MBSTRING_ASC, (unsigned char *)commonName.c_str(), -1, -1, 0) != 1) {
     TSError("[%s] %s: failed to add certificate subject CN", PLUGIN_NAME, __func__);
-    X509_NAME_free(n);
     return nullptr;
   }
-  if (X509_set_subject_name(cert.get(), n) != 1) {
+  if (X509_set_subject_name(cert.get(), n.get()) != 1) {
     TSError("[%s] %s: failed to set certificate subject", PLUGIN_NAME, __func__);
-    X509_NAME_free(n);
     return nullptr;
   }
-  X509_NAME_free(n);
 
   // Set Traffic Server public key
   if (X509_set_pubkey(cert.get(), ca_pkey_scoped.get()) == 0) {
--- a/src/api/InkAPI.cc
+++ b/src/api/InkAPI.cc
@@ -8316,9 +8316,9 @@ TSSslServerCertUpdate(const char *cert_path, const char *key_path)
     }
 
     // Extract common name
-    int              pos              = X509_NAME_get_index_by_NID(X509_get_subject_name(cert.get()), NID_commonName, -1);
-    X509_NAME_ENTRY *common_name      = X509_NAME_get_entry(X509_get_subject_name(cert.get()), pos);
-    ASN1_STRING     *common_name_asn1 = X509_NAME_ENTRY_get_data(common_name);
+    const int              pos              = X509_NAME_get_index_by_NID(X509_get_subject_name(cert.get()), NID_commonName, -1);
+    const X509_NAME_ENTRY *common_name      = X509_NAME_get_entry(X509_get_subject_name(cert.get()), pos);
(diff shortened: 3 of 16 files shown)

### CODE REVIEW

**Comment by JosiahWI** [comment]
Did you consider the `decltype` workaround discussed in #13440? I would like to see an explicit justification for opting to use the less type safe `const_cast` approach.

**Comment by maskit** [comment]
[approve ci]

**Comment by maskit** [comment]
@JosiahWI Thanks for the reminder. The use of `auto` made it a little cleaner. It didn't completely remove the const_casts, but I think this is ok for 10.2. After landing this, we should do cleanup that removes 1.1.1 support on master.

**Comment by JosiahWI** [comment]
I think `decltype` will preserve the CV qualification on the type, unlike `auto`.

**Comment by JosiahWI** [comment]
Oh, I see. The `auto` keyword also determines the correct constness in most cases. This looks good.

**Comment by cmcfarlen** [comment]
I had independently worked up a fix for this (#13482) before seeing this PR — standing that down in favor of yours. Handing over what I found in case it's useful, since some of it isn't covered here yet.

The diagnosis and the three `X509_NAME_get_index_by_NID` call sites match what I landed on exactly. The one thing worth flagging: **neither CI nor a BoringSSL build can validate the OpenSSL 4 side of this.** The Fedora CI job is Fedora 43 (OpenSSL 3.5), and per your own table in #13440 BoringSSL returns *non-const* from `X509_get_subject_name` — i.e. the same shape as 1.1.1. So a green CI run plus a BoringSSL build confirms the 1.1.1 regression is fixed, but can't exercise any of the const-*return* changes that OpenSSL 4 introduces.

I built this branch in a Fedora 45 / OpenSSL 4.0.1 container (`BUILD_EXPERIMENTAL_PLUGINS=ON`, `ENABLE_CRIPTS=ON`, `BUILD_REGRESSION_TESTING=ON`, `ninja -k 0` to get past the first failure). Remaining errors:

| File | Errors |
| --- | --- |
| `include/cripts/Certs.hpp` | 124 |
| `src/cripts/Certs.cc` | 4 |
| `plugins/lua/ts_lua_client_request.cc` | 4 |
| `plugins/lua/ts_lua_client_cert_helpers.h` | 3 |
| `src/iocore/net/unit_tests/test_SSLDHParams.cc` | 1 |

Specifically:

- **`plugins/lua`** — `get_x509_name_string(X509_NAME *)` needs a `const X509_NAME *` param (it only calls `X509_NAME_print_ex`, which is const on every version), and `ts_lua_client_cert_helpers.h` reads `sig->length` / `sig->data` directly, which breaks now that `ASN1_STRING` is opaque. Note this one builds by default — `plugins/lua` is gated on `if(USE_LUAJIT)`, and Fedora 45 has luajit — so it's not an opt-in path.
- **`src/cripts`** — `Certs.cc` has the same opaque-`ASN1_STRING` issue in `Signature::_load` and `_write_ip_address`. `ENABLE_CRIPTS` defaults OFF so this is lower priority.
- **`test_SSLDHParams.cc`** — `make_cert_and_key` mutates the subject name *in place*. Once the accessor returns const you have to dup, then call **both** `X509_set_subject_name` and `X509_set_issuer_name`, or the generated self-signed cert silently loses its CN.
- Also broken on OpenSSL 4 but outside a default build: `example/plugins/c-api/client_context_dump/client_context_dump.cc` and `tests/tools/plugins/ssl_client_verify_test.cc`.

Two notes on approach, take or leave:

1. `const_cast` handles the three call sites fine, but it can't handle Cripts. `CertBase::X509Value::_load_name(X509_NAME *(*getter)(const X509 *))` takes a **function pointer**, and `X509_get_subject_name`'s whole type changes in 4.0 — casting a function pointer and calling through it is UB. `X509_getm_notBefore`/`notAfter` also moved const the *opposite* direction (`const X509 *` in 1.1.1, `X509 *` in 4.0), so no single hardcoded spelling works there either. What worked for me was deducing the type from the accessor — `auto *` for locals, `decltype(&X509_get_subject_name)` for parameter and function-pointer types — which needs no casts and no version macros.
2. The `certifier.cc` change (carried from #13440) leaks the duplicated name on the `X509_NAME_add_entry_by_txt` failure path — only the `X509_set_subject_name` path frees it — and doesn't null-check `X509_NAME_dup`.

Diffs for all of the above are on `cmcfarlen:ossl4-compat` (#13482) if you want to cherry-pick or copy from them; verified green on Fedora 45/OpenSSL 4.0.1, Fedora 43/OpenSSL 3.5.7, and Ubuntu 20.04/OpenSSL 1.1.1f + clang-12, with `test_tscore` and `test_net` passing on all three. Happy to just close mine and let you carry it however you prefer.

File: src/tscore/X509HostnameValidator.cc [hunk 1]
```diff
@@ -269,13 +268,13 @@ validate_hostname(X509 *x, std::string_view hostname, bool is_ip, char **peernam
...
   while ((i = X509_NAME_get_index_by_NID(name, NID_commonName, i)) >= 0) {
-    ASN1_STRING   *str;
-    int            astrlen;
-    unsigned char *astr;
+    const ASN1_STRING *str;
+    int                astrlen;
+    unsigned char     *astr;
     str = X509_NAME_ENTRY_get_data(X509_NAME_get_entry(name, i));
     // Convert to UTF-8
     astrlen = ASN1_STRING_to_UTF8(&astr, str);
```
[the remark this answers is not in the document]
maskit: Confirmed real, but pre-existing on master since the function's original addition in 2015, unrelated to this PR's OpenSSL 4 changes. Filed as #13483.

src/api/InkAPI.cc
```diff
@@ -8316,9 +8316,9 @@ TSSslServerCertUpdate(const char *cert_path, const char *key_path)
...
 
     // Extract common name
-    int              pos              = X509_NAME_get_index_by_NID(X509_get_subject_name(cert.get()), NID_commonName, -1);
-    X509_NAME_ENTRY *common_name      = X509_NAME_get_entry(X509_get_subject_name(cert.get()), pos);
-    ASN1_STRING     *common_name_asn1 = X509_NAME_ENTRY_get_data(common_name);
+    const int              pos              = X509_NAME_get_index_by_NID(X509_get_subject_name(cert.get()), NID_commonName, -1);
+    const X509_NAME_ENTRY *common_name      = X509_NAME_get_entry(X509_get_subject_name(cert.get()), pos);
+    const ASN1_STRING     *common_name_asn1 = X509_NAME_ENTRY_get_data(common_name);
     char *common_name_str = reinterpret_cast<char *>(const_cast<unsigned char *>(ASN1_STRING_get0_data(common_name_asn1)));
     if (ASN1_STRING_length(common_name_asn1) != static_cast<int>(strlen(common_name_str))) {
```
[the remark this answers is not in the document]
maskit: Confirmed real, but pre-existing on master since 2024, unrelated to this PR's OpenSSL 4 changes. Filed as #13484.

maskit:
[approve ci autest 2]

src/tscore/X509HostnameValidator.cc
```diff
@@ -205,12 +205,12 @@ do_check_string(ASN1_STRING *a, int cmp_type, equal_fn equal, const unsigned cha
...
-  if (!a->data || !a->length || cmp_type != a->type) {
+  if (!ASN1_STRING_get0_data(a) || !ASN1_STRING_length(a) || cmp_type != ASN1_STRING_type(a)) {
     return false;
   }
-  retval = equal(a->data, a->length, b, blen);
+  retval = equal(ASN1_STRING_get0_data(a), ASN1_STRING_length(a), b, blen);
   if (retval && peername) {
-    *peername = ats_strndup((char *)a->data, a->length);
+    *peername = ats_strndup((char *)ASN1_STRING_get0_data(a), ASN1_STRING_length(a));
   }
```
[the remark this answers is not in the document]
maskit: Fixed.

plugins/certifier/certifier.cc
```diff
@@ -367,12 +367,18 @@ mkcrt(const std::string &commonName, int serial)
   X509_gmtime_adj(X509_get_notAfter(cert.get()), static_cast<long>(3650) * 24 * 3600);
 
   // Get handle to subject name
-  X509_NAME *n = X509_get_subject_name(cert.get());
+  X509_NAME *n = X509_NAME_dup(X509_get_subject_name(cert.get()));
   // Set common name field
   if (X509_NAME_add_entry_by_txt(n, "CN", MBSTRING_ASC, (unsigned char *)commonName.c_str(), -1, -1, 0) != 1) {
     TSError("[%s] %s: failed to add certificate subject CN", PLUGIN_NAME, __func__);
     return nullptr;
```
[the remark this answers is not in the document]
maskit: Fixed.

maskit:
I've updated the change and addressed all the comments. New code does not use const_cast at all now. I did see `const_cast`s around the changes but those are just few of 500+. Those should be fixed separately.

maskit:
@jeredfloyd I added changes on top of your commits to cover the entire source tree. I'd appreciate it if you could test this change for Fedora 45.

plugins/certifier/certifier.cc
```diff
@@ -401,12 +402,20 @@ mkcrt(const std::string &commonName, int serial)
   X509_gmtime_adj(X509_get_notAfter(cert.get()), static_cast<long>(3650) * 24 * 3600);
 
   // Get handle to subject name
-  X509_NAME *n = X509_get_subject_name(cert.get());
+  scoped_X509_NAME n{X509_NAME_dup(X509_get_subject_name(cert.get())), X509_NAME_free};
+  if (n == nullptr) {
+    TSError("[%s] %s: failed to duplicate certificate subject", PLUGIN_NAME, __func__);
+    return nullptr;
+  }
```
JosiahWI, line 409: Error checking added in this patch. Very good! Thanks @maskit !

plugins/certifier/certifier.cc
```diff
@@ -401,12 +402,20 @@ mkcrt(const std::string &commonName, int serial)
...
   // Set common name field
-  if (X509_NAME_add_entry_by_txt(n, "CN", MBSTRING_ASC, (unsigned char *)commonName.c_str(), -1, -1, 0) != 1) {
+  if (X509_NAME_add_entry_by_txt(n.get(), "CN", MBSTRING_ASC, (unsigned char *)commonName.c_str(), -1, -1, 0) != 1) {
     TSError("[%s] %s: failed to add certificate subject CN", PLUGIN_NAME, __func__);
     return nullptr;
   }
+  if (X509_set_subject_name(cert.get(), n.get()) != 1) {
+    TSError("[%s] %s: failed to set certificate subject", PLUGIN_NAME, __func__);
+    return nullptr;
+  }
```
JosiahWI, line 418: What is the purpose of this added logic?

example/plugins/c-api/verify_cert/verify_cert.cc
```diff
@@ -37,7 +37,7 @@ namespace
 DbgCtl dbg_ctl{PLUGIN_NAME};
 
 static void
-debug_certificate(const char *msg, X509_NAME *name)
+debug_certificate(const char *msg, const X509_NAME *name)
```
JosiahWI, line 40: This is correct! The parameter can be `const` because `X509_NAME_print_ex` takes a const pointer to the name since OpenSSL 1.1.1.

JosiahWI (commented):
This looks great to me. Will wait for @jeredfloyd to check before approving. I have one question about the certifier logic.

plugins/certifier/certifier.cc
maskit: The return type of `X509_get_subject_name` may be const now, and that's why we have to duplicate the returned value before calling `X509_NAME_add_entry_by_txt`. Since the call doesn't directly modify the source, we need to set it as a new value.

JosiahWI: That makes sense. Thanks for walking me through that!

jeredfloyd (approved):
Looks good to me, and confirmed this builds on Fedora 45 and EPEL 10.  (I'm old and prefer explicit types over inferred wherever possible, and this looks like a reasonable balance.)

JosiahWI (approved):
Approved!

cmcfarlen:
Cherry-picked to the 10.2.x branch as 2b7fb195db335311e9e4fd2eb66c05e0420ec661 for the 10.2.0 release.

PR #13480, opened 2026-08-03
Fix enabling per-server metrics that disables the outbound keep-alive minimum 

serrislew:
`ConnectionTracker::Group` kept its connection count in two mutually exclusive places: `TxnState::reserve()` returned the metric gauge *instead of* incrementing `Group::_count` when `metric_enabled` is set, and never touched `_count` at all. `release()` had the same split.
  
This was internally consistent — every read inside the tracker also branched on `_count_metric != nullptr`, so connection-max enforcement and the JSON/`dump` output all saw the right number. But `_count` has one consumer *outside* the tracker, in `ServerSessionPool::eventHandler()`: https://github.com/apache/trafficserver/blob/master/src/proxy/http/HttpSessionManager.cc#L316
  
  ```cpp
  bool connection_count_below_min = s->conn_track_group->_count <= s->conn_track_group->_min_keep_alive_conns;
  ```
With `proxy.config.http.per_server.connection.metric_enabled` non-zero, _count stays at 0, so this is always true and an idle keep-alive session is never closed on inactivity timeout — per_server.connection.min is effectively unbounded and idle origin connections accumulate.

Scope: the keep-alive minimum is the only affected behavior. Connection-max enforcement, blocking, and the exported metric values are unchanged.

Introduced from #12250

### DIFF
--- a/include/iocore/net/ConnectionTracker.h
+++ b/include/iocore/net/ConnectionTracker.h
@@ -449,7 +449,7 @@ ConnectionTracker::TxnState::reserve()
 {
   _reserved_p = true;
   // @a _count is always the authoritative count; the metrics, if enabled, only mirror it.
-  int count = ++_g->_count;
+  auto count = ++_g->_count;
   if (_g->_count_metric != nullptr) {
     ts::Metrics::Gauge::increment(_g->_count_metric);
     ts::Metrics::Counter::increment(_g->_count_total_metric);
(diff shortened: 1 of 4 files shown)


bneradt (changes requested):
I think the change makes sense. Can we add an autest for this though?

bneradt (dismissed):

bneradt (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as f0f99ec9f8a485eb1c90f35962ca5e9dfb4af0f9 for the 10.2.0 release.

PR #13488, opened 2026-08-04
Fix wrong variable in eventloop events max metric

moonchen:
The code for counting eventloop events max used the incorrect `_count` variable, which stores the number of _iterations_, rather than the number of _events_. `proxy.process.eventloop.events.max.*` has therefore been reporting an iteration count. Fix it by changing the code to use `count`.


JosiahWI (approved):
Whoops! Good catch!

cmcfarlen:
Cherry-picked to the 10.2.x branch as 0f07a58a718418fb3f1b0f4e3732966d3a2aab04 for the 10.2.0 release.

PR #13492, opened 2026-08-04
Track minimum-only origin connections

bneradt:
A per-server keep-alive minimum without a connection maximum or
per-server metrics created a tracker group but never reserved it.
The count remained zero, so idle origin sessions were retained
without limit and their eventual release reported invalid accounting.

Reserve the tracker whenever the keep-alive minimum needs it, and add
replay coverage that lets surplus pooled sessions expire while
preserving the configured minimum.

---

This added test fails, giving a value of 0 without this patch, and
showing the tracked value (1) with it.

### DIFF
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -5863,7 +5863,8 @@ HttpSM::do_http_server_open(bool raw, bool only_direct)
     }
 
     ct_state.update_max_count(ccount);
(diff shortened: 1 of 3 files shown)


serrislew (approved):
Nice catch, thanks!

cmcfarlen:
Cherry-picked to the 10.2.x branch as 817c7df56e16239d3e8507c6a993516b310d4905 for the 10.2.0 release.

PR #13498, opened 2026-08-05
Fix bad return in `validate_hostname`

### LINKED ISSUE #13483: validate_hostname() returns -1 on UTF-8 conversion failure, which converts to true in its bool return type
validate_hostname() in src/tscore/X509HostnameValidator.cc has return type bool, but on ASN1_STRING_to_UTF8() failure it does:

```cpp
astrlen = ASN1_STRING_to_UTF8(&astr, str);

if (astrlen < 0) {
  return -1;
}
```

-1 implicitly converts to true when returned from a bool function, so a UTF-8 conversion failure on the certificate CN is reported as a successful hostname match instead of a failure.

This traces back to the original addition of the function in 1649abc30e9 (2015) and is present on current master, independent of any other in-flight work. It was noticed during review of #13476 (OpenSSL 4 build-compatibility fixes), which touches nearby lines for unrelated reasons but does not change this logic.

## Suggested fix

return false; instead of return -1;

JosiahWI:
The function has a `bool` return type, but returned `-1` to indicate an error on UTF-8 conversion failure. This patch corrects the return statement to return `false`.

### DIFF
--- a/src/tscore/X509HostnameValidator.cc
+++ b/src/tscore/X509HostnameValidator.cc
@@ -281,7 +281,7 @@ validate_hostname(X509 *x, std::string_view hostname, bool is_ip, char **peernam
     astrlen = ASN1_STRING_to_UTF8(&astr, str);
 
     if (astrlen < 0) {
-      return -1;
+      return false;
     }
     retval = equal(astr, astrlen, hostname_data, hostname_len);
     if (retval && peername) {


JosiahWI:
[approve ci autest 0]

bneradt (approved):

maskit (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as 79861fb489f11985be527272286a470bbd54ef0c for the 10.2.0 release.

PR #13503, opened 2026-08-05
Anchor the 404 check in tls_check_cert_select_plugin

### LINKED ISSUE #13501: tls_check_cert_select_plugin: unanchored "404" assertion collides with ephemeral port numbers
## Description

`tests/gold_tests/tls/tls_check_cert_select_plugin.test.py` fails intermittently in CI, depending purely on which ephemeral port AuTest happens to assign to the `ts` process.

The test asserts on the bare substring `"404"`:

```python
tr.Processes.Default.Streams.All += Testers.ExcludesExpression("404", "Should make an exchange")
```

That substring is not anchored to an HTTP status line, so it also matches the *port number* in ordinary curl output. When AuTest assigns `ts` a port containing `404` — for example `62404` — the excluded expression matches lines like:

```
* Added bar.com:62404:127.0.0.1 to DNS cache
*   Trying 127.0.0.1:62404...
```

and the run fails:

```
file .../6-tr-Default/stream.all.txt : Should make an exchange - Failed
   Reason: Contents ... contains expression: "404"
     Details:
       * Added bar.com:62404:127.0.0.1 to DNS cache : 1
         ... 0*   Trying 127.0.0.1:62404... : 6
```

Test run 6 ("Test new version of bar cert with bad CA") is the one that fails, because it is the only run using `ExcludesExpression` for `"404"`. The `ContainsExpression("404")` assertions in the other runs are spuriously *satisfied* by the same collision, which is arguably worse — they can pass without a real HTTP exchange having occurred.

## Second, related bug in the same file

Four assertions write to the wrong test run variable. Lines 103, 116, and 130 use `tr` where the surrounding run is `tr2`:

```python
tr2 = Test.AddTestRun("foo.com cert")
...
tr2.Processes.Default.Streams.All = Testers.ExcludesExpression("Could Not Connect", ...)
tr2.Processes.Default.Streams.All += Testers.ContainsExpression("CN=foo.com", ...)
tr2.Processes.Default.Streams.All += Testers.ExcludesExpression("CN=bar.com", ...)
tr.Processes.Default.Streams.All += Testers.ContainsExpression("404", "Should make an exchange")   # <-- tr, not tr2
```

The effect is that the "Should make an exchange" check for runs 1, 2 and 3 is appended to run 0 instead. Run 0's report shows "Should make an exchange" four times, while runs 1-3 never assert it at all. This also partially masks the port collision, since those misdirected checks are `Contains` rather than `Excludes`.

## Suggested fix

1. Anchor the assertion so it cannot match a port number — e.g. match `HTTP/1.1 404` or the curl status line `< 404`, rather than the bare digits.
2. Change the three misdirected `tr.` references to `tr2.` so runs 1-3 assert what they were meant to.

## Affected versions

Both bugs were introduced together in 7dbb6cb188 ("Add hook for loading certificate and key data from plugin", #6609, 2021-06-25), verified with `git log -S` on the exact strings. The file has not been meaningfully changed since in this respect, so this affects every current line: `9.2.x`, `10.0.x`, `10.1.x`, `10.2.x` and `master`. The earliest release tag containing it is `9.2.0`.

## Notes

Not a regression from any recent change — it surfaced while reviewing an unrelated 10.2.x backport batch, where the shard drew port `62404`. Re-running the shard makes it pass, which is consistent with the port-draw explanation.

cmcfarlen:
## Description

`tls_check_cert_select_plugin.test.py` fails intermittently in CI depending only on which ephemeral port AuTest hands the `ts` process. It asserted on the bare substring `"404"`:

```python
tr.Processes.Default.Streams.All += Testers.ExcludesExpression("404", "Should make an exchange")
```

`ContainsExpression`/`ExcludesExpression` are regexes searched per line, and `404` is unanchored, so it also matches the **port number** in ordinary curl output. When the port contains `404` — e.g. `62404` — the excluded expression matches lines like:

```
* Added bar.com:62404:127.0.0.1 to DNS cache
*   Trying 127.0.0.1:62404...
```

and the last test run fails:

```
file .../6-tr-Default/stream.all.txt : Should make an exchange - Failed
   Reason: Contents ... contains expression: "404"
     Details:
       * Added bar.com:62404:127.0.0.1 to DNS cache : 1
         ... 0*   Trying 127.0.0.1:62404... : 6
```

Run 6 ("Test new version of bar cert with bad CA") is the one that fails because it is the only one using `ExcludesExpression`. The `ContainsExpression("404")` assertions elsewhere are spuriously *satisfied* by the same collision, which is arguably worse since they can pass without a real HTTP exchange.

This is currently causing 10.2.x PR runs to fail.

## Why a 404 is the right thing to assert

The origin registers a response only for `GET / HTTP/1.1` with no `Host` header, but `proxy.config.url_remap.pristine_host_hdr` is enabled, so every request forwards its original `Host` and does not match. The 404 from the origin is what proves the exchange completed. That intent is preserved; only the pattern changes.

## Changes

1. Match the response status line via a named constant instead of bare digits:

```python
HTTP_404 = r"HTTP/[\d.]+ 404"
```

This cannot match a port number, and covers both `HTTP/1.1 404 Not Found` and HTTP/2's `HTTP/2 404`, since the TLS port offers h2 via ALPN.

2. Fix three assertions that referenced the wrong test run variable. Lines that belonged to `tr2` blocks used `tr.`, so the "Should make an exchange" check for runs 1-3 was appended to run 0 instead. Run 0's report shows it four times while runs 1-3 never asserted it at all. They now point at their own run, so those three runs actually verify what they were meant to.

## Test plan

I verified the new pattern against the exact lines from the CI failure plus the status lines it must still match:

| Line | new pattern | old `"404"` | want |
|---|---|---|---|
| `* Added bar.com:62404:127.0.0.1 to DNS cache` | no | **yes** | no |
| `*   Trying 127.0.0.1:62404...` | no | **yes** | no |
| `* Connected to bar.com (127.0.0.1) port 62404` | no | **yes** | no |
| `< HTTP/1.1 404 Not Found` | yes | yes | yes |
| `< HTTP/2 404` | yes | yes | yes |
| `< HTTP/1.1 200 OK` | no | no | no |

All nine cases I checked behave correctly; the old pattern false-positived on three port lines that the new one rejects.

`yapf` 0.43.0 (the version pinned in `tools/yapf.sh`) reports the file clean.

Note that fixing the run-variable references *activates* three assertions that were previously inert, so CI is the real verification for those three runs. All four runs go through the same `map /` remap to the same origin with a non-matching `Host`, so all four are expected to 404 identically — but if a run turns out not to, that part can be dropped and handled separately.

## Backport

Both bugs date to 7dbb6cb188 (#6609, 2021-06-25), verified with `git log -S` on the exact strings, so this affects `9.2.x`, `10.0.x`, `10.1.x`, `10.2.x` and `master`. The assertion lines are identical on master and 10.2.x — only the unrelated `ssl_multicert` block differs — so this should cherry-pick cleanly. 10.2.x wants it for the RC.

Fixes #13501

### DIFF
--- a/tests/gold_tests/tls/tls_check_cert_select_plugin.test.py
+++ b/tests/gold_tests/tls/tls_check_cert_select_plugin.test.py
@@ -22,6 +22,13 @@ Test.Summary = '''
 Test ATS offering different certificates based on SNI. Load via plugin
 '''
 
+# The origin only has a response registered for a request with no Host header, so with
+# pristine_host_hdr every request 404s at the origin. A 404 therefore proves the exchange
+# happened. Match the response status line rather than a bare "404", which also matches the
+# ephemeral port number curl prints (a port such as 62404 made this test fail at random).
+# Covers both HTTP/1.1 ("HTTP/1.1 404 Not Found") and HTTP/2 ("HTTP/2 404").
+HTTP_404 = r"HTTP/[\d.]+ 404"
+
 # Define default ATS
 ts = Test.MakeATSProcess("ts", enable_tls=True)
 server = Test.MakeOriginServer("server", ssl=True)
@@ -94,7 +101,7 @@ tr.StillRunningAfter = ts
 tr.Processes.Default.Streams.All = Testers.ExcludesExpression("Could Not Connect", "Curl attempt should have succeeded")
 tr.Processes.Default.Streams.All += Testers.ContainsExpression("CN=bar.com", "Cert should contain bar.com")
 tr.Processes.Default.Streams.All += Testers.ExcludesExpression("CN=foo.com", "Cert should not contain foo.com")
-tr.Processes.Default.Streams.All += Testers.ContainsExpression("404", "Should make an exchange")
+tr.Processes.Default.Streams.All += Testers.ContainsExpression(HTTP_404, "Should make an exchange")
 
 # Should receive a foo.com cert
 tr2 = Test.AddTestRun("foo.com cert")
@@ -106,7 +113,7 @@ tr2.StillRunningAfter = ts
 tr2.Processes.Default.Streams.All = Testers.ExcludesExpression("Could Not Connect", "Curl attempt should have succeeded")
 tr2.Processes.Default.Streams.All += Testers.ContainsExpression("CN=foo.com", "Cert should contain foo.com")
 tr2.Processes.Default.Streams.All += Testers.ExcludesExpression("CN=bar.com", "Cert should not contain bar.com")
-tr.Processes.Default.Streams.All += Testers.ContainsExpression("404", "Should make an exchange")
+tr2.Processes.Default.Streams.All += Testers.ContainsExpression(HTTP_404, "Should make an exchange")
 
 # Should receive random.server.com
 tr2 = Test.AddTestRun("random.server.com cert")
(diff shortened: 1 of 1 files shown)


bneradt (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as e4ea73bd3d97a220f1d16bbee87e6f2869514878 for the 10.2.0 release.

PR #13505, opened 2026-08-06
Add hidden metrics and MAX/MIN/incremental derived metric aggregation

cmcfarlen:
Adds two facilities to `ts::Metrics`, plus a `traffic_ctl` option to inspect the first, and fixes three pre-existing defects found along the way. A follow-up PR uses these for per-upstream-server connection metrics; this PR is independently useful and stands on its own.

### Hidden metrics

A second `Storage` instance, reached through `Metrics::hidden_instance()`, for high-cardinality intermediate values that are worth recording but not worth publishing. `Gauge::createHiddenPtr` / `Counter::createHiddenPtr` return the same correctly typed pointer as `createPtr`, so a hidden metric is read and written with the ordinary typed mutators with no cast at the call site.

A separate store rather than a per-metric "hidden" flag is deliberate: it makes hidden metrics *structurally* unreachable from the published store, so no consumer can expose one by forgetting to check a flag.

Since that also makes them hard to debug, `traffic_ctl metric match --include-hidden` lists them. The rec type bit for this (`RECT_HIDDEN_METRIC = 0x40`) sits deliberately outside `RECT_ALL` (`0x3F`), so hidden metrics are returned only when explicitly asked for and never as a side effect of a broad query.

### Derived metric aggregation

- `MAX` and `MIN` in addition to `SUM`. The accumulator is seeded from the first source rather than from zero, since a zero seed is only correct for `SUM` and would clamp `MIN` to `<= 0`.
- `Derived::add_source()`, for aggregates whose sources are discovered while the process runs rather than known at startup. Repeatedly calling `derive()` for one derived name does not work for this: it appends a separate entry per call, all targeting the same metric, so each update overwrites the others with its own subset and the last writer silently wins. `add_source()` accumulates into a single entry, and re-registering an existing source is a no-op.

### Pre-existing defects fixed

- **`Storage::create()` had no exhaustion check.** Filling the last blob let the following bookkeeping call `addBlob()` and write one past the end of `_blobs`. Verified by temporarily shrinking `MAX_BLOBS`: the current code segfaults, and the `debug_assert` in `addBlob()` does not catch it because it is off by one against the access it guards (`_blobs[++_cur_blob]`) — and being a `debug_assert`, it is compiled out of release builds entirely. Now a `release_assert` against `MAX_BLOBS - 1`, with `create()` refusing the final slot and returning the reserved `bad_id`.
- **Unresolvable derived sources were not skipped.** A source given by name or id that does not resolve was still passed to `lookup()`, which masks the unresolved id down to the reserved `bad_id` slot, so the aggregate silently included that slot's value. Observable under `MAX`/`MIN`, where the `bad_id` value can become the winning one; under `SUM` it was hidden by `bad_id` holding zero.
- **The metrics unit tests were order dependent.** They asserted absolute metric ids and iterator positions, which only hold when the case runs first against an otherwise empty store. Any other case that creates a published metric made them fail. This lands first so the fragility is never introduced.

### Testing

- Unit tests grow from 487 assertions / 32 cases to 6208 / 36, and pass under `--order rand` across many seeds. Each fix was checked against the unfixed code first to confirm the new assertions actually discriminate.
- `tests/gold_tests/jsonrpc/metric_match_include_hidden.test.py` covers the `--include-hidden` RPC round trip end to end. This is load-bearing: the rec type also has to be accepted by the JSONRPC request decoder, which validates each requested type against a whitelist and rejects the entire request otherwise. That was invisible to every build-level check and only showed up end to end.
- Verified against a running `traffic_server` with two temporary hidden metrics registered: absent from `metric match`, present with `--include-hidden`, and absent from a broad `proxy.process` query of ~900 published metrics.
- Docs build clean under `-W` with `nitpicky = True`.

Co-authored-by: @serrislew

### DIFF
--- a/doc/appendices/command-line/traffic_ctl.en.rst
+++ b/doc/appendices/command-line/traffic_ctl.en.rst
@@ -949,12 +949,19 @@ traffic_ctl metric
    Display the current value of the specified statistics.
 
 .. program:: traffic_ctl metric
-.. option:: match REGEX [REGEX...]
+.. option:: match [--include-hidden] REGEX [REGEX...]
 
    :ref:`admin_lookup_records`
 
    Display the current values of all statistics whose names match
    the given regular expression.
+
+.. option:: --include-hidden
+
+   Also match hidden metrics. Hidden metrics are internal metrics that are stored but never
+   published through the normal metrics registry; they are not part of the stable metric
+   contract and may be added, changed, or removed between releases without notice. This
+   option is intended for debugging.
 
 .. program:: traffic_ctl metric
 .. option:: describe RECORD [RECORD...]
--- a/doc/appendices/command-line/traffic_ctl.en.rst
+++ b/doc/appendices/command-line/traffic_ctl.en.rst
@@ -956,6 +956,7 @@ traffic_ctl metric
    Display the current values of all statistics whose names match
    the given regular expression.
 
+.. program:: traffic_ctl metric match
 .. option:: --include-hidden
 
    Also match hidden metrics. Hidden metrics are internal metrics that are stored but never
--- a/src/records/RecCore.cc
+++ b/src/records/RecCore.cc
@@ -612,6 +612,22 @@ RecLookupMatchingRecords(unsigned rec_type, const char *match, void (*callback)(
     });
   }
 
+  if (rec_type & RECT_HIDDEN_METRIC) {
+    // Opt-in only: hidden metrics are never reachable through RECT_ALL, see RecDefs.h.
+    for (auto &&[name, type, val] : ts::Metrics::hidden_instance()) {
+      if (regex.exec(name.data())) {
+        RecRecord tmp;
+
+        tmp.rec_type = RECT_PROCESS;
+
+        tmp.name         = name.data();
+        tmp.data_type    = type == ts::Metrics::MetricType::COUNTER ? RECD_COUNTER : RECD_INT;
+        tmp.data.rec_int = val;
+        callback(&tmp, data);
+      }
+    }
+  }
+
   int num_records = g_num_records;
   for (int i = 0; i < num_records; i++) {
     RecRecord *r = &(g_records[i]);
--- a/src/records/RecCore.cc
+++ b/src/records/RecCore.cc
@@ -614,7 +614,16 @@ RecLookupMatchingRecords(unsigned rec_type, const char *match, void (*callback)(
 
   if (rec_type & RECT_HIDDEN_METRIC) {
     // Opt-in only: hidden metrics are never reachable through RECT_ALL, see RecDefs.h.
-    for (auto &&[name, type, val] : ts::Metrics::hidden_instance()) {
+    auto &hidden = ts::Metrics::hidden_instance();
+    // Slot 0 of every Storage is the reserved bad_id placeholder, so it exists under the same name
+    // in both stores. Skip it here, otherwise a query matching it returns two identically named
+    // records that differ only in value.
+    auto it = hidden.begin();
+
+    ++it;
+    for (; it != hidden.end(); ++it) {
+      auto &&[name, type, val] = *it;
+
       if (regex.exec(name.data())) {
         RecRecord tmp{};
 
--- a/src/records/RecCore.cc
+++ b/src/records/RecCore.cc
@@ -618,7 +618,12 @@ RecLookupMatchingRecords(unsigned rec_type, const char *match, void (*callback)(
       if (regex.exec(name.data())) {
         RecRecord tmp;
 
-        tmp.rec_type = RECT_PROCESS;
+        // Tag both bits so that a caller asking only for RECT_HIDDEN_METRIC passes the rec_type
+        // check the lookup callback applies to every record it is handed. Note this combination
+        // satisfies neither REC_TYPE_IS_STAT nor REC_TYPE_IS_CONFIG (both compare for equality),
+        // so the YAML encoder emits no stat_meta block for hidden metrics. That is intentional:
+        // the meta fields of this synthetic record were never populated.
+        tmp.rec_type = static_cast<RecT>(RECT_PROCESS | RECT_HIDDEN_METRIC);
 
         tmp.name         = name.data();
         tmp.data_type    = type == ts::Metrics::MetricType::COUNTER ? RECD_COUNTER : RECD_INT;
--- a/src/records/RecCore.cc
+++ b/src/records/RecCore.cc
@@ -521,7 +521,7 @@ RecLookupRecord(const char *name, void (*callback)(const RecRecord *, void *), v
   auto         it      = metrics.find(name);
 
   if (it != metrics.end()) {
-    RecRecord r;
+    RecRecord r{};
     auto &&[name, type, val] = *it;
 
     r.rec_type     = RECT_PLUGIN;
@@ -554,7 +554,7 @@ RecLookupRecord(const char *name, void (*callback)(const RecRecord *, void *), v
       auto &strings = ts::Metrics::StaticString::instance();
 
       if (auto m = strings.lookup(std::string{name}); m) {
-        RecRecord r;
+        RecRecord r{};
         r.rec_type                = RECT_PLUGIN;
         r.data_type               = RECD_STRING;
         r.name                    = name;
@@ -585,7 +585,7 @@ RecLookupMatchingRecords(unsigned rec_type, const char *match, void (*callback)(
     // librecords callback with a "pseudo" record.
     for (auto &&[name, type, val] : ts::Metrics::instance()) {
       if (regex.exec(name.data())) {
-        RecRecord tmp;
+        RecRecord tmp{};
 
         tmp.rec_type = RECT_PROCESS;
 
@@ -598,7 +598,7 @@ RecLookupMatchingRecords(unsigned rec_type, const char *match, void (*callback)(
     // Finally check string metrics
     ts::Metrics::StaticString::instance().for_each([&](const std::string &name, const std::string &value) {
       if (regex.exec(name)) {
-        RecRecord tmp;
+        RecRecord tmp{};
 
         tmp.rec_type = RECT_PROCESS;
 
@@ -616,7 +616,7 @@ RecLookupMatchingRecords(unsigned rec_type, const char *match, void (*callback)(
     // Opt-in only: hidden metrics are never reachable through RECT_ALL, see RecDefs.h.
     for (auto &&[name, type, val] : ts::Metrics::hidden_instance()) {
       if (regex.exec(name.data())) {
-        RecRecord tmp;
+        RecRecord tmp{};
 
         // Tag both bits so that a caller asking only for RECT_HIDDEN_METRIC passes the rec_type
         // check the lookup callback applies to every record it is handed. Note this combination
--- a/src/traffic_ctl/traffic_ctl.cc
+++ b/src/traffic_ctl/traffic_ctl.cc
@@ -249,7 +249,9 @@ main([[maybe_unused]] int argc, const char **argv)
     .add_example_usage("traffic_ctl metric get METRIC [METRIC ...]");
   metric_command.add_command("describe", "Show detailed information about one or more metric values", "", MORE_THAN_ONE_ARG_N,
                              Command_Execute); // not implemented
-  metric_command.add_command("match", "Get metrics matching a regular expression", "", MORE_THAN_ZERO_ARG_N, Command_Execute);
+  metric_command.add_command("match", "Get metrics matching a regular expression", "", MORE_THAN_ZERO_ARG_N, Command_Execute)
+    .add_option("--include-hidden", "", "Also match hidden (internal, normally unpublished) metrics")
+    .add_example_usage("traffic_ctl metric match METRIC [--include-hidden]");
   metric_command
     .add_command(
       "monitor",
--- a/src/traffic_ctl/traffic_ctl.cc
+++ b/src/traffic_ctl/traffic_ctl.cc
@@ -251,7 +251,7 @@ main([[maybe_unused]] int argc, const char **argv)
                              Command_Execute); // not implemented
   metric_command.add_command("match", "Get metrics matching a regular expression", "", MORE_THAN_ZERO_ARG_N, Command_Execute)
     .add_option("--include-hidden", "", "Also match hidden (internal, normally unpublished) metrics")
-    .add_example_usage("traffic_ctl metric match METRIC [--include-hidden]");
+    .add_example_usage("traffic_ctl metric match [--include-hidden] METRIC [METRIC ...]");
   metric_command
     .add_command(
       "monitor",
--- a/src/tsutil/Metrics.cc
+++ b/src/tsutil/Metrics.cc
@@ -42,6 +42,16 @@ Metrics::instance()
   return _instance;
 }
 
+Metrics &
+Metrics::hidden_instance()
+{
+  // Separate storage from instance(). Hidden metrics are never published.
+  static std::shared_ptr<Storage> _hidden_store = std::make_shared<Storage>();
+  thread_local Metrics            _instance(_hidden_store);
+
+  return _instance;
+}
+
 void
 Metrics::Storage::addBlob() // The mutex must be held before calling this!
 {
--- a/src/tsutil/Metrics.cc
+++ b/src/tsutil/Metrics.cc
@@ -295,6 +295,26 @@ namespace details
       metrics.push_back(std::move(m));
     }
 
+    void
+    add_source(Metrics::IdType id, Metrics::AtomicType *source, Metrics::Derived::Op op)
+    {
+      if (!source) {
+        return;
+      }
+
+      std::lock_guard l(metrics_lock);
+      auto            it = std::find_if(metrics.begin(), metrics.end(), [id](DerivedMetric const &m) { return m.metric == id; });
+
+      if (it == metrics.end()) {
+        metrics.push_back(DerivedMetric{id, {source}, op});
+        return;
+      }
+      // Already registered sources are skipped so repeated registration is harmless.
+      if (std::find(it->derived_from.begin(), it->derived_from.end(), source) == it->derived_from.end()) {
+        it->derived_from.push_back(source);
+      }
+    }
+
     static DerivativeMetrics &
     instance()
     {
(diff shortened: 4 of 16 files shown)


src/records/RecCore.cc
```diff
@@ -612,6 +612,27 @@ RecLookupMatchingRecords(unsigned rec_type, const char *match, void (*callback)(
     });
   }
 
+  if (rec_type & RECT_HIDDEN_METRIC) {
+    // Opt-in only: hidden metrics are never reachable through RECT_ALL, see RecDefs.h.
+    for (auto &&[name, type, val] : ts::Metrics::hidden_instance()) {
+      if (regex.exec(name.data())) {
+        RecRecord tmp;
```
bryancall: `RecRecord tmp;` leaves `version`, `registered`, `rsb_id`, `order` and `data_default` indeterminate, and `convert<RecRecord>::encode` reads all of them unconditionally (`convert.h` lines 121-126, plus `data_default` in the RECD_INT and RECD_COUNTER cases). The path is reachable: `RecordsUtils.cc` hands each callback record straight to the encoder, so `traffic_ctl --format json metric match --include-hidden` can emit different version, rsb and order values on successive runs. Reading `registered`, an indeterminate `bool`, is undefined behavior that a MemorySanitizer build would flag.

Your comment correctly explains why no `stat_meta` block is emitted, which covers the fields Copilot was most concerned about, but it does not cover these five. `RecRecord tmp{};` zero-initializes the whole struct and costs nothing.

Not blocking, because the published-metric loop about 25 lines up has had the identical pattern for years, so this is not a regression. Since you are already in the file, worth fixing in both places.

src/records/RecCore.cc
```diff
@@ -612,6 +612,27 @@ RecLookupMatchingRecords(unsigned rec_type, const char *match, void (*callback)(
     });
   }
 
+  if (rec_type & RECT_HIDDEN_METRIC) {
+    // Opt-in only: hidden metrics are never reachable through RECT_ALL, see RecDefs.h.
+    for (auto &&[name, type, val] : ts::Metrics::hidden_instance()) {
```
bryancall: The hidden-store iteration starts at `begin()`, which is the reserved `bad_id` slot every `Storage` constructor creates. So `traffic_ctl metric match --include-hidden bad_id` returns two records both named `proxy.process.api.metrics.bad_id`, one from each store, with different values. Cosmetic, but confusing in exactly the debugging situation this flag exists for. Skipping id 0 in this loop avoids it.

src/tsutil/Metrics.cc
```diff
@@ -64,6 +76,13 @@ Metrics::Storage::create(std::string_view name, const MetricType type)
     return it->second;
   }
 
+  // The slot is written below and the bookkeeping only then advances, calling addBlob() once
+  // _cur_off reaches MAX_SIZE. Refusing the final slot of the final blob keeps addBlob() from
+  // ever being reached in an exhausted store, at a cost of one slot out of MAX_BLOBS * MAX_SIZE.
+  if (_cur_blob >= MAX_BLOBS - 1 && _cur_off >= MAX_SIZE - 1) {
```
bryancall, line 82: The new guard covers `create()` only. `createSpan()` does `if (_cur_off + size > MAX_SIZE) addBlob();` and then `_cur_off += size`, so a span landing exactly on the boundary leaves `_cur_off == MAX_SIZE` without allocating a new blob. The next `create()` then indexes `names[1024]` on a `std::array<NameAndId, 1024>`. `end()` also becomes `_makeId(blob, MAX_SIZE)`, which `iterator::next()` can never reach because it wraps on `++offset == MAX_SIZE`.

Latent today, since the `Gauge::createSpan` and `Counter::createSpan` wrappers have no callers in the tree outside the unit tests. I raise it because it is the same class of bookkeeping off-by-one this PR is hardening against, and the new blob-growth-boundary test only drives `create()`. Either a matching guard or a comment noting `createSpan` is currently unused would close it.

doc/appendices/command-line/traffic_ctl.en.rst
```diff
@@ -949,13 +949,20 @@ traffic_ctl metric
...
 .. program:: traffic_ctl metric
-.. option:: match REGEX [REGEX...]
+.. option:: match [--include-hidden] REGEX [REGEX...]
 
    :ref:`admin_lookup_records`
 
    Display the current values of all statistics whose names match
    the given regular expression.
 
+.. option:: --include-hidden
```
bryancall, line 960: `--include-hidden` is a sibling `.. option::` under `.. program:: traffic_ctl metric`, so it renders as another subcommand alongside `match`, `describe` and `get` rather than as a flag on `match`. A reader cannot tell it only applies to `match`.

The synopsis on the line above also puts the flag before the regex, while `add_example_usage` in `traffic_ctl.cc` writes `traffic_ctl metric match METRIC [--include-hidden]`. Worth making the two spellings agree.

bryancall (dismissed):
I went through all 16 files. This is a clean change and the commit hygiene is genuinely good: one concern per commit, each message stating the defect and the reasoning, with the test order-independence fix landed first so the fragility never exists in the history.

A few things worth calling out as done right, because they were easy to get wrong:

- A separate `Storage` rather than a per-metric "hidden" flag. That makes hidden metrics structurally unreachable from the published store instead of relying on every consumer remembering to check a bit.
- `RECT_HIDDEN_METRIC = 0x40` sitting outside `RECT_ALL = 0x3F`, so no existing broad query changes behavior, and the JSONRPC decoder change stays on the read-only lookup path.
- Seeding the derived value from the first source instead of from zero. A zero seed is correct only for SUM and would clamp every MIN result to `<= 0`. The guard tests are built so an aliased-in `bad_id` zero would change the result, strictly positive sources for MIN and strictly negative for MAX, so they actually discriminate.
- The developer guide page documents the sampling semantics of a derived MAX, the 5 second recompute interval, and the hazard that ids are not portable across stores.

One general note that did not fit on a line. `test_RecHiddenMetricLookup.cc` re-implements the requested-type check locally rather than calling the real one, so nothing verifies end to end that a hidden metric survives the check in `get_record_regex_impl`. If that check changed from `(recType & rec_type) == 0` to an equality comparison, this test would still pass, and the gold test would not catch it either because a live traffic_server registers no hidden metrics for the query to return. The mirrored check is documented with a comment pointing at the original and the substance is asserted, so the risk is small. I raise it only because the PR body correctly identifies this interaction as the one thing invisible to every build-level check.

Everything I found is non-blocking. Four inline comments below.

src/tsutil/Metrics.cc
```diff
@@ -168,6 +187,17 @@ Metrics::Storage::createSpan(size_t size, Metrics::MetricType type, Metrics::IdT
   release_assert(size <= MAX_SIZE);
   ts::lock_guard lock(_mutex);
 
+  // On the final blob there is nowhere left to grow, so refuse a span that would fill or overflow
+  // it rather than letting addBlob() assert. Same intent as the guard in create(), and the same
+  // cost: some slots of the last blob go unused.
+  if (_cur_blob >= MAX_BLOBS - 1 && _cur_off + size >= MAX_SIZE) {
```
bryancall, line 193: This guard does not cover the case its comment describes, and the trailing `addBlob()` you added below makes a state that was previously survivable into a release abort.

`createSpan` can now advance the blob index twice in one call: once at line 201 because a span has to be contiguous, and once at line 220 when it ends exactly on the boundary. The guard only inspects `_cur_blob` as it is on entry, so it misses the case where the first advance is what puts you on the last blob.

Walking `_cur_blob == MAX_BLOBS - 2` (8190), `_cur_off == 500`, `size == MAX_SIZE`:

- line 193: `8190 >= 8191` is false, so no refusal
- line 201: `500 + 1024 > 1024`, so `addBlob()` runs, its `release_assert(8190 < 8191)` passes, leaving `_cur_blob == 8191` and `_cur_off == 0`
- line 214: `_cur_off` becomes exactly 1024
- line 220: `1024 >= 1024`, so `addBlob()` runs again and `release_assert(8191 < 8191)` fires

`release_assert` is defined outside the debug gate, so that aborts traffic_server in a release build. Sweeping `_cur_blob` in [8188, 8191] against every `_cur_off` and `size` gives exactly 1023 aborting triples, all at `_cur_blob == 8190`, `_cur_off` in [1, 1023], `size == MAX_SIZE`. The same sweep against the pre-commit body aborts zero times at 8190, so this state is newly fatal rather than a pre-existing hole.

Reachability is why this is not a blocker: it needs about 8.38M live published metrics, roughly half a gigabyte of blob storage plus an 8.4M-entry lookup map, and `createSpan` has no non-test callers. It is still worth closing, because the guard reads as complete and is not.

The obvious rewrite in terms of a target blob and offset is wrong, so I want to save you the detour. With `_cur_blob == 8191`, `_cur_off == 500`, `size == 600` it computes `target_off == 0`, fails the `>= MAX_SIZE` test, and falls through to an `addBlob()` that the current guard correctly refuses. Counting the advances instead subsumes the existing behavior:

```c++
// A span can advance the blob index twice: once because it must be contiguous, and once more
// when it ends exactly on the boundary. Refuse unless both advances are available.
const bool     needs_fresh_blob = (_cur_off + size > MAX_SIZE);
const size_t   end_off          = (needs_fresh_blob ? 0 : _cur_off) + size;
const unsigned needed           = (needs_fresh_blob ? 1u : 0u) + (end_off >= MAX_SIZE ? 1u : 0u);

if (_cur_blob + needed > MAX_BLOBS - 1) {
  if (id) {
    *id = 0; // Slot 0 is the reserved bad_id.
  }
  return {};
}
```

That turns the 1023 aborting triples into graceful bad_id refusals, never leaves `_cur_off == MAX_SIZE`, and matches the current guard on every case it already handles.

doc/appendices/command-line/traffic_ctl.en.rst
```diff
@@ -949,13 +949,21 @@ traffic_ctl metric
...
-.. option:: match REGEX [REGEX...]
+.. option:: match [--include-hidden] REGEX [REGEX...]
 
    :ref:`admin_lookup_records`
 
    Display the current values of all statistics whose names match
    the given regular expression.
 
+.. program:: traffic_ctl metric match
+.. option:: --include-hidden
```
bryancall, line 960: This fixes the cross-reference scope, which is real and worth having, but not the thing I was complaining about. `--include-hidden` is still a column-0 `.. option::` directive, so it still renders as a top-level sibling of `get`, `match`, `describe` and `monitor` rather than as a flag belonging to `match`.

The file already establishes the pattern 45 lines up: `ssl-multicert show` nests its `.. option:: --yaml, -y` inside the body of the subcommand option. Indenting this block three spaces so it sits inside `match` gets the rendering right, and makes both `.. program::` lines here unnecessary.

src/traffic_ctl/traffic_ctl.cc
```diff
@@ -249,7 +249,9 @@ main([[maybe_unused]] int argc, const char **argv)
     .add_example_usage("traffic_ctl metric get METRIC [METRIC ...]");
   metric_command.add_command("describe", "Show detailed information about one or more metric values", "", MORE_THAN_ONE_ARG_N,
                              Command_Execute); // not implemented
-  metric_command.add_command("match", "Get metrics matching a regular expression", "", MORE_THAN_ZERO_ARG_N, Command_Execute);
+  metric_command.add_command("match", "Get metrics matching a regular expression", "", MORE_THAN_ZERO_ARG_N, Command_Execute)
+    .add_option("--include-hidden", "", "Also match hidden (internal, normally unpublished) metrics")
+    .add_example_usage("traffic_ctl metric match [--include-hidden] METRIC [METRIC ...]");
```
bryancall, line 254: The flag position now agrees with the rst, which is what I asked for. The operand name does not: this says `METRIC` while `traffic_ctl.en.rst:952` says `REGEX`, and the subcommand's own description one line up is "Get metrics matching a regular expression".

`METRIC` is the misleading one here, since `metric get` genuinely takes literal names and this reads as if `match` does too. The sibling command at line 124 spells it `traffic_ctl config match [OPTIONS] REGEX [REGEX ...]`. Suggest `traffic_ctl metric match [--include-hidden] REGEX [REGEX ...]`.

src/records/RecCore.cc
```diff
@@ -612,6 +612,36 @@ RecLookupMatchingRecords(unsigned rec_type, const char *match, void (*callback)(
     });
   }
 
+  if (rec_type & RECT_HIDDEN_METRIC) {
+    // Opt-in only: hidden metrics are never reachable through RECT_ALL, see RecDefs.h.
+    auto &hidden = ts::Metrics::hidden_instance();
+    // Slot 0 of every Storage is the reserved bad_id placeholder, so it exists under the same name
+    // in both stores. Skip it here, otherwise a query matching it returns two identically named
+    // records that differ only in value.
+    auto it = hidden.begin();
```
bryancall, line 621: The skip is correct and solves the duplicate. Three follow-on notes, none blocking.

Skipping slot 0 removes the last way to see the hidden store's `bad_id` counter from the CLI. `RecLookupRecord` consults only `ts::Metrics::instance()`, never `hidden_instance()`, so `metric get` cannot reach it either. That counter accumulates every failed `createHiddenPtr`, and the new `Metrics.en.rst` tells operators that unbounded hidden metric names are the likely cause of exhaustion, so the diagnostic the docs point at is now unreadable. Emitting it under a disambiguated name, appending a `.hidden` suffix for the hidden store's slot 0, would resolve the duplicate confusion and keep the signal.

Moving from a range-for to an explicit loop also changed `end()` from evaluated once to evaluated per iteration, so this now picks up entries registered by another thread mid-walk where the published loop 30 lines up does not. I checked termination and it is fine, but the two loops now have different snapshot semantics for no stated reason. `auto const stop = hidden.end();` would keep them consistent.

There is also no test for the skip, and it is implemented as manual iterator arithmetic rather than a condition inside the loop, which is the shape a later refactor back to a range-for quietly undoes. A section in `test_RecHiddenMetricLookup.cc` calling `RecLookupMatchingRecords` with `RECT_PROCESS | RECT_HIDDEN_METRIC` and a `bad_id` pattern, asserting a count of 1, is about four lines and fails pre-fix with a count of 2.

src/tsutil/unit_tests/test_Metrics.cc
```diff
@@ -144,3 +177,466 @@ TEST_CASE("Metrics", "[libtsapi][Metrics]")
...
+
+  // Every pointer must be distinct: no two names should have been aliased onto the same atomic.
+  std::vector<Metrics::Counter::AtomicType *> sorted_ptrs = ptrs;
+  std::sort(sorted_ptrs.begin(), sorted_ptrs.end());
+  REQUIRE(std::adjacent_find(sorted_ptrs.begin(), sorted_ptrs.end()) == sorted_ptrs.end());
+}
+
+TEST_CASE("Metrics span lands exactly on a blob boundary", "[libtsapi][Metrics]")
+{
+  // A span has to be contiguous, so createSpan(MAX_SIZE) always starts a fresh blob and then fills
```
bryancall, line 613: Good test to add, and picking `MAX_SIZE` because it is the one size that reaches the boundary deterministically is the right instinct. Two notes.

The stated mechanism is not quite right. `createSpan(MAX_SIZE)` does not always start a fresh blob: when `_cur_off == 0` on entry, `_cur_off + size > MAX_SIZE` is `1024 > 1024`, which is false, so no `addBlob()` and the span is placed in the existing empty blob. The conclusion holds either way, so this is only comment accuracy, but the comment is the sole explanation of why `MAX_SIZE` is special and a reader who trusts it could conclude a smaller size behaves the same. The property that actually holds is that a span of `MAX_SIZE` always lands at offset 0 of an empty blob, growing one first if the current blob is partly used, and then fills it completely.

The test also does not fail deterministically against the pre-fix code. Without the trailing `addBlob()`, `_cur_off` is left at `MAX_SIZE` and the next `create()` writes `names[1024]`, one past the end of a `std::array<NameAndId, 1024>`. That is undefined behavior a sanitizer catches, but in a plain build it usually lands in adjacent storage and the REQUIREs still pass, so this stops guarding the fix in any non-sanitizer lane. Asserting that the iterator can actually reach `end()` after the span would pin the other half directly, since an offset parked at `MAX_SIZE` makes `end()` unreachable for `iterator::next()`.

bryancall (approved):
Re-reviewed the three new commits. All four of my comments were addressed, two of them beyond what I asked for. Approving again.

**C1, value-initializing the synthetic records: fixed, and over-delivered.** I asked for the hidden-metric site and suggested the published one. You found all five, including both `RecLookupRecord` sites and the `RECD_STRING` `StaticString` path, which is the one that actually reads `data_default.rec_string`. I checked that `{}` genuinely zeroes everything here: `RecRecord` is an aggregate with no user-provided constructors, `RecStatMeta` and `RecConfigMeta` are both plain types so the anonymous union zero-initializes wholesale, and `RecMutex` is itself an aggregate that gets zeroed before any member constructor runs. The synthetic records' mutex is never acquired on these paths, so the all-zero `pthread_mutex_t` is never used.

**C2, the bad_id duplicate: fixed correctly.** The unconditional `++it` is safe for a reason worth having written down, and your comment writes it down: `Storage`'s constructor unconditionally allocates blob 0 and asserts that `create("proxy.process.api.metrics.bad_id")` returns 0, so `begin() != end()` always holds and a hidden store containing only bad_id lands the single increment exactly on `end()`.

**C3, the createSpan boundary: the case I named is closed, but the new guard does not do what its comment says.** Details inline. Short version: `createSpan` can now advance the blob index twice in one call, and the guard only inspects `_cur_blob` as it is on entry.

**C4, the docs: half fixed.** The flag ordering now agrees between the rst synopsis and the help string, and the `.. program::` addition correctly rescopes the cross-reference target without breaking the following `describe` and `get` options. The rendering problem I was actually complaining about is still there, though. Inline.

I am approving rather than holding this, because the one real defect needs roughly 8.38 million live published metrics to reach and `createSpan` has no callers in the tree outside the unit tests. I want to be clear that I am approving on reachability, not because C3 and C4 are fully resolved.

One claim from my own earlier reasoning that I should retract: I suggested the fix might be to make the trailing `addBlob()` conditional. That would not reintroduce the out-of-bounds `names[1024]` write, since `create()`'s own guard already catches that state, but it would reintroduce the unreachable-`end()` half. So the warning against it stands, just for a different reason than I had in mind.

cmcfarlen:
Cherry-picked to 10.2.x

PR #13510, opened 2026-08-06
doc: rate_limit uses max_age, not max-age

moonchen:
The rate_limit YAML parser reads `max_age` with an underscore, in three places:

* the `queue` node — `plugins/experimental/rate_limit/limiter.h`
* the `ip-rep` node — `plugins/experimental/rate_limit/ip_reputation.cc`
* the `perma-block` sub-node — same file

Every mention in the documentation spelled it `max-age`, including all three
worked examples. A configuration copied from the docs therefore left `_max_age`
at its default of `0` in each case, which silently disables queue expiry (the
expiry pass in `sni_queue_cont()` is gated on
`max_age() > std::chrono::milliseconds::zero()`) and both IP-reputation aging
paths.

This changes the documentation rather than the parser, so that configurations
which are inert today stay inert on upgrade instead of suddenly beginning to
expire queue entries or age out LRU entries with whatever value the operator
wrote.

The `--maxage` pparam description also referred to a ``max-age`` that matches
neither the option name nor the YAML key, so it now just says "maximum age".

`max-age` was the only key in the documentation that did not match the parser;
`ip-rep` and `perma-block` really are hyphenated in the code.

### DIFF
--- a/doc/admin-guide/plugins/rate_limit.en.rst
+++ b/doc/admin-guide/plugins/rate_limit.en.rst
@@ -92,7 +92,7 @@ are available:
 
 .. option:: --maxage
 
-   An optional ``max-age`` for how long a transaction can sit in the delay queue.
+   An optional maximum age for how long a transaction can sit in the delay queue.
    The value (default 0) is the age in seconds.
 
 .. option:: --prefix
@@ -149,7 +149,7 @@ and nodes are documented below.
             rate: 200
             queue:
                size: 1000
-               max-age: 30
+               max_age: 30
             metrics:
                tag: example.com
                prefix: ddos
@@ -163,11 +163,11 @@ and nodes are documented below.
             buckets: 10
             size: 15
             percentage: 90
-            max-age: 300
+            max_age: 300
             perma-block:
                limit: 100
                threshold: 1
-               max-age: 1800
+               max_age: 1800
       lists:
          - name: internal
             cidr:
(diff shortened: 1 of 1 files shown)


JosiahWI (changes requested):
This is great. Could you add a mention of the historical reason for the naming inconsistency to the documentation, so that future developers understand that naming it `max-age` could activate previously inert configurations?

moonchen:
> This is great. Could you add a mention of the historical reason for the naming inconsistency to the documentation, so that future developers understand that naming it `max-age` could activate previously inert configurations?

Added a note to the documentation as a separate commit.

moonchen:
[approve ci rocky]

moonchen:
[approve ci rocky]

moonchen:
[approve ci format rat docs freebsd]

moonchen:
@JosiahWI Added — there is now a note under the `max_age` option:

> Note that this option was incorrectly documented as ``max-age`` with a hyphen in previous
> versions.  The code ignores ``max-age`` in all versions of this plugin.  We have chosen
> to keep the behavior the same and update the documentation, so that previously inert
> configurations don't activate unexpectedly with an upgrade.

JosiahWI (dismissed):
Thank you for working on this fix!

moonchen:
@JosiahWI Following up on your review — feedback from the community is that the documentation should not carry notes about how options used to be documented, so I have removed both back-references. The reference docs now describe just the current keys and units; the reasoning for fixing the docs rather than the parser stays in the PR description and commit trail.

I have also filed #13529 so the plugin rejects unrecognized YAML keys, which is the actual fix for this class of typo.

JosiahWI (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as ea106868d0bf4f6d31ca1c9eb7581c133dcdf7c9 for the 10.2.1 release.

PR #13515, opened 2026-08-07
Fix cache read VC replacement after a lost write lock

bneradt:
A transaction that revalidates a stale cached object and cannot take the
cache write lock is sent back through a second cache lookup while it
still holds the cache read connection its first lookup opened. The read
that completes for that second lookup replaces the connection the
transaction is using: debug builds abort on the read connection
assertion in HttpCacheSM::state_cache_open_read(), and release builds
close that connection out from under the stale object saved as the retry
fallback, leaving the fallback pointing into freed memory. The re-lookup
runs for every cache_open_write_fail_action rather than only for the two
that configure a read retry, so fail action 2, which is documented to
serve the stale object instead of retrying anything, aborts a debug
build several times a day under production traffic.

This patch limits the re-lookup to the fail actions that configure a
read retry. A transaction that loses the write lock with a cached object
and no retry configured now hands that object straight to the freshness
handling that serves stale content, with no second lookup. The retry
actions do want that lookup, so this also makes replacing the read
connection explicit and drops the saved stale object along with the
connection that owns it, since neither can outlive the other. This adds
an autest covering both configurations that does not depend on
contention between transactions: denying the write lock through
max_open_write_retries makes the failure synchronous, and each
configuration aborts an unpatched debug build on the production
assertion.

The re-lookup arrived with the fail action 6 work in #12852, which
applied it to every non-default fail action; that commit's own test
notes the stale path is timing sensitive and does not exercise it. The
resulting aborts resemble the ones #13487 fixed, because both land in
HttpCacheSM while a cache write retry dispatches events, but they are a
distinct failure. #13487 stopped HttpSM from canceling its own captive
action, which aborts on the cancellation assertion in
HttpCacheSM.cc:138; this is the read connection assertion ten lines
later, reached with that action perfectly valid. Both fixes are needed,
and neither subsumes the other.

### DIFF
--- a/include/proxy/http/HttpConfig.h
+++ b/include/proxy/http/HttpConfig.h
@@ -420,6 +420,18 @@ enum class CacheOpenWriteFailAction_t {
   TOTAL_TYPES
 };
 
+/** Whether a cache_open_write_fail_action retries the cache read.
+ *
+ * @param[in] action A proxy.config.http.cache.open_write_fail_action value.
+ * @return Whether losing the cache write lock should retry the cache read.
+ */
+inline bool
+is_read_retry_write_fail_action(MgmtByte action)
+{
+  return action == static_cast<MgmtByte>(CacheOpenWriteFailAction_t::READ_RETRY) ||
+         action == static_cast<MgmtByte>(CacheOpenWriteFailAction_t::READ_RETRY_STALE_ON_REVALIDATE);
+}
+
 extern HttpStatsBlock http_rsb;
 
 /////////////////////////////////////////////////////////////
--- a/src/proxy/http/HttpCacheSM.cc
+++ b/src/proxy/http/HttpCacheSM.cc
@@ -43,14 +43,6 @@
 namespace
 {
 DbgCtl dbg_ctl_http_cache{"http_cache"};
-
-// Helper to check if cache_open_write_fail_action has READ_RETRY behavior
-inline bool
-is_read_retry_action(MgmtByte action)
-{
-  return action == static_cast<MgmtByte>(CacheOpenWriteFailAction_t::READ_RETRY) ||
-         action == static_cast<MgmtByte>(CacheOpenWriteFailAction_t::READ_RETRY_STALE_ON_REVALIDATE);
-}
 } // end anonymous namespace
 
 ////
@@ -145,9 +137,13 @@ HttpCacheSM::state_cache_open_read(int event, void *data)
   switch (event) {
   case CACHE_EVENT_OPEN_READ:
     Metrics::Gauge::increment(http_rsb.current_cache_connections);
-    ink_assert((cache_read_vc == nullptr) || master_sm->t_state.redirect_info.redirect_in_process);
+    ink_assert((cache_read_vc == nullptr) || master_sm->t_state.redirect_info.redirect_in_process ||
+               master_sm->t_state.cache_info.write_lock_state == HttpTransact::CacheWriteLock_t::READ_RETRY);
     if (cache_read_vc) {
-      // redirect follow in progress, close the previous cache_read_vc
+      // A redirect follow or a read retry after losing the cache write lock
+      // replaces the read VC. The stale object that a read retry saved as its
+      // fallback lives in the VC being closed, so it cannot outlive it.
+      master_sm->t_state.cache_info.stale_fallback = nullptr;
       close_read();
     }
     cache_read_vc = static_cast<CacheVConnection *>(data);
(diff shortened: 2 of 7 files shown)


cmcfarlen:
I traced this through and the diagnosis holds up. Recording the path since it spans three files and is not obvious from any one of them.

`HttpSM::state_cache_open_write()` handles `CACHE_EVENT_OPEN_WRITE_FAILED` by bailing out to `FAIL` only for `DEFAULT`, for `ERROR_ON_MISS_OR_REVALIDATE`, or when there is no `object_read`. Everything else takes the `INTENTIONAL FALL THROUGH` into `case CACHE_EVENT_OPEN_READ`, and with `object_read != nullptr` that tail unconditionally ends in:

```cpp
t_state.cache_info.write_lock_state = HttpTransact::CacheWriteLock_t::READ_RETRY;
```

So fail actions 2 and 3 arrive at the `CacheWriteLock_t::READ_RETRY` branch of `handle_cache_write_lock()` despite configuring no retry, and the stale sub-branch there saves `stale_fallback` and issues `TRANSACT_RETURN(CACHE_LOOKUP)` while the first lookup's read VC is still held. The `CACHE_EVENT_OPEN_READ` for that second lookup then trips `ink_assert((cache_read_vc == nullptr) || ...)`, and in release builds `close_read()` frees the VC that owns the object `stale_fallback` points at. Fail action 2 is documented as "serve stale ... otherwise go to origin" with no retry in it at all, so it should never have been on that path.

Scoping the branch with `is_read_retry_write_fail_action()` is the right fix rather than widening the assertion alone, and routing 2/3 into `HandleCacheOpenReadHitFreshness()` matches what those actions are documented to do.

**The deferred-hook question resolves in your favor, but only by a non-local invariant.** My first concern was that the new early return calls `HandleCacheOpenReadHitFreshness(s)` with no `cache_lookup_complete_deferred` handling, unlike the fresh path just below it, which fires `API_CACHE_LOOKUP_COMPLETE` when deferred. That turns out to be safe: `cache_lookup_complete_deferred` is only ever set for actions 5 and 6 (`HttpSM.cc:2738-2742`), so it is guaranteed false on a path guarded by `!is_read_retry_write_fail_action(...)`. The two predicates are exact complements.

Two things follow from that, and they are my only real asks:

1. **Assert it.** The guarantee lives in `HttpSM.cc` while the new guard lives in `HttpTransact.cc`; nothing connects them. An `ink_assert(!s->cache_lookup_complete_deferred)` in the new branch would pin the invariant where a future change to either side would notice.
2. **The two predicates read different variables.** `HttpSM` decides deferral from `t_state.txn_conf->cache_open_write_fail_action`, and the existing `is_read_retry_write_fail_action()` call sites in `HttpCacheSM.cc` also use `txn_conf`, but the new call in `HttpTransact.cc` uses the state copy `s->cache_open_write_fail_action`. They are equal wherever this matters, and the one place they diverge is the redirect case, which forces `DEFAULT` and `write_lock_state = FAIL` and so never reaches here. Still, "same predicate, different source of truth" is a trap; a comment or picking one consistently would help.

**On clearing `stale_fallback` before `close_read()`** — I convinced myself this does not cost action 6 its fallback, and it may be worth saying so in the description since it reads like a behavior loss. The clear only happens when a replacement read VC actually arrives, and in that case `object_read` is re-derived from the new VC and freshness re-evaluated, so the saved fallback is genuinely redundant. The case where action 6 needs the fallback is the second lookup *missing*, which yields `CACHE_EVENT_OPEN_READ_FAILED`, never reaches `state_cache_open_read()`, and so leaves `stale_fallback` intact for `HandleCacheOpenReadMiss()`. Trading a use-after-free for nothing at all is the good outcome, but it took a while to be sure of.

Hoisting `is_read_retry_action` out of the anonymous namespace into `HttpConfig.h` as `is_read_retry_write_fail_action` is fine now that two translation units need it, and the rename is clearer.

**This wants backporting to 10.2.x, and I would treat it as RC-blocking.** #12852 is already on 10.2.x as `49ab050e0e`, and the vulnerable block is present there, so the regression ships in 10.2.0 as it stands — with the same "aborts a debug build several times a day under production traffic" exposure you describe. The good news is that it should pick cleanly: the new autest uses only `Test.ATSReplayTest`, and the `autest:` keys in both replay files (`process_config`, `copy_to_config_dir`, `records_config`, `remap_config`, `log_validation`) are all supported by 10.2.x's `ats_replay.test.ext`.

Agreed too that this is distinct from #13487 and neither subsumes the other: that one stopped `HttpSM` from canceling its own captive action via `pending_action = adjust_thread(...)`, which fails the cancellation assertion; this one reaches the read-VC assertion with the action perfectly valid.

Nice touch making the test synchronous via `max_open_write_retries` instead of relying on contention between transactions — that is exactly the weakness in #13487's autest, which needs a `delay: 3s` / `delay: 200ms` overlap to provoke the race.

cmcfarlen (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as 601871fa9576332f309caaf4ac64f3a506c3702d for the 10.2.0 release.

PR #13531, opened 2026-08-11
Change libswoc test for undefined errno to a much higher number

jeredfloyd:
libswoc tests include one that expects errno 134 to be unknown; this is no longer the case in Linux 7.2.  This patch increases the errno to one that should be unknown for the forseeable future.


JosiahWI (approved):

bneradt (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as 5e81c3a116904dabcb5b345c77c78319ebb7f32d for the 10.2.0 release.

PR #13532, opened 2026-08-11
Fix signature on flz_maxcopy to avoid cast error

jeredfloyd:
flz_maxcopy takes void* arguments but calls fastlz_memcopy (taking uint8_t* arguments) without a cast.  I don't see a reason for flz_maxcopy to not take uint8_t* arguments; for arch-specific speedups these get cast to the appropriately-sized pointer.  Resolves build issues on ppc64le architecture.

### DIFF
--- a/lib/fastlz/fastlz.cc
+++ b/lib/fastlz/fastlz.cc
@@ -162,7 +162,7 @@ static void flz_smallcopy(uint8_t* dest, const uint8_t* src, uint32_t count) {
 }
 
 /* special case of memcpy: exactly MAX_COPY bytes */
-static void flz_maxcopy(void* dest, const void* src) {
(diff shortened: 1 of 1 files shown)


bneradt (approved):
Seems reasonable. Maybe submit a PR upstream?

https://github.com/ariya/FastLZ/

The lib/ directory contains third party code. When we update to a newer version, we will likely lose your change here unless this is changed in upstream.

jeredfloyd:
Done: https://github.com/ariya/FastLZ/pull/11

cmcfarlen:
Cherry-picked to the 10.2.x branch as a69c89869d1b7f1b4592b42d0c1d9ef7ce06630d for the 10.2.0 release.

PR #13538, opened 2026-08-12
Fix records_yaml startup lookup race

bneradt:
The records_yaml AuTest intermittently reported that
proxy.config.diags.output.note was missing when a configuration
callback ran while startup was still registering metrics.
Metrics::find() returned the then-current end iterator for a miss, but
RecLookupRecord() compared it with a new end iterator. An intervening
registration made them differ and misread the missing string record as
a numeric metric.

This patch addresses the race by using the direct metrics lookup API,
which reports a miss without comparing two moving end positions. It also
adds a concurrent registration test that exercises string record lookups
during metric creation.

### DIFF
--- a/src/records/unit_tests/test_RecRegister.cc
+++ b/src/records/unit_tests/test_RecRegister.cc
@@ -17,11 +17,16 @@
    or implied. See the License for the specific language governing permissions and limitations under
    the License.
  */
+#include <atomic>
+#include <string>
+#include <thread>
+
 #include <catch2/catch_test_macros.hpp>
 #include "records/RecCore.h"
 #include "iocore/eventsystem/EventSystem.h"
 #include "iocore/eventsystem/RecProcess.h"
 #include "tscore/Layout.h"
+#include "tsutil/Metrics.h"
 #include "test_Diags.h"
 
 TEST_CASE("RecRegisterConfig - Type Dispatch", "[librecords][RecConfig]")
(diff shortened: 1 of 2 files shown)


src/records/unit_tests/test_RecRegister.cc
```diff
@@ -17,11 +17,16 @@
    or implied. See the License for the specific language governing permissions and limitations under
    the License.
  */
+#include <atomic>
+#include <string>
+#include <thread>
```
JosiahWI: Nit: it's better to include standard headers after local project headers to expose missing includes quickly.

src/records/unit_tests/test_RecRegister.cc
```diff
@@ -87,3 +92,43 @@ TEST_CASE("RecRegisterStat - Type Dispatch", "[librecords][RecStat]")
...
+  size_t lookup_count          = 0;
+
+  start.store(true, std::memory_order_release);
+  while (!finished.load(std::memory_order_acquire)) {
+    if (RecGetRecordStringAlloc(record_name) != record_value) {
+      all_lookups_succeeded = false;
+      break;
+    }
+    ++lookup_count;
+  }
```
JosiahWI: This loop can run 0 times in theory, which will fail the test. What was the intent here?

JosiahWI (changes requested):
The fix is good. I have a request for clarification on the concurrency test.

src/records/unit_tests/test_RecRegister.cc
bneradt: Done. Project headers now precede the standard-library headers.

src/records/unit_tests/test_RecRegister.cc
bneradt: Changed to a do/while loop so at least one lookup always occurs, and switched to unnamed Counter::createSpan(1) allocations to avoid the start spin and retaining 100,000 metric names.

JosiahWI (approved):
Thanks!

cmcfarlen:
Cherry-picked to the 10.2.x branch as 43e3d233d8c412ffb48c87a2041e97aefae749b6 for the 10.2.1 release.

PR #13533, opened 2026-08-11
Remove dead config records from NetHandler

brbzull0:
Three proxy.config.net.* records have been unregistered since their introduction in 2015 , so their callbacks never fired and the struct fields they wrote were never read. Removing them shifts default_inactivity_timeout to index 2; the bitset selecting per-thread dependent values is unaffected since those members keep indices 0 and 1. Assertions now pin the field offsets and the bitset width, both of which were previously implicit.

Fixes: #12933

PR #13543, opened 2026-08-13
Index NetHandler config by named enum

brbzull0:
follow up for https://github.com/apache/trafficserver/pull/13533.

Config was addressed as an array by advancing a pointer from its first member, and the update handler recovered the index by subtracting pointers to distinct members. Both are undefined behavior regardless of layout. Name the values instead so indexing is well defined, which also removes the layout assertions and the magic per-thread mask.

The index also arrives from an untyped event cookie, so give the enum a fixed underlying type: converting an out of range integer to an enumeration without one is undefined, which would defeat the check in operator[] before it could run. Scoping the enum keeps an int from silently becoming an index again.

The compiler checks the new switch for exhaustiveness but not for correctness, so add a test that every index reaches a distinct member. A case returning the wrong value would otherwise build cleanly and make a record update write to the wrong setting.

### DIFF
--- a/include/iocore/net/NetHandler.h
+++ b/include/iocore/net/NetHandler.h
@@ -24,8 +24,11 @@
 #pragma once
 
 #include <atomic>
-#include <cstddef>
-#include <type_traits>
+#include <bitset>
+#include <cstdint>
+#include <limits>
+
+#include "tscore/ink_assert.h"
 
 #include "iocore/eventsystem/Continuation.h"
 #include "iocore/eventsystem/EThread.h"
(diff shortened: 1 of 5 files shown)


src/iocore/net/UnixNet.cc
```diff
@@ -41,20 +41,12 @@ std::atomic<bool> net_memory_throttle = false;
...
-static_assert(PER_THREAD_DEPENDENT_CONFIG < (1ULL << NetHandler::CONFIG_ITEM_COUNT));
-} // end anonymous namespace
-
 NetHandler::Config                                     NetHandler::global_config;
 std::bitset<std::numeric_limits<unsigned int>::digits> NetHandler::active_thread_types;
-const std::bitset<NetHandler::CONFIG_ITEM_COUNT> NetHandler::config_value_affects_per_thread_value{PER_THREAD_DEPENDENT_CONFIG};
+/// The values @c NetHandler::configure_per_thread_values reads.
+const std::bitset<NetHandler::CONFIG_ITEM_COUNT> NetHandler::config_value_affects_per_thread_value{
+  (1U << static_cast<unsigned>(NetHandler::Config::Index::MAX_CONNECTIONS_IN)) |
+  (1U << static_cast<unsigned>(NetHandler::Config::Index::MAX_REQUESTS_IN))};
```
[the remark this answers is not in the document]
brbzull0: Took the 1ULL half . std::bitset's constructor takes unsigned long long anyway, so that's the right type to build the mask in.

I left out the static_assert. The case it guards is std::bitset silently dropping bits above its width, but that can't happen here anymore: the shift amounts are Config::Index enumerators, and every enumerator is by construction less than COUNT, which is exactly the width of the bitset (CONFIG_ITEM_COUNT). So the set bits are always in range.

That guard did earn its place in #13533, where the mask was a magic 0x3 with no connection to the struct it described and nothing tying the two together. Deriving the mask from named indices is what makes it redundant, so keeping it would preserve scaffolding for a problem the change removes. Happy to add it back if you'd rather have the belt and braces.

include/iocore/net/NetHandler.h
```diff
@@ -113,30 +116,38 @@ class NetHandler : public Continuation, public EThread::LoopTailHandler
...
   struct Config {
+    /// Identifies a config value, so an update can name it instead of passing a pointer.
+    /// @note The underlying type is fixed and must not be narrowed. The index arrives as
+    /// an untyped event cookie, and a value that wrapped would look like a valid index.
+    enum class Index : int {
+      MAX_CONNECTIONS_IN,
+      MAX_REQUESTS_IN,
+      DEFAULT_INACTIVITY_TIMEOUT,
+      COUNT ///< Number of config values, not a valid index.
+    };
```
JosiahWI, line 127: The bitset initialization would be simpler if you set the enumeration values to powers of two. What you did is good - just throwing the idea out there.

src/iocore/net/unit_tests/test_NetHandler.cc
```diff
@@ -0,0 +1,60 @@
...
+  limitations under the License.
+ */
+
+#include <catch2/catch_test_macros.hpp>
+
+#include <cstddef>
+#include <cstdint>
+#include <set>
+
+#include "iocore/net/NetHandler.h"
```
JosiahWI, line 30: Nit: order should be local headers, then third party headers, then standard headers, to expose missing includes early.

JosiahWI (approved):
Great fix. Thanks for working on this!

include/iocore/net/NetHandler.h
brbzull0: i'll be doing some more work here in the short term, I may re-think this. thanks mate.

PR #13546, opened 2026-08-13
Honor valueless max-stale requests

bneradt:
A valueless Cache-Control: max-stale directive permits clients to
accept responses stale by any amount. Header cooking instead discarded
the directive, causing ATS to revalidate an otherwise acceptable cached
response.

Commit d3196a87a28a64bfd26a0a721ccc24305fe8cc0b ("Ignore malformed
Cache-Control directives per RFC 7234", #12670) introduced the
regression by treating every numeric directive without an equals sign
as malformed.

This patch represents valueless max-stale with the existing unlimited
sentinel and adds parser and end-to-end regression coverage.


bneradt:
Backport note: this change is only needed on master. The 10.x branches already handle valueless max-stale correctly, so no 10.x backport is necessary.

moonchen (approved):

PR #13547, opened 2026-08-13
Add percentage limit for stale cache age

### LINKED ISSUE #12252: Feature request: Add % max_stale_age config
Our proxy.config.http.cache.max_stale_age is currently a number of seconds. The shorter the ttl, the larger the percentage of potential stale time. For instance, the default 7 day max_stale_age would be 700% of a 1 day ttl; and obviously grows the shorter the ttl. Conversely, the default is only 25% of a 1 month ttl and 2% of a 1 year percent. It's much safer to allow an object with 1yr ttl to be served an extra 2% than to allow an 1hr ttl object to be served for so long.

We should really have a percentage setting instead-- for instance setting stale to be 25% would allow a 1hr ttl object to be served stale for an extra 15 min, whereas a 1 month ttl object would be allowed to be served for an extra week.

bneradt:
A fixed stale-age window can let short-lived cached responses remain
usable for many times their original freshness lifetime. This makes
serving stale content disproportionately risky for responses with a
short max-age.

This adds an optional percentage limit and applies the smaller of the
percentage-based and absolute stale-age windows. The default keeps the
existing behavior, while transaction overrides and supported scripting
interfaces can opt in per use case.

Fixes: #12252


src/proxy/http/HttpTransact.cc
```diff
@@ -6396,8 +6396,20 @@ HttpTransact::is_stale_cache_response_returnable(State *s)
...
                                                                  s->cache_info.object_read->response_received_time_get(),
                                                                  cached_response, cached_response->get_date(), s->current.now);
+
+  MgmtInt max_age       = get_max_age(cached_response);
+  MgmtInt max_stale_age = s->txn_conf->cache_max_stale_age;
+  MgmtInt max_stale_percentage =
+    std::clamp(s->txn_conf->cache_max_stale_age_percent, static_cast<MgmtInt>(0), static_cast<MgmtInt>(100));
+
+  if (max_stale_percentage > 0 && max_age >= 0) {
+    MgmtInt percent_max_stale_age = max_age * max_stale_percentage / 100;
```
serrislew, line 6406: The integer division in `max_age * max_stale_percentage / 100` truncates, so the window is 0 for any max-age < 100/percent (ex: max_stale_age_percent: 1 zeroes the stale window for everything under max-age=100 taking those objects from servable for the full 7d max_stale_age to unservable after a second) Intended? If so, maybe add a doc note?

bneradt: Yes, this is intentional. ATS measures age in whole seconds, and because the percentage is an upper bound, rounding up could exceed the configured limit. I documented that calculations round down and that sub-one-second results permit no stale window, including the 1% with max-age=99 example.

serrislew (approved):

PR #13552, opened 2026-08-14
Fix memory leaks reported by Coverity Scan

bryancall:
Part 3 of 3 splitting a Coverity Scan cleanup into independently reviewable pieces. Each leaked allocation is unreachable after the leak, so releasing it changes no observable behavior.

### Repeatable command line options

`regex_revalidate`, `remap_purge`, `xdebug`, `stale_response` and the `uri_signing` issuer id all overwrote a previously duplicated string when an option was given twice. Every one of these fields is null-initialized first (`memset`, an init helper, or an explicit `= nullptr`), so the first pass frees nothing and no string literal ever reaches `free`. `stale_response` compares against `PLUGIN_TAG` by pointer identity, which is the idiom its own destructor already uses.

### Ownership the callers were not honoring

- `TSMgmtStringGet` hands back a copy the caller owns (`ats_strdup` internally). `maxmind_acl` and an API regression test dropped it. Both are safe on the failure path too, since the function does not write `*result` on failure and both callers pre-initialize to null.
- `jax_fingerprint` leaked its configuration on three plugin initialization failure paths. Note the `reserve_user_arg` failure path is deliberately **not** given a `delete`, because the config is captured by a `TSLogFieldRegister` lambda above it.
- The YAML remap parser duplicated a redirect URL that nothing owned. `parse_format_redirect_url` copies what it needs out of the buffer, so the local string's storage can be handed over directly. It does write into that buffer transiently (nul-terminating chunks in place and restoring them), so the buffer must be mutable and outlive the call, which it is and does.
- `traffic_cache_tool` never released its URL set or stripe hash table.

### The one hunk where new code runs

`~Cache()` in `CacheTool.cc` is the only place here that adds executing code rather than deleting or substituting. `Cache` holds `std::list<std::unique_ptr<Span>>`, so copy and move are implicitly deleted and the new destructor cannot double free; the `URLset` entries are `new`ed in exactly one place and deleted nowhere else. The `= nullptr` initializer on `stripes_hash_table` is load bearing, since most instances never build the table.

Two follow-ups I did not fold in, to keep this reviewable: `build_stripe_hash_table()` overwrites `stripes_hash_table` without freeing a previous table (harmless today, one call per instance, but now that the field is owning it is worth guarding), and `regex_revalidate`'s `-l` is a fourth repeatable option that still leaks a `TSTextLogObject`.

### Verification

Clean build with no new warnings and the full unit test suite passing (137/137) on Fedora, GCC 16.1.1.

Two of the touched plugins are not built by default, so verifying them needed extra options: `uri_signing` requires cjose, and `jax_fingerprint` defaults to off and needs `-DENABLE_JAX_FINGERPRINT=ON`. Both are compiled in the run above.

Draft while CI runs.

### Update

Pushed a follow-up commit that replaces the explicit deletes here with a `unique_ptr`, since a reviewer pointed out the deletes were doing by hand what ownership should do on its own.

That turned up a leak the explicit-delete version missed. `TSPluginInit`'s user-argument reservation failure path returned without freeing, which is only correct when the log field callback has captured the configuration, and that capture is conditional on a log symbol being configured. With no `--log-field`, nothing owned it and it leaked. Reserving the index before registering the log field puts every failure exit inside the span the `unique_ptr` still owns, so no path has to reason about who else holds a reference. Note that reordering is a behavior change, small but real: a failed reservation no longer leaves a registered log field behind.

### DIFF
--- a/plugins/experimental/jax_fingerprint/plugin.cc
+++ b/plugins/experimental/jax_fingerprint/plugin.cc
@@ -451,23 +451,23 @@ TSReturnCode
 TSRemapNewInstance(int argc, char *argv[], void **ih, char * /* errbuf ATS_UNUSED */, int /* errbuf_size ATS_UNUSED */)
 {
   Dbg(dbg_ctl, "New instance for client matching %s to %s", argv[0], argv[1]);
-  auto config         = std::make_unique<PluginConfig>();
-  config->plugin_type = PluginType::REMAP;
+  auto owned_config         = std::make_unique<PluginConfig>();
+  owned_config->plugin_type = PluginType::REMAP;
 
   // Parse parameters
-  if (!read_config_option(argc - 1, const_cast<const char **>(argv + 1), *config)) {
+  if (!read_config_option(argc - 1, const_cast<const char **>(argv + 1), *owned_config)) {
     Dbg(dbg_ctl, "Bad arguments");
     return TS_ERROR;
   }
 
-  if (!config->log_symbol.empty()) {
+  if (!owned_config->log_symbol.empty()) {
     TSError("[%s] --log-field is not supported in remap.config. Use it in plugin.config instead.", PLUGIN_NAME);
     return TS_ERROR;
   }
 
   // Create a log file
-  if (!config->log_filename.empty()) {
-    if (!create_log_file(config->log_filename, config->log_handle)) {
+  if (!owned_config->log_filename.empty()) {
+    if (!create_log_file(owned_config->log_filename, owned_config->log_handle)) {
       TSError("[%s] Failed to create log.", PLUGIN_NAME);
       return TS_ERROR;
     } else {
@@ -475,30 +475,30 @@ TSRemapNewInstance(int argc, char *argv[], void **ih, char * /* errbuf ATS_UNUSE
     }
   }
 
-  if (reserve_user_arg(*config) == TS_ERROR) {
+  if (reserve_user_arg(*owned_config) == TS_ERROR) {
     TSError("[%s] Failed to reserve user arg index.", PLUGIN_NAME);
     return TS_ERROR;
   }
 
   // Past here the instance handle owns the configuration and TSRemapDeleteInstance releases it.
-  PluginConfig *instance = config.release();
+  PluginConfig *config = owned_config.release();
 
   // Create continuation
-  if (instance->standalone) {
+  if (config->standalone) {
     Dbg(dbg_ctl, "Standalone mode. Adding hooks.");
-    instance->handler = TSContCreate(main_handler, nullptr);
-    if (instance->method.on_client_hello) {
-      TSHttpHookAdd(TS_SSL_CLIENT_HELLO_HOOK, instance->handler);
+    config->handler = TSContCreate(main_handler, nullptr);
+    if (config->method.on_client_hello) {
+      TSHttpHookAdd(TS_SSL_CLIENT_HELLO_HOOK, config->handler);
     }
-    if (instance->method.type == Method::Type::CONNECTION_BASED) {
-      TSHttpHookAdd(TS_VCONN_CLOSE_HOOK, instance->handler);
+    if (config->method.type == Method::Type::CONNECTION_BASED) {
+      TSHttpHookAdd(TS_VCONN_CLOSE_HOOK, config->handler);
     } else {
-      TSHttpHookAdd(TS_HTTP_TXN_CLOSE_HOOK, instance->handler);
+      TSHttpHookAdd(TS_HTTP_TXN_CLOSE_HOOK, config->handler);
     }
-    TSContDataSet(instance->handler, instance);
+    TSContDataSet(config->handler, config);
   }
 
-  *ih = static_cast<void *>(instance);
+  *ih = static_cast<void *>(config);
 
   return TS_SUCCESS;
 }
--- a/plugins/experimental/jax_fingerprint/plugin.cc
+++ b/plugins/experimental/jax_fingerprint/plugin.cc
@@ -378,12 +378,14 @@ TSPluginInit(int argc, char const **argv)
 
   if (!read_config_option(argc, argv, *config)) {
     TSError("[%s] Failed to parse options.", PLUGIN_NAME);
+    delete config;
     return;
   }
 
   if (!config->log_filename.empty()) {
     if (!create_log_file(config->log_filename, config->log_handle)) {
       TSError("[%s] Failed to create log.", PLUGIN_NAME);
+      delete config;
       return;
     } else {
       Dbg(dbg_ctl, "Created log file.");
@@ -465,6 +467,7 @@ TSRemapNewInstance(int argc, char *argv[], void **ih, char * /* errbuf ATS_UNUSE
   if (!config->log_filename.empty()) {
     if (!create_log_file(config->log_filename, config->log_handle)) {
       TSError("[%s] Failed to create log.", PLUGIN_NAME);
+      delete config;
       return TS_ERROR;
     } else {
       Dbg(dbg_ctl, "Created log file.");
@@ -473,6 +476,7 @@ TSRemapNewInstance(int argc, char *argv[], void **ih, char * /* errbuf ATS_UNUSE
 
   if (reserve_user_arg(*config) == TS_ERROR) {
     TSError("[%s] Failed to reserve user arg index.", PLUGIN_NAME);
+    delete config;
     return TS_ERROR;
   }
 
--- a/plugins/experimental/jax_fingerprint/plugin.cc
+++ b/plugins/experimental/jax_fingerprint/plugin.cc
@@ -373,24 +373,33 @@ TSPluginInit(int argc, char const **argv)
     return;
   }
 
-  PluginConfig *config = new PluginConfig();
-  config->plugin_type  = PluginType::GLOBAL;
+  auto owned_config         = std::make_unique<PluginConfig>();
+  owned_config->plugin_type = PluginType::GLOBAL;
 
-  if (!read_config_option(argc, argv, *config)) {
+  if (!read_config_option(argc, argv, *owned_config)) {
     TSError("[%s] Failed to parse options.", PLUGIN_NAME);
-    delete config;
     return;
   }
 
-  if (!config->log_filename.empty()) {
-    if (!create_log_file(config->log_filename, config->log_handle)) {
+  if (!owned_config->log_filename.empty()) {
+    if (!create_log_file(owned_config->log_filename, owned_config->log_handle)) {
       TSError("[%s] Failed to create log.", PLUGIN_NAME);
-      delete config;
       return;
     } else {
       Dbg(dbg_ctl, "Created log file.");
     }
   }
+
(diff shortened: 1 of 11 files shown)


src/proxy/http/remap/RemapYamlConfig.cc
```diff
@@ -388,7 +388,9 @@ parse_map_referer(const YAML::Node &node, url_mapping *url_mapping)
       !strcasecmp(url.c_str(), "<default_redirect_url>") || !strcasecmp(url.c_str(), "default_redirect_url")) {
     url_mapping->default_redirect_url = true;
   }
-  url_mapping->redir_chunk_list = redirect_tag_str::parse_format_redirect_url(ats_strdup(url.c_str()));
+  // parse_format_redirect_url() copies what it needs out of the buffer, so hand it the local
+  // string's storage rather than a fresh allocation that nothing would own.
+  url_mapping->redir_chunk_list = redirect_tag_str::parse_format_redirect_url(url.data());
```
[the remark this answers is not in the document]
bryancall: You are right, and this is a regression from the change rather than something pre-existing. Fixed in bc46c50.

The original code here passed `ats_strdup(url.c_str())`, so the temporary nul landed inside a malloc allocation of `strlen + 1` and was fine; it just leaked, which is what I was trying to fix. Swapping in `url.data()` removed the leak and moved that write onto `std::string`, where the reference at index `size()` may not be assigned through.

Worth noting the trigger is the common case, not an edge case: the inner scan only stops early on `%r`, `%f`, `%t` or `%o`, so any redirect URL with no format specifier runs to the terminating nul and takes that path.

It now hands the parser an `ats_scoped_str` it owns and releases once the call returns. The chunk list keeps its own copies via `ats_strdup`, so nothing outlives the buffer, and the leak stays fixed. Clean build with no new warnings and 137/137 unit tests on Fedora, GCC 16.1.1.

plugins/experimental/jax_fingerprint/plugin.cc
```diff
@@ -445,19 +451,17 @@ TSReturnCode
 TSRemapNewInstance(int argc, char *argv[], void **ih, char * /* errbuf ATS_UNUSED */, int /* errbuf_size ATS_UNUSED */)
 {
   Dbg(dbg_ctl, "New instance for client matching %s to %s", argv[0], argv[1]);
-  auto config         = new PluginConfig();
+  auto config         = std::make_unique<PluginConfig>();
```
JosiahWI: The naming here and below is inconsistent with the naming change earlier in the file. Please use consistent naming.

plugins/experimental/stale_response/stale_response.cc
```diff
@@ -1070,6 +1070,10 @@ parse_args(int argc, char const *argv[])
       plugin_config->log_info.stale_if_error = true;
       break;
     case 'd':
+      // The option may be repeated; release the previously duplicated name first.
+      if (plugin_config->log_info.filename != PLUGIN_TAG) {
+        free(const_cast<char *>(plugin_config->log_info.filename));
```
JosiahWI: Can we store the filename as non-const instead?

plugins/xdebug/xdebug.cc
```diff
@@ -947,6 +947,7 @@ TSPluginInit(int argc, const char *argv[])
     switch (opt) {
     case 'h':
       Dbg(dbg_ctl, "Setting header: %s", optarg);
+      TSfree(const_cast<char *>(xDebugHeader.str)); // The option can be repeated, so the earlier value is not leaked
```
JosiahWI: Can we store the `str` field as non-const instead?

JosiahWI (changes requested):

plugins/experimental/jax_fingerprint/plugin.cc
bryancall: Fixed in 34323cc. The two functions were using `config` for different things, the `unique_ptr` in `TSRemapNewInstance` and the released raw pointer in `TSPluginInit`. Both now spell the owning handle `owned_config` and the released raw pointer `config`.

plugins/experimental/stale_response/stale_response.cc
bryancall: Yes, done in 34323cc, and it removed more than the cast. The field defaulted to the static `PLUGIN_TAG`, which is what forced both the `const_cast` and the `!= PLUGIN_TAG` pointer identity test that decided whether freeing was safe. It now defaults to null and the tag is substituted at the point of use, so the free is unconditional and the comparison is gone. That comparison was also fragile: if the default ever became a `strdup`, it would have silently degraded to a leak.

plugins/xdebug/xdebug.cc
bryancall: Yes, done in 34323cc. `str` is only ever assigned a `TSstrdup` result, never a literal, so `char *` is accurate and it removed a second `const_cast` at the `TSUserArgSet` call as well.

JosiahWI (dismissed):

bryancall:
Copilot posted three suppressed notes in its last review rather than inline threads, so there is nothing to reply to in place. Recording what I did with them here.

Took two of them in e860bb7:

- The redirect URL comment was describing a hazard the final code never has. It explained why passing `std::string` storage would be wrong, which is not visible in this diff at all, because the net change is from an unowned `ats_strdup` to a scoped one. Reworded to say what is true of the code as written: the parser needs a mutable buffer, it keeps no pointer into that buffer, and the duplicate previously leaked.
- `stale_response` was managing the log filename with `strdup`/`free` while every other allocation in that file uses the ATS wrappers. Those three sites were the only bare ones, so they are now `TSstrdup`/`TSfree`.

Partly took the third. The suggestion was to encode ownership in the types for `Cache`, holding `URLset` as owning smart pointers and wrapping `stripes_hash_table` with an `ats_free` deleter. I agree that is the better design, but it changes the type of a public member that two unrelated code paths iterate, which is more than a leak fix should carry, so I would rather do it separately.

What I did take is the concrete risk behind it: `build_stripe_hash_table()` replaced the table without releasing the previous one, and it is called from two places. That is harmless while each `Cache` builds the table once, but the destructor added here makes that pointer owning, so the function now frees any table already installed.

Verified on Fedora with GCC 16.1.1: clean build, no new warnings, 140/140 unit tests, with `xdebug`, `stale_response`, `jax_fingerprint` and `uri_signing` all compiled.

bryancall:
Another round of suppressed Copilot notes, so again there are no threads to reply to. Recording the dispositions.

**Took the `stale_response` one** in 2b3c27c. The point about a nullable owning pointer pushing the `PLUGIN_TAG` fallback onto every caller was right, and holding it as a `std::string` is better than either version I had: empty means the default tag, the effective name is computed where it is used, and the destructor no longer owns anything. It also turns the defect that started this from fixed into impossible, since assigning to a string releases the previous value and there is no ownership left to get wrong.

**Declined the `uri_signing` one.** The suggestion was to replace the `malloc`/`strcpy` pair for `cfg->id` with `TSstrdup`/`TSfree` for consistency. That file is C style throughout: `config_new()` allocates with `malloc` and `config_delete()` releases with `free`, so switching only the duplication site would create exactly the allocator mismatch the note warns about. Making the whole file consistent is a reasonable cleanup but it is not this change.

**Declined the `CacheTool` one again**, same reasoning as last round: encoding ownership in the member types is the better design, but it changes the type of a public member that two unrelated code paths iterate, and I would rather not carry that in a leak fix. The concrete risk it names, a second `build_stripe_hash_table()` call leaking the previous table, is already handled in e860bb7.

Verified: Debug build clean with 140/140 unit tests, and a Release build with warnings as errors produces no errors beyond the two that master already produces with this compiler.

Note on the CentOS failure on the previous commit: `ci-centos` is the only Release configuration in the matrix and was the only platform that failed, with Debian, Fedora, Rocky and Ubuntu all green on the same commit. I could not reproduce it locally, so I am watching whether this push clears it.

bryancall:
Third round of suppressed notes. Two of them I had declined before, and on a closer look one of those was worth taking after all.

**Took the stripe hash table one** in 639cea0. I had been declining this as a type redesign, but there is an in-tree idiom that makes it a two line change: `ats_scoped_mem<T>` in `tscore/ink_memory.h`. Its assignment operator clears the existing resource first, so holding the table in one removes **both** manual frees at once, the destructor release and the free-before-reassign I added in e860bb7. That is a better answer than either version I had, so thanks for pressing on it.

**Took the `uri_signing` one**, which changed since last round. Previously it suggested `TSstrdup`/`TSfree`, which I declined because that file allocates with `malloc` and releases with `free` throughout, so switching only the duplication site would have created the very mismatch the note warned about. This round it suggests a plain single-call duplication instead, and `strdup` pairs correctly with the `free` already in `config_delete()`, so `malloc` plus `strcpy` is now just `strdup`.

**Took the spelling note.** The tree overwhelmingly writes "null-terminated" (47 occurrences against 2 of the form I used), so the comment now matches.

**Still declining the URL set half.** `URLset` is a public member iterated by two code paths that are otherwise untouched here, so converting it to an owning container changes their loops as well. That is a reasonable cleanup on its own and I would rather it be its own change than ride along in a leak fix.

Verified: Debug build clean with 140/140 unit tests, and a Release build with warnings as errors produces no errors beyond the two master already produces with this compiler.

bryancall:
Fourth round of suppressed notes. I am declining all four, and I want to be specific about why rather than just saying no.

**`reset(...)` on the stripe hash table.** `ats_scoped_mem` does not have a `reset()`. The public API of `ats_scoped_resource` is `clear()`, `get()`, `release()`, and `operator=`, so assignment *is* the ownership transfer API for this type. The suggestion cannot be applied as written.

**Owning container for `URLset`.** Fourth time for this one and my answer is unchanged: `URLset` is a public member iterated by two code paths this change does not otherwise touch, so converting it edits their loops too. Still worth doing, still not in a leak fix.

**Unchecked `TSTextLogObjectCreate`.** This one is legitimate and already identified. It was triaged out of this pull request deliberately: adding a failure branch and an error log is a behavior change, and this pull request is scoped to leaks whose fix is not observable. It is sitting in the behavioral batch with the other changes of that kind, and it will come with an error path rather than as a drive-by.

**`strdup` portability.** This contradicts the previous round, which asked for exactly this change: it suggested a single-call duplication and I used `strdup`. On the substance, plain `strdup` already has 11 call sites across `src/` and `plugins/`, and this file allocates with `malloc` and releases with `free` throughout, so `strdup` is both consistent locally and established in the tree.

More generally, this is the fourth automated review pass on this pull request and the notes have moved from real defects to preferences about the shape of the fixes, including one that contradicts an earlier round and one that targets an API the type does not have. The earlier passes were genuinely valuable and caught a real regression I introduced. This one is not moving the code forward, and every push re-runs the full matrix, so I am going to leave the code here and let human review drive from this point.

State: all checks green on the current head, with the Release build (the only one in the matrix, on CentOS) passing.

bryancall:
[approve ci autest 3]

JosiahWI (approved):

JosiahWI:
[approve ci freebsd osx clang-analyzer autest 0]

JosiahWI:
[approve ci autest 1]

bryancall:
[approve ci autest 2]

PR #13554, opened 2026-08-15
Contention-aware per-event send budget for QUICStream

maskit:
## Summary

Benchmarking qmux (HTTP/3-over-TCP) against H2 showed a throughput deficit
specific to large, single-stream responses, absent for small objects and
mostly recovered once multiple concurrent streams share a connection.

`QUICStream::send_data()` capped how much of one stream's data it sends
per write event to a fixed 16KB, regardless of contention. A large object
on a solo stream needs many separate event-loop round trips to drain,
each paying real per-event overhead (quiche iterator alloc/free, an FFI
crossing, a flush). This makes that per-event budget contention-aware:
it scales down toward the existing 16KB floor under real contention
(preserving today's fairness guarantee exactly), and grows toward a new
256KB ceiling when a stream has the connection's write path to itself.
The budget is sized from the actual number of writable streams for that
same event, counted via one extra `quiche_conn_writable()` drain before
sizing the budget and servicing streams. An earlier version of this sized
the budget from the *previous* event's count instead, to avoid that extra
drain -- but a sudden swing in contention between events (including a
connection's very first write event, which has no previous count at all)
could then hand every stream a budget sized for the wrong count, so this
always counts fresh instead of carrying a lagging estimate.

Also, while touching the HTTP/3 frame-handling path, removed some dead
code and per-request allocations found along the way:

- `Http3FrameDispatcher`'s per-type handler list moved from a
  heap-allocated `std::vector` to fixed inline storage, since at most a
  handful of handlers ever register per frame type and this dispatcher
  is a per-transaction object.
- `Http3Transaction`'s five per-transaction helper objects (frame
  generators/handlers) converted from heap-allocated pointers to value
  members, removing five allocations/frees per HTTP/3 transaction.
- Removed `Http3UnknownFrame` and `Http3HeadersFrame`'s `ats_unique_buf`
  constructor/factory overload, both fully dead code.
- Aligned `Http3HeadersFrame`'s send-path constructor to take
  `IOBufferReader &` like `Http3DataFrame`'s equivalent, instead of `*`.

## Test plan

- [x] New unit test (`test_QUICStream.cc`) covers
      `compute_fair_send_budget()`'s no-contention, heavy-contention,
      mid-range, and floor-boundary cases.
- [x] `test_net` and `test_http3` pass against BoringSSL + quiche + qmux.
- [x] `test_net` and `test_http3` pass against OpenSSL's native QUIC
      backend (a separate, non-quiche `NetVConnection` implementation
      that also compiles this code).

### DIFF
--- a/src/iocore/net/QUICNetVConnection.cc
+++ b/src/iocore/net/QUICNetVConnection.cc
@@ -688,17 +688,22 @@ void
 QUICNetVConnection::_handle_write_ready()
 {
   if (quiche_conn_is_established(this->_quiche_con)) {
+    const size_t budget = QUICStream::compute_fair_send_budget(this->_last_writable_stream_count);
+
     quiche_stream_iter *writable = quiche_conn_writable(this->_quiche_con);
     uint64_t            s        = 0;
+    size_t              count    = 0;
     while (quiche_stream_iter_next(writable, &s)) {
+      ++count;
       QUICStream *stream = static_cast<QUICStream *>(this->_stream_manager->find_stream(s));
       if (stream == nullptr) {
         [[maybe_unused]] QUICConnectionError err;
         stream = this->_stream_manager->create_stream(s, err);
       }
-      stream->send_data(*this);
+      stream->send_data(*this, budget);
     }
     quiche_stream_iter_free(writable);
+    this->_last_writable_stream_count = count;
   }
 
   Ptr<IOBufferBlock> udp_payload;
--- a/src/iocore/net/QUICNetVConnection.cc
+++ b/src/iocore/net/QUICNetVConnection.cc
@@ -688,13 +688,26 @@ void
 QUICNetVConnection::_handle_write_ready()
 {
   if (quiche_conn_is_established(this->_quiche_con)) {
-    const size_t budget = QUICStream::compute_fair_send_budget(this->_last_writable_stream_count);
+    // Count real contention for THIS event before deciding its budget, rather than
+    // sizing it from a previous event's count -- a stale count can be wrong in either
+    // direction whenever contention swings between events, not just on the first
+    // event. writable() is a pure, side-effect-free snapshot (verified against
+    // quiche's source), so draining it twice costs one extra O(n) collect and n extra
+    // FFI calls, n bounded by this connection's stream limit -- cheap next to the
+    // per-stream work that follows.
+    quiche_stream_iter *probe          = quiche_conn_writable(this->_quiche_con);
+    uint64_t            probe_id       = 0;
+    size_t              writable_count = 0;
+    while (quiche_stream_iter_next(probe, &probe_id)) {
+      ++writable_count;
+    }
+    quiche_stream_iter_free(probe);
+
+    const size_t budget = QUICStream::compute_fair_send_budget(writable_count);
 
     quiche_stream_iter *writable = quiche_conn_writable(this->_quiche_con);
     uint64_t            s        = 0;
-    size_t              count    = 0;
     while (quiche_stream_iter_next(writable, &s)) {
-      ++count;
       QUICStream *stream = static_cast<QUICStream *>(this->_stream_manager->find_stream(s));
       if (stream == nullptr) {
         [[maybe_unused]] QUICConnectionError err;
@@ -703,7 +716,6 @@ QUICNetVConnection::_handle_write_ready()
       stream->send_data(*this, budget);
     }
     quiche_stream_iter_free(writable);
-    this->_last_writable_stream_count = count;
   }
 
   Ptr<IOBufferBlock> udp_payload;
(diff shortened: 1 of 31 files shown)


src/iocore/net/quic/QUICStream.cc
```diff
@@ -144,24 +155,28 @@ QUICStream::receive_data(QUICStreamIO &stream_io)
...
   ssize_t                    len = 0;
   [[maybe_unused]] ErrorCode error_code{0};
   size_t                     written_this_event = 0;
+  // _write_vio.nbytes is set once when the VIO is armed and doesn't change over the
+  // course of this call, so query it once instead of re-locking the adapter's mutex
+  // for the same value on every loop iteration below.
+  const uint64_t total_len = this->_adapter->total_len();
 
-  while (written_this_event < MAX_STREAM_SEND_BYTES_PER_EVENT) {
+  while (written_this_event < max_bytes_this_event) {
```
maskit, line 169: This is the biggest improvement. Fixed size cap -> Variable size cap based on the number of streams.

src/proxy/http3/test/test_Http3Frame.cc
```diff
@@ -105,20 +105,28 @@ TEST_CASE("Store HEADERS Frame", "[http3]")
...
+    // consumed above -- it holds its own independent clone of the reader.
+    auto           ibb = frame->to_io_buffer_block();
     IOBufferReader reader;
     reader.block = ibb.get();
     len          = reader.read_avail();
     reader.read(buf, sizeof(buf));
     CHECK(len == 6);
     CHECK(memcmp(buf, expected1, len) == 0);
+
+    free_MIOBuffer(header_block);
```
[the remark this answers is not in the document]
maskit: Fixed

src/iocore/net/quic/QUICStream.cc
bneradt, line 169: **[P2] Cap a carried pending block to this event’s remaining budget.**

`_pending_send_block` can have been created under a previous 256 KB budget and survive a partial write. If contention then reduces `max_bytes_this_event` to 16 KB, the later `write_stream()` call still passes the entire remaining block, allowing the stream to exceed the computed budget. Please cap the submitted length to `max_bytes_this_event - written_this_event` (and only set `fin` when the whole pending block is included). A unit test with a partial first write followed by a smaller budget would cover this transition.

src/iocore/net/QUICNetVConnection.cc
```diff
@@ -688,17 +688,22 @@ void
 QUICNetVConnection::_handle_write_ready()
 {
   if (quiche_conn_is_established(this->_quiche_con)) {
+    const size_t budget = QUICStream::compute_fair_send_budget(this->_last_writable_stream_count);
```
bneradt: **[P2] Avoid using the maximum budget for every stream on the first event.**

`_last_writable_stream_count` starts at 1, so an event whose current iterator contains many streams gives all of them the 256 KB maximum. With the configured 100-stream limit, that can queue or process roughly 25.6 MB in one event instead of the previous 1.6 MB, delaying unrelated connections on the same event thread. The QMux path has the same pattern. Please initialize conservatively, enforce a connection-wide per-event cap, or otherwise account for current contention.

src/iocore/net/QUICNetVConnection.cc
```diff
@@ -688,6 +688,23 @@ void
...
+    // direction whenever contention swings between events, not just on the first
+    // event. writable() is a pure, side-effect-free snapshot (verified against
+    // quiche's source), so draining it twice costs one extra O(n) collect and n extra
+    // FFI calls, n bounded by this connection's stream limit -- cheap next to the
+    // per-stream work that follows.
+    quiche_stream_iter *probe          = quiche_conn_writable(this->_quiche_con);
+    uint64_t            probe_id       = 0;
+    size_t              writable_count = 0;
+    while (quiche_stream_iter_next(probe, &probe_id)) {
+      ++writable_count;
```
bneradt, line 702: **[P2] Count streams that actually have data to send.**

`quiche_conn_writable()` reports every unfinished stream with flow-control capacity, not just streams with queued application data ([quiche 0.28 `StreamMap::writable()`](https://github.com/cloudflare/quiche/blob/0.28.0/quiche/src/stream/mod.rs#L628-L633), [`Stream::is_writable()`](https://github.com/cloudflare/quiche/blob/0.28.0/quiche/src/stream/mod.rs#L786-L793)). For example, with 20 open request streams but response data available on only one, `writable_count` is 20, so the active stream receives the 16 KB floor while the other 19 consume none of their budgets. This repeats every event and effectively restores the original per-stream limitation. Please base the divisor on ATS streams whose `QUICStream::has_data_to_send()` is true; ideally collect that filtered snapshot once and reuse it for the send loop. The QMux path has the same issue.

bneradt (dismissed):

bneradt (approved):

PR #13557, opened 2026-08-18
http: funnel cache key generation through `HttpSM`

JakeChampion:
The four cache key call sites each open-coded the same `Cache::generate_key` call with the same two `txn_conf` arguments, and one of them also open-coded the ATS 9.2 compatibility branch. Collect that into a single `HttpSM::generate_cache_key` so callers stop repeating the configuration plumbing.

### DIFF
--- a/include/proxy/http/HttpSM.h
+++ b/include/proxy/http/HttpSM.h
@@ -429,6 +429,7 @@ private:
   void do_drain_request_body(HTTPHdr &response);
 
   void wait_for_full_body();
+  void generate_cache_key(HttpCacheKey *key, URL *url, bool compat = false);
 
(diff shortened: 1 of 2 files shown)


JosiahWI (commented):
This is an interesting refactor! The new method is has private access, as it should. What do you think of putting the 9.x compatibility case into a separate method? Note that this is the same approach used for the `Cache` methods (there are two methods, not one method with a boolean switch).

moonchen (approved):
Thanks for cleaning this up!

JakeChampion:
not sure why the pr got closed... i also can not re-open this one, i will have to open a new one

PR #13558, opened 2026-08-18
keep buflen in sync in dns_process address copy path

dxbjavid:
dns_process copies each answer record into the fixed HostEnt hostbuf while keeping a running buflen for the space left at the write cursor, but the branch that memcpys an A/AAAA address when the source is not word aligned advances the cursor by the alignment padding and the record length without decrementing buflen. Since every later name expansion hands that same buflen to dn_expand as its output limit, the stale, over-large value lets a crafted response (a large address record followed by a record whose owner name expands near the end of hostbuf) write the expanded name past the buffer and into the adjacent srv_hosts vector. This restores buflen from the cursor after the copy so it stays consistent, matching what the name, CNAME and PTR paths already do.

### DIFF
--- a/src/iocore/dns/DNS.cc
+++ b/src/iocore/dns/DNS.cc
@@ -1889,8 +1889,9 @@ dns_process(DNSHandler *handler, HostEnt *buf, int len)
           memcpy((*hap++ = bp), cp, n);
           Dbg(dbg_ctl_dns, "received %s = %s", QtypeName(type),
               inet_ntop(T_AAAA == type ? AF_INET6 : AF_INET, bp, ip_string, sizeof(ip_string)));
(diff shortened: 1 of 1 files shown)


moonchen (approved):

bryancall (approved):

dxbjavid:
Thank you, much appreciated

cmcfarlen:
Cherry-picked to the 10.2.x branch as b880bea1ac7156ed0f1d2967b07591d66622e41d for the 10.2.1 release.

cmcfarlen:
Cherry-picked to the 10.1.x branch as 227594774caea39b513f4e953fcb00a531e6cb01 for the 10.1.5 release.

PR #13559, opened 2026-08-18
Build the well-known-string table at compile time

moonchen:
`hdrtoken_init()` builds the well-known-string (WKS) table at startup.  This PR does the work at compile-time, which enables more robust checking, and removes an old restriction regarding prefixes.

### 1. Modernize the well-known-string initialization

Make use of `constexpr` to initialize WKS at compile time.  Remove the unnecessary `name` field from the WKS table at runtime.  Trim the `hdrtoken_hash_table` size to just the reachable number of slots, build it at compile time as well, and shrink its buckets from 16 to 8 bytes by storing an index instead of a pointer.

|                              | before             | after                  |
| ---------------------------- | ------------------ | ---------------------- |
| WKS table                    | ~15 KB on the heap | 10,800 B in `.rodata`  |
| `hdrtoken_hash_table`        | 1,052,904 B        | 262,144 B in `.rodata` |
| `sizeof(HdrTokenHeapPrefix)` | 56                 | 48                     |

### 2. Add compile-time checks for WKS correctness

Checks are now done at compile-time, and the new checks are more strict than the old release asserts.

An append-only ledger (`_hdrtoken_strs_frozen`) fingerprints the frozen entries: inserting, reordering, or editing them fails to compile, and appending strings requires appending a matching ledger row.

### 3. Remove the restriction that prevents a WKS from being added after its prefix

In order to maintain cache compatibility, new strings needed to be appended to the list.  Previously, it was not legal to append a string to WKS if its prefix already exists.

This PR removes the prefix restriction.  We can now append strings like `Server-Timing`, `Accept-CH`, and `Expect-CT`.

### DIFF
--- a/include/proxy/hdrs/HdrToken.h
+++ b/include/proxy/hdrs/HdrToken.h
@@ -31,7 +31,6 @@
 #include "tscore/ink_memory.h"
 #include "tscore/ink_string.h"
 #include "tscore/Allocator.h"
-#include "tsutil/Regex.h"
 #include "tscore/ink_apidefs.h"
 
 ////////////////////////////////////////////////////////////////////////////
@@ -57,20 +56,28 @@ enum class HdrTokenInfoFlags : uint32_t {
   PROXYAUTH = 1 << 3,
 };
 
-inline HdrTokenInfoFlags
+constexpr HdrTokenInfoFlags
 operator|(HdrTokenInfoFlags a, HdrTokenInfoFlags b)
 {
   return static_cast<HdrTokenInfoFlags>(static_cast<uint32_t>(a) | static_cast<uint32_t>(b));
 }
 
-inline HdrTokenInfoFlags
+constexpr HdrTokenInfoFlags
 operator&(HdrTokenInfoFlags a, HdrTokenInfoFlags b)
 {
   return static_cast<HdrTokenInfoFlags>(static_cast<uint32_t>(a) & static_cast<uint32_t>(b));
 }
 
-struct HdrTokenFieldInfo {
+// One row of the field initializer table. The name binds the row to a well-known string and is not
+// needed once the row is resolved, so it is absent from HdrTokenFieldInfo below.
+struct HdrTokenFieldInit {
   const char       *name;
+  int32_t           slotid;
+  uint64_t          mask;
+  HdrTokenInfoFlags flags;
+};
+
+struct HdrTokenFieldInfo {
   int32_t           slotid;
   uint64_t          mask;
   HdrTokenInfoFlags flags;
@@ -92,8 +99,7 @@ struct HdrTokenHeapPrefix {
   HdrTokenTypeSpecific wks_type_specific;
 };
 
-extern DFA *hdrtoken_strs_dfa;
-extern int  hdrtoken_num_wks;
+extern int hdrtoken_num_wks;
 
 extern const char       *hdrtoken_strs[];
 extern int               hdrtoken_str_lengths[];
@@ -109,7 +115,6 @@ extern HdrTokenInfoFlags hdrtoken_str_flags[];
 ////////////////////////////////////////////////////////////////////////////
 
 extern void        hdrtoken_init();
-extern int         hdrtoken_tokenize_dfa(const char *string, int string_len, const char **wks_string_out = nullptr);
 extern int         hdrtoken_tokenize(const char *string, int string_len, const char **wks_string_out = nullptr);
 extern int         hdrtoken_method_tokenize(const char *string, int string_len);
 extern const char *hdrtoken_string_to_wks(const char *string);
@@ -141,12 +146,13 @@ hdrtoken_is_valid_wks_idx(int wks_idx)
 /*-------------------------------------------------------------------------
   -------------------------------------------------------------------------*/
 
-// ToDo: This, and dependencies / users should probably be const HdrTokenHeapPrefix * IMO.
-inline HdrTokenHeapPrefix *
+// The well-known strings live in a table that is built at compile time and is read-only, so a
+// prefix is only ever read through.
+inline const HdrTokenHeapPrefix *
 hdrtoken_wks_to_prefix(const char *wks)
 {
   ink_assert(hdrtoken_is_wks(wks));
-  return reinterpret_cast<HdrTokenHeapPrefix *>(const_cast<char *>(wks) - sizeof(HdrTokenHeapPrefix));
+  return reinterpret_cast<const HdrTokenHeapPrefix *>(wks - sizeof(HdrTokenHeapPrefix));
 }
 
 /*-------------------------------------------------------------------------
@@ -194,7 +200,7 @@ hdrtoken_index_to_flags(int wks_idx)
   return hdrtoken_str_flags[wks_idx];
 }
 
-inline HdrTokenHeapPrefix *
+inline const HdrTokenHeapPrefix *
 hdrtoken_index_to_prefix(int wks_idx)
 {
   ink_assert(hdrtoken_is_valid_wks_idx(wks_idx));
@@ -236,7 +242,7 @@ inline uint64_t
 hdrtoken_wks_to_mask(const char *wks)
 {
   ink_assert(hdrtoken_is_wks(wks));
-  HdrTokenHeapPrefix *prefix = hdrtoken_wks_to_prefix(wks);
+  const HdrTokenHeapPrefix *prefix = hdrtoken_wks_to_prefix(wks);
   return prefix->wks_info.mask;
 }
 
--- a/include/proxy/hdrs/HdrToken.h
+++ b/include/proxy/hdrs/HdrToken.h
@@ -24,6 +24,8 @@
 #pragma once
 
 #include <cassert>
+#include <cstddef>
+#include <cstdint>
 #include <sys/types.h>
 #include "tscore/ink_assert.h"
 #include "tscore/ink_atomic.h"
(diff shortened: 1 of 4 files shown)


moonchen:
[approve ci Autest 1]

moonchen:
[approve ci autest 1]

src/proxy/hdrs/HdrToken.cc
```diff
@@ -274,239 +264,309 @@ HdrTokenFieldInfo _hdrtoken_strs_field_initializers[] = {
...
-void
-hdrtoken_hash_init()
+constexpr size_t
+hdrtoken_max_literal_length()
 {
-  uint32_t i;
-  int      num_collisions;
+  const auto longest = std::max_element(std::cbegin(_hdrtoken_strs), std::cend(_hdrtoken_strs),
+                                        [](std::string_view a, std::string_view b) { return a.length() < b.length(); });
+  return longest->length();
```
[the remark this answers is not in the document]
moonchen: `min_element`/`max_element` have been `constexpr` since C++14 (N3471); `std::sort` and `adjacent_find` only became `constexpr` in C++20, which is why the note singles those two out.

Checked against the oldest standard library in the CI matrix rather than inferred: both `max_element` overloads are declared `_GLIBCXX14_CONSTEXPR` on the libstdc++ 9 release branch. The Ubuntu job builds with clang-12 against libstdc++ 9 and passes, as does FreeBSD on libc++. Leaving as is.

src/proxy/hdrs/HdrToken.cc
```diff
@@ -274,239 +264,309 @@ HdrTokenFieldInfo _hdrtoken_strs_field_initializers[] = {
...
+hdrtoken_ascii_toupper(unsigned char c)
+{
+  unsigned char const is_lower = static_cast<unsigned char>((static_cast<unsigned>(c) - 'a') < 26u);
 
-/**
-  basic FNV hash
-**/
-#define TINY_MASK(x) (((uint32_t)1 << (x)) - 1)
+  return static_cast<unsigned char>(c - (is_lower << 5));
+}
```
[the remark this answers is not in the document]
moonchen: Agreed, switched to the plain ASCII-range form.

The terse version was presumably for the runtime hashing path, but that does not hold up: at `-O2` both forms produce equivalent code (19 vs 18 instructions, same number of branch/select instructions), and a `static_assert` sweep confirms they agree on all 256 byte values. The generated well-known-string table is byte-for-byte unchanged.

include/proxy/hdrs/HdrToken.h
```diff
@@ -109,8 +115,7 @@ extern HdrTokenInfoFlags hdrtoken_str_flags[];
 ////////////////////////////////////////////////////////////////////////////
 
 extern void        hdrtoken_init();
-extern int         hdrtoken_tokenize_dfa(const char *string, int string_len, const char **wks_string_out = nullptr);
-extern int         hdrtoken_tokenize(const char *string, int string_len, const char **wks_string_out = nullptr);
+int                hdrtoken_tokenize(const char *string, int string_len, const char **wks_string_out = nullptr);
```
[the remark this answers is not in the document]
moonchen: Good catch, this was unintentional. Upstream has `extern int hdrtoken_tokenize(...)`; the `extern` was lost when the adjacent `hdrtoken_tokenize_dfa` declaration was deleted. Restored, along with the alignment of that declaration block.

include/proxy/hdrs/HdrToken.h
```diff
@@ -141,12 +146,13 @@ hdrtoken_is_valid_wks_idx(int wks_idx)
...
-// ToDo: This, and dependencies / users should probably be const HdrTokenHeapPrefix * IMO.
-inline HdrTokenHeapPrefix *
+// The well-known strings live in a table that is built at compile time and is read-only, so a
+// prefix is only ever read through.
+inline const HdrTokenHeapPrefix *
 hdrtoken_wks_to_prefix(const char *wks)
 {
   ink_assert(hdrtoken_is_wks(wks));
-  return reinterpret_cast<HdrTokenHeapPrefix *>(const_cast<char *>(wks) - sizeof(HdrTokenHeapPrefix));
+  return reinterpret_cast<const HdrTokenHeapPrefix *>(wks - sizeof(HdrTokenHeapPrefix));
```
bneradt: Now that `wks` points to the first element of the `WksEntry::str` array member, subtracting `sizeof(HdrTokenHeapPrefix)` moves outside that array subobject; the C++ pointer-arithmetic rules make that undefined behavior. The `offsetof(WksEntry, str)` assertion confirms physical layout but does not make the cross-subobject subtraction valid. The previous heap representation used one raw byte allocation, so it did not have this particular issue. Could we either preserve byte-buffer storage or obtain the prefix through a valid table/index relationship?

src/proxy/hdrs/HdrToken.cc
```diff
@@ -40,20 +41,9 @@ DbgCtl dbg_ctl_hdr_token{"hdr_token"};
 /*
  WARNING:  Indexes into this array are stored on disk for cached objects.  New strings must be added at the end of the array to
  avoid changing the indexes of pre-existing entries, unless the cache format version number is increased.
```
bneradt, line 42: Because these indexes are persisted in cached objects, could we enforce the append-only rule rather than relying solely on this comment? One option is a compile-time fingerprint of the ordered first `N` frozen entries, where `N` is independent of the current array size. Appending would continue to pass, while reordering, removing, or changing an existing entry would fail compilation and require an intentional cache-version decision.

src/proxy/hdrs/HdrToken.cc
```diff
@@ -274,239 +264,307 @@ HdrTokenFieldInfo _hdrtoken_strs_field_initializers[] = {
...
-/**
- */
-void
-hdrtoken_init()
+static_assert(hdrtoken_names_resolve_uniquely(_hdrtoken_strs_type_initializers),
+              "a token-type initializer names a string that is missing from _hdrtoken_strs or already claimed");
+static_assert(hdrtoken_names_resolve_uniquely(_hdrtoken_strs_field_initializers),
+              "a field initializer names a string that is missing from _hdrtoken_strs or already claimed");
+static_assert(hdrtoken_names_resolve_uniquely(_hdrtoken_strs_cc_initializers),
+              "a Cache-Control initializer names a string that is missing from _hdrtoken_strs or already claimed");
```
bneradt, line 450: Since these initializer rows are now `constexpr`, could validation cover their semantic invariants too? In particular: each non-`NONE` field slot should be in `[0, 31]` and unique; each nonzero presence mask should be single-bit and unique; and each Cache-Control mask should be nonzero, single-bit, unique, and bound to an entry typed `CACHE_CONTROL`. Cheap resolver-contract assertions for case folding and prefix rejection would also directly document the behavior this change is intended to enable.

src/proxy/hdrs/HdrToken.cc
```diff
@@ -274,239 +264,307 @@ HdrTokenFieldInfo _hdrtoken_strs_field_initializers[] = {
...
+ *                        H A S H    T A B L E                         *
+ *                                                                     *
+ ***********************************************************************/
+
+struct HdrTokenHashBucket {
+  const char *wks;
+  uint32_t    hash;
+};
+
+HdrTokenHashBucket hdrtoken_hash_table[HDRTOKEN_HASH_TABLE_SIZE];
```
bneradt: Could we make the hash table `constexpr` as well? A bucket storing `{wks_idx_plus_one, hash}` instead of a pointer can be built from `_hdrtoken_strs` at compile time, use zero as the empty value, and likely shrink each bucket from 16 bytes to 8. That would remove `hdrtoken_hash_init()` and another startup write. Using one `constexpr hdrtoken_hash(std::string_view)` for both construction and `hdrtoken_tokenize()` would also make compile-time/runtime hash parity structural.

include/proxy/hdrs/HdrToken.h
moonchen: Fixed. `hdrtoken_wks_to_prefix()` now recovers the entry index from the string pointer with integer (`uintptr_t`) arithmetic and returns `&hdrtoken_wks_entries[idx].prefix`, so no out-of-bounds pointer is formed. The entry struct moved to the header with fixed 32-byte string storage so call sites know the stride; sizeof(entry) is unchanged at 80, so the table layout and size are identical.

src/proxy/hdrs/HdrToken.cc
moonchen: Added, as a freeze ledger rather than a single constant: `_hdrtoken_strs_frozen` holds {count, fingerprint} rows (reusing the table's FNV-1a step, with a terminator folded in per entry). Every row is validated forever, so editing, inserting, reordering, or removing a frozen entry breaks a fingerprint, and a second assert requires the last row's count to match the table size — appending strings therefore forces appending a new ledger row, making the cache-affecting step deliberate.

src/proxy/hdrs/HdrToken.cc
moonchen: Added. Field rows: non-NONE slot ids must be in [0, 31] and unique; nonzero presence masks single-bit and unique. Cache-Control rows: masks nonzero, single-bit, unique, and the entry typed CACHE_CONTROL, which `HTTPHdr::is_cache_control_set()` asserts for any directive it checks. Also added the resolver-contract asserts: case-insensitive resolution, no prefix matching, and a WKS that prefixes another (`Accept`) resolving to its own entry.

src/proxy/hdrs/HdrToken.cc
moonchen: Done. Buckets are now {wks_idx_plus_one, hash} (8 bytes, zero means empty) built at compile time, `hdrtoken_hash_init()` is removed, and one `constexpr hdrtoken_hash(std::string_view)` is shared by construction and `hdrtoken_tokenize()`, so hash parity is structural. The table drops from 524,288 B of mutable data to 262,144 B in .rodata.

bneradt (approved):

PR #13561, opened 2026-08-18
Test DNS TCP split length reads

bneradt:
Add a unit test for accumulating the DNS-over-TCP length prefix one byte at a time. Move the prefix state update into DNSConnection::TCPData so the split-read behavior can be exercised without constructing a DNSHandler.


JosiahWI:
Do you know whether it's possible to test this through a public interface?

bneradt:
> Do you know whether it's possible to test this through a public interface?

I don't think easily. This came out of another code review. I don't like the tail wagging the dog (cluttering the API for the test). I'm going to close this.

PR #13562, opened 2026-08-18
Test Lua server request guards

bneradt:
Add an AuTest that calls ts.server_request accessors and setters before ATS has built a server request. This covers the guarded failure path and verifies the calls return nil instead of crashing.

### DIFF
--- a/tests/gold_tests/pluginTest/lua/lua_server_request_guard.test.py
+++ b/tests/gold_tests/pluginTest/lua/lua_server_request_guard.test.py
@@ -1,0 +1,83 @@
+'''
+Verify ts.server_request APIs fail safely before a server request exists.
+'''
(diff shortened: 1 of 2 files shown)


moonchen (approved):
Looks good. This adds useful regression coverage for calling `ts.server_request` APIs before ATS has constructed the server request. The assertions verify that the accessors return `nil`, while successfully reaching `SEND_RESPONSE_HDR` verifies that the setter calls do not crash.

Optional: since the test passes deliberately different values to the setters, it could also verify that the eventual upstream request remains unchanged. For example, with Proxy Verifier request expectations in an `ATSReplayTest`. That is not necessary for the crash regression covered here, but it would add explicit coverage that these early setter calls are no-ops.

Approving.

PR #13566, opened 2026-08-18
Test AtomicSharedPtr concurrency

bneradt:
Add Catch2 coverage for AtomicSharedPtr load/store/exchange operations and concurrent readers during writer swaps. Also clarify the fallback implementation comment so it no longer relies on stale library-version shorthand.

### DIFF
--- a/src/tsutil/unit_tests/test_AtomicSharedPtr.cc
+++ b/src/tsutil/unit_tests/test_AtomicSharedPtr.cc
@@ -27,6 +27,7 @@
 
 #include <algorithm>
 #include <atomic>
+#include <chrono>
 #include <memory>
 #include <thread>
 #include <vector>
@@ -49,27 +50,25 @@ struct Payload {
 };
 
 struct ReaderState {
-  std::atomic_flag should_start;
-  std::atomic_flag should_stop;
-  std::atomic<int> invalid_reads{0};
-  std::atomic<int> read_count{0};
+  std::atomic<bool> should_start{false};
+  std::atomic<bool> should_stop{false};
+  std::atomic<int>  invalid_reads{0};
+  std::atomic<int>  read_count{0};
 };
 
 void
 run_reader(AtomicSharedPtr<Payload> &ptr, ReaderState &state)
 {
-  state.should_start.wait(false, std::memory_order_acquire);
+  while (!state.should_start.load(std::memory_order_acquire)) {
+    std::this_thread::yield();
+  }
 
-  while (!state.should_stop.test(std::memory_order_acquire)) {
+  while (!state.should_stop.load(std::memory_order_acquire)) {
     auto current = ptr.load(std::memory_order_acquire);
     if (current == nullptr || !current->is_valid()) {
       state.invalid_reads.fetch_add(1, std::memory_order_relaxed);
     }
     auto const reads = state.read_count.fetch_add(1, std::memory_order_release) + 1;
-    if (reads == 1) {
-      // Release start_readers() now that this reader has completed a read.
-      state.read_count.notify_all();
-    }
     if (reads % 64 == 0) {
       std::this_thread::yield();
     }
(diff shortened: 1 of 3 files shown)


bneradt:
Sanitizer validation completed in the Fedora 44 `asfats5` container with GCC 16.1.1:

- ASan: `test_tsutil "[AtomicSharedPtr]"` passed (11 assertions in 2 test cases) with no sanitizer findings.
- TSan: `test_tsutil "[AtomicSharedPtr]"` passed (11 assertions in 2 test cases) with no sanitizer findings.

src/tsutil/unit_tests/test_AtomicSharedPtr.cc
```diff
@@ -0,0 +1,141 @@
...
+  CHECK(*ptr.load() == 1);
+
+  auto second   = std::make_shared<int>(2);
+  auto previous = ptr.exchange(second);
+  CHECK(previous == first);
+  CHECK(ptr.load() == second);
+  CHECK(*ptr.load() == 2);
+}
+
+TEST_CASE("AtomicSharedPtr supports concurrent readers during writer swaps", "[libts][AtomicSharedPtr]")
```
JosiahWI, line 134: It looks like some helpers could be extracted here.

bneradt: Done. I extracted reader execution, construction, readiness waiting, and shutdown into helpers, leaving the test focused on writer swaps and assertions.

src/tsutil/unit_tests/test_AtomicSharedPtr.cc
```diff
@@ -0,0 +1,162 @@
...
+    return std::all_of(std::begin(values_), std::end(values_), [this](int value) { return value == generation_; });
+  }
+
+  int generation_ = 0;
+  int values_[VALUE_COUNT];
+};
+
+struct ReaderState {
+  std::atomic<bool> should_start{false};
+  std::atomic<bool> should_stop{false};
```
JosiahWI: The `std::atomic_flag` type would be more suitable for these flags and would simplify the waiting.

JosiahWI (changes requested):

src/tsutil/unit_tests/test_AtomicSharedPtr.cc
bneradt: On the two flags: switched both to `std::atomic_flag`. `should_start` now uses `wait(false, acquire)` against the writer's `test_and_set(release)` + `notify_all()`, which drops the spin-with-`yield()` loop, and `should_stop` uses `test(acquire)`.

While doing that I also removed `wait_for_reader()` and its 5s timeout in favor of `read_count.wait(0, acquire)`, with the first reader issuing a single `notify_all()`. The timeout wasn't buying anything: its failure path called `stop_readers()`, which joins the readers, so a reader that never saw `should_start` would have hung in the join anyway. Net effect is 27 lines out, 19 in.

bneradt: Had to back this out, unfortunately. CI caught that the C++20 synchronization library is not available on every toolchain we still build with: the CentOS job uses devtoolset-10, whose libstdc++ predates it.

```
error: 'struct std::atomic_flag' has no member named 'wait'
error: 'struct std::atomic_flag' has no member named 'test'
error: 'struct std::atomic<int>' has no member named 'notify_all'
```

Ubuntu fails the same way.

The pre-C++20 `std::atomic_flag` interface is only `test_and_set` and `clear`, and neither flag can be expressed with those: a reader polling `test_and_set` would set the flag it is polling and release itself on the next iteration. Initialization is awkward too, since the default constructor only became clear-initializing in C++20 and before that needs the now-deprecated `ATOMIC_FLAG_INIT`.

So 7277ac6 restores the `std::atomic<bool>` flags and the spin-with-yield gate. Happy to revisit if the CentOS job moves off devtoolset-10.

JosiahWI (approved):

PR #13567, opened 2026-08-18
Metrics: Revert added locking and annotations

moonchen:
The locking that the thread safety annotations required on the `ts::Metrics::Storage` read paths introduces too much of a performance regression. Revert it for now, until a better solution can be found.

The thread safety analysis infrastructure and its use in `SSLOriginSessionCache` are unaffected.

### DIFF
--- a/include/tsutil/Metrics.h
+++ b/include/tsutil/Metrics.h
@@ -38,7 +38,6 @@
 #include "swoc/MemSpan.h"
 
 #include "tsutil/Assert.h"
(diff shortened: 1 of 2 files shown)


cmcfarlen (approved):
Thanks! I know copilot is not happy about it, but we can address the "data race" with atomics in a followup PR.

cmcfarlen:
Cherry-picked to 10.2.x

PR #13568, opened 2026-08-18
Compact response HTTP version log fields

### LINKED ISSUE #12891: Log field sshv and csshv take 16 bytes for each
Both `sshv` and `csshv` store just an HTTP version, the major version and the minor version. The type of those fields is `dINT` (double int) somehow. A 64 bit integer for the major version and another 64 bit integer for the minor version.

We should change the type to `STRING` and use just 1 byte (or 2 bytes at maximum). Even `sINT` (64 bit integer) is too much.

This change would cause log format incompatibility.

bneradt:
The sshv and csshv access log fields store two one-byte version
components as separate 64-bit integers, consuming 16 bytes per entry and
publishing dINT framing in binary log schemas.

This patch addresses the overhead by marshaling major.minor strings in
one 8-byte-aligned slot while preserving the existing HTTP/major.minor
ASCII form. It updates v3 schema expectations and regression coverage
for both fields.

Fixes: #12891


maskit (approved):
I wonder if we should keep `dINT`.

PR #13571, opened 2026-08-19
Use __atomic builtins as the freelist 128-bit CAS fallback

### LINKED ISSUE #13555: Heads-up: riscv64 build fails at `ink_queue.h` "unsupported processor" (no RISC-V freelist case)
G'day ATS team,

First off — thank you for Traffic Server. We're big fans of the project, and we've been working on packaging **ATS 10** for [Nixpkgs](https://github.com/NixOS/nixpkgs) (the package set behind NixOS). Work-in-progress PR here: https://github.com/NixOS/nixpkgs/pull/474177

While testing that package across a few architectures, we stumbled onto a small issue we thought was worth flagging — **not urgent at all**, purely a friendly heads-up for if/when RISC-V is on your radar.

## What we saw

x86_64 and aarch64 build and run great (we tested aarch64 natively on real hardware). On **riscv64** the build stops at:

```
include/tscore/ink_queue.h:177:2: error: #error "unsupported processor"
```

This reproduces both cross-compiling to riscv64 and building natively on a riscv64 board (SpacemiT K1). It's present on `master` as well as the `10.2.0-rc1` tag.

## Why it happens

The lock-free freelist head in [`include/tscore/ink_queue.h`](https://github.com/apache/trafficserver/blob/master/include/tscore/ink_queue.h) chooses one of two strategies:

1. a **generic path** when the toolchain has a lock-free 128-bit CAS — `#elif TS_HAS_128BIT_CAS` (line ~127), or
2. an **architecture-specific pointer-tagging** layout, hand-coded only for x86_64 (line ~133) and aarch64 (line ~157); everything else hits the `#else` → `#error` (line ~177).

`TS_HAS_128BIT_CAS` is decided by [`cmake/Check128BitCas.cmake`](https://github.com/apache/trafficserver/blob/master/cmake/Check128BitCas.cmake), which compiles:

```c
int main(void) {
    __int128_t x = 0;
    return __sync_bool_compare_and_swap(&x, 0, 10);
}
```

On riscv64 there's no inline lock-free 128-bit CAS, so this lowers to a libatomic call and the probe fails to compile/link; the fallback retry uses `-mcx16`, which is x86-only. So `TS_HAS_128BIT_CAS` ends up **off**, and with no riscv64 branch in `FREELIST_POINTER`, the build falls through to the `#error`.

## Possible directions (just ideas — you know the code far better than we do)

- Add a **riscv64 case** to the `FREELIST_POINTER` / `FREELIST_VERSION` / `SET_FREELIST_POINTER_VERSION` macros, analogous to the existing aarch64 block. One thing to watch: riscv64 Linux virtual-address width isn't fixed (Sv39 / Sv48 / Sv57), so the number of spare high bits available for the version counter varies — the aarch64 block assumes a 52-bit VA, whereas a conservative riscv64 layout might need fewer version bits.
- Or allow the generic `TS_HAS_128BIT_CAS` path on riscv64 (e.g. via libatomic), accepting that 128-bit CAS there may not be lock-free without the Zacas extension.

Either way, we didn't want to presume the right approach — mostly just wanted to put this on your radar.

## Offer to help

We have a riscv64 board (SpacemiT K1) and a cross-compile setup, so we're very happy to **test any patch** on real RISC-V hardware if that's useful.

Thanks again for all the work on Traffic Server!

### Environment
- ATS: `10.2.0-rc1` tag and `master`
- Arch: `riscv64` (reproduced cross-compiling on x86_64 and natively on a SpacemiT K1)
- Toolchain: GCC (via Nixpkgs)

phongn:
### Summary

ATS does not build on riscv64. The build stops in `ink_queue.h` with the error "unsupported processor" (#13555). The freelist head needs one of two mechanisms: a 128-bit CAS, or a hand-written pointer layout with version bits. riscv64 has neither mechanism.

This change adds a portable fallback. When the 16-byte `__sync` builtins are not available, the freelist uses the `__atomic` builtins. The `__atomic` builtins lower to libatomic calls. The build then links against libatomic.

### Why not wait for Zacas hardware

The Zacas extension does not help here:

* GCC does not emit the 128-bit `amocas.q` instruction. The initial Zacas support deferred it, and it never landed.
* LLVM refuses to emit `amocas.q` for ABI-compatibility reasons. See the [LLVM RISC-V usage notes](https://llvm.org/docs/RISCVUsage.html).
* Zacas is an optional extension in the ratified RVA23 profile.

Thus, on riscv64 the 16-byte atomic operations always go through libatomic. This is true on current hardware and on future Zacas hardware.

### Why not a hand-written pointer layout

A packed 64-bit layout is possible, but it is fragile on RISC-V:

* The virtual address width varies (Sv39, Sv48, Sv57). Linux kernels before the mmap window change (~6.9) returned Sv57 addresses by default. A packed pointer silently corrupts on such kernels.
* An Sv57-safe layout leaves only 6 version bits. `LogObject.cc` uses the version field as a reference count, so few version bits are not safe.

The `__atomic` fallback has no address-space assumptions and keeps the full 64-bit version field. A packed riscv64 fast path can come later as an optimization if measurements justify it.

### Correctness

On riscv64, libatomic implements the 16-byte operations with internal locks. This is correct for this code because every access to a shared `head_p` goes through `INK_QUEUE_LD` and `ink_atomic_cas`. All of these operations serialize on the same libatomic lock, so the load/CAS retry loops keep their current semantics. The code comments now record this constraint.

The `__sync` path stays first in the probe order. x86-64 and aarch64 builds produce the same code as before; GCC does not inline the 16-byte `__atomic` CAS even with `-mcx16` (GCC PR80878), so a blanket switch would regress them.

Note: libatomic exports only `__atomic_*` symbols. The 16-byte `__sync` builtins emit undefined `__sync_*_16` references when the compiler cannot inline them, and nothing provides those symbols. That is why the fallback must use the `__atomic` builtins, and why the two paths are mutually exclusive.

### Changes

* `cmake/Check128BitCas.cmake`: probe the `__atomic` builtins when the `__sync` probe fails, first without and then with `-latomic`. New variables: `TS_HAS_128BIT_CAS_LIBATOMIC`, `TS_NEEDS_LIBATOMIC_FOR_CAS`. GCC >= 14 removed the automatic `-latomic` on RISC-V, so the explicit link is necessary.
* `ink_queue.h`: `INK_QUEUE_LD` becomes a 16-byte `__atomic_load` on the fallback tier; `head_p` keeps the plain `{pointer, int64 version}` layout there.
* `ink_atomic.h`: an `ink_atomic_cas<__int128_t>` specialization uses `__atomic_compare_exchange_n` on the fallback tier.
* `src/tscore/CMakeLists.txt`: link `atomic` (PUBLIC) when needed.
* Tests: the old `src/tscore/test_atomic.cc` stress test was orphaned (no build system referenced it). It is replaced with Catch2 tests in `unit_tests/test_InkAtomicList.cc`: a concurrent push/pop/popall conservation test with a double-reachability detector, an `ink_atomiclist_remove` test, and a concurrent `InkFreeList` new/free test.
* Drive-by: remove the dead `INK_QUEUE_NT` conditional (`ink_queue_nt.c` no longer exists) and report the new feature flag in `traffic_layout`.

### Performance

The fallback affects only platforms that had no working build before. A microbenchmark of the freelist access pattern (x86 proxy numbers): the lock-based path is ~1.7x slower than the inline CAS when uncontended (~37 ns vs ~21 ns per op), and equal or slightly faster under contention, because the futex parks waiters instead of burning CAS retries. `ProxyAllocator` thread caches absorb most freelist traffic, so the global head is not on the per-request fast path.

### Testing

* Full build and `test_tscore` pass on x86-64 (`__sync` tier, unchanged codegen).
* The fallback tier was exercised on x86-64 with a forced configuration and a lock-based libatomic stand-in; all stress tests pass.
* Verification on real riscv64 hardware would be very welcome. `traffic_layout info --features` should show `TS_HAS_128BIT_CAS_LIBATOMIC: 1` there.

Fixes: #13555


moonchen (changes requested):
Clang/RISC-V still fails under the default warning-as-error build. With Clang 18, compiling these fallback operations using `--target=riscv64-linux-gnu -march=rv64gc -Werror` produces:

```
error: large atomic operation may incur significant performance penalty; the access size (16 bytes) exceeds the max lock-free size (8 bytes) [-Werror,-Watomic-alignment]
```

The diagnostic is emitted for both `__atomic_load` in `INK_QUEUE_LD` and `__atomic_compare_exchange_n` in the CAS specialization. The CMake probe can succeed because its try-compile does not inherit `CMAKE_COMPILE_WARNING_AS_ERROR`, but the actual C++ build then fails. Please suppress this expected diagnostic for the fallback path or otherwise avoid promoting it to an error.

src/tscore/CMakeLists.txt
```diff
@@ -138,6 +138,14 @@ if(TS_HAS_128BIT_CAS AND TS_NEEDS_MCX16_FOR_CAS)
   target_compile_options(tscore PUBLIC "-mcx16")
 endif()
 
+if(TS_NEEDS_LIBATOMIC_FOR_CAS)
+  target_link_libraries(tscore PUBLIC atomic)
+
+  if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    target_compile_options(tscore PUBLIC "-Wno-error=atomic-alignment")
+  endif()
+endif()
```
moonchen, line 147: `-Wno-error=atomic-alignment` is applied inside the `TS_NEEDS_LIBATOMIC_FOR_CAS` block, but the code that emits the diagnostic -- `INK_QUEUE_LD`'s `__atomic_load` and the `ink_atomic_cas<__int128_t>` specialization -- is gated on `TS_HAS_128BIT_CAS_LIBATOMIC`. The probe sets those two independently: if `CHECK_PROGRAM_ATOMIC` links without an explicit `-latomic`, `USE_LIBATOMIC_CAS` is TRUE while `NEED_LIBATOMIC` stays FALSE, so the `__atomic` path compiles with the diagnostic still fatal and `-Werror` fails again. I couldn't find a toolchain here that actually lands there, so it's latent rather than broken, but keying the flag to the same variable as the code costs nothing.

```suggestion
if(TS_NEEDS_LIBATOMIC_FOR_CAS)
  target_link_libraries(tscore PUBLIC atomic)
endif()

if(TS_HAS_128BIT_CAS_LIBATOMIC AND CMAKE_CXX_COMPILER_ID MATCHES "Clang")
  target_compile_options(tscore PUBLIC "-Wno-error=atomic-alignment")
endif()
```

moonchen (dismissed):
The `-Werror` failure is fixed -- confirmed with clang 18 `--target=riscv64-linux-gnu -march=rv64gc`: both sites error without the flag, warn with it. `test_tscore` passes ASAN-clean including the new `[InkAtomicList]` cases. One optional suggestion inline.

moonchen (approved):
The latest revision addresses the previous diagnostic-gating suggestion. No new blocking findings.

cmcfarlen:
Cherry-picked to the 10.2.x branch as 56b1e10d8513f783ac39fff55e527690ea0fc16c for the 10.2.1 release.

PR #13574, opened 2026-08-19
 Fix TransformVConnection resource leak when abort_tunnel() is called with active request transform

sxia-aviatrix:
## Summary

Fix a resource leak in `HttpSM::state_read_server_response_header()` when
`abort_tunnel()` is called while a request transform plugin registered at
`TS_HTTP_READ_REQUEST_HDR_HOOK` is active.

## Bug

When a POST request has a request transform and the origin server responds
before the full body is forwarded through the transform chain:

1. `state_read_server_response_header()` calls `abort_tunnel()`
2. `abort_tunnel()` cancels I/O on tunnel producers/consumers and calls
   `reset()`, but does **not** close VCs or clean up vc\_table entries
3. `post_transform_info.entry` still references the TransformVConnection
   with `in_tunnel = true`
4. `cleanup_all()` in `kill_this()` calls `cleanup_entry()`, which skips
   `do_io_close()` because `in_tunnel == true`
5. The TransformVConnection and the plugin's transform continuations are
   **never closed** — a resource leak on every affected request

The bug only triggers when the request transform is added before the tunnel
starts (e.g. at `TS_HTTP_READ_REQUEST_HDR_HOOK`). Transforms added at
`TS_HTTP_TUNNEL_START_HOOK` become part of the tunnel chain and are properly
cleaned up by `abort_tunnel()`.

An `ink_release_assert(post_transform_info.entry == nullptr)` placed after
`abort_tunnel()` confirms the stale entry on every request that hits this
path. GDB on the resulting core shows:

```
#7  HttpSM::state_read_server_response_header at HttpSM.cc:2158
post_transform_info = {entry = 0x7f4ee73ba950, vc = 0x5130000202c0}
*post_transform_info.entry = {vc = 0x5130000202c0, vc_type = TRANSFORM_VC,
  in_tunnel = true}
```

## Fix

After `abort_tunnel()`, explicitly close and clean up the orphaned
TransformVConnection:

```cpp
if (post_transform_info.entry != nullptr) {
    post_transform_info.vc->do_io_close();
    vc_table.cleanup_entry(post_transform_info.entry);
    post_transform_info.entry = nullptr;
}
```

This calls `do_io_close()` directly on the transform VC rather than
clearing the `in_tunnel` flag, preserving the ownership semantics that
other call sites rely on. With `in_tunnel == true`, `cleanup_entry()`
skips its own `do_io_close()` and falls through to `remove_entry()`,
so there is no double-close.

`post_transform_info.vc` is left non-null, which correctly tells
`transform_cleanup()` in `kill_this()` that the chain was already closed.

## Test

Added `post_early_response_transform.test.py` with the
`null_transform_request` test plugin. The test sends a partial POST
(`Content-Length: 100000`, sends only small chunks slowly) while the
origin responds immediately. This exercises the `abort_tunnel()` path with
an active request transform, verifying ATS handles the cleanup without
leaking resources.


tests/tools/plugins/null_transform_request.cc
```diff
@@ -0,0 +1,332 @@
...
+  TSPluginRegistrationInfo info;
+
+  info.plugin_name   = PLUGIN_NAME;
+  info.vendor_name   = "Apache Software Foundation";
+  info.support_email = "dev@trafficserver.apache.org";
+
+  if (TSPluginRegister(&info) != TS_SUCCESS) {
+    TSError("[%s] Plugin registration failed", PLUGIN_NAME);
+
+    goto Lerror;
```
JosiahWI: Is it desirable to have two error messages?

sxia-aviatrix: This part is copied from the existing tunnel_transform.cc. I can clean it up if you'd like, but kept it consistent with the original.

JosiahWI: The tunnel_transform.cc plugin looks very similar. What's the difference between this test plugin and that one?

JosiahWI: It's the changes in the PR description. I think that's small enough different conceptually to warrant some amount of de-duplication.

sxia-aviatrix: Good point, I will remove the part we don't need

sxia-aviatrix: I have updated the codes and PR descriptions. Thanks a lot for pointing this out.

src/proxy/http/HttpSM.cc
```diff
@@ -2145,6 +2145,17 @@ HttpSM::state_read_server_response_header(int event, void *data)
...
+      // transform is present, post_transform_info.entry still points at the
+      // TransformVConnection whose chain will be freed by the abort cascade.
+      // Clean it up now so cleanup_all() in kill_this() does not call
+      // do_io_close() on freed memory.
+      if (post_transform_info.entry != nullptr) {
+        post_transform_info.entry->in_tunnel = false;
+        vc_table.cleanup_entry(post_transform_info.entry);
+        post_transform_info.entry = nullptr;
+      }
+      // ink_release_assert(post_transform_info.entry == nullptr);
```
bneradt: Commented out code.

bneradt (changes requested):
Thanks for the fix.

Please reorganize tests/gold_tests/slow_post/post_early_response_transform.test.py as a Test class. See tests/gold_tests/ats_probe/ats_probe.test.py as an example.

src/proxy/http/HttpSM.cc
sxia-aviatrix: Thanks for the comments. I have commented this out and update the test cases to use test class

bneradt (commented):
Thanks for digging into this one — the teardown gap you found looks real. My main concern is that the mechanism described in the PR (and in the code comment and the test docstring) doesn't hold up: `cleanup_entry()` only calls `do_io_close()` when `in_tunnel == false`, and at this point it is `true`, so `cleanup_all()` can't have been dereferencing the stale VC. Details inline.

I do think there's a genuine bug here, just a different one: before this patch the TransformVConnection and the plugin's transform continuations are never closed on the abort path — a leak rather than a use-after-free. If that's what you were chasing, the fix is close to right, but it should say so directly instead of overwriting the `in_tunnel` ownership flag. If there really is a crash, could you attach the stack trace or ASAN report so we can confirm this addresses it?

The rest of the comments are on the test and the test plugin. Also flagging that one of the Copilot comments below is incorrect — replied in that thread.

src/proxy/http/HttpSM.cc
```diff
@@ -2145,6 +2145,17 @@ HttpSM::state_read_server_response_header(int event, void *data)
     // If there is a post body in transit, give up on it
     if (tunnel.is_tunnel_alive()) {
       tunnel.abort_tunnel();
+      // abort_tunnel() does not clean up vc_table entries.  If a request
+      // transform is present, post_transform_info.entry still points at the
+      // TransformVConnection whose chain will be freed by the abort cascade.
+      // Clean it up now so cleanup_all() in kill_this() does not call
+      // do_io_close() on freed memory.
```
bneradt: This comment doesn't match what the code below actually does, and I don't think the crash it describes can happen.

`HttpVCTable::cleanup_entry()` only calls `do_io_close()` when `in_tunnel == false`:

```cpp
void
HttpVCTable::cleanup_entry(HttpVCTableEntry *e)
{
  ink_assert(e->vc);
  if (e->in_tunnel == false) {
    ...
    e->vc->do_io_close();
    e->vc = nullptr;
  }
  remove_entry(e);
}
```

At this point `post_transform_info.entry->in_tunnel` is `true`. It is set in `do_setup_client_request_body_tunnel()` (`case HttpVC_t::TRANSFORM_VC`) and again in `setup_transform_to_server_transfer()`, and it is cleared only in `tunnel_handler_transform_write()` — which `abort_tunnel()` never invokes, since `abort_tunnel()` doesn't call any consumer/producer handlers. So `cleanup_all()` in `kill_this()` falls straight through to `remove_entry()`, which never dereferences `e->vc`. There is no `do_io_close()` on freed memory to prevent.

The two cases:

* `in_tunnel == true` — the real case, and this patch having to clear it is the proof: `cleanup_all()` never touched `e->vc`, so the described use-after-free cannot have occurred.
* `in_tunnel == false` — hypothetically: the assignment is a no-op, and `cleanup_entry()` performs the very same `do_io_close()` on the very same pointer, just earlier in the call chain. Also not a fix.

That said, I think you have found a real bug, just a different one. `abort_tunnel()` only issues `do_io_read/do_io_write(this, 0, nullptr)` on the producers/consumers and then `reset()`; it never closes the transform chain. `kill_tunnel()` afterwards finds an empty tunnel, and `transform_cleanup()` skips the chain because `post_transform_info.vc != nullptr`. So today `TransformVConnection::m_closed` is never set, `TransformTerminus::handle_event()` never reaches `delete m_tvc`, and the TransformVConnection plus the plugin's transform continuations **leak**. Forcing a close here does fix that.

So this looks like a fix for a leak / missing teardown rather than for a use-after-free. Could you attach the actual stack trace or ASAN report? If there genuinely is a crash, I want to be sure this isn't masking it rather than fixing it. And either way the comment, the PR description, and the test docstring should be rewritten around the real mechanism.

src/proxy/http/HttpSM.cc
```diff
@@ -2145,6 +2145,17 @@ HttpSM::state_read_server_response_header(int event, void *data)
...
+      if (post_transform_info.entry != nullptr) {
+        post_transform_info.entry->in_tunnel = false;
```
bneradt: Overwriting `in_tunnel` to coerce a side effect out of `cleanup_entry()` is a blunt instrument, and it works against the rest of the file. Every other post-transform cleanup site deliberately preserves this flag:

* `tunnel_handler_post_or_put()` asserts `post_transform_info.entry->in_tunnel == true` and then calls `cleanup_entry()` *specifically so that it will not close*, with a comment explaining why.
* `state_common_wait_for_transform_read()` and `handle_server_setup_error()` likewise call `cleanup_entry()` without touching the flag.

`in_tunnel` encodes who owns the VC — the tunnel or the SM. If the intent is "close the transform chain now", it reads much better to say that directly:

```cpp
post_transform_info.vc->do_io_close();
vc_table.cleanup_entry(post_transform_info.entry);
post_transform_info.entry = nullptr;
```

There's a robustness cost too. The `in_tunnel == true` check is currently what protects paths where the TVC has already been closed but the entry is still populated — `tunnel_handler_transform_write()`'s `VC_EVENT_ERROR` case does `c->vc->do_io_close(EHTTP_ERROR)` and leaves the entry alone. Clearing the flag unconditionally removes that protection and turns any such path into a genuine use-after-free.

Last thought on placement: `abort_tunnel()` is the function that leaves the vc_table inconsistent, and `kill_tunnel()` has the same shape. Patching this one caller leaves the trap set for the next person. A small SM helper (or handling it on the abort path itself) would cover both.

src/proxy/http/HttpSM.cc
```diff
@@ -2145,6 +2145,17 @@ HttpSM::state_read_server_response_header(int event, void *data)
...
+        vc_table.cleanup_entry(post_transform_info.entry);
+        post_transform_info.entry = nullptr;
```
bneradt, line 2156: Nulling `entry` while leaving `post_transform_info.vc` non-null breaks the pairing that two sites rely on — they check `.vc` and then dereference `.entry`:

* `tunnel_handler_post_or_put()`: `if (post_transform_info.vc != nullptr) { ink_assert(post_transform_info.entry->in_tunnel == true); ...; vc_table.cleanup_entry(post_transform_info.entry); }`
* `handle_server_setup_error()`: `if (post_transform_info.vc) { ...; vc_table.cleanup_entry(post_transform_info.entry); }`

The `handle_server_setup_error()` one is safe here only incidentally: `abort_tunnel()` already called `reset()`, so `tunnel.get_consumer(post_transform_info.vc)` returns nullptr and the inner guard short-circuits. Worth confirming `tunnel_handler_post_or_put()` can't be reached after this abort — in a release build the `ink_assert`s compile out and `cleanup_entry(nullptr)` faults on `e->in_tunnel`.

`state_common_wait_for_transform_read()` already produces this `entry == nullptr && vc != nullptr` state in the TRANSFORM_FAIL case, so it may well be fine. I'd just rather see it checked than assumed.

src/proxy/http/HttpSM.cc
```diff
@@ -2145,6 +2145,17 @@ HttpSM::state_read_server_response_header(int event, void *data)
...
+      }
+      ink_release_assert(post_transform_info.entry == nullptr);
```
bneradt: This assert is tautological: the block immediately above sets `post_transform_info.entry = nullptr`, and if the `if` wasn't taken it was already nullptr. It can never fire. Looks like leftover repro scaffolding — please drop it.

Relatedly, the PR description still says to "uncomment the assert after `abort_tunnel()`" to reproduce, which no longer matches the diff.

tests/gold_tests/slow_post/post_early_response_transform.test.py
```diff
@@ -0,0 +1,112 @@
+"""Verify ATS does not crash when a server replies before receiving the full POST body and a request transform plugin is active.
+
+When a POST request has a request transform and the origin responds before the
+full body is forwarded through the transform chain, abort_tunnel() is called.
+Without the fix, post_transform_info.entry is left stale in the vc_table,
+causing a use-after-free in cleanup_all().
```
bneradt: Two things.

First, this docstring restates the "use-after-free in `cleanup_all()`" mechanism, which I don't think holds — see my comment on `HttpSM.cc`. Whatever the real root cause turns out to be, this needs to match it.

Second, and more important: **does this test fail on master?** AuTest would catch a genuine traffic_server crash — `MakeATSProcess` sets `p.ReturnCode = 0` on the ATS process and adds an `ExcludesExpression("FATAL:")` tester on diags.log — so a real SIGSEGV/SIGABRT does fail the run. But if the pre-patch defect is a leak rather than a crash (which is what I believe it is), this test passes with and without the fix and isn't a regression test at all. Could you post the failing output from master?

I also don't see any CI runs on this branch yet. Worth getting Jenkins green before this goes further.

tests/gold_tests/slow_post/post_early_response_transform.test.py
```diff
@@ -0,0 +1,112 @@
...
+#  limitations under the License.
+
+import os
+from ports import get_port
+import sys
+
+Test.Summary = __doc__
+
+
+class PostEarlyResponseTransformTest:
```
bneradt: This class re-implements `QuickServerTest` in `quick_server.test.py` almost line for line: same DNS server, same `quick_server.py` origin, same remap line, the same four `records.yaml` keys, the same `StartBefore` chain, the same `tr.Timeout = 10`. That file already parameterizes over three booleans.

Adding a `use_request_transform` parameter there — installing the plugin and swapping in the partial-POST client — would fold this in without a second copy of the scaffolding.

tests/gold_tests/slow_post/post_early_response_transform.test.py
```diff
@@ -0,0 +1,112 @@
...
+    """Verify abort_tunnel with a request transform does not crash ATS."""
+
+    _partial_post_client = 'partial_post_client.py'
+    _quick_server = 'quick_server.py'
+    _init_file = '__init__.py'
+    _http_utils = 'http_utils.py'
```
bneradt: +1 to Copilot on this one: `_http_utils` is never used — line 94 rebuilds the path inline. Please drop the class attribute.

tests/gold_tests/slow_post/post_early_response_transform.test.py
```diff
@@ -0,0 +1,112 @@
...
+        tr.Setup.CopyAs(self._init_file, Test.RunDirectory)
+        tr.Setup.CopyAs(http_utils, Test.RunDirectory)
+        tr.Setup.CopyAs(self._quick_server, Test.RunDirectory)
+        tr.Setup.CopyAs(self._partial_post_client, Test.RunDirectory)
+
+        p = tr.Processes.Default
+        p.Command = (f'{sys.executable} {self._partial_post_client} '
+                     f'127.0.0.1 {self._ts.Variables.port}')
+        p.ReturnCode = 0
+        p.Streams.All += Testers.ContainsExpression('Got response', 'Verify client received a response from ATS')
```
bneradt: Both of these assertions are vacuous:

* The client's `main()` unconditionally does `return 0` on every path, including the `socket.timeout` and `ConnectionError` handlers, so `p.ReturnCode = 0` cannot fail.
* `ContainsExpression('Got response', ...)` matches `Got response: timeout (server may still be processing)` and `Got response: connection closed` just as happily as a real response.

So the test currently rests entirely on the implicit "ATS didn't crash" check. Please assert on the actual status line the proxy is expected to return, and have the client exit non-zero when it doesn't get one — compare `quick_server.test.py`, which checks for `HTTP/1.1 200 OK` explicitly.

tests/gold_tests/slow_post/partial_post_client.py
```diff
@@ -0,0 +1,69 @@
...
+    """Run the client."""
+    host = sys.argv[1] if len(sys.argv) > 1 else '127.0.0.1'
+    port = int(sys.argv[2]) if len(sys.argv) > 2 else 8080
+
+    request = (
+        f'POST / HTTP/1.1\r\n'
+        f'Host: quick.server.com\r\n'
+        f'Content-Type: application/octet-stream\r\n'
+        f'Content-Length: 100000\r\n'
+        f'\r\n').encode()
```
bneradt: Minor: none of these `f` prefixes have placeholders, so plain string literals would do (yapf/flake8 will likely flag them).

More generally, this overlaps `slow_post_client.py` in the same directory, which already does the "send a POST and don't finish it" dance. The difference that matters is `Content-Length` vs. `Transfer-Encoding: chunked`, so a `--content-length N --send-bytes M` mode there would avoid a third client script in this directory. Your call — the mechanism is different enough that a separate file is defensible.

tests/tools/plugins/null_transform_request.cc
```diff
@@ -0,0 +1,162 @@
...
+
+#include "ts/ts.h"
+
+#define PLUGIN_NAME "null_transform_request"
+
+typedef struct {
+  TSVIO            output_vio;
+  TSIOBuffer       output_buffer;
+  TSIOBufferReader output_reader;
+} TransformData;
```
bneradt: `typedef struct { ... } TransformData;` is a C idiom; in C++ this is just `struct TransformData { ... };`.

Also `<cstdio>` and `<cinttypes>` (lines 29-30) are unused now that the `Dbg()` calls are gone.

tests/tools/plugins/null_transform_request.cc
```diff
@@ -0,0 +1,162 @@
...
+    break;
+  }
+
+  return 0;
+}
+
+static int
+transform_plugin(TSCont /* contp ATS_UNUSED */, TSEvent event, void *edata)
+{
+  if (event == TS_EVENT_HTTP_READ_REQUEST_HDR) {
```
bneradt: Following up on the de-duplication thread above: after the trimming, the only remaining difference from `tunnel_transform.cc` is this hook point (plus not adding the response transform). The ~70-line transform body is a verbatim copy.

`PrepareTestPlugin` already forwards `plugin_args` into `plugin.config`, so `tunnel_transform` could take an argument selecting `TS_HTTP_READ_REQUEST_HDR_HOOK` vs. `TS_HTTP_TUNNEL_START_HOOK` (and whether to add the response transform), and this file could go away entirely. That's a better outcome than two copies of the same null transform to keep in sync.

tests/tools/plugins/null_transform_request.cc
```diff
@@ -0,0 +1,162 @@
...
+    TSVIONBytesSet(data->output_vio, TSVIONDoneGet(input_vio));
+    TSVIOReenable(data->output_vio);
+    TSContCall(TSVIOContGet(input_vio), TS_EVENT_VCONN_WRITE_COMPLETE, input_vio);
+  }
+}
+
+static int
+null_transform(TSCont contp, TSEvent event, void * /* edata ATS_UNUSED */)
+{
+  if (TSVConnClosedGet(contp)) {
```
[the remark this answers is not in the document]
bneradt: This bot comment is incorrect — please don't act on it. `TSTransformCreate()` returns a `TSVConn` that *is* the continuation, so passing it to `TSVConnClosedGet()` is the standard idiom for every transform plugin in the tree: see `tests/tools/plugins/tunnel_transform.cc` and `example/plugins/c-api/null_transform/null_transform.cc`, which both do exactly this. The code here is correct as written.

src/proxy/http/HttpSM.cc
sxia-aviatrix: Thanks for the review. I think you are right. This is a leak instead of use-after-free. 
Unfortunately, since I think this is a resource leak not memory leak, so I cannot find anything with ASAN or LSAN.

sxia-aviatrix: I do have gdb stack trace and hope this can be helpful: 
```
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `/tmp/sb_pr4/post_early_response_transform/ts/bin/traffic_server --bind_stderr /'.
Program terminated with signal SIGABRT, Aborted.
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44

warning: 44     ./nptl/pthread_kill.c: No such file or directory
[Current thread is 1 (Thread 0x7f4ef591e6c0 (LWP 287043))]
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007f4f014a927e in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007f4f0148c8ff in __GI_abort () at ./stdlib/abort.c:79
#5  0x0000632477666236 in ink_abort (message_format=message_format@entry=0x6324787af000 "%s:%d: failed assertion `%s`") at /src/trafficserver_pr/trafficserver/src/tscore/ink_error.cc:99
#6  0x000063247765f9b6 in _ink_assert (expression=expression@entry=0x6324787d9200 "post_transform_info.entry == nullptr", file=file@entry=0x6324787d3480 "/src/trafficserver_pr/trafficserver/src/proxy/http/HttpSM.cc", line=line@entry=2158) at /src/trafficserver_pr/trafficserver/src/tscore/ink_assert.cc:35
#7  0x00006324777cd115 in HttpSM::state_read_server_response_header (this=0x7f4ee73b8000, event=<optimized out>, data=<optimized out>) at /src/trafficserver_pr/trafficserver/src/proxy/http/HttpSM.cc:2158
#8  0x00006324777deec7 in HttpSM::main_handler (this=0x7f4ee73b8000, event=<optimized out>, data=<optimized out>) at /src/trafficserver_pr/trafficserver/src/proxy/http/HttpSM.cc:2880
#9  0x000063247823737a in Continuation::handleEvent (data=0x7f4ee749b5c8, event=100, this=0x7f4ee73b8000) at /src/trafficserver_pr/trafficserver/include/iocore/eventsystem/Continuation.h:228
#10 Continuation::handleEvent (data=0x7f4ee749b5c8, event=100, this=0x7f4ee73b8000) at /src/trafficserver_pr/trafficserver/include/iocore/eventsystem/Continuation.h:224
#11 read_signal_and_update (event=100, vc=0x7f4ee749b1b0) at /src/trafficserver_pr/trafficserver/src/iocore/net/UnixNetVConnection.cc:86
#12 0x00006324782488b8 in UnixNetVConnection::net_read_io (this=0x7f4ee749b1b0, nh=<optimized out>) at /src/trafficserver_pr/trafficserver/src/iocore/net/UnixNetVConnection.cc:600
#13 0x00006324782f51c4 in NetHandler::process_ready_list (this=this@entry=0x7f4efa3a43a0) at /src/trafficserver_pr/trafficserver/src/iocore/net/NetHandler.cc:265
#14 0x00006324782f5da6 in NetHandler::waitForActivity (this=0x7f4efa3a43a0, timeout=<optimized out>) at /src/trafficserver_pr/trafficserver/src/iocore/net/NetHandler.cc:360
#15 0x0000632478443b90 in EThread::execute_regular (this=this@entry=0x7f4efa3a3800) at /src/trafficserver_pr/trafficserver/src/iocore/eventsystem/UnixEThread.cc:326
#16 0x0000632478444552 in EThread::execute (this=0x7f4efa3a3800) at /src/trafficserver_pr/trafficserver/src/iocore/eventsystem/UnixEThread.cc:383
#17 EThread::execute (this=0x7f4efa3a3800) at /src/trafficserver_pr/trafficserver/src/iocore/eventsystem/UnixEThread.cc:360
#18 0x000063247843876e in spawn_thread_internal (a=a@entry=0x50600000faa0) at /src/trafficserver_pr/trafficserver/src/iocore/eventsystem/Thread.cc:75
#19 0x00007f4f027b1a42 in asan_thread_start (arg=0x7f4f0074f000) at ../../../../src/libsanitizer/asan/asan_interceptors.cpp:234
#20 0x00007f4f01500aa4 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:447
#21 0x00007f4f0158dc6c in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78
#7  0x00006324777cd115 in HttpSM::state_read_server_response_header (this=0x7f4ee73b8000, event=<optimized out>, data=<optimized out>) at /src/trafficserver_pr/trafficserver/src/proxy/http/HttpSM.cc:2158
2158          ink_release_assert(post_transform_info.entry == nullptr);
$1 = {entry = 0x7f4ee73ba950, vc = 0x5130000202c0}
$2 = {vc = 0x5130000202c0, read_buffer = 0x0, write_buffer = 0x0, read_vio = 0x0, write_vio = 0x0, vc_read_handler = NULL, vc_write_handler = NULL, vc_type = HttpVC_t::TRANSFORM_VC, sm = 0x7f4ee73b8000, eos = false, in_tunnel = true}
[ats-develop@b07c50b9e0d7 trafficserver ]$ 
```

sxia-aviatrix: The test and stack trace are got in `upstream/master`.

sxia-aviatrix: I can reproduce the ATS crash with a proprietary plugin under stress, but I can't share that stack trace. This fix can resolve the crash. I will keep current null_transform_request plugin and assert as the way to reproduce the bug. Please let me know if this is not sufficient or anything else we need other than assert and gdb.

sxia-aviatrix: Have updated the title and description

src/proxy/http/HttpSM.cc
sxia-aviatrix: This makes sense. I have use explicit free for vc.

src/proxy/http/HttpSM.cc
sxia-aviatrix: Good catch on the pairing. I traced both sites:

**`tunnel_handler_post_or_put()`** — Cannot be reached after this `abort_tunnel()`. It fires via `HTTP_TUNNEL_EVENT_DONE`, but `abort_tunnel()` sets `active = false`, marks all producers/consumers dead, and calls `reset()`. The tunnel will never deliver that callback again.

**`handle_server_setup_error()`** — Safe as you noted: `abort_tunnel()` calls `reset()`, so `tunnel.get_consumer(post_transform_info.vc)` returns `nullptr` and the inner guard short-circuits before touching `.entry`.

The `entry == nullptr && vc != nullptr` state also matches the existing pattern at line 2929 (`tunnel_handler_post_or_put` itself does this) and in `state_common_wait_for_transform_read()`'s `TRANSFORM_FAIL` case.

So I'll keep `.vc` non-null to stay consistent with the rest of the file.

src/proxy/http/HttpSM.cc
sxia-aviatrix: You're right, removed. The assert was repro scaffolding used to confirm the stale entry via GDB — it's not part of the fix.

tests/gold_tests/slow_post/post_early_response_transform.test.py
sxia-aviatrix: I have removed this.

tests/gold_tests/slow_post/post_early_response_transform.test.py
sxia-aviatrix: That's a good point. I have merge the plugin and test to tunnel transform plugin and quick server test

tests/gold_tests/slow_post/post_early_response_transform.test.py
sxia-aviatrix: I have updated the code to check for `HTTP/1.1 200 OK`

tests/gold_tests/slow_post/partial_post_client.py
sxia-aviatrix: I have updated the `f` prefixes and kept this file

tests/tools/plugins/null_transform_request.cc
sxia-aviatrix: I have merged null transform request plugin to tunnel transform plugin.

tests/tools/plugins/null_transform_request.cc
sxia-aviatrix: This is a good idea. I have added the argument and done the deduplication.

tests/gold_tests/slow_post/post_early_response_transform.test.py
sxia-aviatrix: I have updated the comments.

sxia-aviatrix: Since the problem is resource leak, I don't think AuTest can fail without assert. I have shared the gdb stack trace from master above.

shinrich:
[approve ci]

shinrich:
[ approve ci ]

bneradt:
[approve ci]

bneradt:
Thank you for the updates. I'm going to build a dev rpm of this and try it in production to make sure it's stable for us. If that goes well I'll approve.

bneradt (commented):
This is much better — thanks for turning it around so quickly. The `HttpSM.cc` change now says what it does and does what it says, folding the test into `quick_server.test.py` came out cleaner than I expected, and teaching `tunnel_transform.cc` a mode flag beats a second copy of the plugin. The C++ fix itself looks right to me.

One item I'd call blocking, and it's in the test: the client's connection-reset message is worded so that it satisfies the `ContainsExpression('HTTP/1.1')` check, which means the tester labelled "Verify client received an HTTP response" passes even when no response arrived. Worth nailing down what ATS is actually expected to return here.

Everything else is smaller — one nit on the fix, a note that the `nbytes` change in the plugin is load-bearing and should be commented (I traced why: `INT64_MAX` would make `state_request_wait_for_transform_read()` fail the transaction outright, so without it the new run wouldn't exercise the abort path at all), and a flag that a leak regression won't be caught by this test outside ASAN.

src/proxy/http/HttpSM.cc
```diff
@@ -2145,6 +2145,16 @@ HttpSM::state_read_server_response_header(int event, void *data)
     // If there is a post body in transit, give up on it
     if (tunnel.is_tunnel_alive()) {
       tunnel.abort_tunnel();
+      // abort_tunnel() cancels I/O but does not close VCs or clean up
+      // vc_table entries.  When a request transform is active the
+      // post_transform_info entry still references the TransformVConnection
+      // with in_tunnel=true, which causes cleanup_entry() to skip
+      // do_io_close() — leaking the VC.  Close it explicitly here.
+      if (post_transform_info.entry != nullptr) {
+        post_transform_info.vc->do_io_close();
```
bneradt: This is the shape I was hoping for — the comment now describes the actual defect, and the close is explicit rather than a side effect of clearing `in_tunnel`.

One nit: the guard tests `post_transform_info.entry`, but this line dereferences `post_transform_info.vc`. The invariant does hold today — `do_post_transform_open()` only creates the entry when `vc` is non-null and sets `entry->vc = vc`, and `state_common_wait_for_transform_read()` nulls both together — but `post_transform_info.entry->vc->do_io_close()` is the more direct expression, and it's the very pointer `cleanup_entry()` asserts on (`ink_assert(e->vc)`) one line later. Either that, or guard on `.vc` the way the other sites in this file do.

For the record on the other direction: `TransformVConnection::do_io_close()` early-returns on `m_closed != 0`, so a double close here is harmless.

src/proxy/http/HttpSM.cc
```diff
@@ -2145,6 +2145,16 @@ HttpSM::state_read_server_response_header(int event, void *data)
...
+        vc_table.cleanup_entry(post_transform_info.entry);
+        post_transform_info.entry = nullptr;
```
bneradt, line 2156: Still open from the last round — flagging once rather than re-explaining. This leaves `post_transform_info.vc` non-null with `entry == nullptr`, and `tunnel_handler_post_or_put()` checks only `.vc` before dereferencing `.entry`.

Leaving `.vc` set is deliberate and correct (it's what makes `transform_cleanup()` skip the chain), so the only question is whether `tunnel_handler_post_or_put()` is reachable after this abort. If you've satisfied yourself that it isn't, a one-line comment saying so would save the next reader the trip.

tests/gold_tests/slow_post/quick_server.test.py
```diff
@@ -105,22 +116,30 @@ def run(self):
...
-        p.Command = client_command
-        if self._should_abort_request or self._should_abort_response_headers:
-            p.Streams.All += Testers.ExcludesExpression('HTTP/1.1 200 OK', 'Verify response was received')
+        if self._use_request_transform:
+            tr.Setup.CopyAs(self._partial_post_client, Test.RunDirectory)
+            p = tr.Processes.Default
+            p.Command = (f'{sys.executable} {self._partial_post_client} '
+                         f'127.0.0.1 {self._ts.Variables.port}')
+            p.ReturnCode = 0
+            p.Streams.All += Testers.ContainsExpression('HTTP/1.1', 'Verify client received an HTTP response')
```
bneradt: This is the one item I'd hold the PR on.

The client's reset path prints `HTTP/1.1 connection reset (expected for partial POST)` (`partial_post_client.py:56`), worded such that it satisfies `ContainsExpression('HTTP/1.1')`. So the tester labelled "Verify client received an HTTP response" passes when no response was received at all — and `p.ReturnCode = 0` passes too, because that path returns 0. Someone scanning this file sees a real check where there isn't one.

Please pin down what ATS is actually expected to do here. The origin sends a complete `HTTP/1.1 200 OK` / `Content-Length: 0` before the body finishes, and `state_read_server_response_header()` sets `NO_KEEPALIVE` on both sides after the abort — so I'd expect the 200 to be forwarded and then the connection closed, i.e. `Testers.ContainsExpression('HTTP/1.1 200 OK', ...)` deterministically, matching the other runs in this file.

If it genuinely races between "200 then FIN" and "RST", then say so explicitly: print something that isn't shaped like a status line, assert on it separately, and comment the race. As written, the message and the assertion are engineered to agree with each other regardless of what ATS did.

tests/gold_tests/slow_post/partial_post_client.py
```diff
@@ -0,0 +1,77 @@
...
+        print(f'Sent POST headers (Content-Length: 100000) + {len(partial_body)} bytes')
+
+        try:
+            response = sock.recv(4096)
+        except ConnectionError:
+            # ATS may reset the connection after responding since the POST
+            # body is incomplete.  This is acceptable — the important thing
+            # is that ATS did not crash.
+            print('HTTP/1.1 connection reset (expected for partial POST)')
+            return 0
```
bneradt, line 54: Same point as on the test, from this side: `'HTTP/1.1 connection reset ...'` is doing double duty as a human-readable log line and as the token that `ContainsExpression('HTTP/1.1')` matches over in `quick_server.test.py`. If a reset really is an acceptable outcome, print something that can't be mistaken for a status line (`CONNECTION RESET`) and give the test a tester for that string specifically.

Also worth checking: `except ConnectionError` won't catch a clean FIN — that surfaces as `recv()` returning `b''`, which falls through to the `return 1` branch below. Since ATS sets `NO_KEEPALIVE` on this path, a clean close after the response is the more likely outcome, so make sure the branch you expect to hit is the one you're actually asserting on.

tests/gold_tests/slow_post/quick_server.test.py
```diff
@@ -132,3 +151,7 @@ def run(self):
         for abort_response_headers in [True, False]:
             test = QuickServerTest(abort_request, drain_request, abort_response_headers)
             test.run()
+
+# Partial POST with a request transform plugin: exercises the abort_tunnel()
+# cleanup path for TransformVConnection entries in the vc_table.
+QuickServerTest(abort_request=True, drain_request=False, abort_response_headers=False, use_request_transform=True).run()
```
bneradt: Two things on this line.

`abort_request=True` is inert for the transform run — nothing reads `_should_abort_request` in the `use_request_transform` branch — yet the generated run name will still print "Aborting request: True". Pass `False`, or leave the flags that don't apply out of the name.

More important: now that the root cause is correctly identified as a leaked `TransformVConnection` rather than a use-after-free, this run can't catch a regression on its own — a leak doesn't fail an autest. It only has teeth under ASAN/LSAN. Worth saying that in the comment above, and worth confirming the ASAN autest job actually runs `slow_post`; otherwise this is a "doesn't crash" smoke test and the leak could come back unnoticed.

tests/gold_tests/slow_post/quick_server.test.py
```diff
@@ -27,24 +31,27 @@ class QuickServerTest:
...
     _slow_post_client = 'slow_post_client.py'
+    _partial_post_client = 'partial_post_client.py'
     _quick_server = 'quick_server.py'
 
     _dns_counter = 0
     _server_counter = 0
     _ts_counter = 0
 
-    def __init__(self, abort_request: bool, drain_request: bool, abort_response_headers: bool):
+    def __init__(self, abort_request: bool, drain_request: bool, abort_response_headers: bool, use_request_transform: bool = False):
```
bneradt, line 42: This signature is exactly 132 characters, right at `column_limit` in `.style.yapf`. Please run `cmake --build build -t format` so we know yapf agrees rather than relying on the boundary — it may well want to wrap it.

tests/tools/plugins/tunnel_transform.cc
```diff
@@ -99,7 +107,8 @@ handle_transform(TSCont contp, bool forward)
     data->output_buffer = TSIOBufferCreate();
     data->output_reader = TSIOBufferReaderAlloc(data->output_buffer);
     Dbg(plugin_ctl, "\tWriting %" PRId64 " bytes on VConn", TSVIONBytesGet(input_vio));
-    data->output_vio = TSVConnWrite(output_conn, contp, data->output_reader, INT64_MAX);
+    int64_t nbytes   = (request_hdr_mode && TSVIONBytesGet(input_vio) > 0) ? TSVIONBytesGet(input_vio) : INT64_MAX;
```
bneradt, line 112: This is load-bearing and needs a comment, because it reads like an incidental tweak and someone will eventually "simplify" it back.

`TSVConnWrite()`'s `nbytes` becomes the terminus write VIO's `nbytes`, which is what `TransformTerminus::handle_event()` hands the SM as the `TRANSFORM_READ_READY` payload. `state_request_wait_for_transform_read()` then does:

```cpp
size = *(static_cast<int64_t *>(data));
if (size != INT64_MAX && size >= 0) {
  t_state.hdr_info.transform_request_cl = size;
  ...
} else {
  // No content length from the post.  This is a no go
  event = VC_EVENT_ERROR;
  Log::error("Request transformation failed to set content length");
}
```

So with the original `INT64_MAX`, the request-transform path fails the transaction before the post tunnel is ever set up, and the new test run would never reach `abort_tunnel()` at all. Something like "a request transform must report a real content length — INT64_MAX makes `state_request_wait_for_transform_read()` fail the transaction" would make that clear.

Good that the default mode still passes `INT64_MAX`: `tests/gold_tests/tunnel/tunnel_transform.test.py` loads this plugin with no arguments, so that path is unchanged.

tests/tools/plugins/tunnel_transform.cc
```diff
@@ -280,8 +291,9 @@ transform_plugin(TSCont /* contp ATS_UNUSED */, TSEvent event, void *edata)
 
   Dbg(plugin_ctl, "Entering transform_plugin()");
   switch (event) {
+  case TS_EVENT_HTTP_READ_REQUEST_HDR:
   case TS_EVENT_HTTP_TUNNEL_START:
-    Dbg(plugin_ctl, "\tEvent is TS_EVENT_HTTP_TUNNEL_START");
+    Dbg(plugin_ctl, "\tEvent is %d", event);
```
bneradt: Minor: this drops the readable event name for both cases, and the two modes are now the main thing you'd be using this plugin's debug output to distinguish. Keeping them apart is worth more than sharing the line:

```cpp
Dbg(plugin_ctl, "\tEvent is %s",
    event == TS_EVENT_HTTP_TUNNEL_START ? "TS_EVENT_HTTP_TUNNEL_START" : "TS_EVENT_HTTP_READ_REQUEST_HDR");
```

tests/gold_tests/slow_post/quick_server.test.py
bneradt: we enforce this check in ci

src/proxy/http/HttpSM.cc
sxia-aviatrix: Yeah, `entry->vc` is clear and I have updated the code.

src/proxy/http/HttpSM.cc
sxia-aviatrix: I have double checked that `tunnel_handler_post_or_put()` is not reachable and added the comments

tests/gold_tests/slow_post/quick_server.test.py
sxia-aviatrix: I have updated the code and comments. But I don't think ASAN/LSAN can catch this, since this is resource leak and vc pointer is reachable. I also tried ASAN/LSAN myself and confirmed that they cannot fail the tests.

tests/gold_tests/slow_post/partial_post_client.py
sxia-aviatrix: I have updated the code

tests/gold_tests/slow_post/quick_server.test.py
sxia-aviatrix: I have updated the code for this.

tests/gold_tests/slow_post/quick_server.test.py
sxia-aviatrix: I have run the format check and it seems to be fine.

tests/tools/plugins/tunnel_transform.cc
sxia-aviatrix: I have updated the debug msg

tests/tools/plugins/tunnel_transform.cc
sxia-aviatrix: I have added the comment

bneradt (commented):
Re-reviewed 6c5ccee1bd870a74d91ef7fc1e2b6dbad32d560f. No new blocking findings; the code looks ready from this review.

The earlier blocking test issue is fixed: the run requires an actual HTTP/1.1 200 OK, and reset/EOF messages can no longer satisfy it. The explicit transform close, entry cleanup, and deliberate retention of the .vc marker agree with the abort/reset and transform_cleanup paths. The requested test/plugin consolidation and explanatory comments are also in place.

All 15 reported CI checks are successful. I additionally checked Python syntax and exercised the client's output/return-code checks with simulated 200, reset, EOF, timeout, and 502 outcomes. I did not rebuild ATS or rerun AuTest locally. The new run remains a smoke test for the early-response path, not a direct assertion that the transform was destroyed. A destruction counter/assertion would be useful follow-up coverage.

This assessment does not verify the production trial mentioned in my earlier comment.

bneradt (approved):
I tested this on a production box and it seems to run fine there, judging from our stats.

Approving. Thank you for working on this.

sxia-aviatrix:
> I tested this on a production box and it seems to run fine there, judging from our stats.
> 
> Approving. Thank you for working on this.

Great to hear that it works well in production. It's my pleasure to contribute to this project. But I don't think I have merge access. Could you please help merge this PR or grant me the access? Thanks!

cmcfarlen:
Cherry-picked to the 10.2.x branch as 4de6948efb568e59e6b2d9be094b74d79e280520 for the 10.2.1 release.

PR #13577, opened 2026-08-20
Fix silently dropped Streams.all autest assertions

bneradt:
AuTest's `Streams` entity defines `All`, not `all`, and does not override `__setattr__`, so every `Streams.all = ...` assignment created a dead instance attribute rather than registering a tester. Twelve assertions across seven gold tests had never run, and two of the expectations behind them had rotted unnoticed as a result.

This renames those assignments to `Streams.All` so the assertions register, and corrects the two expectations they exposed: `client_default_inactivity_timeout.gold` opened with three backticks rather than the two that mark an AuTest wildcard, and `tls_hooks_client_verify` left `proxy.config.ssl.client.verify.server.policy` at the default `ENFORCED` while its origin serves a self-signed cert.

`cert_update.test.py` has two more of these, left alone because #13576 is already rewriting that file.

### DIFF
--- a/tests/gold_tests/h3/h3_sni_check.test.py
+++ b/tests/gold_tests/h3/h3_sni_check.test.py
@@ -119,7 +119,7 @@ ssl_multicert:
                 "SNI not found", "ATS should see the SNI presented by client.")
 
         if self.gold_file:
-            tr.Processes.Default.Streams.all = self.gold_file
+            tr.Processes.Default.Streams.All = self.gold_file
 
 
 # TEST 1: Client request with SNI.
--- a/tests/gold_tests/ip_allow/ip_allow.test.py
+++ b/tests/gold_tests/ip_allow/ip_allow.test.py
@@ -310,7 +310,7 @@ ssl_multicert:
                 "client.*allowed by ip-allow policy", "Request should be allowed by ip_allow")
 
         if self.gold_file:
-            tr.Processes.Default.Streams.all = self.gold_file
+            tr.Processes.Default.Streams.All = self.gold_file
 
 
 # ip_allow tests for h3.
--- a/tests/gold_tests/timeout/default_inactivity_timeout.test.py
+++ b/tests/gold_tests/timeout/default_inactivity_timeout.test.py
@@ -93,7 +93,7 @@ class TestDefaultInactivityTimeout:
 
         # Set up expectectations for the timeout closing the connection.
         tr.Processes.Default.ReturnCode = 1
-        tr.Processes.Default.Streams.all = self.client_gold_file
+        tr.Processes.Default.Streams.All = self.client_gold_file
 
 
 test = TestDefaultInactivityTimeout("global config", use_override=False)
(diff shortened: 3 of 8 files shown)


bneradt:
## Verified locally

Built and ran in the CI-equivalent container (`ci-fedora-autest` preset, OpenSSL 3.5.7, `TS_HAS_QUICHE=0`). Two of the revived assertions were right and the expectations around them were wrong; both are fixed in c0213cf.

| Test | Revived | Result |
|---|---|---|
| `tls/allow-plain` | 3 | Live, passing |
| `tls/tls_hooks_client_verify` | 3 | Live — exposed a real config gap, passing after fix |
| `tls/tls_sni_groups` | 2 | Live, passing (both run; OpenSSL 3.5.7) |
| `timeout/default_inactivity_timeout` | 1 | Live — exposed a malformed gold file, passing after fix |
| `timeout/quic_no_activity_timeout` | 1 | Not exercised here — gated on `TS_HAS_QUICHE` |
| `h3/h3_sni_check` | 1 | Unreachable either way |
| `ip_allow/ip_allow` | 1 | Unreachable either way |

**Baseline.** Reverting just the two failing files to their `.all` form makes both pass again — confirming the failures came from the assertions waking up, not from the build or environment.

**Mutation check.** For the tests that passed on the first run, I broke one assertion in each to confirm it was actually running rather than still inert. `allow-plain` and `tls_sni_groups` both failed as they should. `quic_no_activity_timeout` did *not* — its only `gold_file` call site sits behind `if Condition.HasATSFeature('TS_HAS_QUICHE')`, and this build is OpenSSL-QUIC. That one still needs a quiche CI build to be exercised; I have not verified it.

### The two real findings

**`gold/client_default_inactivity_timeout.gold`** opened with ` ``` ` where the AuTest wildcard marker is `` `` ``. The three-backtick line was matched literally against `[DEBUG]: Configured logging at level DEBUG` and could never match. Its sibling `quic_no_activity_timeout.gold` has the correct two. The rest of the file was already right — the output does contain both `PARSE_INCOMPLETE` and `Failed HTTP/1 transaction with key: timeout2`.

**`tls_hooks_client_verify`** maps to an `https://` origin but never set `proxy.config.ssl.client.verify.server.policy`, so it inherited the `ENFORCED` default while `MakeOriginServer(ssl=True)` serves a self-signed cert:

```
WARNING: Core server certificate verification failed for (foo.com).
         Action=Terminate Error=self-signed certificate server=127.0.0.1 depth=0
```

ATS returned its `Could Not Connect` error page. `ReturnCode == 0` passed anyway, because curl did get *an* HTTP response — which is exactly the hole the `ExcludesExpression("Could Not Connect")` assertion existed to close. Set to `PERMISSIVE`, matching `tls_verify_base`, `tls_partial_blind_tunnel`, and `tls_check_cert_selection_reload`. This only affects ATS→origin; the test's subject is inbound client-cert verification, which is unchanged.

That failure also cascaded: with the first test run red, `request bad name` and `request badly signed cert` were skipped, so the pre-existing `traffic_out` "verify callback happens 2 times" checks had nothing to match and failed too. Those were collateral, not separate defects.

Leaving **WIP** on until CI confirms the quiche-gated path.

cmcfarlen (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as 0d51bfac343c373921aef034cdb502f5ad39bf01 for the 10.2.1 release.

PR #13578, opened 2026-08-21
Remove dead JSONRPC server_shutdown handler

masaori335:
Never registered nor called since it was added in #7478, and its void(YAML::Node) signature never matched the method handler shape it sat among.  Shutdown already syncs the cache dir from the event thread in traffic_server.cc; doing it from an RPC thread would race the dir writers.  Dropping it also removes mgmt's relative-path include of iocore's private P_CacheDir.h.


JosiahWI (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as 434e09f4879537e74019dcb300a12f32b69a5e6e for the 10.2.1 release.

PR #13579, opened 2026-08-21
Add non-HttpSM tests for `acquireSession`

JosiahWI:
This patch includes a few simple test cases for `acquireSession`. They cover only the behavior which does not require an `HttpSM` object.

### DIFF
--- a/include/proxy/http/HttpSessionManager.h
+++ b/include/proxy/http/HttpSessionManager.h
@@ -85,14 +85,33 @@ public:
   static bool match(PoolableSession *ss, sockaddr const *addr, CryptoHash const &host_hash,
                     TSServerSessionSharingMatchMask match_style);
 
-  /** Get a session from the pool.
-
-      The session is selected based on @a match_style equivalently to @a match. If found the session
-      is removed from the pool.
-
-      @return A pointer to the session or @c NULL if not matching session was found.
-  */
-  HSMresult_t acquireSession(sockaddr const *addr, CryptoHash const &host_hash, TSServerSessionSharingMatchMask match_style,
+  /** Search the pool for a server session compatible with the given address, hostname, and state machine.
+   *
+   * The selection criteria are controlled by @p match_style. At least one of
+   * @c TS_SERVER_SESSION_SHARING_MATCH_MASK_IP or @c TS_SERVER_SESSION_SHARING_MATCH_MASK_HOSTONLY
+   * must be set for a match to be possible; if neither is set, @c HSMresult_t::NOT_FOUND is
+   * returned unconditionally.
+   *
+   * When a compatible session is found and does not support multiplexed streams, it is removed from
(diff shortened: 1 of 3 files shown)


include/proxy/http/HttpSessionManager.h
```diff
@@ -85,14 +85,33 @@ class ServerSessionPool : public Continuation
...
+   * When a compatible session is found and does not support multiplexed streams, it is removed from
+   * the pool and ownership transfers to the caller. A session that supports multiplexed streams is
+   * not removed and remains available for subsequent acquisitions.
+   *
+   * @param[in]  addr           Remote address and port to match.
+   * @param[in]  hostname_hash  Cryptographic hash of the target hostname.
+   * @param[in]  match_style    Bitmask of @c TSServerSessionSharingMatchMask values specifying which
+   *                            attributes must agree between the candidate session and the request.
+   * @param[in]  sm             The requesting HTTP state machine; consulted for SNI and certificate
+   *                            validation when the corresponding bits are set in @p match_style.
```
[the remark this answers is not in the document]
JosiahWI: I don't want to guarantee that it can ever be nullptr - this is an issue on the test side which I want to fix in a follow-up - setting up an HttpSM will be a big effort.

JosiahWI:
[approve ci autest 3]

src/proxy/http/unit_tests/test_HttpSessionManager.cc
```diff
@@ -0,0 +1,337 @@
...
+
+/// The pool bookkeeping updates this gauge, which is normally initialized via HttpConfig.
+/// Create it here (when needed) and reset it between Catch2 runs so tests don't leak state.
+void
+init_metrics()
+{
+  if (http_rsb.pooled_server_connections == nullptr) {
+    http_rsb.pooled_server_connections = Metrics::Gauge::createPtr("proxy.process.http.pooled_server_connections");
+  }
+  Metrics::Gauge::store(http_rsb.pooled_server_connections, 0);
```
bneradt, line 138: [P2] This reset only happens before each Catch2 execution. Several section paths leave sessions for `pool.purge()`, but `purge()` clears the intrusive maps without decrementing `pooled_server_connections`; in the final section, one session remains and the gauge is left at 1 for later test cases. Please restore the gauge during teardown after `pool.purge()` (or use a scope guard that resets it) so this test does not leak global state.

include/proxy/http/HttpSessionManager.h
bneradt: [P3] This description omits the unconditional use of `sm` by `validate_session_origin_cert()` for multiplexing sessions. `acquireSession()` invokes that helper regardless of the `match_style` bits, and the helper dereferences `sm` to obtain the outbound SNI. Please document that `sm` must be valid whenever multiplexed candidates may be present, in addition to its use for match-style-controlled validation.

bneradt (approved):

PR #13580, opened 2026-08-21
compress: add Vary to cached client responses

traeak:
Related to Vary header for compressible content (#12741).
If a compressible cached object did not have the vary header (like after an ATS upgrade from 10.0.x to 10.1.x), the plugin gets read only access to the cached header and tries to add the Vary header.  This triggers errors in the logs for every matching request and doesn't add the Vary header to a client response.

```
[Aug 20 15:54:54.873] [ET_NET 25] ERROR: <InkAPI.cc:252 (TSError)> [/rpmbuilddir/BUILD/trafficserver-10.1.4/plugins/compress/compress.cc:297] [vary_header] ERROR: cannot add/update the Vary header
[Aug 20 15:54:54.873] [ET_NET 25] ERROR: <InkAPI.cc:252 (TSError)> [/rpmbuilddir/BUILD/trafficserver-10.1.4/plugins/compress/compress.cc:939] [add_vary_header_for_compressible_content] ERROR: failed to add Vary header for compressible content
```

With this fix the Vary header is added at the client response header hook if it does not already exist in the client response.

### DIFF
--- a/plugins/compress/compress.cc
+++ b/plugins/compress/compress.cc
@@ -772,25 +772,37 @@ transformable(TSHttpTxn txnp, bool server, HostConfiguration *host_configuration
 }
 
 static void
-add_vary_header_for_compressible_content(TSHttpTxn txnp, bool server, HostConfiguration * /* hc ATS_UNUSED */)
+add_vary_header_for_compressible_content(TSHttpTxn txnp)
 {
   TSMBuffer resp_buf;
   TSMLoc    resp_loc;
 
-  // Get the response headers
-  if (server) {
-    if (TS_SUCCESS != TSHttpTxnServerRespGet(txnp, &resp_buf, &resp_loc)) {
-      return;
-    }
-  } else {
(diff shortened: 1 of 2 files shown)


plugins/compress/compress.cc
```diff
@@ -772,23 +772,16 @@ transformable(TSHttpTxn txnp, bool server, HostConfiguration *host_configuration
 }
 
 static void
-add_vary_header_for_compressible_content(TSHttpTxn txnp, bool server, HostConfiguration * /* hc ATS_UNUSED */)
+add_vary_header_for_compressible_content(TSHttpTxn txnp)
```
serrislew: nitpicky but maybe change this to match the client function `add_vary_header_to_server_response`

serrislew (dismissed):

plugins/compress/compress.cc
traeak: taken

serrislew (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as 1377bf27146c9e1ae454d51960f02eddc8411b4a for the 10.2.1 release.

PR #13581, opened 2026-08-21
Add focused txn_box comparison test

### LINKED ISSUE #12335: txn_box plugin error in example
Does this txn_box example work?

txn_box:
- when: upstream-rsp
  do:
  - with: [ upstream-rsp-status , proxy-req-path ]
    select:
    - as-tuple:
      - eq: 404
      - match: "example"
      do:
      - debug: "Resetting upstream response"
      - upstream-rsp-status: [ 200 , "OK" ]
      - upstream-rsp-field<Cache-Control>: "max-age=3600"

Error: 
    While parsing as-tuple comparison at Line 5.
    While loading "with" directive at Line 3 in "select" at Line 5.
    While parsing directive at Line 3.
    While loading directives at Line 3.

When i also use 
- with: upstream-rsp-status
  select:
  - eq: 200

 It gives error.

bneradt:
Txn_box's documented tuple example was reported as failing to load, but
its existing coverage is embedded in a broad static-file test and does
not isolate either reported comparison form.

This adds a replay-based regression test that loads the default txn_box
root as a global plugin and verifies heterogeneous tuple selection and
scalar status selection through response mutations.

Fixes: #12335

### DIFF
--- a/tests/gold_tests/pluginTest/txn_box/tuple_comparison/txn_box.yaml
+++ b/tests/gold_tests/pluginTest/txn_box/tuple_comparison/txn_box.yaml
@@ -1,0 +1,33 @@
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+txn_box:
+- when: upstream-rsp
+  do:
+  - with: [ upstream-rsp-status, proxy-req-path ]
+    select:
+    - as-tuple:
+      - eq: 404
+      - match: "example"
+      do:
+      - upstream-rsp-status: [ 200, "OK" ]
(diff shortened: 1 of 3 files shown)


bryancall (approved):
Approving. I built this locally with experimental plugins on and ran the test, and
confirmed it's doing real work rather than passing vacuously: it runs (`Skipped: 0`,
so the `PluginExists` gate isn't quietly skipping it), and it fails as expected both
when the tuple is made unmatchable (`match: "example"` -> `"NOMATCH"`) and when the
config is given a parse error.

One thing that might let us close #12335 on firmer ground. The tuple form here was
already covered and passing at the `10.2.0` tag itself --
`tests/gold_tests/pluginTest/txn_box/static_file/static_file.replay.yaml` has the
identical `with: [ upstream-rsp-status , proxy-req-path ]` + `as-tuple` block, and
that's the snippet `doc/admin-guide/plugins/txn_box/examples.en.rst` literal-includes
as the documented example. No comparison-parsing changes have landed in txn_box since
10.2.0 either (that window is Coverity/ASAN/OpenSSL cleanups). So instead of "it works
on master," I think we can tell the reporter something stronger: the documented form
wasn't broken on the version they reported either, so the difference has to be in
their actual config or environment -- and we'd need the real file to get further.
Might be worth reopening just long enough to ask for it.

Two smaller notes:

- The description says the existing coverage "does not isolate either reported
  comparison form." The tuple form is isolated (and doc-linked) in static_file. What's
  genuinely new here is worth calling out instead: this loads a standalone
  `txn_box.yaml` with the default `txn_box:` root key via `plugin.config`, whereas
  static_file goes through `TxnBoxTestAndRun` with `config_key="meta.txn_box.global"`.
  The standalone path is what an operator actually uses, and only
  `prod/mTLS.txnbox.yaml` otherwise touches it -- that's a good argument for this test
  existing. The scalar select case is new too.

- The reporter's second snippet had no `do:` under `- eq: 200`:

  ```yaml
  - with: upstream-rsp-status
    select:
    - eq: 200
  ```

  This test adds one, so that literal form is still unexercised. Probably not the
  cause, but worth knowing it's untested if the issue comes back.

PR #13583, opened 2026-08-21
Metrics: close the id lookup race and bounds gaps left by the lock revert

cmcfarlen:
Follow-ups to #13567, which reverted the locking that #13310 had added to the `ts::Metrics::Storage`
read paths. That revert restored the performance but left the data race #13310 was closing, plus
some pre-existing bounds problems in the same functions. This closes the race without a lock, and
fixes the bounds.

### One gate for id validation

`valid()`, `lookup(IdType)`, `name()` and `rename()` each carried their own copy of the same range
test, and the copies disagreed. `valid()` rejected an offset past `MAX_SIZE`; the other three did
not. `_splitID` passes the low 16 bits of an id through unmasked and the offset check only applied
when the id named the current blob, so an id such as `0x0000FFFF` indexed well past the end of a
blob's 1024 entry arrays once a second blob existed. Ids reaching these accessors come from plugins
through the `TSStat*` API, so they are untrusted.

All four now go through `Storage::_is_allocated()`. It also fixes an off-by-one: `create()` returns
the id and then advances, so the next free slot was being accepted as allocated. An increment there
landed on the slot `create()` would hand out next, and since `create()` writes only the name and
never the value, the next plugin to call `TSStatCreate()` received a metric already carrying someone
else's count.

Nothing depended on the loose bound: `end()` builds an id at the allocation point that is compared
but never dereferenced, `iterator::next()` keeps the offset in range, and `find()` returns `end()`
on a miss.

### One published position

`_cur_blob` and `_cur_off` are gone, replaced by a single atomic holding the blob index above the
offset, packed as `_makeId` packs one.

Two atomics cannot be read as a coherent pair. `addBlob()` reset the offset before advancing the
blob, so a reader between the two stores saw the old blob index with a zero offset and rejected
every id in the just-completed blob. That is not a dropped increment: with the default
`ENABLE_FAST_SDK=OFF`, `TSStatInt*` feeds the result to `sdk_assert` and aborts through
`_TSReleaseAssert`. Ordering the two stores the other way only trades it for accepting ids in a blob
nothing has been written to yet.

One value removes the window rather than relocating it: crossing a blob is a single release store,
sequenced after the blob pointer it publishes. The position only ever increases, since `(N+1)<<16`
exceeds `N<<16 + offset` for any `offset < MAX_SIZE`, so an id is allocated exactly when it packs
below the bound and `_is_allocated()` is one acquire load and one compare. Acquiring the bound
acquires the blob install, so `_blobs` needs no check of its own. The offset test stays: `_splitID`
takes the low 16 bits, so an id in an earlier blob can name an offset past `MAX_SIZE` and still pack
under the bound. The packed value is also the id of the next free slot, which is exactly what
`end()` wants, so iteration's bound stops being reconstructed from two fields.

Thanks to @moonchen for finding the rollover window and the `rename()` race in review.

### `createSpan` and `rename` removed

It was the only path that could leave a blob partly filled: when a span did not fit it skipped to a
fresh blob, abandoning tail slots that were never handed out but that pack below the bound. With it
gone, blobs fill contiguously and "packs below the bound" means exactly "was handed out", with no
special case. It also has no callers outside the tests.

Two test consumers adapted: a span/rename section became rename-only, and `test_RecRegister.cc` was
using `createSpan(1)` as a cheap anonymous registration while hammering lookups, which `create()`
does as well. The test case covering `createSpan`'s blob boundary goes with it.

`rename()` goes for a different reason: it mutated a slot's name while `name()` and
`lookup(id, &out_name)` read that same `std::string` without the mutex and hand out `string_view`s
into it. Locking `rename()` does not close that — the readers are the lock free paths this PR exists
to keep — and giving names immutable storage with its own lifetime rules is a lot of machinery for a
function nothing outside the tests calls. Removing it leaves the useful invariant: a name is written
once, before the store that publishes it, and never changes, so the unlocked readers are correct by
construction and the `string_view` keys in `_lookups` are stable for the life of the process.

Both are removals from an installed header, so they belong in the 11.0.0 release notes. Neither has
a caller in tree, and `createSpan` handed out unnamed slots that only `rename()` could have named.

### `_extractType` on a negative id

It shifted a signed `IdType`, so `_extractType(NOT_FOUND)` sign extended to `-4` — a `MetricType`
outside its enumeration, returned by `Metrics::type()`. Shifting unsigned is not sufficient on its
own: the sign bit sits above the type field, so `NOT_FOUND` still yields `4`. Masking to the single
bit `_makeId` writes makes the function total for any input.

Two smaller ones: `rename()` computed a reference into the name storage before taking the mutex,
which read nothing but made the critical section look wider than it is, and `Metrics.h` had been
relying on `swoc/MemSpan.h` for `<limits>`.

### Testing

A new test resolves ids from several threads while another registers metrics across a few blob
boundaries. Under the `tsan` preset, making either allocation counter non-atomic again reports a data
race there. It does **not** catch a downgrade of the release/acquire pairs to relaxed — atomics are
race free at any ordering, so TSAN stays quiet and the assertions still hold. The memory orders are
reviewed, not tested, and the test says so.

Two ways that test could pass without testing anything, both found in review and both fixed. Readers
only published their tally on exit and nothing made them run before the writer finished, so on one
CPU every reader could see `stop` and resolve nothing while `resolved > 0` still held; each
resolution is now published as it happens and the writer waits for one before stopping. And an id
that `lookup()` clamps resolves to the `bad_id` slot, whose name is not empty, so the name check
could not detect a clamp; it now compares against the name that id must have.

`tools/benchmark/benchmark_Metrics.cc` is new; nothing in tree measured these paths, which is how a
global mutex on the hottest one went unnoticed. Four cases scaled by thread count, on a 10 core
machine at 20k ops/thread:

| threads | `increment(ptr)` | `increment(id)` | `lookup(id)` | `lookup(name)` |
|---|---|---|---|---|
| 1 | 0.12 ms | 1.81 ms | 1.05 ms | 2.34 ms |
| 4 | 0.24 ms | 2.20 ms | 1.15 ms | 37.6 ms |
| 16 | 4.71 ms | 7.04 ms | 2.09 ms | 80.4 ms |
| 64 | 19.6 ms | 24.0 ms | 6.13 ms | 317 ms |

`lookup(name)` is a deliberate control: it still takes the mutex, so it must degrade with thread
count. It goes 2.3 ms to 317 ms while `lookup(id)` goes 1.05 to 6.13 ms, which is the evidence that
the harness loads the machine rather than the lock free numbers being flat for want of load. Above
10 threads the machine is oversubscribed, so treat the shape as meaningful and the magnitudes as
not.

Comparing a build with and without the atomics commit put every case within noise, the only
consistent signal being 4-8% on `lookup(id)` — two `ldaprh` rather than two `ldrh` on ARM64, and
plain loads on x86. Set against what the mutex costs, it is not a trade worth considering.

Those numbers predate the packed position, so the packing was measured separately, as an A/B of
`bc333f401e` (two atomics, two acquire loads and a null blob test in `_is_allocated`) against the tip
of this branch. Same binary, same machine, `RelWithDebInfo`, 20k ops/thread, seven runs per point,
mean +- sd. This is a 16 logical core machine, so these absolute numbers are not comparable with the
table above.

| threads | case | two atomics | packed | delta |
|---|---|---|---|---|
| 1 | `increment(ptr)` | 33.6 +- 2.1 us | 32.7 +- 0.5 us | -2.9% |
| 1 | `increment(id)` | 63.0 +- 2.3 us | 50.2 +- 3.3 us | **-20.2%** |
| 1 | `lookup(id)` | 47.1 +- 3.7 us | 43.3 +- 3.5 us | **-8.0%** |
| 1 | `lookup(name)` | 318.0 +- 18.0 us | 305.3 +- 6.8 us | -4.0% |
| 4 | `increment(ptr)` | 177.6 +- 9.5 us | 175.9 +- 5.3 us | -1.0% |
| 4 | `increment(id)` | 176.2 +- 9.0 us | 154.4 +- 1.7 us | **-12.4%** |
| 4 | `lookup(id)` | 81.3 +- 18.8 us | 74.4 +- 1.7 us | **-8.5%** |
| 4 | `lookup(name)` | 4396.7 +- 96.0 us | 4457.0 +- 51.2 us | +1.4% |

`increment(ptr)` and `lookup(name)` are the controls: neither calls `_is_allocated`, and both stay
within 4%, which sets the noise floor. The two id paths move well outside it, and by a ratio that
falls out of the code: the benchmark's `increment(id)` case is `valid(id) ? increment(id, 1) : 0`,
which enters `_is_allocated` twice, and its delta is about twice `lookup(id)`'s. So the saving is
localized to the check, which is what removing one acquire load and one null test should do.

At 16 and 64 threads every case lands within noise, controls included — contention on the metric
atomics dominates, with `increment(ptr)` going from 33 us to 19 ms. The packing is a win where the
check is measurable and free where it is not.

### Provenance

The bounds and memory-order findings came out of a review of this code prompted by a production
`perf` profile, in which the `#13310` locking accounted for roughly half of all CPU in futex
contention. I do not have a public link for that review to cite. The parts of it this PR does not
implement — deleting the `sdk_assert` from the `TSStat*` entry points, and an opaque handle API for
plugins — were either out of proportion to the measured benefit or need an upstream decision first.
Removing `rename()` was on that list too, and review showed it was not optional.

Per-blob published counts were the other candidate for the rollover window, and would additionally
have excluded `createSpan`'s abandoned tail slots. Deleting `createSpan` achieves that instead, and
leaves the check at one load rather than two dependent ones.

`addBlob`'s bound assert was part of the same review and landed earlier in #13505.

### DIFF
--- a/src/tsutil/Metrics.cc
+++ b/src/tsutil/Metrics.cc
@@ -62,8 +62,9 @@ Metrics::Storage::addBlob() // The mutex must be held before calling this!
   // The write below is to _blobs[_cur_blob + 1], so the last usable blob index is MAX_BLOBS - 1.
   release_assert(_cur_blob < MAX_BLOBS - 1);
 
-  _blobs[++_cur_blob] = std::move(blob);
-  _cur_off            = 0;
+  _blobs[_cur_blob + 1] = std::move(blob);
+  _cur_off              = 0;
+  ++_cur_blob;
 }
 
 Metrics::IdType
@@ -113,14 +114,16 @@ Metrics::Storage::lookup(const std::string_view name) const
 Metrics::AtomicType *
 Metrics::Storage::lookup(Metrics::IdType id, std::string_view *out_name, Metrics::MetricType *out_type) const
 {
-  auto [blob_ix, offset]         = _splitID(id);
-  Metrics::NamesAndAtomics *blob = _blobs[blob_ix].get();
+  auto [blob_ix, offset] = _splitID(id);
 
-  // Do a sanity check on the ID, to make sure we don't index outside of the realm of possibility.
-  if (!blob || (blob_ix == _cur_blob && offset > _cur_off)) {
-    blob   = _blobs[0].get();
-    offset = 0;
+  // Anything that does not name an allocated slot resolves to the reserved bad_id slot rather than
+  // indexing out of range.
+  if (!_is_allocated(id)) {
+    blob_ix = 0;
+    offset  = 0;
   }
+
+  Metrics::NamesAndAtomics *blob = _blobs[blob_ix].get();
 
   if (out_name) {
     *out_name = std::get<0>(std::get<0>(*blob)[offset]);
@@ -159,14 +162,16 @@ Metrics::Storage::lookup(const std::string_view name, Metrics::IdType *out_id, M
 std::string_view
 Metrics::Storage::name(Metrics::IdType id) const
 {
-  auto [blob_ix, offset]         = _splitID(id);
-  Metrics::NamesAndAtomics *blob = _blobs[blob_ix].get();
+  auto [blob_ix, offset] = _splitID(id);
 
-  // Do a sanity check on the ID, to make sure we don't index outside of the realm of possibility.
-  if (!blob || (blob_ix == _cur_blob && offset > _cur_off)) {
-    blob   = _blobs[0].get();
-    offset = 0;
+  // Anything that does not name an allocated slot resolves to the reserved bad_id slot rather than
+  // indexing out of range.
+  if (!_is_allocated(id)) {
+    blob_ix = 0;
+    offset  = 0;
   }
+
+  Metrics::NamesAndAtomics *blob = _blobs[blob_ix].get();
 
   const std::string &result = std::get<0>(std::get<0>(*blob)[offset]);
 
@@ -225,13 +230,13 @@ Metrics::Storage::createSpan(size_t size, Metrics::MetricType type, Metrics::IdT
 bool
 Metrics::Storage::rename(Metrics::IdType id, std::string_view name)
 {
-  auto [blob_ix, offset]         = _splitID(id);
-  Metrics::NamesAndAtomics *blob = _blobs[blob_ix].get();
-
   // We can only rename Metrics that are already allocated
-  if (!blob || (blob_ix == _cur_blob && offset > _cur_off)) {
+  if (!_is_allocated(id)) {
     return false;
   }
+
+  auto [blob_ix, offset]         = _splitID(id);
+  Metrics::NamesAndAtomics *blob = _blobs[blob_ix].get();
 
   std::string    &cur = std::get<0>(std::get<0>(*blob)[offset]);
   std::lock_guard lock(_mutex);
--- a/src/tsutil/Metrics.cc
+++ b/src/tsutil/Metrics.cc
@@ -199,11 +199,13 @@ Metrics::Storage::rename(Metrics::IdType id, std::string_view name)
     return false;
   }
 
+  // Held across the whole rename: the name is the key _lookups is indexed by, so replacing it has
+  // to be serialized against every other writer of that slot's name.
+  std::lock_guard lock(_mutex);
+
   auto [blob_ix, offset]         = _splitID(id);
   Metrics::NamesAndAtomics *blob = _blobs[blob_ix].get();
-
-  std::string    &cur = std::get<0>(std::get<0>(*blob)[offset]);
-  std::lock_guard lock(_mutex);
+  std::string              &cur  = std::get<0>(std::get<0>(*blob)[offset]);
 
   if (cur.length() > 0) {
     _lookups.erase(cur);
--- a/src/tsutil/Metrics.cc
+++ b/src/tsutil/Metrics.cc
@@ -201,7 +201,10 @@ Metrics::Storage::createSpan(size_t size, Metrics::MetricType type, Metrics::IdT
   // On the final blob there is nowhere left to grow, so refuse a span that would fill or overflow
   // it rather than letting addBlob() assert. Same intent as the guard in create(), and the same
   // cost: some slots of the last blob go unused.
-  if (_cur_blob.load(std::memory_order_relaxed) >= MAX_BLOBS - 1 && _cur_off.load(std::memory_order_relaxed) + size >= MAX_SIZE) {
+  auto cur_blob = _cur_blob.load(std::memory_order_relaxed);
+  auto cur_off  = _cur_off.load(std::memory_order_relaxed);
+
+  if (cur_blob >= MAX_BLOBS - 1 && cur_off + size >= MAX_SIZE) {
     if (id) {
       *id = 0; // Slot 0 is the reserved bad_id.
     }
@@ -209,13 +212,11 @@ Metrics::Storage::createSpan(size_t size, Metrics::MetricType type, Metrics::IdT
   }
 
   // A span has to be contiguous, so one that does not fit in the current blob starts a new one.
-  if (_cur_off.load(std::memory_order_relaxed) + size > MAX_SIZE) {
+  if (cur_off + size > MAX_SIZE) {
     addBlob();
+    cur_blob = _cur_blob.load(std::memory_order_relaxed);
+    cur_off  = _cur_off.load(std::memory_order_relaxed);
   }
-
-  // Re-read: addBlob() above may have moved both.
-  auto const cur_blob = _cur_blob.load(std::memory_order_relaxed);
-  auto const cur_off  = _cur_off.load(std::memory_order_relaxed);
 
   Metrics::IdType           span_start = _makeId(cur_blob, cur_off, type);
   Metrics::NamesAndAtomics *blob       = _blobs[cur_blob].get();
--- a/src/tsutil/Metrics.cc
+++ b/src/tsutil/Metrics.cc
@@ -58,17 +58,16 @@ Metrics::Storage::addBlob() // The mutex must be held before calling this!
 {
   auto blob = std::make_unique<Metrics::NamesAndAtomics>();
 
-  auto const cur_blob = _cur_blob.load(std::memory_order_relaxed);
+  auto const [cur_blob, cur_off] = _splitID(static_cast<IdType>(_next_free.load(std::memory_order_relaxed)));
 
   debug_assert(blob);
   // The write below is to _blobs[cur_blob + 1], so the last usable blob index is MAX_BLOBS - 1.
   release_assert(cur_blob < MAX_BLOBS - 1);
 
   _blobs[cur_blob + 1] = std::move(blob);
-  _cur_off.store(0, std::memory_order_relaxed);
 
-  // Publishes the blob; both writes above are sequenced before it.
-  _cur_blob.store(cur_blob + 1, std::memory_order_release);
+  // Publishes the blob and the offset reset as one value; the write above is sequenced before it.
+  _next_free.store(_pack(cur_blob + 1, 0), std::memory_order_release);
 }
 
 Metrics::IdType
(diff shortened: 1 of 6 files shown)


src/tsutil/Metrics.cc
```diff
@@ -58,12 +58,17 @@ Metrics::Storage::addBlob() // The mutex must be held before calling this!
...
+  auto const cur_blob = _cur_blob.load(std::memory_order_relaxed);
+
   debug_assert(blob);
-  // The write below is to _blobs[_cur_blob + 1], so the last usable blob index is MAX_BLOBS - 1.
-  release_assert(_cur_blob < MAX_BLOBS - 1);
+  // The write below is to _blobs[cur_blob + 1], so the last usable blob index is MAX_BLOBS - 1.
+  release_assert(cur_blob < MAX_BLOBS - 1);
+
+  _blobs[cur_blob + 1] = std::move(blob);
+  _cur_off.store(0, std::memory_order_relaxed);
```
moonchen: Could the blob and offset be published as one coherent state, or via per-blob published bounds?

A reader can load the old `_cur_blob`, then observe `_cur_off == 0` after the reset above but before the `_cur_blob` store. `_is_allocated()` will consequently reject every valid ID in the just-completed blob.

This is more serious than a dropped increment: with the default `ENABLE_FAST_SDK=OFF`, `TSStatInt*` passes that result to `sdk_assert`, which calls the non-returning `_TSReleaseAssert`. If the initial sanity check succeeds but the subsequent lookup lands in this window, lookup instead clamps the operation to `bad_id`.

src/tsutil/unit_tests/test_Metrics.cc
```diff
@@ -640,3 +637,116 @@ TEST_CASE("Metrics span lands exactly on a blob boundary", "[libtsapi][Metrics]"
...
+  }
+
+  for (int i = 0; i < N_CREATE; ++i) {
+    auto const nm = "pub.order." + std::to_string(i);
+
+    REQUIRE(Metrics::Counter::createHiddenPtr(nm) != nullptr);
+    created[i].store(h.lookup(nm), std::memory_order_relaxed);
+  }
+
+  stop.store(true, std::memory_order_relaxed);
```
moonchen, line 684: Please add a barrier or ready-count handshake before starting the writer loop. Nothing currently guarantees that any reader reaches its loop before `stop` becomes true.

I reproduced the failure by pinning this test to one CPU: on run 4, every reader observed `stop == true` before doing work and the test failed at `resolved.load() > 0`.

src/tsutil/unit_tests/test_Metrics.cc
```diff
@@ -640,3 +637,116 @@ TEST_CASE("Metrics span lands exactly on a blob boundary", "[libtsapi][Metrics]"
...
+            continue;
+          }
+
+          // valid() accepted the id, so lookup() must hand back the metric with its name rather
+          // than clamping to the reserved bad_id slot.
+          std::string_view    name;
+          Metrics::MetricType type;
+          auto               *m = h.lookup(id, &name, &type);
+
+          if (m == nullptr || name.empty()) {
```
moonchen: This check does not enforce the preceding comment. An invalid lookup returns the `bad_id` pointer, whose name is non-empty, so neither condition detects that lookup clamped.

Since the expected name is known from `i`, please compare the returned name with `"pub.order." + std::to_string(i)` or otherwise explicitly reject the `bad_id` result.

moonchen (commented):
I left a few inline comments on the blob-rollover state and the concurrency test coverage.

cmcfarlen:
Pushed two commits for this.

**The blob rollover — 96e8e5b42e**

You were right that publishing one coherent value is the fix, and reordering the stores is not. Writing `_cur_blob` first and `_cur_off` second only trades a false reject for a false accept: a reader then sees the new blob with the old offset and accepts ids in a blob nothing has been written to. Two independent atomics have no coherent pair in either order.

`_cur_blob` and `_cur_off` are now one `std::atomic<uint32_t>` holding the blob index above the offset, packed exactly as `_makeId` packs them. Crossing a blob is a single release store:

```cpp
_next_free.store(_pack(cur_blob + 1, 0), std::memory_order_release);
```

Three things fall out of it:

- The value only ever increases — `(N+1)<<16` exceeds `N<<16 + offset` for any `offset < MAX_SIZE` — so an id is allocated exactly when it packs below the bound. `_is_allocated()` becomes one acquire load and one compare.
- Acquiring the bound acquires the blob install that was release-ordered before it, so the `_blobs[blob_ix] != nullptr` check is redundant and gone. That also answers the review note about the second acquire load on the hot path: there is only one now.
- The packed value *is* the COUNTER-typed id of the next free slot, so iteration's end bound is the load rather than something reconstructed from two fields.

The `offset >= MAX_SIZE` check has to stay. `_splitID` takes the low 16 bits, so an id in an earlier blob can name offset 60000 and still pack below the bound.

**createSpan is gone with it**

It had no callers outside the tests, and it was the only path that could leave a blob partly filled: when a span did not fit it skipped to a fresh blob, abandoning tail slots that were never handed out. Those slots pack below the bound, so the packed check would have called them allocated — which is also the substance of the suppressed review comment about `offset < MAX_SIZE` in earlier blobs. With `createSpan` gone, blobs fill contiguously and "packs below the bound" means exactly "was handed out", with no special case.

Two test consumers adapted: the span/rename section became a rename-only section, and `test_RecRegister.cc` was using `createSpan(1)` as a cheap anonymous registration while hammering lookups, which `create()` does just as well. The test case that existed only to cover `createSpan`'s blob-boundary quirk is deleted.

Note this removes a function from an installed header. Given it hands out unnamed slots and is unused in tree, out-of-tree use seems unlikely, but it is worth a line in the release notes.

**The test problems — 5608493eb0**

Both real, and the barrier alone would not have fixed the first one. A ready-count gets every reader into its loop, but on a single CPU the writer can still run to completion before any reader is scheduled, and they would all then see `stop` and resolve nothing while `resolved > 0` still passed. So readers now publish each resolution as it happens and the writer waits for one before setting `stop`. The ready-count is in there too, so the race still happens when there are CPUs for it.

The name check could not detect a clamp, as you said — `bad_id`'s name is not empty. Names are precomputed before the threads start and the reader compares against the one that id must have, so no allocation in the loop.

Full `test_tsutil` and `test_records` pass, and I checked both commits build individually.

cmcfarlen:
Both correct, both mine. Fixed in 580b4d13de.

**`<limits>`** — dropping `createSpan` removed `swoc/MemSpan.h`, and that was what supplied `<limits>` for `NOT_FOUND`'s `numeric_limits`. Worth being precise about the state: the header still compiles standalone today, through some other transitive path, so nothing is broken right now. That is exactly what makes it worth declaring rather than leaving to luck, since the path that happens to work is not one this header controls. Verified a translation unit including only `tsutil/Metrics.h` compiles before and after.

**Unused `cur_off`** — `addBlob()` destructured the packed value and used only the blob half. It now shifts out just the index. No warning locally, but there is no reason to bind a name that is never read.

While applying the second one I broke `create()`, which destructures the same expression and does need both halves; caught by the build, repaired before pushing. Full `test_tsutil` (15725 assertions) and `test_records` (294) pass, and the whole tree builds.

cmcfarlen:
Moved in f0478d5819.

Worth noting the mechanism as described is not quite what was happening. Binding `std::string &cur = std::get<0>(std::get<0>(*blob)[offset])` is address computation; it does not access the string's bytes. The first actual read was `cur.length()`, which was already inside the lock, so there was no unsynchronized access to the string and no data race by the standard's definition.

The change is still worth making. Computing that reference outside the lock made the critical section look wider than it was, and nothing in this function has a reason to sit outside it — the name is the key `_lookups` is indexed by, so the whole replacement belongs in one place. The allocation check stays outside, since it reads only the published bound and is lock free like its neighbours.

Low risk either way: `rename()` has no callers outside the tests. `LogObject::rename` is an unrelated class.

Full `test_tsutil` and `test_records` pass; whole tree builds.

src/tsutil/Metrics.cc
```diff
@@ -179,63 +190,21 @@ Metrics::Storage::type(IdType id) const
...
-
   // We can only rename Metrics that are already allocated
-  if (!blob || (blob_ix == _cur_blob && offset > _cur_off)) {
+  if (!_is_allocated(id)) {
     return false;
   }
 
-  std::string    &cur = std::get<0>(std::get<0>(*blob)[offset]);
+  // The name is the key _lookups is indexed by, so the whole replacement is serialized.
   std::lock_guard lock(_mutex);
```
moonchen: The acquire on `_next_free` only publishes a slot's initial construction. This mutex protects the writer, but `lookup(id, &out_name)` and `name(id)` still read the same `std::string` without `_mutex`, and both expose a `string_view` into it. I reproduced a TSAN race between the assignment below and the read in `lookup()` at line 136. Please give metric names stable immutable storage with suitable synchronization/lifetime semantics, or remove `rename()`.

moonchen (changes requested):
Requesting changes for the remaining name-storage race.

cmcfarlen:
Removed in 2d05898c94.

You are right that locking `rename()` cannot fix this. The readers you point at — `name()` and `lookup(id, &out_name)` — are unlocked on purpose; they are the paths this PR exists to keep off the mutex. So the choice was immutable name storage with its own lifetime rules, or dropping the function. For something with no caller outside the tests, an atomic name pointer per slot plus never reclaiming replaced strings is a lot of machinery to carry.

Dropping it leaves the invariant those readers actually need, which is worth stating rather than assuming: a slot's name is written once, before the release store that publishes it, and never changes afterwards. So the unlocked reads are correct by construction, and the `string_view` keys in `_lookups` stay valid for the life of the process. `_lookups.erase` disappeared with `rename()`, so that map only grows now.

That is the second removal from an installed header in this PR, after `createSpan`, and the two are related: `createSpan` handed out unnamed slots that only `rename()` could ever have named. Both are noted for the 11.0.0 release notes, and the description is updated — it had listed removing `rename()` among the things this PR deliberately did not do, which review has overtaken.

Full `test_tsutil` and `test_records` pass and the tree builds.

moonchen (approved):
The rename race is resolved: metric names are now write-once before publication, and the unsafe mutation API is gone. Verified the focused Metrics and records tests locally, including the Metrics suite under ThreadSanitizer.

cmcfarlen:
Cherry-picked to the 10.2.x branch as c2c2f17cf265d53aca111b99fbf4e0232d6b8999 for the 10.2.1 release.

PR #13584, opened 2026-08-25
Fix unsafe certificate common name extraction

### LINKED ISSUE #13484: TSSslServerCertUpdate() can crash or read out of bounds extracting a certificate CN
In src/api/InkAPI.cc, TSSslServerCertUpdate() extracts the certificate common name like this:

```cpp
const int              pos              = X509_NAME_get_index_by_NID(X509_get_subject_name(cert.get()), NID_commonName, -1);
const X509_NAME_ENTRY *common_name      = X509_NAME_get_entry(X509_get_subject_name(cert.get()), pos);
const ASN1_STRING     *common_name_asn1 = X509_NAME_ENTRY_get_data(common_name);
char *common_name_str = reinterpret_cast<char *>(const_cast<unsigned char *>(ASN1_STRING_get0_data(common_name_asn1)));
if (ASN1_STRING_length(common_name_asn1) != static_cast<int>(strlen(common_name_str))) {
  // Embedded null char
  return TS_ERROR;
}
```

Two problems:

1. pos is -1 when the certificate has no CN. X509_NAME_get_entry(name, -1) returns nullptr, and neither X509_NAME_ENTRY_get_data(nullptr) nor the subsequent dereferences are guarded, so loading a certificate without a CN can null-deref.
2. ASN1_STRING_get0_data() returns a length-prefixed buffer that is not guaranteed to be NUL-terminated. Calling strlen() on it can read past the end of the buffer if there is no NUL within its declared length. The existing length check only catches an embedded NUL (a CN shorter than the buffer), not the case where the buffer has no NUL at all.

This logic predates any in-flight OpenSSL 4 build-compatibility work (traces to commits from 2024) and is present on current master, independent of that work. It was noticed during review of #13476, which touches nearby lines for unrelated reasons but does not change this logic.

## Suggested fix

Guard pos >= 0 and the common_name/common_name_asn1 pointers before use, and build the CN string from ASN1_STRING_get0_data()/ASN1_STRING_length() directly (e.g. std::string_view or std::string with an explicit length) instead of strlen().

RajaMuhammadAwais:
Guard missing common names and use ASN.1 length when extracting certificate CNs. Add a cert_update regression test for certificates without a common name. Fixes #13484.


bneradt:
[approve ci]

tests/gold_tests/pluginTest/cert_update/cert_update.test.py
```diff
@@ -99,6 +99,19 @@
...
+# Server-Cert-Update-No-CN
+# A certificate without a common name must be rejected without crashing ATS.
+tr = Test.AddTestRun("Server-Cert-Update-No-CN")
+tr.Processes.Default.Env = ts.Env
+tr.Processes.Default.Command = (
+    'openssl req -x509 -newkey rsa:2048 -nodes -keyout {0}/no-cn.key -out {0}/no-cn.crt '
+    '-subj /O=NoCN -days 1 >/dev/null 2>&1 && '
+    'cat {0}/no-cn.key {0}/no-cn.crt > {0}/no-cn.pem && '
+    '{1}/traffic_ctl plugin msg cert_update.server {0}/no-cn.pem'.format(ts.Variables.SSLDir, ts.Variables.BINDIR))
+ts.Disk.traffic_out.Content = "gold/update-no-cn.gold"
```
bneradt: This assignment is dead, so the assertion you added never runs.

`Disk.<file>.Content = ...` routes to `TesterSet.Assign()`, which *replaces* the tester list rather than appending to it. Line 144 (`Client-Cert-Update`) assigns `gold/update.gold` to this same `ts.Disk.traffic_out.Content` afterwards, so `gold/update-no-cn.gold` is discarded and never compared against anything.

I verified this by replacing the entire contents of `gold/update-no-cn.gold` with `THIS_STRING_WILL_NEVER_APPEAR_ANYWHERE` and re-running the test: still `Passed: 1`.

There is a second problem hiding behind the first: **gold files are not regexes.** `autest/testers/gold_file.py` does a whole-content diff and only substitutes the wildcard tokens `` `` `` and `{}`; `.*` is compared literally. When I forced the gold to be the live assertion (by moving it after line 144), it failed with exactly that:

```
- ... Failed to update server cert with .*no-cn.pem
+ ... Failed to update server cert with /tmp/sb/cert_update/ts/ssl/no-cn.pem
```

No existing gold file under `tests/gold_tests` uses `.*`, which is consistent with it not being supported.

Suggested fix: drop `gold/update-no-cn.gold` and use a real regex tester with `+=` so it is additive.

```python
ts.Disk.traffic_out.Content += Testers.ContainsExpression(
    r"Failed to update server cert with .*no-cn\.pem", "ATS should reject a certificate that has no common name")
```

Lines 99 and 144 need to become `+=` as well, otherwise the later `Assign()` still wipes this one out. I ran that combination locally: it passes with your `InkAPI.cc` fix, fails when the expression is changed to one that cannot match, and fails against unpatched `InkAPI.cc`.

One more note: the test *does* catch the regression today, but only incidentally. `traffic_ctl` exits non-zero because `traffic_server` segfaults out from under the RPC connection, so the `ReturnCode = 0` check on line 112 is what trips. Adding `tr.StillRunningAfter = ts` would turn the "without crashing ATS" intent in your comment into an actual assertion instead of relying on that side effect.

src/api/InkAPI.cc
```diff
@@ -8316,19 +8316,36 @@ TSSslServerCertUpdate(const char *cert_path, const char *key_path)
...
+    if (common_name_str.find('\0') != std::string::npos) {
       // Embedded null char
       return TS_ERROR;
     }
-    Dbg(dbg_ctl_ssl_cert_update, "Updating from %s with common name %s", cert_path, common_name_str);
 
+    Dbg(dbg_ctl_ssl_cert_update, "Updating from %s with common name %s", cert_path, common_name_str.c_str());
     // Update context to use cert
-    cc = lookup->find(common_name_str);
+    cc = lookup->find(common_name_str.c_str());
```
bneradt: `SSLCertLookup::find()` takes `const std::string &` (`src/iocore/net/P_SSLCertLookup.h:156`), so `.c_str()` discards the length you just computed and makes the compiler build a second `std::string` — including a `strlen()` — for the temporary. Pass the string through directly:

```cpp
cc = lookup->find(common_name_str);
```

Compare line 8029 in this same file, which passes its `std::string` to `find()` unchanged.

Minor, same idea on the `Dbg()` above: the other `dbg_ctl_ssl_cert_update` sites in this file use the `%.*s` form (e.g. lines 8072, 8241), so `"%.*s"` with `static_cast<int>(common_name_str.size()), common_name_str.data()` would match local style. Either way works now that the string is length-bounded.

src/api/InkAPI.cc
```diff
@@ -8316,19 +8316,36 @@ TSSslServerCertUpdate(const char *cert_path, const char *key_path)
...
 
-    // Extract common name
-    const int              pos              = X509_NAME_get_index_by_NID(X509_get_subject_name(cert.get()), NID_commonName, -1);
-    const X509_NAME_ENTRY *common_name      = X509_NAME_get_entry(X509_get_subject_name(cert.get()), pos);
+    // Extract common name. X509_NAME_get_index_by_NID() returns -1 when the
+    // certificate has no commonName, and ASN1_STRING_get0_data() does not
+    // guarantee NUL termination. Check both conditions before using the data.
+    const X509_NAME *subject = X509_get_subject_name(cert.get());
+    const int        pos     = X509_NAME_get_index_by_NID(subject, NID_commonName, -1);
+    if (pos < 0) {
```
bneradt, line 8403: Optional: the three new `return TS_ERROR` paths are silent. The load failure a few lines up uses `SSLError("Failed to load certificate/key from %s", cert_path)`, so an operator who feeds in a CN-less cert gets no diagnostic at all — the only trace is the calling plugin's own `Dbg()`, which needs its debug tag enabled. A single `Dbg(dbg_ctl_ssl_cert_update, ...)` or `SSLError()` naming the reason would make this much easier to support in the field.

bneradt (commented):
Reviewed and tested this locally (Fedora container, cherry-picked onto current `master`, `cmake --build build -t format` clean).

**The core fix is correct and I confirmed it does what it says.** Reverting only `src/api/InkAPI.cc` and running the new test reproduces the crash from #13484:

```
traffic_server: received signal 11 (Segmentation fault)
/lib64/libcrypto.so.3(ASN1_STRING_get0_data+0xc) [0xffffb00d170c]
libtsapi.so(_Z21TSSslServerCertUpdatePKcS0_+0x214) [0xffffb0786464]
cert_update.so(_Z14CB_cert_updateP10tsapi_cont7TSEventPv+0x1f4) [0xffffad19365c]
```

With the patch applied, `traffic_server` survives and logs `Failed to update server cert with .../no-cn.pem`. Both parts of the issue (unguarded `pos == -1`, and `strlen()` on a non-NUL-terminated `ASN1_STRING`) are addressed, and the embedded-NUL rejection is preserved.

Two things to fix before this lands: the new gold file is never actually compared (details inline), and a small `std::string` round-trip in the `lookup->find()` call.

tests/gold_tests/pluginTest/cert_update/cert_update.test.py
RajaMuhammadAwais: Thanks for the detailed review. I removed the unused gold file, switched the testers to additive `+=` with `Testers.ContainsExpression`, added `tr.StillRunningAfter = ts`, and changed the lookup to `find(common_name_str)`. These fixes have been pushed to the existing PR branch.

src/api/InkAPI.cc
```diff
@@ -8316,17 +8316,38 @@ TSSslServerCertUpdate(const char *cert_path, const char *key_path)
       return TS_ERROR;
     }
 
-    // Extract common name
-    const int              pos              = X509_NAME_get_index_by_NID(X509_get_subject_name(cert.get()), NID_commonName, -1);
-    const X509_NAME_ENTRY *common_name      = X509_NAME_get_entry(X509_get_subject_name(cert.get()), pos);
+    // Extract common name. X509_NAME_get_index_by_NID() returns -1 when the
+    // certificate has no commonName, and ASN1_STRING_get0_data() does not
+    // guarantee NUL termination. Check both conditions before using the data.
+    const X509_NAME *subject = X509_get_subject_name(cert.get());
```
maskit: The type probably has to be `auto` to deal with OpenSSL API signature difference.

maskit:
```
../src/api/InkAPI.cc:8354:32: error: no matching function for call to 'X509_NAME_get_index_by_NID'
    const int        pos     = X509_NAME_get_index_by_NID(subject, NID_commonName, -1);
                               ^~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/openssl/x509.h:810:5: note: candidate function not viable: 1st argument ('const X509_NAME *' (aka 'const X509_name_st *')) would lose const qualifier
int X509_NAME_get_index_by_NID(X509_NAME *name, int nid, int lastpos);
    ^
```

src/api/InkAPI.cc
```diff
@@ -8395,17 +8395,38 @@ TSSslServerCertUpdate(const char *cert_path, const char *key_path)
...
-    // Extract common name
-    const int              pos              = X509_NAME_get_index_by_NID(X509_get_subject_name(cert.get()), NID_commonName, -1);
-    const X509_NAME_ENTRY *common_name      = X509_NAME_get_entry(X509_get_subject_name(cert.get()), pos);
-    const ASN1_STRING     *common_name_asn1 = X509_NAME_ENTRY_get_data(common_name);
-    char *common_name_str = reinterpret_cast<char *>(const_cast<unsigned char *>(ASN1_STRING_get0_data(common_name_asn1)));
-    if (ASN1_STRING_length(common_name_asn1) != static_cast<int>(strlen(common_name_str))) {
+    // Extract common name. X509_NAME_get_index_by_NID() returns -1 when the
+    // certificate has no commonName, and ASN1_STRING_get0_data() does not
+    // guarantee NUL termination. Check both conditions before using the data.
+    const auto *subject = X509_get_subject_name(cert.get());
```
maskit: We have to let `auto` decide whether it's const or not. Just `auto *subject = `.

maskit (commented):
Looks good to me. I'll leave the approval to @bneradt.

bneradt (commented):
Follow-up review of cd33db4: the CN extraction implementation looks correct. It checks for a missing CN before accessing its ASN.1 value, uses the explicit length, preserves embedded-NUL rejection, passes the string directly to lookup->find(), and includes debug diagnostics. The OpenSSL subject pointer now correctly lets auto infer constness. All 15 reported CI checks pass. However, merge commit 79120a68 dropped the previously fixed no-CN regression test; the current PR changes only InkAPI.cc. Please restore that coverage before approval, as detailed inline. This was a source/history review; I did not rerun the previously reported local tests or build ATS in this unconfigured worktree.

src/api/InkAPI.cc
```diff
@@ -8395,17 +8395,38 @@ TSSslServerCertUpdate(const char *cert_path, const char *key_path)
...
-    const X509_NAME_ENTRY *common_name      = X509_NAME_get_entry(X509_get_subject_name(cert.get()), pos);
-    const ASN1_STRING     *common_name_asn1 = X509_NAME_ENTRY_get_data(common_name);
-    char *common_name_str = reinterpret_cast<char *>(const_cast<unsigned char *>(ASN1_STRING_get0_data(common_name_asn1)));
-    if (ASN1_STRING_length(common_name_asn1) != static_cast<int>(strlen(common_name_str))) {
+    // Extract common name. X509_NAME_get_index_by_NID() returns -1 when the
+    // certificate has no commonName, and ASN1_STRING_get0_data() does not
+    // guarantee NUL termination. Check both conditions before using the data.
+    auto     *subject = X509_get_subject_name(cert.get());
+    const int pos     = X509_NAME_get_index_by_NID(subject, NID_commonName, -1);
+    if (pos < 0) {
```
bneradt, line 8403: [P2] Restore the no-CN regression test lost during the merge

The current head no longer contains Server-Cert-Update-No-CN in tests/gold_tests/pluginTest/cert_update/cert_update.test.py. Comparing merge 79120a68 with its first parent shows that the merge removed the entire run: certificate generation, the cert_update.server invocation, the additive ContainsExpression for rejection, ReturnCode = 0, and StillRunningAfter = ts. Thus the earlier review fixes were implemented but subsequently lost, and the current green AuTests no longer exercise this missing-CN crash regression. Please restore that run alongside the upstream test additions and verify it fails against the unpatched API and passes with this fix.

PR #13590, opened 2026-08-26
Make plugin API cleanup functions noexcept

bryancall:
Exceptions have never been a supported error channel across the plugin API: API functions report failure through return values, and no header or document has ever promised that a TS API call throws anything a plugin could catch. That contract was never written down, though, so any exception escaping from the core into plugin code is both undiagnosable and invisible to static analysis.

It matters most in destructors. Plugin destructors routinely call the API cleanup functions, destructors are implicitly `noexcept`, and so an exception escaping one of those calls terminates the process with no attribution at all.

### What this does

- Annotates 28 API functions `noexcept` in `ts.h`, the cleanup/teardown family that destructors call: `TSContDestroy`, `TSHandleMLocRelease`, `TSIOBuffer*`/`TSMBufferDestroy`, `TSHttpHdrDestroy`, `TSMimeHdrDestroy`, `TSVConn{Close,Abort,Shutdown}`, `TSActionCancel`, `TSCacheKeyDestroy`, `TSSslContextDestroy`, `TSFetchDestroy`, `TSTextLogObject*`, the `TSMutex*` family, `TSThread{Wait,Destroy}`, `TSContData{Get,Set}`, and `TSError`).
- Converts each definition to a function-try-block whose handler calls `ink_abort("exception escaped %s", __func__)`. An exception that would previously have escaped now produces a named, attributable abort instead of a bare `std::terminate` inside a plugin destructor.
- Annotates `ts::do_abort` `[[noreturn]] noexcept` and contains formatting failures inside it, and does the same for `ts::shared_mutex::_call_fatal`. Both build their message with `swoc::bwprint`/`Strerror`, which allocate, so the abort path could itself throw `bad_alloc` before reaching the abort, from inside `~write_guard`/`~read_guard`.
- Updates the matching prototypes in the developer guide (22 pages) and states the no-exceptions contract in the plugin getting-started chapter.

### Overhead

None on the happy path, as expected for table-driven exception handling. Comparing generated code for a representative wrapped function against the same function unwrapped, the instruction sequences are identical except that the final call can no longer be a tail call. Measured on `InkAPI.cc` built before and after with identical flags: `__text` grew 896 bytes (+0.66%, about 45 bytes per wrapped function) and `__gcc_except_tab` grew 500 bytes of read-only data that is never touched unless an exception unwinds.

In the other direction this is a small optimization enabler: callers that can see `noexcept` no longer need their own landing pads around these calls, so plugin cleanup code should get marginally smaller.

### Coverity

**This should resolve 83 Coverity `UNCAUGHT_EXCEPT` defects.**

The checker flags this pattern because it has to assume any call in a destructor can throw. I arrived at that number by parsing the body of every destructor it flags and classifying what each one calls: 80 whose only calls are functions this PR makes `noexcept`, plus 3 `ts::shared_mutex` guard destructors covered by the `_call_fatal` fix. By area that is 61 in plugins, 13 in `tscpp/api`, 5 in core, 2 in examples, and 2 in test code.

Treat 83 as a projection rather than a measurement. It assumes the checker credits a `noexcept` callee and stops reporting the caller, which is the documented behavior but is not something I can confirm without a scan of the merged tree. The first analysis run after this lands will give the real figure, and I will follow up with it.

Destructors whose throw path is a member subobject rather than an API call are not affected by this change and need their own root cause, as do the CLI tool entry points. Those are follow-up work.

### ABI

Adding `noexcept` does not change the ABI. In the Itanium C++ ABI an exception specification is part of a function *type* but is not encoded in the mangled name of a function declaration, and it does not affect the calling convention.

I verified this rather than assuming it. Building `src/api/InkAPI.cc` before and after with identical flags, both objects export 748 text symbols with zero name differences: `TSContDestroy` is `_Z13TSContDestroyP10tsapi_cont` in both, `TSError` is `_Z7TSErrorPKcz` in both. As an end-to-end check, a caller compiled against the old declaration and linked against the new `noexcept` definition resolves to the same symbol, links, and runs. So an existing plugin binary keeps working against a rebuilt `libtsapi` and vice versa.

Function pointers are also unaffected: a `noexcept` function converts implicitly to a plain function-pointer type, so `void (*p)(TSCont) = TSContDestroy;` still compiles. Mangling would only shift if a parameter or template argument were spelled as a `noexcept` function type, which nothing in the API does. Nothing in the tree takes the address of any annotated function.

The compatibility cost is at the source level, not the binary level, and is described below.

### Notes for reviewers

- **Behavior change:** a plugin that wrapped a TS API call in a `catch` expecting to catch a core-internal exception would now see an abort. Exceptions were never a supported error channel here, so this standardizes existing practice rather than removing a documented one. It is a real change, though, and worth a release note.
- **Source compatibility:** `noexcept` has been part of the function type since C++17, so any translation unit that includes `ts.h` and separately declares one of these functions without `noexcept` becomes ill-formed. This applies equally to a plugin built as C++17 or C++20. Seven in-tree test stubs that redefine API functions needed the annotation added; out-of-tree plugins doing the same will need the same one-word change. The stub edits are mechanical and have to land in the same commit or the build breaks, which is why they are here rather than split out.
- **Selection rule:** the set is "the cleanup/teardown entry points a destructor can reasonably call," derived by parsing what the flagged destructors actually call, then extended to the obvious siblings (`TSMutexDestroy`, `TSMutexLockTry`, `TSMimeHdrDestroy`, `TSSslContextDestroy`) so the boundary is not an artifact of a scanner hit list. Widening it to the whole API surface is a reasonable follow-up if reviewers prefer that; it did not seem like the right first step.


bryancall:
[approve ci autest 2]

bryancall:
[approve ci autest 2]

cmcfarlen (approved):
Went through all 37 files. The write-up is unusually thorough, so rather than take the ABI, overhead, and source-compatibility claims on trust I checked the ones that would hurt if they were wrong:

- **`ink_abort` really is `[[noreturn]]`** (`TS_NORETURN` at `include/tscore/ink_error.h:53`), so flowing off the end of the handler in the eight value-returning wrappers is well-formed rather than undefined.
- **No C exposure.** `ts.h:32` already refuses to compile below C++17, so `noexcept` cannot reach a C translation unit. The source-compatibility cost is genuinely limited to redeclarations, as described.
- **The stub coverage is exactly complete.** Grepping the tree for definitions of all 28 names turns up redefinitions only in `src/api` plus the seven stub files in this PR, so there is no missed target hiding behind a non-default build option.
- **The docs match the header.** The doc hunks add `noexcept` to 26 unique prototypes and that set is exactly the annotated list. The two without doc changes, `TSCacheKeyDestroy` and `TSThreadWait`, have no `.. function::` prototype anywhere, only prose mentions, so nothing is missing.
- **The function-pointer claim holds** — nothing in the tree takes the address of any annotated function.
- **`do_abort` is not a behavior change.** Master already ended with `abort()`; the diff moves it into the try block and adds a second one to the handler. And `[[noreturn]]` on `shared_mutex::_call_fatal` is correct precisely because `fatal_error` expands to `do_abort`, which this PR makes `[[noreturn]]`.

Three observations, none of them blocking.

**The diagnostic family is split.** Only `TSError` gets `noexcept`; `TSStatus`, `TSNote`, `TSWarning`, `TSFatal`, `TSAlert`, and `TSEmergency` sit on adjacent lines in `ts.h` and do not. They are equally leaf-like and equally callable from a destructor, `TSFatal`/`TSAlert`/`TSEmergency` especially since those are what you reach for on a doomed path. Given the stated selection rule of extending to the obvious siblings, these look like the most obvious ones left out.

**Attribution is coarser than the message reads, for the re-entrant members of the set.** Most of it is leaf cleanup, but `TSVConnClose` and `TSVConnAbort` call `vc->do_io_close()`, and `TSContDestroy` calls `INKContInternal::destroy()`, which can `schedule_imm(this)`. Those paths can reach another plugin's code. If that plugin's handler throws, the abort reads `exception escaped TSVConnClose`, naming the caller rather than the thrower. Still strictly better than today's bare `std::terminate`, but a sentence in the getting-started text saying the named function is where the exception surfaced rather than necessarily where it originated would save someone a wrong turn during triage.

**A cost question on the hot accessors.** `TSContDataGet`, `TSContDataSet`, `TSIOBufferReaderAvail`, `TSIOBufferReaderConsume`, and the `TSMutex` lock/unlock trio have bodies a compiler can prove non-throwing: a pointer dereference, an integer return, a pthread macro. For those, a bare `noexcept` gives the identical guarantee with no handler emitted at all, whereas the function-try-block is what costs the roughly 45 bytes and the lost tail call. Since the 896 byte figure is aggregate, it would be worth knowing whether the wrapper is elided for the trivial bodies or paid for on paths that run once per callback. Keeping the wrapper uniform is defensible — it means nobody has to re-audit later which bodies happen to be trivial — I would just rather it be a deliberate choice than a side effect.

The framing here is right, too: the contract was always that exceptions are not an error channel across the API, and this makes the compiler enforce what the code already assumed, with an attributable abort instead of an anonymous terminate inside a plugin destructor. The release note called out in the reviewer notes is worth doing.

LGTM.

PR #13593, opened 2026-08-27
Fix Clang 21 analyzer findings

bneradt:
Newer Clang releases expose latent ownership and error-handling issues,
while the analyzer preset currently produces a GCC compilation database
that Clang cannot reliably consume.

This patch selects Clang explicitly for the analyzer preset, fixes the
reported leaks, unchecked stream calls, and directory scanning under a
mutex, and reshapes the remaining flagged code so the analyzer can
follow it. That gives ATS a clean diagnostic baseline before the job
moves to Ubuntu 26.04.

### DIFF
--- a/src/proxy/http3/Http3Frame.cc
+++ b/src/proxy/http3/Http3Frame.cc
@@ -599,8 +599,7 @@ Http3FrameFactory::create_headers_frame(IOBufferReader *header_block_reader, siz
 {
   ats_unique_buf buf = ats_unique_malloc(header_block_len);
 
-  int64_t nread;
-  while ((nread = header_block_reader->read(buf.get(), header_block_len)) > 0) {
+  while (header_block_reader->read(buf.get(), header_block_len) > 0) {
     ;
   }
 
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -267,6 +267,8 @@
       "description": "CI Pipeline config for running clang-analyzer",
       "inherits": ["ci"],
       "cacheVariables": {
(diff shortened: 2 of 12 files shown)


src/proxy/http3/Http3Frame.cc
```diff
@@ -599,10 +599,9 @@ Http3FrameFactory::create_headers_frame(IOBufferReader *header_block_reader, siz
 {
   ats_unique_buf buf = ats_unique_malloc(header_block_len);
 
-  int64_t nread;
-  while ((nread = header_block_reader->read(buf.get(), header_block_len)) > 0) {
-    ;
-  }
+  int64_t const nread = header_block_reader->read(buf.get(), header_block_len);
+
+  ink_release_assert(nread == static_cast<int64_t>(header_block_len));
```
serrislew: new crash path? intended?

bneradt: It was to satisfy a clang-analyzer complaint.

Removing nread also does this though and the clang-analyzer report as well. Let's do that instead.

serrislew (approved):

cmcfarlen:
Cherry-picked to the 10.2.x branch as 91f21fb4860f659e1cfbdd8e2b31787685cba13b for the 10.2.1 release.

cmcfarlen:
Cherry-picked to the 10.1.x branch as 2ed165fad19c202346a4ab2a575321cdbfee90c3 for the 10.1.5 release.

PR #13595, opened 2026-08-27
Normalize repeated carriage returns in headers

### LINKED ISSUE #12195: ATS forwards extra carriage returns in header line endings
Hi,

If you send ATS a request with extra carriage returns in header line endings, it forwards them as-is.

For example, if you send the following request:
```
GET / HTTP/1.1\r\n
Host: echo\r\n
Extra-CRs: \r\r\r\r\n
\r\n
```

Then, ATS forwards the following:
```
GET / HTTP/1.1\r\n
Host: echo\r\n
Extra-CRs: \r\r\r\r\n
Client-ip: 172.18.0.1\r\n
X-Forwarded-For: 172.18.0.1\r\n
Via: http/1.1 traffic_server[a24326f7-ff3b-4679-9801-53028b257033] (ApacheTrafficServer/10.2.0)\r\n
\r\n
```
Note that the extra carriage returns are still present :)

If you send more than 3 extra (4 total) carriage returns in a single header line ending, though, they all get normalized away. Kinda fun.

-Ben

RajaMuhammadAwais:
## Summary

Fix HTTP header serialization so malformed header line endings containing repeated carriage-return (`CR`) bytes are not forwarded unchanged to the next hop.

This addresses [issue #12195](https://github.com/apache/trafficserver/issues/12195), where ATS was observed forwarding header lines such as `Extra-CRs: \r\r\r\r\n` with the extra carriage returns preserved.

## Problem

ATS intentionally preserves the original input buffer for some valid `CRLF`-terminated fields as a serialization fast path. The parser already trims trailing whitespace, line-feed, and carriage-return characters from the parsed field value, but it could still mark a line with repeated carriage returns as safe for raw-line preservation merely because it ended in `CRLF`.

As a result, the normalized field representation and the serialized wire representation could differ: the parsed value was normalized, while the original malformed line—including the extra `CR` bytes—could be forwarded downstream.

This is undesirable for HTTP intermediaries because downstream components should not receive malformed line endings that ATS has already parsed and normalized.

## Implementation

The parser now disables raw input-line preservation when either of the following is true:

1. The line does not end in the normal `CRLF` sequence; or
2. The byte immediately preceding the final `CRLF` is itself a carriage return, indicating repeated carriage returns at the line ending.

When raw preservation is disabled, ATS serializes the parsed field using its normal canonical format, producing a single `CRLF` terminator. The existing fast path for correctly formed header lines remains unchanged.

## Regression Test

Added a header unit test that:

1. Parses a request containing `Extra-CRs: \r\r\r\r\n`.
2. Serializes the parsed request.
3. Verifies that the output contains the normalized `Extra-CRs: \r\n` form.
4. Verifies that the serialized output does not contain repeated carriage returns.

## Validation

The following checks were completed locally:

- The new focused regression test passed.
- The complete `test_proxy_hdrs` suite passed: **53 test cases and 436,882 assertions**.
- Repository formatting checks passed, including clang-format, yapf, cmake-format, trailing-whitespace, and DOS-line-ending checks.
- `git diff --check` passed.

## Scope and Compatibility

This change is limited to serialization of header fields whose raw line ending contains repeated carriage returns. Properly formed `CRLF` header lines retain the existing behavior. No public API, configuration setting, or wire format for valid HTTP headers is changed.

## Related Issue

Fixes #12195

### DIFF
--- a/src/proxy/hdrs/MIME.cc
+++ b/src/proxy/hdrs/MIME.cc
@@ -2509,7 +2509,8 @@ mime_parser_parse(MIMEParser *parser, HdrHeap *heap, MIMEHdrImpl *mh, const char
       }
       field_name.rtrim_if(&ParseRules::is_ws);
       raw_print_field = false;
-    } else if (parsed.suffix(2) != "\r\n") {
+    } else if (parsed.suffix(2) != "\r\n" || (parsed.size() > 2 && parsed[parsed.size() - 3] == '\r')) {
+      // Do not preserve malformed line endings when forwarding the field.
       raw_print_field = false;
     }
 
--- a/src/proxy/hdrs/unit_tests/test_Hdrs.cc
+++ b/src/proxy/hdrs/unit_tests/test_Hdrs.cc
@@ -2983,6 +2983,32 @@ TEST_CASE("HTTP parser tolerates high-bit bytes without UB", "[proxy][hdrtest]")
   req_hdr.destroy();
 }
 
+TEST_CASE("HTTP parser normalizes repeated carriage returns in header line endings", "[proxy][hdrtest]")
+{
+  constexpr std::string_view message = "GET / HTTP/1.1\r\nHost: example.com\r\nExtra-CRs: \r\r\r\r\n\r\n"sv;
+
+  HTTPParser parser;
+  http_parser_init(&parser);
+
+  HTTPHdr  req_hdr;
+  HdrHeap *heap = new_HdrHeap(HdrHeap::DEFAULT_SIZE + 64);
+  req_hdr.create(HTTPType::REQUEST, HTTP_1_1, heap);
+
+  auto start = message.data();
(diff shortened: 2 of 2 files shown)


bryancall (approved):
Verified locally: at the base commit with only the test applied the new case fails, and it passes with the `MIME.cc` change, so this is a real regression test. Full `test_proxy_hdrs` is green here, 53 test cases and 437,070 assertions.

Approving. This closes the line reported in #12195.

It does not cover the general case: the check looks at the single byte before the terminating CRLF, and the value trim strips SP and HT as well as CR, so `Extra-CRs: \r \r\n` is still forwarded with the bare CR intact. I filed #13624 with the details and a suggested form of the check, and I am happy to take it if you would rather not.

One thing worth knowing either way: your fixture sits exactly on the raw-print pad limit of 7, so adding a CR to "strengthen" it would make the test pass without the fix.

PR #13596, opened 2026-08-27
slice_stale_generation.test.py: Accept either abort shape

bneradt:
The test pinned one Proxy Verifier parse state, PARSE_INCOMPLETE, for a
transaction that slice aborts. How far slice got before aborting decides
what the client sees, so under load the client instead reads a complete
response header whose body never arrives, and Proxy Verifier reports a
Content-Length body underrun.

This patch accepts either shape, since both mean the client received no
complete response, which is what the test checks. The sibling cold child
run already asserts only that the transaction failed.


bneradt:
[approve ci autest 2]

bneradt:
[approve ci clang-analyzer autest 2]

bneradt:
[approve ci clang-analyzer]

traeak (approved):
Looks good, adds robustness to a test that can fail in more than one way.

PR #13600, opened 2026-08-28
rate_limit: catch YAML errors instead of dying on reload

### LINKED ISSUE #13598: rate_limit: YAML parse errors escape yamlParser() and terminate ATS on config reload
`SniSelector::yamlParser()` wraps only `YAML::LoadFile` in a try block
(`sni_selector.cc:35-43`). Every node accessor and `.as<T>()` call after it runs
unguarded, so a malformed value in the rate_limit config throws a
`YAML::Exception` that escapes the function.

On startup that replaces the intended `TSFatal` with an unhandled exception. On
`traffic_ctl config reload` it kills a running server: `sni_config_cont` is the
callback registered with `TSMgmtUpdateRegister`, so the throw at
`sni_selector.cc:191` unwinds into the event loop. The `else` branch directly
below that call already handles a failed reload by logging and keeping the old
config, and it never gets the chance to run.

### Reproduce

Either config throws. Load it, then run `traffic_ctl config reload`.

A selector entry with no `sni` key:

```yaml
selector:
  - limit: 100
```

`sni["sni"].IsSequence()` at `sni_selector.cc:116` throws `YAML::InvalidNode`
on the const node, before the "selector node is not a map or without a name"
error below it can fire.

A `percentage` value taken from the documentation:

```yaml
ip-rep:
  - name: test
    size: 15
    percentage: 0.9
```

`ip_reputation.cc:93` reads it as `uint32_t`, so `0.9` throws
`YAML::TypedBadConversion<unsigned int>`.

Compiled against the vendored yaml-cpp, the two cases report:

```
invalid node; first invalid key: "sni"
yaml-cpp: error at line 1, column 13: bad conversion
```

The same exposure applies to the other conversions: `limit`, `rate`,
`queue.size`, `queue.max_age`, `metrics.prefix`, `metrics.tag` (`limiter.h`),
`buckets`, `size`, `max_age`, `perma-block.*` (`ip_reputation.cc`), `cidr`
(`lists.cc`), and `ip-rep` / `exclude` (`sni_limiter.cc`).

### On `percentage`

`doc/admin-guide/plugins/rate_limit.en.rst:272` gives the default as ``0.9``,
but the parser reads an integer and `_percentage` defaults to `90`
(`ip_reputation.h:249`). The documented value is the one that throws, so the
docs need `90` here.

Present on master (a2011c2fc). The unguarded conversions came in with #10559.

sinhaparth5:
`SniSelector::yamlParser()` guarded only `YAML::LoadFile`, so a malformed value
threw out of the function. On `traffic_ctl config reload` that unwinds into the
event loop from the management update continuation and terminates a running
server, bypassing the `else` branch that exists to log the failure and keep the
previous configuration.

The parsing moves into `parseYamlFile()` and `yamlParser()` becomes the
exception boundary, so a bad config is rejected and the running one kept.

It also checks for the `sni` key before reading it. `sni["sni"].IsSequence()`
throws `YAML::InvalidNode` on the const node, so a selector entry without an
`sni` key never reached the "selector node is not a map or without a name"
error that is already there for it.

The `percentage` documentation gave the default as `0.9`, but the parser reads
an integer and the default is `90`. The documented value is one of the ones
that throws, so the docs are corrected as well.

### Testing

New autest `rate_limit_yaml_reload` covers both shapes: a selector entry with no
`sni` key, and a fractional `percentage`.

Run against the plugin built without this change, ATS dies on the first
malformed reload. The reload command itself reports success, because the plugin
callback is dispatched fire-and-forget on `ET_TASK`, and then the health check
returns `000`, the JSONRPC socket refuses connections, and `diags.log` stops
mid-reload with no FATAL and no shutdown.

With this change the test passes: both configs are rejected, the errors are
logged, and ATS keeps serving. All 7 rate_limit autests pass.

Fixes: #13598

### DIFF
--- a/plugins/experimental/rate_limit/sni_selector.cc
+++ b/plugins/experimental/rate_limit/sni_selector.cc
@@ -38,7 +38,7 @@ SniSelector::yamlParser(const std::string &yaml_file)
   try {
     return parseYamlFile(yaml_file);
   } catch (YAML::BadFile const &e) {
-    TSError("[%s] Cannot load configuration file: %s.", PLUGIN_NAME, e.what());
+    TSError("[%s] Cannot load configuration file %s: %s.", PLUGIN_NAME, yaml_file.c_str(), e.what());
   } catch (std::exception const &e) {
     TSError("[%s] Failed to parse configuration file %s: %s.", PLUGIN_NAME, yaml_file.c_str(), e.what());
   }
@@ -50,8 +50,6 @@ bool
 SniSelector::parseYamlFile(const std::string &yaml_file)
 {
   YAML::Node config = YAML::LoadFile(yaml_file);
-
-  _yaml_file = yaml_file;
 
   // First build the Lists, if any
   const YAML::Node &lists = config["lists"];
@@ -120,10 +118,11 @@ SniSelector::parseYamlFile(const std::string &yaml_file)
 
   if (sel && sel.IsSequence()) {
     for (const auto &i : sel) {
-      const YAML::Node &sni = i;
+      const YAML::Node &sni      = i;
+      const YAML::Node  sni_node = sni.IsMap() ? sni["sni"] : YAML::Node{};
 
-      if (sni.IsMap() && sni["sni"] && !sni["sni"].IsSequence()) {
-        auto name = sni["sni"].as<std::string>();
+      if (sni_node && !sni_node.IsSequence()) {
+        auto name = sni_node.as<std::string>();
 
         if (nullptr != findLimiter(name)) {
           TSError("[%s] Duplicate SNIs being added (%s)", PLUGIN_NAME, name.c_str());
@@ -176,6 +175,7 @@ SniSelector::parseYamlFile(const std::string &yaml_file)
     }
   }
 
+  _yaml_file = yaml_file;
   Dbg(dbg_ctl, "Succesfully loaded YAML file: %s", yaml_file.c_str());
 
   return true;
--- a/plugins/experimental/rate_limit/sni_selector.cc
+++ b/plugins/experimental/rate_limit/sni_selector.cc
@@ -28,8 +28,8 @@ std::atomic<SniSelector *> SniSelector::_instance = nullptr;
 // YAML parser for the global YAML configuration (via plugin.config)
 //
 // This is the exception boundary for the configuration parsing. The node
-// accessors and conversions in parseYamlFile() throw on malformed input, and
-// this runs on the management update continuation during a config reload, so
+// accessors and conversions in parseYamlFile() throw on malformed input, and a
+// config reload runs this on an ET_TASK thread via ConfigUpdateCallback, so
 // letting an exception escape here would terminate the server.
 //
 bool
@@ -121,7 +121,7 @@ SniSelector::parseYamlFile(const std::string &yaml_file)
       const YAML::Node &sni      = i;
       const YAML::Node  sni_node = sni.IsMap() ? sni["sni"] : YAML::Node{};
 
-      if (sni_node && !sni_node.IsSequence()) {
+      if (sni.IsMap() && sni_node && !sni_node.IsSequence()) {
         auto name = sni_node.as<std::string>();
 
         if (nullptr != findLimiter(name)) {
--- a/plugins/experimental/rate_limit/sni_selector.cc
+++ b/plugins/experimental/rate_limit/sni_selector.cc
@@ -27,20 +27,29 @@ std::atomic<SniSelector *> SniSelector::_instance = nullptr;
 ///////////////////////////////////////////////////////////////////////////////
 // YAML parser for the global YAML configuration (via plugin.config)
 //
+// This is the exception boundary for the configuration parsing. The node
+// accessors and conversions in parseYamlFile() throw on malformed input, and
+// this runs on the management update continuation during a config reload, so
+// letting an exception escape here would terminate the server.
+//
 bool
 SniSelector::yamlParser(const std::string &yaml_file)
 {
-  YAML::Node config;
-
   try {
-    config = YAML::LoadFile(yaml_file);
+    return parseYamlFile(yaml_file);
   } catch (YAML::BadFile const &e) {
     TSError("[%s] Cannot load configuration file: %s.", PLUGIN_NAME, e.what());
-    return false;
   } catch (std::exception const &e) {
-    TSError("[%s] Unknown error while loading configuration file: %s.", PLUGIN_NAME, e.what());
-    return false;
+    TSError("[%s] Failed to parse configuration file %s: %s.", PLUGIN_NAME, yaml_file.c_str(), e.what());
   }
+
+  return false;
+}
+
+bool
+SniSelector::parseYamlFile(const std::string &yaml_file)
+{
+  YAML::Node config = YAML::LoadFile(yaml_file);
 
   _yaml_file = yaml_file;
 
@@ -113,7 +122,7 @@ SniSelector::yamlParser(const std::string &yaml_file)
     for (const auto &i : sel) {
       const YAML::Node &sni = i;
 
-      if (sni.IsMap() && !sni["sni"].IsSequence()) {
+      if (sni.IsMap() && sni["sni"] && !sni["sni"].IsSequence()) {
         auto name = sni["sni"].as<std::string>();
 
         if (nullptr != findLimiter(name)) {
--- a/plugins/experimental/rate_limit/sni_selector.cc
+++ b/plugins/experimental/rate_limit/sni_selector.cc
@@ -176,7 +176,7 @@ SniSelector::parseYamlFile(const std::string &yaml_file)
   }
 
   _yaml_file = yaml_file;
-  Dbg(dbg_ctl, "Succesfully loaded YAML file: %s", yaml_file.c_str());
+  Dbg(dbg_ctl, "Successfully loaded YAML file: %s", yaml_file.c_str());
 
   return true;
 }
--- a/tests/gold_tests/pluginTest/rate_limit/rate_limit_yaml_reload.test.py
+++ b/tests/gold_tests/pluginTest/rate_limit/rate_limit_yaml_reload.test.py
@@ -19,6 +19,7 @@ terminating traffic_server.
 #  limitations under the License.
 
 import os
+import shlex
 
 Test.Summary = '''
 rate_limit: a malformed YAML config fails the reload without killing ATS.
@@ -85,7 +86,7 @@ tr.StillRunningAfter = ts
 # each overwrite to make sure it does.
 for description, bad_config in [("selector entry without an sni key", missing_sni), ("a fractional percentage", bad_percentage)]:
     tr = Test.AddTestRun(f"Install {description}")
-    tr.Processes.Default.Command = f"sleep 2 && cp {bad_config} {rate_limit_yaml}"
+    tr.Processes.Default.Command = f"sleep 2 && cp {shlex.quote(bad_config)} {shlex.quote(rate_limit_yaml)}"
     tr.Processes.Default.ReturnCode = 0
     tr.StillRunningAfter = ts
 
--- a/tests/gold_tests/pluginTest/rate_limit/rate_limit_yaml_reload.test.py
+++ b/tests/gold_tests/pluginTest/rate_limit/rate_limit_yaml_reload.test.py
@@ -1,0 +1,107 @@
+'''
+Test that a malformed rate_limit YAML file fails the reload instead of
+terminating traffic_server.
+'''
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


brbzull0:
[approve ci]

plugins/experimental/rate_limit/sni_selector.cc
```diff
@@ -27,21 +27,30 @@ std::atomic<SniSelector *> SniSelector::_instance = nullptr;
...
 
+  return false;
+}
+
+bool
+SniSelector::parseYamlFile(const std::string &yaml_file)
+{
+  YAML::Node config = YAML::LoadFile(yaml_file);
+
   _yaml_file = yaml_file;
```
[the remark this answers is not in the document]
sinhaparth5: Reload parsing is performed on a newly allocated SniSelector, and SniSelector::swap(new_sel) is called only after yamlParser() succeeds. On failure, new_sel is deleted, so partial mutations cannot affect the active configuration. Startup likewise parses into a fresh instance and terminates on failure. I nevertheless moved the _yaml_file assignment to the successful end of parseYamlFile() in d70059b77 to keep the object invariant clear.

tests/gold_tests/pluginTest/rate_limit/rate_limit_yaml_reload.test.py
```diff
@@ -0,0 +1,107 @@
...
+tr.Processes.Default.Command = CURL
+tr.Processes.Default.ReturnCode = 0
+tr.Processes.Default.Streams.stdout = Testers.ContainsExpression("200", "ATS should serve the request")
+tr.StillRunningAfter = ts
+
+# The plugin callback only fires when the file mtime advances, so sleep before
+# each overwrite to make sure it does.
+for description, bad_config in [("selector entry without an sni key", missing_sni), ("a fractional percentage", bad_percentage)]:
+    tr = Test.AddTestRun(f"Install {description}")
+    tr.Processes.Default.Command = f"sleep 2 && cp {bad_config} {rate_limit_yaml}"
```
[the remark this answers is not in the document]
sinhaparth5: Updated in d70059b77. Both paths interpolated into the cp command now use shlex.quote().

plugins/experimental/rate_limit/sni_selector.cc
```diff
@@ -27,21 +27,30 @@ std::atomic<SniSelector *> SniSelector::_instance = nullptr;
...
 bool
 SniSelector::yamlParser(const std::string &yaml_file)
 {
-  YAML::Node config;
-
   try {
-    config = YAML::LoadFile(yaml_file);
+    return parseYamlFile(yaml_file);
   } catch (YAML::BadFile const &e) {
     TSError("[%s] Cannot load configuration file: %s.", PLUGIN_NAME, e.what());
```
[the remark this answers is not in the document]
sinhaparth5: Updated in d70059b77. The YAML::BadFile diagnostic now includes yaml_file.

plugins/experimental/rate_limit/sni_selector.cc
```diff
@@ -113,7 +122,7 @@ SniSelector::yamlParser(const std::string &yaml_file)
     for (const auto &i : sel) {
       const YAML::Node &sni = i;
 
-      if (sni.IsMap() && !sni["sni"].IsSequence()) {
+      if (sni.IsMap() && sni["sni"] && !sni["sni"].IsSequence()) {
         auto name = sni["sni"].as<std::string>();
```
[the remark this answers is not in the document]
sinhaparth5: Updated in d70059b77. The sni child node is now stored once, while preserving the IsMap() guard before indexing the parent node.

sinhaparth5:
Implemented the Copilot review recommendations in commit d70059b77: deferred the YAML filename update until parsing succeeds, improved the BadFile diagnostic, avoided repeated sni node lookups, and quoted the test file paths. I also replied to each review thread with the corresponding details.

brbzull0 (commented):
Reviewed the reload path end to end and ran the plugin locally against both commits. The exception boundary is the right fix and the scope is complete -- `sni_selector.cc` is the only `YAML::LoadFile` in the plugin, and the remap path takes argv rather than YAML.

One blocking item: the `sni_node` hoist in `d70059b7` drops the `sni.IsMap()` guard, which makes a malformed selector entry get accepted and swapped in rather than rejected. Details and a verified one-line fix inline.

Also confirmed the `_yaml_file` deferral is safe -- it's never read during parsing, only in `sni_config_cont()` after the call returns.

plugins/experimental/rate_limit/sni_selector.cc
```diff
@@ -111,10 +118,11 @@ SniSelector::yamlParser(const std::string &yaml_file)
 
   if (sel && sel.IsSequence()) {
     for (const auto &i : sel) {
-      const YAML::Node &sni = i;
+      const YAML::Node &sni      = i;
+      const YAML::Node  sni_node = sni.IsMap() ? sni["sni"] : YAML::Node{};
 
-      if (sni.IsMap() && !sni["sni"].IsSequence()) {
-        auto name = sni["sni"].as<std::string>();
+      if (sni_node && !sni_node.IsSequence()) {
```
brbzull0: issue: (blocking)

Hoisting the lookup is a good change, but moving `sni.IsMap()` into the ternary isn't equivalent to keeping it in the condition, and it stops non-map selector entries from being rejected.

`YAML::Node{}` is truthy -- the default constructor leaves `m_isValid == true` with a null `m_pNode`, so `operator bool()` -> `IsDefined()` returns true, and `IsSequence()` -> `Type()` returns `NodeType::Null`. A non-map entry therefore satisfies `sni_node && !sni_node.IsSequence()` and enters the branch that the `else` used to catch. `as<std::string>()` doesn't throw there either -- `as_if<std::string, void>` returns the literal `"null"` for a Null node -- so a limiter named `"null"` is constructed and parsing continues to a successful return.

I built both commits and reloaded the same config, an entry with one level of extra indentation:

```yaml
selector:
  - - sni: indented.example.com
      limit: 5
```

At `341a0cf4`:
```
[ET_TASK 0] ERROR: [rate_limit] selector node is not a map or without a name
[ET_TASK 0] ERROR: [rate_limit] Failed to reload YAML file: .../rate_limit.yaml
```

At `d70059b7`:
```
[ET_TASK 0] <sni_selector.cc:179 (parseYamlFile)> Succesfully loaded YAML file: .../rate_limit.yaml
[ET_TASK 0] <sni_selector.cc:206 (sni_config_cont)> Reloading YAML file: .../rate_limit.yaml
```

`parseYamlFile()` returns true and `sni_config_cont()` takes the success branch, so `swap()` installs a selector whose only entry is the `"null"` limiter. Every configured SNI limit is dropped, `traffic_ctl config reload` reports success, and nothing lands in `diags.log`. On a plugin whose job is enforcing limits, an indentation slip silently disabling enforcement seems worth catching before this merges.

Restoring the map check in the condition keeps the hoist and the original behavior:

```suggestion
      if (sni.IsMap() && sni_node && !sni_node.IsSequence()) {
```

I applied exactly that on top of `d70059b7` and re-ran: the indentation config is rejected again, `Reloading YAML file` appears zero times so nothing is swapped in, this PR's `rate_limit_yaml_reload` test still passes, and all 8 `rate_limit` autests pass.

Worth a third case in the test loop as well, since the existing two are both maps and take the `else` correctly either way:

```python
non_map = os.path.join(Test.RunDirectory, 'non_map.yaml')
with open(non_map, 'w') as f:
    f.write('selector:\n  - - sni: indented.example.com\n      limit: 5\n')
```

tests/gold_tests/pluginTest/rate_limit/rate_limit_yaml_reload.test.py
```diff
@@ -0,0 +1,108 @@
...
+
+    tr = Test.AddTestRun(f"ATS survives {description}")
+    tr.Processes.Default.Command = f"sleep 2 && {CURL}"
+    tr.Processes.Default.ReturnCode = 0
+    tr.Processes.Default.Streams.stdout = Testers.ContainsExpression("200", "ATS should still be serving traffic")
+    tr.StillRunningAfter = ts
+
+# Both cases have to be reported and rejected. Assigning here replaces the
+# default "diags.log has no ERROR:" testers, since these errors are expected.
+ts.Disk.diags_log.Content = Testers.ContainsExpression(
```
brbzull0, line 111: suggestion:

The test can't currently detect the thing it's guarding, which is why the `sni_node` change in this commit passes it.

Three gaps:

- **The only positive signal is `curl` returning 200**, and the origin returns 200 whether or not the rate_limit config applied. So the test can't distinguish "config rejected, previous one kept" from "malformed config accepted and the limits silently dropped" -- and the second is the outcome that actually hurts.
- **These are existence checks over the whole log, not per-reload assertions.** If the first reload is rejected and the second is silently accepted, `"Failed to reload YAML file"` is still present once and all four testers pass, despite the comment saying "Both reloads should be rejected".
- **`ExcludesExpression("FATAL")` doesn't catch the original crash.** An uncaught exception reaching the event loop is `std::terminate`/SIGABRT -- there is no `try`/`catch` in `UnixEThread.cc` -- so nothing is written to diags.log at all; the log just stops. `StillRunningAfter` and the 200 are what actually detect it. This tester reads like a guard but isn't one.

A direct assertion that no swap happened would close all three, since `sni_config_cont()` only logs that on the success path:

```python
ts.Disk.traffic_out.Content = Testers.ExcludesExpression(
    "Reloading YAML file", "No malformed config should ever be swapped in")
```

plugins/experimental/rate_limit/sni_selector.cc
```diff
@@ -27,22 +27,29 @@ std::atomic<SniSelector *> SniSelector::_instance = nullptr;
 ///////////////////////////////////////////////////////////////////////////////
 // YAML parser for the global YAML configuration (via plugin.config)
 //
+// This is the exception boundary for the configuration parsing. The node
+// accessors and conversions in parseYamlFile() throw on malformed input, and
+// this runs on the management update continuation during a config reload, so
```
brbzull0: nitpick: (non-blocking)

"the management update continuation" reads as though this runs on the management/RPC thread. It's `ConfigUpdateCallback` on an `ET_TASK` thread -- `ConfigUpdateCbTable::invoke()` does `schedule_imm(new ConfigUpdateCallback(contp), ET_TASK)`, two hops from the RPC handler, which is why the original failure presented so oddly: the reload had already reported success by the time the process died. The `ET_TASK 0` prefix on the plugin's own error lines shows it.

Something like "on an ET_TASK thread, via `ConfigUpdateCallback`" would point the next reader at the right place.

plugins/experimental/rate_limit/sni_selector.cc
```diff
@@ -27,22 +27,29 @@ std::atomic<SniSelector *> SniSelector::_instance = nullptr;
...
+// letting an exception escape here would terminate the server.
+//
 bool
 SniSelector::yamlParser(const std::string &yaml_file)
 {
-  YAML::Node config;
-
   try {
-    config = YAML::LoadFile(yaml_file);
+    return parseYamlFile(yaml_file);
```
brbzull0, line 39: praise:

Making the whole parse the guarded region rather than just `LoadFile` is the right call, and it fixes more than the two cases in the test -- `list["name"].as<std::string>()` and `ipr["name"].as<std::string>()` in the `lists` and `ip-rep` loops throw on a non-scalar too, and were reaching the event loop the same way.

The const-node diagnosis behind the `sni` key check was also exactly right: the const `operator[]` returns a `ZombieNode`, `IsSequence()` goes through `Type()` which throws `InvalidNode`, while `operator bool()` goes through `IsDefined()` which returns false without throwing.

plugins/experimental/rate_limit/sni_selector.cc
sinhaparth5: Fixed in ae9257ba2, applying the suggestion as written. I kept the ternary so `sni["sni"]` is still only read on a map node.

Confirmed the Null node behaviour: `as_if<std::string, void>` returns `"null"` for it, so the entry was accepted instead of throwing, and every configured limit went with it on the swap. The indented config is now a third case in the test loop.

tests/gold_tests/pluginTest/rate_limit/rate_limit_yaml_reload.test.py
sinhaparth5: Added the `traffic_out` exclusion in ae9257ba2. `sni_config_cont()` logs `Reloading YAML file` only on the success path, right before `swap()`, so that is the assertion the `sni_node` regression fails.

Dropped the `ExcludesExpression("FATAL")` tester for the reason you give: the abort writes nothing, so `StillRunningAfter` and the 200 are what catch it, and the tester only looked like a guard.

The per-reload point stands, the existence checks still pass if one reload is rejected and another is not. The exclusion covers all three at once, so a silent accept fails now regardless of which one it was.

plugins/experimental/rate_limit/sni_selector.cc
sinhaparth5: Reworded in ae9257ba2 to "a config reload runs this on an ET_TASK thread via `ConfigUpdateCallback`". That points at the right place, and it explains why the reload reported success before the process died.

sinhaparth5:
Pushed ae9257ba2 with the three items from the review:

- Restored `sni.IsMap()` in the condition, keeping the hoist.
- Added the non-map selector entry as a third case in the test loop, and the `traffic_out` exclusion on `Reloading YAML file` so a silent accept fails instead of passing on the 200s. Dropped the `FATAL` tester.
- Reworded the comment to name the ET_TASK thread and `ConfigUpdateCallback`.

brbzull0:
[approve ci]

sinhaparth5:
Fixed the typo in 1350cc3a0.

On the license header: I left the docstring above it, since that is what the rest of the suite does. Every other `.test.py` in `tests/gold_tests/pluginTest/rate_limit/` starts with the docstring, and repo-wide it is 474 files docstring-first against 97 license-first. Happy to flip it if the project would rather standardize the other way.

PR #13601, opened 2026-08-28
rate_limit: fix swapped metric tag and prefix in YAML configs

### LINKED ISSUE #13599: rate_limit: YAML metrics node swaps the metric tag and prefix
`RateLimiter::initializeMetrics()` takes the tag before the prefix:

```cpp
void initializeMetrics(uint type, std::string tag, std::string prefix = RATE_LIMITER_METRIC_PREFIX)
```

The remap path passes them in that order (`txn_limiter.cc:182`). The YAML path
does not (`limiter.h:228`):

```cpp
std::string prefix = metrics["prefix"] ? metrics["prefix"].as<std::string>() : RATE_LIMITER_METRIC_PREFIX;
std::string tag    = metrics["tag"] ? metrics["tag"].as<std::string>() : name();

initializeMetrics(RATE_LIMITER_TYPE_SNI, prefix, tag);
```

`metric_helper()` builds the name as `prefix + "." + type + "." + tag`
(`limiter.cc:60-74`), so the two halves come out reversed. Every SNI limiter
configured with a `metrics` node gets the wrong metric name, including when
only one of the two keys is set, because the defaults get swapped along with
them.

### Reproduce

```yaml
selector:
  - sni: example.com
    limit: 100
    metrics:
      prefix: myprefix
      tag: mytag
```

`traffic_ctl metric match rate` reports:

```
mytag.sni.myprefix.queued
```

instead of `myprefix.sni.mytag.queued`.

With only `tag: mytag` set, the name is `mytag.sni.plugin.rate_limiter.queued`
rather than `plugin.rate_limiter.sni.mytag.queued`. With only
`prefix: myprefix` set it is `example.com.sni.myprefix.queued` rather than
`myprefix.sni.example.com.queued`.

Present on master (a2011c2fc), dating to #10559. No autest covers the metric
names, which is why it has gone unnoticed.

sinhaparth5:
`RateLimiter::initializeMetrics()` takes the tag before the prefix:

```cpp
void initializeMetrics(uint type, std::string tag, std::string prefix = RATE_LIMITER_METRIC_PREFIX)
```

The remap path passes them in that order (`txn_limiter.cc:182`), but the YAML
path passed `(prefix, tag)`. `metric_helper()` builds the name as
`prefix + "." + type + "." + tag`, so every SNI limiter with a `metrics` node
came out as `tag.sni.prefix`.

This affects a `metrics` node even when only one of the two keys is set, since
the unset one takes its default and the two get swapped together.

### Testing

New autest `rate_limit_metric_names` covers all three shapes: both keys set,
only `tag` set, and only `prefix` set.

Run against the plugin built without this change, `traffic_ctl metric match`
reports the reversed names, which is what the test now pins:

```
onlytag.sni.plugin.rate_limiter.queued          (expected plugin.rate_limiter.sni.onlytag.queued)
prefixonly.example.com.sni.onlyprefix.queued    (expected onlyprefix.sni.prefixonly.example.com.queued)
```

All 7 rate_limit autests pass with the change.

Fixes: #13599

### DIFF
--- a/plugins/experimental/rate_limit/limiter.h
+++ b/plugins/experimental/rate_limit/limiter.h
@@ -224,8 +224,8 @@ public:
       std::string prefix = metrics["prefix"] ? metrics["prefix"].as<std::string>() : RATE_LIMITER_METRIC_PREFIX;
       std::string tag    = metrics["tag"] ? metrics["tag"].as<std::string>() : name();
 
-      Dbg(dbg_ctl, "Metrics for selector rule: %s(%s, %s)", name().c_str(), prefix.c_str(), tag.c_str());
-      initializeMetrics(RATE_LIMITER_TYPE_SNI, prefix, tag);
+      Dbg(dbg_ctl, "Metrics for selector rule: %s (prefix=%s, tag=%s)", name().c_str(), prefix.c_str(), tag.c_str());
+      initializeMetrics(RATE_LIMITER_TYPE_SNI, tag, prefix);
     }
 
     return true;
--- a/tests/gold_tests/pluginTest/rate_limit/rate_limit_metric_names.test.py
+++ b/tests/gold_tests/pluginTest/rate_limit/rate_limit_metric_names.test.py
@@ -1,0 +1,97 @@
+'''
+Test that the rate_limit YAML metrics node names metrics as prefix.type.tag.
+'''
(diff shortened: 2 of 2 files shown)


cmcfarlen (approved):
Good catch!

Issue #9636, opened 2023-04-25
☂ Cleanup, add and remove WKS's, making them modern


zwoop:
- [ ] Various old protocol scheme WKS (e.g. NNTP and the streaming protocols). Keep e.g. FTP and other scheme that are good.

PR #13603, opened 2026-08-31
Rebuild well-known string indexes when reading a cached object

bneradt:
Unblocks #9636 by decoupling well-known string (WKS) indexes from the on-disk cache format.

This PR incorporates cache compatibility fixes and tests from **Jake Champion (@JakeChampion), originally contributed in #13637**. Jake is also credited as a co-author of the integration commit.

## Cache compatibility

Upgrading preserves existing cached objects. Stripe validation still checks the major version, and objects older than 24.2 retain their legacy fragment-table decoder. The decoder boundary is fixed at 24.2 rather than following the current cache version; a compile-time assertion protects that boundary.

The minor version becomes 24.3 so older ATS builds reject newly written objects and refetch them instead of interpreting changed WKS indexes. Rolling back does not invalidate objects written before the upgrade. Shared-memory schema versioning remains separate.

## WKS repair and fast path

Cached headers store WKS indexes for field names, the request method, and URL scheme alongside the original strings. A different table can make those indexes—and the derived MIME presence bits and slot accelerators—incorrect.

Newly marshalled alternates carry a distinct tag and an eight-byte compatibility identity trailer. The identity covers the ordered WKS strings, MIME slot IDs, and presence masks. Matching identities skip the field walk and tokenization. Missing or mismatched identities rebuild from the stored strings after both header heaps are unmarshalled. Already-live buffers remain untouched, including buffers shared by RAM-cache readers.

The fixed-size alternate layout and existing header offsets are unchanged. Legacy alternates have no trailer and always rebuild. The identity is a compatibility check, not an integrity checksum.

This permits changes to WKS indexes and cardinality. It does not reconstruct persisted cooked Cache-Control metadata; changes to recognized directives or cooked masks require a separate compatibility decision.

## Other fixes

Jake's changes unify cache scanning and normal reads on the version-aware decoder, reject oversized header lengths and truncated alternate tails, and remove the unused alternate unmarshaller. A failed unmarshal also marks the document corrupt before returning to the cache reader.

`traffic_cache_tool` shares the server's cache-version constants, accepts supported 24.3 stripe metadata, selects the correct fragment-table layout, and consumes identity trailers while reading URLs from their stored strings.

## Validation

Validation in `asfats5` includes the full build/install and formatting, 21 cache/header/cache-tool CTest targets, 16 selected cache AuTest files, the documentation build, and targeted Clang analysis of `CacheVC.cc`.

The tests cover identity matches and mismatches, legacy objects without identities, shared buffers, both fragment layouts, multiple alternates, and truncated headers/trailers. The WKS test hook and unit scrambler now rotate fields in the inline MIME block as well as overflow blocks, and the hook writes a mismatched identity. The cache-tool test accepts 24.1–24.3 stripe metadata and rejects future versions.

The WKS string changes from #9636 remain a separate follow-up.

### DIFF
--- a/doc/developer-guide/cache-architecture/architecture.en.rst
+++ b/doc/developer-guide/cache-architecture/architecture.en.rst
@@ -483,6 +483,47 @@ default). Objects which are in use when the write cursor is near use the same
 underlying evacuation mechanism but are handled automatically and not via the
 explicit ``pinned`` bit in :cpp:class:`Dir`.
 
+Object Versioning
+-----------------
+
+Every ``Doc`` records the cache format version that wrote it, in its ``v_major``
+and ``v_minor`` fields, taken from ``CACHE_DB_MAJOR_VERSION`` and
+``CACHE_DB_MINOR_VERSION`` in ``iocore/cache/CacheDefs.h``.
+
+Bumping the minor version does not clear the cache. Stripe validation looks only
+at the major version, and the current reader still reads every object written at
+an older minor version. What the bump buys is protection in the other direction:
+a reader rejects any object newer than itself and refetches it, rather than
+misreading a shape it does not understand.
+
+Reading an older object sometimes needs work that reading a current one does
+not. Compare against **the fixed version at which that part of the format
+changed**, never against ``CACHE_DB_VERSION``. The latter silently changes
+meaning at the next bump, and sends every object the previous release wrote down
+the wrong path. ``CACHE_DB_FRAG_OFFSET_TABLE_VERSION`` is such a fixed point.
+
+Well-Known Strings
+------------------
+
+A marshalled header stores indexes into the well-known string table
+(``proxy/hdrs/HdrToken.cc``) beside the strings those indexes stand for: the
+index of every MIME field name, of the request method, and of the request URL
+scheme, plus the presence bits and slot accelerators derived from them. Change
+the table and every stored index denotes a different string.
+
+The strings are in the object too, so the indexes are only a cache over them.
+``HTTPInfo::unmarshal()`` rebuilds all of it through
+``HTTPHdrImpl::recompute_wks_indices()`` before anything reads the header,
+unconditionally rather than on a version test, since an object written by a
+same-version build with a different table needs the same treatment as an older
+one. The ``CacheAltMagic`` check keeps this to once per marshalled buffer, on a
+read that already paid for disk I/O or a RAM-cache decompression.
+
+The table is therefore free to change without invalidating anyone's cache. The
+one requirement is that a |TS| predating the rebuild never read an object
+written against a different table; cache version 24.3 is where the rebuild
+landed, and older versions reject anything newer than themselves.
+
 Additional Notes
 ----------------
 
--- a/include/iocore/cache/CacheDefs.h
+++ b/include/iocore/cache/CacheDefs.h
@@ -37,12 +37,34 @@ enum class CacheInitState : int {
 #define CACHE_ALT_INDEX_DEFAULT -1
 #define CACHE_ALT_REMOVED       -2
 
+// Bumping the minor version does not clear anyone's cache: stripe validation looks only at the
+// major version, and this build still reads every object written at an older minor version. What
+// it does mean is that an ATS older than this treats the objects this build writes as corrupt and
+// refetches them, so bump it whenever an object gains a shape an older ATS would misread.
+//
+// 24.2 marshalled the fragment offset table in full; see CACHE_DB_FRAG_OFFSET_TABLE_VERSION below.
+// 24.3 stopped trusting the well-known string indexes stored in an object and started rebuilding
+// them from the header strings stored alongside them, in HTTPHdrImpl::recompute_wks_indices().
+// That is what frees the well-known string table in proxy/hdrs/HdrToken.cc to change: any ATS at
+// 24.3 or newer reads objects written against any table, and anything older refuses them outright
+// rather than resolving their indexes against the wrong table.
 static const uint8_t CACHE_DB_MAJOR_VERSION = 24;
-static const uint8_t CACHE_DB_MINOR_VERSION = 2;
+static const uint8_t CACHE_DB_MINOR_VERSION = 3;
 // This is used in various comparisons because otherwise if the minor version is 0,
 // the compile fails because the condition is always true or false. Running it through
 // VersionNumber prevents that.
 extern const ts::VersionNumber CACHE_DB_VERSION;
+
+// The first version whose objects carry a complete fragment offset table. Before it, an object
+// with more than HTTPCacheAlt::N_INTEGRAL_FRAG_OFFSETS fragments marshalled only the offsets past
+// the integral ones, and the reader rebuilt the combined table; see HTTPInfo::unmarshal_v24_1().
+// Objects older than this need that reader, and objects from this version on need
+// HTTPInfo::unmarshal(). This is a fixed point in the format's history, not the current version:
+// comparing against CACHE_DB_VERSION instead would send every current object through the old
+// reader the moment the cache version is bumped for any other reason.
+static const uint8_t           CACHE_DB_FRAG_OFFSET_TABLE_MAJOR_VERSION = 24;
+static const uint8_t           CACHE_DB_FRAG_OFFSET_TABLE_MINOR_VERSION = 2;
+extern const ts::VersionNumber CACHE_DB_FRAG_OFFSET_TABLE_VERSION;
 
 static const uint8_t CACHE_DIR_MAJOR_VERSION = 18;
 static const uint8_t CACHE_DIR_MINOR_VERSION = 0;
--- a/src/proxy/hdrs/HTTP.cc
+++ b/src/proxy/hdrs/HTTP.cc
@@ -1966,6 +1966,26 @@ HTTPHdrImpl::unmarshal(intptr_t offset)
 }
 
 void
+HTTPHdrImpl::recompute_wks_indices()
+{
+  if (m_polarity == HTTPType::REQUEST) {
+    // http_hdr_method_get() answers from the index when it is set, so a stale index would report a
+    // different method than the one stored here. Tokenize case sensitively, exactly as the parser
+    // does, so a method that only matches a well-known string case insensitively stays untokenized.
+    u.req.m_method_wks_idx = u.req.m_ptr_method != nullptr ?
+                               static_cast<int16_t>(hdrtoken_method_tokenize(u.req.m_ptr_method, u.req.m_len_method)) :
+                               int16_t{-1};
+    if (u.req.m_url_impl != nullptr) {
+      u.req.m_url_impl->recompute_wks_idx();
+    }
+  }
+
+  if (m_fields_impl != nullptr) {
+    m_fields_impl->recompute_accelerators_and_presence_bits();
+  }
+}
+
+void
 HTTPHdrImpl::move_strings(HdrStrHeap *new_heap)
 {
   if (m_polarity == HTTPType::REQUEST) {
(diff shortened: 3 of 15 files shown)


src/proxy/hdrs/HTTP.cc
```diff
@@ -2262,6 +2306,8 @@ HTTPInfo::unmarshal(char *buf, int len, RefCountObj *block_ref)
     alt->m_response_hdr.m_mime = hh->m_fields_impl;
   }
 
+  recompute_alt_wks_indices(alt);
```
moonchen, line 2309: Could we avoid doing this unconditionally when the object was written against the same WKS table? A compact WKS compatibility identity persisted with newly written cache metadata would let the reader rebuild for pre-24.3 objects and for an identity mismatch, while trusting the stored indexes on a match. That preserves the flexibility this PR is aiming for without paying the field walk and tokenization cost in the common case.

src/proxy/hdrs/HdrHeap.cc
```diff
@@ -49,6 +52,95 @@ namespace
...
+        }
+      }
+      break;
+    }
+    case HdrHeapObjType::MIME_HEADER: {
+      MIMEHdrImpl *mh     = reinterpret_cast<MIMEHdrImpl *>(obj);
+      mh->m_presence_bits = MIME_PRESENCE_NONE;
+      for (uint32_t &accelerator : mh->m_slot_accelerators) {
+        accelerator = 0xFFFFFFFF;
+      }
```
moonchen, line 129: `m_first_fblock` is inline in `MIMEHdrImpl`, so this walk only reaches fields that overflowed into a separate `FIELD_BLOCK` object, which takes more than 16 fields. Every header in the autest fits in the first block, so no field index is actually rotated; only the presence bits and accelerators are cleared. Checked by dumping the marshalled heap with `ATS_TEST_WKS_IDX_SHIFT=7`: the method index moves 117 -> 124 while `Host`/`Accept-Encoding` stay at 30/1. Please rotate the live fields in `mh->m_first_fblock` here as well.

```suggestion
      for (uint32_t &accelerator : mh->m_slot_accelerators) {
        accelerator = 0xFFFFFFFF;
      }
      // The first field block is inline in the MIMEHdrImpl, so the walk never sees it as a FIELD_BLOCK.
      for (uint32_t i = 0; i < mh->m_first_fblock.m_freetop; ++i) {
        MIMEField &field = mh->m_first_fblock.m_field_slots[i];
        if (field.is_live()) {
          field.m_wks_idx = test_shift_wks_idx(field.m_wks_idx);
        }
      }
```

With this (and the same in `scramble_marshalled_heap`) the autest and both unit tests still pass here.

src/proxy/hdrs/unit_tests/test_Hdrs.cc
```diff
@@ -2902,6 +2902,198 @@ TEST_CASE("HTTPInfo::unmarshal_v24_1 frag bounds checks", "[proxy][hdrtest][unma
...
+      }
+      break;
+    }
+    case HdrHeapObjType::MIME_HEADER: {
+      MIMEHdrImpl *mh            = reinterpret_cast<MIMEHdrImpl *>(obj);
+      mh->m_presence_bits        = MIME_PRESENCE_NONE;
+      mh->m_slot_accelerators[0] = 0xFFFFFFFF;
+      mh->m_slot_accelerators[1] = 0xFFFFFFFF;
+      mh->m_slot_accelerators[2] = 0xFFFFFFFF;
+      mh->m_slot_accelerators[3] = 0xFFFFFFFF;
```
moonchen, line 2967: Same inline-block gap in the unit-test scrambler: with these test headers, the `FIELD_BLOCK` case never fires. Please rotate the live fields in `mh->m_first_fblock` so the serialized-data test verifies field-index reconstruction rather than only method/scheme indexes and the aggregate metadata.

```suggestion
      mh->m_slot_accelerators[3] = 0xFFFFFFFF;
      for (uint32_t i = 0; i < mh->m_first_fblock.m_freetop; ++i) {
        MIMEField &field = mh->m_first_fblock.m_field_slots[i];
        if (field.is_live()) {
          field.m_wks_idx = rotate_wks_idx(field.m_wks_idx);
        }
      }
```

src/proxy/hdrs/unit_tests/test_Hdrs.cc
```diff
@@ -2902,6 +2902,198 @@ TEST_CASE("HTTPInfo::unmarshal_v24_1 frag bounds checks", "[proxy][hdrtest][unma
...
+
+/// Rotate every well-known string index in one marshalled header heap.
+void
+scramble_marshalled_heap(HdrHeap *heap)
+{
+  char *obj_data = reinterpret_cast<char *>(heap) + sizeof(HdrHeap);
+  char *heap_end = reinterpret_cast<char *>(heap) + heap->m_size;
+
+  // Objects start at the marshalled heap's data offset, which marshal() sets to the header size.
+  obj_data = reinterpret_cast<char *>(heap) + reinterpret_cast<intptr_t>(heap->m_data_start);
```
moonchen, line 2932: The first `obj_data` initializer is dead because it is overwritten below. This can be a single initialization from `heap->m_data_start`.

src/proxy/hdrs/unit_tests/test_Hdrs.cc
```diff
@@ -2902,6 +2902,198 @@ TEST_CASE("HTTPInfo::unmarshal_v24_1 frag bounds checks", "[proxy][hdrtest][unma
...
+  CHECK(got_resp->value_get(static_cast<std::string_view>(MIME_FIELD_VARY)) == "Accept-Encoding"sv);
+  CHECK(got_resp->presence(MIME_PRESENCE_CACHE_CONTROL) != 0);
+  CHECK(got_resp->presence(MIME_PRESENCE_VARY) != 0);
+  // The cooked Cache-Control cache is keyed by directive name, so it survives independently, but
+  // it has to still agree with the rebuilt indices.
+  CHECK(got_resp->get_cooked_cc_mask() & MIME_COOKED_MASK_CC_MAX_AGE);
+  CHECK(got_resp->get_cooked_cc_max_age() == 300);
+
+  req.destroy();
+  resp.destroy();
```
moonchen, line 3094: `info` still owns the alternate allocated by `create()`; please add `info.destroy()` before leaving the test.

doc/developer-guide/cache-architecture/architecture.en.rst
```diff
@@ -483,6 +483,47 @@ default). Objects which are in use when the write cursor is near use the same
...
+
+The strings are in the object too, so the indexes are only a cache over them.
+``HTTPInfo::unmarshal()`` rebuilds all of it through
+``HTTPHdrImpl::recompute_wks_indices()`` before anything reads the header,
+unconditionally rather than on a version test, since an object written by a
+same-version build with a different table needs the same treatment as an older
+one. The ``CacheAltMagic`` check keeps this to once per marshalled buffer, on a
+read that already paid for disk I/O or a RAM-cache decompression.
+
+The table is therefore free to change without invalidating anyone's cache. The
```
moonchen, line 522: The index/cardinality changes motivating this PR are covered, but `m_cooked_stuff` is marshalled as-is. Reordering or reusing WKS IDs does not affect it; changing which Cache-Control directives are recognized, or changing their cooked masks, does. Could this qualify the statement that the table is free to change so it does not imply that those Cache-Control semantics are also reconstructed?

include/iocore/cache/CacheDefs.h
```diff
@@ -37,13 +37,35 @@ enum class CacheInitState : int {
...
+//
+// 24.2 marshalled the fragment offset table in full; see CACHE_DB_FRAG_OFFSET_TABLE_VERSION below.
+// 24.3 stopped trusting the well-known string indexes stored in an object and started rebuilding
+// them from the header strings stored alongside them, in HTTPHdrImpl::recompute_wks_indices().
+// That is what frees the well-known string table in proxy/hdrs/HdrToken.cc to change: any ATS at
+// 24.3 or newer reads objects written against any table, and anything older refuses them outright
+// rather than resolving their indexes against the wrong table.
 static const uint8_t CACHE_DB_MAJOR_VERSION = 24;
-static const uint8_t CACHE_DB_MINOR_VERSION = 2;
+static const uint8_t CACHE_DB_MINOR_VERSION = 3;
```
moonchen, line 52: `traffic_cache_tool` still defines its cache version as 24.1, and `validateMeta()` rejects stripe minor versions above 2. As a result it refuses a stripe written by this 24.3 code before reaching the WKS-safe URL scan added here. Please update the tool's accepted version handling and cover inspection of a 24.3 stripe.

moonchen (changes requested):
The compatibility direction looks sound, but I would like the WKS rebuild to have a cheap compatibility fast path before this lands. The current implementation adds a field-name walk and tokenization to every fresh unmarshal even though the usual writer and reader have the same WKS table. The existing test-validity issues, small unit-test cleanup, documentation scope, and traffic_cache_tool version support also remain.

src/proxy/hdrs/HTTP.cc
bneradt: Implemented in 98f47635ca. New alternates carry a distinct marshalled tag and an eight-byte identity covering the ordered WKS strings, MIME slot IDs, and presence masks. Matches skip rebuilding; legacy objects and mismatches rebuild after both heaps are swizzled. Tests cover matching and mismatching identities, legacy objects, shared buffers, multiple alternates, and truncated trailers.

src/proxy/hdrs/HdrHeap.cc
bneradt: Fixed in 98f47635ca. The test hook now rotates live fields in the inline first block as well as overflow blocks, and writes a mismatched identity so the reader takes the repair path. The WKS AuTest passes with the corrected hook.

src/proxy/hdrs/unit_tests/test_Hdrs.cc
bneradt: Fixed in 98f47635ca. The unit scrambler now rotates the inline first-block fields too, and the serialized alternate carries a mismatched identity. The header unit tests pass.

src/proxy/hdrs/unit_tests/test_Hdrs.cc
bneradt: Fixed in 98f47635ca: obj_data is initialized directly from m_data_start, without the dead assignment.

src/proxy/hdrs/unit_tests/test_Hdrs.cc
bneradt: Fixed in 98f47635ca: the test now destroys info before leaving.

doc/developer-guide/cache-architecture/architecture.en.rst
bneradt: Qualified in 98f47635ca. The documentation now explicitly limits this to WKS index/cardinality changes and explains that persisted m_cooked_stuff is not reconstructed; recognized Cache-Control directives and cooked masks require a separate compatibility decision. The documentation build passes.

include/iocore/cache/CacheDefs.h
bneradt: Fixed in 98f47635ca. The tool now shares the server version constants, accepts supported 24.3 metadata, selects the fragment-table layout by object version, and consumes WKS identity trailers. A new CTest covers supported and future stripe versions. I also exercised the offline decoder against the WKS AuTest cache: four 24.3 metadata records and five cached alternates decoded successfully.

moonchen (approved):
Thanks for addressing the review comments. The compatibility fast path and legacy cache handling look good, and the cache tool changes address my earlier concern. LGTM.

PR #13606, opened 2026-08-31
Track coalesced origin connections

### LINKED ISSUE #13605: Origin connect coalescing can bypass connection tracking
## Summary

An origin connection created through `ConnectingEntry` can be omitted from
`ConnectionTracker`. This makes the live per-server connection gauges
undercount real origin connections and can also let
`proxy.config.http.per_server.connection.max` operate on an incomplete count.

The problem is timing-dependent: it occurs when multiple transactions on one
ATS event thread target the same origin while the first origin connection is
still being established.

## Observed failure

The failure was captured by the multi-group portion of
`origin_connection/per_server_connection_max.test.py`. The test configures:

- `proxy.config.http.per_server.connection.metric_enabled: 1`
- `proxy.config.http.per_server.connection.metric_aggregate: 2`
- `proxy.config.http.per_server.connection.match: both`
- a 500 ms raw-stat synchronization interval
- two concurrent requests to `multi.origin.com:<origin-a-port>`
- three concurrent requests to `multi.origin.com:<origin-b-port>`
- six-second origin response delays so every connection overlaps multiple
  derived-metric update periods

The expected aggregate was:

```text
proxy.process.http.per_server.total_connection.multi.origin.com 5
proxy.process.http.per_server.current_connection.multi.origin.com 5
proxy.process.http.per_server.current_connection_max.multi.origin.com 3
```

After polling for ten seconds, CI consistently observed:

```text
proxy.process.http.per_server.current_connection.multi.origin.com 4
proxy.process.http.per_server.total_connection.multi.origin.com 5
proxy.process.http.per_server.blocked_connection.multi.origin.com 0
proxy.process.http.per_server.current_connection_max.multi.origin.com 2
```

Increasing the metric wait would not address this instance: the published
gauges remained at 4 and 2 until test teardown.

## Preserved sandbox evidence

The archived httpbin output proves all five requests reached the two origins
and overlapped. The timestamps below are response-completion times and the
reported durations are approximately six seconds:

```text
origin A:
19:04:25.6461 duration_ms=6000.55 GET /delay/6
19:04:25.6573 duration_ms=6000.87 GET /delay/6

origin B:
19:04:25.6483 duration_ms=6001.11 GET /delay/6
19:04:25.6607 duration_ms=6001.82 GET /delay/6
19:04:25.6607 duration_ms=6002.24 GET /delay/6
```

Thus the requests began between approximately 19:04:19.645 and
19:04:19.658. ATS also logged five distinct `Http1ServerSession::new_connection`
events during that interval.

Two origin-B requests landed on `ET_NET 3`. The relevant trace was:

```text
[19:04:19.656] [ET_NET 3] [SM 2] Check for existing connect request
[19:04:19.656] [ET_NET 3] [SM 2] Queue multiplexed request
[19:04:19.657] [ET_NET 3] [SM 3] Add entry to connection queue. size=2
[19:04:19.657] [ET_NET 3] [SM 3] Queue behind existing request
[19:04:19.657] [ET_NET 3] ConnectingEntry: Kick off 2 state machines waiting for origin
[19:04:19.657] [ET_NET 3] ConnectingEntry send CONNECT_EVENT_TXN to first 0
[19:04:19.657] [ET_NET 3] [SM 3] Connection handshake complete via CONNECT_EVENT_TXN
[19:04:19.657] [ET_NET 3] ConnectingEntry Pass along CONNECT_EVENT_DIRECT 1
[19:04:19.657] [ET_NET 3] [SM 2] open connection to multi.origin.com:<origin-b-port>
```

Five server sessions were created, but only four remained represented by the
connection tracker: two for origin A and two for origin B.

## Code path

`HttpSM::do_http_server_open()` calls `add_to_existing_request()` before it
obtains and reserves `t_state.outbound_conn_track_state`. A transaction that
joins an existing `ConnectingEntry` returns immediately and therefore has no
active tracker reservation.

When the connection completes, `ConnectingEntry::state_http_server_open()`
selects an arbitrary state machine from its `std::set<HttpSM *>` and calls that
state machine's `create_server_session()`. If it selects the queued state
machine, the new physical server session is created without tracker ownership,
because `HttpSM::create_server_session()` only transfers tracking when that
state machine's `outbound_conn_track_state` is active.

For HTTP/1, the other state machine is sent through `CONNECT_EVENT_DIRECT` and
reserves again. This explains why the cumulative total reached 5 while the
instantaneous count was only 4: the cumulative metric received the retry
reservation, while the physical session created from the unreserved queued
state machine remained untracked.

For a multiplexing origin, the same ownership mismatch can affect the single
connection shared by all queued transactions.

## Expected behavior

The reservation made for the physical connection owned by a
`ConnectingEntry` should remain with that entry until the connection either:

1. succeeds and transfers the reservation to the resulting
   `PoolableSession`, or
2. fails or is cancelled and releases the reservation.

The state machine selected to receive the completed connection should not
determine whether the physical connection is tracked.

## Reproduction characteristics

The exact Uranium scenario passed in six local attempts (five normal runs and
one run with a single ATS event thread). The CI failure depended on two requests
being scheduled on the same event thread during the narrow origin-connect
window. The sandbox trace above captures that path directly, so this should not
be treated as a metric synchronization timeout or fixed by weakening the
expected values.

bneradt:
Connection coalescing could assign an origin socket to a queued
transaction that did not own its connection-tracker reservation. The live
connection was then absent from per-server limits and metrics.

This patch keeps the reservation with the in-progress connection and
transfers it to the established session. Failed connection attempts
release the reservation from the coalescing entry.

Fixes: #13605

### DIFF
--- a/include/proxy/http/ConnectingEntry.h
+++ b/include/proxy/http/ConnectingEntry.h
@@ -35,6 +35,11 @@ class ConnectingEntry : public Continuation
 public:
   ConnectingEntry() = default;
   ~ConnectingEntry() override;
+  /** Retain a connection-tracker reservation until the connection completes.
+   *
+   * @param group Reserved outbound connection group owned by this entry.
+   */
+  void                    enable_outbound_connection_tracking(std::shared_ptr<ConnectionTracker::Group> group);
   void                    remove_entry();
   int                     state_http_server_open(int event, void *data);
   static PoolableSession *create_server_session(HttpSM *root_sm, NetVConnection *netvc, MIOBuffer *netvc_read_buffer,
@@ -51,9 +56,10 @@ public:
   bool               is_no_plugin_tunnel = false;
 
 private:
-  MIOBuffer      *_netvc_read_buffer = nullptr;
-  IOBufferReader *_netvc_reader      = nullptr;
-  Action         *_pending_action    = nullptr;
+  std::shared_ptr<ConnectionTracker::Group> _conn_track_group;
+  MIOBuffer                                *_netvc_read_buffer = nullptr;
+  IOBufferReader                           *_netvc_reader      = nullptr;
+  Action                                   *_pending_action    = nullptr;
 };
 
 struct IpHelper {
--- a/src/proxy/http/ConnectingEntry.cc
+++ b/src/proxy/http/ConnectingEntry.cc
@@ -37,10 +37,20 @@ DbgCtl dbg_ctl_http_connect{"http_connect"};
 
 ConnectingEntry::~ConnectingEntry()
 {
+  if (_conn_track_group) {
+    _conn_track_group->release();
+  }
   if (_netvc_read_buffer != nullptr) {
     free_MIOBuffer(_netvc_read_buffer);
     _netvc_read_buffer = nullptr;
   }
+}
+
+void
+ConnectingEntry::enable_outbound_connection_tracking(std::shared_ptr<ConnectionTracker::Group> group)
+{
+  ink_assert(!_conn_track_group);
+  _conn_track_group = std::move(group);
 }
 
 int
@@ -85,8 +95,11 @@ ConnectingEntry::state_http_server_open(int event, void *data)
       auto prime_iter = connect_sms.rbegin();
       ink_release_assert(prime_iter != connect_sms.rend());
       PoolableSession *new_session = (*prime_iter)->create_server_session(*netvc, _netvc_read_buffer, _netvc_reader);
-      netvc                        = nullptr;
-      _netvc_read_buffer           = nullptr;
+      if (_conn_track_group) {
+        new_session->enable_outbound_connection_tracking(std::move(_conn_track_group));
+      }
+      netvc              = nullptr;
+      _netvc_read_buffer = nullptr;
 
       // Did we end up with a multiplexing session?
       int count = 0;
--- a/src/proxy/http/HttpSM.cc
+++ b/src/proxy/http/HttpSM.cc
@@ -6014,6 +6014,7 @@ HttpSM::do_http_server_open(bool raw, bool only_direct)
       new_entry->sni                 = this->get_outbound_sni();
       new_entry->cert_name           = this->get_outbound_cert();
       new_entry->is_no_plugin_tunnel = plugin_tunnel_type == HttpPluginTunnel_t::NONE;
+      new_entry->enable_outbound_connection_tracking(t_state.outbound_conn_track_state.drop());
       this->t_state.set_connect_fail(EIO);
       new_entry->connect_sms.insert(this);
       ethread->connecting_pool->m_ip_pool.insert(std::make_pair(new_entry->ipaddr, new_entry));


cmcfarlen (changes requested):
Traced the ownership chain for the reservation through all of the entry's exit paths. The handoff itself is sound — `std::shared_ptr<ConnectionTracker::Group>` is already the currency for this transfer (`TxnState::_g` → `PoolableSession::conn_track_group`, and `UnixNetVConnection::conn_track_group` on the inbound side), and `ConnectingEntry` becoming a third holder with a matching `enable_*` signature fits that pattern. Each exit path releases exactly once: the success path `std::move`s into the session so the destructor no-ops, and both delete paths (`delete this` in the failure branch, `delete connecting_entry` in `HttpSM::cancel_pending_server_connection`) run the destructor release. No double-release, no missed release.

The problem is *when* the release happens, not whether.

## Reservation outlives the failed attempt, throttling the retry

In the `VC_EVENT_INACTIVITY_TIMEOUT` / `VC_EVENT_ACTIVE_TIMEOUT` / `VC_EVENT_ERROR` / `NET_EVENT_OPEN_FAILED` branch, the only release is `~ConnectingEntry` at `delete this` — which runs *after* the loop that dispatches the failure event to every queued `HttpSM`. Those SMs retry synchronously on the default configuration:

- `proxy.config.http.connect_attempts_retry_backoff_base` defaults to `0` (`RecordsConfig.cc:520`), so `StateMachineAction_t::ORIGIN_SERVER_OPEN` calls `do_http_server_open()` inline (`HttpSM.cc:8437`) instead of scheduling it.
- `do_http_server_open` re-obtains and re-`reserve()`s the group, so a second reservation is taken while the entry still holds the dead attempt's slot.

The result is two live reservations for one attempted connection. With `per_server.connection.max` at the boundary, `ccount > server_max`, so the retry hits `blocked_connection` / `Warn_Blocked` / `send_origin_throttled_response()` and returns a throttle response rather than retrying.

Worth noting *why* nothing else in the tree absorbs this. Pre-patch, retry safety came from the `t_state.outbound_conn_track_state.clear()` at the top of `do_http_server_open` (`HttpSM.cc:5663`), whose comment says it is there so "the selected group is consistent with the actual upstream in case of retry." Once `drop()` moves the group out of the SM, that `clear()` is a no-op — the SM no longer owns the thing that needs clearing. The safety mechanism ends up attached to the wrong holder.

Suggested fix: give `ConnectingEntry` a named `release_outbound_connection_tracking()` mirroring `PoolableSession.h:222`, and call it at the top of the error branch, before the dispatch loop. Keep the destructor call as a backstop.

## Same ordering on the certificate-validation path

In the success branch, when every queued SM fails `validate_server_certificate_hostname`, each is sent `CONNECT_EVENT_DIRECT` and re-enters `do_http_server_open` inline — reserving again — while the reservation has already been moved into `new_session`, which isn't closed until after the loop. For `Http2ServerSession` the release is later still: `do_io_close` defers to `release_stream`, and the release lives in `destroy()` (`Http2ServerSession.cc:49`), so for a session that never received a stream it isn't obvious the release happens at all. Same spurious-throttling shape, and this path becomes load-bearing more often now, since pre-patch the session only carried a group when the prime SM happened to be the entry's creator.

## Only the creating transaction's configuration is used

The entry carries the *creating* SM's reservation, and `connection_tracker_config` is overridable per remap. A transaction that joins an existing entry returns from `add_to_existing_request()` at `HttpSM.cc:5848`, before the obtain/reserve block at `:5874`, so it contributes nothing. If the creating transaction had tracking disabled (`server_max == 0 && server_min == 0 && !metric_enabled`) while the joiners have it enabled, the resulting shared origin connection stays invisible to per-server limits and metrics — #13605 persists for that mixed-configuration case. Conversely, the connection is accounted to the creator's hostname / match-type group rather than the joiner's. Probably acceptable as a documented limitation, but it should be a deliberate call rather than a silent one.

## Smaller items

- The new `enable_outbound_connection_tracking` call on `new_session` has no guard against `create_server_session` having already installed a group, and `PoolableSession::enable_outbound_connection_tracking` opens with `ink_assert(nullptr == conn_track_group)`. It's safe today only because joining SMs bail at `HttpSM.cc:5848` before ever obtaining a group, so `is_active()` is false at `:1854`. If that early return ever moves below the obtain block, debug builds assert and release builds silently overwrite the `shared_ptr` — dropping a reservation without releasing it, so `_count` never returns to zero and that origin is throttled permanently. Worth a comment naming the invariant.
- The new call site does `drop()` unconditionally while `HttpSM.cc:1854` guards with `if (is_active())`. Both are correct, since a null `shared_ptr` stores and releases harmlessly, but the asymmetry makes a reader stop and check.
- `PoolableSession` and `UnixNetVConnection` both expose a named release helper; this destructor releases inline. The helper suggested above would keep the three holders symmetric.
- No test. Given the `Fixes:` line and that the behavior is observable through the `per_server` metrics, is there a practical way to pin it — an HTTP/2 origin with concurrent requests through Proxy Verifier, checking `current_connection` returns to zero after a failed connect?

Since `_count` never being decremented, or being decremented twice, both degrade into origin-wide throttling that outlives the transaction, I'd like the release ordering sorted before this lands. The structure of the fix is right.

PR #13608, opened 2026-09-01
Bound cache directory bucket walks in all builds

masaori335:
The loop protection in probe/overwrite/remove/dir_clean_bucket sat behind LOOP_CHECK_MODE, which is commented out and not settable from CMake, so production builds walked bucket chains unbounded. A cycle in dir_next spins an event thread while it holds the stripe mutex. 

I have never seen this infinite loop in production. This is defensive change and cleanup of the `LOOP_CHECK_MODE`.

Directory::max_bucket_depth() is the bound, `(DIR_DEPTH - 1) * buckets + 1`: init_segment() frees rows `1`..`DIR_DEPTH-1` of every bucket onto the free list and never row 0, so one chain holds at most every free entry of the segment plus its own head. Exceeding it proves a cycle, so no exact check is needed at the trigger. bucket_length() capped at a hardcoded 100 instead, below the longest legitimate chain, so a chain over that would make check_segment() report corruption and Stripe::_shm_directory_is_valid() reject a healthy shared memory attach.

Readers report and move on, writers repair. probe() throttles its report because it does not repair, so the loop lives until a writer clears it and every lookup that hashes to the bucket lands on the same trigger. insert()'s tail walk is the only chain walk it makes, so that is where a writer meets a cycle, and stopping at the cap linked the new entry into the cycle.

Stripe::dir_check() had no cap at all and wrote chain_tag[] once per step, so a cycle ran off the end of a 65536-entry stack array. It now stops at the bound, and at the first revisited entry rather than printing a cycle line per step. check_bucket_not_contains(), an unbounded walk with no caller since the 2009 import, is deleted.


src/iocore/cache/CacheDir.cc
```diff
@@ -567,7 +548,15 @@ Directory::insert(const CacheKey *key, StripeSM *stripe, Dir *to_part)
...
     last = next_dir(last, seg);
-  } while (last && (++l <= this->buckets * DIR_DEPTH));
+    // Past the longest legitimate chain, so this is provably a cycle. This is the only walk insert() makes, so it is
+    // where a writer meets one: repair and start over rather than linking into the cycle.
+    if (++l > this->max_bucket_depth()) {
+      if (this->bucket_loop_fix(b, s)) {
+        goto Lagain;
+      }
+      break;
+    }
```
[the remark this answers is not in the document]
masaori335: The loop concern doesn't hold, and the suggested fix would cost more than the case it addresses.

bucket_loop_fix() returns 0 only when dir_bucket_loop_check() (Floyd's, CacheDir.cc:179) proves the chain acyclic — it's exact, not a heuristic that can "fail". And even setting that aside, the fall-through does dir_set_next(e, 0); dir_set_next(prev, dir_to_offset(e, seg)), which repoints prev at a chain-terminating entry. That severs a cycle at prev; it cannot link into one.

So the break path is: a chain longer than (DIR_DEPTH - 1) * buckets + 1 that is genuinely acyclic. That needs corrupted next offsets pulling other buckets' row-0 heads into the chain, since the longest acyclic path in a segment is DIR_DEPTH * buckets — a window of buckets - 1 entries above the bound. The consequence is that the tail past prev is orphaned: off the chain, off the freelist, reclaimed at the next init_segment(). No hang, no cycle, no wrong data.

Forcing init_segment() there would discard up to DIR_DEPTH * buckets valid entries to reclaim at most buckets - 1 orphans, on a segment we just proved has no loop. Not a trade I want to make on a path that is already corruption-only.

For context, the link-after-cap behavior is not new — the previous code was while (last && (++l <= this->buckets * DIR_DEPTH)), same fall-through, without the repair attempt this PR adds.

src/iocore/cache/CacheDir.cc
```diff
@@ -683,21 +669,17 @@ Directory::remove(const CacheKey *key, StripeSM *stripe, Dir *del)
...
+    int loop_count = 0;
     do {
-#ifdef LOOP_CHECK_MODE
-      loop_count++;
-      if (loop_count > DIR_LOOP_THRESHOLD) {
-        if (this->bucket_loop_fix(dir_bucket(b, seg), s))
-          return 0;
+      // Past the longest legitimate chain, so this is provably a cycle.
+      if (++loop_count > this->max_bucket_depth()) {
+        this->bucket_loop_fix(dir_bucket(b, seg), s);
```
bneradt, line 680: [P2] Mark the directory dirty when repairing a cycle

If `header->dirty` is 0 when this path runs, `bucket_loop_fix()` clears the segment through `init_segment()`, but neither helper sets the dirty flag, and this return bypasses `delete_entry()` (which normally sets it). The new `overwrite(..., must_overwrite=true)` repair path has the same problem: after resetting the segment it retries and returns 0 before `Lfill`. `CacheSync::mainEvent()` skips directories whose dirty flag is clear, so without a subsequent mutation the repair is never periodically persisted; an already-persisted corrupt chain can return after a crash. Please mark successful repairs dirty, preferably in the shared repair helper, and cover these paths starting with `header->dirty = 0`.

bneradt (commented):
Reviewed the bounded walks, segment reset paths, and updated unit tests. One persistence issue noted inline. All reported PR checks pass; local tests were not run because this worktree has no configured build.

src/iocore/cache/CacheDir.cc
masaori335: Good catch! b2763ee9b00c1e42a52c9985f2131c65e84c2d4a marks the flag in `Directory::init_segment`.

src/iocore/cache/CacheDir.cc
```diff
@@ -683,21 +671,17 @@ Directory::remove(const CacheKey *key, StripeSM *stripe, Dir *del)
...
```
bneradt: [P2] Account for discarded entries when repairing a segment

Confirming the gauge issue mentioned in the automated review summary: this newly enabled production path calls bucket_loop_fix(), which wipes the entire segment through init_segment(), but neither helper decrements cache_rsb.direntries_used or stripe->cache_vol->vol_rsb.direntries_used. For example, repairing a segment containing five counted entries leaves both gauges five too high; a subsequent insert increments them again even though only one entry remains. Ordinary removals cannot correct the error because the discarded entries no longer exist. Please account for the occupied entries discarded by runtime repairs in both gauges, and extend the repair tests to assert the before/after counts. Count physical entries rather than walking the corrupt chain.

bneradt (commented):
Follow-up review of b2763ee9b00c1e42a52c9985f2131c65e84c2d4a: the earlier persistence finding is addressed. init_segment() now marks the directory dirty, and the new tests check the remove and must-overwrite early-return paths as well as the no-repair case.

One remaining P2 issue is confirmed inline: repair leaves the used-entry gauges inflated (also mentioned in the automated review summary). I found no additional correctness findings in the traversal caps. The maximum valid chain is 49,153 entries at the maximum segment size, below the 65,536-element diagnostic array.

All 15 reported CI checks pass. This was a source review; I did not run the cache unit tests locally because this worktree has no configured ATS build.

src/iocore/cache/CacheDir.cc
masaori335: Fixed by [8c29ff0](https://github.com/apache/trafficserver/pull/13608/commits/8c29ff0b16aad9cffdcf16fc11b954e5be63bb53)

bneradt (commented):
Follow-up review of 8c29ff0: the earlier persistence and used-entry gauge findings are addressed. Runtime segment resets count physical occupied rows, debit both gauges, and mark the directory dirty; the tests now check these effects and exercise dir_check(). All 14 reported CI checks pass. One remaining test reliability issue is noted inline. The automated concern about a null head offset applies specifically to bucket 0, not every bucket; abandoning an already-corrupt acyclic chain in cleanup is not an additional blocker for this bounded-walk change. Source review only; no local ATS build or test run was performed.

src/iocore/cache/unit_tests/test_CacheDir.cc
```diff
@@ -164,78 +164,157 @@ class CacheDirTest : public CacheInit
...
+      }
+      return used;
+    };
+
+    // Give a bucket a fresh chain long enough that a walk has to leave the bucket's own rows, then loop it.
+    auto corrupt_fresh_chain = [&]() {
+      for (int i = 0; i < 5; i++) {
+        stripe->directory.insert(&key, stripe, &dir1);
+      }
+      dir_corrupt_bucket(dir_bucket(b1, stripe->directory.get_segment(s1)), s1, stripe);
```
bneradt, line 186: [P2] Make the corruption helper reliably create a cycle

When the randomly selected b1 is 0 and dir_corrupt_bucket() chooses l == 0, e is the segment base. dir_to_offset(e, seg) is then 0, the null-link sentinel, so dir_set_next(e, 0) truncates the chain instead of creating a cycle. The new CHECK(!directory.check()), bucket_length == -1, and repair/dirty assertions can therefore fail intermittently. Other bucket heads have nonzero offsets, so this is specifically the bucket-0 case (not a 20% failure rate across all buckets). Please select a non-head entry for the self-loop, or deterministically loop the second entry; the helper always inserts enough entries for that.

PR #13612, opened 2026-09-01
clang-tidy: add low-churn checks

bneradt:
These checks catch avoidable copies, blocked moves, narrow loop
variables, and unclear ownership transfers. This patch fixes the current
findings and enables the checks in CI and editor configuration.


src/iocore/net/SSLSecret.cc
```diff
@@ -88,8 +88,8 @@ std::string
 SSLSecret::loadFile(const std::string &name)
 {
   Dbg(dbg_ctl_ssl_secret, "SSLSecret::loadFile(%s)", name.c_str());
-  std::error_code   error;
-  std::string const data = swoc::file::load(swoc::file::path(name), error);
+  std::error_code error;
+  std::string     data = swoc::file::load(swoc::file::path(name), error);
```
bneradt, line 92: This is unintuitive, but the local const prevents automatic move when NRVO is not performed: returning it produces a const std::string&&, which cannot bind to std::string’s move constructor and therefore requires a copy. Since data is not otherwise modified and its lifetime ends at the return, removing const preserves the function’s semantics while allowing the fallback move. An explicit std::move would inhibit NRVO and would not move from a const string anyway.

cmcfarlen (dismissed):

cmcfarlen (approved):

bneradt:
[approve ci clang-analyzer]

